PR #3, opened 2020-11-18
ECF-NONE: Add govuk-design-system-formbuilder and try it out

ethanmills:
### Context
Not one to be merged in, but just learning how the formbuilder works.

Added a toy model and controller, just to be able to play with the form. The validation isn't quite right (if the date is invalid, the other fields are blanked), but that wasn't what I was focussing on here.

### UI changes
image
image
image

### Guidance to review
The view for new is the big interest here, as that's where all the formbuilder goodness is.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,3 +62,7 @@ end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw jruby]
+
+gem 'wdm', '>= 0.1.0' if Gem.win_platform?
+
+gem "govuk_design_system_formbuilder", "~> 2.1"
--- a/Gemfile
+++ b/Gemfile
@@ -57,6 +57,7 @@ group :development do
 end
 
 group :test do
+  gem 'rails-controller-testing'
   gem "webdrivers", "~> 4.4"
 end
 
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -1,0 +1,38 @@
+class SchoolsController < ApplicationController
+  def show
+    @school = School.find(params[:id])
+  end
+
+  def new; end
+
+  def create
+    validate_opened_date
+    return if performed?
+
+    @school = School.new(params.require(:school).permit(:name, :opened, :school_type))
+
+    if @school&.save
+      redirect_to @school
+    else
+      render "new"
+    end
+  end
+
+  private
+  def validate_opened_date
+    unless Date.valid_date?(*opened_date_parts)
+      @school = School.new
+      @school.errors.add(:opened, 'date validation error text goes here')
+      render "new"
+    end
+  end
+
+  def opened_date_parts
+
+    [
+      params.dig(:school, :'opened(1i)').to_i,
+      params.dig(:school, :'opened(2i)').to_i,
+      params.dig(:school, :'opened(3i)').to_i,
+    ]
+  end
+end
--- a/app/views/schools/new.html.erb
+++ b/app/views/schools/new.html.erb
@@ -1,0 +1,22 @@
+<h1 class="govuk-heading-xl">New school</h1>
+
+<%= form_with model: @school ||= School.new, url: schools_path, local: true do |form| %>
+  <% if @school&.errors&.any? %>
+    <%= form.govuk_error_summary %>
+  <% end %>
+  <%= form.govuk_text_field :name %>
+
+  <%= form.govuk_date_field :opened,
+                            legend: { text: "Enter the date the school was opened" },
+                            hint: { text: "For example, 31 3 1980" }
+  %>
+
+  <%= form.govuk_collection_radio_buttons :school_type,
+                                          School::SCHOOL_TYPES,
+                                          ->(option) { option },
+                                          ->(option) { option },
+                                          legend: { text: "What type of school is it?" }
+  %>
+
+  <%= form.submit %>
+<% end %>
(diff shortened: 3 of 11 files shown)


Gemfile
```diff
@@ -57,8 +57,13 @@ group :development do
...
 
 group :test do
+  gem 'rails-controller-testing'
   gem "webdrivers", "~> 4.4"
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw jruby]
+
+gem 'wdm', '>= 0.1.0' if Gem.win_platform?
```
kicferk1, line 67: Interesting, why do we need it? Is Windows an inferior platform? :D

app/controllers/schools_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      render "new"
+    end
+  end
+
+  private
+  def validate_opened_date
+    unless Date.valid_date?(*opened_date_parts)
+      @school = School.new
+      @school.errors.add(:opened, 'date validation error text goes here')
+      render "new"
```
kicferk1, line 26: Probably not the pattern we want to stick with - I am pretty sure this will wipe all the user-entered values.

I think a better idea would be to always make a new school in `create` and add a validator to the model checking for date being correct. Arguably that's where validation belongs in MVC anyway :) 

As a side benefit it will let you get rid of
```
return if performed?
```
line.

app/views/schools/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<h1 class="govuk-heading-xl">New school</h1>
+
+<%= form_with model: @school ||= School.new, url: schools_path, local: true do |form| %>
```
kicferk1, line 3: What does `local` do?

app/views/schools/new.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <%= form.govuk_date_field :opened,
+                            legend: { text: "Enter the date the school was opened" },
+                            hint: { text: "For example, 31 3 1980" }
+  %>
+
+  <%= form.govuk_collection_radio_buttons :school_type,
+                                          School::SCHOOL_TYPES,
+                                          ->(option) { option },
+                                          ->(option) { option },
+                                          legend: { text: "What type of school is it?" }
```
kicferk1, line 18: I imagine the two functions are one for value and one for text that option shows?

kicferk1: Also, the massive white space, uuh, can't say I'm a fan :P But I appreciate it might be rubocop making you do it.

spec/requests/schools_request_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      post "/schools", params: { school:
+                                   { name: "",
+                                     "opened(1i)": "2020",
+                                     "opened(2i)": "1",
+                                     "opened(3i)": "1",
+                                   } }
+
+      expect(response).to have_http_status(:success)
+      expect(response).to render_template(:new)
+      expect(assigns(:school).errors.empty?).to be false
```
kicferk1, line 54: Might also be worth testing the total number of schools to be 0?

spec/requests/schools_request_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    [
+      [2020, 2, 30],
+      [2020, 1, 32],
+    ].each do |invalid_date|
```
kicferk1, line 73: Ooooh, sweet :O Parametrised tests <3

kicferk1 (commented):
Looks pretty good to me, nice job. So nice I merged and unmerged it xD

Gemfile
ethanmills: Because it told me to basically: 
Please add the following to your Gemfile to avoid polling for changes:
    gem 'wdm', '>= 0.1.0' if Gem.win_platform?

It was red and ugly and this made it go away 😛

app/views/schools/new.html.erb
ethanmills: Prevents the form being submitted ajax

app/views/schools/new.html.erb
ethanmills: > I imagine the two functions are one for value and one for text that option shows?

Yeah, that's right. For something more complex, we might want something more like the example in [the docs ](https://govuk-form-builder.netlify.app/form-elements/radios/):
```
departments = [
  OpenStruct.new(id: 1, name: 'Sales'),
  OpenStruct.new(id: 2, name: 'Marketing'),
  OpenStruct.new(id: 3, name: 'Finance')
]
```
and
```
<%= f.govuk_collection_radio_buttons :new_department_id, departments, :id, :name, legend: { text: 'Which department do you work for?' }, hint: { text: 'There should be a sign near your desk' } %>
```

ethanmills: As for the whitespace, we could have something like
```
  <%= form.govuk_collection_radio_buttons(
        :school_type,
        School::SCHOOL_TYPES,
        ->(option) { option },
        ->(option) { option },
        legend: { text: "What type of school is it?" }
      )
  %>
```

spec/requests/schools_request_spec.rb
ethanmills: I would prefer something like 
```
    it "will not create a school without a name" do
      expect do

        post "/schools", params: { school:
                                     { name: "",
                                       "opened(1i)": "2020",
                                       "opened(2i)": "1",
                                       "opened(3i)": "1",
                                     } }

        expect(response).to have_http_status(:success)
        expect(response).to render_template(:new)
        expect(assigns(:school).errors.empty?).to be false
      end.not_to(change{School.count})
    end
```
but it's a really good point that it's something we should test!

PR #5, opened 2020-11-18
ECF-None: Formbuilder

kicferk1:
### Context
Copy of #3 
Not one to be merged in, but just learning how the formbuilder works.

Added a toy model and controller, just to be able to play with the form. The validation isn't quite right (if the date is invalid, the other fields are blanked), but that wasn't what I was focussing on here.

### UI changes
image
image
image

### Guidance to review
The view for new is the big interest here, as that's where all the formbuilder goodness is.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,3 +63,7 @@ end
 
 # do not rely on host’s timezone data, which can be inconsistent
 gem "tzinfo-data"
+
+gem 'wdm', '>= 0.1.0' if Gem.win_platform?
+
+gem "govuk_design_system_formbuilder", "~> 2.1"
--- a/Gemfile
+++ b/Gemfile
@@ -58,6 +58,7 @@ group :development do
 end
 
 group :test do
+  gem 'rails-controller-testing'
   gem "webdrivers", "~> 4.4"
 end
 
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -9,9 +9,9 @@ class SchoolsController < ApplicationController
     validate_opened_date
     return if performed?
 
-    @school = School.new(params.require(:school).permit(:name, :opened, :school_type))
+    @school = School.new(school_parameters)
 
-    if @school&.save
+    if @school.save
       redirect_to @school
     else
       render :new
@@ -20,10 +20,14 @@ class SchoolsController < ApplicationController
 
 private
 
+  def school_parameters
+    params.require(:school).permit(:name, :opened_at, :school_type)
+  end
+
   def validate_opened_date
     unless Date.valid_date?(*opened_date_parts)
       @school = School.new(params.require(:school).permit(:name, :school_type))
-      @school.errors.add(:opened, 'date validation error text goes here')
+      @school.errors.add(:opened_at, "date validation error text goes here")
       render :new
     end
   end
@@ -31,9 +35,9 @@ private
   def opened_date_parts
 
     [
-      params.dig(:school, :'opened(1i)').to_i,
-      params.dig(:school, :'opened(2i)').to_i,
-      params.dig(:school, :'opened(3i)').to_i,
+      params.dig(:school, :"opened_at(1i)").to_i,
+      params.dig(:school, :"opened_at(2i)").to_i,
+      params.dig(:school, :"opened_at(3i)").to_i,
     ]
   end
 end
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -1,0 +1,38 @@
+class SchoolsController < ApplicationController
+  def show
+    @school = School.find(params[:id])
+  end
+
+  def new; end
+
+  def create
+    validate_opened_date
+    return if performed?
+
+    @school = School.new(params.require(:school).permit(:name, :opened, :school_type))
+
+    if @school&.save
+      redirect_to @school
+    else
+      render "new"
+    end
+  end
+
+  private
+  def validate_opened_date
+    unless Date.valid_date?(*opened_date_parts)
+      @school = School.new
+      @school.errors.add(:opened, 'date validation error text goes here')
+      render "new"
+    end
+  end
+
+  def opened_date_parts
+
+    [
+      params.dig(:school, :'opened(1i)').to_i,
+      params.dig(:school, :'opened(2i)').to_i,
+      params.dig(:school, :'opened(3i)').to_i,
+    ]
+  end
+end
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -14,16 +14,17 @@ class SchoolsController < ApplicationController
     if @school&.save
       redirect_to @school
     else
-      render "new"
+      render :new
     end
   end
 
-  private
+private
+
   def validate_opened_date
     unless Date.valid_date?(*opened_date_parts)
-      @school = School.new
+      @school = School.new(params.require(:school).permit(:name, :school_type))
       @school.errors.add(:opened, 'date validation error text goes here')
-      render "new"
+      render :new
     end
   end
 
--- a/app/views/schools/new.html.erb
+++ b/app/views/schools/new.html.erb
@@ -6,17 +6,21 @@
   <% end %>
   <%= form.govuk_text_field :name %>
 
-  <%= form.govuk_date_field :opened,
-                            legend: { text: "Enter the date the school was opened" },
-                            hint: { text: "For example, 31 3 1980" }
+  <%= form.govuk_date_field(
+        :opened,
+        legend: { text: "Enter the date the school was opened" },
+        hint: { text: "For example, 31 3 1980" }
+      )
   %>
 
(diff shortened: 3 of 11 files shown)


app/controllers/schools_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      render "new"
+    end
+  end
+
+  private
+  def validate_opened_date
+    unless Date.valid_date?(*opened_date_parts)
+      @school = School.new
+      @school.errors.add(:opened, 'date validation error text goes here')
+      render "new"
```
kicferk1, line 26: Copied from before:
Probably not the pattern we want to stick with - I am pretty sure this will wipe all the user-entered values.

I think a better idea would be to always make a new school in `create` with all the parameters, and add a validator to the model checking for date being correct. Arguably that's where validation belongs in MVC anyway :)
Edit: Ah, but our model expects a date. Much fun. Well, I guess we would need to add custom date validation somewhere in controller then. Sadness :(

As a side benefit it will let you get rid of
```
return if performed?
```
line.

app/views/schools/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<h1 class="govuk-heading-xl">New school</h1>
+
+<%= form_with model: @school ||= School.new, url: schools_path, local: true do |form| %>
```
kicferk1, line 3: We will likely get rid of `local: true` once we cut out turbolinks and rails-ujs.

app/views/schools/new.html.erb
```diff
@@ -0,0 +1,22 @@
...
+                            legend: { text: "Enter the date the school was opened" },
+                            hint: { text: "For example, 31 3 1980" }
+  %>
+
+  <%= form.govuk_collection_radio_buttons :school_type,
+                                          School::SCHOOL_TYPES,
+                                          ->(option) { option },
+                                          ->(option) { option },
+                                          legend: { text: "What type of school is it?" }
+  %>
```
kicferk1, line 19: Something we should discuss and stick to: What whitespace format do we want? 
1. Long single-line
2. Multiline (like here)
3. Multiline with parentheses: 
```
  <%= form.govuk_collection_radio_buttons(
        :school_type, 
        School::SCHOOL_TYPES,
        ->(option) { option },
        ->(option) { option },
        legend: { text: "What type of school is it?" }
      )
  %>
```
We should probably decide which multiline to use (sometimes single line makes sense imo), and stick to it. I think I like 3 more than 2, because it prevents indentation depending on length of form builder method we want to use.

spec/requests/schools_request_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      post "/schools", params: { school:
+                                   { name: expected_name,
+                                     "opened(1i)": expected_date.year.to_s,
+                                     "opened(2i)": expected_date.month.to_s,
+                                     "opened(3i)": expected_date.day.to_s,
+                                   } }
+
+      expect(response).to redirect_to assigns(:school)
+      expect(School.find_by(name: expected_name)).not_to be_nil
+      expect(School.find_by(name: expected_name).opened).to eq(expected_date)
```
kicferk1, line 20: As Ethan pointed out,
```
expect do
(...)
end.not_to(change{School.count})
```
would be a good addition :)

app/views/schools/new.html.erb
```diff
@@ -0,0 +1,22 @@
...
+
+  <%= form.submit %>
```
kicferk1, line 21: There is a `form.govuk_submit` :O

ethanmills: Oops, yeah!

app/controllers/schools_controller.rb
ethanmills: I've made this a bit better, but suspect what I did doesn't scale well. Very keen to hear how other people would do this

Gemfile
```diff
@@ -58,8 +58,13 @@ group :development do
...
+  gem 'rails-controller-testing'
   gem "webdrivers", "~> 4.4"
 end
 
 # do not rely on host’s timezone data, which can be inconsistent
 gem "tzinfo-data"
+
+gem 'wdm', '>= 0.1.0' if Gem.win_platform?
+
+gem "govuk_design_system_formbuilder", "~> 2.1"
```
amazimbe, line 70: I wonder if the team has strong opinions on when to use double or single quotes. DfE probably don't mind either otherwise they would have set up a rule in rubocop

ethanmills: I have a rubocop rule enforcing double quotes. Not sure whether that came from DfE or elsewhere though

ethanmills: Ah, here https://github.com/alphagov/rubocop-govuk/blob/master/config/style.yml#L52

app/controllers/schools_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    @school = School.find(params[:id])
+  end
+
+  def new; end
+
+  def create
+    validate_opened_date
+    return if performed?
+
+    @school = School.new(params.require(:school).permit(:name, :opened, :school_type))
```
amazimbe, line 12: Better to move params.require(:school).permit(:name, :opened, :school_type) to a private method. It would change a lot as more school attributes are added. Also, that way it could be reused for another action.

app/controllers/schools_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    if @school&.save
```
amazimbe, line 14: I doubt the & here is necessary

ethanmills: Good spot 🙂

app/controllers/schools_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+      redirect_to @school
+    else
+      render :new
+    end
+  end
+
+private
+
+  def validate_opened_date
```
amazimbe, line 23: Definitely better to define a custom validation in the model. It looks a bit out of place in here.

ethanmills: I tried that, but couldn't quite get it, since the date seemed to be cast before model validation ran. How would you do that?

config/routes.rb
```diff
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
   get "/pages/:page", to: "pages#show"
 
+  resources :schools
```
amazimbe, line 4: Since the controller only supports show, new and create, for security reasons I would whitelist the supported actions here.

db/migrate/20201116191346_create_schools.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSchools < ActiveRecord::Migration[6.0]
+  def change
+    create_table :schools do |t|
+      t.timestamps
+      t.column :name, :string, null: false
+      t.column :opened, :date
```
amazimbe, line 6: We can just follow the rails convention and call the date field opened_at

spec/requests/schools_request_spec.rb
```diff
@@ -0,0 +1,107 @@
+require 'rails_helper'
+
+RSpec.describe "Schools", type: :request do
+
+  describe "Creating a school" do
+    it "successfully creates a school" do
+      # Given
+      expected_name = "Test School"
```
amazimbe, line 8: Using let(:expected_name) { "Test School" } etc above the it would be better.

spec/requests/schools_request_spec.rb
```diff
@@ -0,0 +1,107 @@
+require 'rails_helper'
+
+RSpec.describe "Schools", type: :request do
+
+  describe "Creating a school" do
```
amazimbe, line 5: The convention is to add the url endpoint as part of this description. It gives helpful info about the routes etc so this would be along the lines of: describe 'Creating a school - POST /schools'

spec/requests/schools_request_spec.rb
ethanmills: Ah good to know. Why is this better?

amazimbe: The let blocks are lazily instantiated so objects tend to just get created on demand at run time. And these testing variables tend to be applicable to multiple tests as well. With let blocks you don't duplicate them for every spec. The overall result is that the specs look less cluttered and are much easier to read and understand. https://www.betterspecs.org/#let has some good guidelines for writing ruby specs.

app/controllers/schools_controller.rb
amazimbe: Missed this yesterday. I'll take a look this morning or perhaps we can pair via google meets or something

db/migrate/20201116191346_create_schools.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSchools < ActiveRecord::Migration[6.0]
+  def change
+    create_table :schools do |t|
+      t.timestamps
+      t.column :name, :string, null: false
+      t.column :opened_at, :date
```
amazimbe, line 6: Needs null: false too given the validation in the model

PR #6, opened 2020-11-19
Supplier dashboard from prototype

kicferk1:
### Context
The prototype in question is here: https://ecf-proto-supplier.herokuapp.com/dashboard_sup

I think even if we will want some copy changes or to remove some tabs, we will probably keep that page in one form or another.

I added a "new service" banner below the header for every page we make, because we probably want to display it everywhere.

Added a minimal endpoint and controller for displaying the page.

The page itself: 
image

### Guidance to review
I will probably add some tests for the endpoint, `Rspec.feature` was it?

Also, do we want to internationalise our strings?

The urls from the dashboard point to nowhere, I think before we start adding any mroe pages we want the form builder stuff (#5) in.


app/views/layouts/application.html.erb
```diff
@@ -61,7 +61,20 @@
     </header>
 
     <div class="govuk-width-container">
-      <main class="govuk-main-wrapper " id="main-content" role="main">
+      <div class="govuk-phase-banner app-phase-banner--no-border govuk-!-display-none-print">
+        <p class="govuk-phase-banner__content">
+          <strong class="govuk-tag govuk-phase-banner__content__tag ">
+            Alpha
```
ethanmills, line 67: Our favourite reference project has a Development banner, would that be more appropriate? I think eventually we will need a different banner depending on the hosting environment

kicferk1: Truth be told I just copied text from the prototype :D 

I imagine this will likely be changed to Beta soon, and yeah, we might add environment into the mix when we deploy it and have multiple environments. I'm happy to switch it to development for now :)

spec/features/view_pages_spec.rb
```diff
@@ -6,4 +6,10 @@
 
     expect(page).to have_text("Lorem")
   end
+
```
amazimbe, line 9: We definitely want request specs rather than feature specs. They are a pain. Let the QA guy do the feature specs as they will just slow us down. They tend to be very brittle and time consuming.

kicferk1: Thanks, request tests look great!

config/routes.rb
```diff
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   get "/pages/:page", to: "pages#show"
+  get "/supplier_dashboard", to: "suppliers#dashboard"
```
amazimbe, line 3: This looks like a singleton resource rather than a static page to me. https://guides.rubyonrails.org/routing.html#singular-resources

 If you do resource : supplier_dashboard, only: :index  you can get rid of this unrestful route.

Also, the dashboard would fetch all sorts of data and I don't think you want that in the suppliers controller. A SupplierDashboardController would probably be a more suitable place for this. The index action would just conventionally render the right view rather than having to do render template: "pages/supplier_dashboard"

kicferk1: I'm not sure which of `index` vs `show` to use. 

I think `index` is meant to show a list of resources, which is probably not what we are doing here. But admittedly I didn't use singular resources much, so I'm not too familiar with the standard conventions.

amazimbe (approved):

PR #8, opened 2020-11-19
ECF-30: Add formbuilder

ethanmills:
### Context
Adding some dependencies from #5. We don't want to merge that, so splitting out the useful bits here

Add:
- govuk_design_system_formbuilder
- rails-controller-testing
- wdm

### Guidance to review
None, really

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,6 +57,7 @@ group :development do
 end
 
 group :test do
+  gem "rails-controller-testing"
   gem "webdrivers", "~> 4.4"
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -62,3 +62,7 @@ end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw jruby]
+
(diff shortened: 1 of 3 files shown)


kicferk1:
While we are at it, can we remove turbolinks and rails-ujs?

ethanmills:
> While we are at it, can we remove turbolinks and rails-ujs?

#7 😉

Gemfile
```diff
@@ -58,8 +58,13 @@ group :development do
 end
 
 group :test do
+  gem "rails-controller-testing"
```
kicferk1, line 61: Hmm, do we want it? What does it give us exactly?

kicferk1: Oh, I just ran into it on a different PR. Lets us check what was rendered. 

Let's keep it :)

kicferk1 (approved):
LGTM

Gemfile
ethanmills: Everywhere you see `assigns` or `render_template` in #5, that's this gem

ethanmills: Perhaps we want to test in a different way though

amazimbe (approved):

cfq (approved):

PR #9, opened 2020-11-20
EF-32: Replicate invite schools page from prototype

kicferk1:
### Context
Another page to replicate from the prototype: https://ecf-proto-supplier.herokuapp.com/new

Add some lovely view helper methods stolen from `apply for teacher training`. 

Use them and govuk form builder to make a neat copy of prototype page.

Add a form (ActiveModel, idea stolen from `apply...`), so we can validate the input without database model.

Test the above thoroughly, more for practice than because it needs to.

The look:
image

### Guidance to review

Not much happens when we save the page - cause, well the next page doesn't exist yet.

I added a gem I stole from `apply...` to test validations. The tests for the form might be an overkill (feel free to disagree), but the gem is probably a good idea to keep.

### DIFF
--- a/app/controllers/invite_schools_controller.rb
+++ b/app/controllers/invite_schools_controller.rb
@@ -1,0 +1,20 @@
+class InviteSchoolsController < ApplicationController
+  def show
+    @find_school_form = FindSchoolForm.new
+  end
+
+  def create
+    @find_school_form = FindSchoolForm.new(*find_school_form_params)
+    if @find_school_form.valid?
+      redirect_to supplier_dashboard_path
+    else
+      render :show
+    end
+  end
+
+private
+
+  def find_school_form_params
+    params.require(:find_school_form).permit(%i[search_type name_url_postcode local_authority network geography])
+  end
+end
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -1,0 +1,47 @@
+module ViewHelper
+  def govuk_link_to(body, url, html_options = {}, &_block)
+    html_options[:class] = prepend_css_class("govuk-link", html_options[:class])
+
+    return link_to(url, html_options) { yield } if block_given?
+
+    link_to(body, url, html_options)
+  end
+
+  def govuk_back_link_to(url = :back, body = "Back")
+    classes = "govuk-back-link"
+
+    if url == :back
+      url = controller.request.env["HTTP_REFERER"] || "javascript:history.back()"
+      classes += " app-back-link--fallback"
+    end
+
+    if url == "javascript:history.back()"
+      classes += " app-back-link--no-js"
+    end
+
+    link_to(body, url, class: classes)
+  end
+
+  def govuk_button_link_to(body, url, html_options = {}, &_block)
+    html_options = {
+      class: prepend_css_class("govuk-button", html_options[:class]),
+      role: "button",
+      data: { module: "govuk-button" },
+      draggable: false,
+    }.merge(html_options)
+
+    return link_to(url, html_options) { yield } if block_given?
+
+    link_to(body, url, html_options)
+  end
+
+private
+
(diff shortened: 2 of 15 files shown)


spec/requests/invite_schools_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "rails_helper"
+
+RSpec.describe "Invite schools", type: :request do
+  it "renders the school search page" do
+    get "/invite_schools"
+    expect(response).to render_template(:show)
+  end
+
+  it "re-renders the school search page when given invalid request" do
+    get "/invite_schools"
```
amazimbe, line 10: Do we need this?

amazimbe, line 10: Do we need this line

app/controllers/invite_schools_controller.rb
```diff
@@ -0,0 +1,20 @@
+class InviteSchoolsController < ApplicationController
```
amazimbe, line 1: 👍 but I think SchoolInvitesController might be a better name. It certainly makes more sense in the routes file where you declare the resource.

config/routes.rb
```diff
@@ -2,6 +2,7 @@
   get "/pages/:page", to: "pages#show"
 
   resource :supplier_dashboard, controller: :supplier_dashboard, only: :show
+  resource :invite_schools, only: %i[show create]
```
amazimbe, line 5: I think we want a noun as the resource and not a verb so probably school_invites

app/helpers/view_helper.rb
```diff
@@ -0,0 +1,47 @@
+module ViewHelper
```
amazimbe, line 1: I just wonder; if these apply for the whole application, and they look like they do, then perhaps they should go in application helper.

spec/forms/find_school_form_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe FindSchoolForm, type: :model, with_audited: true do
+  describe "validations" do
+    it { is_expected.to validate_presence_of(:search_type).with_message("Please select an option") }
```
ethanmills, line 5: Do we get useful error messages when these tests fail? I'd normally expect them to have a description

spec/requests/invite_schools_spec.rb
kicferk1: Nope :) Good spot!

app/helpers/view_helper.rb
kicferk1: I think the concerns of various link helpers shouldn't really be used outside views - they talk about css, html and js.

I'd start by keeping them in the view helper, and if/when we have a use case for using them in a wider scope, we can move them.

spec/forms/find_school_form_spec.rb
kicferk1: This is what you get when these fail:
```
     FindSchoolForm validations is expected to validate that :search_type cannot be empty/falsy, producing a custom validation error on failure
     Failure/Error: it { is_expected.to validate_presence_of(:search_type).with_message("Please select an invalid option") }
     
       Expected FindSchoolForm to validate that :search_type cannot be
       empty/falsy, producing a custom validation error on failure, but this
       could not be proved.
         After setting :search_type to ‹nil›, the matcher expected the
         FindSchoolForm to be invalid and to produce the validation error
         "Please select an invalid option" on :search_type. The record was
         indeed invalid, but it produced these validation errors instead:
     
         * search_type: ["Please select an option"]
     # ./spec/forms/find_school_form_spec.rb:5:in `block (3 levels) in <main>'
```
It looks pretty readable to me.

ethanmills: Ah nice, looks good :)

ethanmills (approved):

amazimbe (approved):

PR #10, opened 2020-11-22
ECF-30 Some refactoring from the pairing session with Ethan

amazimbe:
### Context
Some refactoring from the pairing session with Ethan
Added additional spec
Some code cleanup


Gemfile
```diff
@@ -58,7 +58,8 @@ group :development do
 end
 
 group :test do
-  gem 'rails-controller-testing'
+  gem "rails-controller-testing"
+  gem "shoulda-matchers"
```
amazimbe, line 62: I added this for the convenient syntax it brings to testing models especially

app/controllers/schools_controller.rb
```diff
@@ -23,21 +22,4 @@ def create
   def school_parameters
     params.require(:school).permit(:name, :opened_at, :school_type)
   end
-
-  def validate_opened_date
```
kicferk1, line 27: Sooo, do we do any date validation? It's removed here, but I don't see it added anywhere?

ethanmills: Doesn't look like it, my validation tests fail on this branch

ethanmills (changes requested):

app/views/schools/new.html.erb
```diff
@@ -1,6 +1,6 @@
 <h1 class="govuk-heading-xl">New school</h1>
 
-<%= form_with model: @school ||= School.new, url: schools_path, local: true do |form| %>
+<%= form_with model: @school, url: schools_path, local: true do |form| %>
```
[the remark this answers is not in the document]
amazimbe: I think the url is needed: https://apidock.com/rails/ActionView/Helpers/FormHelper/form_with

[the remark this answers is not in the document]
kicferk1: Ah, yes, you are right :)

app/controllers/schools_controller.rb
amazimbe: The model validates presence of opened_at and rails automatically checks that it's a date. Personally, I'd just change the spec to check that the bad data results in an exception.

spec/rails_helper.rb
```diff
@@ -58,4 +58,7 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+
+  config.include(Shoulda::Matchers::ActiveModel, type: :model)
+  config.include(Shoulda::Matchers::ActiveRecord, type: :model)
```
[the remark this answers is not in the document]
amazimbe: Yes, I just wanted this on the models. It's also useful on controller specs but then we opted to go for request specs.

app/controllers/schools_controller.rb
ethanmills: But we don't want an unhandled exception. We want a validation message on the form, right? I don't think this does that...

spec/requests/schools_request_spec.rb
```diff
@@ -88,7 +88,7 @@
     end
 
     [
-      [2020, 2, 30],
+      [2020, 2, 32],
```
kicferk1, line 91: 30th of February should be invalid date too, right?

app/controllers/schools_controller.rb
```diff
@@ -1,14 +1,18 @@
+require 'byebug'
+
 class SchoolsController < ApplicationController
+  rescue_from ActiveRecord::MultiparameterAssignmentErrors,
```
amazimbe, line 4: @ethanmills @kicferk1 I just decided to handle the exception. It looks less messy than the controller validations in my view

ethanmills: What happens if we have two dates? This is a pattern we will need to use a lot I think

spec/requests/schools_request_spec.rb
amazimbe: Rails tries to be clever and automatically changes it to 1st March.

app/controllers/schools_controller.rb
amazimbe: We can worry about that when we get to it. We may get there and there are a few potential solutions.

amazimbe:
> Why do we need to change this? This is an invalid case, and this should fail. This was a valuable test before it was changed

Rails tries to be too clever with 30th February and automatically converts it to 1st March. 32 February is in the same equivalence class so I think it's Ok.

ethanmills:
> Rails tries to be too clever with 30th February and automatically converts it to 1st March. 32 February is in the same equivalence class so I think it's Ok.

But what happens if a user enters 2020/2/30?

kicferk1:
> > Rails tries to be too clever with 30th February and automatically converts it to 1st March. 32 February is in the same equivalence class so I think it's Ok.
> 
> But what happens if a user enters 2020/2/30?

I imagine https://stackoverflow.com/questions/597328/how-do-i-validate-a-date-in-rails/597776 :P 

Soo, everything passes, and the date saved into the database is 1st of March. Probably not what the user expected :(

amazimbe:
> > Rails tries to be too clever with 30th February and automatically converts it to 1st March. 32 February is in the same equivalence class so I think it's Ok.
> 
> But what happens if a user enters 2020/2/30

It will create the school with opened_at as 1st March.

ethanmills:
> will create the school with opened_at as 1st March.

I don't think  that's what we want. That's why I wrote a test to ensure that doesn't happen

amazimbe:
> > > Rails tries to be too clever with 30th February and automatically converts it to 1st March. 32 February is in the same equivalence class so I think it's Ok.
> > 
> > 
> > But what happens if a user enters 2020/2/30?
> 
> I imagine https://stackoverflow.com/questions/597328/how-do-i-validate-a-date-in-rails/597776 :P
> 
> Soo, everything passes, and the date saved into the database is 1st of March. Probably not what the user expected :(

It's not an unreasonable assumption

PR #11, opened 2020-11-23
Add automated linting

ethanmills:
### Context
Let's use github actions to lint our code. We can use this until we have more CI set up.

Add github actions to run rubocop and eslint on every push

### Guidance to review
A lot of the changes here are just fixing the linting, probably safe to ignore

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -44,53 +44,51 @@ RSpec.configure do |config|
   # triggering implicit auto-inclusion in groups with matching metadata.
   config.shared_context_metadata_behavior = :apply_to_host_groups
 
-# The settings below are suggested to provide a good initial experience
-# with RSpec, but feel free to customize to your heart's content.
-=begin
-  # This allows you to limit a spec run to individual examples or groups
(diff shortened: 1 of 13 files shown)


amazimbe (approved):

spec/spec_helper.rb
```diff
@@ -44,53 +44,51 @@
...
-
-  # Seed global randomization in this process using the `--seed` CLI option.
-  # Setting this allows you to use `--seed` to deterministically reproduce
-  # test failures related to randomization by passing the same `--seed` value
-  # as the one that triggered the failure.
-  Kernel.srand config.seed
-=end
+  # The settings below are suggested to provide a good initial experience
+  # with RSpec, but feel free to customize to your heart's content.
+  #   # This allows you to limit a spec run to individual examples or groups
```
kicferk1, line 49: That looks a bit weird (double comment?), but I guess it doesn't really matter.

kicferk1 (approved):
Looks good to me :)

spec/spec_helper.rb
ethanmills: Agreed, but 🤷

PR #12, opened 2020-11-24
ECF-93: Integrate with notify

kicferk1:
### Context
We will use notify for sending registration emails (probably forgetting password as well, and I am sure a bunch more of use cases). 

Install `mail-notify` gem.

Add `GOVUK_NOTIFY_API_KEY` to our environment.

Copy `action_mailer` config from `apply-for-teacher-training`.

Use the govuk mailer in code.

### Guidance to review
I will chat to Chris about whether we want a separate service to the one that prototype will use (they already registered one, but haven't used it yet). But changes should be limited to updating the key in the environment (not visible here), and maybe the template id.


.github/workflows/lint.yml
```diff
@@ -21,7 +21,11 @@ jobs:
           ruby-version: 2.7.1
       - name: Install dependencies
         run: bundle install
+      - name: Create mail log file
```
ethanmills, line 24: Do we need this to run the linter?

kicferk1: Yes :(

ethanmills (approved):
Looks great!

app/mailers/registration_mailer.rb
```diff
@@ -0,0 +1,16 @@
+class RegistrationMailer < ApplicationMailer
```
amazimbe, line 1: You need some mailer specs for this: https://relishapp.com/rspec/rspec-rails/docs/mailer-specs

config/environments/development.rb
```diff
@@ -12,6 +12,22 @@
   # Show full error reports.
   config.consider_all_requests_local = true
 
+  # Do not care if the mailer cannot send.
+  config.action_mailer.raise_delivery_errors = false
+  config.action_mailer.perform_caching = false
+  config.action_mailer.delivery_method = :notify
+  config.action_mailer.notify_settings = {
+    api_key: ENV.fetch("GOVUK_NOTIFY_API_KEY"),
```
amazimbe, line 20: We probably don't want to actually be sending emails in dev and testing. It just causes too many mishaps. There are many good solutions for this one of them being: https://github.com/ryanb/letter_opener . It just opens the email in the browser on localhost which is useful for debugging and eliminates the risk of sending emails from dev to actual users.

ethanmills: When you create the API key for notify, you can create a dev key, which will not send any emails

amazimbe: 👍

kicferk1: Yep, like Ethan said Notify is pretty good at managing it, and I do like actually sending emails, makes me feel safer that it will actually work in production :)

app/mailers/registration_mailer.rb
kicferk1: Soo, I added a simple test for checking who email is sent to and from. 

I am pretty certain writing one for email body and subject is counter-productive, because I opted to use notify email templates.

It is possible to store emails in our codebase (in erb files), but templates let non-technical people edit copy whenever they please, so long as we invite them to our notify space. Happy to discuss trade-offs.

ethanmills: Sounds good to me. We can always move to erb if it turns out to be a nightmare

amazimbe (approved):

PR #13, opened 2020-11-26
Add migrations and data model

kicferk1:
### Context
We want a simple data model to be able to do more work on it. In particular we need schools, lead providers and partnerships. We also added networks just because it is easy. 

Add migrations and models for the objects mentioned above.

Add factory bot.

Add tests for the models and relations between them.

### Guidance to review
Feel free to go commit by commit.

I'm not quite sure what the convention is for naming boolean methods (`confirmed?` vs `is_confirmed?`) and methods changing state of an object (`confirm` vs `confirm!`). Feel free to offer advice :)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,8 @@ group :development, :test do
   gem "capybara", "~> 3.33"
 
   gem "dotenv-rails"
+
+  gem "factory_bot_rails"
 end
 
 group :development do
--- a/README.md
+++ b/README.md
@@ -29,6 +29,9 @@
 It should be possible to run just the database from docker, if you want to.
 Check docker-compose file for username and password to put in your `.env` file.
 
+If you want to seed the database you can either run `db:drop` and `run:setup` tasks with your preferred method,
+or `db:seed`. 
+
 ### Govuk Notify
 Register on [GOV.UK Notify](https://www.notifications.service.gov.uk). 
 Ask someone from the team to add you to our service.
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@
 It should be possible to run just the database from docker, if you want to.
 Check docker-compose file for username and password to put in your `.env` file.
 
-If you want to seed the database you can either run `db:drop` and `run:setup` tasks with your preferred method,
+If you want to seed the database you can either run `db:drop` and `db:setup` tasks with your preferred method,
 or `db:seed`. 
 
 ### Govuk Notify
--- a/app/models/partnership.rb
+++ b/app/models/partnership.rb
@@ -3,11 +3,11 @@ class Partnership < ApplicationRecord
   belongs_to :lead_provider
 
   def confirmed?
-    !confirmed.nil?
+    confirmed_at.present?
   end
 
   def confirm
-    self.confirmed = Date.current
+    self.confirmed_at = Time.zone.now
     save!
   end
 end
--- a/app/models/partnership.rb
+++ b/app/models/partnership.rb
@@ -1,0 +1,13 @@
+class Partnership < ApplicationRecord
+  belongs_to :school
+  belongs_to :lead_provider
+
+  def confirmed?
+    !confirmed.nil?
+  end
+
+  def confirm
+    self.confirmed = Date.current
+    save!
+  end
+end
--- a/db/migrate/20201125151816_create_schools.rb
+++ b/db/migrate/20201125151816_create_schools.rb
@@ -1,0 +1,18 @@
+class CreateSchools < ActiveRecord::Migration[6.0]
+  def change
+    create_table :schools do |t|
+      t.timestamps
+      t.column :urn, :string, null: false
+      t.column :name, :string, null: false
+      t.column :address, :string, null: false
+      t.column :school_type, :string
+      t.column :capacity, :integer
+      t.column :high_pupil_premium, :boolean, null: false, default: false
+      t.column :is_rural, :boolean, null: false, default: false
+    end
+    add_index :schools, :urn, unique: true
+    add_index :schools, :name
+    add_index :schools, :high_pupil_premium, where: :high_pupil_premium
+    add_index :schools, :is_rural, where: :is_rural
+  end
+end
--- a/db/migrate/20201125151816_create_schools.rb
+++ b/db/migrate/20201125151816_create_schools.rb
@@ -1,6 +1,6 @@
 class CreateSchools < ActiveRecord::Migration[6.0]
   def change
-    create_table :schools do |t|
+    create_table :schools, id: :uuid do |t|
       t.timestamps
       t.column :urn, :string, null: false
       t.column :name, :string, null: false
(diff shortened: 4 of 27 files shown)


db/migrate/20201125151816_create_schools.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.column :name, :string, null: false
+      t.column :address, :string, null: false
+      t.column :school_type, :string
+      t.column :capacity, :integer
+      t.column :high_pupil_premium, :boolean, null: false, default: false
+      t.column :is_rural, :boolean, null: false, default: false
+    end
+    add_index :schools, :urn, unique: true
+    add_index :schools, :name
+    add_index :schools, :high_pupil_premium, where: :high_pupil_premium
```
ethanmills, line 15: I added these partial indices since I assume we're only interested in finding schools where this is true. Not particularly confident about this in general, so would be great to hear someone else's view

db/schema.rb
```diff
@@ -2,15 +2,60 @@
...
+    t.datetime "updated_at", precision: 6, null: false
+    t.string "name", null: false
+  end
+
+  create_table "partnerships", force: :cascade do |t|
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.bigint "school_id", null: false
+    t.bigint "lead_provider_id", null: false
+    t.date "confirmed"
```
amazimbe, line 35: confirmed_at would be more conventional and probably should be a datetime rather than a date for better precision

Gemfile
```diff
@@ -49,6 +49,8 @@ group :development, :test do
   gem "capybara", "~> 3.33"
 
   gem "dotenv-rails"
+
+  gem "factory_bot_rails"
```
amazimbe, line 53: 👍

app/models/partnership.rb
```diff
@@ -0,0 +1,13 @@
+class Partnership < ApplicationRecord
+  belongs_to :school
+  belongs_to :lead_provider
+
+  def confirmed?
+    !confirmed.nil?
```
amazimbe, line 6: Could be just me but I find double negations hard to understand, why not just confirmed.present?

amazimbe (commented):
Looks good but I've left some minor comments.

My biggest recommendation, especially for a greenfield project like this, would be to use uuids rather than ids. There are many reasons for this some of which are in this article: https://itnext.io/using-uuids-to-your-rails-6-application-6438f4eeafdf. What do you guys think?

db/migrate/20201125151816_create_schools.rb
amazimbe: Sounds like a great idea! 👍

kicferk1:
> Looks good but I've left some minor comments.
> 
> My biggest recommendation, especially for a greenfield project like this, would be to use uuids rather than ids. There are many reasons for this some of which are in this article: https://itnext.io/using-uuids-to-your-rails-6-application-6438f4eeafdf. What do you guys think?

I do like the idea of uuids, although they might lead to some unwieldy urls. Is that a concern at all? @cfq

db/seeds.rb
```diff
@@ -1,7 +1,11 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+# TODO: Remove network and school seeding when we have a way of getting them from GIAS
+
+network = Network.create!(name: "Test school network")
```
ethanmills, line 3: Is this idempotent with the `!`? We want to be able to run this multiple times

kicferk1: Hmmm, good point. I think `create` and `create!` only differ in what happens when validation fails.

So really, if we want them idempotent we would probably need to use `find_or_create_by`. But I am not sure we would want to run seeding multiple times without deleting the data in between - when would that happen?

README.md
```diff
@@ -29,6 +29,9 @@
 It should be possible to run just the database from docker, if you want to.
 Check docker-compose file for username and password to put in your `.env` file.
 
+If you want to seed the database you can either run `db:drop` and `run:setup` tasks with your preferred method,
```
amazimbe, line 32: Is that a typo run:setup -> db:setup

kicferk1: Good spot!

db/migrate/20201125151816_create_schools.rb
```diff
@@ -0,0 +1,18 @@
+class CreateSchools < ActiveRecord::Migration[6.0]
+  def change
+    create_table :schools, id: :uuid do |t|
+      t.timestamps
+      t.column :urn, :string, null: false
+      t.column :name, :string, null: false
+      t.column :address, :string, null: false
```
amazimbe, line 7: Not sure about this. I haven't seen many applications where storing the address as just one string is useful. Perhaps we need to discuss what we might use the address.

kicferk1: I am pretty sure most of our data model will need overhauling later down the line (I had a chat about that this morning with Chris, he and Brandon are working on a more fully-fledged entity model). We aren't using address for much atm, it's just there as a placeholder. Kind of like number of pupils, school type, and boolean flags.

spec/models/lead_provider_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "rails_helper"
+
+RSpec.describe LeadProvider, type: :model do
+  it "can be created" do
+    expect {
+      LeadProvider.create(name: "Test Lead Provider")
```
amazimbe, line 6: I kind of feel this is testing rails itself. Not sure it's strictly necessary but it does no harm.

spec/models/lead_provider_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    }.to change { LeadProvider.count }.by(1)
+  end
+
+  it "can have a partnership" do
+    lead_provider = LeadProvider.create!(name: "Test Lead Provider")
```
amazimbe, line 11: I think for these boilerplate tests for associations or validations the baked in matchers are better: https://github.com/thoughtbot/shoulda-matchers#usage . I mean this sort of thing: describe 'associations' do it {is_expected.to have_many (:blah)} end etc.

db/migrate/20201125151816_create_schools.rb
amazimbe: @kicferk1 of course most of the data model will change though probably having a separate address table with t.string "address_line1"
    t.string "address_line2"
    t.string "address_line3"
    t.string "address_line4"
    t.string "country"
    t.string "postcode" is probably a better starting point. Happy to approve. I'll leave the final call to you.

amazimbe (approved):

PR #17, opened 2020-12-07
[ECF-29] Implement passwordless login

amazimbe:

Passwordless authentication based on devise

### Guidance to review

 Only the changes required for login are in here. The registration changes will come in a separate pr shortly.


db/migrate/20201130093110_devise_create_users.rb
```diff
@@ -0,0 +1,43 @@
...
+      t.string :email,              null: false, default: ""
+      t.string :encrypted_password, null: false, default: ""
+
+      ## Passwordless authenticatable
+      t.string :login_token
+      t.datetime :login_token_valid_until
+
+      ## Recoverable
+      t.string   :reset_password_token
+      t.datetime :reset_password_sent_at
```
kicferk1, line 19: We probably don't need this.

db/migrate/20201130093110_devise_create_users.rb
```diff
@@ -0,0 +1,43 @@
...
+      ## Confirmable
+      t.string   :confirmation_token
+      t.datetime :confirmed_at
+      t.datetime :confirmation_sent_at
+
+      t.timestamps null: false
+    end
+
+    add_index :users, :email,                unique: true
+    add_index :users, :reset_password_token, unique: true
```
kicferk1, line 40: Or this.

lib/devise/strategies/passwordless_authenticatable.rb
```diff
@@ -0,0 +1,29 @@
...
+            login_token_valid_until: 60.minutes.from_now)
+
+            url = Rails.application.routes.url_helpers.email_confirmation_url(
+              login_token: user.login_token,
+              host: Rails.application.config.domain,
+            )
+
+            UserMailer.sign_in_email(user, url).deliver_now
+            fail!("An email was sent to you with a magic link.")
+          end
```
kicferk1, line 22: I find it a bit hard to understand what's going on here. 

If the user exists, we try to update their login token. And if that succeeds we send them an email, and we `fail!` with a confirmation message? That seems like a hack. If it's a standard way to do it, I would add a comment with link to devise documentation. 

And what happens when the user does not exist? Devise magic catches that afterwards? Beforehand? 

Can we do something like https://gist.github.com/r00k/906356 does? With `success!` method for the good case, and a `fail` for the bad case?

spec/mailers/user_mailer_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      login_token: user.login_token, host: Rails.application.config.domain,
+    )
+  end
+
+  describe "#sign_in_email" do
+    it "renders the right headers and content" do
+      expect(mail.subject).to eq("Sign in into ECF")
+      expect(mail.to).to eq([user.email])
+      expect(mail.from).to eq(["from@example.com"])
+      expect(mail.body.encoded).to include(sign_in_link)
```
kicferk1, line 18: Does that test pass now that you moved to use notify templates? It would be a little surprising to me - after all, how would our tests know the email body? Make an api call to notify, using a fake key, to get it? :/

amazimbe: Good catch! I've updated it

lib/devise/strategies/passwordless_authenticatable.rb
amazimbe: >> If the user exists, we try to update their login token. And if that succeeds we send them an email, and we fail! with a confirmation message? That seems like a hack.

The fail! is because we haven't logged in the user, we have just sent them the magic link. success! would try and set the cookie etc which is not what we need.

kicferk1: Alright, that makes sense.

We will probably revisit this sooner or later (when designers get their hands on the login flow I imagine), but it can probably stay this way for time being.

kicferk1:
So, just to make sure, do we want to fix registration in this PR? Or will we make another one?

amazimbe:
> So, just to make sure, do we want to fix registration in this PR? Or will we make another one?

No, I'm fixing registration in https://github.com/DFE-Digital/early-careers-framework/pull/18

kicferk1 (approved):

PR #18, opened 2020-12-07
[ECF-29] Implement registration

amazimbe:
User registrations and confirmations using Devise.


app/mailers/user_mailer.rb
```diff
@@ -13,4 +14,22 @@ def sign_in_email(user, url)
       },
     )
   end
+
+  def confirmation_instructions(user, token, _options = {})
```
kicferk1, line 18: Do we need the `_options` argument in here?

spec/mailers/user_mailer_spec.rb
```diff
@@ -1,19 +1,37 @@
...
-  let(:sign_in_link) do
-    Rails.application.routes.url_helpers.email_confirmation_url(
-      login_token: user.login_token, host: Rails.application.config.domain,
-    )
-  end
-
   describe "#sign_in_email" do
+    let(:sign_in_email) { UserMailer.sign_in_email(user, sign_in_link) }
+
+    let(:sign_in_link) do
```
kicferk1, line 9: Are we using this link for anything?

spec/mailers/user_mailer_spec.rb
```diff
@@ -1,19 +1,37 @@
...
+      expect(sign_in_email.from).to eq(["mail@example.com"])
+    end
+  end
+
+  describe "#confirmation_instructions" do
+    let(:confirmation_email) do
+      UserMailer.confirmation_instructions(user, confirmation_link)
+    end
+
+    let(:confirmation_link) do
```
kicferk1, line 26: Ditto, is this used?

config/initializers/devise.rb
```diff
@@ -17,10 +17,10 @@
   config.secret_key = Rails.application.secrets.secret_key_base
 
   # Configure the class responsible to send e-mails.
-  # config.mailer = "UserMailer"
+  config.mailer = "UserMailer"
```
kicferk1, line 20: Cool! Is that why we name a method in the mailer `confirmation_instructions`? Does it need to match what devise is expecting?

app/mailers/user_mailer.rb
amazimbe: Yes, devise requires it and provides the options param when it calls this method. Taking it out will cause an error

spec/mailers/user_mailer_spec.rb
amazimbe: Not really. I just use it to call the mailer as in UserMailer.sign_in_email(user, sign_in_link) . I could pass in any string but since I have already created the link I just decided to keep it.

config/initializers/devise.rb
amazimbe: Yes, it has to be confirmation_instructions

kicferk1 (approved):

PR #19, opened 2020-12-07
ECF-TECH-36: GitHub actions automated testing

ethanmills:
### Context
We want automated tests to run on every push

- Add a new workflow to run automated tests
- Fix said tests

### Guidance to review
None really


package.json
```diff
@@ -18,6 +18,9 @@
     "jest": "^26.6.3",
     "webpack-dev-server": "^3.11.0"
   },
+  "scripts": {
+    "test": "jest --passWithNoTests"
```
ethanmills, line 22: we'll remove this --passWithNoTests when we have some tests...

db/schema.rb
```diff
@@ -13,6 +13,7 @@
 ActiveRecord::Schema.define(version: 2020_11_30_093110) do
 
   # These are extensions that must be enabled in order to support this database
+  enable_extension "pgcrypto"
```
kicferk1, line 16: What does this do?

ethanmills: Required for UUID generation

kicferk1: Ooh, what? That's weird - I thought we added uuids (and tests that should make use of objects using uuids as ids) a while ago :/

Or is this enabling uuids at a database level when our old code does it on application level?

Basically, I'm wondering how did you get an error that needed this fix?

amazimbe: Could be dependent on the db version

ethanmills: This enables the feature in the db that your code depended on

ethanmills: It's off by default on my local db and the db provided by the github action

kicferk1 (approved):

amazimbe (approved):

PR #21, opened 2020-12-09
ECF-TECH-33: Staging and secrets

ethanmills:
### Context
This addresses secrets management and environments. I have created a staging environment, and made it easier to set up production.

- Use rails credentials for environment specific secrets
- Split github action deploy into two parts
- Make the healthcheck endpoint more use (environment name and git hash)


ethanmills:
@cfq I'm currently using a tiny-unencrypted-11 db on staging, because it's free. We might want something more, what do you think?

.github/workflows/deploy.yml
```diff
@@ -53,23 +59,39 @@ jobs:
       - name: Build and push docker image from builder target
         uses: docker/build-push-action@v2
         with:
-          build-args: BUILDKIT_INLINE_CACHE=1
+          build-args: |
+            BUILDKIT_INLINE_CACHE=1
+            VERSION=${{ env.TAG }}
           cache-from:
-            ${{ env.DOCKERHUB_REPOSITORY }}:develop
+            ${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ github.event.inputs.environment }}:${{ github.event.inputs.environment }}
```
kicferk1, line 66: Hmm, I was thinking the tag would correspond to git ref rather than environment. After all, we know which environment it is from repository name.

.github/workflows/deploy.yml
```diff
@@ -53,23 +59,39 @@ jobs:
...
           push: true
           tags: |
-            ${{ env.DOCKERHUB_REPOSITORY }}:${{ env.TAG }}
-            ${{ env.DOCKERHUB_REPOSITORY }}:develop
+            ${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ github.event.inputs.environment }}:${{ env.TAG }}
+            ${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ github.event.inputs.environment }}:${{ github.event.inputs.environment }}
```
kicferk1, line 70: Ditto

.github/workflows/deploy.yml
```diff
@@ -53,23 +59,39 @@ jobs:
...
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           TF_VAR_paas_user: ${{ secrets.GOVPAAS_DEV_USERNAME }}
           TF_VAR_paas_password: ${{ secrets.GOVPAAS_DEV_PASSWORD }}
         run: |
-          export TF_VAR_paas_app_docker_image=${{ env.DOCKERHUB_REPOSITORY }}:develop
+          export TF_VAR_paas_app_docker_image=${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ github.event.inputs.environment }}:${{ env.TAG }}
           cd terraform/app
           terraform init -reconfigure -input=false -backend-config="bucket=paas-s3-broker-prod-lon-7f2ca242-9929-4662-a79c-c454ea56ea7b"
```
kicferk1, line 83: I wonder if the bucket name could be put in `../workspace-variables/xxx.tfvars`?

.github/workflows/deploy.yml
```diff
@@ -53,23 +59,39 @@ jobs:
...
-          terraform apply -input=false -auto-approve -var-file ../workspace-variables/dev.tfvars -var='secret_paas_app_env_values={"GOVUK_NOTIFY_API_KEY":"${{ secrets.GOVUK_NOTIFY_DEV_API_KEY }}","SECRET_KEY_BASE":"${{secrets.SECRET_DEV_KEY_BASE}}"}'
+          terraform apply -input=false -auto-approve -var-file ../workspace-variables/dev.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}"}'
+
+      - name: Deploy to staging
+        if: ${{ github.event.inputs.environment == 'staging' }}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID_STAGING }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY_STAGING }}
+          TF_VAR_paas_user: ${{ secrets.GOVPAAS_STAG_USERNAME }}
+          TF_VAR_paas_password: ${{ secrets.GOVPAAS_STAG_PASSWORD }}
```
kicferk1, line 92: We can probably do it in a separate PR, but I wonder if we can move any of those to our Rails secrets file?

.gitignore
```diff
@@ -41,4 +41,7 @@ terraform.tfstate.d/
 .terraform/
 .terraform.tfstate.lock.info
 .terraform.lock.hcl
-tokens
\ No newline at end of file
+tokens
+
+# Rails secrets
+/config/credentials/*.key
```
kicferk1, line 47: Where will we store this key? I guess that's a wider question of 'where do we store secrets in a way visible to team members?'

config/database.yml
```diff
@@ -20,6 +20,14 @@ test:
...
     idle_in_transaction_session_timeout: 0
 
+deployed_development:
+  <<: *default
+  sslmode: require
+
+staging:
+  <<: *default
+  sslmode: require
+
```
kicferk1, line 30: Do we need different configs for `deployed_development` and `staging`?

config/environments/deployed_development.rb
```diff
@@ -0,0 +1,107 @@
+Rails.application.configure do
+  # Settings specified here will take precedence over those in config/application.rb.
```
kicferk1, line 2: If we are adding many of them, can we remove comments from them to make them easier to compare? 

Actually, before we do. Is this any different to `production` one? And is it any different to staging?

kicferk1 (commented):
Can we add a note to readme on how to manage rails credentials? Even if it is just a link to documentation?

.github/workflows/deploy.yml
ethanmills: No, we want to cache from the environment here, as before. We then tag is as both e.g. `dev` and the commit hash. That way, we can deploy the commit hash, but use `dev` as a starting point for the next image build

ethanmills: Happy to chat about this if that didn't make sense

.github/workflows/deploy.yml
ethanmills: No can do, we can't access variables in setting up the backend

ethanmills: We might be able to store it in a different partial config file, I'll check

.github/workflows/deploy.yml
ethanmills: Not as far as I'm aware. I think we can only access those from inside rails

.gitignore
ethanmills: Zoho? 😛

ethanmills: Seriously though, no idea. @cfq ?

config/database.yml
ethanmills: Different API keys, so different environments, so different configs

config/environments/deployed_development.rb
ethanmills: The only difference with production is that has some ENV["IGNORE_SECRETS_FOR_BUILD"]

No difference with staging at the moment, but I expect that to change

.gitignore
cfq: In our local ENV files?

ethanmills: I have a secret key. I need to share it with the team. In other projects we've had a shared password manager for that sort of thing. How do we do it here?

cfq:
> @cfq I'm currently using a tiny-unencrypted-11 db on staging, because it's free. We might want something more, what do you think?

There's a cost attached to it, so unless we run into problems we should carry on with tiny. We have allocated two medium-ha-11s can get more if we need to.

kicferk1 (approved):

PR #22, opened 2020-12-10
[Tech Debt] Freeze ruby string literals

amazimbe:
### Context
# frozen_string_literal: true is a magic comment, supported for the first time in Ruby 2.3, that tells Ruby that all string literals in the file are implicitly frozen, as if #freeze had been called on each of them. That is, if a string literal is defined in a file with this comment, and you call a method on that string which modifies it, such as <<, you'll get RuntimeError: can't modify frozen String.

The comment must be on the first line of the file.

Frozen strings literals will be the literal in ruby 3.0 and many tools in the ruby ecosystem already use them in generated code. Some of their advantages are discussed below.

https://stackoverflow.com/questions/37799296/ruby-what-does-the-comment-frozen-string-literal-true-do
https://freelancing-gods.com/2017/07/27/an-introduction-to-frozen-string-literals.html

I have added # frozen_string_literal: true to all ruby files and changed rubocop to enforce this cop moving forward.


ethanmills:
Could we enable it project wide with the `RUBYOPT="--enable-frozen-string-literal"` env var?

kicferk1 (approved):
Looks good to me. I was thinking about enabling it for all of our code with something like 
```
RUBYOPT="--enable-frozen-string-literal"
```
in our environment ([stolen from here](https://freelancing-gods.com/2017/07/27/friendly-frozen-string-literals.html)) but then again, maybe some of the gems we use, or will want to use, didn't adapt to it yet :/

ethanmills (approved):

amazimbe:
> Could we enable it project wide with the `RUBYOPT="--enable-frozen-string-literal"` env var?

Yes, as @kicferk1 said many gems out there aren’t written in ways that treat string literals as frozen by default, so enabling this globally would cause problems.

PR #24, opened 2020-12-10
Govspeak preview

kicferk1:
### Context
We will import CIP content from wordpress. We will use govspeak to show it. It would be convenient to see a quick preview of what the markdown will look like when displayed. 

Adds new endpoints, only available in development, that let you preview markdown.

Also adds govspeak as a dependency.

image

### Guidance to review
Feel free to check it out and see how it works. Not that much to it. I guess we could make the endpoints into proper resources if we wanted to - happy to do it if we think it's important.


config/routes.rb
```diff
@@ -21,5 +21,10 @@
   get "/422", to: "errors#unprocessable_entity", via: :all
   get "/500", to: "errors#internal_server_error", via: :all
 
+  if Rails.env.development?
```
ethanmills, line 24: Do we also want this on deployed_development?

kicferk1: Good shout!

config/application.rb
```diff
@@ -16,6 +16,7 @@
 # require "action_cable/engine"
 # require "sprockets/railtie"
 require "rails/test_unit/railtie"
+require "govspeak"
```
amazimbe, line 19: Looks quite odd to require this here why not just do so in  app/controllers/govspeak_test_controller.rb?

ethanmills (approved):

amazimbe (approved):

PR #25, opened 2020-12-10
ECF-78 Implement logout

amazimbe:
### Context
Implement user logout

Code changes to enable users to log out

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -45,12 +45,15 @@
           <button type="button" role="button" data-module="govuk-button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
           <nav>
             <ul id="navigation" class="govuk-header__navigation " aria-label="Top Level Navigation">
-              <li class="govuk-header__navigation-item govuk-header__navigation-item--active">
-                <%= link_to "Navigation item 1", "#", class: "govuk-header__link" %>
-              </li>
+              <% if user_signed_in? %>
               <li class="govuk-header__navigation-item">
-                <%= link_to "Navigation item 2", "#", class: "govuk-header__link" %>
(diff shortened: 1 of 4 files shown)


ethanmills (approved):

app/views/layouts/application.html.erb
```diff
@@ -45,12 +45,15 @@
           <button type="button" role="button" data-module="govuk-button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
           <nav>
             <ul id="navigation" class="govuk-header__navigation " aria-label="Top Level Navigation">
-              <li class="govuk-header__navigation-item govuk-header__navigation-item--active">
-                <%= link_to "Navigation item 1", "#", class: "govuk-header__link" %>
-              </li>
+              <% if user_signed_in? %>
               <li class="govuk-header__navigation-item">
-                <%= link_to "Navigation item 2", "#", class: "govuk-header__link" %>
+                <%= link_to "Logout", destroy_user_session_path, method: "delete", class: "govuk-header__link" %>
```
kicferk1, line 50: So we `delete` here, why do we have `config.sign_out_via = :get` line in our config? And in our tests?

Also, I would argue that `post` or `delete` are in fact much better than `get` to represent a sign out action.

amazimbe: Delete is definitely preferable but very difficult to use from an anchor unless you are handling the anchor with an Ajax call. method: "delete" here just adds data-method=delete to the link and by itself doesn't do anything. As I have no desire to handle this link with ajax, I should have taken it out but forgot.

amazimbe: I have taken it out

kicferk1: Sorry, I don't quite understand, what is the behaviour that's difficult to get exactly? 

I really don't like the idea of signing out with `get`. Signing out doesn't seem like an idempotent operation for me for one.

And I had a look at teaching vacancies, and apply for teacher training - they both use navigation items with delete for logout.

kicferk1 (approved):

PR #28, opened 2020-12-15
[ECF-83-WS1] School induction coordinator registration workflow

amazimbe:


.rubocop.yml
```diff
@@ -8,6 +8,7 @@ AllCops:
     - 'node_modules/**/*'
     - 'bin/**/*'
     - 'db/schema.rb'
+    - 'vendor/**/*'
```
amazimbe, line 11: Is this just for your local env?

ethanmills: I'm not quite sure where this came from, but it's being linted

ethanmills: (and it shouldn't be). It seems to be a gem cache

app/controllers/users/induction_coordinators/registrations_controller.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+class Users::InductionCoordinators::RegistrationsController < Devise::RegistrationsController
```
amazimbe, line 3: I think there's more potential to reuse code and avoid duplication if we organise this by feature rather than by type of user. I'd have it as Users::Registrations::InductionCoordinatorsController. This will add the endpoints specifically required by InductionCoordinators and anything shared between profiles would be in Users::RegistrationsController. What do you think?

ethanmills: That sounds like a good idea

app/views/users/induction_coordinators/registrations/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h2 class="govuk-heading-l">Sign up</h2>
+
+    <%= form_for(resource, as: resource_name, url: induction_coordinator_registration_register_path) do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <%= f.govuk_email_field :email, autocomplete: "email", value: @email, readonly: true %>
+
+      <%= f.govuk_text_field :school_name, label: { text: "School name" }, value: @school.name, readonly: true %>
```
amazimbe, line 10: If they have already confirmed the school, probably there's no need to show that on the form. Better to have a hidden field with the school_id that we can then later use to retrieve the school unambiguously without worrying about upcasing / downcasing or getting rid of spaces.

config/routes.rb
```diff
@@ -5,6 +5,14 @@
     sessions: "users/sessions",
   }
 
+  as :user do
+    get "/induction_coordinator/registration/school_confirmation", to: "users/induction_coordinators/registrations#confirm_school"
+    get "/induction_coordinator/registration/check_email", to: "users/induction_coordinators/registrations#start_registration"
```
amazimbe, line 10: This can definitely be tidied up with namespaces and scopes. I'll take a look this afternoon

db/migrate/20201215111206_create_school_domains.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateSchoolDomains < ActiveRecord::Migration[6.0]
```
amazimbe, line 3: 👍

app/views/users/induction_coordinators/registrations/new.html.erb
ethanmills: There is also a hidden school ID 🙂

spec/models/school_spec.rb
```diff
@@ -12,5 +12,7 @@
 
     it { is_expected.to have_one(:partnership) }
     it { is_expected.to have_one(:lead_provider).through(:partnership) }
+    it { is_expected.to have_and_belong_to_many(:induction_coordinator_profiles) }
+    it { is_expected.to have_and_belong_to_many(:school_domains) }
```
amazimbe, line 16: This just looks wrong to me. I don't see why we should have a school with many domains. Even if it did, there'd only be a handful and the data wouldn't change that often we could store it in an array field

ethanmills: We can always change this, but this was motivated by the idea that you might have an "@multi-academy-trust.com" type email shared by multiple schools

ethanmills:
Closed and reopened to recreated test app. Easiest way to rollback schema changes

app/controllers/users/induction_coordinators/registrations_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      if @schools.any?
+        handle_matching_schools
+      else
+        redirect_to induction_coordinator_registration_check_email_path, alert: "No schools matched your email"
+      end
+    end
+  end
+
+  def handle_matching_schools
```
amazimbe, line 31: 👍 but I think this should go at the bottom under 'private' as it's just a helper rather than an action handler

app/controllers/users/induction_coordinators/registrations_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+    else
+      redirect_to root_path, alert: "Someone from your school has already signed up"
+    end
+  end
+
+  def new
+    @email = params[:email]
+    @school = School.find(params[:school_id])
+    super do
+      render "users/induction_coordinators/registrations/new" and return
```
amazimbe, line 45: 'and return' is probably not needed. There are no more statements to execute so it will just return in any case.

app/controllers/users/induction_coordinators/registrations_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def create
+    school = School.find(params[:user][:school_id])
+    super do
+      if resource.persisted?
+        profile = InductionCoordinatorProfile.create!(user: resource, schools: [school])
+      end
+    rescue StandardError => e
+      profile&.destroy!
```
amazimbe, line 56: I don't think profile and resource would automatically be in scope here if a StandardError was raised. Also, if InductionCoordinatorProfile.create! fails, profile won't be created so profile&.destroy! wouldn't  be necessary. 

I think what we are doing here can best be handled using ActiveRecord::Base.transaction on line 51. If profile or user or anything else goes wrong the whole transaction will rollback and we don't need to do the cleaning up and error handling  manually.

app/models/school.rb
```diff
@@ -4,4 +4,16 @@ class School < ApplicationRecord
   belongs_to :network, optional: true
   has_one :partnership
   has_one :lead_provider, through: :partnership
+  has_and_belongs_to_many :induction_coordinator_profiles
+
+  def full_address
+    address = <<~ADDRESS
```
amazimbe, line 10: I think we should add a unit test for this. We don't need to assign it to address if we are just building and returning the string

app/controllers/users/induction_coordinators/registrations_controller.rb
ethanmills: The super yields to a block then renders afterwards, so I think we do need this

app/controllers/users/induction_coordinators/registrations_controller.rb
ethanmills: FWIW, this does work as it, I tried it. But I didn't know we could have transactions out here, that's cool. That's much better, I'll do that. Thanks!

app/controllers/users/induction_coordinators/registrations_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+  end
+
+private
+
+  def email_domain
+    @email.split("@")[1]
+  end
+
+  def validate_creation_parameters
```
ethanmills, line 65: @amazimbe we need to do something like this, but I'm not sure this is the best way. What do you think?

amazimbe: I think it's ok.

ethanmills (approved):

PR #29, opened 2020-12-16
Ecf tech 37 rails upgrade

ethanmills:
### Context
Rails 6.1 is out and we have no code, so it looks like a good time to upgrade

Pretty simple upgrade here, the big graft was getting some of our dependencies to support 6.1. A couple of issues and a PR later and they all do 🎉


.githooks/pre-commit
```diff
@@ -6,11 +6,12 @@ git apply -R unstaged.diff
```
ethanmills, line 6: This isn't related to the upgrade, but I needed it to be able to commit the files 😛

config/initializers/permissions_policy.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Define an application-wide HTTP permissions policy. For further
+# information see https://developers.google.com/web/updates/2018/06/feature-policy
+#
+# Rails.application.config.permissions_policy do |f|
+#   f.camera      :none
```
amazimbe, line 7: All the code has been commented out in here. Is this file needed?

amazimbe (approved):

config/initializers/permissions_policy.rb
ethanmills: No, and I don't expect we'll need them in the future, but there's no harm in leaving them there in case. I think this now comes as part of the rails template

PR #32, opened 2020-12-17
Env var cleanup

ethanmills:
### Context
We were setting env vars in the production container that we shouldn't be. This meant the secret key base on the deployed app was "TestKey". I have verified this is no longer the case on the review app

- clean up env vars
- clean up gemfile

### Guidance to review
Commit by commit here, details in commit messages


ethanmills:
The review app deployed fine, the check just timed out. The deploy took ~6mins, but it was in the queue for about 10 mins before, and the timeout is 15 mins

amazimbe (approved):

PR #38, opened 2021-01-05
update the version of govspeak

xswoodsx:
### Context
We need to display more components than the alphagov/govspeak version allows so we are using our own version.

change version in the gemfile

### Guidance to review
try out secondary button on the review app.


kicferk1 (approved):
LGTM

PR #39, opened 2021-01-05
ECF-1-98: Admin interface for CRUD of LPs

ethanmills:
### Context
DfE admins need to be able to manage LP entities and users from LPs. [Ticket](https://trello.com/c/A2l05WCW/98-sprint-3-crud-interface-for-dfe-admin-to-manage-lps)

- Add pages for admins to view, update and create lead providers
- Add govuk components and replace our own
- Add simplecov for code coverage

### Guidance to review
Commit by commit will work


app/views/supplier_dashboard/show.html.erb
```diff
@@ -27,22 +27,22 @@
           Alternatively you can upload your own selected group of schools to invite.
         </p>
 
-        <%= govuk_button_link_to "Invite Schools", school_invites_path %>
-        <%= govuk_button_link_to "Upload School group", "/upload" %>
+        <%= govuk_link_to "Invite Schools", school_invites_path, button: true %>
```
kicferk1, line 30: Do they still look like buttons?

app/controllers/admin/lead_providers_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  def new
+    @lead_provider = LeadProvider.new
+  end
+
+  def create
+    @lead_provider = LeadProvider.new(name: params.dig(:lead_provider, :name))
+
+    if @lead_provider.save
+      redirect_to admin_lead_providers_path
+    else
```
kicferk1, line 23: I was thinking whether we could use `LeadProvider.create(...)` for something shorter here, but I guess the fact that we also want to assign value to `@lead_provider` when validation fails means we should have too much to put in an if statement anyway. 

Out of curiosity, does ruby let you do something as naughty as
```
if @lead_provider = LeadProvider.create(name: params.dig(:lead_provider, :name))
  redirect_to admin_lead_providers_path
else
  render :new
end
```
?

app/controllers/admin/lead_providers_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    @lead_provider = LeadProvider.find(params.require(:id))
+  end
+
+  def update
+    @lead_provider = LeadProvider.find(params.fetch(:id))
+
+    if params[:lead_provider][:name].blank?
+      @lead_provider.errors.add(:name, "Enter a name")
+      render :edit and return
+    end
```
kicferk1, line 38: This should definitely live in `LeadProvider` model.

Then doing `@lead_provider.update(...)` will give you boolean saying whether it worked or not (I think), and we can use similar pattern to create.

app/controllers/admin/lead_providers_controller.rb
```diff
@@ -10,6 +10,12 @@ def new
   end
 
   def create
+    if params[:lead_provider][:name].blank?
+      @lead_provider = LeadProvider.new
+      @lead_provider.errors.add(:name, "Enter a name")
+      render :new and return
+    end
```
kicferk1, line 17: If it lives in the model you don't need this - save will do it for free :O

spec/requests/admin/lead_providers_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+      # Then
+      expect(response).to redirect_to("/admin/lead_providers")
+    end
+
+    it "creates a new lead provider" do
+      expect {
+        post "/admin/lead_providers", params: { lead_provider: {
+          name: lead_provider_name,
+        } }
+      }.to change { LeadProvider.count }.by(1)
```
kicferk1, line 35: Is there a keyword `increase` that we could use?

Gemfile
```diff
@@ -82,5 +82,6 @@ group :test do
   gem "faker"
   gem "rails-controller-testing", ">= 1.0.5"
   gem "shoulda-matchers", "~> 4.4"
+  gem "simplecov"
```
kicferk1, line 85: ooh, are we adding coverage? are we gonna add some rules on 'not decreasing test coverage' or something similar?

kicferk1 (commented):
Mostly looks good, only big thing is moving lead provider name validation to model to make things easier for ourselves.

app/views/supplier_dashboard/show.html.erb
ethanmills: With `button: true`, yes

Gemfile
ethanmills: I wasn't planning to. We could, but I just like to have the info while writing tests to make sure I don't overlook something obvious

app/controllers/admin/lead_providers_controller.rb
ethanmills: I don't think that would work. From https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-create:

> Creates an object (or multiple objects) and saves it to the database, if validations pass. The resulting object is returned whether the object was saved successfully to the database or not.

ethanmills: I think something like 
`
@lead_provider = LeadProvider.create(name: params.dig(:lead_provider, :name))
if @lead_provider.valid?
  redirect_to admin_lead_providers_path
else
  render :new
end
`
might work, but I'm not 100%

kicferk1: Pretty sure it would, but it wouldn't really be any nicer than the present implementation.

ethanmills: Perhaps `@lead_provider.persisted?` would be better

spec/requests/admin/lead_providers_spec.rb
ethanmills: No, but this tests for increase. If the number decreased by one this test would fail. We would need `change { LeadProvider.count }.by(-1)

app/models/lead_provider.rb
```diff
@@ -3,6 +3,7 @@
 class LeadProvider < ApplicationRecord
   has_many :partnerships
   has_many :schools, through: :partnerships
+  has_many :lead_provider_profiles
```
kicferk1, line 6: Hmm, I'm not sure that sounds right to me.

`has_many :users, through: lead_provider_profiles` sounds a bit more like what I would expect. 

Mainly because profiles mean very little to the lead provider - almost all the operations I can think of relate to users rather than their profiles.

app/controllers/admin/lead_provider_users_controller.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class Admin::LeadProviderUsersController < ApplicationController
+  before_action :set_lead_provider
+
+  def index
+    @users = User.joins(:lead_provider_profile).where(lead_provider_profiles: { lead_provider: @lead_provider })
```
kicferk1, line 7: That's where having `users through profiles` relation would be useful. You could do `@users = @lead_provider.users`.

app/controllers/admin/lead_provider_users_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  def new
+    @user = User.new
+  end
+
+  def create
+    @user = User.new(
+      first_name: params.dig(:user, :first_name),
+      last_name: params.dig(:user, :last_name),
+      email: params.dig(:user, :email),
+    )
```
kicferk1, line 19: Maybe we can put that in the transaction and use `get_or_create`?

app/views/admin/lead_provider_users/new.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <%= f.govuk_text_field(
+        :last_name,
+        label: { text: 'Last name' })
+  %>
+  <%= f.govuk_email_field(
+        :email,
+        label: { text: 'Email' })
+  %>
+  <%= f.govuk_submit "Create" %>
+<% end %>
```
kicferk1, line 15: Can we commonise it with the update form? The only difference I can see is the text on submit button.

app/controllers/admin/lead_provider_users_controller.rb
ethanmills: Why would we want `get_or_create`, as opposed to just `create!`?

app/views/admin/lead_provider_users/new.html.erb
ethanmills: And the form submission path

app/controllers/admin/lead_provider_users_controller.rb
kicferk1: Maybe not particularly useful in here, but I think `get_or_create` is the better pattern to use when creating user profiles - because one user can have many profiles. 

It is by coincidence of current business logic that we think lead provider won't have any other profiles.

Although, now that I wrote it, we might want to do something like 'make emails unique' and 'have one type of flow for making a new user with a role, and another with just assigning a role to a user'. Arghh, the rabbit holes.

ethanmills: From this page, I would be very surprised if an existing user was given new permissions, as opposed to the creation of a new user

ethanmills: Emails are already unique

kicferk1: Alright, I think it's fine to keep `create!` here. I do have a feeling we will have to worry about assigning people new roles in an intuitive way (with all the admin DfE users will need to do), but definitely not here and not now.

ethanmills: Agreed

app/controllers/admin/dashboard_controller.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class AdminDashboardController < ApplicationController
+class Admin::DashboardController < Admin::BaseController
   before_action :authenticate_user!
```
kicferk1, line 4: Do we not inherit this from `Admin::BaseController`?

app/views/admin/lead_provider_users/edit.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= form_for @user, url: admin_lead_provider_user_path do |f| %>
+  <%= render "form_fields", locals={f: f} %>
+  <%= f.govuk_submit "Update" %>
+<% end %>
+
+<%= govuk_back_link text: "Back", href: admin_lead_provider_users_path(@lead_provider) %>
```
kicferk1, line 6: Nitpick, but I am not sure we want back links to be at the bottom of a page - top of the page seems to be a more common pattern. Presumably so you don't need to scroll down to find it on long pages, and to keep it consistent with them on short pages.

app/views/admin/lead_provider_users/index.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <tbody class="govuk-table__body">
+<% @users.each do |user| %>
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell"><%= user.first_name %></td>
+      <td class="govuk-table__cell"><%= user.last_name %></td>
+      <td class="govuk-table__cell"><%= user.email %></td>
+      <td class="govuk-table__cell"><%= link_to "edit", edit_admin_lead_provider_user_path(@lead_provider, user) %></td>
+    </tr>
+  <% end %>
+  </tbody>
```
kicferk1, line 20: Something is not quite right with whitespace here, in particular `tr` inside `thead` is not indented, and `users.each` tag is off.

app/views/admin/lead_provider_users/index.html.erb
```diff
@@ -0,0 +1,25 @@
...
+</table>
+
+<%= govuk_link_to "Add", new_admin_lead_provider_user_url, button: true %>
+<br />
+<%= govuk_back_link text: "Back", href: edit_admin_lead_provider_path(@lead_provider) %>
```
kicferk1, line 25: Ditto about back link - in here it can probably get annoying if you get enough LP admins. (and ditto about all other back links, not gonna spam you with comments on each of them :P )

app/views/admin/lead_providers/new.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= form_for @lead_provider, url: admin_lead_providers_path do |f| %>
+  <%= f.govuk_text_field(
+        :name,
+        label: { text: 'Lead Provider name' })
+  %>
+  <%= f.govuk_submit "Create" %>
+<% end %>
```
kicferk1, line 7: Fine, commonizing this particular form would probably be overkill :D

kicferk1 (commented):
Looks pretty good to me, main thing is back button placement. 

Probably worth checking it with a designer, especially if you got inspired by one of our designs in your placement of back buttons :)

kicferk1:
Looks like the controller refactor might have broken some redirects :/

app/controllers/admin/lead_provider_users_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+      LeadProviderProfile.create!(user: @user, lead_provider: @lead_provider)
+    end
+
+    redirect_to admin_lead_provider_users_path
+  rescue ActiveRecord::RecordInvalid
+    render :new
+  end
+
+  def edit
+    @user = User.find(params[:id])
```
amazimbe, line 32: We can probably restrict this with pundit but even then scoping like @lead_provider.users.find would probably be better than just User.find

app/views/users/registrations/_single_school.html.erb
```diff
@@ -25,5 +25,5 @@
   </tbody>
 </table>
 
-<%= govuk_button_link_to "Confirm", users_register_path(email: @email, school_id: @school) %>
-<%= govuk_button_link_to "No, change details", users_check_email_path, class: "govuk-button--secondary" %>
+<%= govuk_link_to "Confirm", users_register_path(email: @email, school_id: @school), button: true %>
```
amazimbe, line 28: Just curious why this has been changed

spec/models/user_spec.rb
```diff
@@ -9,9 +9,28 @@
...
+    it { is_expected.to validate_presence_of(:last_name).with_message("Enter a last name") }
+    it { is_expected.to validate_presence_of(:email).with_message("Enter an email") }
+    it {
+      is_expected.to validate_uniqueness_of(:email)
+                       .case_insensitive
+                       .with_message("This email address is already in use")
+    }
+
+    it "rejects an invalid email" do
+      user = FactoryBot.create(:user)
```
amazimbe, line 23: You could just do user = FactoryBot.build(:user, email: 'invalid'). Saves you having to do user.email = "invalid"

spec/models/user_spec.rb
```diff
@@ -9,9 +9,28 @@
   end
 
   describe "validations" do
-    it { is_expected.to validate_presence_of(:first_name) }
-    it { is_expected.to validate_presence_of(:last_name) }
-    it { is_expected.to validate_presence_of(:email) }
+    subject { FactoryBot.create(:user) }
```
amazimbe, line 12: I doubt this line is necessary

app/views/admin/lead_provider_users/edit.html.erb
ethanmills: Ah yeah, force of habit. NHS style guide is at the bottom of the page, their research showed that screenreader users found them distracting at the top of the page. I'll move them to the top. @cfq any views on this?

spec/requests/admin/lead_providers_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+    it "renders the correct template" do
+      # When
+      get "/admin/lead_providers"
+
+      # Then
+      expect(response).to render_template(:index)
+    end
+  end
+
+  describe "POST /admin/lead_providers" do
```
amazimbe, line 19: Having every check in a separate request test though tidy is probably not a great idea. It's Ok for now as our test suite is small but as it gets bigger it will become painfully slow for the test suite to complete. For the easy checks you are probably better off combining them in future.

app/views/admin/lead_provider_users/edit.html.erb
kicferk1: Maybe we want to use [the back link](https://design-system.service.gov.uk/components/back-link/)? 

Which says about itself:
> Always place back links at the top of a page, before the <main> element. Placing them here means that the “Skip to main content” link allows the user to skip all navigation links, including the back link.

ethanmills: Yeah, that's what I'm using, just put it in the wrong place. I've internalised the NHS design system, which reads

> The GOV.UK Design System says that you should put back links at the top of a page, but a couple of NHS services (NHS 111 online and Find out why your NHS data matters) found that it works better at the bottom of the page below the primary action. That's because we don't want to suggest to people who use a screen reader that they leave the page prematurely.

ethanmills: https://service-manual.nhs.uk/design-system/components/back-link

kicferk1: Ah, my bad, for some reason I thought it was a button 😕

app/views/users/registrations/_single_school.html.erb
ethanmills: We added the `govuk-components` gem and stripped out our own `govuk_button_link_to` helper method

spec/models/user_spec.rb
ethanmills: We need this because of the `validate_uniqueness_of`. Read this for details: https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_record/validate_uniqueness_of_matcher.rb#L56

ethanmills: Please correct me if there's a better way!

spec/requests/admin/lead_providers_spec.rb
ethanmills: I don't know what the overhead is, but each test is running in about 60ms at the moment, so I figured there was no harm

app/views/admin/lead_provider_users/index.html.erb
ethanmills: By default, rubymine does not intent children of ```html
body
thead
tbody
tfoot```. What is your preference?

spec/models/user_spec.rb
amazimbe: 👍 Looks like that's the only way

spec/requests/admin/lead_providers_spec.rb
amazimbe: As I said, it's Ok for now as the project is very young  but let's keep that in mind in future. Things will probably be different in 3 to 5 years.  I've had to refactor a few test suites due to tests taking more than 45 minutes to run.

app/views/admin/lead_provider_users/index.html.erb
kicferk1: I strongly prefer we indent them. I find indentation really helpful when reading html, especially something as potentially complicated as tables.

kicferk1: I guess you already found https://youtrack.jetbrains.com/issue/WEB-2230, but it seems changing it is not too difficult, although it needs to be done for each rubymine user :/

spec/models/user_spec.rb
ethanmills: That would be expecting the FactoryBot invocation to raise a RecordInvalid, which just seems wrong somehow

app/views/admin/lead_provider_users/index.html.erb
ethanmills: Yeah I've found the settings. I'll see if I can track the editor config file?

kicferk1: Sounds good to me :)

spec/models/user_spec.rb
amazimbe: FactoryBot.build(:user, email: 'invalid') won't cause an error. It's exactly what you need here as it will build the record but not automatically save it.

ethanmills: Sorry, was using create 🙄

app/views/admin/lead_provider_users/index.html.erb
```diff
@@ -0,0 +1,26 @@
...
+        <td class="govuk-table__cell"><%= user.last_name %></td>
+        <td class="govuk-table__cell"><%= user.email %></td>
+        <td class="govuk-table__cell"><%= link_to "edit", edit_admin_lead_provider_user_path(@lead_provider, user) %></td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
+
+<%= govuk_link_to "Add", new_admin_lead_provider_user_url, button: true %>
+<br/>
```
kicferk1, line 26: We probably don't need this break here, or any other end of files :)

kicferk1 (commented):
Tiny nitpick with brs, other than that it looks good.

kicferk1 (approved):

PR #41, opened 2021-01-06
ECF-97 DFE admin login

amazimbe:
A DfE administrator can log in with their email address
They can see an administrator dashboard on successful login

### DIFF
--- a/app/controllers/admin_dashboard_controller.rb
+++ b/app/controllers/admin_dashboard_controller.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AdminDashboardController < ApplicationController
+  before_action :authenticate_user!
+
+  def show; end
+end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,11 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
+  def profile_dashboard_url(user)
+    if user.admin_profile
+      admin_dashboard_url
+    else
+      dashboard_url
+    end
+  end
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -14,7 +14,7 @@ unless LeadProvider.first
   LeadProvider.create!(name: "Test Lead Provider")
 end
 
-unless AdminProfile.first
+unless AdminProfile.first || Rails.env.production?
   user = User.find_or_create_by!(email: "ecf@mailinator.com") do |u|
     u.first_name = "Admin"
     u.last_name = "User"
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -13,3 +13,11 @@ end
 unless LeadProvider.first
   LeadProvider.create!(name: "Test Lead Provider")
 end
+
+unless AdminProfile.first
+  user = User.find_or_create_by!(email: "ecf@mailinator.com") do |u|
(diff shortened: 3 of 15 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,4 +1,11 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
+  def profile_dashboard_url(user)
```
amazimbe, line 4: Just generating the right dashboard url based on the user's profile.

app/controllers/admin_dashboard_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AdminDashboardController < ApplicationController
+  before_action :authenticate_user!
```
kicferk1, line 4: Can non-admin users access this dashboard? Do we need an extra check saying 'if user not admin, kick them out'?

kicferk1: Probably worth adding a test describing this behaviour too.

amazimbe: @kicferk1 that did cross my mind but I felt it was out of scope for this login pr. I'd typically use pundit for that and add policies and policy tests. I won't add them to this pr but I can do a separate one after this on when we beef up this controller.

kicferk1: Alright, happy for it to be a separate PR

kicferk1 (approved):
Should have approved yesterday, sorry if it delayed you :/

ethanmills:
Could we seed an admin account for deployed-dev using a [mailinator ](https://www.mailinator.com/) email address? (and add that to the notify permitted list)

amazimbe:
> Could we seed an admin account for deployed-dev using a [mailinator ](https://www.mailinator.com/) email address? (and add that to the notify permitted list)

@ethanmills I'm probably missing something. What do we need mailinator for?

ethanmills:
> > Could we seed an admin account for deployed-dev using a [mailinator ](https://www.mailinator.com/) email address? (and add that to the notify permitted list)
> 
> @ethanmills I'm probably missing something. What do we need mailinator for?

Doesn't have to be that, I just thought it might make a convenient joint mailbox.

I would like to be able to log in as different user types on dev, but we can't all add lots of different emails addresses, because notify limits the number of permitted email addresses. Therefore I think it might be good to have some shared test accounts. That means we need some shared mailboxes.

kicferk1:
Do we already have shared email accounts? Like `ecf-dev@digital.education.gov.uk` that we use for accessing paas when deploying?

ethanmills:
That's a good point, I'd forgotten about those

ethanmills:
The advantage of mailinator is that we'd be able to easily share it with the wider team, though.

db/seeds.rb
```diff
@@ -13,3 +13,11 @@
 unless LeadProvider.first
   LeadProvider.create!(name: "Test Lead Provider")
 end
+
+unless AdminProfile.first
```
ethanmills, line 17: This is great, exactly what I was thinking of. We potentially want this as 
```
unless AdminProfile.first  || Rails.env.production?
```
however

ethanmills (approved):

PR #42, opened 2021-01-08
Govspeak styles

kicferk1:
### Context
A lot of govspeak we present is perfectly valid html, but perhaps not looking as slick as it could. We found https://govspeak-preview.herokuapp.com/ and the repository for them, and decided to copy their approach - which is 'steal the styling of govuk publishing components'. 

Add `govuk_publishing_components` to gemfile - I feel explicit dependency makes sense, given we want to steal their styles.

Use [the asset pipeline](https://guides.rubyonrails.org/asset_pipeline.html) to import css from a gem - apparently [webpacker does not believe it should be supported](https://github.com/rails/webpacker/issues/488).

Tests were crashing and burning around some css - and I used the solution found in https://github.com/alphagov/govuk-frontend/issues/1350 to fix it. 

### Guidance to review
```

|                 |                                                                                                                                                                                                                                                                                                                                                                             |
| --------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Questioning** | Effective instruction checks for understanding and extends pupil thinking (Rosenshine, 2012). Ms Brophy can refine her questioning to ensure she gets quality answers from her pupils, helping her to decide whether to revisit content or go deeper into topics.                                                                                                           |
| **Talk**        | Pupils can refine their understanding and gain new insights from their knowledge through effective peer talk (Jay et al. 2017; Kirschner et al., 2018). Ms Brophy needs to check whether pupils have enough knowledge to talk effectively and use her instruction to facilitate successful talk.                                                                            |
| **Feedback**    | Good quality feedback is one the best-evidenced and most effective ways to improve pupil progress (EEF, 2018). Ms Brophy needs to practise ‘decision-driven data collection’: considering what kind of feedback she has the capacity to give and pupils have the capacity to act upon, and how she can do this in a way which supports pupils to manage their own learning. |

```
is markdown for a table. Compare it between [dev app](https://ecf-dev.london.cloudapps.digital/govspeak_test), the review app and https://govspeak-preview.herokuapp.com/

If you can think of a better way of handling assets (maybe bending webpacker to our whims), let me know.

### DIFF
--- a/app/webpacker/packs/application.js
+++ b/app/webpacker/packs/application.js
@@ -1,6 +1,9 @@
 /* eslint-disable import/first */
 require.context("govuk-frontend/govuk/assets");
 
+//= require govuk_publishing_components/dependencies
+//= require govuk_publishing_components/all_components
+
 import "../styles/application.scss";
 import { initAll } from "govuk-frontend";
 
--- a/app/webpacker/packs/application.js
+++ b/app/webpacker/packs/application.js
@@ -1,9 +1,6 @@
 /* eslint-disable import/first */
 require.context("govuk-frontend/govuk/assets");
 
-//= require govuk_publishing_components/dependencies
-//= require govuk_publishing_components/all_components
-
 import "../styles/application.scss";
 import { initAll } from "govuk-frontend";
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,6 +63,9 @@ jobs:
       - name: Set up test database
         run: bin/rails db:create db:schema:load
 
+      - name: Precompile assets
+        run: rake assets:precompile
+
       - name: Run tests
         run: bundle exec rspec
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,9 +63,6 @@ jobs:
       - name: Set up test database
         run: bin/rails db:create db:schema:load
 
-      - name: Precompile assets
-        run: rake assets:precompile
-
       - name: Run tests
         run: bundle exec rspec
 
--- a/Gemfile
+++ b/Gemfile
@@ -41,6 +41,9 @@ gem "tzinfo-data"
 
 gem "govuk_design_system_formbuilder", "~> 2.1", ">= 2.1.5"
 
+gem "sass-rails"
+gem "govuk_publishing_components"
+
 platform :mswin, :mingw, :x64_mingw do
   gem "wdm", ">= 0.1.0"
 end
--- a/Gemfile
+++ b/Gemfile
@@ -41,8 +41,8 @@ gem "tzinfo-data"
 
 gem "govuk_design_system_formbuilder", "~> 2.1", ">= 2.1.5"
 
-gem "sass-rails"
 gem "govuk_publishing_components"
+gem "sass-rails"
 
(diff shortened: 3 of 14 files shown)


app/webpacker/packs/application.js
```diff
@@ -1,6 +1,9 @@
 /* eslint-disable import/first */
 require.context("govuk-frontend/govuk/assets");
 
+//= require govuk_publishing_components/dependencies
+//= require govuk_publishing_components/all_components
```
ethanmills, line 5: What does this do?

kicferk1: Got it from https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md

I presume it adds any javascript that is needed to make govuk publishing components work.

That being said, maybe I should switch to a version that is not deprecated.

kicferk1: Also, might be worth digging to see if there is any that we want, actually - maybe all we want is their css.

ethanmills:
I'm not convinced the css bundling is working correctly, we now seem to be loading an additional large css file

kicferk1:
> I'm not convinced the css bundling is working correctly, we now seem to be loading an additional large css file

That is more or less what I would expect - maybe we can work on the size of that file.

But if you look at `app/views/layouts/application.html.erb`, I added a new stylesheet link there. So I kind of would expect the page to load an additional file. I imagine it's large because it's all of css of publishing components. Or did you mean that it should be split into smaller bits? 

The reason I kept both is that keeping it in one seemed like a bad idea. Purely webpacker approach seems to be really hard, and not what webpacker wants. Purely asset pipeline (reportedly) doesn't hot reload as efficiently, if at all.

ethanmills:
> That is more or less what I would expect - maybe we can work on the size of that file.
> 
> But if you look at `app/views/layouts/application.html.erb`, I added a new stylesheet link there. So I kind of would expect the page to load an additional file. I imagine it's large because it's all of css of publishing components. Or did you mean that it should be split into smaller bits?
> 
> The reason I kept both is that keeping it in one seemed like a bad idea. Purely webpacker approach seems to be really hard, and not what webpacker wants. Purely asset pipeline (reportedly) doesn't hot reload as efficiently, if at all.

If that's what you expected then I guess that's fine, it just seemed surprising to me. When using webpack(er) I'm used to having a single bundled file. This seems like an unusual compromise

kicferk1:
> If that's what you expected then I guess that's fine, it just seemed surprising to me. When using webpack(er) I'm used to having a single bundled file. This seems like an unusual compromise

I mean, I can make it a single webpacker file - by copy-pasting css from publishing components into a directory and telling webpack to look for it. It just seemed like a bad idea - if they update their styles we would have to redo the copy paste, rather than just updating version.

app/webpacker/packs/application.js
kicferk1: So I had a quick read through their js, and I am tempted to just not use it. Until some extensive and deep testing shows we need it, we can probably cut it.

ethanmills (approved):

kicferk1:
Deployment of review app went fine, but adding a comment on the PR apparently not. I don't think it matters much, so I'm gonna merge it.

PR #44, opened 2021-01-11
Secondary button fix

kicferk1:
### Context
Secondary buttons approach we used in old govspeak had annoying styling when we added govuk publishing components styles. This bumps the version to something that should work better.

### Guidance to review
The following should be grey on review app (and not white). 
```
{button secondary}[Download this module (PDF)](http://www.early-career-framework.education.gov.uk/ambition/wp-content/uploads/sites/3/2020/09/EarlyCareerTeachers_2020_TeacherHandbook_I1.pdf){/button}
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,7 +48,7 @@ platform :mswin, :mingw, :x64_mingw do
   gem "wdm", ">= 0.1.0"
 end
 
-gem "govspeak", git: "https://github.com/DFE-Digital/ecf-govspeak.git", ref: "eaf8ce3"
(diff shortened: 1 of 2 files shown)


ethanmills (approved):
For design PRs like this it might be nice to have before/after screenshots in future. What do you think?

ethanmills (changes requested):
500s on the review app

kicferk1:
> 500s on the review app

Oops, good catch!

As for design changes, yeah, probably makes sense. especially that it should be easy - we have the old version of it on our dev paas space app after all.

ethanmills (approved):
Looks good!

PR #46, opened 2021-01-11
Govuk styling

xswoodsx:
### Context
Highlights in the govspeak need some additional styling. Speaking with Mark A he said that highlight sections should be kept black and white for govifying. 

Styling of .call-to-action and information classes so that they resemble what is published on govuk. 

### Guidance to review
Having the .call-to-action change in the _highlights file didn't work and the govuk_publishing_components seemed to overwrite changes in there. I Moved them to the application.scss and it worked.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -7,21 +7,5 @@
 @import "govuk_publishing_components/components/textarea";
 
 @import "./govspeak/highlights";
-
-// Proof of concept that our styling takes priority
-.gem-c-govspeak {
-  table {
-    td,
-    th {
-      padding: 10px 20px;
-    }
-    td:last-child {
-      padding-right: 0;
-    }
-  }
-  .call-to-action {
-    background-color: #fff;
-    border-style: dashed;
-    border-radius: 9px;
-  }
-}
+@import "./govspeak/call-to-action";
+@import "./govspeak/tables";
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -17,4 +19,9 @@
       padding-right: 0;
     }
   }
+  .call-to-action {
+    background-color: #fff;
```
kicferk1, line 23: What happens without setting it to white?

app/assets/stylesheets/govspeak/_highlights.scss
```diff
@@ -0,0 +1,6 @@
+.information {
+  border-style: solid;
+  border-radius: 9px;
+  padding: 40px;
+  margin: 50px;
+}
```
kicferk1, line 6: I think we should choose how we organise styles and stick with it - I am happy with putting styles for each govspeak component into their own css directory, like this one.

Although, I think it should be 
```
. gem-c-govspeak {
  .information {...}
}
```

So that if we have a website `information` class we don't get the govspeak styling in there.

app/assets/stylesheets/application.scss
xswoodsx: it goes grey. i think it's the default for cta

app/assets/stylesheets/govspeak/_highlights.scss
xswoodsx: no problem, will separate them out and do tables as well

ethanmills:
It would be really good to have before/after screenshots as part of the PR description for design changes like this. They're really useful for communicating intent

amazimbe (approved):

kicferk1 (approved):

ethanmills (approved):

PR #47, opened 2021-01-11
Pundit

ethanmills:
# Not ready yet, just want to talk about the linter

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -2,4 +2,12 @@
 
 class Admin::BaseController < ApplicationController
   before_action :authenticate_user!
+
+  def policy_scope(scope)
+    super([:admin, scope])
+  end
+
+  def authorize(record, query = nil)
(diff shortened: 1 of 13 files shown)


app/controllers/admin/base_controller.rb
```diff
@@ -1,5 +1,16 @@
 # frozen_string_literal: true
 
 class Admin::BaseController < ApplicationController
+  include Pundit
+
   before_action :authenticate_user!
+  before_action :ensure_admin
+  after_action :verify_authorized, except: :index
```
ethanmills, line 8: Rubocop is very unhappy about this, saying
```
app/controllers/admin/base_controller.rb:8:3: C: Rails/LexicallyScopedActionFilter: index is not explicitly defined on the class. (https://rails.rubystyle.guide#lexically-scoped-action-filter)
					after_action :verify_authorized, except: :index
```

Do I just define an `index` here? Seems messy

PR #48, opened 2021-01-11
School search page

kicferk1:
### Context
We would like to have a page on which suppliers can search for schools with a bunch of filters.

Mostly frontend, vast majority of the things we want to search by are in GIAS, so coming up with architecture around the search is probably best done when we have some dummy data from them.

I went for 'show searches for schools based on query params of the url' and 'post redirects to show with correct parameters' - that seemed to result in nicest url that can be copy-pasted or bookmarked. `get` form was a close second in that. 

### Guidance to review
Review app is probably a good place to play around with the search form. 

I used https://ecf-beta-prototype.herokuapp.com/school-search as inspiration.

### DIFF
--- a/app/controllers/school_search_controller.rb
+++ b/app/controllers/school_search_controller.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "govspeak"
-
 class SchoolSearchController < ApplicationController
   def show
     @search_schools_form = SchoolSearchForm.new(*form_params_show)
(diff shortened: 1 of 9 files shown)


app/controllers/school_search_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require "govspeak"
```
amazimbe, line 3: Do we need this in here?

kicferk1: Nope, copy-pasted the controller without thinking too much.

amazimbe (approved):
Code looks good to me. There are a few instantiations of OpenStruct in the views that I would rather they were done in the controller / presenter for better testability but that can wait until things start getting complicated.

PR #49, opened 2021-01-12
School search pagination

kicferk1:
### Context
We expect thousands of schools in our system, and on some queries we will display a lot of them. Paginating seems a must.

Add [kaminari](https://github.com/kaminari/kaminari) - a gem that does pagination for us.

Add pagination to school search.

Add empty school search state, link to school search from dashboard if user is a lead provider admin, show the blue tag if the school is already partnered with someone else. Do we need a green tag for when it is partnered with us?

### Guidance to review
I added `paginates_per` attribute to show off pagination in test app with 4 schools, happy to get rid of it (and maybe add a bunch more schools to the seed file if we want it). 

Tests to follow.

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class School < ApplicationRecord
+  paginates_per 2
   belongs_to :network, optional: true
   has_one :partnership
   has_one :lead_provider, through: :partnership
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class School < ApplicationRecord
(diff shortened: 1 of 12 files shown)


app/models/school.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class School < ApplicationRecord
+  paginates_per 3
```
ethanmills, line 4: What does this do? Three results per page?

ethanmills: Ignore, just read the description properly... 🙄

ethanmills:
Would be good to get a results count I think. That was definitely on the designs at some point, not sure if it still is...

ethanmills (approved):
LGTM

PR #50, opened 2021-01-12
Pundit

ethanmills:
### Context
We are going to have lots of entities with fairly complex access permissions. Pundit is a framework to manage that

- Add Pundit to the app
- Setup policies for some of the existing models
- Punditise the admin endpoints

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -2,4 +2,12 @@
 
 class Admin::BaseController < ApplicationController
   before_action :authenticate_user!
+
+  def policy_scope(scope)
+    super([:admin, scope])
+  end
+
+  def authorize(record, query = nil)
+    super([:admin, record], query)
+  end
 end
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -1,13 +1,16 @@
 # frozen_string_literal: true
 
 class Admin::BaseController < ApplicationController
+  include Pundit
+
   before_action :authenticate_user!
+  before_action :ensure_admin
+  after_action :verify_authorized
+  after_action :verify_policy_scoped
 
-  def policy_scope(scope)
-    super([:admin, scope])
-  end
+private
 
-  def authorize(record, query = nil)
-    super([:admin, record], query)
+  def ensure_admin
+    raise Pundit::NotAuthorizedError, "Forbidden" unless current_user.admin?
   end
 end
--- a/app/controllers/admin/lead_provider_users_controller.rb
+++ b/app/controllers/admin/lead_provider_users_controller.rb
@@ -17,6 +17,7 @@ class Admin::LeadProviderUsersController < Admin::BaseController
   def create
     @user = User.new(permitted_attributes(User))
     authorize @user
+    authorize LeadProviderProfile
 
     ActiveRecord::Base.transaction do
       @user.save!
--- a/app/controllers/admin/lead_provider_users_controller.rb
+++ b/app/controllers/admin/lead_provider_users_controller.rb
@@ -2,17 +2,21 @@
 
 class Admin::LeadProviderUsersController < Admin::BaseController
   before_action :set_lead_provider
+  skip_after_action :verify_authorized, only: :index
+  skip_after_action :verify_policy_scoped, except: :index
 
   def index
-    @users = @lead_provider.users
+    @users = policy_scope(@lead_provider.users)
   end
 
   def new
     @user = User.new
+    authorize @user
   end
 
   def create
-    @user = User.new(user_params)
+    @user = User.new(permitted_attributes(User))
+    authorize @user
 
     ActiveRecord::Base.transaction do
       @user.save!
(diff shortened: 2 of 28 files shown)


app/controllers/admin/base_controller.rb
```diff
@@ -1,5 +1,16 @@
 # frozen_string_literal: true
 
 class Admin::BaseController < ApplicationController
+  include Pundit
+
   before_action :authenticate_user!
+  before_action :ensure_admin
+  after_action :verify_authorized
+  after_action :verify_policy_scoped
```
kicferk1, line 9: Hmm, why do we ensure a user is admin before action, but verify that they are authorised and scoped after?

app/policies/user_policy.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class UserPolicy < ApplicationPolicy
+  def create?
+    admin_only
+  end
```
kicferk1, line 6: Does that break user registration (school induction coordinators in particular)?

app/controllers/admin/lead_provider_users_controller.rb
```diff
@@ -2,17 +2,22 @@
...
   def new
     @user = User.new
+    authorize @user
   end
 
   def create
-    @user = User.new(user_params)
+    @user = User.new(permitted_attributes(User))
+    authorize @user
+    authorize LeadProviderProfile
```
kicferk1, line 20: What does `authorize` do exactly? 

Check what the `current_user` can do the things that are authorised?

app/controllers/admin/lead_providers_controller.rb
```diff
@@ -1,16 +1,21 @@
 # frozen_string_literal: true
 
 class Admin::LeadProvidersController < Admin::BaseController
+  skip_after_action :verify_authorized, only: :index
+  skip_after_action :verify_policy_scoped, except: :index
+
   def index
-    @lead_providers = LeadProvider.all
+    @lead_providers = policy_scope(LeadProvider)
```
kicferk1, line 8: `policy_scope` I imagine loads all lead providers that the current user can see?

spec/models/user_spec.rb
```diff
@@ -38,4 +38,19 @@
       expect(subject.password_required?).to be false
     end
   end
+
+  describe "#admin?" do
+    it "is expected to be true when the user has an admin profile" do
+      user = create(:admin_profile).user
```
amazimbe, line 44: This logic looks circular and a bit confusing. Why not just create the user first i.e. user = create(:user) and then create the profile with that user i.e create(:admin_profile, user: user)

app/controllers/admin/lead_providers_controller.rb
ethanmills: Correct

app/controllers/admin/lead_provider_users_controller.rb
ethanmills: Something along the lines of `LeadProviderProfilePolicy.new(LeadProviderProfile, current_user).create?`. Probably worth reading over the [Pundit docs](https://github.com/varvet/pundit)

spec/models/user_spec.rb
ethanmills: Laziness mostly... I might create a helper method that does that, because we do this often enough that it would be good to keep as a oneliner. Same for lead provider users

app/controllers/admin/base_controller.rb
ethanmills: I wasn't sure about this, but these do fundamentally different things. For example, if I'm not an admin, without the `ensure_admin` I'll still be able to index lead providers by going to /admin/lead_providers. Pundit controls models, not pages, so I just wouldn't be able to load the /new or /edit pages.

The alternative is to use headless policies to achieve this, but that seemed a little excessive

ethanmills: For example, for the admin dashboard, we could have something like
```
# app/policies/dashboard_policy.rb
class Admin::DashboardPolicy < Struct.new(:user, :dashboard)
  def show?
    return true if user.admin?

    false
  end
end
```
and
```
# app/controllers/admin/dashboard_controller.rb
...
def show
   authorize([:admin, :dashboard], :show?)
end
```
which would do exactly the same thing, but through pundit

app/policies/user_policy.rb
ethanmills: No, since we don't call `authorize` there. If we did, then yes, it would

ethanmills: I'm expect to have several base controllers which all mandate after_action :verify_authorized & after_action :verify_policy_scoped for different site areas, but that the registration stuff will not inherit from them

spec/models/user_spec.rb
amazimbe: Yes, I think you should fix this. In addition to what I said earlier , I'm sure it creates 2 admins when it runs and the extra admin is unnecessary.

ethanmills: Could you explain how it creates two please?

ethanmills: How does that look?

amazimbe (approved):
I agree adding a trait for this is a very good idea. You have a conflict to resolve but other than that LGTM.

app/controllers/admin/base_controller.rb
kicferk1: Alright, makes sense.

kicferk1 (approved):

PR #51, opened 2021-01-12
ECF-tech-41: Upgrade terraform cloudfoundry provider

ethanmills:


kicferk1 (approved):
Well, the review app is deployed and the initial page looks reasonable - are there any other, more in depth checks we should do before making it a 'proper' pull request and hitting merge?

ethanmills:
> are there any other, more in depth checks we should do before making it a 'proper' pull request and hitting merge?

Don't think so, just wanted to make sure the thing deployed before opening a PR

PR #52, opened 2021-01-13
[ECF-1-101] Updated school coordinator login workflow

amazimbe:
### Context

Updated school coordinator login workflow based on the latest designs. The wireframes can be found here: https://miro.com/app/board/o9J_ldVNkCY=/?moveToWidget=3074457352728112506&cot=14

### Guidance to review

This pr just focuses on the login workflow. There'll be follow up prs on user registration and school registration in due course.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,16 +1,10 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
-  class EmailNotFoundError < StandardError; end
-
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
   before_action :configure_permitted_parameters, if: :devise_controller?
-  before_action :set_current_user
-
-  rescue_from EmailNotFoundError do
-    render "users/sessions/email_not_found"
-  end
+  before_action :set_current_user, unless: :devise_controller?
 
   def check
     render json: { status: "OK", version: ENV["SHA"], environment: Rails.env }, status: :ok
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,10 +1,16 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
+  class EmailNotFoundError < StandardError; end
+
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :set_current_user
+
+  rescue_from EmailNotFoundError do
+    render "users/sessions/email_not_found"
+  end
 
   def check
     render json: { status: "OK", version: ENV["SHA"], environment: Rails.env }, status: :ok
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -1,9 +1,17 @@
 # frozen_string_literal: true
 
 class Users::SessionsController < Devise::SessionsController
+  class EmailNotFoundError < StandardError; end
+
   def new
     @email = params[:email]
     super
+  end
+
+  def create
+    super
+  rescue EmailNotFoundError
+    render :email_not_found
   end
 
   def sign_in_with_token
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -10,7 +10,7 @@ class Users::SessionsController < Devise::SessionsController
 
   def create
     super
-  rescue EmailNotFoundError => e
+  rescue EmailNotFoundError
     render :email_not_found
   end
 
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -1,17 +1,9 @@
 # frozen_string_literal: true
 
 class Users::SessionsController < Devise::SessionsController
-  class EmailNotFoundError < StandardError; end
-
   def new
     @email = params[:email]
     super
-  end
-
-  def create
-    super
-  rescue EmailNotFoundError
-    render :email_not_found
   end
 
   def sign_in_with_token
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -1,9 +1,17 @@
 # frozen_string_literal: true
 
 class Users::SessionsController < Devise::SessionsController
+  class EmailNotFoundError < StandardError; end
+
   def new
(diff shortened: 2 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,11 +1,17 @@
...
 
 class ApplicationController < ActionController::Base
+  class EmailNotFoundError < StandardError; end
+
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :set_current_user
 
+  rescue_from EmailNotFoundError do
```
amazimbe, line 11: I've had to add this in here as the before_action :set_current_user above is triggering the authentication workflow rather than on th intended request to /users/sign_in

lib/devise/strategies/passwordless_authenticatable.rb
```diff
@@ -20,7 +20,9 @@ def authenticate!
             )
 
             UserMailer.sign_in_email(user, url).deliver_now
-            fail!("An email was sent to you with a magic link.")
+            fail!("Follow the instructions in your email to continue.")
+          else
+            raise ::ApplicationController::EmailNotFoundError
```
amazimbe, line 25: I raise an exception here which I then catch in the controller so that I can render the new page that is in the designs

app/controllers/application_controller.rb
kicferk1: I don't quite understand. I thought `current_user` from devise would give me either the currently logged in user, or nil if there isn't any. Does it have a side effect of triggering the authentication flow?

Is so, maybe we want to wrap it in a `user_signed_in?` check. 

I really don't like this rescue here, since really, we only want to show it in a very specific case of user trying to sign in and the email not existing.

amazimbe: @ethanmills @kicferk1 no @current_user = current_user if user_signed_in? throws the same error. Anyway, why do we need to set_current_user in application_controller?

amazimbe: @ethanmills @kicferk1 I found something that works can you have another look please.

ethanmills: Looks like the `raise` is still there? Now that we have the `unless :devise_controller?` can we not just render the  `email_not_found` directly?

kicferk1: So I want to set current user as high up as possible, because I think vast majority of our app will behave differently for different users. And I think it is really convenient to be able to access current user from any view, without worrying that it will inadvertently throw an exception :/

amazimbe: > Looks like the `raise` is still there? Now that we have the `unless :devise_controller?` can we not just render the `email_not_found` directly

@ethanmills I think you are looking at stale code

kicferk1: Although, I am perfectly happy to not do it for devise controllers :)

app/views/users/sessions/new.html.erb
```diff
@@ -1,17 +1,16 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-l">Log in</h2>
-
+  <h1 class="govuk-heading-m">Sign in with your email address</h1>
     <%= form_for(resource, as: resource_name, url: session_path(resource_name)) do |f| %>
       <div class='govuk-form-group'>
-        <%= f.govuk_email_field :email, autocomplete: "email", value: @email %>
+        <%= f.govuk_email_field :email, label: { text: '' }, autocomplete: "email", value: @email %>
```
kicferk1, line 6: It's probably not very accessible to have empty labels.

I would add some text, could even be 'email' and set `hidden` to true.

```
label: {text: "Emain", hidden: true},
```

app/controllers/application_controller.rb
ethanmills: I'm looking at 271e78ce which has `raise Users::SessionsController::EmailNotFoundError`

app/controllers/users/sessions_controller.rb
```diff
@@ -1,11 +1,19 @@
...
   def new
     @email = params[:email]
     super
   end
 
+  def create
+    super
+  rescue EmailNotFoundError
+    render :email_not_found
+  end
```
kicferk1, line 15: Are school coordinators the only people we expect to stumble onto this page without being registered?

If not, we might need to worry about distinguishing who tried to log in. Definitely not something we want solved in this PR though.

app/views/users/sessions/new.html.erb
ethanmills: I'm not sure that will do what you want: https://www.w3.org/WAI/tutorials/forms/labels/#note-on-hiding-elements

app/controllers/application_controller.rb
amazimbe: When the user email exists we render the flash message that the user should check their email. We are rendering email_not_found directly but only when the exception is raised. Not sure, if I have completely understood your point. @ethanmills

app/views/users/sessions/new.html.erb
kicferk1: Hehe, but we are simply telling the experienced technology at our formbuilder to visually hide it this way.

Really what it does is applies a class `.govuk-visually-hidden` to the label, and that class is 1px by 1px in an unassuming place :)

kicferk1: Or so I think anyway, judging by how hidden labels on https://govuk-form-builder.netlify.app/building-blocks/fieldsets/ are handled.

ethanmills: That's good, but surprising. I would have expected a `visually_hidden` attribute or similar 🙂

app/controllers/application_controller.rb
ethanmills: I think this is fine now, sorry

app/controllers/users/sessions_controller.rb
amazimbe: Yes, most, if not all, users will use the same passwordless login. There'll be a lot of conditional template rendering for sure but that can be looked at in future pr's. Probably when we start getting content.

amazimbe:
@ethanmills @kicferk1 I've addressed all the feedback. Could you take a final look, please.

kicferk1 (approved):

kicferk1:
> The base branch requires all commits to be signed.

Ah, screw it, that's what we have admin rights for, eh? @cfq

PR #54, opened 2021-01-14
update ecf-govspeak version

xswoodsx:
### Context
Figures have been added to the latest ECF-govspeak.

Bumping the ECF-govspeak version so that images will now be rendered in the app.

### Guidance to review

the below should output as shown in the picture
```
$Figure
$Alt Example of the part-whole model using the number 28. The number 20 has already been added to the lower left field, the lower right field is blank ready to be filled in. 
$EndAlt 
$URL https://www.early-career-framework.education.gov.uk/teachfirst/wp-content/uploads/sites/4/2020/08/Part-whole-model.jpg 
$EndURL 
$Caption Figure 1: Part-whole model. 
$EndCaption
$EndFigure
```

Screenshot 2021-01-14 at 11 39 03

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ platform :mswin, :mingw, :x64_mingw do
   gem "wdm", ">= 0.1.0"
 end
 
-gem "govspeak", git: "https://github.com/DFE-Digital/ecf-govspeak.git", ref: "1b56c44"
+gem "govspeak", git: "https://github.com/DFE-Digital/ecf-govspeak.git", ref: "114c888"
 
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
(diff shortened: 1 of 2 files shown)


amazimbe (approved):
LGTM!

kicferk1:
I think the figcaption looks a bit weird - it's not centred under the image as I would expect.

We should probably double check govspeak handling of it, and bump this version if we find any bugs.

But other than that it looks great!

ethanmills (approved):
The caption still seems to be left aligned, and I'm a little surprised by the div class="img", but assuming that's all expected then this looks good 🙂

kicferk1:
> The caption still seems to be left aligned, and I'm a little surprised by the div class="img", but assuming that's all expected then this looks good 🙂

Yeah, we copied the approach that govspeak takes with their other images, I hoped that in combination with govuk publishing components would center the caption, but oh well, I guess we will do it later in our app if we decide we want to.

kicferk1 (approved):

PR #58, opened 2021-01-18
ECF-EL-73 add CIP structure

kicferk1:
### Context
We want to start the work of migrating CIP content into our Rails app. Since it will be stored in our db, the first step is to write some migrations and models. There are no routes or views, and it's by design - so more work is unblocked faster. 

Add `course_year`, `course_module` and `course_lesson` to our db, add models describing their relations, write tests for them. 

### Guidance to review
https://miro.com/app/board/o9J_kj-a9uA=/?moveToWidget=3074457353294903980&cot=14 is dev corner part describing the structure I think CIP will eventually take - and this PR is just taking care of year-module-lesson part.

### DIFF
--- a/db/migrate/20210118102533_add_cip_content.rb
+++ b/db/migrate/20210118102533_add_cip_content.rb
@@ -6,7 +6,7 @@ class AddCipContent < ActiveRecord::Migration[6.1]
       t.timestamps
       t.column :is_year_one, :boolean, null: false
       t.column :title, :string, null: false
-      t.column :content, :string, null: false
+      t.column :content, :text, null: false, limit: 200_000
 
       t.references :lead_provider, null: false, foreign_key: true, type: :uuid
     end
@@ -14,9 +14,9 @@ class AddCipContent < ActiveRecord::Migration[6.1]
     create_table :course_modules, id: :uuid do |t|
       t.timestamps
       t.column :title, :string, null: false
-      t.column :content, :string, null: false
+      t.column :content, :text, null: false, limit: 200_000
 
-      t.references :course_modules, null: true, foreign_key: { to_table: :course_modules }, type: :uuid
+      t.references :next_module, null: true, foreign_key: { to_table: :course_modules }, type: :uuid
       t.references :previous_module, null: true, foreign_key: { to_table: :course_modules }, type: :uuid
       t.references :course_year, null: false, foreign_key: true, type: :uuid
     end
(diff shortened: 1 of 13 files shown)


spec/models/course_lesson_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      CourseLesson.create(
+        title: "Test Course lesson",
+        content: "No content",
+        course_module: FactoryBot.create(:course_module),
+      )
+    }.to change { CourseLesson.count }.by(1)
+  end
+
+  describe "associations" do
+    it { is_expected.to belong_to(:course_module) }
```
amazimbe, line 17: Could we test these associations as well: has_one :next_lesson, class_name: "CourseLesson"
  has_one :previous_lesson, class_name: "CourseLesson"

spec/models/course_module_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      CourseModule.create(
+        title: "Test Course module",
+        content: "No content",
+        course_year: FactoryBot.create(:course_year),
+      )
+    }.to change { CourseModule.count }.by(1)
+  end
+
+  describe "associations" do
+    it { is_expected.to belong_to(:course_year) }
```
amazimbe, line 17: Same here, I think we can add similar checks for the other associations in the class

db/migrate/20210118102533_add_cip_content.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class AddCipContent < ActiveRecord::Migration[6.1]
+  def change
+    create_table :course_years, id: :uuid do |t|
+      t.timestamps
+      t.column :is_year_one, :boolean, null: false
```
amazimbe, line 7: What do we need this for?

db/migrate/20210118102533_add_cip_content.rb
```diff
@@ -0,0 +1,34 @@
...
+      t.column :title, :string, null: false
+      t.column :content, :string, null: false
```
amazimbe, line 9: Not sure what content looks like but if it's anything the size of  a document then a text field might be more appropriate.

db/migrate/20210118102533_add_cip_content.rb
kicferk1: I think CIP content is two years in total, and there is likely going to be some logic around what lessons can be completed when. And this is the most basic thing I could think of to keep track of it.

db/migrate/20210118102533_add_cip_content.rb
kicferk1: Good call!

amazimbe (approved):
LGTM!

PR #59, opened 2021-01-18
Update school search

xswoodsx:
### Context
The school search form can currently only search for school group. Additional functionality is needed.  

- Adding a partnership when seeding the database.
- The 'Show schools partnered with another provider' check box is only seen if a lead provider is logged in.
- You can now search by URN
- The partnership checkbox now filters to only 'Show schools partnered with another provider'. 

### Guidance to review
To check the Partnerships filters on localhost you'll need to run the seed file again to create the partnership in the db. 
Unless you're logged in as an LP you'll need to comment out the if statement on the school_search/show.html.erb view.

The linter has failed because of line 42 on the school_search_form_spec (unused variable), unless we want to use it for any testing, i'll remove it.  I spotted this afterwards.

### DIFF
--- a/app/forms/school_search_form.rb
+++ b/app/forms/school_search_form.rb
@@ -6,12 +6,20 @@ class SchoolSearchForm
   attr_accessor :school_name, :location, :search_distance, :search_distance_unit, :characteristics, :partnership
 
   def find_schools(page)
-    if partnership&.include? "partnered_with_another_provider"
-      School.where(id: Partnership.select(:school_id).map(&:school_id)).page(page)
-    else
-      School.where("lower(name) LIKE ? OR lower(urn) LIKE ?", "%#{(school_name || '').downcase}%", "%#{(school_name || '').downcase}%").includes(
-        :network, :lead_provider
-      ).page(page)
-    end
+    schools = School.where("lower(name) LIKE ? OR
+                            lower(urn) LIKE ?",
+                           "%#{(school_name || '').downcase}%",
+                           "%#{(school_name || '').downcase}%")
+    .includes(:network, :lead_provider)
+
+    schools = schools.where(id: Partnership.pluck(:school_id)) if filter_by_partnership_status
+
+    schools.page(page)
+  end
+
+private
+
+  def filter_by_partnership_status
+    partnership&.include? "partnered_with_another_provider"
   end
 end
--- a/app/forms/school_search_form.rb
+++ b/app/forms/school_search_form.rb
@@ -20,6 +20,6 @@ class SchoolSearchForm
 private
 
   def filter_by_partnership_status
-    partnership&.include? "partnered_with_another_provider"
+    partnership&.include? "in_a_partnership"
   end
 end
--- a/app/forms/school_search_form.rb
+++ b/app/forms/school_search_form.rb
@@ -6,10 +6,12 @@ class SchoolSearchForm
   attr_accessor :school_name, :location, :search_distance, :search_distance_unit, :characteristics, :partnership
 
   def find_schools(page)
-    School.where(
-      "lower(name) like ?", "%#{(school_name || '').downcase}%"
-    ).includes(
-      :network, :lead_provider
-    ).page(page)
+    if partnership&.include? "partnered_with_another_provider"
+      School.where(id: Partnership.select(:school_id).map(&:school_id)).page(page)
+    else
+      School.where("lower(name) LIKE ? OR lower(urn) LIKE ?", "%#{(school_name || '').downcase}%", "%#{(school_name || '').downcase}%").includes(
+        :network, :lead_provider
+      ).page(page)
+    end
   end
 end
--- a/app/views/school_search/show.html.erb
+++ b/app/views/school_search/show.html.erb
@@ -10,22 +10,18 @@
     <div style="width: 100%; height: 100%; background-color: #dee0e2; padding: 8px;">
       <%= form_with model: @school_search_form, url: school_search_path, method: :post do |f| %>
         <h2 class="govuk-heading-m">Filter Schools</h2>
-
         <%= f.govuk_submit "Apply filters", classes: "govuk-button govuk-!-margin-bottom-0" %>
         <p>
           <%= govuk_link_to "Clear filters", "?" %>
         </p>
-
         <%= f.govuk_text_field(
               :school_name,
               value: @school_search_form.school_name,
               label: {text: "School name or group", class: "govuk-label govuk-label--s"}) %>
-
         <%= f.govuk_text_field(
               :location,
               label: {text: "Location", class: "govuk-label govuk-label--s"},
               hint: {text: "Enter a postcode, town or region"}) %>
-
         <%= f.govuk_fieldset legend: {text: 'Search distance', class: "govuk-hint"} do %>
           <div class="govuk-grid-row">
             <div class="govuk-grid-column-one-half">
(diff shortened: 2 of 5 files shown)


spec/factories/partnership.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :partnership do
```
amazimbe, line 4: Do we need some stuff in here? If there's nothing in it then we probably don't need to have this factory

db/seeds.rb
```diff
@@ -14,6 +14,11 @@
   LeadProvider.create!(name: "Test Lead Provider")
 end
 
+# TODO: Remove this when we have a way of adding partnerships
+unless Partnership.first
```
amazimbe, line 18: You probably want unless Partnership.first || Rails.env.production?

app/forms/school_search_form.rb
```diff
@@ -6,10 +6,12 @@ class SchoolSearchForm
   attr_accessor :school_name, :location, :search_distance, :search_distance_unit, :characteristics, :partnership
 
   def find_schools(page)
-    School.where(
-      "lower(name) like ?", "%#{(school_name || '').downcase}%"
-    ).includes(
-      :network, :lead_provider
-    ).page(page)
+    if partnership&.include? "partnered_with_another_provider"
+      School.where(id: Partnership.select(:school_id).map(&:school_id)).page(page)
```
amazimbe, line 10: I think you can replace Partnership.select(:school_id).map(&:school_id) with Partnership.pluck(:school_id)

app/forms/school_search_form.rb
```diff
@@ -6,10 +6,12 @@ class SchoolSearchForm
   attr_accessor :school_name, :location, :search_distance, :search_distance_unit, :characteristics, :partnership
 
   def find_schools(page)
-    School.where(
-      "lower(name) like ?", "%#{(school_name || '').downcase}%"
-    ).includes(
-      :network, :lead_provider
-    ).page(page)
+    if partnership&.include? "partnered_with_another_provider"
```
amazimbe, line 9: Probably better to have this logic in a private method

spec/forms/school_search_form_spec.rb
```diff
@@ -22,5 +22,30 @@
...
+    it "finds all schools with an empty query" do
+      form = SchoolSearchForm.new(school_name: "")
+      schools = form.find_schools(1)
+      expect(schools.count).to eq(3)
+    end
+
+    it "finds schools that have a partnership" do
+      school = schools[2]
+      lead_provider = FactoryBot.create(:lead_provider)
+      partnership = FactoryBot.create(:partnership, school: school, lead_provider: lead_provider)
```
kicferk1, line 42: The purpose of factories is to save us time when setting up for tests. The partnership factory is probably doing the opposite - I would just create a `Partnership` directly. 

And then you can painlessly remove the partnership factory like Alfred suggests :)

app/forms/school_search_form.rb
```diff
@@ -6,10 +6,12 @@ class SchoolSearchForm
...
+      School.where(id: Partnership.select(:school_id).map(&:school_id)).page(page)
+    else
+      School.where("lower(name) LIKE ? OR lower(urn) LIKE ?", "%#{(school_name || '').downcase}%", "%#{(school_name || '').downcase}%").includes(
```
ethanmills, line 12: I think this would be easier to read split over 4 lines

app/forms/school_search_form.rb
ethanmills, line 10: I'm not sure this is the behaviour we want. I think at a minimum this should work in conjunction with the other filters. I'm pretty sure Dom's design was that partnered schools would be excluded from the results, and only be included when this box was checked. I don't think that's what is happening here, right?

app/forms/school_search_form.rb
ethanmills, line 12: You seem to be searching against the school name twice

xswoodsx: Definitely, my bad on that.

spec/forms/school_search_form_spec.rb
xswoodsx: Makes sense, I wasn't 100% on whether to include the factory bot for a partnership or not so went somewhere in the middle.

spec/factories/partnership.rb
xswoodsx: I'll remove it as Michal has mentioned below. It'd be helpful to go through some testing with you, especially Factorybot, how it works with traits etc would be good to fully understand and know how best to use.

app/forms/school_search_form.rb
xswoodsx: At the minute that's how it's working, so you either get a search result from the name or by this filter. 

Definitely agree it needs to work in conjunction other filters, especially location and characteristics. I think whilst we only have this searching by school or urn I hadn't thought about working alongside just that. Happy to do that if it's worth doing from the start?

spec/factories/partnership.rb
amazimbe: Sure, let me know when you're ready.

app/forms/school_search_form.rb
ethanmills: I think so, especially as it should just be a one-liner

app/forms/school_search_form.rb
```diff
@@ -6,12 +6,20 @@ class SchoolSearchForm
...
-        :network, :lead_provider
-      ).page(page)
-    end
+    schools = School.where("lower(name) LIKE ? OR
+                            lower(urn) LIKE ?",
+                           "%#{(school_name || '').downcase}%",
+                           "%#{(school_name || '').downcase}%")
+    .includes(:network, :lead_provider)
+
+    schools = schools.where(id: Partnership.pluck(:school_id)) if filter_by_partnership_status
```
kicferk1, line 15: I think the purpose of the filter is, when viewed by lead provider XXX:
- checked: "show schools that have been partnered with some other lead provider than XXX" 
- unchecked: "only show schools partnered with XXX or with nobody"

What this does, when viewed by lead provider XXX:
- checked: "show schools that have been partnered with some lead provider" 
- unchecked: "only show schools partnered with with nobody"

I think what it does now is actually reasonably useful, but we might want to rename the label and double check it with Isabelle, cause it feels like a product decision.

xswoodsx: ahhh ok, i understood the use of the filter as what i've done but can understand why the other purpose would be used. I'll speak with Isabelle about this part of the filters intended use.

xswoodsx: I had a chat with Alaric and he said that the filter should work as it has been. I've made changes to reflect that its all schools that are partnered

app/views/school_search/show.html.erb
```diff
@@ -55,7 +55,7 @@
           <%= f.govuk_collection_check_boxes(
               :partnership,
               [OpenStruct.new(
-                id: :partnered_with_another_provider, name: 'Show schools partnered with another provider')],
+                id: :in_a_partnership, name: 'Schools in a partnership')],
```
kicferk1, line 58: "Show schools in a partnership"? 

Also, let's use double quotes on frontend strings (I know there were single quotes here, that's on me :/ )

kicferk1 (approved):
Copy question. Happy for it to go in after resolving it :)

kicferk1 (approved):

PR #61, opened 2021-01-21
[ECF-1-138] Replace first name and last name with fullname

amazimbe:
### Context

The design wireframes use full name when capturing user information. And following some discussions with the team it was decided that we should use full name rather than first name and last name. 
https://miro.com/app/board/o9J_lZ2k8oc=/

The main change in this pr is the migration to rename first name to full name and drop the last name column. Unfortunately this has far reaching repercussions when it comes to the tests. The tests have been updated accordingly.

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -11,7 +11,7 @@ class UserMailer < ApplicationMailer
       rails_mailer: mailer_name,
       rails_mail_template: action_name,
       personalisation: {
-        full_name: user.full_name.titleize,
+        full_name: user.full_name,
         sign_in_url: url,
       },
     )
@@ -29,7 +29,7 @@ class UserMailer < ApplicationMailer
       rails_mailer: mailer_name,
       rails_mail_template: action_name,
       personalisation: {
-        full_name: user.full_name.capitalize,
+        full_name: user.full_name,
         confirmation_url: confirmation_url,
       },
     )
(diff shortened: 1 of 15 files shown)


spec/requests/admin/lead_provider_users_spec.rb
```diff
@@ -149,40 +130,37 @@
...
       put "/admin/lead_providers/#{lead_provider.id}/users/#{user.id}", params: { user: {
-        first_name: first_name,
-        last_name: user.last_name,
+        full_name: full_name,
         email: user.email,
       } }
 
       # Then
-      expect(User.find(user.id).first_name).to eq(first_name)
+      expect(user.reload.full_name).to eq(full_name)
```
kicferk1, line 138: Cool, `reload` is a neat trick :)

app/mailers/user_mailer.rb
```diff
@@ -29,7 +29,7 @@ def confirmation_instructions(user, token, _options = {})
       rails_mailer: mailer_name,
       rails_mail_template: action_name,
       personalisation: {
-        first_name: user.first_name.capitalize,
+        full_name: user.full_name.capitalize,
```
kicferk1, line 32: I think we should be consistent with how we show names in our emails. Sign in one uses `titleize` instead of `capitalize` - probably causes some dissonance.

Also, I wonder if we can just use `user.full_name` in both - if the user enters lower case name and don't like it, I imagine we might let them change it? (later down the line)

kicferk1 (commented):
Tiny comment on email name case.

kicferk1:
Looks like the review app crashed when trying to apply migrations :(

kicferk1 (approved):

PR #62, opened 2021-01-22
Add cip views

kicferk1:
### Context
We want to display core induction programme content

Add a cip page, showing some introduction and all years.

Add a year page showing year page content and its modules.

Add a module page showing module page contents and its lessons.

Add a lesson page.

### Guidance to review
Go to `/core_induction_programme` on web app, we should have some dummy data.

Majority of content is to follow in a later massive seeding PR.

### DIFF
--- a/app/models/course_lesson.rb
+++ b/app/models/course_lesson.rb
@@ -7,4 +7,8 @@ class CourseLesson < ApplicationRecord
 
   validates :title, presence: { message: "Enter a title" }
   validates :content, presence: { message: "Enter content" }
+
+  def content_in_html
+    Govspeak::Document.new(content, options: { allow_extra_quotes: true }).to_html
+  end
 end
--- a/app/models/course_lesson.rb
+++ b/app/models/course_lesson.rb
@@ -8,7 +8,7 @@ class CourseLesson < ApplicationRecord
   validates :title, presence: { message: "Enter a title" }
   validates :content, presence: { message: "Enter content" }
 
-  def content_in_html
+  def content_to_html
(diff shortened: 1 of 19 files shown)


app/models/course_lesson.rb
```diff
@@ -7,4 +7,8 @@ class CourseLesson < ApplicationRecord
 
   validates :title, presence: { message: "Enter a title" }
   validates :content, presence: { message: "Enter content" }
+
+  def content_in_html
```
amazimbe, line 11: Why not just call this to_html ?

kicferk1: Maybe `content_to_html`? 

There is a bunch of elements on lesson that won't be in this html - like title, previous / next lessons, module. Just `to_html` sounds like it would want to include them.

app/views/core_induction_programme/modules/show.html.erb
```diff
@@ -0,0 +1,25 @@
...
+
+    <div class="gem-c-govspeak govuk-govspeak ">
+      <%= @course_module.content_in_html.html_safe %>
+    </div>
+
+    <ol class="app-task-list">
+      <li>
+        <p><strong>Topic materials</strong></p>
+        <ul class="app-task-list__items">
+          <% @course_module.course_lessons.each do |lesson| %>
```
amazimbe, line 13: What happens when there are no course lessons? Do we need to show a message for that?

kicferk1: Probably. What happens is, quality control is annoyed that we called something a 'module' when it might not be 😂

amazimbe (approved):
Just a couple of questions but otherwise it looks very good.

PR #64, opened 2021-01-22
ECF-EL-73: Add cip content

kicferk1:
### Context
We got a go-ahead from important people to put the content on github, so it's time to commit and load it.

Split the seeding file into several - one per CIP provider.

Add version to the year-module-lesson models - so that we can have an easier time modifying the content. When we make changes to the content, we would simply bump the version and next time seed is run it will update them. Not sure how that will work (if at all) on the dev app, but on review app it should work just fine.

### Guidance to review

I would not bother too much with reading through the seeding files - one thing they are not is pretty. Playing around with review app might be a better idea. There are some issues (like titles of some years not being what is expected), but I imagine we will fix them in due time through content review and improvements.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,35 +1,15 @@
 # frozen_string_literal: true
 
-# TODO: Remove network and school seeding when we have a way of getting them from GIAS
-unless Network.first
-  network = Network.create!(name: "Test school network")
-  School.create!(urn: "TEST_URN_1", name: "Test school one", address_line1: "Test address", country: "England", postcode: "TEST1", network: network, domains: %w[testschool1.sch.uk network.com digital.education.gov.uk])
-  School.create!(urn: "TEST_URN_2", name: "Test school two", address_line1: "Test address London", country: "England", postcode: "TEST2", network: network, domains: %w[testschool2.sch.uk network.com digital.education.gov.uk])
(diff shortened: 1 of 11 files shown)


db/seeds.rb
```diff
@@ -1,35 +1,15 @@
...
 
-# TODO: Remove network and school seeding when we have a way of getting them from GIAS
-unless Network.first
-  network = Network.create!(name: "Test school network")
-  School.create!(urn: "TEST_URN_1", name: "Test school one", address_line1: "Test address", country: "England", postcode: "TEST1", network: network, domains: %w[testschool1.sch.uk network.com digital.education.gov.uk])
-  School.create!(urn: "TEST_URN_2", name: "Test school two", address_line1: "Test address London", country: "England", postcode: "TEST2", network: network, domains: %w[testschool2.sch.uk network.com digital.education.gov.uk])
-  School.create!(urn: "TEST_URN_3", name: "Test school three", address_line1: "Test address Oxford", country: "England", postcode: "TEST3", domains: ["testschool3.sch.uk"])
-  School.create!(urn: "TEST_URN_4", name: "Test school four", address_line1: "Test address Newcastle", country: "England", postcode: "TEST4", domains: ["testschool4.sch.uk"])
-end
+# store all seeds inside the folder db/seeds
```
amazimbe, line 3: Should we remove the file then. Not sure if rails would allow that though

amazimbe (approved):
I haven't checked the large content files though.

kicferk1:
> I haven't checked the large content files though.

Yeah, I don't think reading markdown content is the most efficient way to go about checking them :P

PR #66, opened 2021-01-25
[ECF-1-101] Coordinator registration workflow

amazimbe:
### Context

Coordinator registration workflow and page content as depicted in these designs:
https://miro.com/app/board/o9J_lZ2k8oc=/
https://miro.com/app/board/o9J_lZ2k8oc=/

Mostly rendering new pages based on the workflow.

Removing old pages that are no longer relevant


app/controllers/users/registrations_controller.rb
```diff
@@ -1,45 +1,36 @@
...
-  def new
-    @email = params[:email]
-    @school = School.find(params[:school_id])
-    super
+    render :school_partially_registered if @school.partially_registered?
   end
 
   def create
     validate_creation_parameters
+    build_resource(sign_up_params)
```
kicferk1, line 27: What does `build_resource` do?

app/controllers/users/registrations_controller.rb
```diff
@@ -1,45 +1,36 @@
...
+
     ActiveRecord::Base.transaction do
-      super do
-        if resource.persisted?
-          InductionCoordinatorProfile.create!(user: resource, schools: [@school])
-        end
-      end
+      resource.save!
```
kicferk1, line 30: Is this some lovely Rails magic, something like controller having a `resource` that it understands from controller name or something like that? Is it a devise thing?

app/controllers/users/registrations_controller.rb
```diff
@@ -49,23 +40,17 @@ def email_domain
...
-    unclaimed_schools = @schools.filter { |school| school.induction_coordinator_profiles.none? }
-
-    if unclaimed_schools.any?
-      redirect_to controller: "users/registrations", action: :confirm_school, school_ids: unclaimed_schools, email: @email
-    else
-      redirect_to root_path, alert: "Someone from your school has already signed up"
-    end
+  def render_start_registration
+    flash.now[:alert] = "Your details did not match any schools."
+    render :start_registration
```
kicferk1, line 45: I don't know if we want to get into pattern of flash messages - do we render any validation on `start_registration`? Could we add an error there, that will show in error summary?

app/controllers/users/registrations_controller.rb
```diff
@@ -49,23 +40,17 @@ def email_domain
...
   end
 
   def validate_creation_parameters
-    @school = School.find(params[:user][:school_id])
+    @school = School.find_by(id: params[:user][:school_id])
```
kicferk1, line 49: I thought those are identical. Are they not?

app/controllers/users/registrations_controller.rb
```diff
@@ -49,23 +40,17 @@ def email_domain
...
     @email = params[:user][:email]
 
-    raise ActionController::BadRequest if @email.nil?
+    raise ActionController::BadRequest if @email.blank?
+    raise ActionController::BadRequest if @school.blank?
```
kicferk1, line 53: Do the bad requests get rendered in a user friendly way?

app/views/users/confirmations/confirmed.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="govuk-grid-row">
+  <h2 class="govuk-heading-m confirmed-banner">Your school is registered</h2>
```
kicferk1, line 2: Is this header gonna be full-width? Do we want to wrap it in a div saying so? 
```
<div class="govuk-grid-column-full-width">
```

Or should it be inside the two-thirds column?

app/views/users/registrations/school_partially_registered.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h2 class="govuk-heading-l">Your need to verify your email</h2>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>check your email for our verification link (it may be in your spam or junk folder)</li>
+      <li>or <%= govuk_link_to "register again", users_new_path %> in 24 hours</li>
```
kicferk1, line 6: Out of curiosity, does the 24h part work at the moment?

app/views/users/registrations/start_registration.html.erb
```diff
@@ -1,18 +1,22 @@
...
-            autocomplete: "email")
+            value: @email,
+            label: { text: "Your school email address", class: "govuk-label govuk-label--s" })
       %>
 
-      <%= f.govuk_submit "Sign up" %>
+      <%= f.govuk_submit "Continue" %>
     <% end %>
 
   </div>
```
kicferk1, line 22: This file and `users/registrations/new` look remarkably similar. Can we commonize them a little? The form for example.

app/views/users/registrations/verification_email_sent.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h2 class="govuk-heading-l">Verify your email</h2>
+    <p>We have sent you a link to verify your email. This will expire in 24 hours.</p>
+    <p>It may be in your spam or junk folder</p>
+    <br>
```
kicferk1, line 6: There are govuk design classes for spacing, we should use them over brs. 

https://design-system.service.gov.uk/styles/spacing/

`govuk-!-margin-bottom-2` or something like that will do I think.

app/views/users/sessions/redirect_from_magic_link.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-l">Continue to your account</h2>
-
+    <h3 class="govuk-heading-l">Sign in successful</h3>
```
kicferk1, line 3: Damn, we will need to sort out our headers I think. (as in h1, h2, h3, and where to use them to be semantically correct :P )
I would keep using `h2` here though.

app/webpacker/styles/registration.scss
```diff
@@ -0,0 +1,7 @@
+.confirmed-banner {
+  background-color: #00703c;
+  padding: 50px 30px;
+  margin: 15px;
+  text-align: center;
+  color: govuk-colour("white") !important;
+}
```
kicferk1, line 7: I imagine you took the styles from a site like https://design-system.service.gov.uk/patterns/confirmation-pages/

I guess that means we don't have them in our govuk styling by default?

spec/models/school_spec.rb
```diff
@@ -15,6 +15,72 @@
...
+      expect(school.partially_registered?).to be false
+    end
+
+    context "when school has an unconfirmed induction coordinator" do
+      let(:user) { create(:user, confirmed_at: nil) }
+      let!(:coordinator) { create(:induction_coordinator_profile, user: user, schools: [school]) }
+
+      it "returns false if no one has registered a school" do
+        expect(school.partially_registered?).to be true
+      end
```
kicferk1, line 71: Eh, the test description doesn't match what it is checking.

spec/requests/users/registrations/check_details_spec.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Users::Registrations /check-details", type: :request do
+  let(:school) { FactoryBot.create(:school) }
+  let(:email) { Faker::Internet.email(domain: school.domains.first) }
+
+  describe "POST /users/check_email" do
+    it "renders correct template when no schools are found" do
```
kicferk1, line 10: Rather than 'correct template' in many places in this file, I would expect the names of templates in the test description - so that when someone has a failing test they won't be tempted to just quickly change the template in it.

kicferk1 (commented):
Looks pretty good overall, most comments are minor or questions :)

app/controllers/users/registrations_controller.rb
amazimbe: It's like User.build(params) but it comes from devise. I had to override devise on this controller action so that I can render a template rather than a flash message.

app/controllers/users/registrations_controller.rb
amazimbe: Yes, devise magic! Had to override devise as explained earlier

app/controllers/users/registrations_controller.rb
amazimbe: No we are not rendering any validation errors at the moment. The form has school as well as  user information and we are running checks on the relationship between these two and rendering the correct template based on that. This one is the only error message I think the rest are just information pages. Our designer was quite adamant that I can't replace these pages with error / flash messages as the two are fundamentally serving different purposes.

app/controllers/users/registrations_controller.rb
amazimbe: find_by won't cause an exception if there is no record. It will just return nil instead.

app/controllers/users/registrations_controller.rb
amazimbe: I don't think they are, not in development anyway. Let me check a prod-like environment for that.

app/controllers/users/registrations_controller.rb
kicferk1: Well, yeah, I agree with not removing pages.

What I was thinking about was showing the same `"Your details did not match any schools."` message in form of a form error, rather than a flash message. Do you think that would be possible?

ethanmills:
Let's make use of the review app here 🙂 Would be really good to put together a set of input parameters to trigger different screen and send them over to Al and Shani and check they're happy

app/views/users/sessions/redirect_from_magic_link.html.erb
amazimbe: The wireframes show different sizes so I just picked the sizes that seemed to match the designs better. But yes, we may need to revisit this for consistency and to be semantically correct. I'll probably also get some feedback from the designer.

app/webpacker/styles/registration.scss
amazimbe: No, I just made the styles and wasn't aware of  https://design-system.service.gov.uk/patterns/confirmation-pages/. I'll try and use the classes from there as we probably already have the styles.

amazimbe:
> Let's make use of the review app here 🙂 Would be really good to put together a set of input parameters to trigger different screen and send them over to Al and Shani and check they're happy

@ethanmills I was thinking more along the lines of doing a quick demo with them so they can give me some feedback but probably after merging the pr. I could then create smaller tickets to address any recommendations they make.

cfq:
@amazimbe @ethanmills I think a preview app is a good idea. It's why we set them up in the first place – to be able to demo before merging.

amazimbe:
> @amazimbe @ethanmills I think a preview app is a good idea. It's why we set them up in the first place – to be able to demo before merging.

Agreed @ethanmills @cfq

app/controllers/users/registrations_controller.rb
amazimbe: I think what you describe is better for model validations rather than when checking connections between models like we are doing here. I'm sure it can be done. One way would be to define a new entity that has a user and a school and does the checks. I just don't think it would be worth the cost.

kicferk1: Alright, flash message it is then :)

app/models/school.rb
```diff
@@ -17,4 +17,24 @@ def full_address
...
+  end
+
+  def not_registered?
+    induction_coordinator_profiles.none?
+  end
+
+  def partially_registered?
+    return false if not_registered?
+
+    induction_coordinator_profiles
```
ethanmills, line 35: This is just `!fully_registered?`, right?

app/views/users/registrations/new.html.erb
```diff
@@ -1,19 +1,25 @@
...
-      <%= f.govuk_error_summary %>
+    <%= form_for(:induction_coordinator_profile, url: users_check_details_path) do |f| %>
+      <%= f.govuk_text_field(
+            :school_urn,
+            value: @school&.urn,
+            hint: { text: "6 digit number", class: "govuk-hint" },
+            label: { text: "Your school's Unique Reference Number (URN)", class: "govuk-label govuk-label--s" })
+      %>
+      <p class="govuk-body govuk-!-font-weight-bold">
+        <%= govuk_link_to "Find my URN", "#", class: "govuk-link" %>
```
ethanmills, line 13: Would be good to have a todo here so it doesn't get lost

app/models/school.rb
amazimbe: No, I think there's more to it than just !fully_registered? especially as we have 3 states i.e. not_registered? as well.

ethanmills:
Noticed a couple of things. Tried logging in to the review app (with no account). Saw this 
image
then clicking on the button showed me this
image

app/models/school.rb
ethanmills: Sorry, specifically lines 35-38

ethanmills: The whole function is just `!(fully_registered? || not_registered?)` I believe

amazimbe:
> Noticed a couple of things. Tried logging in to the review app (with no account). Saw this
> image
> then clicking on the button showed me this
> image

@ethanmills that was in the design. I went through it with Al and Shani this morning and that's one of the things they will be redesigning.

app/models/school.rb
amazimbe: Probably! More than happy to simplify it.

app/views/users/registrations/new.html.erb
amazimbe: There was some talk of removing that altogether in the demo this morning. But yeah, I can add a todo.

app/controllers/users/registrations_controller.rb
amazimbe: @kicferk1 yes you get a nice looking page that something went wrong.

app/controllers/users/registrations_controller.rb
amazimbe: @kicferk1 I did a demo of this with our designers and they will be splitting this into two screens so we can validate a single entity. That should address the error messaging concerns you had but it may be part of another pr.

kicferk1: Thanks for following up on it, sounds good to me to make it a separate PR.

app/controllers/users/registrations_controller.rb
kicferk1: Alright, good enough for now, eventually would be nice to have some more specific pages describing the error and how to fix it, but not on this PR.

app/views/users/registrations/school_partially_registered.html.erb
amazimbe: No, I have a trello ticket for it though

app/views/users/sessions/email_not_found.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-l">You do not have an account</h2>
-    <p>Your school may not be registered. </p>
-    <%= govuk_link_to "Register your school", users_check_email_path %>
+    <h2 class="govuk-heading-l">Your school has not registered</h2>
+    <p>You must be responsible for running your induction programme to register your school.</p>
+    <%= govuk_link_to "Register your school", users_information_path, button: true %>
```
ethanmills, line 5: Is this not meant to be `users_new_path`?

amazimbe: No, first to those information pages you had in your screenshots earlier

app/views/users/confirmations/confirmed.html.erb
ethanmills, line 2: This wants to be `<%= govuk_panel title: "Your school is registered", body: nil %>`

app/views/users/sessions/email_not_found.html.erb
ethanmills: Are you working from these designs? 
image

ethanmills: Which design are you working from?

amazimbe: https://miro.com/app/board/o9J_lZ2k8oc=/

app/views/users/confirmations/confirmed.html.erb
amazimbe: @ethanmills does this apply to all the page titles?

app/views/users/sessions/email_not_found.html.erb
amazimbe: Al and Shani have been updating it so it's slightly diff from what it was. I'll have to update it later when they are done.

app/views/users/confirmations/confirmed.html.erb
ethanmills: No, this gives you a [panel](https://design-system.service.gov.uk/components/panel/)

amazimbe: It will be full width

app/views/users/registrations/start_registration.html.erb
amazimbe: I can't see anything we can commonise and I think trying to do so for things which are likely to need frequent changes like these two would probably not be a good idea.

amazimbe:
@ethanmills @kicferk1 could you take a final look please. I have addressed all the feedback for now and any further design changes from Al and Shani will be implemented in separate prs.

app/views/users/registrations/start_registration.html.erb
kicferk1: Well, the form is similar.
We have 
```
    <%= form_for(:induction_coordinator_profile, url: users_check_details_path) do |f| %>
      <%= f.govuk_text_field(
            :school_urn,
            value: @school&.urn,
            hint: { text: "6 digit number", class: "govuk-hint" },
            label: { text: "Your school's Unique Reference Number (URN)", class: "govuk-label govuk-label--s" })
      %>
      <p class="govuk-body govuk-!-font-weight-bold">
        <%# TODO replace with actual link when available %>
        <%= govuk_link_to "Find my URN", "#", class: "govuk-link" %>
    </p>
      <%= f.govuk_email_field(
            :email,
            value: @user&.email,
            autocomplete: "email",
            label: { text: "Your school email address", class: "govuk-label govuk-label--s" })
      %>
      <%= f.govuk_submit "Continue" %>
    <% end %>
```
in the other one. The only differences I can spot are that in `registrations/new`, we are passing urn and email through other objects (user and school), while here they are their own variables.

Well, that and one has a TODO on a dead link, and the other doesn't. 

If you think they will very likely change and it is just a coincidence they look the same I'm happy to let them be coincidentally similar though :)

app/webpacker/styles/registration.scss
```diff
@@ -0,0 +1,3 @@
+.confirmed-banner {
+  margin: 15px;
```
kicferk1, line 2: Nice :)

kicferk1 (approved):

ethanmills (approved):

app/webpacker/styles/registration.scss
amazimbe: We can have a chat when I implement updates to the design but yes, I think it's better to keep these separate rather than having a form with conditionals and all that.

PR #67, opened 2021-01-26
Add gias import from public url

kicferk1:
### Context
We want some gias data, and we want it now!
I thought this was a more useful first thing to do than speculative model edits.

Shamelessly rip off approach for `https://github.com/DFE-Digital/claim-additional-payments-for-teaching`. 

The gist of it is:
1. Have a task for importing schools (that makes a request to gias, gets csv file, parses it, updates our db with it).
2. Add a rake task for it and a cron job for it.
3. Add a bunch of tests and setup around cron jobs because it's easier than untangling it from the original code :P

Adding more fields will probably be its own PR.

### Guidance to review
If you want a review app to show you all the cool schools, you will need to schedule a task for it. It should be similar to seeding (check readme), but task name is `schools_data:import`.


app/models/school_data_importer.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require "file_download"
+require "csv"
+
+class SchoolDataImporter
+  def run
```
ethanmills, line 7: We need to think about how we handle schools that close

kicferk1: Well, the good part is that GIAS data tells us when school closes.

I imagine the approach, when we decide what to do, is to have an `after_update` hook on a school model, that will do _something_. That will probably be enough for school closing.

For school change, uhh, most of them will just be saved in the db, thanks to urn matching. For schools merging, or splitting, argh, I don't even want to think about it. Not sure how it gets represented in gias, and I agree we will need to dig around that. I can add a TODO here, maybe with a ticket number describing the problem?

ethanmills: I think that would be good for now 👍

spec/services/school_data_importer_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        expect(imported_school.name).to eql("Penistone Grammar School")
+        expect(imported_school.school_type).to eql("1")
+        expect(imported_school.capacity).to eql(1700)
+        expect(imported_school.address_line1).to eql("Huddersfield Road")
+        expect(imported_school.address_line2).to eql("Penistone")
+        expect(imported_school.address_line3).to eql("Sheffield")
+        expect(imported_school.address_line4).to eql("South Yorkshire")
+        expect(imported_school.country).to eql("")
+        expect(imported_school.postcode).to eql("S36 7BX")
+        expect(imported_school.domains).to eql(["penistone-gs.uk/"])
```
amazimbe, line 44: Can we get rid of that / at the end

Gemfile
```diff
@@ -47,8 +47,13 @@ gem "tzinfo-data"
 gem "govuk-components", ">= 1.0.0"
 gem "govuk_design_system_formbuilder", "~> 2.1", ">= 2.1.5"
 gem "govuk_publishing_components", ">= 23.11.0"
+gem "httpclient"
```
amazimbe, line 50: Just curious is this what's mostly used in dfe? Most organizations tend to stick to one http client and probably there's one in dfe that's widely used. It could be this one though.

app/jobs/school_data_importer_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class SchoolDataImporterJob < CronJob
+  self.cron_expression = "0 6 * * *"
```
amazimbe, line 4: Just wondering if this should be moved to a constant or even to a config item.

Gemfile
kicferk1: I got that out of `claim-additional-payments-for-teaching`.

DfE clearly doesn't stick to one - I saw `http` gem in `apply-for-teacher-training`, `httparty` in `teaching-vacancies`, and `httpclient` in `claim-additional-payments-for-teaching`.

The convenience of using this one is that copy-pasting it alongside a bunch of other things from `claim-additional-payments-for-teaching` 'just works'.

app/jobs/school_data_importer_job.rb
kicferk1: Eh, maybe? I reckon doing it when we have a single job to keep track of is a bit of an overkill. Once we have many, and we want them to happen in particular order, I think it would make sense to have a single file specifying that.

app/models/school_data_importer.rb
```diff
@@ -0,0 +1,41 @@
...
+    date_string = Time.zone.now.strftime("%Y%m%d")
+    "https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public/edubasealldata#{date_string}.csv"
+  end
+
+  def schools_data_file
+    FileDownload.new(gias_schools_csv_url).fetch
+  end
+
+  def row_to_school(row)
+    school = School.find_or_initialize_by(urn: row.fetch("URN"))
```
amazimbe, line 27: School.find_or_initialize_by(urn: row.fetch("URN")) do |school|
   school.name = row.fetch("EstablishmentName") .... would save you having to have school as the last statement just so we can return. Even more I doubt we want to be doing these assignments for schools that have already been created.

app/models/school_data_importer.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require "file_download"
```
amazimbe, line 3: Is this coming from the httpclient?

kicferk1: Nah, it's a file we added, `lib/file_download`

app/models/school_data_importer.rb
kicferk1: I think we should make those assignments for schools we already have in database - cause they can change, and that's an easy way to update them.

app/models/school_data_importer.rb
```diff
@@ -0,0 +1,41 @@
...
+      school = row_to_school(row)
+      school.save!
+    end
+  end
+
+private
+
+  def gias_schools_csv_url
+    date_string = Time.zone.now.strftime("%Y%m%d")
+    "https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public/edubasealldata#{date_string}.csv"
```
amazimbe, line 19: The base part of this definitely needs to go in an env file. They could change their server in future

lib/file_download.rb
```diff
@@ -0,0 +1,38 @@
...
+    else
+      raise DownloadError, "#{response.status} response for #{url}"
+    end
+  end
+
+private
+
+  def temp_file_from_response(response)
+    body = response.body.force_encoding(encoding)
+    file = Tempfile.new(encoding: encoding)
```
amazimbe, line 33: Where are we writing this temp file to? Hopefully tmp

app/jobs/school_data_importer_job.rb
amazimbe: I was thinking more along the lines of not having to do a pr when some stakeholder says run the import at 8 am on Saturdays instead. In the config we would just update it but in the code it would be more involved.

kicferk1: Well, thing is, this config value would need to have a default, and if the stakeholder was important enough, then we probably would have to change the default value as well. 

I'm happy to move it to config once we get such a request though :)

lib/file_download.rb
kicferk1:
```
"/var/folders/hk/nd5czf2n0x3flmmtdhzxcqdr2_nmnl/T/20210126-12080-r1pbpc"
```
is what I'm getting when I print file path.

I don't actually see it on my filesystem, presumably because it gets deleted quickly, or is stored in memory.

app/jobs/school_data_importer_job.rb
amazimbe: We can leave it for now. Anyway, the default would be what we think is reasonable and the value would be what the stakeholder requests.

Also, for now running this at 6 in the morning might be ok but on a bust site we'd probably wan't to do the run between 12-4am to avoid slow performance for users.

amazimbe (approved):

PR #68, opened 2021-01-26
Fix docker installation for local development

callumacrae:
### Context

Docker had a couple errors when ran locally.

Adding a log file that was missing and causing an error, and make sure all dependencies are installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,8 @@ FROM ${BASE_RUBY_IMAGE} AS builder
 
 ARG BUILD_DEPS="git gcc libc-dev make nodejs yarn postgresql-dev build-base libxml2-dev libxslt-dev ttf-ubuntu-font-family"
 
+RUN mkdir log && touch log/mail.log
+
 WORKDIR /app
 
 COPY Gemfile Gemfile.lock package.json yarn.lock .ruby-version ./
@@ -22,7 +24,7 @@ RUN apk -U upgrade && \
     bundler -v && \
     bundle config set no-cache 'true' && \
     bundle config set no-binstubs 'true' && \
-    bundle --retry=5 --jobs=4 --without=development && \
+    bundle --retry=5 --jobs=4 && \
     yarn install --check-files && \
     apk del .gem-installdeps && \
     rm -rf /usr/local/bundle/cache && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN apk -U upgrade && \
     bundler -v && \
     bundle config set no-cache 'true' && \
     bundle config set no-binstubs 'true' && \
-    bundle --retry=5 --jobs=4 --without=development && \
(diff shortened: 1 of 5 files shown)


kicferk1 (approved):
LGTM!

Dockerfile
```diff
@@ -22,7 +24,7 @@ RUN apk -U upgrade && \
     bundler -v && \
     bundle config set no-cache 'true' && \
     bundle config set no-binstubs 'true' && \
-    bundle --retry=5 --jobs=4 --without=development && \
```
ethanmills, line 25: I would this `--without=development` to stay ideally. This was written as a production dockerfile really, not a development one. Maybe we should create a Dockerfile.dev if you're keen on using it for dev?

ethanmills (changes requested):

Dockerfile
kicferk1: This dockerfile installs a lot of things that we never use in production - all the test dependencies are in it (and so dependencies in dev+test are in it too), because we run tests on it.

If we want to use separate dockerfiles, it's probably ok to add a tech debt card for it somewhere, that we can pick up when we do load or penetration testing and find out our dev/test dependencies are a bad idea.

kicferk1: Hmm, actually, are we running tests on it?

kicferk1: I guess we aren't. But we are definitely installing test dependencies :/

ethanmills: nope, see https://github.com/DFE-Digital/early-careers-framework/blob/develop/.github/workflows/test.yml
If anything I'd like to remove the test dependencies, not add the dev ones

ethanmills: Aside from fluff it production, it will make our builds take longer

kicferk1: Ah, I think that last part convinced me, our builds do take a while.

callumacrae: Done - a bit uneasy with this approach as I feel it could go out of date (it looks like I'm the only person using it and I'm not a permanent member of staff), but have added a note to the README to make it clear to anyone in the future that it could be an issue.

Dockerfile
```diff
@@ -9,6 +9,8 @@ FROM ${BASE_RUBY_IMAGE} AS builder
 
 ARG BUILD_DEPS="git gcc libc-dev make nodejs yarn postgresql-dev build-base libxml2-dev libxslt-dev ttf-ubuntu-font-family"
 
+RUN mkdir log && touch log/mail.log
+
```
kicferk1, line 13: We probably don't need it in our proper production file :)

ethanmills: Not 100% either way

kicferk1 (approved):

ethanmills (approved):
Looks great, thanks! Sorry to have slowed your ramp-up
We can probably slim down the dev dockerfile over time to speed things up if needed.

PR #69, opened 2021-01-27
Cip easier updating

kicferk1:
### Context
We will likely do some content editing by hand - either ourselves or through a content person. That might include manually changing a few pieces that are easier to deal with on a case by case basis in a text area than in a seed file, up to a full content review.

And once these changes are made, we want to check them into our repository, and probably recreate them on other versions of the app.

Add `seed_dump` for dumping the contents of our database.

Add `activerecord-import` to make the import faster, with more options, and most importantly, to make foreign keys not shout at me during bulk insert.

Add a download button on cip page that admins can see.

Move from using several files with a lot of lines to using one file with few lines but a lot of width. That's how seed dump generates them, and I didn't bother to change it.

### Guidance to review
I didn't bother writing policy for cip controller, because it seems more effort than it's worth.

I imagine this setup might get removed at some point when we are happy enough with our content.


spec/requests/core_induction_programme/core_induction_programme_spec.rb
```diff
@@ -4,7 +4,12 @@
 
 RSpec.describe "Core Induction Programme", type: :request do
   it "renders the core_induction_programme page" do
-    get "/core_induction_programme"
+    get "/core-induction-programme"
     expect(response).to render_template(:show)
   end
+
+  it "redirects to cip path when user is not admin" do
```
amazimbe, line 11: Naturally, I'd expect a corollary spec for when a user is an admin

amazimbe (commented):
Just one question. Otherwise looks good. A short and sweet pr 😄

spec/requests/core_induction_programme/core_induction_programme_spec.rb
kicferk1: So would I.

Turns out it's not trivial to test `send_file` method. Internet tells me to 'just stub it' in a controller test, but I didn't get it to work easily. Do you have any ideas on how we can handle it?

app/controllers/core_induction_programme_controller.rb
```diff
@@ -4,4 +4,17 @@ class CoreInductionProgrammeController < ApplicationController
   def show
     @course_years = CourseYear.all
   end
+
+  def download_export
+    unless @current_user&.admin_profile
```
ethanmills, line 9: `@current_user&.admin?`

ethanmills (approved):
LGTM - one small thing

spec/requests/core_induction_programme/core_induction_programme_spec.rb
amazimbe: Did you try: response.body.should eql data
response.header['Content-Type'].should eql 'image/png'

kicferk1: I got it to work, the problem was that tests didn't want to generate the file in the place I wanted them to. After a long attempt at stubbing that file out I gave up and decided to just check it in.

app/controllers/core_induction_programme_controller.rb
```diff
@@ -4,4 +4,17 @@ class CoreInductionProgrammeController < ApplicationController
   def show
     @course_years = CourseYear.all
   end
+
+  def download_export
+    unless @current_user&.admin?
+      redirect_to cip_path and return
```
amazimbe, line 10: When you redirect you don't need the and return unlike when you render

amazimbe (approved):
Need to get rid of that "and return" though.

app/controllers/core_induction_programme_controller.rb
kicferk1: I think if you don't return, the method keeps going :/ At least failing tests think so.

I refactored it to not need a return though.

PR #70, opened 2021-01-27
Ecf cip add edit of content

xswoodsx:
### Context
DfE admin need to be able to edit CIP content. Who can view the edit and make updates needs to be restricted also. I've started with editing lessons before moving on to modules and years. 

 - Updating CourseLessons controller to allow editing, previewing and saving of changes to lesson content.
 - Adding a policy so that only DfE admin can access these views
 - Adding tests for CourseLessonPolicy
 - Updating routes
 
### Guidance to review
Just spotted the create in routes, it's not being used in the controller so imagine this needs to be removed. I was going to use the same logic/methods in lesson for modules and years after this PR.


app/controllers/core_induction_programme/lessons_controller.rb
```diff
@@ -1,7 +1,36 @@
...
   def show
     @course_lesson = CourseLesson.find(params[:lesson_id])
   end
+
+  def edit
+    @lesson = CourseLesson.find(params[:lesson_id])
+    authorize @lesson
+
+    @lesson_preview = params[:lesson_preview] || @lesson.content
+    @course_lesson = Govspeak::Document.new(@lesson_preview, options: { allow_extra_quotes: true }).to_html
```
kicferk1, line 18: We might want to make a helper, say `GovspeakHelper` and add a method there `content_to_html` that will do this lengthy thing for us - we duplicate it in a bunch of places and it's getting annoying.

app/controllers/core_induction_programme/lessons_controller.rb
```diff
@@ -1,7 +1,36 @@
...
+  end
+
+  def update
+    @lesson = CourseLesson.find(params[:lesson_id])
+    authorize @lesson
+
+    respond_to do |format|
```
kicferk1, line 25: Do we need this line? We only ever respond to html in this method anyway.

config/routes.rb
```diff
@@ -46,7 +46,8 @@
   resource :core_induction_programme, controller: :core_induction_programme, only: :show, as: "cip", path: "/core_induction_programme" do
     resource :years, controller: "core_induction_programme/years", only: :show, path: "/:year_id" do
       resource :modules, controller: "core_induction_programme/modules", only: :show, path: "/:module_id" do
-        resource :lessons, controller: "core_induction_programme/lessons", only: :show, path: "/:lesson_id"
+        resource :lessons, controller: "core_induction_programme/lessons", only: %i[show create edit update], path: "/:lesson_id"
+        put "/:lesson_id/edit", to: "core_induction_programme/lessons#update"
```
kicferk1, line 50: Probably worth writing some request tests for new routes.

config/routes.rb
```diff
@@ -46,7 +46,8 @@
   resource :core_induction_programme, controller: :core_induction_programme, only: :show, as: "cip", path: "/core_induction_programme" do
     resource :years, controller: "core_induction_programme/years", only: :show, path: "/:year_id" do
       resource :modules, controller: "core_induction_programme/modules", only: :show, path: "/:module_id" do
-        resource :lessons, controller: "core_induction_programme/lessons", only: :show, path: "/:lesson_id"
+        resource :lessons, controller: "core_induction_programme/lessons", only: %i[show create edit update], path: "/:lesson_id"
```
amazimbe, line 49: You could try and define this as a member route: resource :lessons, controller: "core_induction_programme/lessons", only: %i[show create edit update], path: "/:lesson_id" do
  member do
    put 'edit'
  end
end
https://guides.rubyonrails.org/routing.html#adding-member-routes and it's more like the rails way.

app/controllers/core_induction_programme/lessons_controller.rb
amazimbe, line 25: We are only rendering html so you can take out respond_to do |format| everywhere really.

amazimbe: Just render and redirect, you don't need the format.html

app/controllers/core_induction_programme/lessons_controller.rb
xswoodsx: No problem, will add that in.

i'll update the govspeaktest controller while i've added it.

app/policies/course_lesson_policy.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def show?
+    true
+  end
+
+  def create?
+    admin_only
+  end
+
+  def edit?
```
ethanmills, line 17: Don't think we need the edit here, since it's the same as update. Double check ApplicationPolicy, but I'm pretty sure `edit?` just returns `update?` by default, so we can leave it like that

app/policies/course_lesson_policy.rb
```diff
@@ -0,0 +1,32 @@
...
+    admin_only
+  end
+
+  def update?
+    admin_only
+  end
+
+private
+
+  def admin_only
```
ethanmills, line 27: Don't need to redefine this here, we get this by inheriting from `ApplicationPolicy`

app/views/core_induction_programme/lessons/show.html.erb
```diff
@@ -1,10 +1,11 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
+    <% if @current_user&.admin? %>
```
ethanmills, line 3: Maybe `if policy(@lesson).edit?` would be more suitable here?

app/controllers/core_induction_programme/lessons_controller.rb
ethanmills: Not sure exactly what's going on here. I find it a little surprising that `@lesson` is a `CourseLesson`, while `@course_lesson` is a... something else?

ethanmills: Especially since `@course_lesson` is a `CourseLesson` in `show`

xswoodsx: I think this is just inconsistent naming on my part. @course_lesson and @lesson are the same thing. Will stick with one.

app/policies/course_lesson_policy.rb
xswoodsx: We redefine the admin_only slightly in the course_lesson policy because we need non-users to be able to access the content. It better just to put that in the ApplicationPolicy?

def admin_only
    return true if user&.admin?

ethanmills: Ah sorry, hadn't spotted the difference between the two. Yes, I think I would just update ApplicationPolicy

kicferk1 (approved):
I would follow Ethan's suggestion to change admin method of application policy, rather than overriding it in lesson policy. Other than that I think it looks good.

amazimbe (approved):

PR #71, opened 2021-01-28
Add more fields to schools from GIAS extract

kicferk1:
### Context
We want more GIAS fields. 

### Changes proposed in this pull request
Add Local Authority and Local Authority District models, add more school fields.

### Guidance to review
Need to add some tests, probably remove some old schools from seeds. Also, the school group data doesn't seem to be in the public extract we are getting :( I commented the fields out, might remove that, or find a better way to deal with it.

PR #72, opened 2021-01-29
[ECF-144] Email school primary contact when school is confirmed

amazimbe:
### Context

Send email to school primary contact once an induction coordinator has registered a school  and confirmed the account.

Only send an email to the primary contact if the email they registered with is different from that of the school primary contact.

### DIFF
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -17,6 +17,9 @@ private
 
   def notify_school_primary_contact
     school = resource.induction_coordinator_profile.schools.first
-    UserMailer.primary_contact_notification(resource, school)
+
+    if school.primary_contact_email != resource.email
+      UserMailer.primary_contact_notification(resource, school)
+    end
   end
 end
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -6,9 +6,17 @@ class Users::ConfirmationsController < Devise::ConfirmationsController
     self.resource = resource_class.confirm_by_token(params[:confirmation_token])
 
     if resource.errors.empty?
+      notify_school_primary_contact
       render :confirmed
     else
       respond_with_navigational(resource.errors, status: :unprocessable_entity) { render :new }
     end
+  end
+
+private
+
+  def notify_school_primary_contact
+    school = resource.induction_coordinator_profile.schools.first
+    UserMailer.primary_contact_notification(resource, school)
   end
 end
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -16,6 +16,8 @@ class Users::ConfirmationsController < Devise::ConfirmationsController
 private
 
   def notify_school_primary_contact
+    return unless resource.induction_coordinator?
+
     school = resource.induction_coordinator_profile.schools.first
 
     if school.primary_contact_email != resource.email
--- a/db/migrate/20210128143531_add_primary_contact_email_to_school.rb
+++ b/db/migrate/20210128143531_add_primary_contact_email_to_school.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPrimaryContactEmailToSchool < ActiveRecord::Migration[6.1]
+  def change
+    add_column :schools, :primary_contact_email, :string
+  end
+end
--- a/spec/requests/users/confirmations_spec.rb
+++ b/spec/requests/users/confirmations_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe "Users::Confirmations", type: :request do
   let(:school) { create(:school) }
   let(:induction_coordinator) { user.induction_coordinator_profile }
 
-  let!(:user) do
+  let(:user) do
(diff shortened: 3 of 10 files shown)


db/migrate/20210128143531_add_primary_contact_email_to_school.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPrimaryContactEmailToSchool < ActiveRecord::Migration[6.1]
+  def change
+    add_column :schools, :primary_contact_email, :string
+  end
```
kicferk1, line 6: https://github.com/DFE-Digital/early-careers-framework/pull/71 adds that too... Arghhhhhh

Whatever. This likely should go in first.

app/controllers/users/confirmations_controller.rb
```diff
@@ -6,9 +6,20 @@ def show
...
   end
+
+private
+
+  def notify_school_primary_contact
+    school = resource.induction_coordinator_profile.schools.first
+
+    if school.primary_contact_email != resource.email
+      UserMailer.primary_contact_notification(resource, school)
+    end
```
kicferk1, line 23: So we only notify the primary email if the coordinator email is different? Do we not want to send an email to coordinator saying that the process was successful?

app/controllers/users/confirmations_controller.rb
```diff
@@ -6,9 +6,20 @@ def show
     self.resource = resource_class.confirm_by_token(params[:confirmation_token])
 
     if resource.errors.empty?
+      notify_school_primary_contact
```
ethanmills, line 9: We only want to do this for school induction coordinators, right?

db/migrate/20210128143531_add_primary_contact_email_to_school.rb
```diff
@@ -0,0 +1,7 @@
...
+end
```
ethanmills, line 7: This is going to conflict with @kicferk1 's work https://github.com/DFE-Digital/early-careers-framework/pull/71/files#diff-206b706d03f65d61ef14a0064609851a14599e1a986ebffe83307e4fd219e6c2R34

spec/requests/users/confirmations_spec.rb
```diff
@@ -4,16 +4,38 @@
 
 RSpec.describe "Users::Confirmations", type: :request do
   let(:confirmation_token) { "soC7rF8i_BgYdUxafFcP" }
+  let(:school) { create(:school) }
+  let(:induction_coordinator) { user.induction_coordinator_profile }
 
   let!(:user) do
-    create(:user, confirmed_at: nil, confirmation_token: confirmation_token)
+    create(:user, :induction_coordinator, confirmed_at: nil, confirmation_token: confirmation_token)
   end
```
ethanmills, line 12: We're going to have users who aren't induction coordinators...

app/controllers/users/confirmations_controller.rb
amazimbe: No coordinator get a message on the ui for now but later they will just get logged in if confirmation is successful

app/controllers/users/confirmations_controller.rb
amazimbe: Yes. I agree, best to check that here

app/controllers/users/confirmations_controller.rb
kicferk1: When I buy a plane ticket I also see a confirmation, but I also like to get an email 😂 

Just wanna make sure it's something a designer or a product person specified.

spec/requests/users/confirmations_spec.rb
amazimbe: We are not sure at this point but I agree I need to restrict the notification to induction coordinator confirmations

app/controllers/users/confirmations_controller.rb
ethanmills: The flow is user signs up -> user gets "confirm your email" email -> If user confirms email, they get account and we send notification email to school primary contact, right?

amazimbe: Yes, @ethanmills . @kicferk1 we are here because we are confirming a registration. Eventually, the user will just get logged in. We can't send them another confirmation that they have confirmed their account.

kicferk1: This is not confirming user registration, this is confirming school registration. I think those are quite different in the text they contain, and in the purpose they serve.

Again, I'm happy to not send that email in some scenarios, as long as we confirmed with someone (Abs?) that it's ok to do.

amazimbe: This is what was designed and presented to the team. Obviously it's not perfect but for now my job is to implement what was agreed.  I agree the workflow is a bit fuzzy as registering a coordinator automatically leads to registering a school and I'll raise your concerns with  the designer and Abs. If they agree I'll do another pr to change it.

kicferk1: Alright, sounds good to me.

amazimbe:
I've applied the feedback. Could you take another look please. @kicferk1 , @ethanmills

kicferk1 (approved):

ethanmills (approved):

PR #73, opened 2021-01-29
Link CourseYear to CIPs

kicferk1:
### Context
We recently added CIPs. It makes much more sense for a CourseYear to be linked to a CIP than it is to be linked to a lead provider - because some lead providers use others' CIPs.

Remove course years reference to LPs, add it to CIPs.

This requires some changes to the way we seed data - foreign keys are no fun when trying to dump data. Nothing major though. 

### Guidance to review
Running seed command currently removes CIPS that don't have particular ids. That's because the only way I found to dump foreign key relations automatically is to just dump the ids. And that means our CIPs need to have the correct ids if we want to maintain their links to course years via seed dumping.

Luckily we can easily insert objects with arbitrary ids into our database - thanks Alfred for suggesting we use uuids, I really appreciate it now :)


lib/tasks/cip_seed_dump.rake
```diff
@@ -2,5 +2,5 @@
 
 desc "Dump CIP content into a file"
 task cip_seed_dump: :environment do
-  system "bundle exec rake db:seed:dump FILE=db/seeds/cip_seed_dump.rb EXCLUDE='[:created_at, :updated_at]' MODELS='LeadProvider, CourseYear, CourseModule, CourseLesson' IMPORT=true"
+  CoreInductionProgrammeExporter.new.run
```
kicferk1, line 5: The old task did not want to fire from controller in production env :(

db/seeds/dummy_structures.rb
```diff
@@ -23,16 +23,10 @@
   user = User.find_or_create_by!(email: "ecf@mailinator.com") do |u|
     u.full_name = "Admin User"
   end
+  user.confirm
```
ethanmills, line 26: Yay!

app/models/core_induction_programme.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
 
 class CoreInductionProgramme < ApplicationRecord
+  has_many :course_years
```
seb-siem-softwire, line 4: If we destroy Core Induction Programme - do we expect the associated records to be deleted as well?

https://stackoverflow.com/questions/29560805/how-to-use-dependent-destroy-in-rails

app/controllers/core_induction_programme_controller.rb
```diff
@@ -7,7 +7,8 @@ def show
 
   def download_export
     if @current_user&.admin?
-      system "bundle exec rake cip_seed_dump"
+      CoreInductionProgrammeExporter.new.run
```
seb-siem-softwire, line 10: Nice and clean!

ethanmills (approved):
LGTM

lib/tasks/cip_seed_dump.rake
seb-siem-softwire: Nice and clean

app/models/core_induction_programme.rb
kicferk1: Hmm, 'who knows' is the short answer. 

From a dev perspective, probably yes. 

From user perspective, I don't think we got to the point of figuring out whether we want to create / delete cips at all. But when we eventually do, I think this whole thing will change into some many-to-many relation. 

For time being I will add a cascade delete here. Good spot :)

seb-siem-softwire: Sweet!

seb-siem-softwire (approved):

amazimbe (approved):

PR #60, opened 2021-01-19
ECF-RP-115: Create suppliers

ethanmills:
### Context
DfE admins need to create new suppliers (LPs&DPs). This is that.

### Changes proposed in this pull request
[Wireframes](https://miro.com/app/board/o9J_ldVNkCY=/?moveToWidget=3074457353201992591&cot=14)

### Guidance to review
Commit by commit

PR #74, opened 2021-02-01
ECF-RP-134: Create users for suppliers

ethanmills:
### Context
Allow DfE administrators to create users for a lead provider or delivery partner
Ticket: https://trello.com/c/pCqb60yC/134-sprint-4-admins-can-create-users-for-suppliers
Miro: https://miro.com/app/board/o9J_ldVNkCY=/?moveToWidget=3074457353712711111&cot=14

Pretty similar multistep form flow to creating a supplier
I have omitted the role screen from the miro flow for now, since we don't really know what that means

### Guidance to review
Commit by commit


app/forms/supplier_user_form.rb
```diff
@@ -0,0 +1,45 @@
...
+  include ActiveModel::Serialization
+
+  attr_accessor :full_name, :email, :supplier
+  validates :supplier, presence: { message: "Select one" }, on: :supplier
+  validates :full_name, presence: { message: "Enter a name" }, on: :details
+  validates :email, presence: { message: "Enter email" }, on: :details
+  validate :email_not_taken, on: :details
+
+  def attributes
+    { "full_name" => nil, "email" => nil, "supplier" => nil }
```
kicferk1, line 14: We can probably use the new ruby way of creating hashes 
```
{ full_name: nil, ... }
```

app/forms/supplier_user_form.rb
```diff
@@ -0,0 +1,45 @@
...
+  def chosen_supplier
+    LeadProvider.find_by(id: supplier) || DeliveryPartner.find_by(id: supplier)
+  end
+
+  def save!
+    profile = if chosen_supplier.instance_of?(LeadProvider)
+                LeadProviderProfile.new(lead_provider: chosen_supplier)
+              else
+                DeliveryPartnerProfile.new(delivery_partner: chosen_supplier)
+              end
```
kicferk1, line 26: Ech, I really dislike that we manage LeadProviders and DeliveryPartners in the same places and in the same forms :(

But I guess it's what our designers want.

app/forms/supplier_user_form.rb
```diff
@@ -0,0 +1,45 @@
...
+    ActiveRecord::Base.transaction do
+      @user = User.create!(
+        full_name: full_name,
+        email: email,
+        confirmed_at: Time.zone.now.utc, # Skip confirmation email
+      )
+      profile.user = @user
+      profile.save!
+    end
+    @user
```
kicferk1, line 37: Does this need to be a `@` variable?

app/controllers/admin/supplier_users_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+class Admin::SupplierUsersController < Admin::BaseController
+  skip_after_action :verify_authorized, only: :index
+  skip_after_action :verify_policy_scoped, except: :index
+
+  before_action :set_suppliers, only: %i[new receive_supplier]
+  before_action :form_from_session, only: %i[user_details review create success]
+
+  def index
+    lead_providers = policy_scope(LeadProvider).joins(:users).includes(:users)
+    delivery_partners = policy_scope(DeliveryPartner).joins(:users).includes(:users)
```
kicferk1, line 12: do we need both `join` and `includes` users?

app/views/admin/supplier_users/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      <th scope="col" class="govuk-table__header">Email address</th>
+      <th scope="col" class="govuk-table__header">Organisation</th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+    <% @users.each do |user| %>
+      <tr class="govuk-table__row">
+        <td class="govuk-table__cell"><%= govuk_link_to user.full_name, "#" %></td>
+        <td class="govuk-table__cell"><%= user.email %></td>
+        <td class="govuk-table__cell"><%= user.supplier_name %></td>
```
kicferk1, line 16: Does that lead to n+1 queries? Or is Rails clever enough to remember it loaded the suppliers before it loaded users (in controller)

app/views/admin/suppliers/_layout.html.erb
```diff
@@ -0,0 +1,20 @@
+<h1 class="govuk-heading-l">Suppliers</h1>
+
+
```
kicferk1, line 3: I hate whitespace indeed

app/views/admin/suppliers/_layout.html.erb
```diff
@@ -0,0 +1,20 @@
...
+<div style="display: flex; flex-direction: row">
```
kicferk1, line 4: Let's move that to css file.

kicferk1: Or use something like govuk grid row, and two columns with width half inside of it.

spec/requests/admin/supplier_users_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+  let(:full_name) { Faker::Name.name }
+  let(:email) { Faker::Internet.email }
+
+  before do
+    admin_user = create(:user, :admin)
+    sign_in admin_user
+  end
+
+  describe "GET /admin/suppliers/users" do
+    it "renders the correct template" do
```
kicferk1, line 17: `correct template` should probably be specifying which template exactly.

`renders the index template`?

spec/requests/admin/supplier_users_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+      get "/admin/suppliers/users"
+
+      expect(response).to render_template(:index)
+    end
+  end
+
+  describe "GET /admin/suppliers/users/new" do
+    it "renders the correct template" do
```
kicferk1, line 25: Ditto here and around this file.

app/views/layouts/application.html.erb
```diff
@@ -17,6 +17,7 @@
     <%= stylesheet_pack_tag 'application', media: 'all' %>
     <%= javascript_pack_tag 'application', defer: true %>
     <%= stylesheet_link_tag 'application', media: 'all' %>
+    <%= yield(:javascript_pack_tags) %>
```
kicferk1, line 20: Do we need this? 

I would think `<%= javascript_pack_tag 'application', defer: true %>` above would add all webpacker javascript, including autcomplete. Maybe we need to add `supplier-users` to `application.js`?

app/controllers/admin/supplier_users_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+    sorted_users = (lead_providers + delivery_partners).flat_map(&:users).sort_by(&:full_name)
+    @users = Kaminari.paginate_array(sorted_users).page(params[:page]).per(20)
+    @page = @users.current_page
+    @total_pages = @users.total_pages
+  end
+
+  def new
+    authorize User, :new?
```
kicferk1, line 20: It's not just user that we want to authorise here right? We are gonna create some profiles as well, and that's really the part that needs authorisation.

app/views/admin/supplier_users/new.html.erb
```diff
@@ -0,0 +1,22 @@
...
+
+<% content_for :before_content, govuk_back_link(
+  text: "Back",
+  href: admin_supplier_users_path)
+%>
+
+<h1 class="govuk-heading-l">What supplier does this user belong to?</h1>
+
+<%= form_with model: @supplier_user_form, url: new_admin_supplier_user_path, method: :post do |f| %>
+  <%= f.govuk_collection_select(
```
kicferk1, line 11: So is this enhanced by autocomplete? Or is it some other field?

Does it work for collection select? :O

app/forms/supplier_user_form.rb
ethanmills: I want the keys to be strings, not symbols

app/forms/supplier_user_form.rb
ethanmills: 🤷 wireframes, innit

app/controllers/admin/supplier_users_controller.rb
ethanmills: See https://github.com/DFE-Digital/early-careers-framework/pull/60#discussion_r564694120

package.json
```diff
@@ -7,6 +7,7 @@
   },
   "dependencies": {
     "@rails/webpacker": "^5.2.1",
+    "accessible-autocomplete": "^2.0.3",
```
amazimbe, line 10: What widget is this js for?

app/forms/supplier_user_form.rb
ethanmills: Yes, since I create it inside the transaction block and want to return it. Out of scope otherwise

app/views/admin/suppliers/index.html.erb
```diff
@@ -1,8 +1,5 @@
-<% content_for :before_content, govuk_breadcrumbs(breadcrumbs: {
-  "Dashboard" => admin_dashboard_path,
-  "Manage suppliers" => nil
-}) %>
-<h1 class="govuk-heading-l">Manage suppliers</h1>
+<%= render "layout", locals={page: "suppliers"} %>
```
amazimbe, line 1: I'm surprised rubocop didn't complain about {page: "suppliers"} not being spaced out correctly

app/forms/supplier_user_form.rb
kicferk1: Will something like 
```
    user = ActiveRecord::Base.transaction do
      user = User.create!(
        full_name: full_name,
        email: email,
        confirmed_at: Time.zone.now.utc, # Skip confirmation email
      )
      profile.user = user
      profile.save!
      user
    end
    user
```
work?

app/forms/supplier_user_form.rb
amazimbe: Just curious why you specifically want strings.

app/forms/supplier_user_form.rb
```diff
@@ -0,0 +1,45 @@
...
+                LeadProviderProfile.new(lead_provider: chosen_supplier)
+              else
+                DeliveryPartnerProfile.new(delivery_partner: chosen_supplier)
+              end
+
+    ActiveRecord::Base.transaction do
+      @user = User.create!(
+        full_name: full_name,
+        email: email,
+        confirmed_at: Time.zone.now.utc, # Skip confirmation email
```
amazimbe, line 32: Why do we need to confirm their accounts? I'd have expected them to do this themselves, if anything to confirm the email address they will use to sign in.

ethanmills: This is what was designed. The expectation is that this account may be created months before it's first used, so any confirmation link would expire. We also expect the email entered by the administrator to be correct

ethanmills: See https://miro.com/app/board/o9J_ldVNkCY=/?openComment=3074457353853155754

app/views/admin/suppliers/index.html.erb
ethanmills: Doesn't lint erb files :/

package.json
ethanmills: This https://miro.com/app/board/o9J_ldVNkCY=/?moveToWidget=3074457353712710918&cot=14

app/views/admin/supplier_users/new.html.erb
ethanmills: Yes. I think to most of those questions...

ethanmills: Try it out 🙂 https://ecf-review-pr-74.london.cloudapps.digital/admin/suppliers/users/new

ethanmills: image

app/views/layouts/application.html.erb
ethanmills: I was trying to avoid loading it on every page 🤷

app/views/admin/supplier_users/index.html.erb
ethanmills: It does, not sure how else to do it though

ethanmills: Specific DTO type thing for the view? How much do we care?

kicferk1: I mean, we care a little. I would not say enough for a separate DTO, but maybe enough to examine how we are loading the data. 

If I understand it right, we load the providers, then we load their users, then we sort the users.

I wonder if what we want to do instead is load the users who have the right profiles, along the suppliers they have through the profiles. Then sort them and return them. 

Thoughts?

app/views/layouts/application.html.erb
kicferk1: Oh, my bad, didn't read it thoroughly enough. 

How big is the autocomplete js? Is it big enough that we want to justify the extra complexity of working with it introduced by only loading it when it is used?

app/forms/supplier_user_form.rb
ethanmills: Misunderstood the [docs](https://api.rubyonrails.org/classes/ActiveModel/Serialization.html). They had strings, so I figured that's how the underlying methods worked

app/views/admin/supplier_users/index.html.erb
ethanmills: That to me looks something like 
```
User
                     .left_outer_joins(:lead_provider_profile)
                     .left_outer_joins(:delivery_partner_profile)
                     .where.not(lead_provider_profile: { lead_provider_id: nil })
                     .or(User
                           .left_outer_joins(:lead_provider_profile)
                           .left_outer_joins(:delivery_partner_profile)
                           .where.not(delivery_partner_profile: { delivery_partner_id: nil })
                     )
```
Would you agree?

kicferk1: Do we need to check lead provider . delivery partner ids not being nil? 

Surely we don't allow provider profiles with no provider?

kicferk1: We could also have two queries, one for lead provider users and one for delivery partner users. Better than N+1 :P

ethanmills: That doesn't solve the N+1 though

ethanmills: I don't know how to do that

kicferk1:
```
lead_provider_users = User.joins(:lead_provider_profile).includes(:lead_provider)
delivery_partner_users = Users.join(:delivery_partner_profile).includes(:delivery_partner)
<<some code to sort users by full name in ruby>>
```
Is where I would start from.

kicferk1: Happy to pair if we somehow find some time when we are free

app/views/layouts/application.html.erb
ethanmills: Looks to be about 61kb minified. I thought it was ~400kb, but that's the dev-server

app/controllers/admin/supplier_users_controller.rb
ethanmills: You happy for that to go in the create? When we know what type of profile we are creating

kicferk1: Did I mention how much I hate that we have different entities being managed in the same place?

Fine, create is a good place for that.

app/views/layouts/application.html.erb
kicferk1: My personal opinion (after looking through some more mature project) is that our js is not really complex enough to selectively load it. And I would lean towards 61kb not being large enough to warrant its own load.

ethanmills: Will move. I just feels _wrong_ somehow to have it loading and running on every page 🤷

app/controllers/admin/supplier_users_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+      authorize DeliveryPartnerProfile
+    end
+    user = @supplier_user_form.save!
+    redirect_to admin_new_supplier_user_success_path(user_id: user.id)
+  end
+
+  def success
+    @user = User.find(params[:user_id])
+    authorize @user, :show?
+    @supplier = @supplier_user_form.chosen_supplier.name
```
kicferk1, line 75: `@supplier_name`? I know, nitpick.

app/views/admin/suppliers/_layout.html.erb
```diff
@@ -0,0 +1,18 @@
+<h1 class="govuk-heading-l">Suppliers</h1>
+<div class="page-tab-container">
+  <div class="page-tab govuk-!-margin-right-8 <%= "page-tab__selected" if page == "suppliers" %>">
+    <% if page == "suppliers" %>
+      All suppliers
+    <% else %>
+      <%= govuk_link_to "All suppliers", admin_suppliers_path %>
+    <% end %>
+  </div>
```
kicferk1, line 9: This is like a page navbar, right? Kind of like the one on https://design-system.service.gov.uk/components/?

If it makes any difference, they don't bother with making current element not a link - and it looks a bit odd to see two if statements checking the same thing in two consecutive lines.

spec/requests/users/sessions_spec.rb
```diff
@@ -7,23 +7,23 @@
 
   describe "POST /users/sign_in" do
     context "when sign in email has been sent" do
-      it "renders the correct template" do
+      it "renders the login_email_sent template" do
```
kicferk1, line 10: Nice and thorough :)

kicferk1 (commented):
Just some nitpicks

app/views/admin/suppliers/_layout.html.erb
ethanmills: I take issue with that navigation: there is no indication of the current page to a screen reader user. @cfq do you know of good patterns for this?

ethanmills: I expect something like leaving it as a link and adding a `visually-hidden` span saying "Current page" or similar might work?

kicferk1: I think it would, yes :)

app/views/admin/suppliers/_layout.html.erb
```diff
@@ -1,17 +1,13 @@
 <h1 class="govuk-heading-l">Suppliers</h1>
 <div class="page-tab-container">
   <div class="page-tab govuk-!-margin-right-8 <%= "page-tab__selected" if page == "suppliers" %>">
-    <% if page == "suppliers" %>
-      All suppliers
-    <% else %>
-      <%= govuk_link_to "All suppliers", admin_suppliers_path %>
+    <%= govuk_link_to admin_suppliers_path do %>
+      All suppliers <%= "<span class='govuk-visually-hidden'>Current page</span>".html_safe if page == "suppliers" %>
```
kicferk1, line 5: I think it's good enough solution for now.

For the future, if we want to keep using this navigation more often, I think we should make a more general component. Pity it's not a thing in govuk design system.

kicferk1 (approved):

app/views/admin/suppliers/_layout.html.erb
ethanmills: How's this @kicferk1 ?

app/views/admin/suppliers/_layout.html.erb
ethanmills: Agreed

amazimbe (approved):

spec/requests/admin/supplier_users_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+    it "redirects to the supplier type page" do
+      when_i_choose_a_supplier(lead_provider)
+
+      expect(response).to redirect_to("/admin/suppliers/users/new/user-details")
+    end
+
+    it "Sets the correct lead provider" do
+      when_i_choose_a_supplier(lead_provider)
+
+      # Then
```
seb-siem-softwire, line 42: Gherkin language ;) Kudos for that - the comment can be removed though - the tests are self-explanatory

spec/requests/admin/supplier_users_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+      given_i_have_entered_user_details(full_name, email)
+      given_i_have_confirmed_my_choices
+
+      expect(User.find_by_email(email).lead_provider).to eq lead_provider
+    end
+
+    it "Sets the correct delivery partner" do
+      when_i_choose_a_supplier(delivery_partner)
+
+      # Then
```
seb-siem-softwire, line 52: You can remove the comment - the code already tells a story

spec/requests/admin/supplier_users_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+      when_i_enter_user_details(full_name, email)
+
+      # Then
+      expect(response).to redirect_to("/admin/suppliers/users/new/review")
+    end
+
+    it "sets the correct user details" do
+      when_i_enter_user_details(full_name, email)
+
+      # Then
```
seb-siem-softwire, line 95: To not repeat yourself too many times - your tests are very clear and easy to read - no need for putting additional comments

kicferk1: Boom, PR merged anyway 😂 

I guess Ethan didn't check his email in time...

PR #75, opened 2021-02-01
ECF-Tech-43: Mock login on dev

ethanmills:
### Context
Ticket: https://trello.com/c/VJjCt8n7/43-mock-out-login-for-dev
We want to bypass login on pr apps and dev

Create admin@example.com, lead-provider@example.com, school-leader@example.com
Bypass login on dev and deployed_dev


db/seeds/dummy_structures.rb
```diff
@@ -19,12 +19,25 @@
...
+    u.confirmed_at = Time.zone.now.utc
   end
-  user.confirm
   AdminProfile.create!(user: user)
+
+  user = User.find_or_create_by!(email: "lead-provider@example.com") do |u|
+    u.full_name = "Lp User"
+    u.confirmed_at = Time.zone.now.utc
+  end
+  user.confirm
```
kicferk1, line 33: That can be canned now I think

amazimbe:
The code changes look good I'm just not sure this is a good idea. Personally, I prefer the code working in the same way regardless of where it is run. My fear is we could have a broken sign in feature and not realise it for a long time as prod would be different from dev / local.

ethanmills:
> The code changes look good I'm just not sure this is a good idea. Personally, I prefer the code working in the same way regardless of where it is run. My fear is we could have a broken sign in feature and not realise it for a long time as prod would be different from dev / local.

I understand your concerns. I'm not planning on merging this any time soon, I think is something we need to talk about as a team - this is just a talking point.

It's important to note that we also have a staging environment 🙂

kicferk1:
Even if we don't merge it, it would be good to have different users with different roles.

seb-siem-softwire:
@amazimbe I hear your concerns Alfred - which is the reason for having a comprehensive test suite. At the same time an angle that we could look at it - is - do we really want to test sign-in functionality with every test? Does the additional time running the test suite justifies it when we could be testing login functionality both independently and also as a smoke test. Ethan raised a good point where we can discuss it as a team and see the issue from multiple perspectives and hear each other out. Given the experience that we have and also as problem solvers - there is always more than one way to approach it. Shall we pick it up tomorrow(or today - or well whenever someone reads this comment)

amazimbe:
> @amazimbe I hear your concerns Alfred - which is the reason for having a comprehensive test suite. At the same time an angle that we could look at it - is - do we really want to test sign-in functionality with every test? Does the additional time running the test suite justifies it when we could be testing login functionality both independently and also as a smoke test. Ethan raised a good point where we can discuss it as a team and see the issue from multiple perspectives and hear each other out. Given the experience that we have and also as problem solvers - there is always more than one way to approach it. Shall we pick it up tomorrow(or today - or well whenever someone reads this comment)

We could discuss it at one of our afternoon team catchups.

app/controllers/users/sessions_controller.rb
```diff
@@ -1,11 +1,19 @@
...
   class LoginIncompleteError < StandardError; end
 
   def create
-    super
+    if Rails.env.development? || Rails.env.deployed_development?
+      user = User.find_by_email(params.dig(:user, :email))
+      sign_in(user, scope: :user)
+      redirect_to profile_dashboard_url(user)
+    else
+      super
```
kicferk1, line 15: Can we do something more ugly here? Like, check if the email of the user is in the list of permitted test values, check if we are in dev / deployed dev, and only then log them in without email?

I think that's the bit that we discussed yesterday - so that our test accounts get in without fuss, but login mechanism still works for other users (in particular, I would keep `ecf@mailinator.com` as not automatically logged in).

ethanmills: Where do you think would be a good place to maintain that list of emails?

kicferk1: Wherever convenient and not disgusting.

Constant in the controller is fine by me.

db/seeds/dummy_structures.rb
kicferk1: Can still be canned.

kicferk1 (approved):
LGTM, don't think we need to both confirm and set confirmed time for a user in the seeding file, but that's a tiny nitpick.

db/seeds/dummy_structures.rb
amazimbe: Indeed! Even if it weren't for this change, it would still be unnecessary as all it does is set confirmed_at

amazimbe (approved):

rupollbabu (approved):
Able to Login using 3 roles(Admin, Lead-provider ,school-provider) in DEV env.

kicferk1 (approved):
LGTM, pls merge so I can add another user in another PR :)

PR #77, opened 2021-02-02
[ECF-1-131] Allow registrations if school is not confirmed within 24 hours

amazimbe:
### Context
When an induction coordinator registers a school they get a confirmation email with a confirmation link. They are given 24hrs to confirm otherwise we'll allow other users to register the school.

What should happen to the user who didn't confirm registration? That is up for discussion but my view would be to remove the induction coordinator profile as well as user profile.

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -16,10 +16,10 @@ class Users::RegistrationsController < Devise::RegistrationsController
     render :user_already_registered and return if @user.persisted?
     render_start_registration and return if @school.nil?
     render :school_not_eligible and return unless @school.eligible?
-    render :school_not_registered and return if @school.not_registered?
     render :school_fully_registered and return if @school.fully_registered?
+    render :school_partially_registered and return if @school.partially_registered?
 
-    render :school_partially_registered if @school.partially_registered?
+    render :school_not_registered
   end
 
   def create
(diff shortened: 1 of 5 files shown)


app/controllers/users/registrations_controller.rb
```diff
@@ -28,6 +28,7 @@ def create
 
     ActiveRecord::Base.transaction do
       resource.save!
+      @school.induction_coordinator_profiles.destroy_all
```
amazimbe, line 31: I'm not sure what we need to do with the coordinator who hasn't confirmed their email. Here I'm just removing them from the school. They remain induction coordinators but they no longer belong to a school.  Not sure what the best thing to do should be. We could just destroy the user record and profile.

spec/requests/users/registrations/register_spec.rb
```diff
@@ -90,5 +90,24 @@
         } }
       }.to raise_error(ActionController::BadRequest)
     end
+
+    context "when school has old unconfirmed registrations" do
+      let(:user) { create(:user, confirmed_at: nil) }
+      let!(:coordinator) { create(:induction_coordinator_profile, user: user, schools: [school]) }
+      let(:created_user) { User.find_by_email(email) }
```
kicferk1, line 97: I think for a single test we can put
```
      created_user = User.find_by_email(email)
```
in the test.

On the other hand, that line is repeated a few times throughout this test file - so maybe we want `created_user` to be defined and used one context higher?

kicferk1 (approved):
LGTM

app/controllers/users/registrations_controller.rb
seb-siem-softwire: Who do you think would be the best person to answer the question - the designers? BA's - It's a valid question and it seems like it is not about technical implementation but the product functionality

PR #78, opened 2021-02-02
Add editing module content

xswoodsx:
### Context

We need to include functionality for editing module content. The changes included in the PR enable DfE admin to do this. 

- enable editing of module content
- including request and policy tests for being able to edit module content
- refactoring of cip routes to improve readability and minimise duplication.
- amending cip request tests to reflect updates made in the routes. 
- amending cip views to reflect changes made in the routes

### Guidance to review
Much of what has been used has been replicated from what is done editing lessons.

### DIFF
--- a/app/controllers/core_induction_programme/modules_controller.rb
+++ b/app/controllers/core_induction_programme/modules_controller.rb
@@ -18,7 +18,7 @@ class CoreInductionProgramme::ModulesController < ApplicationController
   def update
     if params[:commit] == "Save changes"
       @course_module.update!(content: params[:module_preview])
-      redirect_to cip_years_modules_url
+      redirect_to cip_year_module_url
     else
       @module_preview = params[:module_preview]
       @html_content = content_to_html(@module_preview)
@@ -29,7 +29,7 @@ class CoreInductionProgramme::ModulesController < ApplicationController
 private
 
   def load_and_authorize_course_module
-    @course_module = CourseModule.find(params[:module_id])
+    @course_module = CourseModule.find(params[:id])
     authorize @course_module
   end
 end
--- a/app/controllers/core_induction_programme/modules_controller.rb
+++ b/app/controllers/core_induction_programme/modules_controller.rb
@@ -18,6 +18,7 @@ class CoreInductionProgramme::ModulesController < ApplicationController
   def update
     if params[:commit] == "Save changes"
       @course_module.update!(content: params[:module_preview])
+      flash[:success] = "Your changes have been saved"
       redirect_to cip_year_module_url
     else
       @module_preview = params[:module_preview]
--- a/app/controllers/core_induction_programme/modules_controller.rb
+++ b/app/controllers/core_induction_programme/modules_controller.rb
@@ -1,7 +1,35 @@
 # frozen_string_literal: true
 
 class CoreInductionProgramme::ModulesController < ApplicationController
-  def show
-    @course_module = CourseModule.includes(:course_lessons).find(params[:module_id])
+  include Pundit
+  include GovspeakHelper
+
+  after_action :verify_authorized, except: :show
+  before_action :authenticate_user!, except: :show
+  before_action :load_and_authorize_course_module
+
+  def show; end
+
+  def edit
+    @module_preview = params[:module_preview] || @course_module.content
+    @html_content = content_to_html(@module_preview)
+  end
+
+  def update
+    if params[:commit] == "Save changes"
+      @course_module.update!(content: params[:module_preview])
+      redirect_to cip_years_modules_url
+    else
+      @module_preview = params[:module_preview]
+      @html_content = content_to_html(@module_preview)
+      render action: "edit"
+    end
+  end
+
+private
+
+  def load_and_authorize_course_module
+    @course_module = CourseModule.find(params[:module_id])
+    authorize @course_module
   end
 end
--- a/app/views/core_induction_programme/modules/edit.html.erb
+++ b/app/views/core_induction_programme/modules/edit.html.erb
@@ -1,7 +1,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
     <h1 class="govuk-heading-l">Content change preview</h1>
-    <%= form_with url: cip_years_modules_path, method: :put do |f| %>
+    <%= form_with url: cip_year_module_path, method: :put do |f| %>
       <%= f.govuk_text_area :module_preview, label: { text: "Markdown string for course module content" }, rows: 10, value: @module_preview %>
       <%= f.govuk_submit "See preview" %>
       <%= f.govuk_submit "Save changes" %>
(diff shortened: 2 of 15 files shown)


config/routes.rb
```diff
@@ -86,13 +86,16 @@
...
-          member do
-            put "edit"
-          end
-        end
+
+    concern :editable do
+      member do
+        put "edit"
+      end
+    end
```
kicferk1, line 94: Interesting, what do we need it for?

spec/requests/core_induction_programme/course_module_spec.rb
```diff
@@ -3,9 +3,79 @@
...
+    end
+
+    describe "GET /core-induction-programme/years/:years_id/modules/module_id" do
+      it "renders the cip module page" do
+        get course_module_url
+        expect(response).to render_template(:show)
+      end
+    end
+
+    describe "GET /core-induction-programme/years/:years_id/modules/module_id/edit" do
```
kicferk1, line 22: There is a break between tests here, but between some other tests down below. Let's make it consistent :)

app/controllers/core_induction_programme/modules_controller.rb
```diff
@@ -1,7 +1,35 @@
...
+  def show; end
+
+  def edit
+    @module_preview = params[:module_preview] || @course_module.content
+    @html_content = content_to_html(@module_preview)
+  end
+
+  def update
+    if params[:commit] == "Save changes"
+      @course_module.update!(content: params[:module_preview])
```
amazimbe, line 20: Just curious what should happen if updating fails for some reason.

kicferk1: I imagine "nobody knows" is the answer. All of content editing is a bit speculative, and doesn't have designs - mainly because we will likely use it internally before release a lot more than it will be used afterwards.

So I think for time being "user sees an error caused by failed update" is fine.

spec/requests/core_induction_programme/course_module_spec.rb
```diff
@@ -3,9 +3,79 @@
...
+      it "renders a preview of changes to module" do
+        put "#{course_module_url}/edit"
+
+        expect(response).to render_template(:edit)
+      end
+    end
+    describe "PUT /core-induction-programme/years/:years_id/modules/module_id" do
+      it "redirects to the module page when saving content" do
+        put course_module_url, params: { commit: "Save changes", module_preview: course_module.content }
+        expect(response).to redirect_to(course_module_url)
```
amazimbe, line 39: I think you should also test some side effects here like changes to the record in the database.

config/routes.rb
xswoodsx: As i understand it, it adds the put "edit" in to each nested resource, reducing the duplication of putting a member in each resource.

app/views/core_induction_programme/modules/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <h1 class="govuk-heading-l">Content change preview</h1>
+    <%= form_with url: cip_year_module_path, method: :put do |f| %>
```
ethanmills, line 4: Does this work without js? Do we care?

config/routes.rb
xswoodsx: Yeah, if we remove it the test ```put "#{course_lesson_url}/edit"``` fails, saying there's no path that matches. 

The editable concern adds  
```edit_cip_year_module_lesson PUT    /core-induction-programme/years/:year_id/modules/:module_id/lessons/:id/edit(.:format)            core_induction_programme/lessons#edit```. Without it we only have a get with this path. 

I had a play around with it to try and not use it but I think preview rendered again but it removed the /edit from the end of the url still in the preview. This was the way that I found keeps it in.

File: app/controllers/core_induction_programme/modules_controller.rb [hunk 6]
```diff
@@ -1,7 +1,35 @@
...
+      redirect_to cip_year_module_url
```
amazimbe, line 21: So when we redirect we won't even show a message to the user than the update was a success?

kicferk1: They will be redirected to the page they just edited, where the success should be easily verifiable.

Are you thinking a flash message? I definitely don't want to add a confirmation page for it.

app/views/core_induction_programme/modules/edit.html.erb
kicferk1: Fascinatingly enough, when I disabled js, updating a module actually worked.

Preview didn't though :/ 

Might be because we are using a different route for preview (one that we defined ourselves), and Rails has some magic to handle default resource requests that fire off without js.

app/controllers/core_induction_programme/modules_controller.rb
amazimbe: Yes, a flash message exactly. Confirmation pages are best left alone unless the designers demand them.

config/routes.rb
xswoodsx: The concern has been removed as it wasn't necessary to include.

app/views/core_induction_programme/modules/edit.html.erb
xswoodsx: Removing the concern in routes has done something so that preview now renders with JS disabled.

app/controllers/core_induction_programme/modules_controller.rb
xswoodsx: i've done a notice that adds a message on the redirected page.  Are we bothered whether it's an alert or notice? 
Screenshot 2021-02-03 at 11 43 05

ethanmills: This is a decision for a designer. I expect we don't need this at all, see [this thread](https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1612181773094100?thread_ts=1612181226.088100&cid=C6DMEH5R6)

ethanmills: > I kinda think you only need these Success messages if it’s not otherwise obvious from the page. Like, you don’t need one for updating a thing, if the page shows you that updated thing anyway.

ethanmills: I've invited Dom to the repo, once she accepts we should be able to tag her on discussions like this. Might be worth doing the same for Mark

amazimbe (approved):

app/controllers/core_induction_programme/modules_controller.rb
xswoodsx: Thanks for sharing that Ethan. I had a chat with Mark A and he said as long as the message is clear it doesn't matter to much whether it's a notice or alert. I'll include it for now.

kicferk1 (approved):

PR #79, opened 2021-02-03
Add ect profile

kicferk1:
### Context
One of sprint goals on engage and learn this (next?) sprint is adding an ECT dashboard and tracking ECT progress through CIP. Before we can do any of that, we need to add something representing an ECT.

Add a new profile, `EarlyCareerTeacherProfile`. 

Make it link to reasonable things - like user, school, cohort and cip. One day we might rethink whether the connection to cip lives on the profile, on the school, or on the cohort (or somewhere else entirely) - but for time being I opted for the simplest solution for the upcoming work.

### Guidance to review
I moved the order of seed files loaded, because I want core induction programmes to exist by the time I generate my dummy ECT.

### DIFF
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class Cohort < ApplicationRecord
+  has_many :early_career_teacher_profiles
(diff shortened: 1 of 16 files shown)


amazimbe (approved):
LGTM

PR #83, opened 2021-02-04
Lesson order in modules

kicferk1:
### Context
After editing a lesson, it messes up the order of lessons displayed. That's bad. 

I found an amazing fix, which is short enough to fit inside a margin!

I just don't like it much. 

Turns out the underlying reason for incorrect order was another bug - `next_lesson` always being `nil`. Why would it be nil you ask? Because we had an unused reference (that I forgot to remove) - and because when Rails asked me for a foreign key for `next_lesson`, I mindlessly put in the reference that matched the name - yup, the one that wasn't used and hence was empty. So the one-liner, margin-fitting fix is to correct the name of the foreign key.

The cleanup is to write some migrations removing it for good, and hence the change in the seeding files. 

Ok, so far so good. But we haven't gotten to the best part of the story yet!

While I was working on a fix, I thought "maybe I should add a custom ordered lessons function", so I wrote the following in `CourseModule`:
```
  def course_lessons_in_order
    preloaded_lessons = course_lessons.includes(:previous_lesson, :next_lesson)

    first_lesson = preloaded_lessons.find { |lesson| !preloaded_lessons.include?(lesson.previous_lesson) }
    return [] unless first_lesson

    ordered_lessons = [first_lesson]
    current_lesson = first_lesson
    next_lesson = preloaded_lessons.find { |lesson| lesson.previous_lesson == current_lesson }

    while preloaded_lessons.include?(next_lesson)
      ordered_lessons += [next_lesson]
      current_lesson = next_lesson
      next_lesson = preloaded_lessons.find { |lesson| lesson.previous_lesson == current_lesson }
    end

    ordered_lessons
  end
```
Which led me to figure out that something funky was going on with next lessons.

So I fixed it by giving the relation correct foreign key, and I decided to write some tests for my new shiny method. And the tests tell me the order of `course_lessons` was the same as the order of my fancy function - whaaaat?

Turns out it's the result of `inverse_of: :next_lesson` on the `previous_lesson` relation. When I remove it, the order differs.

And that is the scary thing - is this a 'well-known fact'? Or just a quirk of underlying Rails mechanisms? Do we feel confident in relying on it, or should we add the functions like the one above to CourseModule and CourseYear to be on the safe side?


app/models/course_module.rb
```diff
@@ -2,10 +2,14 @@
 
 class CourseModule < ApplicationRecord
   belongs_to :course_year
-  has_one :next_module, class_name: "CourseModule", foreign_key: :next_module_id
-  belongs_to :previous_module, class_name: "CourseModule", inverse_of: :next_module, optional: true
   has_many :course_lessons, dependent: :delete_all
 
+  # We use previous_module_id to store the connections between lessons
```
amazimbe, line 7: Should that be connections between modules.

amazimbe: I just wonder if we should perhaps update the schema rather than adding the comments. Is this more a result of a bad schema.

kicferk1: > Should we replace lesson with module in the text.
Good spot!

> I just wonder if we should perhaps update the schema rather than adding the comments. Is this more a result of a bad schema.
I am happy to make a better schema, just don't know how. The little I understand about Rails tells me there isn't a pretty solution.

I want a chain of lessons. I think pointing to the previous lesson from current one makes perfect sense to achieve that. Hence `belongs_to` previous lesson, cause id lives on current lesson. (at least I think that's what Rails expects - id lives on our object = our object belongs to some other object)

Then if we want to access next lesson, it needs to be via `has_one`, cause we don't store that id on our current lesson. And if it has to be `has_one`, then it needs a field from the target object to tell it what is the name, on the target objet, to look for. And `previous_lesson_id` makes sense as well.

I don't think it's a bad schema, because it seems that the `next_lesson` requiring `previous_lesson_id` as foreign key is a perfectly natural result of Rails philosophy coupled with reasonable database structure. I put a comment in because it is counter-intuitive to me. And I don't know how to make it intuitive in Rails. 

We could, I guess, duplicate the connections - have both previous lesson and next lesson ids on the model - but then we risk them getting out of sync. And we store the same relation twice.

spec/models/course_year_spec.rb
```diff
@@ -23,4 +23,26 @@
...
+
+  describe "course_modules" do
+    it "returns modules in order after they get updated" do
+      course_year = FactoryBot.create(:course_year)
+      course_module_one = FactoryBot.create(:course_module, title: "One", course_year: course_year)
+      course_module_two = FactoryBot.create(:course_module, title: "Two", course_year: course_year)
+      course_module_three = FactoryBot.create(:course_module, title: "Three", course_year: course_year)
+      course_module_four = FactoryBot.create(:course_module, title: "Four", course_year: course_year)
+
+      course_module_two.previous_module = course_module_one
```
amazimbe, line 35: Just wondering if course_module_two.update!(previous_module: course_module_one) wouldn't be better

spec/models/course_module_spec.rb
```diff
@@ -25,4 +25,26 @@
...
+      course_lesson_two.previous_lesson = course_lesson_one
+      course_lesson_four.previous_lesson = course_lesson_two
+      course_lesson_three.previous_lesson = course_lesson_four
+      course_lesson_two.save!
+      course_lesson_three.save!
+      course_lesson_four.save!
+
+      expected_lessons_with_order = [course_lesson_one, course_lesson_two, course_lesson_four, course_lesson_three]
+      expected_lessons_with_order.zip(course_module.course_lessons).each do |expected, actual|
+        expect(expected).to eq(actual)
```
seb-siem-softwire, line 46: Hmm reads kind of funny - expect the expected and you never be disappointed - if this was a story and you would read it as a bedtime story - how could you improve the plotline ?

spec/models/course_year_spec.rb
```diff
@@ -23,4 +23,26 @@
...
+      course_module_four.previous_module = course_module_two
+      course_module_three.previous_module = course_module_four
+      course_module_two.save!
+      course_module_three.save!
+      course_module_four.save!
+
+      expected_modules_with_order = [course_module_one, course_module_two, course_module_four, course_module_three]
+      expected_modules_with_order.zip(course_year.course_modules).each do |expected, actual|
+        expect(expected).to eq(actual)
```
seb-siem-softwire, line 44: You can expect better expectation naming in here as well.

seb-siem-softwire:
:D Liked the suspense reading the story description

seb-siem-softwire:
Also - if it works - and it's railsy - it tends to be reliable. Good investigation. Reminds me of Miller from Expanse

amazimbe (approved):

seb-siem-softwire (approved):

PR #84, opened 2021-02-04
Sign in defects

ethanmills:
### Context

> Defect1:User able to land on sign in
> successful page with expire token
> Given I am on sign in page.
> When I entered emailId on sign in page
> And I click on Expired notification link in mail inbox.
> Then I am landing on “Sign in successful page”
> But I should land on sign in page.

 - Ensure no success message is displayed when clicking an invalid link
 - Ensure logged in users are redirected to the dashboard when clicking a sign-in link
 - Ensure logged in users are redirected to the dashboard when clicking Start now

### Guidance to review
Now added tests

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,7 +11,7 @@ class ApplicationController < ActionController::Base
   end
 
   def after_sign_in_path_for(user)
-    helpers.profile_dashboard_url(user)
+    stored_location_for(user) || helpers.profile_dashboard_url(user)
   end
 
 protected
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,6 +10,10 @@ class ApplicationController < ActionController::Base
     render json: { status: "OK", version: ENV["SHA"], environment: Rails.env }, status: :ok
   end
 
+  def after_sign_in_path_for(user)
+    helpers.profile_dashboard_url(user)
+  end
+
 protected
 
   def configure_permitted_parameters
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -34,13 +34,13 @@ private
   def ensure_login_token_valid
     @user = User.find_by(login_token: params[:login_token])
 
-    if @user.blank? || login_token_expired
+    if @user.blank? || login_token_expired?
       flash[:alert] = "There was an error while logging you in. Please enter your email again."
       redirect_to new_user_session_path
     end
   end
 
-  def login_token_expired
+  def login_token_expired?
     Time.zone.now > @user.login_token_valid_until
   end
 end
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -33,9 +33,14 @@ private
 
   def ensure_login_token_valid
     @user = User.find_by(login_token: params[:login_token])
-    unless @user.present? && Time.zone.now < @user.login_token_valid_until
+
+    if @user.blank? || login_token_expired
       flash[:alert] = "There was an error while logging you in. Please enter your email again."
       redirect_to new_user_session_path
     end
+  end
+
+  def login_token_expired
+    Time.zone.now > @user.login_token_valid_until
   end
 end
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -4,6 +4,9 @@ class Users::SessionsController < Devise::SessionsController
   class EmailNotFoundError < StandardError; end
   class LoginIncompleteError < StandardError; end
 
+  before_action :redirect_to_dashboard, only: %i[sign_in_with_token redirect_from_magic_link]
+  before_action :ensure_login_token_valid, only: %i[sign_in_with_token redirect_from_magic_link]
+
   def create
     super
   rescue LoginIncompleteError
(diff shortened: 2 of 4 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -13,19 +16,26 @@ def create
...
+
+private
+
+  def redirect_to_dashboard
+    redirect_to helpers.profile_dashboard_url(current_user) if current_user.present?
+  end
+
+  def ensure_login_token_valid
+    @user = User.find_by(login_token: params[:login_token])
+    unless @user.present? && Time.zone.now < @user.login_token_valid_until
```
seb-siem-softwire, line 36: Small stylistic for better readability - adding an empty line improves the speed of reading the code when there is a block. A case of oranges over apples. You could also put Time.zone.now < @user.login_token_valid_until in a variable for better readability or even put it to user class. An opinion - not a hard rule.

seb-siem-softwire:
Small comments otherwise its fine

app/controllers/users/sessions_controller.rb
amazimbe: I prefer to use ! @user.present?  in this case rather than  unless @user.present? . When we have a compound predicate with unless I can never quite remember if the unless applies to (@user.present? && Time.zone.now < @user.login_token_valid_until) or just to @user.present?.  I find ! @user.present? just clearer

app/controllers/application_controller.rb
```diff
@@ -10,6 +10,10 @@ def check
     render json: { status: "OK", version: ENV["SHA"], environment: Rails.env }, status: :ok
   end
 
+  def after_sign_in_path_for(user)
+    helpers.profile_dashboard_url(user)
```
amazimbe, line 14: Probably stored_location_for(user) || helpers.profile_dashboard_url(user) would be better.  stored_location_for(user) is the link that was stored in the session. With it we can send the user back to the page they were on before closing the browser. If there isn't one then we fallback on helpers.profile_dashboard_url(user)

app/controllers/users/sessions_controller.rb
kicferk1: Or maybe `if @user.blank? or Time.zone.now > @user.login_token_valid_until`, hehe. So many ways to write a conditional in Ruby.

I would second putting the login token time check outside of this line. My brain gets dizzy with more than two logical complexities at the same time - one being `unless` and the other joint statement without a clear name.

app/controllers/users/sessions_controller.rb
```diff
@@ -13,19 +16,26 @@ def create
...
   end
 
   def redirect_from_magic_link
     @login_token = params[:login_token] if params[:login_token].present?
   end
+
+private
+
+  def redirect_to_dashboard
+    redirect_to helpers.profile_dashboard_url(current_user) if current_user.present?
```
kicferk1, line 31: Is `current_user` always gonna be the same as `@user`? I think so, so maybe let's just use `@user` here?

app/controllers/users/sessions_controller.rb
amazimbe: if @user.blank? works for me too and is better than  ! @user.present?

app/controllers/users/sessions_controller.rb
ethanmills: Not in the before actions 🙂

app/controllers/users/sessions_controller.rb
ethanmills: This is one of the only lines I didn't change 😂 I'll smarten it up 👍

app/controllers/application_controller.rb
ethanmills: Ooh interesting, thanks! I think I'll double check with UX that that's the behaviour we want

ethanmills: Ah, I misunderstood what the method did. This sounds good 👍

app/controllers/users/sessions_controller.rb
```diff
@@ -13,19 +16,31 @@ def create
...
+
+    if @user.blank? || login_token_expired
       flash[:alert] = "There was an error while logging you in. Please enter your email again."
       redirect_to new_user_session_path
     end
   end
 
-  def redirect_from_magic_link
-    @login_token = params[:login_token] if params[:login_token].present?
+  def login_token_expired
```
amazimbe, line 43: If it returns a boolean then login_token_expired? would be more conventional and reads a tad better.

spec/requests/users/sessions_spec.rb
```diff
@@ -28,6 +46,32 @@
...
+
+    context "when the token has expired" do
+      before { user.update!(login_token_valid_until: 1.hour.ago) }
+
+      it "redirects to sign in when the token has expired" do
+        get "/users/confirm_sign_in?login_token=#{user.login_token}"
+
+        expect(response).to redirect_to "/users/sign_in"
+      end
+    end
```
kicferk1, line 64: I think that a context for a single test is a bit excessive.

Especially since we mention that token expired both in context and in test itself.

spec/requests/users/sessions_spec.rb
```diff
@@ -28,6 +46,32 @@
...
+
+    context "when already signed in" do
+      before { sign_in user }
+
+      it "redirects to the dashboard when already signed in" do
+        get "/users/confirm_sign_in?login_token=aaaaaaaaaa"
+
+        expect(response).to redirect_to "/dashboard"
+      end
+    end
```
kicferk1, line 74: Ditto, context for a single test with duplicated context message.

amazimbe (approved):
Looks good to me. Is it now time for the capybara end to end tests 😉  ?

spec/requests/users/sessions_spec.rb
kicferk1: Also, do we want to test for showing an error message?

kicferk1 (commented):
Looks pretty good, just some small things in tests

seb-siem-softwire (approved):

ethanmills:
> Looks good to me. Is it now time for the capybara end to end tests 😉 ?

This is time for @rupollbabu to tell us what's missing

spec/requests/users/sessions_spec.rb
ethanmills: I was just following @amazimbe 's lead with the other tests in this file

ethanmills: I think it makes it a lot easier to read actually

kicferk1: I'm happy to keep the context, but I don't think we need to duplicate the message "when the token has expired".

spec/requests/users/sessions_spec.rb
amazimbe: Personally, I prefer this way especially if one is actually setting the context like Ethan did in before { sign_in user }. There are times when people just add unnecessary contexts too but for me this is not one of them.

spec/requests/users/sessions_spec.rb
ethanmills: Ah yes, my mistake, added the context after. Good spot!

kicferk1 (approved):

rupollbabu (approved):
Retested Successfully. Meets the A.C

PR #86, opened 2021-02-05
ECF-RP-128: pupil premium

ethanmills:
### Context
[Ticket](https://trello.com/c/EJKi12x9/128-sprint-4-generate-fixture-from-pupil-premium-data)

- Set up model for pupil premium
- Import data from csv
- Add utility methods to School model

### Guidance to review
Commit-wise


app/services/school_data_importer.rb
```diff
@@ -4,6 +4,10 @@
 require "csv"
```
ethanmills, line 4: I moved this from /models to /services, since that seemed more appropriate. What do you think?

app/models/school.rb
```diff
@@ -42,6 +43,27 @@ def partially_registered?
...
+        OR pupil_premium_eligibilities.percent_secondary_pupils_eligible >= ?
+      )
+    QUERY
+    joins(:pupil_premium_eligibilities)
+      .where(
+        query,
+        start_year,
+        PupilPremiumEligibility::THRESHOLD_PERCENTAGE,
+        PupilPremiumEligibility::THRESHOLD_PERCENTAGE,
+      )
```
kicferk1, line 64: I wonder, is this part of the scope placed appropriately? 

Or should we make a scope for it on `PupilPremiumEligibility`? 

`only_uplifting_premiums` or something along those lines?

db/schema.rb
```diff
@@ -39,7 +39,7 @@
   end
 
   create_table "core_induction_programmes", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
-    t.string "name", null: false
+    t.string "name"
```
kicferk1, line 42: Is this a change that's real?

kicferk1: Maybe drop, create and re-migrate your db?

app/models/school.rb
kicferk1: Also, I know you are gonna love it, buut...
Do we want a test for `with_pupil_premium_uplift`?

ethanmills: No.

ethanmills: 😛

app/services/school_data_importer.rb
kicferk1: I think it's fine - especially since a test for it already lives in services directory.

app/models/school.rb
ethanmills: Done the first part, working on tests

kicferk1 (approved):
Just out of curiosity - is there a url, like with GIAS, from which we can download the CSV file?

kicferk1:
Oh, also, sorry about schema conflict, that's on me :P

ethanmills:
> Just out of curiosity - is there a url, like with GIAS, from which we can download the CSV file?

I got it from google drive, not sure where data got it from. I'll try to find out

amazimbe:
@ethanmills I can't review straightaway but if you're not in a hurry I'd be happy to take a look later today.

app/services/pupil_premium_importer.rb
```diff
@@ -0,0 +1,34 @@
...
+  def run
+    CSV.foreach(data_file, headers: true, encoding: "ISO-8859-1:UTF-8") do |row|
+      update_school_premium(row)
+    end
+  end
+
+private
+
+  def data_file
+    @source_file || __dir__ + "/../../data/pupil_premium.csv"
```
amazimbe, line 21: This will probably fail on native windows. Try and use something like this Rails.root.join('tmp', 'blah', 'caching-dev.txt' ) https://apidock.com/rails/Rails/root/class

app/services/pupil_premium_importer.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require "csv"
+
+class PupilPremiumImporter
+  def initialize(logger, start_year = Time.zone.now.year, source_file = "")
```
amazimbe, line 6: Better to have source_file default to nil rather than "". Was there a good reason for picking ""

app/services/school_data_importer.rb
```diff
@@ -4,6 +4,10 @@
...
 
 class SchoolDataImporter
+  def initialize(logger)
+    @logger = logger
```
amazimbe, line 8: It could be just personal preference, but I find it neater and a little easier on the eyes to add attr_reader :logger and then just logger.info rather than @logger.info

db/migrate/20210202134141_create_pupil_premium_eligibilities.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreatePupilPremiumEligibilities < ActiveRecord::Migration[6.1]
+  def change
+    create_table :pupil_premium_eligibilities do |t|
+      t.references :school, null: false, foreign_key: true, type: :uuid
+      t.column :start_year, :integer, limit: 2, null: false
```
amazimbe, line 7: t.column? Never come across that before.

db/schema.rb
```diff
@@ -213,6 +213,16 @@
     t.index ["lead_provider_id"], name: "index_provider_relationships_on_lead_provider_id"
   end
 
+  create_table "pupil_premium_eligibilities", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
amazimbe, line 216: I think the table should just have been called PupilPremium. The eligibility part seems to be more about the operations we perform on the data rather than the data itself.

db/migrate/20210202134141_create_pupil_premium_eligibilities.rb
kicferk1: Found some reference to it [here](https://api.rubyonrails.org/classes/ActiveRecord/Migration.html).

I agree t.integer is better.

kicferk1: Oh, also, we might have used `t.column` a lot in earlier migrations, oops.

db/schema.rb
ethanmills: The table contains %ages of pupils _eligible_ for pupil premium, hence the name. Happy to change it, I'm really not enjoying typing it out 😛

app/services/pupil_premium_importer.rb
ethanmills: None at all, will change

ethanmills:
[Sprint 4: Generate fixture from Pupil Premium data](https://trello.com/c/EJKi12x9/128-sprint-4-generate-fixture-from-pupil-premium-data)

app/models/pupil_premium.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  belongs_to :school
+
+  def uplift?
+    percentage_eligible >= THRESHOLD_PERCENTAGE
+  end
+
+  scope :only_with_uplift, lambda { |start_year|
+    where(start_year: start_year)
+      .where("(CAST(eligible_pupils AS float) / CAST(total_pupils AS float)) * 100 >= ?", THRESHOLD_PERCENTAGE)
```
amazimbe, line 14: Just curious if we need to cast both eligible_pupils and total_pupils.

app/services/pupil_premium_importer.rb
```diff
@@ -0,0 +1,41 @@
...
+  def run
+    CSV.foreach(data_file, headers: true, encoding: "ISO-8859-1:UTF-8") do |row|
+      update_school_premium(row)
+    end
+  end
+
+private
+
+  def data_file
+    source_file || Rails.root.join("data/pupil_premium.csv")
```
amazimbe, line 25: Rails.root.join("data", "pupil_premium.csv") would be platform independent. I also wonder if we could just do this in the initialiser i.e. @source_file = source_file || Rails.root.join("data", "pupil_premium.csv")

amazimbe (approved):
A couple of minor comments but otherwise LGTM!

app/services/pupil_premium_importer.rb
kicferk1: I think that will lead to 
```
RuboCop: Please use `Rails.root.join('path/to')` instead. [Rails/FilePath]
```

ethanmills: Yep, exactly

app/models/pupil_premium.rb
```diff
@@ -11,7 +11,7 @@ def uplift?
 
   scope :only_with_uplift, lambda { |start_year|
     where(start_year: start_year)
-      .where("(CAST(eligible_pupils AS float) / CAST(total_pupils AS float)) * 100 >= ?", THRESHOLD_PERCENTAGE)
+      .where("total_pupils > 0 AND (CAST(eligible_pupils AS float) / total_pupils) * 100 >= ?", THRESHOLD_PERCENTAGE)
```
kicferk1, line 14: Good old division by 0, I love it

app/services/pupil_premium_importer.rb
```diff
@@ -30,8 +30,8 @@ def update_school_premium(row)
     school = School.find_by(urn: urn)
     logger.info "Could not find school with URN #{urn}" and return unless school
 
-    total_pupils = row.fetch("Number of pupils on roll (7)")
-    eligible_pupils = row.fetch("Total number of pupils eligible for the Deprivation Pupil Premium")
+    total_pupils = row.fetch("Number of pupils on roll (7)").delete(",").to_i
```
kicferk1, line 33: Do we need to worry about spaces too?

rupollbabu (approved):
Covers the DB Operations.

app/services/pupil_premium_importer.rb
ethanmills: I don't think so... This is excel's fault, and while it puts commas in numbers, I don't think it's quite pathological enough to put spaces in. On the other hand... 🤷

kicferk1 (approved):

PR #87, opened 2021-02-05
Ect dashboard

kicferk1:
### Context
We want dashboard to show some CIP info to a ECT.

1. Completely unrelated: Axe the `encrypted_password` from the database. That ended up requiring us to axe a few more things, but the fairly exhaustive tests still pass, and I checked I can log in. I also read through more devise code than I wanted to.
2. A little related: Link course years to CIPs - I thought I did it already, but clearly not. That's the difference in seed files.
3. Show some CIP related contend on dashboard - link to course year seems like enough for time being.
4. Unrelated: Add `Testing` to our PR template - for some reason it didn't show automatically when I made this PR, boo hoo.

### Guidance to review
Commit by commit probably makes sense.
I ended up axing devise validation and going with our own - https://github.com/heartcombo/devise/issues/4913 is what inspired me to do that. Good thing we unit tested email validations already.

### Testing
1. That's pretty thoroughly tested throughout the app - loads of login in request tests, plus plenty of specific unit tests for login.
2. I don't think we are currently testing the content of the seed files. Not sure if we want to.
3. We don't really have front end tests. I guess I could add a request test and check contents of response, looking for some text, but is that how we want to test it?
4. Hopefully next PR it will just be added automatically.


app/models/user.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
-  devise :database_authenticatable, :registerable, :trackable,
-         :rememberable, :validatable, :confirmable, :passwordless_authenticatable
+  devise :registerable, :trackable, :rememberable, :confirmable, :passwordless_authenticatable
```
amazimbe, line 4: Validatable validates the format of the email so we probably need it. And it also validates password and I'm not sure it will play ball to have one and not the other.

Based on the feedback from the designer, we need to remove rememberable as well. I have altredy removed it from the UI.

app/models/user.rb
```diff
@@ -26,12 +26,12 @@ def induction_coordinator?
     induction_coordinator_profile.present?
   end
 
-  def early_career_teacher?
-    early_career_teacher_profile.present?
+  def lead_provider_admin?
+    lead_provider_profile.present?
```
amazimbe, line 30: I don't quite get this: so every lead provider is a lead_provider_admin? Why would we have both if that is the case?

kicferk1: Every lead provider user is a lead provider admin. For time being anyway - we can change what this function does when we have more than one type of lead provider user.

app/views/dashboard/show.html.erb
```diff
@@ -2,8 +2,19 @@
...
     <p class="govuk-body"> You are logged in</p>
-    <% if @current_user.lead_provider_profile %>
+
+    <% if @current_user.lead_provider_admin? %>
       <%= govuk_link_to "Go to school search", school_search_path %>
     <% end %>
+
+    <% if @current_user.early_career_teacher? %>
+      <h2 class="govuk-heading-l">Your Core Induction Programme</h2>
+      <% if @current_user.early_career_teacher_profile&.core_induction_programme&.course_years&.each do |course_year| %>
```
amazimbe, line 12: Can we not have a helper / method for this @current_user.early_career_teacher_profile&.core_induction_programme&.course_years&.each . https://en.wikipedia.org/wiki/Law_of_Demeter

app/models/user.rb
amazimbe: I suppose my question was why we didn't just call this lead_provider? especially when lead_provider and lead_provider_admin? are the same. lead_provider_admin? introduces the concept of an admin that we don't have nor seem to need right now.

kicferk1: Cause `lead_provider?` sounds weird in my head as a method describing a type of the user. And it is very similar to an existing method, `lead_provider` (only `?` of difference) - meaning it's easier to make a mistake between the two.

I think the only user idea that we have around lead providers is the person looking for schools, and eventually filling in a bunch of forms around partnerships - seems admin enough to me. Happy to change admin to something more descriptive, just don't know what.

ethanmills: I'm with Alfred on this one

amazimbe: Both points are valid and you don't need to change it if that is the case. I just needed the explanation / clarification.

kicferk1: Fine, but I reserve the right to "I told you so" when someone forgets the question-mark 😆

ethanmills: How un rails-y would `is_lead_provider?` be?

amazimbe: As un-railsy as having a rubocop against it but most rails codebases tend to have some of these.

amazimbe (approved):

rupollbabu (commented):
Requested @kicferk1 add early_career_teacher profile.

app/views/dashboard/show.html.erb
```diff
@@ -2,8 +2,19 @@
   <div class="govuk-grid-column-full">
```
ethanmills, line 2: At some point we need to decide on a unified approach to this. For the admin, we're sending them to a totally different page. Maybe partials for each user type that drop in?

db/migrate/20210205133627_remove_encrypted_password_from_user.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class RemoveEncryptedPasswordFromUser < ActiveRecord::Migration[6.1]
+  def change
+    remove_column :users, :encrypted_password, type: :string
```
ethanmills, line 5: `add_column :users, :cleartext_password, type: :string`

kicferk1: What?

kicferk1: Surely you mean `plaintext` 😂

kicferk1:
> Requested @kicferk1 add early_career_teacher profile.

Added :)

db/migrate/20210205133627_remove_encrypted_password_from_user.rb
seb-siem-softwire: What do we replace encrypted password with?

ethanmills: Sorry, this was a joke, didn't mean to confuse things

seb-siem-softwire: Cool cool

rupollbabu (commented):
Requested to add Test scenario ,When user sign in then Assert the URL/dashboard/Any unique value at least one test scenario for each role)

PR #88, opened 2021-02-08
update ambition years page in cip

xswoodsx:
### Context
CIP seed content needed updating so that the Ambition years page used the same page taken from the NERO site as other providers
Updating CIP seed and seed dump content

### Guidance to review
I moved the Ambition content back to the 0 index in CourseYear import content as editing it locally moved them down to the bottom of the list. Ambition should still be first listed when seeding.


db/seeds/cip_seed.rb
```diff
@@ -94,6 +94,8 @@
...
   ["4b0c92ea-f9ea-4794-9ebf-e651ebea4719", "2.6: Reflecting on learning", "## Time allocation\n\n10 minutes\n\n## Instructions\n\n- Use your notepad to reflect on your learning in this Block and further professional development.\n- Write just a couple of sentences for each question.\n\n$CTA\n\n<h2>In your notepad</h2>\n<ol>\n  <li>What have you learned in this Block?</li>\n  <li>What do you need to learn more about?</li>\n</ol>\n$CTA\n\n", "701a5401-d77b-48af-8c4d-3bee197896f0", "48c66862-8a65-4908-8fdc-eff63d632940", 4],
```
kicferk1, line 99: I wonder why they moved. If you move them back where they were, re-seed and re-export, do they stay in their old place, or do they move back here?

kicferk1 (commented):
Just a question about the rogue lessons moving around in our seed file.

db/seeds/cip_seed.rb
xswoodsx: I'm not sure, i put it back to the original place in the seed, re-seeded and the seed dump moved it back to lines 97 and 98.

kicferk1: Weird.

I got the old files, I updated ambition year content in seed file, I dropped my db, recreated it, seeded it, and then nothing else moved.

I _think_ it's to do with the last time an object was updated. I think `CoreInductionProgrammeExporter` orders objects by default Rails order, which might be when they were last updated. That's annoying. But since we control the queries we might be able to specify that we want the order the elements were created by to be the deciding factor. Wanna try it? 

In `CoreInductionProgrammeExporter` we would add
```
.order('created_at ASC')
```
to each of the `.all` queries. And see if it is any better :)

amazimbe (approved):
Difficult to review 😁 but I trust you've tested it.

kicferk1:
> Difficult to review 😁 but I trust you've tested it.

I recommend exploiting the PR app.
dev: https://ecf-dev.london.cloudapps.digital/core-induction-programme/years/a4dc302c-ab71-4d7b-a10a-3116a778e8d5
review app: https://ecf-review-pr-88.london.cloudapps.digital/core-induction-programme/years/a4dc302c-ab71-4d7b-a10a-3116a778e8d5

Why yes, the id is the same, thanks for asking. That's because we export and import object IDs as well (to make foreign keys work).

kicferk1:
Had a chat with Babu earlier today, I believe we are not gonna test the content of cip pages, at least not yet, so this should be good to go. The investigation into changing order can happen separately :)

PR #90, opened 2021-02-08
Add view_component and create tabs component

callumacrae:
### Context

I've moved the existing tabs HTML into a view_component component - mostly to test view_component and give an example of how it works.

There is a slight behaviour change - instead of the supplier and supplier user page passing a local variable in, it's using `current_page?` to check the current path against the specified one. This will break if child pages use the same layout, but nothing does that at the moment and would require other changes if desired in the future.

### Guidance to review

To test pagination without manually adding hundreds of providers, set `.per(20)` to `.per(2)` or something like that in the suppliers controller (or revert 35a22e9 where I left it in by mistake).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -451,6 +451,7 @@ DEPENDENCIES
   spring
   spring-watcher-listen (~> 2.0.0)
   tzinfo-data
+  view_component
   wdm (>= 0.1.0)
   web-console (>= 4.1.0)
   webdrivers (~> 4.4, >= 4.4.1)
@@ -461,4 +462,4 @@ RUBY VERSION
    ruby 2.7.1p83
 
 BUNDLED WITH
-   2.1.4
+   2.2.8
--- a/app/components/tabs_component/tab_component.html.erb
+++ b/app/components/tabs_component/tab_component.html.erb
@@ -1,0 +1,5 @@
+<div class="page-tab govuk-!-margin-right-8 <%= "page-tab__selected" if current_page?(path) %>">
+  <%= govuk_link_to path do %>
+    <%= content %> <%= "<span class='govuk-visually-hidden'>Current page</span>".html_safe if current_page?(path) %>
+  <% end %>
+</div>
--- a/app/views/admin/suppliers/_layout.html.erb
+++ b/app/views/admin/suppliers/_layout.html.erb
@@ -1,14 +1,10 @@
 <h1 class="govuk-heading-l">Suppliers</h1>
-<div class="page-tab-container">
-  <div class="page-tab govuk-!-margin-right-8 <%= "page-tab__selected" if page == "suppliers" %>">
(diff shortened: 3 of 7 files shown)


Gemfile.lock
```diff
@@ -461,4 +462,4 @@ RUBY VERSION
    ruby 2.7.1p83
 
 BUNDLED WITH
-   2.1.4
+   2.2.8
```
callumacrae, line 465: Does this mean I have an incorrect version of something installed that I should change?

ethanmills (approved):
This is awesome! Much nicer than my mess..

Gemfile.lock
amazimbe: It looks like you have the latest version of bundle on your system and we were all on 2.1.4

amazimbe: We can all upgrade to 2.2.8 I suppose

app/views/admin/suppliers/_layout.html.erb
```diff
@@ -1,14 +1,10 @@
...
-  <div class="page-tab govuk-!-margin-right-8 <%= "page-tab__selected" if page == "users" %>">
-    <%= govuk_link_to admin_supplier_users_path do %>
-      All users <%= "<span class='govuk-visually-hidden'>Current page</span>".html_safe if page == "users" %>
-    <% end %>
-  </div>
-</div>
-<hr class="govuk-!-margin-top-0"/>
+
+<%= render TabsComponent.new do |c| %>
+  <%= c.tab(path: admin_suppliers_path) do %>
```
amazimbe, line 4: Yeh, that's a lot cleaner.

app/components/tabs_component/tab_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="page-tab govuk-!-margin-right-8 <%= "page-tab__selected" if current_page?(path) %>">
```
kicferk1, line 1: Clever.

I imagine if we had more conditional logic I would expect two components - `active_tab_component` and `tab_component` - and the check to be done in tabs. But two if statements are just about less complex than that :)

kicferk1 (approved):

amazimbe (approved):

app/components/tabs_component/tab_component.html.erb
callumacrae: The logic for the check would ideally be done in the controller, I guess - but when I tried that I got an error, and as I said, I'm trying to avoid changing too many things in this PR.

kicferk1: Yeah, something for later, when we deem we need it :)

seb-siem-softwire (approved):

rupollbabu (approved):

rupollbabu (approved):

PR #92, opened 2021-02-11
ECF-RP-129: Handle schools changing local authorities and lads

ethanmills:
### Context
This is work in preparation for adding sparsity data. Splitting this ticket up into several PRs, lots to change

Add new models to link schools to local authorities and local authority districts
Update the school data import to match 

### Testing
I've updated the unit tests, but there's nothing much new here. Just moving the same data to new places

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -5,9 +5,11 @@ class School < ApplicationRecord
 
   belongs_to :network, optional: true
 
-  # TODO: Register and Partner 150: Handle schools changing LAs and LADs
-  belongs_to :local_authority_district, optional: true
-  belongs_to :local_authority, optional: true
+  has_many :school_local_authorities
+  has_many :local_authorities, through: :school_local_authorities
+  has_many :school_local_authority_districts
+  has_many :local_authority_districts, through: :school_local_authority_districts
+
   has_one :partnership
   has_one :lead_provider, through: :partnership
   has_many :pupil_premiums
@@ -41,6 +43,14 @@ class School < ApplicationRecord
     unconfirmed_induction_coordinators
       &.where("users.confirmation_sent_at > ?", CONFIRMATION_WINDOW.hours.ago)
       &.any?
+  end
+
+  def local_authority
+    school_local_authorities.find_by(end_year: nil)&.local_authority
+  end
+
+  def local_authority_district
+    school_local_authority_districts.find_by(end_year: nil)&.local_authority_district
   end
 
   def pupil_premium_uplift?(start_year)
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -46,11 +46,11 @@ class School < ApplicationRecord
   end
 
   def local_authority
-    school_local_authorities.find_by(end_year: nil)&.local_authority
+    school_local_authorities.latest.first&.local_authority
   end
 
   def local_authority_district
-    school_local_authority_districts.find_by(end_year: nil)&.local_authority_district
+    school_local_authority_districts.latest.first&.local_authority_district
   end
 
   def pupil_premium_uplift?(start_year)
--- a/app/services/school_data_importer.rb
+++ b/app/services/school_data_importer.rb
@@ -95,7 +95,7 @@ private
 
     if school_local_authority.new_record?
       school_local_authority.start_year = start_year
-      SchoolLocalAuthority.find_by(school: school, end_year: nil)&.update!(end_year: start_year)
+      SchoolLocalAuthority.latest.find_by(school: school)&.update!(end_year: start_year)
     end
 
     school_local_authority.save!
@@ -109,7 +109,7 @@ private
 
     if school_lad.new_record?
       school_lad.start_year = start_year
-      SchoolLocalAuthorityDistrict.find_by(school: school, end_year: nil)&.update!(end_year: start_year)
+      SchoolLocalAuthorityDistrict.latest.find_by(school: school)&.update!(end_year: start_year)
     end
 
     school_lad.save!
--- a/app/services/school_data_importer.rb
+++ b/app/services/school_data_importer.rb
@@ -5,16 +5,22 @@ require "csv"
 
 class SchoolDataImporter
   attr_reader :logger
+  attr_reader :start_year
 
-  def initialize(logger)
+  def initialize(logger, start_year = Time.zone.now.year)
     @logger = logger
+    @start_year = start_year
   end
 
   # TODO: Register and Partner 147: Figure out how exactly we want to handle school updating
   def run
     CSV.foreach(schools_data_file.path, headers: true, encoding: "ISO-8859-1:UTF-8") do |row|
       school = row_to_school(row)
-      school.save!
+      local_authority = row_to_local_authority(row)
+      local_authority_district = row_to_lad(row)
+
+      link_school_to_local_authority(school, local_authority)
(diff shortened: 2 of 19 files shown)


app/services/school_data_importer.rb
```diff
@@ -87,4 +89,29 @@ def row_to_lad(row)
...
+      local_authority_district: lad,
+    )
+
+    if school_lad.new_record?
+      school_lad.start_year = start_year
+      SchoolLocalAuthorityDistrict.find_by(school: school, end_year: nil)&.update!(end_year: start_year)
+    end
+
+    school_lad.save!
+  end
```
ethanmills, line 116: I commonised this with some very rails-y metaprogramming, then hated it and deleted it immediately

```
def link_school_to_local_authority(school, local_authority)
  link_school_to_model(school, SchoolLocalAuthority, local_authority)
end

def link_school_to_lad(school, lad)
  link_school_to_model(school, SchoolLocalAuthorityDistrict, lad)
end

def link_school_to_model(school, model, instance)
  school_model = model.find_or_initialize_by(school: school, instance.class.to_s.underscore.to_sym => instance)

  if school_model.new_record?
    school_model.start_year = start_year
    model.find_by(school: school, end_year: nil)&.update!(end_year: start_year)
  end

  school_model.save!
end
```

kicferk1: I think I hate this commonisation too.

app/services/school_data_importer.rb
```diff
@@ -5,16 +5,22 @@
...
   def run
     CSV.foreach(schools_data_file.path, headers: true, encoding: "ISO-8859-1:UTF-8") do |row|
       school = row_to_school(row)
-      school.save!
+      local_authority = row_to_local_authority(row)
+      local_authority_district = row_to_lad(row)
+
+      link_school_to_local_authority(school, local_authority)
+      link_school_to_lad(school, local_authority_district)
     end
```
kicferk1, line 24: You know our new favourite question? 
How are we gonna test that? :P 

I guess we could set up a school with a different LAD / LA than the one in our csv file, import, and check the behaviour is as expected?

db/seeds/dummy_structures.rb
```diff
@@ -32,35 +36,27 @@
 end
 
 if Rails.env.development? || Rails.env.deployed_development?
-  if AdminProfile.none?
```
ethanmills, line 35: @kicferk1 you happy with this?

kicferk1: Looks like it will work to me :)

app/models/school.rb
```diff
@@ -43,6 +45,14 @@ def partially_registered?
       &.any?
   end
 
+  def local_authority
+    school_local_authorities.find_by(end_year: nil)&.local_authority
```
amazimbe, line 49: I don't quite get why we need end_year to be nil.

kicferk1: Cause that's how we find the latest one - without any end time (I think)

ethanmills: Because they have not left that local authority. End year is the year a school is no longer a member of that local authority

seb-siem-softwire: We could introduce a friendly scope so that it would be easier to read - scope :latest could do onschool_local_authority

app/models/school.rb
```diff
@@ -43,6 +45,14 @@ def partially_registered?
...
+  end
+
+  def local_authority_district
+    school_local_authority_districts.find_by(end_year: nil)&.local_authority_district
```
seb-siem-softwire, line 53: Same as above so that it would not be confusing when reading it 6 months after

app/services/school_data_importer.rb
seb-siem-softwire: 👍

app/services/school_data_importer.rb
```diff
@@ -87,4 +89,29 @@ def row_to_lad(row)
     local_authority_district.save!
     local_authority_district
   end
+
+  def link_school_to_local_authority(school, local_authority)
+    school_local_authority = SchoolLocalAuthority.find_or_initialize_by(school: school, local_authority: local_authority)
+
+    if school_local_authority.new_record?
+      school_local_authority.start_year = start_year
+      SchoolLocalAuthority.find_by(school: school, end_year: nil)&.update!(end_year: start_year)
```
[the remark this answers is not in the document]
ethanmills: If we start with 
`LA1, start_year: 2019, end_year: nil`
and want to change to LA2, we need to end up with
`LA2, start_year: 2021, end_year: nil; LA1, start_year: 2019, end_year: 2021`

db/migrate/20210208173154_create_school_local_authorities.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateSchoolLocalAuthorities < ActiveRecord::Migration[6.1]
+  def change
+    create_table :school_local_authorities do |t|
+      t.references :school, null: false, foreign_key: true, type: :uuid
```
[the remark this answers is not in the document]
ethanmills: Which entities?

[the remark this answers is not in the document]
ethanmills: This is just an association between a school and a local authority for a given period of time. The names live on the school and the local authority

app/services/school_data_importer.rb
```diff
@@ -5,16 +5,22 @@
 
 class SchoolDataImporter
   attr_reader :logger
+  attr_reader :start_year
 
-  def initialize(logger)
+  def initialize(logger, start_year = Time.zone.now.year)
     @logger = logger
+    @start_year = start_year
```
kicferk1, line 12: Do we need both attr reader and a class variable for `start_year`?

app/models/school.rb
ethanmills: For a scope it would have to be `where(end_year: nil)`, which we would need to call `.first` on. Is it worth it?

app/services/school_data_importer.rb
ethanmills: Pretty sure that's how attr_reader works, right? attr_reader, not attr_accessor

kicferk1: Ehh, so let me double check. Do we set a class variable, and then never use it directly again? And instead of using it directly(in a few methods), we access it by using attr_reader? 

Oh, is this how getter and setters work in Rails? Uhh. Fine.

ethanmills: You can use an attr_accessor, which creates a getter and setter. But we only want a getter, so I used attr_reader

app/services/school_data_importer.rb
```diff
@@ -87,4 +89,29 @@ def row_to_lad(row)
...
+
+  def link_school_to_lad(school, lad)
+    school_lad = SchoolLocalAuthorityDistrict.find_or_initialize_by(
+      school: school,
+      local_authority_district: lad,
+    )
+
+    if school_lad.new_record?
+      school_lad.start_year = start_year
+      SchoolLocalAuthorityDistrict.find_by(school: school, end_year: nil)&.update!(end_year: start_year)
```
kicferk1, line 112: Will something like `SchoolLocalAuthorityDistrict.latest.find_by(school:school)` find the right district?

rupollbabu (approved):

kicferk1 (approved):

PR #93, opened 2021-02-11
Add demo cypress setup + tests

callumacrae:
This PR is adding Cypress to power end to end tests - at the moment, there aren't really any tests except from a couple demo tests (which will be removed before merging) and some meta tests to test the tests.

Issues:

- ~Cypress on rails is having issues with symbol name of factory trait. Have worked around it but obviously needs a proper fix~
- ~Sometimes the user created by factorybot isn't an admin causing a test to fail~
- ~database_cleaner probably isn't required - need to figure out how the existing tests are doing it~
- example tests need to be removed before merging


package.json
```diff
@@ -13,14 +13,17 @@
...
     "eslint-config-airbnb-base": "^14.2.1",
     "eslint-plugin-import": "^2.22.1",
     "jest": "^26.6.3",
     "webpack-dev-server": "^3.11.0"
   },
   "scripts": {
-    "test": "jest --passWithNoTests"
+    "test": "jest --passWithNoTests",
+    "cypress:open": "cypress open --project ./spec",
+    "cypress:run": "cypress run --project ./spec"
```
kicferk1, line 26: What's the difference between the two?

spec/cypress/app_commands/activerecord_fixtures.rb
```diff
@@ -0,0 +1,22 @@
+# you can delete this file if you don't use Rails Test Fixtures
+
+fixtures_dir = command_options.try(:[], 'fixtures_dir')
+fixture_files = command_options.try(:[], 'fixtures')
+
+if defined?(ActiveRecord)
+  require "active_record/fixtures"
+
+  fixtures_dir ||= ActiveRecord::Tasks::DatabaseTasks.fixtures_path
+  fixture_files ||= Dir["#{fixtures_dir}/**/*.yml"].map { |f| f[(fixtures_dir.size + 1)..-5] }
```
kicferk1, line 10: Uhh, what does this do?

In particular, 
```
.map { |f| f[(fixtures_dir.size + 1)..-5] }
```
looks... not very readable.

spec/cypress/cypress_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+begin
+  require 'factory_bot_rails'
+rescue LoadError => e
+  puts e.message
+  begin
+    require 'factory_girl_rails'
+  rescue LoadError => e
+    puts e.message
+  end
+end
```
kicferk1, line 18: I think we shouldn't check for `factory_girl_rails`. 

Also, I would just require without begin-rescue blocks, unless we expect this file to run in production environment.

spec/cypress/cypress_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+
+require 'cypress_on_rails/smart_factory_wrapper'
+
+factory = CypressOnRails::SimpleRailsFactory
+factory = FactoryBot if defined?(FactoryBot)
+factory = FactoryGirl if defined?(FactoryGirl)
```
kicferk1, line 24: Again, let's just use the one we want.

spec/cypress/app_commands/activerecord_fixtures.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  logger.debug "loading fixtures: { dir: #{fixtures_dir}, files: #{fixture_files} }"
+  ActiveRecord::FixtureSet.reset_cache
+  ActiveRecord::FixtureSet.create_fixtures(fixtures_dir, fixture_files)
+  "Fixtures Done" # this gets returned
+else # this else part can be removed
```
kicferk1, line 16: Let's remove it then :)

spec/cypress/cypress_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+
+CypressOnRails::SmartFactoryWrapper.configure(
+    always_reload: !Rails.configuration.cache_classes,
+    factory: factory,
+    files: [
+      Rails.root.join('spec', 'factories.rb'),
+      Rails.root.join('spec', 'factories', '**', '*.rb')
+    ]
+)
```
kicferk1, line 33: Actually, I have a feeling we have another file like this one. 

```
spec/cypress/app_commands/factory_bot.rb
```

Do we need both?

ethanmills:
@kicferk1 I think most of this is generated code as-is

ethanmills:
I know why your admin test is failing. We had factory bot inadvertently creating two users, the second of which was an admin. It always returns the admin to ruby code, but because you were taking `User.last` not `User.order(:created_at).last` and we're using UUID primary keys, it was random chance as to whether your user was admin or not. Will fix both 🙂

spec/cypress/integration/example-tests.spec.js
```diff
@@ -0,0 +1,125 @@
+describe("Example tests", () => {
+  beforeEach(() => {
+    cy.app("clean");
+  });
+
+  after(() => {
+    cy.app("clean");
+  });
```
ethanmills, line 8: Added this to leave the db clean

spec/cypress/app_commands/activerecord_fixtures.rb
callumacrae: No idea - this was added by cypress-on-rails

spec/cypress/integration/example-tests.spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+  it("should not be editable when logged in as non-admin user", () => {
+    cy.appFactories([
+      ["create", "user"],
+      ["create", "course_lesson"],
+    ]);
+
+    // @todo appFactories returns this
+    cy.appEval("User.last").then((res) => {
```
ethanmills, line 43: When fetching something we just created, we should include an `.order(:created_at)` before the last

package.json
callumacrae: open opens the interface for running tests during development like I showed yesterday, run runs all the tests in the command line (and is basically for ci)

spec/cypress/cypress_helper.rb
callumacrae: Happy to clean this up, but again, this is factory-on-rails added code

package.json
ethanmills: run is headless, open is interactive

spec/cypress/cypress_helper.rb
callumacrae: Good point, looks like the cypress-on-rails docs were out of date

.idea/codeStyles/Project.xml
```diff
@@ -11,5 +11,8 @@
       <option name="HTML_DO_NOT_INDENT_CHILDREN_OF" value="html,body" />
       <option name="HTML_ENFORCE_QUOTES" value="true" />
     </HTMLCodeStyleSettings>
+    <JSCodeStyleSettings version="0">
+      <option name="SPACES_WITHIN_OBJECT_LITERAL_BRACES" value="true" />
+    </JSCodeStyleSettings>
```
seb-siem-softwire, line 16: Hmmm - I thought we were suppose to exclude .idea files?

package.json
seb-siem-softwire: Seems like resolved

spec/cypress/app_commands/scenarios/basic.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# You can setup your Rails state here
+# MyModel.create name: 'something'
+Post.create!(title: "I am a Postman")
```
seb-siem-softwire, line 5: Is this needed?

spec/cypress/integration/example-tests.spec.js
```diff
@@ -0,0 +1,119 @@
...
+    cy.get("#navigation").contains("Core Induction Programme").click();
+
+    cy.get("h1").should("contain", "What is the early career framework?");
+
+    cy.contains("Test Course year").click();
+
+    cy.get("h1").should("contain", "Test Course year");
+    cy.get(".govuk-govspeak").should("contain", "No content");
+
+    cy.get("a").contains("Test Course module").click();
```
seb-siem-softwire, line 25: This is great as a proof of concept as long as we are mindful that this pattern will potentially create a testing architecture nightmare if not trimmed/refactored continuously - One of the more elegant test patterns is page object pattern where you abstract the things that you see on a page in different classes and as such finding what a specific page has would continuously improve.

spec/cypress/support/on-rails.js
```diff
@@ -0,0 +1,54 @@
...
+});
+
+Cypress.Commands.add('appFactories', function (options) {
+  return cy.app('factory_bot', options)
+});
+
+Cypress.Commands.add('appFixtures', function (options) {
+  cy.app('activerecord_fixtures', options)
+});
+// CypressOnRails: end
```
seb-siem-softwire, line 36: Is this needed given that it is commented out

spec/cypress/support/on-rails.js
```diff
@@ -0,0 +1,54 @@
...
+
+// The next is optional
+// beforeEach(() => {
+//  cy.app('clean') // have a look at cypress/app_commands/clean.rb
+// });
```
seb-siem-softwire, line 41: Cleaning the state before every test - is a good idea

.idea/codeStyles/Project.xml
ethanmills: Why?

kicferk1: Cause it is in gitignore 😂 

Just not quite right ;) we should make it so all files in idea are ignored, not just the first level of them.

kicferk1: On a more serious note, out choice of linting rules lives in non-tool specific tools, like rubocop.

ethanmills: Yes, but why is commiting this a bad thing? Eslint should complain about this regardless, but it makes sense for me for all devs using RubyMine to have it just work

kicferk1: My guess is that it's because they tend to be a messy, unreadable xml gabage. And, if one person changes their IDE settings, then everyone on the project has to too, or else! 

And those settings include a lot of cosmetic things, or preferences rather than hard rules. Do you like your IDE in light mode? No? Though shit, I just committed an IDE change to make it dark 😂

ethanmills: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

kicferk1: Also (perish the thought) what if two people use different IDEs that both use `.idea` directory to store settings and they are not fully compatible with each other? ooof, I don't even want to think about it :P

spec/cypress/app_commands/scenarios/basic.rb
callumacrae: no, but I wanted to keep it so that people know scenarios exist instead of removing the directory completely - it can be deleted when someone adds an actual scenario

spec/cypress/integration/example-tests.spec.js
callumacrae: yep, this is all demo code

spec/cypress/support/on-rails.js
callumacrae: 🤷‍♂️ 

I left it because it makes it clear that it's generated code. It can be removed though

spec/cypress/support/on-rails.js
callumacrae: I'd be wary of doing this as it seems a bit magic, and it's not inconceivable that we'd have tests where we _wouldn't_ want to clean the state as the state isn't used—e.g., testing logged out pages, or testing pages on production.

However, I left the commented out code so that we could have this discussion :)

spec/cypress/cypress_helper.rb
seb-siem-softwire: ;) Aaa thoughtworks FactoryGirl - a great example of removing sexism from the code - quite an interesting case study

.idea/codeStyles/Project.xml
callumacrae: can i add my vim config to the repo?

spec/cypress/integration/aaa-meta.js
```diff
@@ -0,0 +1,70 @@
+/**
+ * !!!!!
+ *
+ * This file name starts with aaa so that it runs before all the other tests
+ * which rely on the functionality being tested in this file.
+ *
+ * Please don't rename it!
```
kicferk1, line 7: Gotta love this xD

.idea/codeStyles/Project.xml
kicferk1: I don't think it needs to be done in this PR, but we should probably add a ticket for 'sorting out idea files'. Maybe after talking it through on a dev catchup.

spec/cypress/support/on-rails.js
kicferk1: I'd be for cleaning before every test. 

If we want to test that something is displayed after the action, I think we should explicitly set it up in that test - maybe commonising it when needed.

kicferk1 (commented):
Looks pretty good, I am on the side of cleaning before each test, so

**I will never approve until it happens** - or until more people prefer otherwise, hehe.

spec/cypress/support/on-rails.js
callumacrae: Have changed it to clean before every test!

kicferk1 (approved):

ethanmills (approved):
Happy to go with this and change anything as it comes up

PR #95, opened 2021-02-15
Adds api for returning list of schools

seb-siem-softwire:
### Context

This is to bring autocomplete functionality to school search

Using accessible autocomplete we can now search character by character.

This pull request does not implement any authentication protocol for our API endpoint.

For serialization I used blueprinter 

### Guidance to review

Dont burn me too much 

### Testing

I used unit tests for testing serialization and model

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,6 +44,9 @@ gem "mail-notify", ">= 1.0.3"
 # do not rely on host's timezone data, which can be inconsistent
 gem "tzinfo-data"
 
+# serialization gem that offers more features than active model serializer
+gem "blueprinter"
+
 gem "govuk-components", ">= 1.0.2"
 gem "govuk_design_system_formbuilder", "~> 2.1", ">= 2.1.5"
 gem "httpclient"
--- a/app/controllers/api/school_search_controller.rb
+++ b/app/controllers/api/school_search_controller.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class Api::SchoolSearchController < Api::ApiController
+  def index
+    if params[:search_key]
+      schools = School.search_by_name_or_urn(params[:search_key]).limit(10)
+      decorated_schools = schools.map { |school| ::Decorators::SchoolDecorator.new(school) }
+      render json: SchoolSerializer.render(decorated_schools)
+    else
+      render json: []
+    end
+  end
+end
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -16,6 +16,10 @@ class School < ApplicationRecord
   has_many :early_career_teacher_profiles
   has_many :early_career_teachers, through: :early_career_teacher_profiles, source: :user
 
+  scope :search_by_name_or_urn, lambda { |search_key|
+    where("urn ILIKE ?", "%#{search_key}%").or(School.where("name ILIKE ?", "%#{search_key}%"))
+  }
+
   def full_address
     address = <<~ADDRESS
       #{address_line1}
--- a/app/serializers/school_serializer.rb
+++ b/app/serializers/school_serializer.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class SchoolSerializer < Blueprinter::Base
+  identifier :id
+  fields :full_address_formatted, :name
+end
--- a/app/views/school_search/show.html.erb
+++ b/app/views/school_search/show.html.erb
@@ -14,10 +14,12 @@
         <p>
           <%= govuk_link_to "Clear filters", "?" %>
         </p>
-        <%= f.govuk_text_field(
-              :school_name,
-              value: @school_search_form.school_name,
-              label: {text: "School name or group", class: "govuk-label govuk-label--s"}) %>
+
+        <div class="govuk-form-group">
+          <label for="school-search-form-school-name-field" class="govuk-label govuk-label--s school-search-label">School name or group</label>
+          <div id="school-search-form-autocomplete" class="autocomplete-wrapper"></div>
+        </div>
+
         <%= f.govuk_text_field(
               :location,
               label: {text: "Location", class: "govuk-label govuk-label--s"},
--- a/app/webpacker/packs/school_search.js
+++ b/app/webpacker/packs/school_search.js
@@ -1,5 +1,5 @@
 import accessibleAutocomplete from "accessible-autocomplete";
-import fetch from "node-fetch";
+import "whatwg-fetch";
 
 const element = document.querySelector("#school-search-form-autocomplete");
 const id = "school-search-form-input";
--- a/app/webpacker/packs/school_search.js
+++ b/app/webpacker/packs/school_search.js
@@ -1,5 +1,4 @@
 import accessibleAutocomplete from "accessible-autocomplete";
-import "whatwg-fetch";
 
 const element = document.querySelector("#school-search-form-autocomplete");
 const id = "school-search-form-input";
--- a/app/webpacker/packs/school_search.js
+++ b/app/webpacker/packs/school_search.js
@@ -1,0 +1,32 @@
+import accessibleAutocomplete from "accessible-autocomplete";
+import fetch from "node-fetch";
+
+const element = document.querySelector("#school-search-form-autocomplete");
+const id = "school-search-form-input";
+
+function inputValueTemplate(result) {
+  return result && result.name;
+}
+
+function suggestionTemplate(result) {
+  return result && `<strong>${result.name}</strong> (${result.full_address_formatted})`;
+}
+
+function suggest(query, populateResults) {
+  fetch(`/api/school_search?search_key=${query}`)
+    .then((response) => response.json())
+    .then((data) => populateResults(data));
+}
+
+accessibleAutocomplete({
+  element,
+  id,
+  source: suggest,
+  showNoOptionsFound: false,
+  minLength: 1,
+  name: "school_search_form[school_name]",
+  templates: {
+    inputValue: inputValueTemplate,
+    suggestion: suggestionTemplate,
+  },
+});
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "@rails/webpacker": "^5.2.1",
     "accessible-autocomplete": "^2.0.3",
     "govuk-frontend": "^3.10.2",
+    "node-fetch": "^2.6.1",
     "serialize-javascript": "^5.0.1",
     "set-value": "^3.0.2"
   },
--- a/package.json
+++ b/package.json
@@ -9,9 +9,9 @@
     "@rails/webpacker": "^5.2.1",
     "accessible-autocomplete": "^2.0.3",
     "govuk-frontend": "^3.10.2",
-    "node-fetch": "^2.6.1",
     "serialize-javascript": "^5.0.1",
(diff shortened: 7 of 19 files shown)


app/models/school.rb
```diff
@@ -27,6 +31,10 @@ def full_address
     address.squeeze("\n")
   end
 
+  def full_address_formatted
```
amazimbe, line 34: I'd recommend adding a presenter and moving this method as well as :full_address there. I prefer the logic in the models to be business like / related to the subject domain rather than just presentation.

Gemfile
```diff
@@ -44,6 +44,9 @@ gem "mail-notify", ">= 1.0.3"
 # do not rely on host's timezone data, which can be inconsistent
 gem "tzinfo-data"
 
+# serialization gem that offers more features than active model serializer
+gem "blueprinter"
```
amazimbe, line 48: 👍 I haven't used this one before. I'll check it out!

app/controllers/api/school_search_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Api::SchoolSearchController < Api::ApiController
+  def index
```
amazimbe, line 4: @DFE-Digital/ecf Does this new feature mean we can remove https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/controllers/school_search_controller.rb

app/models/school.rb
```diff
@@ -16,6 +16,10 @@ class School < ApplicationRecord
   has_many :early_career_teacher_profiles
   has_many :early_career_teachers, through: :early_career_teacher_profiles, source: :user
 
+  scope :search_by_name_or_urn, lambda { |search_key|
```
amazimbe, line 19: Is this specifically for searching by name and urn. I know for sure we'll have to search by other parameters like location etc. If we want it to be generic then the scope name probably needs generalising.

app/controllers/api/school_search_controller.rb
kicferk1: Almost certainly not. We will still need `school_search_controller` (or any other controller) to display the school search page. But we can probably prune away the search logic from there, with enough effort. I would do it in a separate PR though.

app/serializers/school_serializer.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class SchoolSerializer < Blueprinter::Base
```
amazimbe, line 3: It would be good to add unit tests for these serializers as well though this one doesn't have much logic yet.

app/models/school.rb
```diff
@@ -16,6 +16,10 @@ class School < ApplicationRecord
...
+    where("urn ILIKE ?", "%" + search_key + "%").or(School.where("name ILIKE ?", "%" + search_key + "%"))
```
kicferk1, line 20: String interpolation over `+`s please :)

app/webpacker/packs/school_search.js
```diff
@@ -0,0 +1,35 @@
...
+const element = document.querySelector("#school-search-form-autocomplete");
+const id = "school-search-form-input";
+
+function inputValueTemplate(result) {
+  return result && result.name;
+}
+
+function suggestionTemplate(result) {
+  return result && `<strong>${result.name}</strong> (${result.full_address_formatted})`;
+}
```
kicferk1, line 14: Ah, crap, we have to compute this in js because autocomplete constructs its own html using js? argh. Oh well.

app/models/school.rb
ethanmills: We only want name and URN in the accessible autocomplete component though, right?

kicferk1 (commented):
Looks pretty good.

app/views/school_search/show.html.erb
```diff
@@ -14,10 +14,12 @@
...
         </p>
-        <%= f.govuk_text_field(
-              :school_name,
-              value: @school_search_form.school_name,
-              label: {text: "School name or group", class: "govuk-label govuk-label--s"}) %>
+
+        <div class="govuk-form-group">
+          <label for="school-search-form-school-name-field" class="govuk-label govuk-label--s school-search-label">School name or group</label>
+          <div id="school-search-form-autocomplete" class="autocomplete-wrapper"></div>
+        </div>
```
ethanmills, line 21: Interesting! Do we definitely want this here? Has it been specified?

app/controllers/api/school_search_controller.rb
seb-siem-softwire: No - this is only for school search api access

seb-siem-softwire: It's namespaced so it doesn't conflict

app/models/school.rb
seb-siem-softwire: Do we have any presenter/decorator patterns?

app/models/school.rb
seb-siem-softwire: Yes - we only have requirement for URN and name

app/webpacker/packs/school_search.js
seb-siem-softwire: Yep - took some banging on the monitor and quite a few swear words to figure it right

app/views/school_search/show.html.erb
seb-siem-softwire: Good point - the ticket specified the proof of concept and that we might use it in other parts of the app - this was the most logical place as I could test whenever the params are being sent correctly and that it doesn't interfere with existing functionality

app/models/school.rb
seb-siem-softwire: Ahhh - yeah make sense - sowwwwyyyy

app/serializers/school_serializer.rb
seb-siem-softwire: I do test it indirectly spec/requests/api/school_search_spec.rb - so I could extend that to also have name attribute - or would you prefer pure serializer tests

seb-siem-softwire: @amazimbe

Gemfile
seb-siem-softwire: Originally I thought about active model serializers and then I remembered how hard sometimes was to see the output of the serialization - and then found this little gem - pretty straightforward and easy to use

app/models/school.rb
amazimbe: No yet, we can always add one though, draper, delegator

app/serializers/school_serializer.rb
amazimbe: Yes, I'd prefer pure serializer tests as well. Not so much for this one but to declare the intention that we test exhaustively 😉 . If we don't add them now probably the next person won't too.

seb-siem-softwire: Yeah - you do have a point there - plus since we are using a new serializer we need to know the patterns for adding more

package.json
```diff
@@ -8,6 +8,7 @@
   "dependencies": {
     "@rails/webpacker": "^5.2.1",
     "accessible-autocomplete": "^2.0.3",
+    "axios": "^0.21.1",
```
callumacrae, line 11: Is this required? It's a pretty large library and only being used for a simple request.

Fetch when polyfilled with https://github.com/github/fetch covers our browser support and is considerably smaller :)

app/webpacker/packs/school_search.js
```diff
@@ -0,0 +1,35 @@
+import accessibleAutocomplete from "accessible-autocomplete";
+
+const axios = require("axios");
```
callumacrae, line 3: There should probably be a linting rule for this, but it's generally bad practice to mix `import` and `require` unless there's a specific reason you're doing it

app/models/school.rb
seb-siem-softwire: Done

app/serializers/school_serializer.rb
seb-siem-softwire: Done

app/models/school.rb
seb-siem-softwire: Done

package.json
seb-siem-softwire: We can change it later on - for now it does the work

app/webpacker/packs/school_search.js
seb-siem-softwire: More than happy to incorporate any suggestions - not a javascript specialist so I will rely on any input

app/views/school_search/show.html.erb
seb-siem-softwire: @ethanmills are you ok with it?

app/webpacker/packs/school_search.js
kicferk1: I prefer using `import`s. :) Much more readable imo, perhaps not here, but when you get into more complicated structures it makes all the difference.

kicferk1 (approved):

kicferk1:
Oh no, Rubocop is angry!

rupollbabu (approved):

package.json
callumacrae: Is there a specific reason it can't be done now such as the work having a tight deadline? Or do you have a plan to remove the library once this PR is merged?

callumacrae (changes requested):
See axios comment :)

package.json
seb-siem-softwire: Given today is sprint planning

amazimbe (approved):

app/webpacker/packs/school_search.js
seb-siem-softwire: Done

package.json
seb-siem-softwire: Done

app/views/school_search/show.html.erb
ethanmills: Ok for now. I think it will need to be removed from this page in the upcoming sprint, as finding schools is one of the things we'll be working on.

seb-siem-softwire: 👍

app/webpacker/packs/school_search.js
```diff
@@ -0,0 +1,32 @@
+import accessibleAutocomplete from "accessible-autocomplete";
+import fetch from "node-fetch";
```
callumacrae, line 2: The package for this is https://www.npmjs.com/package/whatwg-fetch, node-fetch is something backend JS!

You won't need to import anything, too - just add `import 'whatwg-fetch'` to application.js and then use `fetch` in this file as it's a browser API.

seb-siem-softwire: Done

app/webpacker/packs/school_search.js
```diff
@@ -1,5 +1,5 @@
 import accessibleAutocomplete from "accessible-autocomplete";
-import fetch from "node-fetch";
+import "whatwg-fetch";
```
callumacrae, line 2: it's a very minor thing, but this would be best in application.js as probably won't be specific to this file for very long if any other JS functionality is added

callumacrae (approved):

app/webpacker/packs/school_search.js
seb-siem-softwire: Done

PR #98, opened 2021-02-16
Updated school registration workflow

amazimbe:
### Guidance to review
A few tickets combined in this pr and it is a big one but I'm hoping most of the code will be familiar. 

Copy / content updates
https://trello.com/c/VrvFJqzk/201-copy-updates-for-school-registration-workflow

New school registration:
https://trello.com/c/XnJZfG4j/203-remove-email-from-urn-form

Re architecting the backend:
Mostly creating separate controllers for each page and redirecting to the next action as opposed to just rendering another template. Hopefully, this will reduce complexity in the controllers and make them easier to test as well as easier to add additional workflow pages.


app/views/registrations/school_profile/show.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% content_for :before_content, govuk_back_link(
+      text: "Back",
+      href: registrations_question_two_path)
```
callumacrae, line 5: Should the back links all go back to the same page or should they go back to the previous step?

app/views/registrations/no_participants/show.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    href: registrations_question_one_path)
+  %>
+    <h1 class="govuk-heading-l">You can register later</h1>
+    <p class="govuk-body">You only need to register if:</p>
+    <ul class="govuk-list govuk-list--bullet govuk-!-margin-bottom-6">
+      <li>you are expecting early career teachers</li>
+      <li>you are ready to confirm details about your induction programme</li>
+    </ul>
+    <p class="govuk-body">You must register by <span class="govuk-!-font-weight-bold">30 September</span>.</p>
+    <p class="govuk-body">In the meantime, check our guidance for <a href="#">schools choosing an induction programme</a>.</p>
```
ethanmills, line 14: Do we want a TODO on the link?

app/views/registrations/school_profile/show.html.erb
amazimbe: To the previous step

app/views/registrations/no_participants/show.html.erb
amazimbe: Sure

app/views/registrations/school_not_confirmed/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+      href: registrations_school_profile_path)
+    %>
+    <h2 class="govuk-heading-l">Your school needs to complete registration</h2>
+    <p>You or one of your colleagues may have tried to register before and need verify your email.</p>
+    <p>This email will be used to sign in so check with your colleagues if you need access.</p>
+    <p>What you can do</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>check your inbox for an email from <span>TWD Early Careers Framework</span> (it may be in your spam or junk folder)</li>
+      <li>or your colleague can verify, using the link sent to their email.</li>
+      <li>if the link has expired, you can <%= govuk_link_to "register", registrations_question_one_path %> again.</li>
```
kicferk1, line 14: A bit weird - two of the options end with a `.`, and all start with a lowercase, hmm. 

Then again, I imagine it came from designs, so not sure if we want to pursue changing that.

app/views/registrations/school_not_eligible/show.html.erb
```diff
@@ -1,8 +1,12 @@
...
   <div class="govuk-grid-column-two-thirds">
+    <% content_for :before_content, govuk_back_link(
+      text: "Back",
+      href: registrations_school_profile_path)
+    %>
     <h2 class="govuk-heading-l">Sorry, your school is not eligible for the ECF</h2>
-    <p>Only schools in England are eligible for the ECF.</p>
-    <p>The following institutions are also not eligible:</p>
+    <p class="govuk-body">Only schools in England are eligible for the ECF.</p>
+    <p class="govuk-body">The following institutions are also not eligible:</p>
```
kicferk1, line 9: Does that change the style?

I thought we add `govuk-body` to our main element
```
      <main class="govuk-main-wrapper govuk-body" id="main-content" role="main">
```
so that we don't have to add `govuk-body` everywhere.

app/views/registrations/school_profile/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+    <h2 class="govuk-heading-l">What is your school Unique Reference Number(URN)?</h2>
+
+    <%= form_for(:school, model: @school_profile_form, url: registrations_school_profile_path) do |f| %>
+      <label class="govuk-hint">
+        This is a 6 digit number. Not sure? <a target='_blank' href='https://www.get-information-schools.service.gov.uk/'>Search for your school</a> on our register.
+        <%= f.govuk_text_field(:urn, width: 20, label:  nil) %>
+      </label>
+      <%= f.govuk_submit "Continue" %>
+    <% end %>
```
kicferk1, line 16: Just an idea that you might want to relay to Al (if it isn't in the works already) - when we have the component for school search that Sebastian was working on, it might be a lot easier to use that for finding the URN.

app/views/registrations/school_not_confirmed/show.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% content_for :before_content, govuk_back_link(
+      text: "Back",
+      href: registrations_school_profile_path)
+    %>
+    <h2 class="govuk-heading-l">Your school needs to complete registration</h2>
+    <p>You or one of your colleagues may have tried to register before and need verify your email.</p>
+    <p>This email will be used to sign in so check with your colleagues if you need access.</p>
+    <p>What you can do</p>
```
kicferk1, line 10: Do we want `:` after the text?

spec/forms/school_profile_form_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe SchoolProfileForm, type: :model, with_audited: true do
```
ethanmills, line 5: What does the `with_audited` mean here?

amazimbe: copy, paste. I'm not even sure I need it. I'll try and take it out

ethanmills: Was it copied from somewhere else in this project? If we don't need it could you take it out from there too please? 🙂

app/controllers/registrations/account_not_found_controller.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+class Registrations::AccountNotFoundController < ApplicationController
+end
```
kicferk1, line 4: How do empty controllers work exactly? I'm wondering out of my own curiosity,

app/views/registrations/account_not_found/show.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+  <% content_for :before_content, govuk_back_link(
+    text: "Back",
+    href: new_user_session_path)
+  %>
+
+    <h1 class="govuk-heading-l">You do not have an account</h1>
+    <h2 class="govuk-heading-m">School coordinators or leads</h2>
```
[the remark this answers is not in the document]
amazimbe: Not sure what it ought to be: should <% content_for be on the same column as <h1>?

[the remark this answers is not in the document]
callumacrae: yep!

app/views/registrations/school_not_confirmed/show.html.erb
amazimbe: It did come from the designs, that's right. I'll speak to Shani about it. Personally, I don't have any strong feelings about it.

app/controllers/registrations/user_profile_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
```
ethanmills, line 1: Am I right in thinking that at the moment we're not checking the email and school match in any way? If so, could we have a ticket for adding that, and a TODO against that ticket?

ethanmills: We might not know how to do it just yet, but it's important we don't forget 😛

app/views/registrations/account_not_found/show.html.erb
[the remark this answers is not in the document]
kicferk1: I would be tempted to move the `content_for` outside of the row and column divs - I think it gets ripped out of them and put before main anyway (in `application.html.erb`.

app/views/registrations/school_not_eligible/show.html.erb
amazimbe: I noticed the designer did have govuk-body on every paragraph in his prototype. I haven't seen it make any difference so you're probably correct that it's not necessary in the paragraphs

app/views/registrations/school_profile/show.html.erb
amazimbe: Yes, he's got a ticket for it already and the plan is to use it there

app/views/registrations/school_not_confirmed/show.html.erb
amazimbe: The content doesn't have one https://miro.com/app/board/o9J_ldVNkCY=/?moveToWidget=3074457354086350072&cot=14 but I think it would be more correct to add it. I'll speak to shani

app/controllers/registrations/account_not_found_controller.rb
amazimbe: Convention; they render the view that corresponds to the action.

app/controllers/registrations/user_profile_controller.rb
amazimbe: Yes, we are no longer matching schools and user emails. Happy to add a ticket but I don't think I need a todo as well. For the broken links etc, I agree that can be taken care of with a todo but for a feature, I would rather just have the ticket.

app/views/registrations/no_decision/show.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+  <% content_for :before_content, govuk_back_link(
+    text: "Back",
+    href: registrations_question_two_path)
+  %>
+    <h1 class="govuk-heading-l">You can register later</h1>
+    <p class="govuk-body">See our guidance for <a href="#">schools choosing an induction programme</a>.</p>
```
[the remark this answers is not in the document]
callumacrae: From an accessibility point of view, it would be best if the entire thing were a link - but failing that, what's there already is better than just "our guidance".

Some screen reader users browse websites using links only (without the surrounding text), so if there's a link saying something like "see more", they have no idea what it's about.

app/controllers/registrations/user_profile_controller.rb
ethanmills: Ok, sure

ethanmills (approved):
Couple of small things, looks good otherwise

app/views/registrations/no_decision/show.html.erb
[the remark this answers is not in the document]
kicferk1: Makes sense, I guess that's a good reason to avoid links intertwined with text :P Oh well, something to think about later. Maybe something to flag to our designers.

app/controllers/registrations/school_profile_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    params.require(:school).permit(:urn)
+  end
+
+  def load_school
+    session["school_urn"] = school_params[:urn]
+    @school = School.find_by(urn: school_params[:urn])
+  end
+
+  def redirect
+    if !@school.eligible?
```
seb-siem-softwire, line 31: could we create a helper method for not_eligable?

app/forms/school_profile_form.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class SchoolProfileForm
+  include ActiveModel::Model
+
+  attr_accessor :urn
+
+  validates :urn, presence: { message: "Enter a school URN" }
+
```
seb-siem-softwire, line 9: Could we remove one line so that all validations are next to each other

app/controllers/registrations/school_profile_controller.rb
amazimbe: In here it is hust a rails method. But that will be changing with some of the work Ethan is doing. I don't think a helper would be correct for domain knowledge too.

app/views/users/sessions/new.html.erb
```diff
@@ -1,5 +1,10 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+  <% content_for :before_content, govuk_back_link(
```
seb-siem-softwire, line 3: This could be on one line

app/controllers/registrations/school_profile_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      load_school
+      redirect
+    else
+      render :show
+    end
+  end
+
+private
+
+  def school_params
```
seb-siem-softwire, line 21: Nice!

rupollbabu (approved):
I have created URN and Email Validations tickets for test coverage. Edge test cases for Email required.

app/controllers/registrations/school_profile_controller.rb
ethanmills: What about `unless`?

seb-siem-softwire: Bad pattern @ethanmills mixing if and unless - it takes more mental computation when reading the code you either stick to one or another

kicferk1: Bad pattern is Ruby including unless imo.

But while it's there, a one line method using `unless` while some other methods use `if` doesn't look like a bad pattern to me.

app/views/users/sessions/new.html.erb
amazimbe: Yes, I probably copied it from somewhere where it was split. I'll change it

app/forms/school_profile_form.rb
amazimbe: sure

app/controllers/registrations/school_profile_controller.rb
ethanmills: For me personally, I find it easier to read an `unless` than an `if !`. But if we don't ever want to use unless, then we should enforce that through linting.

amazimbe: Ruby had a lofty go to read just as natural language hence rubysims like unless. That said, I find it ok as a guard in a method call but would steer well clear of unless xxxx else xx end or as in this case, using it with a load of other conditionals. And it could just be my bad English but sometimes I kind of forget what unless means.

amazimbe:
@here could I have a final look, please. I have addressed all the feedback. I picked up a few new commits from develop while trying to fix some conflicts. I had to merge develop into my branch to avoid having to rebase 40+ commits. All will be well after merging though

kicferk1 (approved):
Mostly looks good. Small comment about govuk-body classes.

app/controllers/registrations/school_profile_controller.rb
ethanmills: Yep, I agree, I hadn't realised this was the start of a chain of elsifs

ethanmills (approved):

PR #100, opened 2021-02-16
ECF-RP-129: Add sparsity data

ethanmills:
### Context
PR 3/3. This one actually adds the sparsity

- Add a model for recording when a district is sparse
- Import sparsity data from csv

### Testing
- Unit testing


app/services/sparsity_importer.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    @previous_district_sparsities = DistrictSparsity.latest.to_a
+
+    CSV.foreach(data_file, headers: true, encoding: "ISO-8859-1:UTF-8") do |row|
+      new_district = update_lad_sparsity(row)
+      @previous_district_sparsities.delete_if { |old_district| old_district.local_authority_district == new_district.local_authority_district }
+    end
+
+    @previous_district_sparsities.each { |district_sparsity| district_sparsity.update!(end_year: start_year) }
+  end
```
kicferk1, line 27: Hmm, what is this doing exactly? 

Delete all sparsity data for a year, recreate them from rows. That's fine, I can get that from code.

And what do we do with the latest sparsities? I don't really understand that. Can we handle these things separately? Double the csv processing time is annoying, but code understandability trumps efficiency at our stage of development :/

ethanmills: No, keeping track of anything was there before, but is not in the new CSV. We then set the end_year for that

kicferk1: Right, and why do we need to do either?
Is the CSV going to miss some sparsity data? Is this related to lads changing?

We should make some of these 'why's answered via naming things more explicitly. Even if it means we add a method that is a one-liner.

Basically, I find it really hard to understand what's going on without the exact context. 

Also, `previous_district_sparsities` probably doesn't need to be an `@` variable, right?

rupollbabu (approved):
Test Cover for Sparsity data module.

app/services/sparsity_importer.rb
ethanmills: This is not a fixed list. In 2021, a given list of LADs will be sparse. In 2022, we will get a new list, and it may be different. We do not get a list of all LADs with a boolean, we get a list of sparse LADs. As such, we need to work out which LADs we currently have marked as sparse, but are now being told are not sparse anymore. We then need to set the `end_year` on those records so we know when the stopped being sparse

ethanmills: I am

1.  Creating a list of all sparse LADs before the import
2. Ensuring a valid record exists for a given row
3. Removing that LAD from my list
4. When I have imported all rows, I am left with a list of LADs we think are sparse, but aren't any more
5. I'm then setting the `end_year` on those records

kicferk1: I think that logic will fail if you import a year from the past. Which I guess "will never happen". Ehhh, I already don't like the pattern of including years in the initialiser for unit testing. 

I think there is too many things going on at the same time here - could you have one method to `initialise_sparsity_records` and another of `get_sparsity_records_to_finish`?

Then the whole thing would be something like
```
new_sparsity_records = initialise_sparsity_records(...)
sparsity_records_to_finish = get_sparsity_records_to_finish(new_sparsity_records)

DistrictSparsity.where(start_year: start_year).destroy_all
new_sparsity_records.save_all # (or however it is you want to save)
sparsity_records_to_finish.finish_all # (or however it is you want to finish them)
```

I think it would be a lot easier to understand what is happening then - at least to me it would.

kicferk1: You can even just store the updated records rather than initialised records (and move destroying existing ones further up) - but I think separate loops for inserting new records and for computing the records we want to finish are going to be a big improvement in readability.

ethanmills: We can provide the year to #run, no big difference

ethanmills: How's that?

app/services/sparsity_importer.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def districts_no_longer_sparse
+    DistrictSparsity.latest
+                    .joins(:local_authority_district)
+                    .where.not(local_authority_district: { code: latest_sparse_districts })
+  end
+
+  def latest_sparse_districts
+    @csv ||= CSV.read(data_file, headers: true, encoding: "ISO-8859-1:UTF-8")
+    @csv["LAD_CODE"]
```
kicferk1, line 52: Whoa, will that give the list of entries in that table? Amazing :O

app/services/sparsity_importer.rb
kicferk1: Much, much better, will have another look tomorrow before approving though :)

app/models/local_authority_district.rb
```diff
@@ -3,5 +3,18 @@
...
+
+  def sparse?(year = nil)
+    if year.nil?
+      district_sparsities.latest.any?
+    else
+      district_sparsities.for_year(year).any?
+    end
+  end
+
+  scope :only_with_uplift, lambda { |year|
```
amazimbe, line 16: I can't see a unit test for this scope. Any reason why.

ethanmills: Added

kicferk1 (approved):
Just a small question

app/models/district_sparsity.rb
```diff
@@ -0,0 +1,15 @@
...
+  belongs_to :local_authority_district
+
+  scope :latest, -> { where(end_year: nil) }
+  scope :for_year, lambda { |year|
+    query = <<~SQLSNIPPET
+      district_sparsities.start_year <= ?
+          AND (district_sparsities.end_year > ?
+              OR district_sparsities.end_year IS NULL)
+    SQLSNIPPET
+    where(query, year, year)
```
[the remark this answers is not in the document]
ethanmills: I don't think that can happen. When we import, we delete everything that starts that year. Starting and ending the same year is the same as not existing

PR #101, opened 2021-02-16
Fix bug where school search JS ran on every page, throwing an error

callumacrae:
image

This was being thrown on every page but the school search page - wrapping it in a conditional fixes it.

### DIFF
--- a/app/webpacker/packs/school_search.js
+++ b/app/webpacker/packs/school_search.js
@@ -17,15 +17,17 @@ function suggest(query, populateResults) {
     .then((data) => populateResults(data));
 }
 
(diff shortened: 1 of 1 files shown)


seb-siem-softwire (approved):

ethanmills (approved):

kicferk1 (approved):
And no amount of automated testing would have helped us (maybe?)

callumacrae:
> And no amount of automated testing would have helped us (maybe?)

On the contrary - the reason this blocked the cypress work is because cypress fails on uncaught errors, so this was making every single test fail!

PR #102, opened 2021-02-17
Improve front-end linting setup

callumacrae:
Some things (such as me just using completely random quotes all over the place) were getting through the linting setup, so I've made it a bit more strict.

I've also added prettier, which I'm a big fan of. I've ran it in this pull request so you can see the kind of changes it makes. Feel free to discuss.


.eslintrc.js
```diff
@@ -1,17 +1,21 @@
 module.exports = {
+  root: true,
   env: {
     browser: true,
     es2021: true,
   },
   extends: [
+    "eslint:recommended",
     "airbnb-base",
-    "plugin:cypress/recommended",
```
ethanmills, line 8: Why remove this?

callumacrae: I've moved it into the eslintrc in the cypress directory

spec/cypress/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  extends: ["plugin:cypress/recommended"],
+  rules: {
+    "import/no-commonjs": ["off"],
+  },
+};
```
ethanmills, line 6: Does this "stack" with the main file? Do those rules apply too?

callumacrae: yep! eslint looks upwards until it finds a `root: true` and merges everything it finds

ethanmills: Nice!

kicferk1 (approved):
Just to make sure, this is just adding prettier and result of running it, right?

callumacrae:
> Just to make sure, this is just adding prettier and result of running it, right?

yep

ethanmills (approved):

rupollbabu (approved):

PR #105, opened 2021-02-22
[ECF-1-251] Persist induction programme choice

amazimbe:
### Context
New page to allow school induction coordinators to make a choice about the induction programme they want to run.

New model school_cohorts to store information about the choice of induction programme for a school.


app/controllers/induction_programme_choices_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class InductionProgrammeChoicesController < ApplicationController
+  before_action :authenticate_user!
+  before_action :ensure_induction_coordinator
+
+  def show
+    @programme_choices = [
+      OpenStruct.new(id: "funded_training_provider", name: "Use a training provider funded by the Department for Education"),
+      OpenStruct.new(id: "free_development_materials", name: "Use the free development materials"),
```
ethanmills, line 10: Would it make sense to set IDs as `full_induction_programme` and `core_induction_programme` so we're always using the same terms?

app/controllers/induction_programme_choices_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      OpenStruct.new(id: "design_our_own", name: "Design your own induction based on the Early Career Framework"),
+      OpenStruct.new(id: "not_yet_known", name: "I don't know yet"),
+    ]
+  end
+
+  def create
+    @cohort = Cohort.find_or_create_by!(start_year: 2021)
```
kicferk1, line 17: Can we have a todo here to remove the hard-coded value?

kicferk1: Or do we expect to require devs to change this value yearly?

rupollbabu (approved):

app/controllers/induction_programme_choices_controller.rb
amazimbe: Happy to do that but I'm not completely sure which one is which. Would funded_training_provider be cip? What about free_development_materials and the rest?

app/controllers/induction_programme_choices_controller.rb
amazimbe: Yes, happy to move it to a constant actually.

.env.template
```diff
@@ -2,6 +2,7 @@ DB_USERNAME=
 DB_PASSWORD=
 GOVUK_NOTIFY_API_KEY=
 DOMAIN=http://localhost:3000
+CURRENT_COHORT=2021
```
kicferk1, line 5: I think you might have just made more problems for yourself :D 

I was thinking more along the lines of adding a comment linking to a trello card saying "TODO: Figure out how to change cohort". I am not convinced that sticking it in environment makes it any easier. 

But, what I am certain of, is that if you want it in an environment variable, you will have to add it to a few more places. If you globally search for `GIAS_API_ROOT` you will see what I mean. Otherwise, I don't think it's going to work very well :/

amazimbe: I noticed I'll need to add it to the env files. But yeh I get your point. I'll just add the todo.

kicferk1 (approved):
```
@cohort = Cohort.find_or_create_by!(start_year: 2020)
```

Well, we probably want the year to be 2021, but other than that looks good to me :)

PR #107, opened 2021-03-01
Add automated accessibility tests for some pages

callumacrae:

### DIFF
--- a/app/views/users/sessions/login_email_sent.html.erb
+++ b/app/views/users/sessions/login_email_sent.html.erb
@@ -1,6 +1,6 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-l">Check your email</h2>
+    <h1 class="govuk-heading-l">Check your email</h1>
     <p>We have sent you an email to complete your sign in.</p>
     <p>If you do not get an email:</p>
     <ul class="govuk-list govuk-list--bullet">
--- a/spec/cypress/integration/a11y.js
+++ b/spec/cypress/integration/a11y.js
@@ -1,0 +1,48 @@
+describe("Accessibility", () => {
+  it("Landing page should be accessible", () => {
+    cy.visit("/");
+    cy.checkA11y();
+  });
+
+  it("Login should be accessible", () => {
+    cy.visit("/users/sign_in");
+    cy.checkA11y();
+
+    // School not registered page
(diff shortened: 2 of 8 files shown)


app/views/users/sessions/login_email_sent.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-l">Check your email</h2>
+    <h1 class="govuk-heading-l">Check your email</h1>
```
kicferk1, line 3: I always wanted to sort out our header mess, potentially there is still some, but slightly less now :)

spec/cypress/integration/a11y.js
```diff
@@ -0,0 +1,48 @@
+describe("Accessibility", () => {
```
seb-siem-softwire, line 1: At some point, I would recommend sitting down and exploring page object pattern architecture or those tests will become unmaintainable in the future. Good for now but in months time lets have a chat about improving the architecture

seb-siem-softwire (approved):

kicferk1 (approved):

spec/cypress/integration/a11y.js
callumacrae: There are architectural problems with having these tests separate to the behavioural tests, but I don't see how using the page object pattern would fix them in this context.

Remember that these are functional tests, not unit tests - functionality shouldn't be split across multiple tests, so the way we're using `.as()` with Cypress already solves the problem of selectors changing.

Also, using `data-cy` or `data-test` attributes as the Cypress best practices suggests seems a much nicer approach than maintaining a library of selectors in our tests.

kicferk1: Cypress best practices can suggest what they want, I don't really like the idea of having a bunch of attributes stumbling around, with the sole purpose of making my test architecture easier.

Key reason being, when I read the html and try to understand what it's doing and why, the last thing I want to be thinking about it 'ah, nice attribute, it makes our tests less brittle'.

I think I'd rather maintain the selectors :P

callumacrae: There's rails plugins to remove attributes when they're sent to actual users. We have attributes for our CSS and JavaScript in the HTML, why is attributes to stop us having to maintain a (brittle) selector library any worse?

PR #111, opened 2021-03-04
ECF-RP-240: Update school import

ethanmills:
### Context
We now import school data from the private GIAS API, not the public file

- Create a GIAS API client
- Update the import to use the custom GIAS extract

### Guidance to review
commit-wise

### Testing
- unit tests for import
- unit tests for api client


ethanmills:
[Update school data import to use GIAS extract](https://trello.com/c/I0eihe4z/240-update-school-data-import-to-use-gias-extract)

lib/gias_api_client.rb
```diff
@@ -0,0 +1,56 @@
...
+    parts.second.body.to_s
+  end
+
+  def temp_files_from_zipped_response(data)
+    files = {}
+
+    # We need to write the zip file to disk rather than read it from a buffer because of this:
+    # https://github.com/rubyzip/rubyzip#notice-about-zipinputstream
+    zipfile = Tempfile.new(encoding: "ASCII-8BIT")
+    zipfile.write(data.force_encoding("ASCII-8BIT"))
```
amazimbe, line 31: I doubt we need this line if we have already specified the encoding in zipfile = Tempfile.new(encoding: "ASCII-8BIT")

amazimbe (approved):
Just one minor comment / question. Otherwise, all looks good.

Dockerfile
```diff
@@ -47,7 +47,6 @@ FROM ${BASE_RUBY_IMAGE_WITH_GEMS_AND_NODE_MODULES} AS assets-precompile
 
 ENV GOVUK_APP_DOMAIN="http://localhost:3000" \
     GOVUK_WEBSITE_ROOT="http://localhost:3000" \
-    GIAS_API_ROOT="https://ea-edubase-api-prod.azurewebsites.net" \
```
kicferk1, line 50: Do we not want to add the new env variables here?

app/services/school_data_importer.rb
```diff
@@ -12,9 +12,10 @@ def initialize(logger, start_year = Time.zone.now.year)
     @start_year = start_year
   end
 
-  # TODO: Register and Partner 147: Figure out how exactly we want to handle school updating
   def run
-    CSV.foreach(schools_data_file.path, headers: true, encoding: "ISO-8859-1:UTF-8") do |row|
+    files = school_data_files
+
+    CSV.foreach(files["ecf_tech.csv"].path, headers: true, encoding: "ISO-8859-1:UTF-8") do |row|
```
kicferk1, line 18: Are there other files in there? What's in them?

app/services/school_data_importer.rb
```diff
@@ -26,40 +27,29 @@ def run
...
     school.school_type_name = row.fetch("TypeOfEstablishment (name)")
-    school.capacity = row.fetch("SchoolCapacity").to_i
     school.address_line1 = row.fetch("Street")
     school.address_line2 = row.fetch("Locality")
     school.address_line3 = row.fetch("Town")
-    school.address_line4 = row.fetch("County (name)")
-    school.country = row.fetch("Country (name)")
     school.postcode = row.fetch("Postcode")
     school.ukprn = row.fetch("UKPRN")
-    school.previous_school_urn = row.fetch("PreviousEstablishmentNumber")
```
kicferk1, line 51: oof, scary

without this data, are we gonna have troubles figuring out which school is which after a change?

lib/gias_api_client.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def fetch_extract
+    client = Savon.client(wsdl: Rails.application.config.gias_api_schema)
+    res = client.call(
+      :get_extract,
+      message: { "tns:Id" => Rails.application.config.gias_extract_id },
+      wsse_auth: [Rails.application.config.gias_api_user, Rails.application.config.gias_api_password],
+      wsse_timestamp: true,
+    )
+    parts = parse_body(res.http)
```
kicferk1, line 21: is our request using `http`? or `https`?

Not that I know what the hell `res.http` does, just wondering.

lib/gias_api_client.rb
```diff
@@ -0,0 +1,56 @@
...
+    parts.second.body.to_s
```
kicferk1, line 22: Is this how we get rid of the request headers?

lib/gias_api_client.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    files
+  end
+
+  def parse_body(http)
+    Mail::Part.new(
+      headers: http.headers,
+      body: http.body,
+    ).body.split!(boundary(http)).parts
```
kicferk1, line 50: Are we re-purposing an email parser to deal with out zip file? :O

Dockerfile
ethanmills: Nope, we don't need this var for the build

app/services/school_data_importer.rb
ethanmills: Yes. There's the groups file, the group links file, and the links file. Contents per https://www.get-information-schools.service.gov.uk/Downloads. Will be using the groups file later to set up MATs

app/services/school_data_importer.rb
ethanmills: That's what the links file is for 🙂

lib/gias_api_client.rb
ethanmills: It's not that. The initial request will be https to https://ea-edubase-api-prod.azurewebsites.net/edubase/schema/service.wsdl
, and subsequent requests dictated by the contents of that schema. This just return the HTTPI object that has header and body information

lib/gias_api_client.rb
ethanmills: Yes, in essence. I adapted some code from a long dead library: [savon-multipart](https://github.com/savonrb/savon-multipart/blob/5e5a247f3fc5b5e2b25e2eaa76893fa34bda8adf/lib/savon/multipart/response.rb)

lib/gias_api_client.rb
ethanmills: See previous comment. It's basically a MIME parser, and the multipart response is most commonly found in emails

kicferk1: Ech, well, not like I can think of a better way.

kicferk1 (approved):

lib/gias_api_client.rb
ethanmills: This is incredibly touchy about encodings. If you're happy with it, I'd like to leave this as it is, as it definitely works like this

PR #113, opened 2021-03-05
Adds first component tests

seb-siem-softwire:
This pull request adds a test suite for components.

Some of the surprises and difficulties observed on content_for components as they don't render the normal way - solved but required digging deeper into the code for view components.

Overall - I think components are great - self-contained lego blocks that are very easy to test and as long as we keep views as flat as possible and not nest components - we should see plenty of benefits.

The one major benefit is the speed and simplicity of the tests - as long as we put the right variables in - we would always get predictable results.

There is much more to components and their configurability - for example using builder pattern (for example .with_header, .with_footer)

### DIFF
--- a/spec/components/school_information/details/breadcrumb_component_spec.rb
+++ b/spec/components/school_information/details/breadcrumb_component_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ::SchoolInformation::Details::BreadcrumbComponent, type: :component do
+  let(:selected_cohort){ create(:cohort, start_year: 2021) }
+  let(:component){ described_class.new(selected_cohort: selected_cohort) }
+  let(:render_component){ render_inline(component) }
+  let(:before_content_output){ component.content_for(:before_content) }
+
+  before(:each) do
+    render_component
+  end
+
+  it 'includes Schools breadcrumb' do
+    expect(before_content_output).to include 'Schools'
+    expect(before_content_output).to include lead_providers_your_schools_path
+  end
+
+  it 'includes Cohort breadcrumb' do
+    expect(before_content_output).to include '2021 cohort'
+    expect(before_content_output).to include lead_providers_your_schools_path(selected_cohort_id: selected_cohort.id )
+  end
+end
--- a/spec/components/school_information/details/breadcrumb_component_spec.rb
+++ b/spec/components/school_information/details/breadcrumb_component_spec.rb
@@ -3,22 +3,22 @@
 require "rails_helper"
 
 RSpec.describe ::SchoolInformation::Details::BreadcrumbComponent, type: :component do
-  let(:selected_cohort){ create(:cohort, start_year: 2021) }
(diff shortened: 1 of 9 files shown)


ethanmills:
Could you use github's draft PR feature please?

callumacrae:
In my opinion, a lot of this would be better as visual tests - which are going to be added anyway. Testing that a component contains some text when rendered doesn't actually check that the component is being displayed properly, but still requires maintenance when that text is changed in the future.

seb-siem-softwire:
We don't have to choose between one or the other - the same way as you can have capybara, cypress, and unit tests. Component tests are at level of unit tests

callumacrae:
Sure, but by testing things twice you're adding unnecessary maintenance costs.

seb-siem-softwire:
Components tests are not end to end tests - they are discreet self-contained behaviour. I believe there are cases for both and given how fast are those and that one component equals one file - if the component changes or gets deleted - it's quite easy to modify the corresponding file. It's different for end to end testing

spec/components/school_information/details/breadcrumb_component_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  let(:selected_cohort) { create(:cohort, start_year: 2021) }
+  let(:component) { described_class.new(selected_cohort: selected_cohort) }
+  let(:render_component) { render_inline(component) }
+  let(:before_content_output) { component.content_for(:before_content) }
+
+  before(:each) do
+    render_component
+  end
+
+  it "includes Schools breadcrumb" do
```
amazimbe, line 15: I like it! The spec is pretty simple and easy to read

amazimbe (approved):

amazimbe:
Looks really good to me!

amazimbe:
> In my opinion, a lot of this would be better as visual tests - which are going to be added anyway. Testing that a component contains some text when rendered doesn't actually check that the component is being displayed properly, but still requires maintenance when that text is changed in the future.

@callumacrae just curious what visual tests are. Is that manually looking at the app?

callumacrae:
> Components tests are not end to end tests - they are discreet self-contained behaviour.

So are visual tests, when used on components.

> @callumacrae just curious what visual tests are. Is that manually looking at the app?

In the context of testing components, it would render the components in a browser, take a screenshot of them, then if something changed it would flag it in the PR. Basically it fully tests the appearance of a component, not just some of the copy :)

amazimbe:
> > Components tests are not end to end tests - they are discreet self-contained behaviour.
> 
> So are visual tests, when used on components.
> 
> > @callumacrae just curious what visual tests are. Is that manually looking at the app?
> 
> In the context of testing components, it would render the components in a browser, take a screenshot of them, then if something changed it would flag it in the PR. Basically it fully tests the appearance of a component, not just some of the copy :)

That would be really good. Would those be in cypress as well then?

callumacrae:
> That would be really good. Would those be in cypress as well then?

For components tests they'd probably be ran outside of cypress, but we'd use them in cypress too for other things :)

PR #114, opened 2021-03-05
ECF-1-307 Allow DFE admins to edit lead provider users

amazimbe:
### Context
No designs available

Allow DFE admins to edit lead provider users.

Currently they can only edit email and full name


app/controllers/admin/lead_provider_users_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class Admin::LeadProviderUsersController < Admin::BaseController
+  skip_after_action :verify_policy_scoped
+  before_action :load_lead_provider, only: %i[edit update]
+  before_action :load_lead_provider_user, only: %i[edit update]
```
kicferk1, line 6: Do we need `only`s? Are there any other actions here?

app/views/admin/lead_provider_users/edit.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      <%= f.govuk_text_field(
+            :full_name,
+            value: @lead_provider_user.full_name,
+            label: { text: "Full name" })
+      %>
+      <%= f.govuk_email_field(
+            :email,
+            value: @lead_provider_user.email,
+            label: { text: "Email" })
+      %>
```
kicferk1, line 15: That feels a bit boilerplatey. Aren't the fields clever enough to know what value the object has? 

Maybe we want `form_with model: @lead_provider_user`?

app/controllers/admin/lead_provider_users_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    @lead_provider = LeadProvider.find(params[:lead_provider])
+  end
+
+  def load_lead_provider_user
+    @lead_provider_user = User.find(params[:id])
+  end
+
+  def user_params
+    params.require(:user).permit(:full_name, :email)
+  end
```
kicferk1, line 34: @ethanmills that's the repetition I was thinking about - if admins can edit all users, we will have a ton of places with these types of parameters :/

ethanmills: Perhaps Pundit's `permitted_attributes`? https://github.com/varvet/pundit#strong-parameters

ethanmills: Specifically: https://github.com/varvet/pundit#user-content-strong-parameters:~:text=If%20you%20want%20to%20permit%20different,a%20permitted_attributes_for_%23%7Baction%7D%20method%20on%20your%20policy

kicferk1: holy crap, that sounds exactly like what would solve it

ethanmills:
How do you get to the edit page? I can't figure it out in the review app

app/controllers/admin/lead_provider_users_controller.rb
amazimbe: @kicferk1 What's 'only s'

kicferk1:
```
only: %i[edit update]
```
in each before_action

app/views/admin/lead_provider_users/edit.html.erb
amazimbe: That was my oversight. I'll correct it

app/controllers/admin/lead_provider_users_controller.rb
amazimbe: Got you

amazimbe: There will be more actions coming but be can add the restriction when those actions have been added

kicferk1: If what's coming is 'show', I would imagine we can remove it. Cause surely show will need it.

If what's coming is 'index', we can keep it. Cause index won't need it.

If something else, uh, I will let you choose :)

amazimbe:
> How do you get to the edit page? I can't figure it out in the review app

I'm mostly focusing on the endpoint but I'll probably be enabling the link on the admin users index page.  Another thing that crossed my mind was building /admin/lead-providers/:id/users. Just not sure if that wouldn't just be duplicating what /admin/suppliers/users is already doing. Happy to postpone that decision until I have added edit actions for the other types of user.

ethanmills:
I suspect we will need  /admin/lead-providers/:id to show users, delivery partners etc. for a lead provider. When Al is back on Monday I'll get him to unhide the wireframes for that and we can draw inspiration. I'll try to find some time with Nicola and maybe Gemma to go through what we designed and work out the minimum we need from it

Leercourt:
[Card 307 - Test Cases.docx](https://github.com/DFE-Digital/early-careers-framework/files/6100101/Card.307.-.Test.Cases.docx)

Hopefully I've done this right......

app/views/admin/lead_providers/lead_provider_users/edit.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">Edit user details</h1>
+    <%= form_for(@lead_provider_user, url: admin_lead_provider_user_path(@lead_provider_user)) do |f| %>
+      <%= f.govuk_error_summary %>
+      <%= f.govuk_text_field(
```
seb-siem-softwire, line 6: you could have this on one line, suggestion though - not a hard rule

spec/requests/admin/lead_providers/lead_provider_users_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Admin::LeadProviders::LeadProviderUsers", type: :request do
+  let(:lead_provider_profile) { create(:lead_provider_profile) }
+  let(:lead_provider_user) { lead_provider_profile.user }
+
+  before do
+    admin_user = create(:user, :admin)
```
seb-siem-softwire, line 10: You could put it to let and just leave admin user

amazimbe: Indeed. I left it as it's only used in this one place so I didn't see any use in having it as a variable

seb-siem-softwire: :) But it did end up as variable - plus there is a level of elegance of having variables in lets and grouping them together - not hard rule - I'll leave the decision to you. Just it looks nicer when you have all your declarations in lets

app/controllers/admin/lead_providers/lead_provider_users_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    class LeadProviderUsersController < Admin::BaseController
+      skip_after_action :verify_policy_scoped
+      before_action :load_lead_provider_user
+
+      def edit
+        authorize User
+      end
+
+      def update
+        authorize User
```
ethanmills, line 14: Might also be good to authorize LeadProviderProfile?

spec/requests/admin/lead_providers/lead_provider_users_spec.rb
kicferk1: Personally I find it more elegant to have variables that are as local as possible. I would expect things in `let` to be used in more than one place, precisely because that's the power of `let`s. Kind of how I would not make a variable an `@`-variable, unless I have to.

app/controllers/admin/lead_providers/lead_provider_users_controller.rb
ethanmills: Or even better, authorize @lead_provider_user. That way in future, we can write clever policies that check which profiles a user has before permitting editing

amazimbe:
> [Card 307 - Test Cases.docx](https://github.com/DFE-Digital/early-careers-framework/files/6100101/Card.307.-.Test.Cases.docx)
> 
> Hopefully I've done this right......

@Leercourt yes that looks good. We are building the user management incrementally though so best to test again when we are done as a checklist.

seb-siem-softwire (approved):

app/controllers/admin/lead_providers/lead_provider_users_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+      before_action :load_lead_provider_user
+
+      def edit
+        authorize @lead_provider_user
+      end
+
+      def update
+        authorize @lead_provider_user
+
+        if @lead_provider_user.update(permitted_attributes(@lead_provider_user))
```
kicferk1, line 16: Whoa, I was expecting it to require playing with our policies, but it seems someone had enough foresight to put `permitted_attributes` for the user in :O

kicferk1 (approved):

ethanmills:
[Allow DFE admins (super user) to edit a user.](https://trello.com/c/b0GGlRLe/307-allow-dfe-admins-super-user-to-edit-a-user)

PR #115, opened 2021-03-05
ECF-RP-315: refactor admin supplier routes

ethanmills:
### Context
Clean up the mess I made when creating suppliers. Namespace things better

### Guidance to review
Start with the routes file. Everything else is just a result of that. No functional changes


config/routes.rb
```diff
@@ -57,61 +57,48 @@
...
-
-      scope "lead-provider" do
-        get "choose-cip", controller: :lead_providers, action: :choose_cip, as: :new_lead_provider_cip
-        post "choose-cip", controller: :lead_providers, action: :receive_cip
-        get "choose-cohorts", controller: :lead_providers, action: :choose_cohorts, as: :new_lead_provider_cohorts
-        post "choose-cohorts", controller: :lead_providers, action: :receive_cohorts
-        get "review", controller: :lead_providers, action: :review, as: :new_lead_provider_review
-        post "/", controller: :lead_providers, action: :create, as: :create_lead_provider
-        get "success", controller: :lead_providers, action: :success, as: :new_lead_provider_success
+    scope :suppliers, module: "suppliers" do
```
amazimbe, line 60: This I really like i.e. using resources routes as much as possible. Looks a lot cleaner and requires less customisation.

app/controllers/admin/suppliers/delivery_partners_controller.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+module Admin
+  module Suppliers
```
amazimbe, line 4: I agree with this logical organisation too

amazimbe (approved):
@ethanmills It's definitely an improvement. Happy to merge

PR #112, opened 2021-03-04
ECF-RP-313: admins create admins

ethanmills:
### Context
Enable administrators to create other administrators

### Changes proposed in this pull request
Add a (very ugly) nav bar to the admin pages
Add a list of administrators
Form to create new administrators

### Testing
Request specs and Cypress

### Visual changes
image
image
image
image

PR #117, opened 2021-03-05
ECF-RP-315: edit delivery partners

ethanmills:
- [x] show delivery partner details
- [x] delete confirmation page
- [x] soft delete
- [x] edit
- [x] Cypress tests


db/schema.rb
```diff
@@ -171,7 +173,7 @@
     t.datetime "updated_at", precision: 6, null: false
     t.uuid "school_id", null: false
     t.uuid "lead_provider_id", null: false
-    t.uuid "cohort_id", default: "f85b3d42-9f6e-41e2-a014-0a09caea7e93", null: false
+    t.uuid "cohort_id", default: "d97d273b-7519-4b1b-a62f-a12c39c4b98c", null: false
```
seb-siem-softwire, line 176: I think we should remove the default cohort id - or in every pull request we will see a different id from each one of us after changing db

ethanmills: Happy for this to be a new PR? https://github.com/DFE-Digital/early-careers-framework/pull/112#discussion_r589345975

ethanmills:
[Allow DfE admin to edit and delete delivery partners (organisations, not users)](https://trello.com/c/A7hOWF6N/315-allow-dfe-admin-to-edit-and-delete-delivery-partners-organisations-not-users)

app/forms/delivery_partner_form.rb
```diff
@@ -3,20 +3,22 @@
...
   include ActiveModel::Model
 
-  attr_accessor :name, :lead_providers, :provider_relationships
+  attr_accessor :name, :lead_providers, :provider_relationships, :provider_relationship_hashes
   validate :lead_providers_and_cohorts_validation
 
+  def self.provider_relationship_value(lead_provider, cohort)
+    "{\"lead_provider_id\": \"#{lead_provider.id}\", \"cohort_id\": \"#{cohort.id}\"}"
+  end
+
```
ethanmills, line 12: This is pretty hacky. If anyone else can come up with a way to pre-populate nested checkboxes I'd love to hear it

app/helpers/supplier_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module SupplierHelper
+  def supplier_link(supplier)
+    supplier.instance_of?(LeadProvider) ? nil : admin_delivery_partner_path(supplier)
```
ethanmills, line 5: This will change when we have something to show for lead providers

app/forms/delivery_partner_form.rb
```diff
@@ -3,20 +3,22 @@
 class DeliveryPartnerForm
   include ActiveModel::Model
 
-  attr_accessor :name, :lead_providers, :provider_relationships
+  attr_accessor :name, :lead_providers, :provider_relationships, :provider_relationship_hashes
   validate :lead_providers_and_cohorts_validation
 
+  def self.provider_relationship_value(lead_provider, cohort)
+    "{\"lead_provider_id\": \"#{lead_provider.id}\", \"cohort_id\": \"#{cohort.id}\"}"
```
callumacrae, line 10: i'm not entirely sure what's going on here but can you use a hash and `to_json` instead of this?

ethanmills: I wasn't convinced that would guarantee order. Might be wrong though

spec/cypress/integration/admin/suppliers/create-delivery-partner.js
```diff
@@ -0,0 +1,147 @@
...
+    cy.appScenario("admin/suppliers/create_supplier");
+    const deliveryPartnerName = "New delivery partner";
+
+    cy.visit("/admin/suppliers");
+    cy.get(".govuk-button").contains("Add a new supplier").click();
+
+    cy.chooseSupplierName(deliveryPartnerName);
+    cy.clickBackLink();
+    cy.get("input[type=text]").should("have.value", deliveryPartnerName);
+    cy.get(".govuk-button").click();
```
callumacrae, line 76: This will break if another button is added to the page - I've been using `[name="commit"]` which I still don't think is great, but won't break if, say, a reset button is added to the form

spec/cypress/integration/admin/suppliers/create-lead-provider-user.js
```diff
@@ -0,0 +1,142 @@
...
+
+    cy.appEval(`User.find_by(email: "${userEmail}").present?`).then((result) =>
+      expect(result).to.equal(true)
+    );
+    cy.appEval(
+      `User.find_by(email: "${userEmail}").lead_provider.present?`
+    ).then((result) => expect(result).to.equal(true));
+  });
+
+  it.only("allows changing name choice", () => {
```
callumacrae, line 108: this looks like it was left in by mistake

(we should probably add https://www.npmjs.com/package/eslint-plugin-no-only-tests at some point - the single greatest linting rule there is)

spec/cypress/integration/admin/suppliers/delete-delivery-partner.js
```diff
@@ -0,0 +1,34 @@
+describe("Admin user deleting delivery partner", () => {
+  beforeEach(() => {
+    cy.login("admin");
+  });
+
+  it("should create a new delivery partner", () => {
```
callumacrae, line 6: create -> delete? :D

app/forms/delivery_partner_form.rb
callumacrae: that's a fair point

callumacrae:
image

you've added an empty file (or github is broken)

spec/cypress/support/admin/suppliers/commands.js
```diff
@@ -0,0 +1,61 @@
+Cypress.Commands.add("logout", () => {
```
callumacrae, line 1: There's already a logout command isn't there?

spec/cypress/integration/admin/suppliers/edit-delivery-partner.js
```diff
@@ -0,0 +1,68 @@
+describe("Admin user editing delivery partner", () => {
+  beforeEach(() => {
+    cy.login("admin");
+  });
+
+  it("should create a new delivery partner", () => {
```
ethanmills, line 6: I saw it first

spec/cypress/support/admin/suppliers/commands.js
ethanmills: Oops

.eslintrc.js
```diff
@@ -17,5 +17,7 @@ module.exports = {
   },
   rules: {
     "import/no-commonjs": ["error"],
+    "no-only-tests/no-only-tests": "error",
```
callumacrae, line 20: 🥳

Gemfile
```diff
@@ -24,8 +24,8 @@ gem "puma", "~> 5.0"
 # Transpile app-like JavaScript. Read more: https://github.com/rails/webpacker
 gem "webpacker", ">= 5.2.1"
 
-# Use ActiveStorage variant
-# gem 'mini_magick', '~> 4.8'
+# Soft delete
+gem "discard", "~> 1.2", ">= 1.2.0"
```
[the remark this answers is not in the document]
ethanmills: Nope, just had bad experiences with not locking versions in the past

app/controllers/admin/suppliers/delivery_partners_controller.rb
```diff
@@ -3,10 +3,10 @@
 module Admin
   module Suppliers
     class DeliveryPartnersController < Admin::BaseController
-      before_action :show, only: %i[show_users show_lps show_schools]
-
+      before_action :set_delivery_partner, only: %i[show delete edit update destroy]
       skip_after_action :verify_policy_scoped
 
+      # region create
```
amazimbe, line 9: I don't get this comment

ethanmills: Probably means the controller is too big :P

amazimbe: Probably adding IDE specific comments in the codebase isn't a good idea.

app/forms/delivery_partner_form.rb
amazimbe: I don't get why this has to be a string

app/helpers/supplier_helper.rb
amazimbe: instance_of reads a bit like java. Can we use is_a?

amazimbe: I think this is missing a unit test as well.

config/routes.rb
```diff
@@ -93,6 +93,12 @@
         get "review", controller: :supplier_users, action: :review, as: :new_supplier_user_review
         get "success", controller: :supplier_users, action: :success, as: :new_supplier_user_success
       end
+
+      resources :delivery_partners, only: %i[show edit update destroy], path: "delivery-partners" do
```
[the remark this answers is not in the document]
ethanmills: The primary nav component checks for `admin/suppliers` to mark the suppliers tab as active. What would be the problem with long paths?

app/controllers/admin/suppliers/delivery_partners_controller.rb
ethanmills: They're pretty widely supported https://code.visualstudio.com/updates/v1_17#_folding-regions

ethanmills: Happy to remove if you feel strongly though

spec/cypress/integration/admin/suppliers/create-delivery-partner.js
```diff
@@ -0,0 +1,147 @@
...
+      "input[name='delivery_partner_form[provider_relationship_hashes][]'][type=checkbox]"
+    ).should("have.length", 1);
+    cy.chooseFirstLeadProviderAndCohort();
+
+    cy.location("pathname").should(
+      "equal",
+      "/admin/suppliers/new/delivery-partner/review"
+    );
+    cy.get("main").should("contain", deliveryPartnerName);
+    cy.appEval(`LeadProvider.first.name`).then((leadProviderName) =>
```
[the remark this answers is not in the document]
ethanmills: What do you think about hardcoding the names in the scenario?

app/forms/delivery_partner_form.rb
seb-siem-softwire: https://stackoverflow.com/questions/31418673/is-order-of-a-ruby-hash-literal-guaranteed

spec/cypress/integration/admin/suppliers/create-delivery-partner.js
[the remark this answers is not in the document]
amazimbe: I agree, the db checks can stay in the request specs.

app/forms/delivery_partner_form.rb
ethanmills: It's being set as the `value` of a checkbox, so needs to be a string. The issue is I'm trying to create hypothetical `ProviderRelationship` object for each of the possible combinations. They don't have IDs because they don't exist

spec/cypress/integration/admin/suppliers/create-delivery-partner.js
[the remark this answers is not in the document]
ethanmills: This wasn't to check the db per-se, just to find the lead provider name in order to know what to look for in the frontend

app/controllers/admin/suppliers/delivery_partners_controller.rb
ethanmills: It even works for vim 😛 https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/vim-syntax-highlighting/#:~:text=Vim%20provides%20one%20final%20level,synGroup1%2CsynGroup2%2C...

app/views/admin/suppliers/delivery_partners/show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+
+<h1 class="govuk-heading-xl"><%= @delivery_partner.name %></h1>
+<h2 class="govuk-heading-m">Lead Providers</h2>
+<ul>
+<% @delivery_partner.lead_providers.uniq.each do |lead_provider| %>
+  <li><%= lead_provider.name %></li>
+    <% @delivery_partner.cohorts_with_provider(lead_provider).each do |cohort| %>
+    <%= cohort.display_name %>
+      <% end %>
+<% end %>
```
kicferk1, line 11: Some indentation shenanigans here

spec/models/provider_relationship_spec.rb
```diff
@@ -3,6 +3,13 @@
 require "rails_helper"
 
 RSpec.describe ProviderRelationship, type: :model do
+  let(:lead_provider) { create(:lead_provider) }
+  let(:delivery_partner) { create(:delivery_partner) }
+  let(:cohort) { create(:cohort) }
+  let!(:provider_relationship) do
+    ProviderRelationship.create(lead_provider: lead_provider, delivery_partner: delivery_partner, cohort: cohort)
+  end
```
kicferk1, line 11: oh oh, what does the `let!` do that `let` does not?

app/controllers/admin/suppliers/delivery_partners_controller.rb
callumacrae: > It even works for vim 😛

Not by default, from what I can tell!

app/views/admin/suppliers/delivery_partners/show.html.erb
callumacrae: https://github.com/prettier/plugin-ruby 👀

package.json
```diff
@@ -18,11 +18,12 @@
   "devDependencies": {
     "cypress": "^6.5.0",
     "cypress-intellij-reporter": "^0.0.6",
-    "eslint": "^7.13.0",
+    "eslint": "^7.21.0",
     "eslint-config-airbnb-base": "^14.2.1",
     "eslint-config-prettier": "^7.2.0",
     "eslint-plugin-cypress": "^2.11.2",
     "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-no-only-tests": "^2.4.0",
```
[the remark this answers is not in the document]
callumacrae: I suggested this! 😄 

Cypress is good at letting you run a single file, but not a single test in that file - that's where `.only` is super useful.

app/views/admin/suppliers/delivery_partners/show.html.erb
ethanmills: https://github.com/prettier/plugin-ruby/issues/371

spec/models/provider_relationship_spec.rb
ethanmills: It works like `before`. Runs even if you don't call it

kicferk1:
Also, a side comment, which might be taken as moaning (well, it is a little about that):

This PR adds a lot. It could almost certainly be broken up in several smaller parts - separate for editing delivery partners, separate for deleting them, separate for soft-deleting them, possibly more. I appreciate we want to get things through before the release, but lumping together large chunks of work is counterproductive - for example cause it makes reviews harder and longer. 

And I know we have 2 PRs adding `discard` now - I hope they don't conflict too badly, but that means we have you ans Seb figuring out ins and outs of that library in parallel, without basis in code. Which I think is a bit of a waste. 

If it was a single ticket, maybe it should have been broken up. Could be something to flag up.

spec/policies/lead_provider_policy_spec.rb
```diff
@@ -12,12 +12,14 @@
 
     it { is_expected.to permit_new_and_create_actions }
     it { is_expected.to permit_edit_and_update_actions }
+    it { is_expected.to forbid_action(:destroy) }
```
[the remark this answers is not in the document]
ethanmills: Yep. When I edited the policy for DPs I realised we weren't explicitly testing that deleting was forbidden for either DPs or LPs. Seemed like a good thing to test

spec/cypress/integration/admin/suppliers/create-delivery-partner.js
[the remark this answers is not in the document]
callumacrae: I'd be okay with hardcoding the names - the visual tests will break if names change anyway

app/views/admin/suppliers/delivery_partners/show.html.erb
callumacrae: https://github.com/adamzapasnik/prettier-plugin-erb 👀

app/forms/delivery_partner_form.rb
kicferk1: If we want a hypothetical relation, can we use `ProviderRelationship.new`? They still won't 'exist' in our database, but should be real enough for our code, right?

package.json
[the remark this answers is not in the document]
ethanmills: @kicferk1 the cypress support plugin for rubymine adds `only` to your code to run a single test then removes it again. It seems to be the only way

app/controllers/admin/suppliers/delivery_partners_controller.rb
kicferk1: Hehe, does vim require a lot of configuration to make it any good? Well yes, weeks of it ;)

I think there is a larger issue here, and the issue is "our controller is bloated because it serves both standard edit / delete actions and a multi-page flow from before". Maybe it's worth thinking about having separate controllers for the two? Maybe not in this PR, maybe as a tech debt ticket.

app/views/admin/suppliers/delivery_partners/choose_lead_providers.html.erb
```diff
@@ -1,36 +1,11 @@
...
-                :display_name,
-                legend: { hidden: true },
-                hint: { text: "What cohort/academic year are they working with?" }
-              )
-          %>
-        <% end %>
-      <% end %>
-    <% end %>
-  <% end %>
+  <%= render "lead_providers_cohorts", locals = { f: f } %>
```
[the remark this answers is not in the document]
kicferk1: Not that there is any action here, I just wanted to vent :P

app/forms/delivery_partner_form.rb
ethanmills: No. How do you refer to it across controller actions? It won't be in the DB, so won't have an ID

app/views/admin/suppliers/delivery_partners/_lead_providers_cohorts.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        hint: { text: "What cohort/academic year are they working with?" },
+        ) do %>
+        <% lead_provider.cohorts.each do |cohort| %>
+          <%= f.govuk_check_box(
+                :provider_relationship_hashes,
+                DeliveryPartnerForm.provider_relationship_value(lead_provider, cohort),
+                label: { text: cohort.display_name },
+                link_errors: true)
+          %>
+        <% end %>
```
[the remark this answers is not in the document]
ethanmills: It's also used in the controller

app/controllers/admin/suppliers/delivery_partners_controller.rb
amazimbe: I'd rather you took them out. What does the team think? @DFE-Digital/ecf

kicferk1: My opinion is: we need 2 controllers for it, done in a later tech-debt PR.

cfq:
> This PR adds a lot. It could almost certainly be broken up in several smaller parts
> If it was a single ticket, maybe it should have been broken up. Could be something to flag up.

You're right @kicferk1, I agree it should've been broken up. It hasn't been managed well and I'd like to pull us back. This was probably the most rushed bit of work we had until now, apart from @seb-siem-softwire's big PR. Let's not make this mistake again and bring it to the dev channel if you feel it's more productive to split PRs/tickets.

At least @ethanmills has managed it well by creating a draft PR, pushing often and getting feedback as he goes along.

app/forms/delivery_partner_form.rb
amazimbe: Something just feels wrong here. Surely, there has to be a better way. That said, I checked it and it works.

app/forms/delivery_partner_form.rb
```diff
@@ -3,20 +3,22 @@
...
 
   def chosen_provider_relationships
-    provider_relationships.map do |relationship|
-      ProviderRelationship.new(
-        cohort: Cohort.find(relationship["cohort_id"]),
-        lead_provider: LeadProvider.find(relationship["lead_provider_id"]),
-      )
-    end
+    provider_relationship_hashes
+      &.map { |value| JSON.parse(value) }
```
seb-siem-softwire, line 19: What does value represent in here?

app/forms/delivery_partner_form.rb
```diff
@@ -64,6 +50,19 @@ def save!
...
+  def update!(delivery_partner)
+    delivery_partner.name = name
+    ActiveRecord::Base.transaction do
+      delivery_partner.save!
+      chosen_provider_relationships.each do |provider_relationship|
+        provider_relationship.delivery_partner = delivery_partner
+        provider_relationship.save!
+      end
+
+      delivery_partner.provider_relationships.where.not(id: chosen_provider_relationships).discard_all!
```
seb-siem-softwire, line 62: Nice - using not syntax

app/views/admin/suppliers/delivery_partners/_lead_providers_cohorts.html.erb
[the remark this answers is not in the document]
kicferk1: Naturally what I'm suggesting higher up would lead to 'hmm, should we use value objects to store our form data, rather than Rails models?', and maybe the answer is 'not yet, we are happy with attribute accessors on the models', or maybe yes, lets have a separate PR to add that pattern here'.

app/forms/delivery_partner_form.rb
```diff
@@ -74,8 +73,8 @@ def lead_providers_and_cohorts_validation
 
     # Ensure all selected lead providers have at least one selected cohort
     # This is indicated by the presence of a provider relationship for that lead provider
-    lead_providers.each do |lead_provider|
-      unless provider_relationships.filter { |provider_relationship| provider_relationship.lead_provider == lead_provider }.any?
+    lead_providers.each do |lead_provider_id|
+      unless provider_relationships.filter { |provider_relationship| provider_relationship.lead_provider_id == lead_provider_id }.any?
```
seb-siem-softwire, line 77: Possibly use pluck i.e getting all provider ids and checking the array if it includes lead_provider id

spec/requests/admin/suppliers/delivery_partners_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+
+    it "updates provider relationships" do
+      old_cohort = create(:cohort)
+      old_lead_provider = create(:lead_provider, cohorts: [old_cohort])
+      ProviderRelationship.create!(delivery_partner: delivery_partner, lead_provider: old_lead_provider, cohort: old_cohort)
+      ProviderRelationship.create!(delivery_partner: delivery_partner, lead_provider: lead_provider, cohort: old_cohort)
+
+      patch "/admin/suppliers/delivery-partners/#{delivery_partner.id}", params: { delivery_partner_form: {
+        name: delivery_partner.name,
+        lead_providers: [lead_provider.id],
```
amazimbe, line 177: ditto

spec/requests/admin/suppliers/delivery_partners_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+  describe "GET /admin/suppliers/delivery-partners/{delivery_partner.id}/edit" do
+    it "renders the edit template" do
+      get "/admin/suppliers/delivery-partners/#{delivery_partner.id}/edit"
+      expect(response).to render_template("admin/suppliers/delivery_partners/edit")
+      expect(response.body).to include(CGI.escapeHTML(delivery_partner.name))
+    end
+  end
+
+  describe "GET /admin/suppliers/delivery-partners/{delivery_partner.id}/delete" do
+    it "render the delete template" do
```
amazimbe, line 133: renders ?

spec/requests/admin/suppliers/delivery_partners_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+
+    it "creates a new delivery partner" do
+      expect {
+        when_I_confirm_my_choices
+      }.to change { DeliveryPartner.count }.by(1)
+    end
+
+    it "creates a delivery partner with the correct name" do
+      # When
+      post "/admin/suppliers/new/delivery-partner", params: {}
```
amazimbe, line 109: I don't quite understand this: we post empty params and it creates a dps with the correct name? Where is that name coming from?

app/controllers/admin/suppliers/delivery_partners_controller.rb
```diff
@@ -55,17 +55,51 @@ def delivery_partner_success
...
 
-      def show_lps; end
+      def edit
+        authorize @delivery_partner
 
-      def show_schools; end
+        @delivery_partner_form = DeliveryPartnerForm.new(
+          name: @delivery_partner.name,
+          lead_providers: @delivery_partner.lead_providers.map(&:id),
+          provider_relationship_hashes: @delivery_partner.provider_relationships.map do |relationship|
```
seb-siem-softwire, line 74: One option would be to put the helper method to provider relationships - as such you would have all variables to create a string

app/forms/delivery_partner_form.rb
```diff
@@ -3,20 +3,22 @@
...
   def chosen_provider_relationships
-    provider_relationships.map do |relationship|
-      ProviderRelationship.new(
-        cohort: Cohort.find(relationship["cohort_id"]),
-        lead_provider: LeadProvider.find(relationship["lead_provider_id"]),
-      )
-    end
+    provider_relationship_hashes
+      &.map { |value| JSON.parse(value) }
+      &.filter { |relationship_params| lead_providers.include?(relationship_params["lead_provider_id"]) }
```
seb-siem-softwire, line 20: lead_providers gives the impression that it is a collection of whole objects - alternatively using lead_provider_ids as a name would make it easier to understand the intention

spec/requests/admin/suppliers/delivery_partners_spec.rb
ethanmills: The session, look at the before block

spec/requests/admin/suppliers/delivery_partners_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+      follow_redirect!
+      expect(response.body).to include(CGI.escapeHTML(delivery_partner.name))
+    end
+
+    it "update the name of the delivery partner" do
+      old_name = delivery_partner.name
+      new_name = Faker::Name.name
+      patch "/admin/suppliers/delivery-partners/#{delivery_partner.id}", params: { delivery_partner_form: {
+        name: new_name,
+        lead_providers: [lead_provider.id],
```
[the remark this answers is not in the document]
ethanmills: But you change both on the same page. Would that not delete all the provider relationships?

spec/models/provider_relationship_spec.rb
```diff
@@ -3,6 +3,13 @@
 require "rails_helper"
 
 RSpec.describe ProviderRelationship, type: :model do
+  let(:lead_provider) { create(:lead_provider) }
+  let(:delivery_partner) { create(:delivery_partner) }
+  let(:cohort) { create(:cohort) }
+  let!(:provider_relationship) do
```
seb-siem-softwire, line 9: Out of curiosity - what is your general rule of thumb between useing before each block and let!

config/routes.rb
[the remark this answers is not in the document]
ethanmills: Interesting read, thanks. I'll bear it in mind

app/views/admin/suppliers/delivery_partners/_lead_providers_cohorts.html.erb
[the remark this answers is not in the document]
ethanmills: I think that makes sense, I'm generally in favour of moving logic out of controllers and especially views. I don't think I'll have time to do that here, but feel free to open a ticket for it (or a PR 😛 )

app/forms/delivery_partner_form.rb
ethanmills: I read that SO post as more "no one is really sure" 😂 Besides, it's not just the hash itself I'm worried about, it's the hash => json.  If you're both convinced it will never break when the serializer implementation changes, then that's good enough for me and I'll change to hash.to_json

app/views/admin/suppliers/delivery_partners/_lead_providers_cohorts.html.erb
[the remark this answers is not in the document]
ethanmills: Anything you want changed for this in this PR @kicferk1 ?

spec/models/provider_relationship_spec.rb
amazimbe: I avoid let!  as much as possible these days but still use it. The problem is when you do it in a top scope and it ends up creating other objects in other specs where those objects are not required. In those cases it's probably better to use before.

app/views/admin/suppliers/delivery_partners/_lead_providers_cohorts.html.erb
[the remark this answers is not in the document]
kicferk1: A `todo` comment with link to https://trello.com/c/ursP6wtZ/328-refactor-how-we-present-partnerships-when-creating-updating-them would be nice. Either here or in the form itself.

app/forms/delivery_partner_form.rb
ethanmills: Good shout

app/controllers/admin/suppliers/delivery_partners_controller.rb
ethanmills: Not when we build it in the view

spec/models/provider_relationship_spec.rb
ethanmills: No particular reason for it, but generally when I use the named thing that was created I use `let!`, otherwise a before to just set up state for the test

app/forms/delivery_partner_form.rb
amazimbe: Sure, let's change to to_json. When it breaks in future we'll fix it and perhaps we'll have a better fix then.

amazimbe (approved):

spec/cypress/integration/admin/suppliers/create-lead-provider.js
```diff
@@ -0,0 +1,137 @@
...
+    cy.chooseLeadProviderType();
+    cy.chooseFirstCIPForLeadProvider();
+    cy.chooseFirstCohortForLeadProvider();
+
+    cy.get("a").contains("Change name").click();
+    cy.chooseSupplierName(`{selectall}${leadProviderName}`);
+
+    cy.clickCommitButton();
+    cy.clickCommitButton();
+    cy.clickCommitButton();
```
callumacrae, line 125: sorry, didn't notice this the first time - is this supposed to click three different buttons on different pages? If so, I'd worry that double clicks leading to a failing test if something is slow - maybe testing the URL between clicks could make the test less likely to break

(this is minor though - if it's not already breaking I don't really see why it would start)

ethanmills: Yes, that's what it does. Good to know, will avoid in future, and fix this if it breaks

seb-siem-softwire (approved):

PR #119, opened 2021-03-08
Delete users functionality

seb-siem-softwire:
### Context

This pull request adds Delete Users functionality

Using Discard gem we can soft delete the users

### Guidance to review

Japanese, English or Polish comments encouraged 

### Testing

WIP


app/models/admin_profile.rb
```diff
@@ -2,4 +2,13 @@
 
 class AdminProfile < ApplicationRecord
   belongs_to :user
+
+  include Discard::Model
+  default_scope -> { kept }
+
+  scope :kept, -> { undiscarded.joins(:user).merge(User.kept) }
```
kicferk1, line 9: Interesting. How do default scopes work with one-to-one / many-to-one relations in Rails? 

If we have a user, who has an admin profile, and then we discard the admin profile, will the user check for it being admin work? 

And I guess, if we have a school which has many teacher profiles, and we discard some of them, will the school only get the non-discarded when we ask for the list of its teachers?

(and I hope it propagates through `through` keyword?) 

Just something to test and think about, maybe even to write a unit test about. Pretty sure you are thinking about it already :)

app/models/early_career_teacher_profile.rb
```diff
@@ -5,4 +5,13 @@ class EarlyCareerTeacherProfile < ApplicationRecord
...
   belongs_to :cohort, optional: true
+
+  include Discard::Model
+  default_scope -> { kept }
+
+  scope :kept, -> { undiscarded.joins(:user).merge(User.kept) }
+
+  def kept?
+    undiscarded? && user.kept?
+  end
```
ethanmills, line 16: Can we move this onto a base profile?

seb-siem-softwire: Good point - I'll move it

app/models/admin_profile.rb
seb-siem-softwire: 1. Q: If we have a user, who has an admin profile, and then we discard the admin profile, will the user check for it being admin work?
2. A: if we would discard User profile - it would hide/soft delete the corresponding profile. Should we consider the opposite scenario i.e if profile is soft deleted - the corresponding user should be hidden as well - Good question - what do you think should happen @kicferk1 
3. Q And I guess, if we have a school which has many teacher profiles, and we discard some of them, will the school only get the non-discarded when we ask for the list of its teachers?
4. A: Given that the default scope is for kept records - we would see only the non discarded ones - and having a test will prove it
5. Q: (and I hope it propagates through through keyword?) - 
6. A: The unit test will show us the answer 

Good points - if you think of any other scenarios - let me know

ethanmills:
[Allow DFE admins to delete a user](https://trello.com/c/NV09coTN/308-allow-dfe-admins-to-delete-a-user)

ethanmills:
Either way, we should be very sure that we want to introduce new controllers and views, rather using existing ones:
https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/controllers/admin/administrators/administrators_controller.rb
https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/controllers/admin/induction_coordinators_controller.rb
https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/controllers/admin/suppliers/supplier_users_controller.rb
https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/controllers/admin/lead_providers/lead_provider_users_controller.rb

(The last two should probably be combined anyway)

cfq:
Thanks for these Seb. I can't find a way of accessing your work though. Is it linked from anywhere? I've been putting in the URL to access it: https://ecf-review-pr-119.london.cloudapps.digital/admin/profiles/admin_profiles

Also delete didn't work for me. Is it a soft-delete issue?

seb-siem-softwire:
Closing due to issues with deployments -> https://github.com/DFE-Digital/early-careers-framework/pull/126

PR #120, opened 2021-03-08
Removes default value for cohorts in partnership

seb-siem-softwire:
### Context

Removes default value for cohorts in partnership

Removes default cohort id in partnerships table that tends to change

### Guidance to review

Be or not to be a nil Cohort - that is the question

### Testing

Tests pass


kicferk1 (approved):
If you create a db from scratch and seed it, does it work? 

Just a sanity check really, I imagine we will know he answer through review app being created :)

amazimbe (approved):

ethanmills (approved):

seb-siem-softwire:
All good @kicferk1 it does work

PR #122, opened 2021-03-09
[ECF-1-307] Index, edit and update induction coordinators

amazimbe:
## Context

Added index, edit and update actions for induction coordinators

### DIFF
--- a/app/controllers/admin/induction_coordinators_controller.rb
+++ b/app/controllers/admin/induction_coordinators_controller.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module Admin
+  class InductionCoordinatorsController < Admin::BaseController
+    skip_after_action :verify_authorized, only: :index
+    skip_after_action :verify_policy_scoped, except: :index
+    before_action :load_induction_coordinator, only: %i[edit update]
+
+    def index
+      @induction_coordinators = policy_scope(User).induction_coordinators
+    end
+
+    def edit
+      authorize User
+    end
+
+    def update
+      authorize User
+
+      if @induction_coordinator.update(permitted_attributes(@induction_coordinator))
+        redirect_to :admin_induction_coordinators, notice: "Changes saved successfully"
+      else
+        render :edit
+      end
+    end
+
+  private
+
+    def load_induction_coordinator
+      @induction_coordinator = User.find(params[:id])
+    end
+  end
+end
--- a/app/controllers/admin/induction_coordinators_controller.rb
+++ b/app/controllers/admin/induction_coordinators_controller.rb
@@ -11,11 +11,11 @@ module Admin
     end
 
     def edit
-      authorize User
+      authorize @induction_coordinator
     end
 
     def update
-      authorize User
+      authorize @induction_coordinator
 
       if @induction_coordinator.update(permitted_attributes(@induction_coordinator))
         redirect_to :admin_induction_coordinators, notice: "Changes saved successfully"
--- a/app/views/admin/induction_coordinators/index.html.erb
+++ b/app/views/admin/induction_coordinators/index.html.erb
@@ -1,0 +1,21 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">Induction coordinators</h1>
+    <table class="govuk-table">
+      <thead>
+        <tr class="govuk-table__row">
+          <th scope="col" class="govuk-table__header">Name</th>
+          <th scope="col" class="govuk-table__header">Email address</th>
+        </tr>
+      </thead>
+      <tbody class="govuk-table__body">
+        <% @induction_coordinators.each do |induction_coordinator| %>
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell"><%= govuk_link_to induction_coordinator.full_name, admin_edit_user_path(induction_coordinator) %></td>
+            <td class="govuk-table__cell"><%= induction_coordinator.email %></td>
+          </tr>
+        <% end %>
+      </tbody>
+    </table>
+  <div>
+</div>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -106,7 +106,7 @@ Rails.application.routes.draw do
     scope module: "lead_providers", path: "lead-providers" do
       resources :lead_provider_users, only: %i[edit update], path: "users"
     end
-    resources :induction_coordinators, only: %i[edit update], path: "induction-coordinators"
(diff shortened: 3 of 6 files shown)


amazimbe:
[Allow DFE admins (super user) to edit a user.](https://trello.com/c/b0GGlRLe/307-allow-dfe-admins-super-user-to-edit-a-user)

app/views/admin/induction_coordinators/index.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      <thead>
+        <tr class="govuk-table__row">
+          <th scope="col" class="govuk-table__header">Name</th>
+          <th scope="col" class="govuk-table__header">Email address</th>
+        </tr>
+      </thead>
+      <tbody class="govuk-table__body">
+        <% @induction_coordinators.each do |induction_coordinator| %>
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell"><%= govuk_link_to induction_coordinator.full_name, admin_edit_user_path(induction_coordinator) %></td>
```
ethanmills, line 14: I'm surprised it's just `admin_edit_user_path`

spec/requests/admin/induction_coordinators_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    it "renders the index template" do
+      get "/admin/induction-coordinators"
+      expect(response).to render_template("admin/induction_coordinators/index")
+      expect(response.body).to include(CGI.escapeHTML(induction_coordinator.email))
+      expect(response.body).to include(CGI.escapeHTML(induction_coordinator.full_name))
+    end
+  end
+
+  describe "GET /admin/induction-coordinators/:id/edit" do
+    it "renders the index template" do
```
ethanmills, line 26: edit template

spec/requests/admin/induction_coordinators_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      get "/admin/induction-coordinators/#{induction_coordinator.id}/edit"
+
+      expect(response.body).to include("Edit user details")
+      expect(response).to render_template("admin/induction_coordinators/edit")
+    end
+  end
+
+  describe "PATCH /admin/induction-coordinatorsrs/:id" do
```
ethanmills, line 34: typo, rsrs

app/controllers/admin/induction_coordinators_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    skip_after_action :verify_authorized, only: :index
+    skip_after_action :verify_policy_scoped, except: :index
+    before_action :load_induction_coordinator, only: %i[edit update]
+
+    def index
+      @induction_coordinators = policy_scope(User).induction_coordinators
+    end
+
+    def edit
+      authorize User
```
kicferk1, line 14: Should it be `authorize @ induction_coordinator`? Here and in update?

config/routes.rb
```diff
@@ -106,7 +106,7 @@
     scope module: "lead_providers", path: "lead-providers" do
       resources :lead_provider_users, only: %i[edit update], path: "users"
     end
-    resources :induction_coordinators, only: %i[edit update], path: "induction-coordinators"
+    resources :induction_coordinators, only: %i[index edit update], path: "induction-coordinators"
```
kicferk1, line 109: Huh, did we have dead routes before this PR? Oh well, they aren't dead anymore.

kicferk1 (commented):
I get a sense of... repetition.

Are we sure we need separate pages to edit different types of users? The forms we have, the index pages, the controllers, the tests, all look very similar between lead providers and induction coordinators.

app/views/admin/induction_coordinators/index.html.erb
kicferk1: I think it's a helper method added in an earlier PR?

kicferk1: Although, if we go for solution without much repetition (one controller to rule and edit all users), it will be this short without the need for a helper method.

amazimbe: I thought really hard about the repetition and had a quick chat with Seb too. Our conclusion was that we not worry about it too much, these requirements are too uncertain for us to worry about optimisation. My gut feeling is that these endpoints will diverge more as we gather more requirements and it's often easier to go from separate endpoints to one shared endpoint than the other way round.

config/routes.rb
amazimbe: Yes, I have to add them in the earlier pr to test the helper

app/controllers/admin/induction_coordinators_controller.rb
amazimbe: rebase went wrong, will fix

ethanmills (approved):
Couple of typos, but otherwise looks good!

ethanmills:
> I get a sense of... repetition.
> 
> Are we sure we need separate pages to edit different types of users? The forms we have, the index pages, the controllers, the tests, all look very similar between lead providers and induction coordinators.

In future I'd like to surface more information about the user's organisation on these pages. Lead provider name, school name & URN etc. I imagine a support function will approach this thinking about an organisation a user belongs to, rather than an individual. If not, we can commonise in future

seb-siem-softwire:
Just some typos in the specs - all else looks good

seb-siem-softwire (approved):

app/views/admin/induction_coordinators/index.html.erb
kicferk1: Alright. I think it tends to be easier to go from one endpoint to many - cause you can start from copy-pasting the pages, controllers, and routes, and then delete things you don't want in each. 

But I think I'm convinced that there will be enough divergence between these pages that we should set up the skeleton separate flows for them.

app/controllers/admin/induction_coordinators_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+      else
+        render :edit
+      end
+    end
+
+  private
+
+    def load_induction_coordinator
+      @induction_coordinator = User.find(params[:id])
+    end
```
kicferk1, line 31: I guess we could authorise here if we cared about one line of repetition - I'll let you decide if we do :)

kicferk1 (approved):

app/controllers/admin/induction_coordinators_controller.rb
amazimbe: I didn't want to make the edit action codeless and then have to do def edit; end which I find very ugly 😢 . I may do that in my next pr for admins editing other admins

PR #123, opened 2021-03-09
[ECF-1-307] Admins edit admins

amazimbe:
### Context
Allow admins to edit admins

### DIFF
--- a/app/views/admin/administrators/administrators/index.html.erb
+++ b/app/views/admin/administrators/administrators/index.html.erb
@@ -12,7 +12,7 @@
       <tbody class="govuk-table__body">
         <% @administrators.each do |administrator| %>
           <tr class="govuk-table__row">
-            <td class="govuk-table__cell"><%= administrator.full_name %></td>
+            <td class="govuk-table__cell"><%= govuk_link_to administrator.full_name, edit_admin_administrator_path(administrator) %></td>
             <td class="govuk-table__cell"><%= administrator.email %></td>
           </tr>
         <% end %>
--- a/app/views/layouts/admin/base.html.erb
+++ b/app/views/layouts/admin/base.html.erb
@@ -6,6 +6,9 @@
     <%= component.nav_item(path: "/admin/administrators") do %>
       Administrators
     <% end %>
+    <%= component.nav_item(path: "/admin/induction-coordinators") do %>
+      Induction Coordinators
+    <% end %>
   <% end %>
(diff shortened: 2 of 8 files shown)


app/views/layouts/admin/base.html.erb
```diff
@@ -6,6 +6,9 @@
     <%= component.nav_item(path: "/admin/administrators") do %>
       Administrators
     <% end %>
+    <%= component.nav_item(path: "/admin/induction-coordinators") do %>
+      Induction Coordinators
+    <% end %>
```
kicferk1, line 11: Good spot

kicferk1 (approved):

app/views/admin/administrators/administrators/index.html.erb
```diff
@@ -12,7 +12,7 @@
       <tbody class="govuk-table__body">
         <% @administrators.each do |administrator| %>
           <tr class="govuk-table__row">
-            <td class="govuk-table__cell"><%= administrator.full_name %></td>
+            <td class="govuk-table__cell"><%= govuk_link_to administrator.full_name, edit_admin_administrator_path(administrator) %></td>
```
ethanmills, line 15: Are we not using the helper here? Why only in some places?

ethanmills (approved):
Looks good, just one question

app/views/admin/administrators/administrators/index.html.erb
amazimbe: I don't think it's strictly needed. My main motivation for the helper was on the suppliers page where we have a list that mixes up delivery partner users and lead provider users and I needed to send each user to the right endpoint. The admins and induction coordinators don't get combined that why I haven't used it here. @ethanmills

ethanmills: Fair enough. You do use it for induction coordinators though, right? https://github.com/DFE-Digital/early-careers-framework/pull/122/files#diff-9e81e35702dad35ed3b19ae78f92ad76e63185d92a3158b0818ba73110b36927R14

ethanmills: Ok 🙂 Happy for this to go in as-is

app/views/layouts/admin/base.html.erb
seb-siem-softwire: Nice

PR #124, opened 2021-03-10
ECF-RP-320: Eligibility of schools

ethanmills:
### Context
Not all schools we get from GIAS can be part of ECF. We only want to think about those which can. Brandon gave us a set of criteria (on ticket)

- Add required information to school model
- Make some fields integers, for ease of comparison
- Implement eligibility, and make eligible the default scope on schools

### Testing
- updated unit tests for import
- added tests for scopes and predicate function


kicferk1 (approved):
A part of me wonders if it would be better to not have any ineligible schools in the db, but then again, eligibility rules change, and we don't want to axe a bunch of schools when they do, so this is probably a better solution.

ethanmills:
> A part of me wonders if it would be better to not have any ineligible schools in the db, but then again, eligibility rules change, and we don't want to axe a bunch of schools when they do, so this is probably a better solution.

Yeah I wasn't 100% on this, but I wanted to keep business logic out of the GIAS import. We're going to need to keep track of schools as they close anyway, so we're going to need something like this

PR #127, opened 2021-03-11
Fix cookie banner

callumacrae:
### Context

This was a bit of a nightmare to test due to differences between local and the dev environment, but I think I've got it!

I've also changed the tests so that they're more likely to catch failures—although it still doesn't catch the original problem.

### Guidance to review

This bug didn't happen on a PR environment last time either, so it's probably a code review only on this one…

### DIFF
--- a/app/webpacker/packs/cookie-banner.js
+++ b/app/webpacker/packs/cookie-banner.js
@@ -8,16 +8,23 @@ if (cookieBannerEl) {
       return;
     }
 
+    const body = new FormData(cookieFormEl);
+    body.append("cookies_form[analytics_consent]", e.target.value);
+
     fetch(cookieFormEl.action, {
       method: "PUT",
       headers: {
         Accept: "application/json",
-        "Content-Type": "application/json",
       },
-      body: JSON.stringify({
-        cookies_form: { analytics_consent: e.target.value },
-      }),
+      body,
     })
+      .then((res) => {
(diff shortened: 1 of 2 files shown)


kicferk1 (commented):
So does the no-js version work on dev? Maybe reloading a page is not the end of the world?

ethanmills (approved):
How does this work for whoever you stole this from? Did they disable forgery protection for this?

callumacrae:
> So does the no-js version work on dev? Maybe reloading a page is not the end of the world?

The no-js version worked all along, and now it's using all the form data not manually getting the bits I knew about, so this won't happen again, even if something changes in the way rails handles auth tokens in forms.

seb-siem-softwire (approved):

kicferk1 (approved):
So the long-term solution to prevent other devs falling into the same pit is "whenever you see a form submitted using manual javascript, make sure it sends all of form data"?

I guess we are re-doing a bunch of work that rails-ujs does (we cut it out because it didn't work well with rendering errors server-side). Maybe if it happens again we want to look into making a general solution work.

callumacrae:
The long-term solution is to not develop your stuff in the test environment, because it disables the csrf protection 😆 

Yeah, it's pretty much just rails-ujs' `remote` behaviour. I don't see that it's something that will end up happening much in this project, but if it does an abtraction would be worth looking into.

PR #128, opened 2021-03-11
Visual improvements

callumacrae:
Draft PR but comments welcome - if something isn't an improvement feel free to let me know before review stage.

This work isn't particularly structured, I'm just going through and changing stuff as I find it.

### DIFF
--- a/app/views/registrations/school_profile/show.html.erb
+++ b/app/views/registrations/school_profile/show.html.erb
@@ -1,15 +1,19 @@
 <% content_for :before_content, govuk_back_link(text: "Back", href: registrations_question_two_path) %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-l">
-      <label for="school-profile-form-urn-field">What is your school Unique Reference Number (URN)?</label>
-    </h2>
     <%= form_with(model: @school_profile_form, url: registrations_school_profile_path) do |f| %>
       <%= f.govuk_error_summary %>
-      <p class="govuk-hint">
-        This is a 6 digit number. Not sure? <a target='_blank' href='https://www.get-information-schools.service.gov.uk/'>Search for your school</a> on our register.
-      </p>
-      <%= f.govuk_text_field(:urn, width: 20, label:  nil, link_errors: true) %>
+      <%= f.govuk_text_field(
+            :urn,
+            width: 20,
+            label: -> { tag.h1("What is your school Unique Reference Number (URN)?", class: "govuk-heading-l") },
+            hint: -> { tag.p do
+              concat("This is a 6 digit number. Not sure? ")
+              concat(govuk_link_to("Search for your school", "https://www.get-information-schools.service.gov.uk/", target: "_blank"))
+              concat(" on our register.")
+            end
+            },
+            link_errors: true) %>
       <%= f.govuk_submit "Continue" %>
     <% end %>
   </div>
--- a/app/views/registrations/school_profile/show.html.erb
+++ b/app/views/registrations/school_profile/show.html.erb
@@ -9,7 +9,7 @@
             label: { text: "What is your school Unique Reference Number (URN)?", tag: "h1", size: "l" },
             hint: -> { tag.p do
               concat("This is a 6 digit number. Not sure? ")
-              concat(govuk_link_to("Search for your school", "https://www.get-information-schools.service.gov.uk/", target: "_blank"))
+              concat(govuk_link_to("Search for your school", "https://www.get-information-schools.service.gov.uk/", target: "_blank", rel: "noopener noreferrer"))
               concat(" on our register.")
             end
             },
--- a/app/views/registrations/school_profile/show.html.erb
+++ b/app/views/registrations/school_profile/show.html.erb
@@ -1,13 +1,15 @@
 <% content_for :before_content, govuk_back_link(text: "Back", href: registrations_question_two_path) %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h2 class="govuk-heading-l">What is your school Unique Reference Number(URN)?</h2>
+    <h2 class="govuk-heading-l">
+      <label for="school-profile-form-urn-field">What is your school Unique Reference Number (URN)?</label>
+    </h2>
     <%= form_with(model: @school_profile_form, url: registrations_school_profile_path) do |f| %>
       <%= f.govuk_error_summary %>
-      <label class="govuk-hint">
+      <p class="govuk-hint">
         This is a 6 digit number. Not sure? <a target='_blank' href='https://www.get-information-schools.service.gov.uk/'>Search for your school</a> on our register.
-        <%= f.govuk_text_field(:urn, width: 20, label:  nil, link_errors: true) %>
-      </label>
+      </p>
+      <%= f.govuk_text_field(:urn, width: 20, label:  nil, link_errors: true) %>
       <%= f.govuk_submit "Continue" %>
     <% end %>
   </div>
--- a/app/views/registrations/school_profile/show.html.erb
+++ b/app/views/registrations/school_profile/show.html.erb
@@ -6,7 +6,7 @@
       <%= f.govuk_text_field(
             :urn,
             width: 20,
-            label: -> { tag.h1("What is your school Unique Reference Number (URN)?", class: "govuk-heading-l") },
+            label: { text: "What is your school Unique Reference Number (URN)?", tag: "h1", size: "l" },
             hint: -> { tag.p do
               concat("This is a 6 digit number. Not sure? ")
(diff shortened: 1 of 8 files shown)


app/webpacker/styles/primary-nav.scss
```diff
@@ -1,20 +1,29 @@
...
+    margin-right: govuk-spacing(8);
+  }
 
   &.nav-item__selected {
     border-bottom: solid govuk-colour("blue") 4px;
   }
 
   a {
     text-decoration: none;
+    line-height: $primaryNavHeight;
```
callumacrae, line 27: Tip for the future @ethanmills - this is a neat trick to vertically align text in navigation and avoid the `padding-top: 0.25rem` that was there before (which breaks when the text size changes)

ethanmills: Thanks!

app/views/registrations/school_profile/show.html.erb
```diff
@@ -1,13 +1,15 @@
...
+    </h2>
     <%= form_with(model: @school_profile_form, url: registrations_school_profile_path) do |f| %>
       <%= f.govuk_error_summary %>
-      <label class="govuk-hint">
+      <p class="govuk-hint">
         This is a 6 digit number. Not sure? <a target='_blank' href='https://www.get-information-schools.service.gov.uk/'>Search for your school</a> on our register.
-        <%= f.govuk_text_field(:urn, width: 20, label:  nil, link_errors: true) %>
-      </label>
+      </p>
+      <%= f.govuk_text_field(:urn, width: 20, label:  nil, link_errors: true) %>
```
ethanmills, line 12: I personally prefer this method: https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/views/admin/suppliers/suppliers/new.html.erb#L5

ethanmills: This would become 
```
<%= f.govuk_text_field(
            :urn,
            width: 20,
            label: -> { tag.h2("Supplier name", class: "govuk-heading-l") },
            hint: -> { tag.p do
              concat("This is a 6 digit number. Not sure? ")
              concat(govuk_link_to("Search for your school", "https://www.get-information-schools.service.gov.uk/"))
              concat(" on our register.")
            end
            },
            link_errors: true) %>
```

callumacrae: I tried this first but was under the impression from the docs that `hint` has to be plain text only - I'll give it another go.

callumacrae: Yep just tested, it's text only.

ethanmills: Works for me 
image

app/webpacker/styles/header_overrides.scss
```diff
@@ -0,0 +1,9 @@
+.govuk-header {
+  border-bottom: 10px $govuk-brand-colour solid;
+
+  &__container {
+    // Doesn't seem to be any need to do this but leaving it in case there's
+    // any strange bugs in other browsers in the future
+    /* border-bottom: 0; */
+  }
```
kicferk1, line 8: Uhh, if there is no need for it, I'd much rather not have a comment saying that.

app/views/registrations/school_profile/show.html.erb
```diff
@@ -1,13 +1,19 @@
...
-    <h2 class="govuk-heading-l">What is your school Unique Reference Number(URN)?</h2>
     <%= form_with(model: @school_profile_form, url: registrations_school_profile_path) do |f| %>
       <%= f.govuk_error_summary %>
-      <label class="govuk-hint">
-        This is a 6 digit number. Not sure? <a target='_blank' href='https://www.get-information-schools.service.gov.uk/'>Search for your school</a> on our register.
-        <%= f.govuk_text_field(:urn, width: 20, label:  nil, link_errors: true) %>
-      </label>
+      <%= f.govuk_text_field(
+            :urn,
+            width: 20,
```
kicferk1, line 8: I'd rather use govuk rows and columns classes to get the width right - unless 20 (characters? percent?) is explicitly what we want.

app/views/registrations/school_profile/show.html.erb
```diff
@@ -1,13 +1,19 @@
...
+            label: -> { tag.h1("What is your school Unique Reference Number (URN)?", class: "govuk-heading-l") },
+            hint: -> { tag.p do
+              concat("This is a 6 digit number. Not sure? ")
+              concat(govuk_link_to("Search for your school", "https://www.get-information-schools.service.gov.uk/", target: "_blank"))
```
kicferk1, line 12: Do we want some no-referrer and no-opener here?

In general, I wonder if there is a better way of specifying custom label, this is getting a bit annoying with all the concats. Not that I have any specific ideas :/ Maybe `render` method will work here? And we can have a partial for the label?

app/views/registrations/school_profile/show.html.erb
```diff
@@ -1,13 +1,19 @@
...
     <%= form_with(model: @school_profile_form, url: registrations_school_profile_path) do |f| %>
       <%= f.govuk_error_summary %>
-      <label class="govuk-hint">
-        This is a 6 digit number. Not sure? <a target='_blank' href='https://www.get-information-schools.service.gov.uk/'>Search for your school</a> on our register.
-        <%= f.govuk_text_field(:urn, width: 20, label:  nil, link_errors: true) %>
-      </label>
+      <%= f.govuk_text_field(
+            :urn,
+            width: 20,
+            label: -> { tag.h1("What is your school Unique Reference Number (URN)?", class: "govuk-heading-l") },
```
kicferk1, line 9: Did you mention on another PR that we can specify tag separately?

amazimbe: Yeh, do as I say not as I do 😄

app/webpacker/styles/header_overrides.scss
callumacrae: There are two borders overlaid over each other right now - that's strange and unexpected behaviour, this serves as documentation for that. A future update of govuk-frontend could break it, or there could be issues in browsers we haven't tested yet.

Equally, removing the other border could break in future updates of govuk-frontend too.

I think they're both pretty unlikely to happen, but think it's important to write down somewhere what's happening.

app/views/registrations/school_profile/show.html.erb
callumacrae: This adds the `govuk-input--width-20` to the outputted markup and seems to be the recommended way of doing it: https://govuk-form-builder.netlify.app/form-elements/text-field/#generating-text-inputs-with-custom-widths

app/views/registrations/school_profile/show.html.erb
callumacrae: > Do we want some no-referrer and no-opener here?

🤷‍♂️ this is from the old code. Happy to change it if that's required

> In general, I wonder if there is a better way of specifying custom label, this is getting a bit annoying with all the concats. Not that I have any specific ideas :/ Maybe render method will work here? And we can have a partial for the label?

I don't think I know enough about rails to really have an opinion on this, but having a partial just for this seems like overkill to me

app/views/registrations/school_profile/show.html.erb
kicferk1: Right, and is urn 20 characters long?

My guess is that 20 is what looks like designs, rather than specifies expected length exactly - and if we want exact length, umm, maybe it should be like 6? And if we don't want the exact length, then maybe let's not pretend that we do.

callumacrae: Right, got you!

There's no 6 character width, just 5 and 10. Shall I change it to 10?

kicferk1: I guess. Check with Al or Shani?

app/views/registrations/school_profile/show.html.erb
kicferk1: I think the no-referrer and no-opener aren't really needed, unless we expect `https://www.get-information-schools.service.gov.uk/` to try some naughty things on us - but at the same time, I don't see why we would like to allow them to interact with our page via javascript. 

Ech, I dislike programatically cobbling together html - might be just me. 3 lines aren't the end of the world though.

app/views/registrations/school_profile/show.html.erb
ethanmills: I think if we're doing it for design reasons it should just be fractional, right?

kicferk1: Question is, did our designers think about the width here, or did they added a field that kind of look good - I would definitely double check with them.

callumacrae: If it involves roping someone else in—especially as they sound like they're super busy—I'd prefer to leave it for another ticket as it isn't actually related to this PR so much—the width hasn't changed in this PR.

Are you okay with that?

kicferk1: Sure

kicferk1 (approved):

PR #129, opened 2021-03-11
ECF-RP-322: Add success flash

ethanmills:
### Context
Set up for flash success notifications. Eventually I want to go through all of our `flash` and use either a green or blue notification banner, but Al doesn't have time for this right now.

I've not used this anywhere in this PR, just set up work for the ones to follow.

[Here](https://design-system.service.gov.uk/components/notification-banner/#reacting-to-something-the-user-has-done:~:text=Use%20the%20green%20version%20of%20the,they%E2%80%99re%20expecting%20to%20happen%20has%20happened.) is the design system component.
image

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,6 +14,10 @@ class ApplicationController < ActionController::Base
     stored_location_for(user) || helpers.profile_dashboard_url(user)
   end
 
+  def set_success_message(title: "Success", heading: "", content: "")
+    flash[:success] = { title: title, heading: heading, content: content }
+  end
+
 protected
 
   def configure_permitted_parameters
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -80,21 +80,26 @@
 
       <main class="govuk-main-wrapper govuk-body" id="main-content" role="main">
         <% flash.each do |type, msg| %>
-        <% if type == "alert" %>
-          <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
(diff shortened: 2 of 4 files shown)


callumacrae (approved):

app/controllers/application_controller.rb
```diff
@@ -14,6 +14,10 @@ def after_sign_in_path_for(user)
     stored_location_for(user) || helpers.profile_dashboard_url(user)
   end
 
+  def set_success_message(title: "Success", heading: "", content: "")
```
amazimbe, line 17: Where does this get called then?

amazimbe (approved):
LGTM! Just one question to aid my own understanding.

app/controllers/application_controller.rb
kicferk1: I imagine in any controller inheriting form application controller?

app/views/layouts/application.html.erb
```diff
@@ -80,21 +80,26 @@
...
+          <% if type == "alert" %>
+            <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
+              <h2 class="govuk-error-summary__title" id="error-summary-title">
                 <%= msg %>
               </h2>
             </div>
-          </div>
-          <%end%>
+          <% elsif type == "success" %>
+            <%= govuk_notification_banner title: msg["title"], success: true do |banner| %>
```
kicferk1, line 90: Does `govuk_notification_banner` come from govuk-components?

kicferk1 (approved):

app/views/layouts/application.html.erb
ethanmills: Yes

PR #134, opened 2021-03-12
ECF-1-316 Admins view and reject partnerships

amazimbe:
Screenshot 2021-03-12 at 12 33 34
Screenshot 2021-03-12 at 11 38 25

### Context

Added a show and reject action for DfE admins

### DIFF
--- a/app/controllers/admin/partnerships_controller.rb
+++ b/app/controllers/admin/partnerships_controller.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module Admin
+  class PartnershipsController < Admin::BaseController
+    skip_after_action :verify_policy_scoped
+    before_action :load_partnership, only: %i[show reject]
+
+    def show
+      authorize @partnership
+      @reasons_for_rejection = [
+        OpenStruct.new(id: :partnered_with_another_provider, name: "I have already partnered with a different training provider"),
+        OpenStruct.new(id: :provider_not_known, name: "I don't recognise this training provider"),
+        OpenStruct.new(id: :change_of_mind, name: "I have changed my mind"),
+        OpenStruct.new(id: :no_induction, name: "I'm not doing any inductions this year"),
+        OpenStruct.new(id: :mistake, name: "This looks like a mistake"),
+        OpenStruct.new(id: :other, name: "Other"),
+      ]
+    end
+
+    def reject
+      authorize @partnership, :update?
+      @partnership.update!(reject_params)
+      redirect_to admin_partnership_path(@partnership), notice: "Partnership successfully rejected"
+    end
+
+  private
+
+    def load_partnership
+      @partnership = Partnership.find(params[:id])
+    end
+
+    def reject_params
+      params
+      .require(:partnership)
+      .permit(:reason_for_rejection)
+      .merge(status: "rejected", rejected_at: Time.zone.now)
+    end
+  end
+end
--- a/app/controllers/admin/partnerships_controller.rb
+++ b/app/controllers/admin/partnerships_controller.rb
@@ -19,8 +19,11 @@ module Admin
 
     def reject
       authorize @partnership, :update?
-      @partnership.update!(reject_params)
-      redirect_to admin_partnership_path(@partnership), notice: "Partnership successfully rejected"
+      if @partnership.update(reject_params)
+        redirect_to admin_partnership_path(@partnership), notice: "Partnership successfully rejected"
+      else
+        render :show
+      end
     end
 
   private
--- a/app/models/partnership.rb
+++ b/app/models/partnership.rb
@@ -1,6 +1,21 @@
 # frozen_string_literal: true
 
 class Partnership < ApplicationRecord
+  enum status: {
+    pending: "pending",
+    accepted: "accepted",
+    rejected: "rejected",
+  }
+
+  enum reason_for_rejection: {
+    partnered_with_another_provider: "I have already partnered with a different training provider",
+    provider_not_known: "I don't recognise this training provider",
+    change_of_mind: "I have changed my mind",
+    no_induction: "I'm not doing any inductions this year",
+    mistake: "This looks like a mistake",
+    other: "Other",
+  }, _prefix: :rejected_with
+
   belongs_to :school
   belongs_to :lead_provider
   belongs_to :cohort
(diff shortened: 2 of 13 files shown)


app/views/admin/partnerships/show.html.erb
```diff
@@ -0,0 +1,95 @@
...
+  </div>
+  <% unless @partnership.rejected? %>
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l govuk-!-margin-bottom-10">
+      Reject partnership
+    </h1>
+    <%= form_for @partnership, url: reject_admin_partnership_path, method: :patch do |f| %>
+      <%= f.govuk_collection_radio_buttons(
+            :reason_for_rejection, @reasons_for_rejection, :id, :name,
+            legend: -> { tag.h1("Tell us why" , class: "govuk-heading-m") }
```
callumacrae, line 89: this can be simplified:

```
legend: { text: 'Tell us why', tag: 'h1', size: 'm' }
```

Also, there are multiple h1s on this page

callumacrae:
cypress tests? 🙏

ethanmills:
These options were designed for schools, not admins. Are we sure we want them here?

ethanmills:
We need to distinguish between partnerships that were challenged by a school, and those that were removed for another reason. I think this work is good for admins updating the system to reflect support actions on behalf of schools, but that leaves other things missing

app/models/partnership.rb
```diff
@@ -1,6 +1,21 @@
 # frozen_string_literal: true
 
 class Partnership < ApplicationRecord
+  enum status: {
+    pending: "pending",
+    accepted: "accepted",
+    rejected: "rejected",
+  }
```
ethanmills, line 8: Don't think we have a "pending"

cfq:
@amazimbe can you check with Al and Shani that you're doing the right thing?

app/models/partnership.rb
amazimbe: So what would be the status before the school has confirmed the partnership then, nil?

amazimbe:
> These options were designed for schools, not admins. Are we sure we want them here?

We definitely wan't the options to be the same but the copy could be written with an admin rather that a coordinator in mind. Shani can help with that later.

amazimbe:
> We need to distinguish between partnerships that were challenged by a school, and those that were removed for another reason. I think this work is good for admins updating the system to reflect support actions on behalf of schools, but that leaves other things missing

Yes, there are plenty of unanswered questions and there's no grain of clarity in the ticket. My goal was to just add something that we can use to explore and clarify future requirements on this.

app/controllers/admin/partnerships_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+        OpenStruct.new(id: :mistake, name: "This looks like a mistake"),
+        OpenStruct.new(id: :other, name: "Other"),
+      ]
+    end
+
+    def reject
+      authorize @partnership, :update?
+      @partnership.update!(reject_params)
+      redirect_to admin_partnership_path(@partnership), notice: "Partnership successfully rejected"
+    end
```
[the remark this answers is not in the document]
amazimbe: I'll add a check for that. Most likely that would have a nil which we don't want

app/models/partnership.rb
ethanmills: Schools don't confirm partnerships in the [current designs](https://miro.com/app/board/o9J_leK0NUo=/?moveToWidget=3074457355028269710&cot=14)

ethanmills:
> > We need to distinguish between partnerships that were challenged by a school, and those that were removed for another reason. I think this work is good for admins updating the system to reflect support actions on behalf of schools, but that leaves other things missing
> 
> Yes, there are plenty of unanswered questions and there's no grain of clarity in the ticket. My goal was to just add something that we can use to explore and clarify future requirements on this.

This would be really good context for the PR description

callumacrae:
As I said on Slack, I feel like the partnerships page is _instructing_ admins to reject partnerships, instead of giving them the option to if they want to, which is probably not the intent!

This is partly because "Reject partnership" is an h1, but I think it could also be improved by moving the reject partnership form to another page and having a button link to it from the current page—this would bring the emphasis of the page back to the information, not the rejection.

app/views/admin/partnerships/show.html.erb
```diff
@@ -0,0 +1,95 @@
...
+          ) %>
+      <%= f.govuk_submit "Submit" %>
```
callumacrae, line 91: This should be a warning button. It might also be clearer if it said "Reject partnership" or just "Reject", instead of "Submit"

amazimbe:
> As I said on Slack, I feel like the partnerships page is _instructing_ admins to reject partnerships, instead of giving them the option to if they want to, which is probably not the intent!
> 
> This is partly because "Reject partnership" is an h1, but I think it could also be improved by moving the reject partnership form to another page and having a button link to it from the current page—this would bring the emphasis of the page back to the information, not the rejection.
> 
> I also think it could be clearer that the action is being taken on behalf of the school, but I can't think of a way to do that that is less clunk than writing "Reject partnership on behalf of school" 😂 maybe some text below the reject partnership heading?

Completely agree. Thought about adding 'Reject partnerships' as a separate page but then decided to do the simplest thing so I could get some feedback.

app/controllers/admin/partnerships_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+module Admin
+  class PartnershipsController < Admin::BaseController
+    skip_after_action :verify_policy_scoped
+    before_action :load_partnership, only: %i[show reject]
+
+    def show
+      authorize @partnership
+      @reasons_for_rejection = [
+        OpenStruct.new(id: :partnered_with_another_provider, name: "I have already partnered with a different training provider"),
```
seb-siem-softwire, line 11: We could make it into a solid class as opposed to using openstruct

amazimbe: I don't think that would be a good idea. This is quite simple, it has no behaviour just some data and some convenient accessors. I can't see any justification for making this a full blown class. And there's always the testing overhead that comes with new units like classes.

amazimbe:
Closing as this will be revisited in a future sprint. I'll keep the branch

PR #136, opened 2021-03-15
Admin suppliers layout

ethanmills:
### Context
Just some frontend tweaks for consistency

- Only link delivery partners
- Remove edit links
- Move lead providers to top of table

### DIFF
--- a/app/helpers/supplier_helper.rb
+++ b/app/helpers/supplier_helper.rb
@@ -2,6 +2,6 @@
 
 module SupplierHelper
   def supplier_link(supplier)
-    supplier.is_a?(LeadProvider) ? nil : edit_admin_delivery_partner_path(supplier)
+    supplier.is_a?(LeadProvider) ? supplier.name : govuk_link_to(supplier.name, edit_admin_delivery_partner_path(supplier))
   end
 end
--- a/app/views/admin/administrators/administrators/index.html.erb
+++ b/app/views/admin/administrators/administrators/index.html.erb
@@ -1,22 +1,18 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">Administrators</h1>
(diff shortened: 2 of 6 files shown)


seb-siem-softwire (approved):

app/helpers/supplier_helper.rb
```diff
@@ -2,6 +2,6 @@
 
 module SupplierHelper
   def supplier_link(supplier)
-    supplier.is_a?(LeadProvider) ? nil : edit_admin_delivery_partner_path(supplier)
+    supplier.is_a?(LeadProvider) ? supplier.name : govuk_link_to(supplier.name, edit_admin_delivery_partner_path(supplier))
```
kicferk1, line 5: Eh, not really a link anymore, is it? :/

app/views/admin/administrators/administrators/index.html.erb
```diff
@@ -1,22 +1,18 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
```
kicferk1, line 2: Do we not want it to be two-thirds? Just making sure.

app/helpers/supplier_helper.rb
kicferk1: Not that I have a good name. `supplier_details_element`?

amazimbe: I think this is just temporary. I would just keep the name

amazimbe (approved):

app/views/admin/administrators/administrators/index.html.erb
callumacrae: I'd say that table-based pages like this should probably be full width

callumacrae (approved):

app/views/admin/administrators/administrators/index.html.erb
ethanmills: The other table page is full width and looks better, just changing this one for consistency

PR #138, opened 2021-03-15
Frontend tweaks from review session

ethanmills:
- [x] cohort display name
- [x] back links from edit and delete delivery partner

### DIFF
--- a/spec/cypress/integration/admin/suppliers/delete-delivery-partner.js
+++ b/spec/cypress/integration/admin/suppliers/delete-delivery-partner.js
@@ -40,7 +40,7 @@ describe("Admin user deleting delivery partner", () => {
     cy.location("pathname").should("contain", "/delete");
 
     cy.get(".govuk-button").contains("Back").click();
-    cy.location("pathname").should("contain", "Edit");
+    cy.location("pathname").should("contain", "/edit");
     cy.get("main").should("contain", deliveryPartnerName);
   });
 });
--- a/spec/cypress/integration/admin/suppliers/delete-delivery-partner.js
+++ b/spec/cypress/integration/admin/suppliers/delete-delivery-partner.js
@@ -24,4 +24,23 @@ describe("Admin user deleting delivery partner", () => {
     cy.location("pathname").should("equal", "/admin/suppliers");
     cy.get("main").should("not.contain", deliveryPartnerName);
   });
+
+  it("has a back button to the edit page", () => {
+    cy.appScenario("admin/suppliers/manage_delivery_partner");
+
+    cy.visit("/admin/suppliers");
+    cy.get("a").contains(deliveryPartnerName).click();
+
+    cy.location("pathname").should(
+      "contain",
+      "/admin/suppliers/delivery-partner"
+    );
+    cy.get(".govuk-button").contains("Delete").click();
+
+    cy.location("pathname").should("contain", "/delete");
+
+    cy.get(".govuk-button").contains("Back").click();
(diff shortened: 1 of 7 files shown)


spec/cypress/integration/admin/suppliers/delete-delivery-partner.js
```diff
@@ -24,4 +24,23 @@ describe("Admin user deleting delivery partner", () => {
...
     cy.get("main").should("not.contain", deliveryPartnerName);
   });
+
+  it("has a back button to the edit page", () => {
+    cy.appScenario("admin/suppliers/manage_delivery_partner");
+
+    cy.visit("/admin/suppliers");
+    cy.get("a").contains(deliveryPartnerName).click();
+
+    cy.location("pathname").should(
```
kicferk1, line 34: I wonder if path names are the best thing to use to determine what page we are on.

Do we want to make use of `<title>` tags at any point? Maybe they would be a better solution?

kicferk1 (approved):

spec/cypress/integration/admin/suppliers/delete-delivery-partner.js
amazimbe: I'm pretty sure pathnames are the best for this sort of check. What reservations do you have about them?

kicferk1: If (when?) we restructure our url structure, these tests will need to change, even though the behaviour won't.

That or we aren't very precise - if we check that url contains `delete`, we aren't guarding against a dev giving incorrect delete path to some redirect or some url.

amazimbe: > If (when?) we restructure our url structure, these tests will need to change, even though the behaviour won't.
> 
> That or we aren't very precise - if we check that url contains `delete`, we aren't guarding against a dev giving incorrect delete path to some redirect or some url.

That's a valid point though if the routing were resourceful the endpoints could be relied on to be very stable. Unfortunately, most of our endpoints are to support workflows / operations and these change often. That said, with titles we could still have a similar issue i.e. people changing the content on the page but not updating the title. And when they do remember to update the title the test would need am update too.

amazimbe (approved):

spec/cypress/integration/admin/suppliers/delete-delivery-partner.js
callumacrae: > @callumacrae thoughts?

Neither are optimal, titles change too. On public sites, URLs usually change less often as it breaks bookmarks and stuff, so I usually prefer testing them - but if it's a few levels deep I'll often only test part of the URL instead of the whole thing so that if it's moved the test still works.

I wonder if there's a way of having something like `cy.location('pathname').should('match', 'edit_admin_delivery_partner_path')` or something like that

seb-siem-softwire: Hmm still not sure about the overall structure and architecture of the tests - I might do some research around it over the weekend - the issue I see in the future is that we will be repeating ourselves quite much with atomic tasks as opposed to creating reusable test structure. Just a feeling but I saw on "Not on the high street" how using atomic actions in scenarios can get out of hand when we don't slow down and ask - is there a better way. Going to do some digging and come back on that

ethanmills: Great, looking forward to seeing what you come up with. I'm going to merge this for now, this sounds like a bigger discussion

PR #139, opened 2021-03-15
Remove success screen at end of admin creation journey

ethanmills:
### Context
We're replacing most of the success pages with notification banners. This is that for the admin creation journey

- Replace success panel with notification banner
- Make use of data-test attributes in a couple of places

### Testing
Updated existing

### DIFF
--- a/app/controllers/admin/administrators/administrators_controller.rb
+++ b/app/controllers/admin/administrators/administrators_controller.rb
@@ -49,6 +49,9 @@ module Admin
           AdminProfile.create!(user: user)
         end
         session.delete(:administrator_user)
+
+        set_success_message(heading: "User added", content: "They have been sent an email to sign in")
+        redirect_to admin_administrators_path
       end
 
       def edit; end
--- a/app/views/admin/administrators/administrators/index.html.erb
+++ b/app/views/admin/administrators/administrators/index.html.erb
@@ -1,5 +1,5 @@
 <h1 class="govuk-heading-l">Administrators</h1>
-<%= govuk_link_to "Create a new administrator", new_admin_administrator_path, button: true %>
+<%= govuk_link_to "Create a new administrator", new_admin_administrator_path, button: true, "data-test" => "create-admin-button" %>
 <table class="govuk-table">
   <thead>
     <tr class="govuk-table__row">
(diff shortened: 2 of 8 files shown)


amazimbe (approved):

app/controllers/admin/administrators/administrators_controller.rb
```diff
@@ -49,6 +49,9 @@ def create
           AdminProfile.create!(user: user)
         end
         session.delete(:administrator_user)
+
+        set_success_message(heading: "User added", content: "They have been sent an email to sign in")
```
kicferk1, line 53: Is this true? Or make belief?

kicferk1 (approved):

app/controllers/admin/administrators/administrators_controller.rb
seb-siem-softwire: Have you double checked the message with Shani?

app/views/admin/administrators/administrators/index.html.erb
```diff
@@ -1,5 +1,5 @@
 <h1 class="govuk-heading-l">Administrators</h1>
-<%= govuk_link_to "Create a new administrator", new_admin_administrator_path, button: true %>
+<%= govuk_link_to "Create a new administrator", new_admin_administrator_path, button: true, "data-test" => "create-admin-button" %>
```
seb-siem-softwire, line 2: Nice - there is however a problem with submit button just to be aware - a workaround is to simply add a class when working with submit buttons

seb-siem-softwire (approved):
Couple of small things but otherwise good - Double check with Shani if the wording is correct

app/views/admin/administrators/administrators/index.html.erb
callumacrae: Sending a PR fixing that today! :)

app/controllers/admin/administrators/administrators_controller.rb
ethanmills: This is Shani's message. Alfred is implementing this. Once we know when we will be sending emails to lead providers I will get some updated copy for that one

PR #140, opened 2021-03-16
A11y tests

ethanmills:
Add accessibility tests for admin journey pages

### Guidance to review
Let me know if you're happy with this structure

### Testing
This is testing


spec/cypress/integration/admin/suppliers/delete-lead-provider-user.js
```diff
@@ -0,0 +1,57 @@
...
+      cy.appScenario("/admin/suppliers/create_lead_provider_user");
+
+      cy.visit("/admin/suppliers/users");
+      cy.get("[data-test=edit-supplier-user-link]")
+        .contains(leadProviderUserName)
+        .click();
+
+      cy.location("pathname").should(
+        "match",
+        /\/admin\/suppliers\/lead-providers\/users\/.*\/edit/
```
ethanmills, line 35: @kicferk1 what do you think of this as a stopgap on path checking? If you're happy with it I will update everywhere to use it

app/views/admin/suppliers/delivery_partners/_lead_providers_cohorts.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= f.govuk_check_boxes_fieldset(
   :lead_provider_ids,
-  legend: { text: "Which lead providers and cohorts are #{@delivery_partner_form.name} working with?", size: "m" },
+  legend: { text: "Which lead providers and cohorts are #{@delivery_partner_form.name} working with?", tag: "h1", size: "l" },
```
kicferk1, line 3: I don't think it's wise to have structurally important headers in a component that doesn't say it does that - `app/views/admin/suppliers/delivery_partners/edit.html.erb` will have two h1 headers because of this change.

spec/cypress/integration/admin/suppliers/delete-lead-provider-user.js
kicferk1: I think that's reasonable. More likely to catch a dev error.

spec/cypress/support/commands.js
```diff
@@ -47,3 +47,9 @@ Cypress.Commands.add("clickBackLink", () => {
 Cypress.Commands.add("clickCommitButton", () => {
   cy.get("[name=commit]").click();
 });
+
+Cypress.Commands.add("chooseNameAndEmailForUser", (name, email) => {
+  cy.get("input[name*=full_name").type(name);
```
callumacrae, line 52: I'd be tempted to just go with `*=name` here! Not that it makes much difference

ethanmills: I'd be worried that would catch names for other things, like organisations

spec/cypress/integration/admin/administrators/create-administrator.js
```diff
@@ -16,9 +16,7 @@ describe("Admin user creating another admin user", () => {
     cy.get("[data-test=create-admin-button").click();
 
     cy.location("pathname").should("equal", "/admin/administrators/new");
-    cy.get("input[name='user[full_name]']").type(fullName);
-    cy.get("input[name='user[email]']").type(email);
-    cy.get(".govuk-button").contains("Continue").click();
+    cy.chooseNameAndEmailForUser(fullName, email);
```
seb-siem-softwire, line 19: Nice!

app/views/admin/suppliers/delivery_partners/_lead_providers_cohorts.html.erb
ethanmills: Good catch

kicferk1 (approved):

seb-siem-softwire (approved):

callumacrae (approved):

PR #141, opened 2021-03-16
Email testing - add unit tests

kicferk1:
### Context
We want more testing around emails. This PR deals with adding them to RSpec - I went through the places I expected to be sending emails and added unit tests to them. That said, devise likes sending some emails implicitly - I guess if the user is confirmed when we create them it won't send a confirmation email? I wonder if I want to test that in all the places that deal with making new users. 

1. A bunch of unit tests - checking what calls `UserMailer` received
2. A primary email notification was created, but never sent. I am not sure that's a good idea - if we don't want emails to be sent to the real world emails, well, that's what the non-live notify keys are for, right?

### Guidance to review
Keen to hear opinions about primary email notifications.
There will be another PR coming at some point, when I figure out how best test emails in cypress.

### Testing
To quote Ethan, "this is testing".

### DIFF
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -22,7 +22,7 @@ private
     school = resource.induction_coordinator_profile.schools.first
 
     if school.primary_contact_email != resource.email
-      UserMailer.primary_contact_notification(resource, school)
+      UserMailer.primary_contact_notification(resource, school).deliver_now
     end
   end
 end
--- a/app/views/users/sessions/new.html.erb
+++ b/app/views/users/sessions/new.html.erb
@@ -1,10 +1,10 @@
+<% content_for :before_content, govuk_back_link(
+  text: "Back",
+  href: registrations_root_path)
+%>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <% content_for :before_content, govuk_back_link(
-      text: "Back",
-      href: registrations_root_path)
-    %>
-
     <h1 class="govuk-heading-l">Sign in</h1>
     <%= form_for(resource, as: resource_name, url: session_path(resource_name)) do |f| %>
       <div class="govuk-form-group">
--- a/app/views/users/sessions/new.html.erb
+++ b/app/views/users/sessions/new.html.erb
@@ -1,9 +1,9 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-  <% content_for :before_content, govuk_back_link(
-    text: "Back",
-    href: registrations_root_path)
-  %>
+    <% content_for :before_content, govuk_back_link(
+      text: "Back",
+      href: registrations_root_path)
+    %>
 
     <h1 class="govuk-heading-l">Sign in</h1>
     <%= form_for(resource, as: resource_name, url: session_path(resource_name)) do |f| %>
@@ -15,6 +15,6 @@
     <% end %>
     <%= govuk_link_to "Register your school", registrations_question_one_path %>
     <p class="govuk-!-margin-bottom-0 govuk-!-margin-top-15"><strong>Problems signing in</strong></p>
-    <p class="govuk-!-margin-top-1">Contact us: <%= mail_to "ecf-support@digital.education.gov.uk" %> </p>
+    <p class="govuk-!-margin-top-1">Contact us: <%= mail_to "ecf-support@digital.education.gov.uk" %></p>
   </div>
 </div>
--- a/spec/requests/registrations/user_profile_spec.rb
+++ b/spec/requests/registrations/user_profile_spec.rb
@@ -23,38 +23,103 @@ RSpec.describe "Registrations::UserProfile", type: :request do
   end
 
   describe "POST /registrations/user-profile" do
-    it "creates an induction coordinator profile" do
-      post "/registrations/user-profile", params: { user: {
-        full_name: full_name,
-        email: email,
-      } }
+    before do
+      mail = instance_double(ActionMailer::MessageDelivery, deliver_now: true)
+      allow(UserMailer).to receive(:confirmation_instructions).and_return(mail)
+    end
 
-      expect(created_user.full_name).to eq(full_name)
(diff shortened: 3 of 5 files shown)


app/views/users/sessions/new.html.erb
```diff
@@ -1,9 +1,9 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-  <% content_for :before_content, govuk_back_link(
-    text: "Back",
-    href: registrations_root_path)
-  %>
+    <% content_for :before_content, govuk_back_link(
+      text: "Back",
+      href: registrations_root_path)
+    %>
```
ethanmills, line 6: I prefer this right at the top of the file, since it doesn't sit inside this div. (I know you didn't put it here, but you touched it last 😛 )

spec/requests/registrations/user_profile_spec.rb
```diff
@@ -23,39 +23,104 @@
...
+      allow(UserMailer).to receive(:confirmation_instructions).and_return(mail)
+    end
+
+    context "when request is valid" do
+      let(:send_request) do
+        post "/registrations/user-profile", params: { user: {
+          full_name: full_name,
+          email: email,
+        } }
+      end
```
ethanmills, line 37: What does everyone think about this being a `let`? I personally would make this a private method, but I'd like to know what the consensus is. Since this is per-test memoised, it feels wrong to use `let` for side effects

ethanmills: Looking at the implementation, this maybe lives in a `before` block? (I think I would still put it in a private method too in order to give it a nice name)

spec/requests/registrations/user_profile_spec.rb
```diff
@@ -23,39 +23,104 @@
...
+
+      it "redirects to verification sent page" do
+        send_request
+        expect(response).to redirect_to(:registrations_verification_sent)
+      end
+
+      it "sends a confirmation email" do
+        send_request
+        expect(UserMailer).to have_received(:confirmation_instructions).with(created_user, /.*/, {})
```
ethanmills, line 52: What are the params to `with` here?

spec/requests/users/sessions_spec.rb
```diff
@@ -24,11 +24,23 @@
   end
 
   describe "POST /users/sign_in" do
-    context "when sign in email has been sent" do
+    let(:login_url_regex) { /http:\/\/localhost:3000\/users\/confirm_sign_in\?login_token=.*/ }
```
ethanmills, line 27: Nice

spec/requests/registrations/user_profile_spec.rb
kicferk1: I should probably name them. They are a user, a token, and options.

app/controllers/users/confirmations_controller.rb
```diff
@@ -22,7 +22,7 @@ def notify_school_primary_contact
     school = resource.induction_coordinator_profile.schools.first
 
     if school.primary_contact_email != resource.email
-      UserMailer.primary_contact_notification(resource, school)
+      UserMailer.primary_contact_notification(resource, school).deliver_now
```
amazimbe, line 25: 👍

spec/requests/registrations/user_profile_spec.rb
amazimbe: I've seen let used for this sort of thing loads of times too but I think a private method would be a better idea as it encapsulates an action / behaviour

kicferk1: It doesn't really live in a `before` block, sadly.

```
context "when school has old unconfirmed registrations" do
```
context needs to do its `before` block before sending off a request. 

I guess I could separate it out from the `when request is valid` context though...

spec/requests/registrations/user_profile_spec.rb
```diff
@@ -23,39 +23,104 @@
...
-    it "creates an induction coordinator profile" do
-      post "/registrations/user-profile", params: { user: {
-        full_name: full_name,
-        email: email,
-      } }
-
-      expect(created_user.full_name).to eq(full_name)
-      expect(response).to redirect_to(:registrations_verification_sent)
-      expect(created_user.induction_coordinator_profile.schools).to include(school)
+    before do
```
amazimbe, line 26: So did we just decide to go with the tried and tested mocks / stubs in the end. I remember there were some complex alternatives in the discussions earlier. Probably best to keep it simple though. I don't think there was a strong case for the complex solution.

kicferk1: Well, there is a second part in cypress tests coming up. They do look more 'real life'. For the purposes of unit tests, I don't believe there is much point in having anything more complicated than a stub checking what method was called with what arguments on mailers.

In cypress tests we play with `ActionMailer::Base.deliveries` which look a lot more like real requests to notify :)

amazimbe: One of the alternatives was along the lines of building something that could mimic notify I think. That's what I was referring to.

kicferk1: When working on cypress tests I really warmed up to the default Railsy idea of 'when tests are running, put the emails in an array rather than sending them'. That imo leads to a clean and pain-free way of verifying our emails, without the need to include a notify stub. 

And well, for unit test purposes, I really think stubbing the mailer is the way to go - there is barely any logic in the mailer itself, and if there is we can unit test that specifically at a lower level. Stubbing notify project-wide in tests doesn't seem to provide much benefit.

amazimbe (approved):

spec/requests/users/sessions_spec.rb
```diff
@@ -24,11 +24,23 @@
...
+
+    before do
+      mail = instance_double(ActionMailer::MessageDelivery, deliver_now: true)
```
amazimbe, line 30: By the way, I can't see why you need this?

ethanmills (approved):

spec/requests/users/sessions_spec.rb
kicferk1: Good point, calling the original method is a better idea :)

PR #142, opened 2021-03-16
Improve deletion flow consistency

callumacrae:
### Context

There were a couple different designs for the deletion pages before, this makes them all looks the same and (hopefully) makes them look a bit nicer

Button groups

Also, there was no flash message after deleting a delivery provider - I fixed that

### Guidance to review

Percy would help here :D

### Testing

yes

### DIFF
--- a/app/views/admin/administrators/administrators/delete.html.erb
+++ b/app/views/admin/administrators/administrators/delete.html.erb
@@ -1,6 +1,8 @@
 <h1>Do you want to delete this user?</h1>
 <p class="govuk-!-font-weight-bold">Admin user: <span class="govuk-!-font-weight-regular"><%= @administrator.full_name %></span></p>
 <%= form_for :confirmation, url: admin_administrator_path(@administrator), method: :delete do |f| %>
-  <%= f.govuk_submit "Delete", classes: "govuk-button--warning data-test-delete-submit-button" %>
+  <div class="govuk-button-group">
+    <%= f.govuk_submit "Delete", classes: "govuk-button--warning data-test-delete-submit-button" %>
+    <%= govuk_link_to "Back", edit_admin_administrator_path(@administrator), button: true, class: "govuk-button--secondary" %>
+  </div>
 <% end %>
-<%= govuk_link_to "Back", edit_admin_administrator_path(@administrator), button: true, class: "govuk-button--secondary" %>
--- a/app/views/admin/administrators/administrators/edit.html.erb
+++ b/app/views/admin/administrators/administrators/edit.html.erb
@@ -13,10 +13,12 @@
             :email,
             label: { text: "Email" })
       %>
-      <%= f.govuk_submit "Save" %>
-      <% if current_user.id != @administrator.id %>
-        <%= govuk_link_to "Delete", delete_admin_administrator_path(@administrator), button: true, class: "govuk-button--warning", "data-test" => "delete-button" %>
-      <% end %>
+      <div class="govuk-button-group">
+        <%= f.govuk_submit "Save" %>
+        <% if current_user.id != @administrator.id %>
+          <%= govuk_link_to "Delete", delete_admin_administrator_path(@administrator), button: true, class: "govuk-button--warning", "data-test" => "delete-button" %>
+        <% end %>
+      </div>
     <% end %>
   </div>
 </div>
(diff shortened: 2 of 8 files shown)


app/views/admin/administrators/administrators/delete.html.erb
```diff
@@ -1,6 +1,8 @@
-<h1>Do you want to delete this user?</h1>
-<p class="govuk-!-font-weight-bold">Admin user: <span class="govuk-!-font-weight-regular"><%= @administrator.full_name %></span></p>
+<h1>Confirm deletion?</h1>
+<p class="govuk-body-l">Once deleted, <%= @administrator.full_name %> won't be recoverable.</p>
```
callumacrae, line 2: I didn't really know what else to write here, but `Admin user: [user]` felt clunky to me

kicferk1: `Once deleted, <%= @administrator.full_name %>'s account won't be recoverable`?

callumacrae: Hah, it's the `won't be recoverable` I thought was a bit weird (isn't it technically not even true?), but yeah, that's a good suggestion

kicferk1 (approved):

app/views/admin/administrators/administrators/delete.html.erb
kicferk1: Well, there is no way for the user to recover it yet, soooo, it's not technically lying :D

ethanmills (changes requested):
This was specified by Shani, please check this thread before changing anything: https://ukgovernmentdfe.slack.com/archives/C01RSEKR56C/p1615559271010100

callumacrae:
Have changed them to match the copy from that slack thread - but the buttons should definitely be next to each other!

app/views/admin/administrators/administrators/delete.html.erb
seb-siem-softwire: This was exact text that Shani had requested - double-check with her if the text is correct

app/views/admin/administrators/administrators/delete.html.erb
```diff
@@ -1,6 +1,8 @@
-<h1>Do you want to delete this user?</h1>
-<p class="govuk-!-font-weight-bold">Admin user: <span class="govuk-!-font-weight-regular"><%= @administrator.full_name %></span></p>
+<h1>Do you want to delete this admin user?</h1>
```
seb-siem-softwire, line 1: Double check with Shani if the messaging is right. As long as she is happy it should be ok

app/views/admin/administrators/administrators/edit.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l">Edit user details</h1>
+    <h1 class="govuk-heading-l">Edit administrator details</h1>
```
seb-siem-softwire, line 5: Double check with Shani

spec/cypress/integration/admin/administrators/delete-administrator-user.js
```diff
@@ -12,10 +12,10 @@ describe("Admin user deleting another admin user", () => {
     cy.get("[data-test=edit-admin-link]").contains(adminUserName).click();
     cy.get("[data-test=delete-button]").click();
     // @todo replace with visual test
-    cy.get("main").should("contain", "Confirm deletion?");
+    cy.get("main").should("contain", "Do you want to delete this admin user?");
```
kicferk1: Hehe, don't you wish we had visual tests that you would have to recreate instead of this? :D

callumacrae: I wouldn't have needed to change it three times in this PR so far if they were visual tests because visual tests only need approving once! (and they're approved by pressing a button, not by having to modify a test file)

ethanmills (approved):

spec/cypress/integration/admin/administrators/delete-administrator-user.js
kicferk1: Oh, I thought they work more like snapshot tests. Interesting, now I hope we get them even more.

callumacrae:
I've reverted all the changes but the button groups and missing flash message, this PR wasn't worth the amount of time it was taking to get it through the review process

amazimbe (approved):

seb-siem-softwire (approved):

PR #143, opened 2021-03-16
Email testing in cypress

kicferk1:
### Context
We would like to test that certain user flows fire off certain emails in cypress.

Use our existing test setup, which stores sent emails in `ActionMailer::Base.deliveries`.

### Guidance to review
I added a method to test for each type of email we have - but I am not convinced we have cypress tests triggering either confirmation or primary school contact email. Do we (am I missing them)? If not, should I add them? I heard rumours that the flow sending them will be redesigned and re-done, so is it worth our time to write cypress tests for it? If we release before it's re-written, probably yes? 

### Testing
This is testing.


callumacrae:
what's up with the git history in this pr!

kicferk1:
> what's up with the git history in this pr!

Uhh, I don't know, but I will get rid of some old things :/

callumacrae:
> Uhh, I don't know, but I will get rid of some old things :/

Did you create this branch off develop then send to a different branch in this PR? If so, you can probably fix it by rebasing the branch you're sending this to onto develop, pushing that, then rebasing this branch onto the branch you're sending to and pushing again

kicferk1:
Ah, I set it to be on top of the other email testing branch first, and then realised I didn't need to - so I rebased on top of develop, force pushed and forgot to change the branch to `develop` xD

callumacrae:
Or that!

spec/cypress/integration/accessibility.js
```diff
@@ -12,6 +12,9 @@ describe("Accessibility", () => {
     cy.get('[name="user[email]"]').type("doesntexist@example.com{enter}");
     cy.get("h1").should("contain", "You do not have an account");
     cy.checkA11y();
+    cy.appEval("ActionMailer::Base.deliveries").then((result) => {
```
callumacrae, line 15: This is neat! Imo though, it should probably be separated out into a command to make it a bit more readable - e.g. `cy.appSentEmails().then(` or something

amazimbe:
I'm not convinced these tests need to be in Cypress. Why can't we just do them in rspec and avoid all the configuration? I don't think there's anything user-facing / UI oriented in the emails to justify using cypress.

kicferk1:
> I'm not convinced these tests need to be in Cypress. Why can't we just do them in rspec and avoid all the configuration? 

What is the configuration you hope to avoid? This is adding barely anything - a few cypress commands and a call to clean emails sent before each cypress test. The rest is 'for free'. 

> I don't think there's anything user-facing / UI oriented in the emails to justify using cypress.

We don't check that emails sent to users look in a certain way - because the display (and in fact a lot of the content) lives on Notify's end.

We check that a certain user flow triggered a correct type of email with correct params being sent to notify. 

Do we need to do this for the sign in emails? Well, maybe not. Is it useful to have an example of testing emails in cypress? Probably yes.

seb-siem-softwire (approved):
Looks simple enough. Having said that I will do some research over the weekend about the architecture of cypress tests

amazimbe:
> > I'm not convinced these tests need to be in Cypress. Why can't we just do them in rspec and avoid all the configuration?
> 
> What is the configuration you hope to avoid? This is adding barely anything - a few cypress commands and a call to clean emails sent before each cypress test. The rest is 'for free'.
> 
> > I don't think there's anything user-facing / UI oriented in the emails to justify using cypress.
> 
> We don't check that emails sent to users look in a certain way - because the display (and in fact a lot of the content) lives on Notify's end.
> 
> We check that a certain user flow triggered a correct type of email with correct params being sent to notify.
> 
> Do we need to do this for the sign in emails? Well, maybe not. Is it useful to have an example of testing emails in cypress? Probably yes.

Personally, I just feel this sort of test belongs in ruby / rails land. Checking the array of emails etc just naturally belongs in a ruby / rails ecosystem. I can live with it here though as an example of how to do it in cypress.

kicferk1:
Well, on a philosophical level, I think this goes to a question of "what role(s) do cypress tests serve in our process?".

I guess one role it currently serves is front end tests - checking accessibility, checking navigation, checking that certain content appears in certain places. And compared to that, checking what emails were sent looks odd.

But I think the role cypress fills in our testing is larger - it's essentially what we use for full system tests. It's the highest fidelity tool we have for simulating user experience - it runs in a browser, it operates on the level of what elements it can access on each page, it navigates to pages through user stories. 

And for that role, I think testing which emails we send is perfectly suitable. I wouldn't want to exclude a part of our system from our system tests, especially if it's a part we care about - and sending emails is. And, well, the fact that we need to check a Ruby array to get that is a limitation of cypress as a tool that runs on the client.

ethanmills (approved):

amazimbe:
> Well, on a philosophical level, I think this goes to a question of "what role(s) do cypress tests serve in our process?".
> 
> I guess one role it currently serves is front end tests - checking accessibility, checking navigation, checking that certain content appears in certain places. And compared to that, checking what emails were sent looks odd.
> 
> But I think the role cypress fills in our testing is larger - it's essentially what we use for full system tests. It's the highest fidelity tool we have for simulating user experience - it runs in a browser, it operates on the level of what elements it can access on each page, it navigates to pages through user stories.
> 
> And for that role, I think testing which emails we send is perfectly suitable. I wouldn't want to exclude a part of our system from our system tests, especially if it's a part we care about - and sending emails is. And, well, the fact that we need to check a Ruby array to get that is a limitation of cypress as a tool that runs on the client.

But the principle of different types of tests is always to go with the cheapest first i.e. unit tests and then the next level which would be some integration tests for the things which unit tests don't cover adequately etc. It's not to use the tool that best mimics what our users will be using for everything just because we can. But as I said, it's not a big deal, the most important thing is that we have automated tests.

PR #144, opened 2021-03-16
ECF-RP-370: Create github action create PR to release to staging

ethanmills:
### Context
Process for releasing to staging:
 - Manually trigger the new workflow, specify a git ref to deploy and a version number
 - PR is opened automatically
 - PR is approved and merged
 - Tag is created from commit message
 - Deploy to staging happens

- Create workflow to create pull request of correct format
- Tag release before deploying to staging (deploy fails if no tag)

### DIFF
--- a/.github/workflows/trigger_staging_deploy.yml
+++ b/.github/workflows/trigger_staging_deploy.yml
@@ -1,0 +1,38 @@
+name: Trigger release to staging
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Which version are you creating?
+        required: true
+      ref:
+        description: Git ref to deploy
+        required: true
+        default: develop
+jobs:
+  create-pr:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        name: Checkout Code
+        with:
+          ref: ${{ github.event.inputs.ref }}
+
+      - name: Update master
+        run: |
+          git fetch
+          git branch tmp-branch
+          git switch master
+          git reset --hard tmp-branch --
+          git branch -D tmp-branch
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{ secrets.PERSONAL_TOKEN }}
+          branch: staging-release
+          title: Release version ${{ github.event.inputs.version }} to staging
+          team-reviewers: DFE-Digital/ecf
+          base: master
+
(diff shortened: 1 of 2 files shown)


.github/workflows/trigger_staging_deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+        name: Checkout Code
+        with:
+          ref: ${{ github.event.inputs.ref }}
+
+      - name: Update master
+        run: |
+          git fetch
+          git branch tmp-branch
+          git switch master
+          git reset --hard tmp-branch --
```
ethanmills, line 27: Unfortunately this doesn't mean we get to keep the original commits, as we still have the PR merge

.github/workflows/trigger_staging_deploy.yml
```diff
@@ -0,0 +1,38 @@
+name: Trigger release to staging
+
+on:
```
seb-siem-softwire, line 3: For the purpose of self-education and understanding the syntax - which resources would you recommend to decipher this (I'm assuming it's terraform or its variant)

ethanmills: No, this is a github action https://docs.github.com/en/actions

ethanmills:
[Release to staging](https://trello.com/c/uhHaouQH/370-release-to-staging)

.github/workflows/trigger_staging_deploy.yml
```diff
@@ -0,0 +1,38 @@
+name: Trigger release to staging
```
kicferk1, line 1: Why do we need this action exactly?

So we don't have to create PRs manually and instead fire them off from the actions tab?

ethanmills: Yes, because the format of the commit message is important (it's where the version number comes from). It also allows us to specify any ref to deploy, which is slightly easier than doing it manually. 

Another advantage is it is extensible: I can imagine us automatically running smoke tests as a result of this or similar

kicferk1: That just leaves one question. Do we want to prevent any other ways of getting code into master? Is that even possible?

ethanmills: Everything has to go through PR, and any commits which don't have the right message won't trigger a deploy. I don't know that there's a way to disable manual PRs, especially since it will be using your personal token

kicferk1 (approved):

.github/workflows/trigger_staging_deploy.yml
seb-siem-softwire: 👍

seb-siem-softwire (approved):

PR #145, opened 2021-03-17
[ECF-1-364] Email newly created admin

amazimbe:
### Context

We need to send a welcome email to admins who have had their accounts created by other admins.

Added a new admin mailers and some tests.

### DIFF
--- a/app/controllers/admin/administrators/administrators_controller.rb
+++ b/app/controllers/admin/administrators/administrators_controller.rb
@@ -47,6 +47,9 @@ module Admin
         ActiveRecord::Base.transaction do
           user.save!
           AdminProfile.create!(user: user)
+          AdminMailer.account_created_email(
+            user, helpers.new_user_session_url
+          ).deliver_now
         end
         session.delete(:administrator_user)
 
--- a/spec/requests/admin/administrators/administrators_spec.rb
+++ b/spec/requests/admin/administrators/administrators_spec.rb
@@ -97,6 +97,15 @@ RSpec.describe "Admin::Administrators::Administrators", type: :request do
         }),
       )
     end
+
+    it "sends new admin an account created email" do
+      url = "http://www.example.com/users/sign_in"
+      allow(AdminMailer).to receive(:account_created_email).and_call_original
+
+      given_a_user_is_created
+
+      expect(AdminMailer).to have_received(:account_created_email).with(new_user, url)
+    end
   end
 
   let(:admin_profile) { create(:admin_profile) }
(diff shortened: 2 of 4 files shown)


spec/requests/admin/administrators/administrators_spec.rb
```diff
@@ -97,6 +97,16 @@
         }),
       )
     end
+
+    it "sends new admin an account created email" do
+      mail = instance_double(ActionMailer::MessageDelivery, deliver_now: true)
+      url = "http://www.example.com/users/sign_in"
+      allow(AdminMailer).to receive(:account_created_email).and_return(mail)
```
kicferk1, line 104: I know I introduced the `mail = ...` stuff, but I since found a better solution - `and_call_original` :)

kicferk1 (approved):

app/controllers/admin/administrators/administrators_controller.rb
```diff
@@ -47,6 +47,9 @@ def create
         ActiveRecord::Base.transaction do
           user.save!
           AdminProfile.create!(user: user)
+          AdminMailer.account_created_email(
+            user, helpers.new_user_session_url
```
amazimbe, line 51: So the welcome email would have a link to the sign in page. However, personally I think it would be more convenient for the user if this were a magic link that they could click and log in. Just wondering what others think.

spec/requests/admin/administrators/administrators_spec.rb
amazimbe: I did replace that, not sure why it's still there

app/controllers/admin/administrators/administrators_controller.rb
ethanmills: It should be a link to the sign in page, because chances are the magic link will have expired by the time they get to the email. The user isn't the one who triggers this

ethanmills: That's what Al specified anyway

spec/requests/admin/administrators/administrators_spec.rb
amazimbe: Forgot to push the latest, thanks

ethanmills (approved):

seb-siem-softwire (approved):

kicferk1:
Did paas improve error messages? 

> Error: Oh no. Something's gone wrong. I've tried to roll back but you should check to see if everything is OK.: The app name is taken: ecf-review-pr-145

amazimbe:
[Email admins on account creation](https://trello.com/c/nmpcuz3g/364-email-admins-on-account-creation)

PR #146, opened 2021-03-17
ECF-RP-329: login page updates

ethanmills:
### Context
Prevent user enumeration by not showing a different page when no account exists

This spiralled a little bit. I ended up getting loads of copy changes, and bundling them together. It should be ok if you go commit-wise, but happy to split if needed

### DIFF
--- a/app/controllers/admin/suppliers/supplier_users_controller.rb
+++ b/app/controllers/admin/suppliers/supplier_users_controller.rb
@@ -67,7 +67,7 @@ module Admin
         @supplier_user_form.save!
         session.delete(:supplier_user_form)
 
-        set_success_message(heading: "User added", content: "")
+        set_success_message(heading: "User added")
         redirect_to admin_supplier_users_path
       end
 
--- a/app/controllers/admin/suppliers/supplier_users_controller.rb
+++ b/app/controllers/admin/suppliers/supplier_users_controller.rb
@@ -67,7 +67,7 @@ module Admin
         @supplier_user_form.save!
         session.delete(:supplier_user_form)
 
-        set_success_message(heading: "User added", content: "They have been sent an email to sign in")
+        set_success_message(heading: "User added", content: "")
         redirect_to admin_supplier_users_path
       end
 
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -17,7 +17,7 @@ class Users::SessionsController < Devise::SessionsController
   rescue LoginIncompleteError
     render :login_email_sent
   rescue EmailNotFoundError
-    redirect_to :registrations_account_not_found
+    render :login_email_sent
   end
 
   def sign_in_with_token
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -42,7 +42,7 @@
           <% end %>
         </div>
         <div class="govuk-header__content">
-          <%= link_to "Early careers framework", "/", class: "govuk-header__link govuk-header__link--service-name" %>
+          <%= link_to "Early career framework", "/", class: "govuk-header__link govuk-header__link--service-name" %>
           <button type="button" role="button" data-module="govuk-button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
           <% if user_signed_in? %>
           <nav>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template ">
   <head>
-    <title>Early careers framework</title>
+    <title><%= (yield(:title) + " - " unless yield(:title).blank?).to_s + "Early career framework" %></title>
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
     <%= canonical_tag %>
--- a/spec/cypress/integration/accessibility.js
+++ b/spec/cypress/integration/accessibility.js
@@ -10,7 +10,7 @@ describe("Accessibility", () => {
 
     // School not registered page
     cy.get('[name="user[email]"]').type("doesntexist@example.com{enter}");
-    cy.get("h1").should("contain", "You do not have an account");
+    cy.get("h1").should("contain", "we have sent you an email");
     cy.checkA11y();
     cy.appSentEmails().then((emails) => {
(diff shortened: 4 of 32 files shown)


ethanmills:
[Login page changes](https://trello.com/c/6Wix3bMY/329-login-page-changes)

app/controllers/users/sessions_controller.rb
```diff
@@ -17,7 +17,7 @@ def create
   rescue LoginIncompleteError
     render :login_email_sent
   rescue EmailNotFoundError
-    redirect_to :registrations_account_not_found
+    render :login_email_sent
```
amazimbe, line 20: This doesn't look right. If the email doesn't exist I think depending on the user type they may have to contact support.

app/views/layouts/application.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template ">
   <head>
-    <title>Early careers framework</title>
+    <title><%= (yield(:title) + " - " unless yield(:title).blank?).to_s + "Early career framework" %></title>
```
amazimbe, line 4: I think if you interpolate yield(:title) + " - " you do away with the need for to_s

spec/cypress/integration/accessibility.js
```diff
@@ -10,7 +10,7 @@ describe("Accessibility", () => {
 
     // School not registered page
     cy.get('[name="user[email]"]').type("doesntexist@example.com{enter}");
-    cy.get("h1").should("contain", "You do not have an account");
+    cy.title().should("equal", "Check your email - Early career framework");
```
kicferk1, line 13: I love this

spec/cypress/integration/admin/administrators/create-administrator.js
```diff
@@ -10,15 +10,15 @@ describe("Admin user creating another admin user", () => {
     const email = "j.smith@example.com";
 
     cy.visit(basePath);
-    cy.get("[data-test=create-admin-button").click();
+    cy.get("[data-test=create-admin-button]").click();
```
kicferk1, line 13: Huh, this was passing, right? Sometimes Cypress surprises me.

app/controllers/admin/suppliers/supplier_users_controller.rb
```diff
@@ -67,7 +67,7 @@ def create
         @supplier_user_form.save!
         session.delete(:supplier_user_form)
 
-        set_success_message(heading: "User added", content: "They have been sent an email to sign in")
+        set_success_message(heading: "User added", content: "")
```
kicferk1, line 70: Can we set content to be empty by default in the method, and not pass empty content?

app/controllers/users/sessions_controller.rb
kicferk1: This PR also changes the template for this action - it says something like
"If you have an account we sent you an email. You can check if you have an account under this link"

And checking the account is "contact the right people to know".

kicferk1 (commented):
I like page titles :)

app/controllers/users/sessions_controller.rb
ethanmills: The point is to make it so that you can't tell whether an account exists or not just by entering the email address

spec/cypress/integration/admin/administrators/create-administrator.js
amazimbe: browsers are very lenient

app/controllers/admin/suppliers/supplier_users_controller.rb
callumacrae: Is this temporary? If not, could I suggest `heading: "Success", content: "User added"`?

spec/cypress/integration/accessibility.js
callumacrae: I don't think we should be including ` - Early career framework` in every test

spec/cypress/integration/admin/administrators/create-administrator.js
ethanmills: That's the browser, not Cypress. That's behaviour of `document.querySelector`

kicferk1: Makes sense - that's a little less surprising considering what other things they let you get away with (broken html? no worries, lemme fix that for you :D)

spec/cypress/integration/admin/administrators/create-administrator.js
```diff
@@ -10,15 +10,15 @@ describe("Admin user creating another admin user", () => {
...
 
     cy.location("pathname").should("equal", `${basePath}/new`);
     cy.chooseNameAndEmailForUser(fullName, email);
 
     cy.location("pathname").should("equal", `${basePath}/new/confirm`);
     cy.get("main").should("contain", fullName);
     cy.get("main").should("contain", email);
-    cy.get("input.govuk-button").contains("Create administrator user").click();
+    cy.clickCommitButton();
```
seb-siem-softwire, line 21: I think having helper methods like this is better and more readable - we could extend that in future PR's

amazimbe (approved):

spec/cypress/integration/accessibility.js
ethanmills: Agreed, we'll need a helper

app/controllers/admin/suppliers/supplier_users_controller.rb
ethanmills: Nope, Success is the title

ethanmills: image
Double success would be weird

spec/cypress/integration/admin/administrators/create-administrator.js
ethanmills: It will all be cucumber

seb-siem-softwire: Stay tuned this weekend I want to spend some time on cypress , cucumber and page object pattern

app/views/layouts/application.html.erb
ethanmills: @amazimbe I think otherwise when :title is blank, you get `nil` and can't do `nil + string`

kicferk1 (approved):
Soo, how about changing 
```
set_success_message(heading: "User added", content: "")
```
to
```
set_success_message(heading: "User added")
```

I don't really think it makes a difference, I just feel sad seeing an empty string being passed :)

PR #147, opened 2021-03-17
ECF-RP-379: Initial seed data

ethanmills:
### Context
CIPs, Cohorts, and LPs are to be seeded for initial release

### Guidance to review
- Check against details on ticket
- Make sure it's idempotent

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,4 +1,7 @@
 # frozen_string_literal: true
 
 Dir[Rails.root.join("db/seeds/initial_seed.rb")].each { |seed| load seed }
-Dir[Rails.root.join("db/seeds/dummy_structures.rb")].each { |seed| load seed }
+
+if Rails.env.development? || Rails.env.deployed_development?
+  Dir[Rails.root.join("db/seeds/dummy_structures.rb")].each { |seed| load seed }
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-Dir[Rails.root.join("db/seeds/initial_seed.rb")].each { |seed| load seed }
+require_relative "seeds/initial_seed"
 
 if Rails.env.development? || Rails.env.deployed_development?
-  Dir[Rails.root.join("db/seeds/dummy_structures.rb")].each { |seed| load seed }
+  require_relative "seeds/dummy_structures"
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,3 +1,4 @@
 # frozen_string_literal: true
 
+Dir[Rails.root.join("db/seeds/initial_seed.rb")].each { |seed| load seed }
 Dir[Rails.root.join("db/seeds/dummy_structures.rb")].each { |seed| load seed }
--- a/db/seeds/dummy_structures.rb
+++ b/db/seeds/dummy_structures.rb
@@ -1,44 +1,24 @@
 # frozen_string_literal: true
 
-if Rails.env.development? || Rails.env.deployed_development?
-  SchoolDataImporter.new(Rails.logger).delay.run
+SchoolDataImporter.new(Rails.logger).delay.run
 
-  standard_user_emails = ["admin@example.com", "lead-provider@example.com", "second-school-leader@example.com", "school-leader@example.com"]
-  (1..5).each do |number|
-    standard_user_emails << "second-admin-#{number}@example.com"
-    standard_user_emails << "second-lead-provider-#{number}@example.com"
-    standard_user_emails << "second-school-leader-#{number}@example.com"
-  end
+standard_user_emails = ["admin@example.com", "lead-provider@example.com", "second-school-leader@example.com", "school-leader@example.com"]
+(1..5).each do |number|
+  standard_user_emails << "second-admin-#{number}@example.com"
+  standard_user_emails << "second-lead-provider-#{number}@example.com"
+  standard_user_emails << "second-school-leader-#{number}@example.com"
+end
 
-  User.with_discarded.where(email: standard_user_emails).each do |user|
-    user.undiscard
-    user.admin_profile&.undiscard
-    user.lead_provider_profile&.undiscard
-    user.induction_coordinator_profile&.undiscard
-  end
+User.with_discarded.where(email: standard_user_emails).each do |user|
+  user.undiscard
+  user.admin_profile&.undiscard
+  user.lead_provider_profile&.undiscard
+  user.induction_coordinator_profile&.undiscard
+end
 
-  (1..5).each do |number|
-    user = User.with_discarded.find_or_create_by!(email: "second-admin-#{number}@example.com") do |u|
-      u.full_name = "John Doe - #{number}"
-      u.confirmed_at = Time.zone.now.utc
-    end
-
-    user.undiscard
-
-    AdminProfile.with_discarded.find_or_create_by!(user: user).undiscard
-
-    user = User.with_discarded.find_or_create_by!(email: "second-lead-provider-#{number}@example.com") do |u|
-      u.full_name = "Merry Doe - #{number}"
-      u.confirmed_at = Time.zone.now.utc
-    end
-
-    user.undiscard
-
-    LeadProviderProfile.with_discarded.find_or_create_by!(user: user, lead_provider: LeadProvider.first).undiscard
-  end
-
-  user = User.with_discarded.find_or_create_by!(email: "admin@example.com") do |u|
-    u.full_name = "Admin User"
+(1..5).each do |number|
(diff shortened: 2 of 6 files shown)


ethanmills:
[Seed data for providers](https://trello.com/c/AE4wyyWG/379-seed-data-for-providers)

db/seeds/initial_seed.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+cohort2021 = Cohort.find_or_create_by!(start_year: 2021)
```
kicferk1, line 3: nitpick, but can we have a `_` between words and numbers? `cohort_2021`?

db/seeds/initial_seed.rb
```diff
@@ -0,0 +1,32 @@
...
+
+ambition_cip = CoreInductionProgramme.find_or_create_by!(name: "Ambition Institute")
+edt_cip = CoreInductionProgramme.find_or_create_by!(name: "Education Development Trust")
+teach_first_cip = CoreInductionProgramme.find_or_create_by!(name: "Teach First")
+ucl_cip = CoreInductionProgramme.find_or_create_by!(name: "UCL Institute of Education")
+
+ambition = LeadProvider.find_or_create_by!(name: "Ambition Institute")
+ambition.update!(cohorts: [cohort2021]) unless ambition.cohorts.any?
+LeadProviderCip.find_or_create_by!(lead_provider: ambition, cohort: cohort2021, core_induction_programme: ambition_cip)
```
kicferk1, line 12: Hmmmm, now that I see it, I wonder if `cohorts` field on `LeadProvider` is duplicating data included in `lead_provider_cips`. And if there is a clever way of avoiding it.

Thoughts?

db/seeds/dummy_structures.rb
```diff
@@ -1,32 +1,8 @@
...
-if CoreInductionProgramme.none?
-  CoreInductionProgramme.find_or_create_by!(name: "Ambition Institute")
-  CoreInductionProgramme.find_or_create_by!(name: "Education Development Trust")
-  CoreInductionProgramme.find_or_create_by!(name: "Teach First")
-  CoreInductionProgramme.find_or_create_by!(name: "UCL")
-end
-
 if Rails.env.development? || Rails.env.deployed_development?
+  SchoolDataImporter.new(Rails.logger).delay.run
+
```
kicferk1, line 5: Soo, since this won't run on seed, will we have to run it manually first? Or will we wait till next day to check if it worked?

amazimbe:
Just that question we touched on yesterday: do we want to use a seed file, a migration or a rake task. Personally, I would have used a rake task that I would probably remove after we go live.

kicferk1:
> Just that question we touched on yesterday: do we want to use a seed file, a migration or a rake task. Personally, I would have used a rake task that I would probably remove after we go live.

Well, seeding is a rake task, technically.

And for our sake, it will be convenient to have that data in our PR apps (I think) - so sticking it in `seed` task makes sense to me.

db/seeds/initial_seed.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
```
amazimbe, line 1: What about just calling the file production.rb? Generally, I wish we had environment specific seeds rather than having the environment checks in the files.

ethanmills: Is there anywhere we don't want this to run?

db/seeds/initial_seed.rb
ethanmills: Rubocop told me off for that https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/VariableNumber

ethanmills: Happy to disable the cop if we have consensus

db/seeds/initial_seed.rb
amazimbe: No, if it's good enough for prod then it's good enough for every other env. What I feel queasy about is that we are running dummy structures as well and that has env checks in it. I'd sleep better if we had an env specific file that only has what we need to run in prod.

db/seeds/initial_seed.rb
ethanmills: I think so, but I really don't want to change that here. That seems complex

db/seeds/initial_seed.rb
kicferk1: I guess we could move the environment check out of the dummy structures file and into the main seed file?

db/seeds/dummy_structures.rb
ethanmills: We don't need to know about schools in production yet, and I'm not confident that it won't change before we do. I'm hoping to simplify any potential migrations by not having any school data. We can move this to the other seed file when we're ready

db/seeds/initial_seed.rb
kicferk1: Ech, it does indeed. Can we have a tech-debt ticket for it?

ethanmills: https://trello.com/c/IKKuN8dC/384-tech-debt-consider-merging-cohortsleadproviders-and-leadprovidercips-to-eliminate-duplication-of-data

db/seeds/dummy_structures.rb
kicferk1: Right, but we have a cron job that will import them every night, right? Am I missing something? Will we disable it?

db/seeds/initial_seed.rb
amazimbe: I remember not so long ago the cop was against having no underscores. I feel like they just dream up some of these cops

kicferk1: I would try to configure it to `snake_case`. Maybe not disabling it all-together. 

And seeing if that breaks things too much :P

db/seeds/dummy_structures.rb
ethanmills: You never triggered that to actually run...

ethanmills: As far as I can see, we never call [schedule](https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/models/cron_job_scheduler.rb#L8)

db/seeds/initial_seed.rb
ethanmills: Happy with this is @amazimbe is

db/seeds/dummy_structures.rb
kicferk1: Ooops, you are right 😂 I didn't follow the trail to the end...

I guess that means we don't need to disable it.

ethanmills: We just need to not enable it. Which is much easier

kicferk1: I mean, we need to remember not to enable it by accident when we add some other cron jobs, if we ever do

db/seeds/initial_seed.rb
amazimbe: Which file will we run for prod then?

ethanmills: We run `db:seed` everywhere, the environment determines what runs

kicferk1 (approved):

db/seeds/dummy_structures.rb
```diff
@@ -2,122 +2,96 @@
 
 SchoolDataImporter.new(Rails.logger).delay.run
 
-if Cohort.none?
-  Cohort.find_or_create_by!(start_year: 2021)
-  Cohort.find_or_create_by!(start_year: 2022)
+standard_user_emails = ["admin@example.com", "lead-provider@example.com", "second-school-leader@example.com", "school-leader@example.com"]
```
amazimbe, line 5: I know this won't run in prod but still why do we need it?

ethanmills: Because this runs on every review app so that we can log in

ethanmills: This is just moved, it was there before

db/seeds.rb
```diff
@@ -1,3 +1,7 @@
 # frozen_string_literal: true
 
-Dir[Rails.root.join("db/seeds/dummy_structures.rb")].each { |seed| load seed }
+Dir[Rails.root.join("db/seeds/initial_seed.rb")].each { |seed| load seed }
```
amazimbe, line 3: Way easier to just require / require_relative the files. It's not as if there's loads of them

ethanmills: I blame @kicferk1 😛

ethanmills: Would you like this changed, or are you happy to leave this?

amazimbe: Change it please.

seb-siem-softwire (approved):

amazimbe (approved):

PR #148, opened 2021-03-17
Cypress changes

callumacrae:
- [x] Make overly small tests bigger
- [x] Figure out how to make path testing less brittle
- [ ] ~~Maybe undo the pattern that seems to be happening of one-test-per-file~~ - will do in another PR as it's bigger and more controversial

### DIFF
--- a/spec/cypress/integration/admin/suppliers/edit-delivery-partner.js
+++ b/spec/cypress/integration/admin/suppliers/edit-delivery-partner.js
@@ -6,7 +6,7 @@ describe("Admin user editing delivery partner", () => {
     cy.login("admin");
   });
 
-  it("should update a delivery partner", () => {
+  it("should be able to edit a delivery partner", () => {
     cy.appScenario("admin/suppliers/manage_delivery_partner");
     const newName = "New delivery partner";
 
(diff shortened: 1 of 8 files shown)


kicferk1 (approved):
I'm surprised you cut out 80 lines (according to git), but small fixes add up ;)

I like extracting base paths to variables, hopefully it will make things easier to fix if we change high-level url structure.

spec/cypress/integration/admin/suppliers/edit-delivery-partner.js
```diff
@@ -1,67 +1,44 @@
...
-
-    cy.get("a").contains(deliveryPartnerName).click();
-
     cy.get(
       `[name='delivery_partner_form[lead_provider_ids][]'][value=${leadProviderId}]`
     ).uncheck();
     cy.clickCommitButton();
 
-    cy.get("a").contains(deliveryPartnerName).click();
+    cy.location("pathname").should("equal", basePath);
```
seb-siem-softwire, line 30: I do feel that cy.location("pathname") will be used a lot. Perhaps a helper method like current_path? Just a suggestion

callumacrae: I prefer `cy.url()` and incomplete matching - but this is an area that needs more discussion (we could also go with titles, once we have them)

ethanmills (approved):

PR #149, opened 2021-03-17
Extract support email to a config variable, use it in feedback url

kicferk1:
### Context
We might change support email in future. When that happens, we don't want to change it in many places in code.

1. Extract the support email to Rails config
2. Add a component printing it out
3. Made feedback link go to support email

### Guidance to review
I don't know if a component is the best way to do it - but I didn't like the idea of `Rails.config.support_email` floating around in our regular views, so I encapsulated it in a component.

And yes, I know we mention the email 4 times now - once in each environment - which is the same number of times as before. But it's a lot more scalable. 

Although now that I think about it... Should we look into making a shared config? Is that a good idea? Do we care? 

### Testing
I am not convinced there is much to test on this that doesn't amount to "does Rails set config variables correctly?".


kicferk1:
https://trello.com/c/gyFEiDH8/361-make-support-email-configurable

https://trello.com/c/xqHVg38u/321-update-feedback-link-in-phase-banner-with-support-email

kicferk1:
[Update feedback link in phase banner with support email](https://trello.com/c/xqHVg38u/321-update-feedback-link-in-phase-banner-with-support-email)

kicferk1:
[Make support email configurable](https://trello.com/c/gyFEiDH8/361-make-support-email-configurable)

ethanmills:
I think I'd like to see a Cypress test that loads the homepage and checks for a feedback link to the configured value

amazimbe (approved):
I just feel we could minimise the number of files to change by using an env variable though I understand that complicates things a bit. Anyway, your call.

spec/cypress/integration/feedback_email.js
```diff
@@ -0,0 +1,11 @@
+describe("Home page", () => {
```
callumacrae, line 1: I'd name the file home-page.js to match this

callumacrae (approved):

ethanmills (approved):

PR #150, opened 2021-03-17
ECF-RP-372: Add first user

ethanmills:
### Context
When we first release, we need to create the first user

- rake task to create first user
- update docs
- make rspec use configured domain in tests

### Testing
Unit test

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,18 +41,10 @@ or `db:seed`.
 ### Govuk Notify
 Register on [GOV.UK Notify](https://www.notifications.service.gov.uk). 
 Ask someone from the team to add you to our service.
-Generate an api key for yourself and set it in your `.env` file.
+Generate a limited api key for yourself and set it in your `.env` file.
 
 ### Set up git hooks
 Run `git config core.hooksPath .githooks` to use the included git hooks.
-
-## Whats included in this boilerplate?
-
-- Rails 6.0 with Webpacker
-- [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend)
-- RSpec
-- Dotenv (managing environment variables)
-- Travis with Heroku deployment
 
 ## Running specs, linter(without auto correct) and annotate models and serializers
 ```
@@ -69,11 +61,7 @@ bundle exec rspec
 It's best to lint just your app directories and not those belonging to the framework, e.g.
 
 ```bash
-bundle exec rubocop app config db lib spec Gemfile --format clang -a
-
-or
-
-bundle exec scss-lint app/webpacker/styles
+bundle exec rake lint:ruby
 ```
 
 ## End to end tests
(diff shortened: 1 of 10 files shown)


lib/tasks/manage_admins.rake
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+desc "Manage admin users"
+namespace :admin do
+  desc "Create admin user"
+  task :create, %i[full_name email] => :environment do |_task, args|
+    include Rails.application.routes.url_helpers
+    default_url_options[:host] = Rails.configuration.domain
+
+    puts "Creating user"
```
amazimbe, line 10: I would much prefer it if all this task did was pass the params to another object that does the actual creating along the same lines as the data imports. One option could be to define a CreateAdmin class that does what we are doing in here. That would be more reusable in the controllers too. I didn't do it when I worked on the controller coz the code was just being used once.

ethanmills:
[Add First user](https://trello.com/c/AtdoxU1V/372-add-first-user)

lib/tasks/manage_admins.rake
ethanmills: What about a static method on AdminProfile?

README.md
```diff
@@ -98,24 +86,24 @@ yarn cypress:open
...
-The app should be available at https://govuk-rails-boilerplate.london.cloudapps.digital
+Check the [documentation](./documentation/terraform.md) for detailed information on terraform.
+
+### Creating an initial admin user
+1. Follow the [debugging instructions](./documentation/debugging_in_govpaas.md) to gain SSH access to the instance and cd to the app dir
+2. Run `/usr/local/bin/bundle exec rake "admin:create[user name,email@example.com]"`. For example, the command for a user named `John Smith` with the email
+`john.smith@example.com` would be `/usr/local/bin/bundle exec rake "admin:create[John Smith,john.smith@example.com]"`. 
+   
+**The format here is important!
+   Notice that there are no extra spaces in the square brackets, and no quote marks inside the square brackets**
```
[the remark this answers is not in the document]
ethanmills: @cfq this one is for you

README.md
```diff
@@ -98,24 +86,24 @@ yarn cypress:open
...
+
+## Importing School data
+1. Make sure you have copied `.env.template` to `.env` and filled in the GIAS information with details from a teammate
+2. Run `bundle exec rake schools_data:import` to import the latest schools data from GIAS. This takes around 10 minutes
+3. Run `bundle exec rake sparsity:import` to populate the sparsity tables
+4. Run `bundle exec rake pupil_premium:import` to populate the pupil premium tables
```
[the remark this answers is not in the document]
ethanmills: Yep, when I get a chance I'll make one that calls all three

spec/rails_helper.rb
```diff
@@ -68,6 +68,9 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+  config.before(:each, type: :request) do
+    host! Rails.configuration.domain
+  end
```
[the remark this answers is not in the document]
ethanmills: Rails works out it's domain in controllers by looking at the request headers. This sets the rspec request headers to have the host specified in config

lib/tasks/manage_admins.rake
amazimbe: That would work too. Might be better actually

spec/tasks/manage_admins_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "rake admin:create", type: :task do
```
kicferk1, line 5: I don't think I like testing rake tasks explicitly. It requires plenty of setup, and at times it looks like half the thing this is testing is "did we set up our rake task testing correctly?".

I guess this test is checking that we didn't mess up the order of arguments and that we get the domain from Rails config correctly, which is something. 

That said, well, it's set up and seems to pass, so I guess since we put in the work for it already we might as well keep it.

kicferk1 (commented):
A moan about testing rake tasks, can be ignored. 

I still wanna know whether we want this task to be generic (so you can use it for many users) at the cost of having very specific instructions - or if we want it to use hard-coded values.

spec/models/admin_profile_spec.rb
```diff
@@ -4,4 +4,33 @@
 
 RSpec.describe AdminProfile, type: :model do
   it { is_expected.to belong_to(:user) }
+
+  describe "create_admin" do
```
amazimbe, line 8: I think the convention is to add a . in front of create_admin for class methods

amazimbe (approved):

README.md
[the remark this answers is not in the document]
cfq: No, let's keep it generic. I like it. Many cli apps work like this anyway 🤷‍♂️

cfq:
I'm happy for this task to be a generic one. We just need to be careful when running it. Could add a little bit of CLI application embellishment if you're keen like replaying the data back and asking if you're sure? (y/n) but happy for this to go as is. It's not a massive faff if anything goes wrong, just create another user and then delete all the erroneously entered users.

PR #152, opened 2021-03-17
[ECF-1-248] Persist logit setup in terraform

amazimbe:
### Context
We have a service set up in logit where we are piping our logs. Documentation. However, the configuration gets wiped out everytime the develop branch is deployed. Additionally, we need to replicate the setup for the staging and prod environments.

This pr stores the config changes in terraform so that we can automatically set up staging and prod as well as making the dev setup persistent.

I've used https://github.com/DFE-Digital/teacher-training-api/blob/7e7c91d55a18790bbb8d7fc0db6a88314441f97e/terraform/modules/paas/main.tf as a reference

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -119,7 +119,7 @@ jobs:
           export TF_VAR_paas_app_docker_image=${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ env.DOCKERHUB_ENV }}:${{ env.TAG }}
           cd terraform/app
           terraform init -reconfigure -input=false -backend-config="bucket=paas-s3-broker-prod-lon-7f2ca242-9929-4662-a79c-c454ea56ea7b"
-          terraform apply -input=false -auto-approve -var-file ../workspace-variables/dev.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}"}' -var='environment=${{github.event.inputs.environment}}' -var='secret_paas_app_env_values={"SYSLOG_DRAIN_URL":"${{secrets.SYSLOG_DRAIN_URL}}"}'
+          terraform apply -input=false -auto-approve -var-file ../workspace-variables/dev.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}"}' -var 'logstash_url=${{secrets.SYSLOG_DRAIN_URL}}'
 
       - name: Deploy review app to dev
         if: ${{ startsWith(github.event.inputs.environment, 'review') }}
@@ -132,7 +132,7 @@ jobs:
           export TF_VAR_paas_app_docker_image=${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ env.DOCKERHUB_ENV }}:${{ env.TAG }}
           cd terraform/app
           terraform init -reconfigure -input=false -backend-config="bucket=paas-s3-broker-prod-lon-7f2ca242-9929-4662-a79c-c454ea56ea7b" -backend-config="key=${{ github.event.inputs.environment }}/terraform.tfstate"
-          terraform apply -input=false -auto-approve -var-file ../workspace-variables/review.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}"}' -var='environment=${{github.event.inputs.environment}}' -var='secret_paas_app_env_values={"SYSLOG_DRAIN_URL":"${{secrets.SYSLOG_DRAIN_URL}}"}'
+          terraform apply -input=false -auto-approve -var-file ../workspace-variables/review.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}"}' -var='environment=${{github.event.inputs.environment}}' -var 'logstash_url=${{secrets.SYSLOG_DRAIN_URL}}'
 
 
       - name: Deploy to staging
@@ -146,5 +146,5 @@ jobs:
           export TF_VAR_paas_app_docker_image=${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ env.DOCKERHUB_ENV }}:${{ env.TAG }}
           cd terraform/app
           terraform init -reconfigure -input=false -backend-config="bucket=paas-s3-broker-prod-lon-e2123d0b-d394-4594-8056-315300e7d8a8"
-          terraform apply -input=false -auto-approve -var-file ../workspace-variables/staging.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_STAGING}}"}' -var='environment=${{github.event.inputs.environment}}' -var='secret_paas_app_env_values={"SYSLOG_DRAIN_URL":"${{secrets.SYSLOG_DRAIN_URL}}"}'
+          terraform apply -input=false -auto-approve -var-file ../workspace-variables/staging.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_STAGING}}"}' -var 'logstash_url=${{secrets.SYSLOG_DRAIN_URL}}'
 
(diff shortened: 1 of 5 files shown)


terraform/app/modules/paas/main.tf
```diff
@@ -35,3 +40,9 @@ resource cloudfoundry_route web_app_route {
   space    = data.cloudfoundry_space.space.id
   hostname = local.web_app_name
 }
+
+resource cloudfoundry_user_provided_service logging {
+  name             = local.logging_service_name
+  space            = data.cloudfoundry_space.space.id
+  syslog_drain_url = var.logstash_url
```
ethanmills, line 47: Where is this being set?

amazimbe: Good question. I need to set that up in terraform.tf

terraform/app/modules/paas/main.tf
```diff
@@ -18,6 +18,11 @@ resource cloudfoundry_app web_app {
   stopped = var.app_stopped
   strategy = var.web_app_deployment_strategy
   timeout = var.app_start_timeout
+
+  service_binding {
+    service_instance = cloudfoundry_user_provided_service.logging.id
+  }
+
```
kicferk1, line 25: We have dynamic service binding below. I wonder if they could be put together.

terraform/app/modules/paas/main.tf
ethanmills: And also supply it from a github secret in .github/workflows/deploy.yml

ethanmills: Much like RAILS_MASTER_KEY

terraform/app/modules/paas/main.tf
ethanmills: This would involve adding `cloudfoundry_user_provided_service.logging.id` to `app_cloudfoundry_service_instances` on line 54 of terraform/app/modules/paas/variables.tf, and removing this service_binding

amazimbe:
[Set up exporting logs to logit](https://trello.com/c/xYb7RSeG/248-set-up-exporting-logs-to-logit)

.github/workflows/deploy.yml
```diff
@@ -119,7 +119,7 @@ jobs:
           export TF_VAR_paas_app_docker_image=${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ env.DOCKERHUB_ENV }}:${{ env.TAG }}
           cd terraform/app
           terraform init -reconfigure -input=false -backend-config="bucket=paas-s3-broker-prod-lon-7f2ca242-9929-4662-a79c-c454ea56ea7b"
-          terraform apply -input=false -auto-approve -var-file ../workspace-variables/dev.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}"}'
+          terraform apply -input=false -auto-approve -var-file ../workspace-variables/dev.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}"}' -var='environment=${{github.event.inputs.environment}}' -var='secret_paas_app_env_values={"SYSLOG_DRAIN_URL":"${{secrets.SYSLOG_DRAIN_URL}}"}'
```
kicferk1, line 122: I think we already have `secret_paas_app_env_values` field that we pass to terraform - no need to add another one, just use the existing one (with RAILS_MASTER_KEY).

With this, you get a failed deployment with 
```
Error: Reference to undeclared input variable

  on terraform.tf line 47, in module "paas":
  47:   logstash_url                      = var.SYSLOG_DRAIN_URL

An input variable with the name "SYSLOG_DRAIN_URL" has not been declared. This
variable can be declared with a variable "SYSLOG_DRAIN_URL" {} block.
```

I imagine because when terraform gets two variables with the same name in options (like you do here) it takes the first one?

terraform/app/modules/paas/main.tf
kicferk1: I think with the current setup, you pass it to `secret_paas_app_env_values`, which gets merged into `paas_app_env_values`, which is stored in `app_env_values`, which I _think_ is already a variable that you could use.

So in here, you could have `var.app_env_values. SYSLOG_DRAIN_URL`. 

If you prefer having a variable for it, then when you assign it in another file, I think that's what you should refer to for it to work - then again, deploying PR app is probably the best test for that.

.github/workflows/deploy.yml
amazimbe: @kicferk1 where do you see that error log? I couldn't see it on Github

terraform/app/modules/paas/main.tf
ethanmills: This is not an app env var though, so it would seem strange to put it there. This is for terraform to use directly in setting up PaaS

.github/workflows/deploy.yml
kicferk1: I got an email pointing me to https://github.com/DFE-Digital/early-careers-framework/actions/runs/663581993

terraform/app/terraform.tf
```diff
@@ -44,4 +44,5 @@ module paas {
   web_app_instances                 = var.paas_web_app_instances
   web_app_memory                    = var.paas_web_app_memory
   web_app_start_command             = var.paas_web_app_start_command
+  logstash_url                      = var.SYSLOG_DRAIN_URL
```
ethanmills, line 47: Let's keep this lowercase, like the others

ethanmills: This needs to be declared in terraform/app/variables.tf

.github/workflows/deploy.yml
ethanmills, line 122: We should pass in a `var syslog_drain_url='=${{secrets.SYSLOG_DRAIN_URL}}'`, although maybe it should be called `logstash_url`

ethanmills: We don't want to sent the environment here, that was omitted deliberately

.github/workflows/deploy.yml
```diff
@@ -145,4 +146,5 @@ jobs:
           export TF_VAR_paas_app_docker_image=${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ env.DOCKERHUB_ENV }}:${{ env.TAG }}
           cd terraform/app
           terraform init -reconfigure -input=false -backend-config="bucket=paas-s3-broker-prod-lon-e2123d0b-d394-4594-8056-315300e7d8a8"
-          terraform apply -input=false -auto-approve -var-file ../workspace-variables/staging.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_STAGING}}"}'
+          terraform apply -input=false -auto-approve -var-file ../workspace-variables/staging.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_STAGING}}"}' -var='environment=${{github.event.inputs.environment}}' -var='secret_paas_app_env_values={"SYSLOG_DRAIN_URL":"${{secrets.SYSLOG_DRAIN_URL}}"}'
```
ethanmills, line 149: Please don't set `environment` here

.github/workflows/deploy.yml
```diff
@@ -132,7 +132,8 @@ jobs:
           export TF_VAR_paas_app_docker_image=${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ env.DOCKERHUB_ENV }}:${{ env.TAG }}
           cd terraform/app
           terraform init -reconfigure -input=false -backend-config="bucket=paas-s3-broker-prod-lon-7f2ca242-9929-4662-a79c-c454ea56ea7b" -backend-config="key=${{ github.event.inputs.environment }}/terraform.tfstate"
-          terraform apply -input=false -auto-approve -var-file ../workspace-variables/review.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}"}' -var='environment=${{github.event.inputs.environment}}'
+          terraform apply -input=false -auto-approve -var-file ../workspace-variables/review.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}"}' -var='environment=${{github.event.inputs.environment}}' -var 'logstash_url=${{secrets.SYSLOG_DRAIN_URL}}'
+
```
kicferk1, line 136: Hehe, you know who hates whitespace? That's right, I do 😂

kicferk1 (approved):
Looks good to me - I hope it works.

Do we have a single url for all environments? I imagine yes, but worth making sure.

ethanmills (approved):
Looks good, seeing logs come through!

.github/workflows/deploy.yml
amazimbe: I'll take it out. You are not the only one to hate white space. At my first company we didn't accept a pr with white space changes at all.

amazimbe:
> Looks good to me - I hope it works.
> 
> Do we have a single url for all environments? I imagine yes, but worth making sure.

Yes, we have a single url for all environments. Yes, it works. I can see the logs

PR #153, opened 2021-03-18
Adds health check functionality to the application

seb-siem-softwire:
Adds configuration for database, migration, and email

### Context

We would like to get a health check for our application at a minimum for email, database, and migrations

Adds configuration for health_check gem that allows us to see if there are any errors/issues

### Testing

Added basic unit test

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,9 @@ gem "devise", ">= 4.7.3"
 gem "kaminari", ">= 1.2.0"
 gem "pundit"
 
+# Adds health check functionality
+gem "health_check", ">= 3.0.0"
+
 # Use postgresql as the database for Active Record
 gem "pg", ">= 0.18", "< 2.0"
 
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem "kaminari", ">= 1.2.0"
 gem "pundit"
 
 # Adds health check functionality
-gem "health_check", ">= 3.0.0"
+gem "health_check", github: "/ianheggie/health_check", ref: "0b799ead604f900ed50685e9b2d469cd2befba5b"
 
 # Use postgresql as the database for Active Record
 gem "pg", ">= 0.18", "< 2.0"
--- a/config/initializers/health_check.rb
+++ b/config/initializers/health_check.rb
@@ -2,13 +2,13 @@
 
 HealthCheck.setup do |config|
   # uri prefix (no leading slash)
-  config.uri = "health_check"
+  config.uri = "healthcheck"
 
   # Text output upon success
   config.success = "success for following checks: ['database', 'migrations', 'email']"
 
   # Timeout in seconds used when checking smtp server
-  config.smtp_timeout = 30.0
+  # config.smtp_timeout = 30.0
 
   # http status code used when plain text error message is output
   # Set to 200 if you want your want to distinguish between partial (text does not include success) and
--- a/config/initializers/health_check.rb
+++ b/config/initializers/health_check.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+HealthCheck.setup do |config|
+  # uri prefix (no leading slash)
+  config.uri = "health_check"
+
+  # Text output upon success
+  config.success = "success for following checks: ['database', 'migrations', 'email']"
+
+  # Timeout in seconds used when checking smtp server
+  config.smtp_timeout = 30.0
+
+  # http status code used when plain text error message is output
+  # Set to 200 if you want your want to distinguish between partial (text does not include success) and
+  # total failure of rails application (http status of 500 etc)
+
+  config.http_status_for_error_text = 500
+
+  # http status code used when an error object is output (json or xml)
+  # Set to 200 if you want your want to distinguish between partial (healthy property == false) and
+  # total failure of rails application (http status of 500 etc)
+
+  config.http_status_for_error_object = 500
+
+  # You can customize which checks happen on a standard health check, eg to set an explicit list use:
+  config.standard_checks = %w[database migrations email]
+
+  # Or to exclude one check:
+  # config.standard_checks -= [ 'emailconf' ]
+
+  # You can set what tests are run with the 'full' or 'all' parameter
+  config.full_checks = %w[database migrations email]
+
+  # Add one or more custom checks that return a blank string if ok, or an error message if there is an error
+  # config.add_custom_check do
+  #   CustomHealthCheck.perform_check # any code that returns blank on success and non blank string upon failure
+  # end
+
+  # Add another custom check with a name, so you can call just specific custom checks. This can also be run using
+  # the standard 'custom' check.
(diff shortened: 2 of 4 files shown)


spec/requests/health_check_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+RSpec.describe "Health Check", type: :request do
+  describe "GET /health_check" do
+    it "returns success message for current health checks" do
+      get "/health_check"
+
+      expected = "success for following checks: ['database', 'migrations', 'email']"
+      expect(response.body).to eq(expected)
+    end
+  end
+end
```
kicferk1, line 14: Can we have tests for when things are broken?

config/initializers/health_check.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  # http status code used when the ip is not allowed for the request
+  # config.http_status_for_ip_whitelist_error = 403
+
+  # When redis url is non-standard
+  # config.redis_url = 'redis_url'
+
+  # Disable the error message to prevent /health_check from leaking
+  # sensitive information
+  # config.include_error_in_response_body = false
```
kicferk1, line 69: I hope that sensitive information bit is not included in the default checks that we are performing. But I qould like to double check that.

config/initializers/health_check.rb
```diff
@@ -0,0 +1,70 @@
...
+  # max-age of response in seconds
+  # cache-control is public when max_age > 1 and basic_auth_username is not set
+  # You can force private without authentication for longer max_age by
+  # setting basic_auth_username but not basic_auth_password
+  config.max_age = 1
+
+  # Protect health endpoints with basic auth
+  # These default to nil and the endpoint is not protected
+  # config.basic_auth_username = 'my_username'
+  # config.basic_auth_password = 'my_password'
```
kicferk1, line 55: Do we care that this is on public internet? I guess we don't really, cause everything else is on public internet too. 

Do health-check requests show up in our logs? So that if someone tries a ddos through health-checks we will see it in our logs?

config/initializers/health_check.rb
```diff
@@ -0,0 +1,70 @@
...
+
+HealthCheck.setup do |config|
+  # uri prefix (no leading slash)
+  config.uri = "health_check"
+
+  # Text output upon success
+  config.success = "success for following checks: ['database', 'migrations', 'email']"
+
+  # Timeout in seconds used when checking smtp server
+  config.smtp_timeout = 30.0
```
kicferk1, line 11: Huh, that looks like a lot. We sure 30s is how long we want to let it run?

kicferk1 (commented):
A few questions around the ins and outs of config

config/initializers/health_check.rb
amazimbe: Perhaps we could remove this comment as it reveals the password and username.

kicferk1: Uhhh, I don't think this comment reveals much. My local username and password are different!

config/initializers/health_check.rb
kicferk1: Good point...

What is the email check doing exactly?

amazimbe: I don't think we do. Notify takes care of it all.

config/initializers/health_check.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+HealthCheck.setup do |config|
+  # uri prefix (no leading slash)
+  config.uri = "health_check"
```
amazimbe, line 5: We'd usually want this dasherised rather than snake cased. But this is ok as this isn't quite user facing.

config/initializers/health_check.rb
seb-siem-softwire: This is from basic configuration and set as an example - personally - I'm not sure about passwords - from one side - yes it feels more secure - however Denial of Service attack could simply attack other pages - plus message stating everything is ok is of very little value. The question then becomes what should we reveal when things are not ok and are we happy with the outcome.

config/initializers/health_check.rb
seb-siem-softwire: I think I will change it to one word - Matt gave me an example of other services which don't use hyphens and use the word healthcheck https://pp-services.signin.education.gov.uk/healthcheck

config/initializers/health_check.rb
kicferk1: > Denial of Service attack could simply attack other pages

I agree. Hence the question: do the healthcheck requests show up in our logs, like all other pages do (I think)?

config/initializers/health_check.rb
seb-siem-softwire: @ethanmills should or should we not include sensitive information in response body?

ethanmills: We should not. They are public

seb-siem-softwire: Done

config/initializers/health_check.rb
seb-siem-softwire: Yes the healthcheck does show up in our logs like any other request

config/initializers/health_check.rb
seb-siem-softwire: Removed

config/initializers/health_check.rb
seb-siem-softwire: changed to healthcheck

Gemfile
```diff
@@ -13,7 +13,7 @@ gem "kaminari", ">= 1.2.0"
 gem "pundit"
 
 # Adds health check functionality
-gem "health_check", ">= 3.0.0"
+gem "health_check", github: "/ianheggie/health_check", ref: "0b799ead604f900ed50685e9b2d469cd2befba5b"
```
kicferk1, line 16: Wait, so where did the file with the comments come from, if not from the 'official' release?

seb-siem-softwire: This was utterly irritating - the option to not include the sensitive information was not in 3.0.0 but in edge release. So the readme file on that repo is inaccurate - if you use 3.0.0 version you will not get the configuration to disable sensitive info. Had to compare the two files side by side

amazimbe (approved):

spec/requests/health_check_spec.rb
seb-siem-softwire: Sure - I'll add those

config/initializers/health_check.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+HealthCheck.setup do |config|
+  # uri prefix (no leading slash)
+  config.uri = "healthcheck"
+
+  # Text output upon success
+  config.success = "success for following checks: ['database', 'migrations', 'email']"
```
kicferk1, line 8: So I get what database and migrations checks do (or can do, more or less). 

What does email check do?

seb-siem-softwire: I don't think we should be checking email as we are using GOV Notify service

PR #154, opened 2021-03-18
ECF-RP-370: Update app and db params for staging

ethanmills:
### Context
We want staging to be as similar to prod as possible. These are the values we will start with in prod. It is especially important to change the DB now, since it is hard to upgrade from an unencrypted db to an encrypted one

### DIFF
--- a/terraform/workspace-variables/staging.tfvars
+++ b/terraform/workspace-variables/staging.tfvars
@@ -5,9 +5,9 @@ app_environment = "staging"
 # Gov.UK PaaS
 paas_api_url = "https://api.london.cloud.service.gov.uk"
 paas_space_name = "early-careers-framework-staging"
-paas_postgres_service_plan = "tiny-unencrypted-11"
+paas_postgres_service_plan = "small-11"
 paas_app_start_timeout = "180"
 paas_app_stopped = false
 paas_web_app_deployment_strategy = "blue-green-v2"
 paas_web_app_instances = 1
-paas_web_app_memory = 512
+paas_web_app_memory = 8192


terraform/workspace-variables/staging.tfvars
```diff
@@ -5,9 +5,9 @@ app_environment = "staging"
...
 paas_space_name = "early-careers-framework-staging"
-paas_postgres_service_plan = "tiny-unencrypted-11"
+paas_postgres_service_plan = "small-11"
 paas_app_start_timeout = "180"
 paas_app_stopped = false
 paas_web_app_deployment_strategy = "blue-green-v2"
 paas_web_app_instances = 1
-paas_web_app_memory = 512
\ No newline at end of file
+paas_web_app_memory = 8192
```
amazimbe, line 13: What are these based on? Or did you just pick what you deemed suitable?

ethanmills: Asked @cfq. I think JC picked a number months ago

kicferk1 (approved):
I guess this has been talked about and approved, right? As in, which sizes we are going for?

terraform/workspace-variables/staging.tfvars
ethanmills: Are you just curious, or do you disagree with the number?

ethanmills:
Yep, Mehmet gave these as initial values, then we monitor and adjust

terraform/workspace-variables/staging.tfvars
amazimbe: Just curious

amazimbe: Will we have just one web server then?

ethanmills: I think at least initially, yes. We're only going to have 2 users for the next month... Hopefully ramp up will be quite gradual, so we can be reactive to load. It's something that's easy to change, so I think we're ok to start small

terraform/workspace-variables/staging.tfvars
```diff
@@ -5,9 +5,9 @@ app_environment = "staging"
 # Gov.UK PaaS
 paas_api_url = "https://api.london.cloud.service.gov.uk"
 paas_space_name = "early-careers-framework-staging"
-paas_postgres_service_plan = "tiny-unencrypted-11"
+paas_postgres_service_plan = "small-11"
```
ethanmills, line 8: For reference:

```
small-11                        100GB Storage, Dedicated Instance, Storage Encrypted, Max 200 Concurrent Connections. Postgres Version 11. DB Instance Class: db.t3.small.
```

amazimbe (approved):

PR #155, opened 2021-03-18
ECF-RP-371: release prod

ethanmills:
## Proposed release mechanism
All releases to staging and production would be as a result of a manual starting of a Github action. 

Cutting a release to staging would be as follows:
- User specifies ref to deploy and version number
- Addition branch protection on master is temporarily disabled (These anyone from prevent pushing to master)
- Master is reset to the specified ref
- The release is tagged with the specified version
- Branch protections are re-enabled
- Deployment of the specified ref is triggered

Promoting a release to production would be as follows:
- User specifies the tagged version to deploy to production
- Deployment of the specified tag is triggered (terraform apply)

## Proposed changes to build mechanism
- Remove call to separate build/deploy action - inline everything
- Images only built for review & dev - dev images promoted through staging and production
- Use dev dockerhub repo for intermediate build images and review builds
- Use prod dockerhub repo for dev builds (promoted through environments)
- Retag docker image with version when releasing to staging
- Display release version at /check endpoint


.github/workflows/deploy.yml
```diff
@@ -147,3 +147,16 @@ jobs:
...
+        if: ${{ github.event.inputs.environment == 'production' }}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID_PRODUCTION }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY_PRODUCTION }}
+          TF_VAR_paas_user: ${{ secrets.GOVPAAS_PROD_USERNAME }}
+          TF_VAR_paas_password: ${{ secrets.GOVPAAS_PROD_PASSWORD }}
+        run: |
+          export TF_VAR_paas_app_docker_image=${{ env.DOCKERHUB_REPOSITORY_BASE }}${{ env.DOCKERHUB_ENV }}:${{ env.TAG }}
+          cd terraform/app
+          terraform init -reconfigure -input=false -backend-config="bucket=paas-s3-broker-prod-lon-037149fa-bf9a-4577-a066-f627d277d6c4"
```
kicferk1, line 161: Nice fresh bucket, well done to remember it :D

terraform/workspace-variables/staging.tfvars
```diff
@@ -5,9 +5,9 @@ app_environment = "staging"
 # Gov.UK PaaS
 paas_api_url = "https://api.london.cloud.service.gov.uk"
 paas_space_name = "early-careers-framework-staging"
-paas_postgres_service_plan = "tiny-unencrypted-11"
+paas_postgres_service_plan = "small-11"
```
kicferk1, line 8: Needs a rebase I think

.github/workflows/deploy.yml
ethanmills: It's so well documented 😛

.github/workflows/deploy_to_production.yml
```diff
@@ -0,0 +1,41 @@
...
+          echo "INPUTS=${INPUTS}" >> $GITHUB_ENV
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: trigger-deploy
+        uses: benc-uk/workflow-dispatch@v1.1
+        with:
+          workflow: Deploy to Environment # Workflow name, in deploy.yml
+          token: ${{ secrets.PERSONAL_TOKEN }}
```
amazimbe, line 26: Curious what this is for.

ethanmills: Same as the deploy to every other environment. Used for permissions to trigger a workflow: [docs](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token:~:text=If%20you%20need%20a%20token%20that%20requires%20permissions%20that%20aren't%20available)

kicferk1: This is so I get an email notification when you open a production release at 2am on a Saturday

.github/workflows/deploy_to_staging.yml
```diff
@@ -1,41 +1,111 @@
...
+          allow_force_pushes: |
+            true
+          required_linear_history: |
+            true
+          allow_deletions: |
+            null
+          restrictions: |
+            null
+        env:
+          GITHUB_TOKEN: ${{ secrets.PERSONAL_TOKEN }}
```
kicferk1, line 43: Rather than doing it in actions, can we set some protections on `master` branch of our repo that admins don't care about, and use an admin token to bypass them in this action?

kicferk1: (I really don't like the idea of unprotecting a branch so we can make it look prettier on a release :P)

ethanmills: No, because we're all admins. This is the accepted way of doing it

ethanmills: I spent some time looking, and found no combination of branch protection rules that do what we want

ethanmills: Someone has even written a bot for it (which I'm not a big fan of) https://github.com/benjefferies/branch-protection-bot

ethanmills: So what we're doing here is disabling the extra protections on master - the ones you're advocating for disabling permanently. This branch will still only be writable by people with write permissions on the repo

ethanmills: It becomes the same as any other branch for a short time

callumacrae: I can't comment on the rest of the PR but this specifically seems like a sensible approach to me

kicferk1: Oh, so `null` in the workflow variables means 'do not change this'? I guess that makes sense then.

kicferk1 (approved):

.github/workflows/deploy_to_staging.yml
ethanmills: See https://docs.github.com/en/rest/reference/repos#update-branch-protection

ethanmills: Discussed offline. This is changing branch protection rules, even when they're all off this branch is as "protected" as any other, only team members have push access

amazimbe (approved):

seb-siem-softwire (approved):

amazimbe (commented):
Just wondering why there seems to be more changes than I expected. Other than that it looks good.

.github/workflows/deploy_to_staging.yml
```diff
@@ -1,45 +1,128 @@
 name: Deploy to Staging
 
 on:
-  push:
-    branches:
-      - master
-    paths-ignore:
-      - 'documentation/**'
-
+  workflow_dispatch:
```
amazimbe, line 4: Why have we changed this file?

ethanmills: > Just wondering why there seems to be more changes than I expected. Other than that it looks good.

@amazimbe the recent rails issue highlighted some issues: because we were rebuilding each time, we couldn't even deploy old versions to staging. This way, we build once and promote though, which is  best practice anyway

amazimbe (approved):

PR #158, opened 2021-03-18
Don't check the yarn.lock file for AWS keys

callumacrae:
### Context

There are loads of AWS keys in the yarn.lock file, according to the old version of this script. This stops that script from checking the yarn.lock file.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -14,10 +14,16 @@ fi
 exec 1>&2
 
 # Check changed files for an AWS keys
-KEY_ID=$(git diff --cached --name-only -z $against | xargs -0 cat | perl -nle'print $& if m{(?<![A-Z0-9])[A-Z0-9]{20}(?![A-Z0-9])}')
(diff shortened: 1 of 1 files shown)


.githooks/pre-commit
```diff
@@ -14,10 +14,16 @@ fi
...
 # Check changed files for an AWS keys
-KEY_ID=$(git diff --cached --name-only -z $against | xargs -0 cat | perl -nle'print $& if m{(?<![A-Z0-9])[A-Z0-9]{20}(?![A-Z0-9])}')
-KEY=$(git diff --cached --name-only -z $against | xargs -0 cat | perl -nle'print $& if m{(?<![A-Z0-9])[A-Z0-9]{20}(?![A-Z0-9])}')
-
-if [ "$KEY_ID" != "" -o "$KEY" != "" ]; then
+KEY=$(
+    git diff --cached --name-only -z HEAD~2 |
+        tr '\0' '\n' |
+        grep --invert-match --fixed-strings 'yarn.lock' |
+        xargs ls -d 2> /dev/null |
```
callumacrae, line 21: For some reason xargs when ran with null characters as separators ignores missing files, but it doesn't when files are piped in individually, so this line filters out missing files

This is the only line i copied off stack overflow 👀

kicferk1 (approved):
At least those keys aren't ours, right? :D

PR #159, opened 2021-03-19
Call confirm inside the transaction

ethanmills:
### Context
When email sending fails, you currently end up with an orphaned user with no profiles 

Confirm calls save, so this should be inside the transaction

### DIFF
--- a/app/models/admin_profile.rb
+++ b/app/models/admin_profile.rb
@@ -5,9 +5,9 @@ class AdminProfile < BaseProfile
 
   def self.create_admin(full_name, email, sign_in_url)
     user = User.new(full_name: full_name, email: email)
-    user.confirm
 
     ActiveRecord::Base.transaction do
+      user.confirm
       user.save!
       AdminProfile.create!(user: user)
       AdminMailer.account_created_email(user, sign_in_url).deliver_now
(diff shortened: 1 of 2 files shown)


kicferk1 (commented):
If confirm calls save do we need to call save explicitly?

Well, for readability's and sanity's sake maybe yes :) 

Another trick would be to just append confirmed_at: Time.now to the new params rather than calling confirm. That's all confirm does but that implementation could change in  future.

seb-siem-softwire (approved):

app/models/admin_profile.rb
```diff
@@ -5,9 +5,9 @@ class AdminProfile < BaseProfile
 
   def self.create_admin(full_name, email, sign_in_url)
     user = User.new(full_name: full_name, email: email)
-    user.confirm
 
     ActiveRecord::Base.transaction do
+      user.confirm
```
amazimbe, line 10: The unit test needs to cover this confirmation as well

amazimbe, line 10: Just curious if this isn't actually calling save. Most likely it is.

amazimbe (approved):

app/models/admin_profile.rb
amazimbe: My other comment which I deleted was about checking this confirmed in the unit test. We missed that in the first pr but it's small enough to add here

kicferk1 (approved):

PR #160, opened 2021-03-19
Small changes

ethanmills:
### Context
Little things that don't quite look right on staging

- Sort admins alphabetically
- Update phase banner tag

### DIFF
--- a/app/controllers/admin/administrators/administrators_controller.rb
+++ b/app/controllers/admin/administrators/administrators_controller.rb
@@ -8,7 +8,7 @@ module Admin
       before_action :load_admin, only: %i[edit update delete destroy]
 
       def index
-        @administrators = policy_scope(User).admins
+        @administrators = policy_scope(User).admins.order(:full_name)
       end
 
       def new
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -63,7 +63,13 @@
       <div class="govuk-phase-banner app-phase-banner--no-border govuk-!-display-none-print">
         <p class="govuk-phase-banner__content">
           <strong class="govuk-tag govuk-phase-banner__content__tag ">
-            Development
+            <% if Rails.env.production?  %>
+              Beta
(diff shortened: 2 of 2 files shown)


app/controllers/admin/administrators/administrators_controller.rb
```diff
@@ -8,7 +8,7 @@ class AdministratorsController < Admin::BaseController
       before_action :load_admin, only: %i[edit update delete destroy]
 
       def index
-        @administrators = policy_scope(User).admins
+        @administrators = policy_scope(User).admins.order(:full_name)
```
callumacrae, line 11: Is this case insensitive? Ideally it would be

kicferk1 (approved):

app/controllers/admin/administrators/administrators_controller.rb
amazimbe: Yes, it's a db search and the default is case insensitive. I find it helpful to specify asc or desc regardless of the default

amazimbe (approved):

app/controllers/admin/administrators/administrators_controller.rb
ethanmills: Nope. How much do we care? It would become `order("LOWER(full_name)")` I think

callumacrae: I don't think it matters _that_ much as hopefully people will be casing names correctly, but the lead providers page for example sorts "z" before "A" which is kinda weird

ethanmills: Sorting on postgres is a bit of a minefield: https://dba.stackexchange.com/questions/106964/why-is-my-postgresql-order-by-case-insensitive
I'll leave it like this for now, and if people are putting in silly data we can always change it 🙂

amazimbe: Or we could capitalize the name

PR #164, opened 2021-03-19
Add lograge

ethanmills:
### Context
Rails currently gives three separate logs per request. This makes life very difficult. Lograge makes life easier again. 
Also both Teaching Vacancies AND Apply are doing it, so it must be a good idea!

### Guidance to review
Take a look at logit if you're curious, I recommend the following filter: `cf.app:"ecf-review-pr-164" and not cf_tags.source_type:"RTR"`

I've excluded /check logs, which I think is a good idea. Do you disagree?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,12 +8,19 @@ ruby File.read(".ruby-version").chomp
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem "rails", "~> 6.1.2", ">= 6.1.2.1"
 
+# User management and rbac
 gem "devise", ">= 4.7.3"
-gem "kaminari", ">= 1.2.0"
 gem "pundit"
+
+# Pagination
+gem "kaminari", ">= 1.2.0"
 
 # Adds health check functionality
 gem "health_check", github: "/ianheggie/health_check", ref: "0b799ead604f900ed50685e9b2d469cd2befba5b"
+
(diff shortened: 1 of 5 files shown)


kicferk1 (approved):
Soo, what are the differences in different environments? Just log level and whether we log sql? 

Round 2 of "wo de want some default environment config to commonize those things? (Not necessarily here, just in general)

Also, do we have a space somewhere to put some useful log queries? Like in our documentation, readme, confluence?

ethanmills:
> Soo, what are the differences in different environments? Just log level and whether we log sql?

Deployed envs, yep pretty much

> Round 2 of "wo de want some default environment config to commonize those things? (Not necessarily here, just in general)

Quite possibly, Apply have something interesting for that

> Also, do we have a space somewhere to put some useful log queries? Like in our documentation, readme, confluence?

Not yet, I don't think we really know what's interesting yet

amazimbe:
> > Do want some default environment config to commonize those things? (Not necessarily here, just in general)

In the past, I have always added common configs to environment.rb. However, the way our environment.rb looks seem to convey that the bolerplate did not recommend this. Are there any gov guidelines on this? What have other DfE projects done?

amazimbe (approved):
LGTM! We may need to discuss about where to put common configs but that doesn't need to block this.

PR #166, opened 2021-03-19
Convert delivery partner tests to cucumber specs

callumacrae:
- [x] Create
- [x] Edit
- [x] Delete


spec/cypress/support/step_definitions/common-interaction.js
```diff
@@ -10,6 +10,12 @@ const elements = {
   "notification banner": "[data-test=notification-banner]",
   "edit admin link": "[data-test=edit-admin-link]",
   "delete button": "[data-test=delete-button]",
+  "create delivery partner button":
+    '.govuk-button:contains("new delivery partner")',
+  "delivery partner name": 'input[type="text"]',
+  "change name link": 'a:contains("Change name")',
+  "name input": '[name*="name"]',
+  "back button": '.govuk-button:contains("Back")',
```
ethanmills, line 18: Are you not concerned this list is going to get unmanageable?

ethanmills (commented):
Couple of little things. Also, tests are failing 🙁

spec/cypress/support/step_definitions/common-interaction.js
callumacrae: i am concerned yes, but I also can't think of anything better at the moment

app/views/admin/suppliers/delivery_partners/edit.html.erb
```diff
@@ -11,7 +11,7 @@
       <%= render "lead_providers_cohorts", locals = { f: f, legend_options: { size: "m" } } %>
       <div class="govuk-button-group">
         <%= f.govuk_submit "Save" %>
-        <%= govuk_link_to "Delete", delete_admin_delivery_partner_path(@delivery_partner), button: true, class: "govuk-button--warning" %>
+        <%= govuk_link_to "Delete", delete_admin_delivery_partner_path(@delivery_partner), button: true, class: "govuk-button--warning", "data-test": "delete-button" %>
```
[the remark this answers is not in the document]
callumacrae: I'm not sure we'd always want the same data attribute or even text - for example, the other delete button in this PR could easily say "Confirm deletion". Worth making a helper just to add three attributes?

spec/cypress/integration/admin/DeliveryPartnerManagement.feature
```diff
@@ -0,0 +1,112 @@
...
+  Scenario: It should allow changing name choice during creation
+    When I click on "create delivery partner button"
+    And I type "wrong name" into "delivery partner name" field
+    And I click the submit button
+    And I click on "Lead Provider 1" label
+    And I click the submit button
+    And I click on "2021" label
+    And I click the submit button
+    And I click on "change name link"
+    And I type "{selectall}New delivery partner" into "delivery partner name" field
```
kicferk1, line 65: > I type "{selectall}New delivery partner"

I don't expect to be typing `{selectall}` into any field. What does it do? And why?

spec/cypress/integration/admin/DeliveryPartnerManagement.feature
```diff
@@ -0,0 +1,112 @@
...
+    Then I should be on "delivery partner index" page
+    And "main" should contain "New delivery partner"
+
+  Scenario: Admins should be able to edit delivery partners
+    When I click on "a" containing "Delivery Partner 1"
+    Then I should be on "delivery partner edit" page
+    And "name input" should have value "Delivery Partner 1"
+    # And the page should be accessible
+
+    When I type "{selectall}New delivery partner" into "name input" field
```
kicferk1, line 81: Ditto with selectall - can we remove all of its occurrences?

spec/cypress/integration/admin/DeliveryPartnerManagement.feature
```diff
@@ -0,0 +1,112 @@
...
+    And I click on "Lead Provider 1" label
+    And I click the submit button
+    Then I should be on "delivery partner index" page
+    And "main" should contain "New delivery partner"
+    And "main" should not contain "Delivery Partner 1"
+
+    When I click on "a" containing "New delivery partner"
```
kicferk1, line 88: When I click on `a link`?

spec/cypress/support/step_definitions/common-interaction.js
kicferk1: We could split this object into a few more specific ones - input fields, buttons, text display elements. 

That would also help enforce naming conventions. At the moment we have a `name` which is an input for `full_name` and a `name_input` which is an input for `name`...

spec/cypress/support/step_definitions/common-navigation.js
```diff
@@ -3,14 +3,24 @@ import { Given, When, Then } from "cypress-cucumber-preprocessor/steps";
 Given("I am logged in as a {string}", (user) => cy.login(user));
 Given("I am logged in as an {string}", (user) => cy.login(user));
 
-Given("scenario {string} has been ran", (scenario) => cy.appScenario(scenario));
+Given("scenario {string} has been run", (scenario) => cy.appScenario(scenario));
```
kicferk1, line 6: Nice

spec/cypress/support/step_definitions/common-navigation.js
```diff
@@ -3,14 +3,24 @@ import { Given, When, Then } from "cypress-cucumber-preprocessor/steps";
...
+  "delivery partner index": "/admin/suppliers",
+  "choose new delivery partner name":
+    "/admin/suppliers/new/delivery-partner/choose-name",
+  "choose new delivery partner lead providers":
+    "/admin/suppliers/new/delivery-partner/choose-lps",
+  "choose delivery partner cohorts":
+    "/admin/suppliers/new/delivery-partner/choose-cohorts",
+  "new delivery partner review": "/admin/suppliers/new/delivery-partner/review",
+  "delivery partner edit": /\/delivery-partners\/.*\/edit/,
+  "delivery partner delete": /\/delivery-partners\/.*\/delete/,
```
kicferk1, line 23: Uh oh, double lines in json make me sad. Do we want to bump the char limit or look into some commonisation for long urls?

spec/cypress/integration/admin/DeliveryPartnerManagement.feature
```diff
@@ -0,0 +1,112 @@
...
+
+    When I click on "Lead Provider 1" label
+    And I click the submit button
+    Then I should be on "choose delivery partner cohorts" page
+    And the page should be accessible
+
+    When I click on "2021" label
+    And I click the submit button
+    Then I should be on "new delivery partner review" page
+    And "main" should contain "New delivery partner"
```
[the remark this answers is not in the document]
callumacrae: Probably, but most of these would be best off replaced with visual tests anyway

spec/cypress/integration/admin/DeliveryPartnerManagement.feature
callumacrae: It's a bit of a weird syntax, but I do this all the time when using websites! It's the equivalent of pressing cmd+a to replace the contents of an input

spec/cypress/support/step_definitions/common-navigation.js
callumacrae: I'd be very against changing the char limit in prettier as it isn't a limit, it's a target - it will also affect function definitions and other object declarations

I also think it looks fine! (also, this isn't json)

spec/cypress/integration/admin/DeliveryPartnerManagement.feature
kicferk1: Soo, it's like deleting all content and entering what we want?

It there some way of writing it in more straightforward english? Like "I delete contents of ... field And I type ... into ... field"?

spec/cypress/support/step_definitions/common-interaction.js
callumacrae: That's a good shout - but I'd be a bit worried about separating things that are on the same page from each other

It's definitely better than what we have now though!

spec/cypress/integration/admin/DeliveryPartnerManagement.feature
callumacrae: We could add "And I clear the input" as a separate step but it seems a bit strange to me - personally I don't think of clearing and then typing in an input as two separate steps, I think of replacing the contents of an input as a single step

Also, there's other stuff like `{enter}` which is useful as users do that too - but definitely doesn't make sense as a separate step to me

kicferk1: I think by using braces and syntax that doesn't sound like spoken english we lose some of the benefits we think cucumber will provide - namely the documentation that's understandable to a non-technical person (be it product managers, content designers, testers, anyone else who wants to know). 

Could we make `And I type` clear input before typing by default?

And another `And I type {string} into {string} field without clearing`?

callumacrae: That feels a bit magic to me and not the behaviour I would expect

kicferk1: Does `I type {string} into {string} field after clearing it` sound less magic?

callumacrae: Sure, but it violates cucumber best practice which says anything like `after` and `and` in a step should be two separate steps instead 😅 

I feel like the easiest way to resolve this would be to just make it two separate steps

kicferk1: I am more happy with two steps than with `{selectall}`.

I guess `I replace content of {string} with {string}` might be a more natural way of describing it, but I'm not sure that fits the style of our descriptions. 

Arghhh, why is english so hard.

spec/cypress/support/step_definitions/common-navigation.js
ethanmills: FWIW, the ruby line length is longer than the prettier one. I prefer *slightly* longer lines than this personally

spec/cypress/support/step_definitions/common-interaction.js
callumacrae: I've made the names a bit more consistent in this PR and will go for some bigger improvements in another PR (I just can't think of what they should be at the moment)

spec/cypress/integration/admin/DeliveryPartnerManagement.feature
```diff
@@ -55,35 +55,37 @@ Feature: Admin user modifying delivery partners
...
 
   Scenario: Admins should be able to edit delivery partners
     When I click on "a" containing "Delivery Partner 1"
     Then I should be on "delivery partner edit" page
     And "name input" should have value "Delivery Partner 1"
     # And the page should be accessible
 
-    When I type "{selectall}New delivery partner" into "name input" field
+    When I clear "name input"
+    And I type "New delivery partner" into "name input"
```
kicferk1: Wrong input I think

callumacrae: What do you mean?

kicferk1: Hmm, I thought it should be `"delivery partner name input"`. But maybe not, because these are different pages...

callumacrae: Ah I just simplified https://github.com/DFE-Digital/early-careers-framework/pull/166/files#diff-205f09c150777a69c6aefe2dd47f2206e4ecb46b977e5494bcd55326f348fd32L18

`"delivery partner name input"` would work too, because the selector is literally just `input[type="text"]` 😆

kicferk1 (approved):

PR #167, opened 2021-03-22
[ECF-1-391] Add paper-trail auditing

amazimbe:
### Context
As... DfE
I need to know what actions have been performed on the system by which DfE users
So that I have an audit trail if something goes wrong and know who to contact if I have a question

### Testing
Acceptance criteria

- It's done when we have a record of all create/edit/delete actions for School Induction Co-ordinators 
- It's done when we have a record of all create/edit/delete actions for Delivery Partners
- It's done when we have a record of all create/edit/delete actions for Lead Provider USERS (not organisations)
- It's done when we have a record of all create/amend/de-register actions for partnerships between Lead Providers and Schools


app/controllers/admin/base_controller.rb
```diff
@@ -3,6 +3,7 @@
 class Admin::BaseController < ApplicationController
   include Pundit
 
+  before_action :set_paper_trail_whodunnit
```
kicferk1, line 6: Should this be the first action? What happens someone tries to view the page without logging in?

db/migrate/20210319153015_create_versions.rb
```diff
@@ -0,0 +1,37 @@
...
+  # when serializing very large objects, `text` might not be big enough.
+  TEXT_BYTES = 1_073_741_823
+
+  def change
+    create_table :versions do |t|
+      t.string   :item_type, { null: false }
+      t.bigint   :item_id,   null: false
+      t.string   :event,     null: false
+      t.string   :whodunnit
+      t.text     :object, limit: TEXT_BYTES
```
kicferk1, line 18: So does it store a serialised object? Json? Something else?

db/migrate/20210319153015_create_versions.rb
```diff
@@ -0,0 +1,37 @@
...
+      # MySQL timestamp columns do not support fractional seconds unless
+      # defined with "fractional seconds precision". MySQL users should manually
+      # add fractional seconds precision to this migration, specifically, to
+      # the `created_at` column.
+      # (https://dev.mysql.com/doc/refman/5.6/en/fractional-seconds.html)
+      #
+      # MySQL users should also upgrade to at least rails 4.2, which is the first
+      # version of ActiveRecord with support for fractional seconds in MySQL.
+      # (https://github.com/rails/rails/pull/14359)
+      #
```
kicferk1, line 32: Do we care about mysql comment? We use postgres after all.

db/migrate/20210319153015_create_versions.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+# This migration creates the `versions` table, the only schema PT requires.
+# All other migrations PT provides are optional.
+class CreateVersions < ActiveRecord::Migration[6.1]
+  # The largest text column available in all supported RDBMS is
+  # 1024^3 - 1 bytes, roughly one gibibyte.  We specify a size
+  # so that MySQL will use `longtext` instead of `text`.  Otherwise,
+  # when serializing very large objects, `text` might not be big enough.
```
kicferk1, line 9: Ditto, I wonder if we need this in postgres.

I also wonder how large we expect our objects to get - I would be shocked if they were longer than 10k characters.

db/schema.rb
```diff
@@ -171,6 +171,10 @@
     t.uuid "school_id", null: false
     t.uuid "lead_provider_id", null: false
     t.uuid "cohort_id", null: false
+    t.string "status", default: "pending", null: false
+    t.string "reason_for_rejection"
```
ethanmills, line 175: Think you need to regenerate the schema?

app/controllers/admin/base_controller.rb
amazimbe: I've moved it to the end

db/migrate/20210319153015_create_versions.rb
amazimbe: I've changed the schema to store this as json

db/migrate/20210319153015_create_versions.rb
amazimbe: The comment came with the auto generator. I've removed the mysql comment

spec/requests/admin/administrators/administrators_spec.rb
```diff
@@ -167,6 +167,16 @@
       expect(admin_profile.discarded?).to be false
       expect(admin_user.discarded?).to be false
     end
+
+    describe "when an audited action", versioning: true do
```
amazimbe, line 171: I added a few of these tests for sanity checking and mostly to confirm whodunnit is being set to the current user.

db/migrate/20210319153015_create_versions.rb
amazimbe: I've removed it and moved to json rather than text.

db/schema.rb
amazimbe: 👍 That's now been sorted

amazimbe:
[Implement audit trail for DfE Admin journey](https://trello.com/c/uAiEFAX8/391-implement-audit-trail-for-dfe-admin-journey)

app/models/delivery_partner.rb
```diff
@@ -4,6 +4,8 @@ class DeliveryPartner < DiscardableRecord
   has_many :provider_relationships
   has_many :lead_providers, through: :provider_relationships
 
+  has_paper_trail
```
kicferk1, line 7: I wonder if we want to form a convention that paper trail is the very first thing in the model, if it has it - so we don't spend time looking for it. Thoughts?

kicferk1 (approved):
Looks good.

app/models/delivery_partner.rb
amazimbe: No strong opinions either way. I'm sure there will be other things we add that we may want to come first and then soon or later we would have to make a choice and I see that convention breaking subsequently.

kicferk1: Yeah, I guess that convention might evolve to "things from external gems come first, then things from Rails, then whatever custom methods we wrote", or something to that effect

PR #169, opened 2021-03-22
Add annotate gem for models

plisovin:
### Context

Quite handy to have the current schema of the ActiveRecord models


app/models/school_local_authority_district.rb
```diff
@@ -1,5 +1,27 @@
...
 
+# == Schema Information
+#
+# Table name: school_local_authority_districts
+#
+#  id                          :uuid             not null, primary key
+#  end_year                    :integer
+#  start_year                  :integer          not null
+#  created_at                  :datetime         not null
+#  updated_at                  :datetime         not null
```
kicferk1, line 11: I think `id`, `created_at` and `updated_at` are gonna be on every single model we have. Not sure it is worth putting them in here - they might obscure the meaningful things.

app/models/school_local_authority_district.rb
```diff
@@ -1,5 +1,27 @@
...
+# Indexes
+#
+#  index_school_local_authority_districts_on_school_id  (school_id)
+#  index_schools_lads_on_lad_id                         (local_authority_district_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (local_authority_district_id => local_authority_districts.id)
+#  fk_rails_...  (school_id => schools.id)
+#
```
kicferk1, line 24: If we can see the id names in the comment above, and there is a relation on the model saying what this model `belongs_to`, do we gain anything by this foreign key comment?

app/models/school_local_authority_district.rb
```diff
@@ -1,5 +1,27 @@
...
+#  start_year                  :integer          not null
+#  created_at                  :datetime         not null
+#  updated_at                  :datetime         not null
+#  local_authority_district_id :uuid             not null
+#  school_id                   :uuid             not null
+#
+# Indexes
+#
+#  index_school_local_authority_districts_on_school_id  (school_id)
+#  index_schools_lads_on_lad_id                         (local_authority_district_id)
```
kicferk1, line 18: I am not sure how often I would ask myself a question about indexes when looking at a model - but I guess not particularly often since I can't recall last time I did. Probably while doing a ticket on optimising something, but then I know I can find them in schema.

kicferk1 (commented):
I definitely like the part where it tells me what fields a model has - particularly those fields that are specific to that model (I don't really care about a model having an id or timestamps, all of them do).

The other parts, not so sure about. Foreign keys are probably easier to read from the model associations (`belongs_to`), and I would keep indexes to the schema - I don't think the knowledge about them is all that useful.

I guess we could keep them to make spotting errors easier, but I fear we would start treating the comments about it like white noise and it wouldn't actually do its job.

I like the long configuration list, cause it gives me hope all of the above can be configured :)

lib/tasks/auto_annotate_models.rake
```diff
@@ -0,0 +1,59 @@
...
+      "exclude_controllers" => "true",
+      "exclude_helpers" => "true",
+      "exclude_sti_subclasses" => "false",
+      "ignore_model_sub_dir" => "false",
+      "ignore_columns" => nil,
+      "ignore_routes" => nil,
+      "ignore_unknown_models" => "false",
+      "hide_limit_column_types" => "integer,bigint,boolean",
+      "hide_default_column_types" => "json,jsonb,hstore",
+      "skip_on_db_migrate" => "false",
```
ethanmills, line 42: Am I right in thinking this generates the comments on every `db:migrate`? In which case, it would be handy for me if it also ran in `RAILS_ENV=test`. Would that cause issues?

lib/tasks/auto_annotate_models.rake
```diff
@@ -0,0 +1,59 @@
...
+      "format_bare" => "true",
+      "format_rdoc" => "false",
+      "format_yard" => "false",
```
ethanmills, line 45: Personally this would be more useful for me in YARD format, because that's what my IDE understands. I know everyone else is likely to prefer bare, however

lib/tasks/auto_annotate_models.rake
plisovin: you are right, it will run on every ```db:migrate``` and we could add ```test``` here

app/models/school_local_authority_district.rb
plisovin: This could be tweaked in defaults i.e. ```show_indexes" => "false```

seb-siem-softwire:
Some of the fields are available on each and every model - id, created at, modified at are pretty standard and I don't think adding them would add too much value as we can assume they will be there by default - happy with other fields - no opinion on foreign keys

plisovin:
> @plisovin Can we close this?

don't know, was a suggestion, if no one likes it then could be closed

PR #173, opened 2021-03-23
ECF-1-401 Send nomination email to schools

amazimbe:
### Context
We need to send invitation / nomination emails to school primary contacts to invite them to register on our service.

The email will contain a unique token that will be used to track the school and some fields to allow us to track when the email was opened.

### DIFF
--- a/app/mailers/school_mailer.rb
+++ b/app/mailers/school_mailer.rb
@@ -3,16 +3,16 @@
 class SchoolMailer < ApplicationMailer
   NOMINATION_EMAIL_TEMPLATE = "a7cc4d19-c0cb-4187-a71b-1b1ea029924f"
 
-  def nomination_email(params)
+  def nomination_email(recipient:, reference:, school_name:, nomination_url:)
     template_mail(
       NOMINATION_EMAIL_TEMPLATE,
-      to: params[:recipient],
-      reference: params[:reference],
+      to: recipient,
+      reference: reference,
       rails_mailer: mailer_name,
       rails_mail_template: action_name,
       personalisation: {
-        school_name: params[:school_name],
-        nomination_link: params[:nomination_url],
+        school_name: school_name,
+        nomination_link: nomination_url,
       },
     )
   end
--- a/app/mailers/school_mailer.rb
+++ b/app/mailers/school_mailer.rb
@@ -3,15 +3,16 @@
 class SchoolMailer < ApplicationMailer
   NOMINATION_EMAIL_TEMPLATE = "a7cc4d19-c0cb-4187-a71b-1b1ea029924f"
 
-  def nomination_email(recipient_email, nomination_url, reference)
+  def nomination_email(params)
     template_mail(
       NOMINATION_EMAIL_TEMPLATE,
-      to: recipient_email,
-      reference: reference,
+      to: params[:recipient],
+      reference: params[:reference],
       rails_mailer: mailer_name,
       rails_mail_template: action_name,
       personalisation: {
-        nomination_link: nomination_url,
+        school_name: params[:school_name],
+        nomination_link: params[:nomination_url],
       },
     )
   end
--- a/app/mailers/school_mailer.rb
+++ b/app/mailers/school_mailer.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class SchoolMailer < ApplicationMailer
+  NOMINATION_EMAIL_TEMPLATE = "a7cc4d19-c0cb-4187-a71b-1b1ea029924f"
+
+  def nomination_email(recipient_email, nomination_url, reference)
+    template_mail(
+      NOMINATION_EMAIL_TEMPLATE,
+      to: recipient_email,
+      reference: reference,
+      rails_mailer: mailer_name,
+      rails_mail_template: action_name,
+      personalisation: {
+        nomination_link: nomination_url,
+      },
+    )
+  end
+end
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+class InviteSchools
+  def run(school_ids)
+    logger = Rails.logger
+    logger.info "Emailing schools"
+
+    school_ids.each do |school_id|
+      school = School.find_by(id: school_id)
+
+      if school.nil?
+        logger.info "School not found: #{school_id} ... skipping"
+        next
+      end
+
+      token = reference
+      recipient = recipient_email(school)
+
+      school.nomination_emails.create!(
+        token: token,
+        sent_to: recipient,
+        sent_at: Time.zone.now,
+      )
+      send_nomination_email(recipient, token)
+    rescue StandardError
+      logger.info "Error emailing school: #{school_id} ... skipping"
+    end
+  end
+
+private
+
+  def send_nomination_email(recipient, token)
+    SchoolMailer.nomination_email(
+      recipient, nomination_url(token), token
+    ).deliver_now
+  end
+
+  def recipient_email(school)
+    return school.secondary_contact_email unless school.primary_contact_email
+
+    school.primary_contact_email
+  end
+
+  def reference
+    loop do
+      value = SecureRandom.hex(16)
+      break value unless NominationEmail.exists?(token: value)
+    end
+  end
+
+  def nomination_url(token)
+    Rails.application.routes.url_helpers.nominations_url(
+      token: token,
+      host: Rails.application.config.domain,
+    )
+  end
+end
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -1,15 +1,15 @@
 # frozen_string_literal: true
 
 class InviteSchools
-  def run(school_ids)
+  def run(school_urns)
     logger = Rails.logger
     logger.info "Emailing schools"
 
-    school_ids.each do |school_id|
-      school = School.find_by(id: school_id)
+    school_urns.each do |urn|
+      school = School.find_by(urn: urn)
 
       if school.nil?
-        logger.info "School not found: #{school_id} ... skipping"
+        logger.info "School not found, urn: #{urn} ... skipping"
         next
       end
 
@@ -23,7 +23,7 @@ class InviteSchools
       )
       send_nomination_email(recipient, token, school.name)
     rescue StandardError
-      logger.info "Error emailing school: #{school_id} ... skipping"
+      logger.info "Error emailing school, urn: #{urn} ... skipping"
     end
   end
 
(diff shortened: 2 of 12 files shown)


db/migrate/20210323145018_add_email_tracking_to_school.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddEmailTrackingToSchool < ActiveRecord::Migration[6.1]
+  def change
+    change_table :schools, bulk: true do |t|
+      t.string   :nomination_email_token
+      t.datetime :nomination_email_sent_at
+      t.datetime :nomination_email_opened_at
+    end
```
ethanmills, line 9: I think we might be better tracking this in a separate table. There is a mechanism to send this email several times, and I think we should keep all the data

db/migrate/20210323145018_add_email_tracking_to_school.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddEmailTrackingToSchool < ActiveRecord::Migration[6.1]
+  def change
+    change_table :schools, bulk: true do |t|
```
ethanmills, line 5: We will also need to record the ID we get from notify

amazimbe: Sure! Thought about adding a notify status too.

db/migrate/20210323145018_add_email_tracking_to_school.rb
amazimbe: Completely agree

db/migrate/20210323145018_add_email_tracking_to_school.rb
ethanmills: That would be good

seb-siem-softwire:
Do you think adding a factory would be a good idea @amazimbe so that we could create nomination emails on the fly

db/migrate/20210324105239_create_nomination_emails.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class CreateNominationEmails < ActiveRecord::Migration[6.1]
+  def change
+    create_table :nomination_emails, id: :uuid do |t|
+      t.string     :token
+      t.string     :notify_status
+      t.string     :notify_id, null: false
+      t.datetime   :sent_at
+      t.datetime   :opened_at
+      t.references :school, null: false, foreign_key: true, type: :uuid
```
ethanmills, line 11: Let's also record the email address we sent to here: the fields on the school can change as GIAS is updated

app/mailers/school_mailer.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def nomination_email(recipient_email, nomination_url, reference)
+    template_mail(
+      NOMINATION_EMAIL_TEMPLATE,
+      to: recipient_email,
+      reference: reference,
+      rails_mailer: mailer_name,
+      rails_mail_template: action_name,
+      personalisation: {
+        nomination_link: nomination_url,
```
ethanmills, line 14: We need to include the school name here, see https://ukgovernmentdfe.slack.com/archives/C01J1J4FM50/p1616759041050100

db/migrate/20210323145018_add_email_tracking_to_school.rb
amazimbe: I'm sending a unique reference to notify when I send the email. The reference is the same as the token

amazimbe:
> Do you think adding a factory would be a good idea @amazimbe so that we could create nomination emails on the fly

I did generate one but then removed it as I'm not using it in my tests. We can always add it later when there's a need.

app/services/invite_schools.rb
```diff
@@ -0,0 +1,57 @@
...
+    school_ids.each do |school_id|
+      school = School.find_by(id: school_id)
+
+      if school.nil?
+        logger.info "School not found: #{school_id} ... skipping"
+        next
+      end
+
+      token = reference
+      recipient = recipient_email(school)
```
ethanmills, line 17: What about if this is blank?

app/services/invite_schools.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      school.nomination_emails.create!(
+        token: token,
+        sent_to: recipient,
+        sent_at: Time.zone.now,
+      )
+      send_nomination_email(recipient, token)
```
ethanmills, line 24: We need to record the ID back from notify

seb-siem-softwire:
@amazimbe cool cool

lib/tasks/schools.rake
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+namespace :schools do
+  desc "Send nomination invitations to schools"
+  task send_invites: :environment do
+    InviteSchool.run(ARGV[1..-1])
```
ethanmills, line 6: I think this will take URNs, not IDs?

app/services/invite_schools.rb
```diff
@@ -0,0 +1,57 @@
...
+    rescue StandardError
+      logger.info "Error emailing school: #{school_id} ... skipping"
```
ethanmills, line 26: We need to record this against the NomintionEmail in the db

app/services/invite_schools.rb
amazimbe: It will fail as we need a recipient address to send the email. I think I've covered this in the specs

app/services/invite_schools.rb
amazimbe: I'm passing that token as a reference to notify. When it's time to poll we'll be using that token to identify the email

app/services/invite_schools.rb
amazimbe: I doubt it. This covers things like no email recipient being available so the email can't be sent or generally when the nomination email record is invalid. If we want to record those then we need to loosen the validations at the db level

lib/tasks/schools.rake
amazimbe: Agree! Probably urns are more meaningful than our ids

kicferk1: Are we planning to trigger nomination emails through a rake task, that gets passed a list of URNs?

I don't particularly like the idea of devs scheduling a task and passing a thousand urns to it. I kind of thought that we would have a task that fires off all of them, and afterwards the admins could resend the emails for schools that didn't get it.

app/services/invite_schools.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+class InviteSchools
+  def run(school_ids)
```
kicferk1, line 4: I might be missing some context, but...

How many times do we expect it to run, and with how many schools exactly?

If we fire it off with 20k schools, it will `deliver_now` 20k emails one after another (one request after another), and fire off some 40k read queries and 20k write queries to the database. 

Queries we can probably handle, I hope. We kind of do that in school import. But firing off that many requests to notify in a short time sounds scary. Do they have some 'send in bulk' option? (It's scary because 1. notify might not like it and 2. because deliver_now is synchronous, I think, and making 20k api calls to the internet will take a looong time).

Actually, I checked it out. Their send in bulk option is quite crap. Manually upload all the email addresses to notify and send it. Ugh.

Do we worry about the load at all? Can we even load test this without spamming schools?

amazimbe: All legitimate concerns you raise. Quite happy to look into how we can space out the sending of the emails using delayed job. I would do that in a separate pr though.

amazimbe: @kicferk1 forgot to add that from what product have said, they will tell us the list of schools to send to or perhaps the criteria of which schools we send to in order to avoid overwhelming our infrastructure. Until, those requirements become clear it's difficult to be exact with the numbers

amazimbe: My understanding is we'll probably have another rake task that selects the schools we want to send to and then calls this one.

kicferk1: Alright, I think it would be convenient to explore options of handling this amount of emails - perhaps other DfE services need to deal with something similar? I am quite happy for it to live in its own PR. 

After all, if we end up with something that can handle all schools at once, it might save our team a lot of time and effort coming up with batches and triggering it for all of those batches.

app/mailers/school_mailer.rb
```diff
@@ -0,0 +1,19 @@
...
+      to: params[:recipient],
+      reference: params[:reference],
+      rails_mailer: mailer_name,
+      rails_mail_template: action_name,
+      personalisation: {
+        school_name: params[:school_name],
+        nomination_link: params[:nomination_url],
+      },
+    )
+  end
```
kicferk1, line 18: Hmm, I don't like the blanket `params` in a method definition. 

I'm happy for it to be a hash (so we don't need to remember an arbitrary order of several arguments), but I would still like the parameter names to be specified.

Also, I do wonder, are some of the computations we do in the rake task going to be needed each time we send a nomination email? Like making a new reference or deciding which school email we send the message to? If so, maybe we want to pass a `school` to this method, and do those calculations here.

amazimbe: I don't wan't to add any of our logic in the mailer. These are just for sending the email and I wan't them to be free of the record keeping or deciding who to send to or when to send and when not to send.

amazimbe: I don't see anything wrong with the hash tbh. Passing in a hash is so ubiquitous when ones wants flexibility with the arguments e.g. model.new / model.create. What is it you don't like about it?

kicferk1: I want the method signature to look like
```
def nomination_email(recipient:, reference:, school_name:, nomination_url:)
```
so I don't need to read the method body to understand what fields I should include in my hash.

amazimbe: Just curious what your cutoff point would be with the number of arguments. I love keyword arguments too but I'd probably move to a hash if there were at least 4 arguments. Happy to change it

kicferk1 (approved):

app/mailers/school_mailer.rb
kicferk1: My cut off point is more to do with "how many different things does the method need to know to function?". 

If a method needs to know about 10 things, and they are all important, distinct, and can't be grouped, then I prefer having 10 named arguments over a single `params`. Having single `params` saves me some time and space when writing the method signature, but it makes it a lot harder to read, and therefore to use effectively.

So really, I try to limit the number of things a method needs to know about. That's where I was going with my suggestion to put some logic in the mailer - so we could pass a `school` rather than `school_name` and `recipient`, which are quite closely tied together in our data. 

For mailers a lot of things that we end up passing is personalisation - if we had like 10 of those, maybe we should think about computing it in the mailer (and unit testing it there), or about computing it somewhere else and passing the complete personalisation object to the mailer. Not sure we need it right now, just a thought :)

PR #174, opened 2021-03-23
Add google analytics

callumacrae:
### Context

https://trello.com/c/2D8PyMn9/377-google-analytics-set-up

### Testing

Do we think this needs tests? Cypress catches it if it completely blows up and throws an error, but doesn't check whether the script is actually present (and I'm not sure whether it should)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -16,6 +16,22 @@
     <%= favicon_link_tag asset_pack_path('media/images/govuk-apple-touch-icon-180x180.png'), rel: 'apple-touch-icon', type: 'image/png', size: '180x180' %>
     <%= stylesheet_pack_tag 'application', media: 'all' %>
     <%= javascript_pack_tag 'application', defer: true unless params[:nojs] === "nojs" %>
+
+    <% if cookies[:cookie_consent_1] == "on" %>
+      <script async src="https://www.googletagmanager.com/gtag/js?id=G-LR5YHTJD65"></script>
+    <% else %>
+      <script>
+        // For adding the script once the user has given permission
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -16,6 +16,22 @@
     <%= favicon_link_tag asset_pack_path('media/images/govuk-apple-touch-icon-180x180.png'), rel: 'apple-touch-icon', type: 'image/png', size: '180x180' %>
     <%= stylesheet_pack_tag 'application', media: 'all' %>
     <%= javascript_pack_tag 'application', defer: true unless params[:nojs] === "nojs" %>
+
+    <% if cookies[:cookie_consent_1] == "on" %>
+      <script async src="https://www.googletagmanager.com/gtag/js?id=G-LR5YHTJD65"></script>
```
ethanmills, line 21: I think this is specific to staging. We need to be able to select different scripts for different environments. Happy for this to come later though

ethanmills:
Happy for this to be manually tested for now. There are a few examples out there of people writing Cypress tests for this, but I don't think we need it right now

ethanmills (approved):

ethanmills:
We seem to have a user in the States?

amazimbe (approved):

app/views/layouts/application.html.erb
callumacrae: Ah, the ticket didn't say that. That would seem a bit strange to me, though - why would you want to track users on staging?

ethanmills: image

ethanmills: 🤷 I think it's mostly because we don't know the prod domain

PR #176, opened 2021-03-24
ECF-410 Fix paper trail audit schema datatype

amazimbe:
### Context

Item Id / record Id in the paper-trail generated schema were generated as bigint whereas we use uuids. Everything still worked as rails was automatically casting to big int but sooner or later we were going to have a key collision. This ticket corrects the datatype to uuid.

### DIFF
--- a/db/migrate/20210323180540_change_item_id_type_to_uuid.rb
+++ b/db/migrate/20210323180540_change_item_id_type_to_uuid.rb
@@ -4,7 +4,7 @@ class ChangeItemIdTypeToUuid < ActiveRecord::Migration[6.1]
   def change
     remove_column :versions, :item_id, :bigint
     remove_index :versions, column: %i[item_type item_id], if_exists: true
-    add_column :versions, :item_id, :uuid, null: false
+    add_column :versions, :item_id, :uuid, default: "gen_random_uuid()", null: false
     add_index :versions, %i[item_type item_id]
   end
 end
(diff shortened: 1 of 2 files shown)


kicferk1:
Uhhh, will we have to manually rollback and redo a migration on dev environment for that? 

Would it make sense to add a new migration afterwards?

db/migrate/20210323180540_change_item_id_type_to_uuid.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class ChangeItemIdTypeToUuid < ActiveRecord::Migration[6.1]
+  def change
+    remove_column :versions, :item_id, :bigint
+    remove_index :versions, column: %i[item_type item_id], if_exists: true
+    add_column :versions, :item_id, :uuid, default: "gen_random_uuid()", null: false
```
ethanmills, line 7: Not sure about the default here...

amazimbe: The default was just to appease rubocop as otherwise the build was failing.  I don't believe we have any important data in that table anyway so I'm quite happy to erase it on dev

amazimbe:
> Uhhh, will we have to manually rollback and redo a migration on dev environment for that?
> 
> Would it make sense to add a new migration afterwards?

@kicferk1 not sure I've understood your comment. Anyway, this pr is meant to fix the migration we have on dev. Why would we need to manually rollback?

kicferk1:
> > Uhhh, will we have to manually rollback and redo a migration on dev environment for that?
> > Would it make sense to add a new migration afterwards?
> 
> @kicferk1 not sure I've understood your comment. Anyway, this pr is meant to fix the migration we have on dev. Why would we need to manually rollback?

I might have got confused by something - for some reason I thought you edited a migration file, rather than making a new one.

ethanmills (approved):

seb-siem-softwire (approved):

kicferk1 (approved):

PR #178, opened 2021-03-24
Remove dependencies on broken rails components

ethanmills:
See https://github.com/rails/rails/issues/41750


ethanmills:
Currently the following depend on the whole of rails:
canonical-rails 
govuk-components 
health_check

ethanmills:
Lint issues are mostly because the behaviour depends on the version of rails 🙄 Will add ignore comments for now

config/environments/staging.rb
```diff
@@ -37,7 +37,7 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
-  config.active_storage.service = :local
+  # config.active_storage.service = :local # TODO enable when rails is upgraded
```
kicferk1, line 40: Do we even need to enable that? Are we handling any files? Is active storage needed for anything else?

kicferk1 (approved):

config/environments/staging.rb
ethanmills: I think we'll be taking csv uploads, so will need

seb-siem-softwire:
You do what you got to do to get deployments working again - a temporary solution to get things moving again - as long as we go back and undo things - it should be ok. It's hard to believe how one gem broke the internet. Mime has lost its magic

seb-siem-softwire (approved):

amazimbe (approved):

PR #180, opened 2021-03-25
ECF-RP-335: start page and service name

ethanmills:
### Context
Two small changes here: start page copy, and we now have a service name 🎉 

### Guidance to review
I haven't updated the service name in notify yet, so there's still a couple of references to that. Apart from that, let me know if you think I've missed anything

### DIFF
--- a/app/views/cookies/show.html.erb
+++ b/app/views/cookies/show.html.erb
@@ -18,10 +18,10 @@
 
     <h1 class="govuk-heading-l">Cookies</h1>
     <p>Cookies are small files saved on your phone, tablet or computer when you visit a website.</p>
-    <p>We use cookies to make the Early Career Framework (ECF) work and collect information about how you use our service.</p>
+    <p>We use cookies to make the Manage training for early career teachers service (the service) work and collect information about how you use our service.</p>
 
     <h2 class="govuk-heading-m">Essential cookies</h2>
-    <p>Essential cookies keep your information secure while you use ECF. We do not need to ask permission to use them. </p>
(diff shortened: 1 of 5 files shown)


kicferk1 (approved):
There are a few more references
1. `TWD Early Careers Framework` in `school_not_confirmed/show`
2. Lots of `early-careers-framework` in our dev ops. I don't think we want to change all of it, maybe not even some of it, just thought I would mention it in case you disagree :)

ethanmills:
> Lots of early-careers-framework in our dev ops. I don't think we want to change all of it, maybe not even some of it, just thought I would mention it in case you disagree :)

That's a more difficult, but less immediate issue. I decided I didn't want to deal with that today

ethanmills:
> TWD Early Careers Framework in school_not_confirmed/show

That's the notify thing I mentioned. That's still true at the moment

app/views/cookies/show.html.erb
```diff
@@ -18,10 +18,10 @@
 
     <h1 class="govuk-heading-l">Cookies</h1>
     <p>Cookies are small files saved on your phone, tablet or computer when you visit a website.</p>
-    <p>We use cookies to make the Early Career Framework (ECF) work and collect information about how you use our service.</p>
+    <p>We use cookies to make the Manage training for early career teachers service (the service) work and collect information about how you use our service.</p>
```
callumacrae, line 21: Is MTFECTS not catchy enough?

callumacrae (approved):

seb-siem-softwire (approved):

PR #182, opened 2021-03-25
ECF-RP-396: Choose programme

ethanmills:
Choose between CIP and FIP

This is the first thing an induction coordinator sees when signing in, and they don't see it again once they have answered


ethanmills:
Still outstanding is a custom validation message, TODO in code

app/controllers/schools/choose_programme_controller.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class Schools::ChooseProgrammeController < Schools::BaseController
+  skip_after_action :verify_authorized
+  skip_after_action :verify_policy_scoped
```
kicferk1, line 5: Are we sure we want to skip them?

Or should we have a policy for this?

app/controllers/schools/choose_programme_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    @induction_choice_form = InductionChoiceForm.new
+  end
+
+  def create
+    @induction_choice_form = InductionChoiceForm.new(params.require(:induction_choice_form).permit(:programme_choice))
+
+    render :show and return unless @induction_choice_form.valid?
+
+    @cohort = Cohort.current
+    @school = current_user.induction_coordinator_profile.schools.first
```
kicferk1, line 20: Let's make these local, not `@` variables.

app/controllers/schools/choose_programme_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    @school_cohort = SchoolCohort.find_or_create_by!(
```
kicferk1, line 22: That variable is unused, I think. All the more reason to use local variables when possible - RubyMine will tell you the variable is unused if it's local, but not if it's `@`.

callumacrae (approved):

app/controllers/users/confirmations_controller.rb
```diff
@@ -8,6 +8,7 @@ def show
     if resource.errors.empty?
       notify_school_primary_contact
       sign_in(resource)
+      @current_user = current_user
```
kicferk1, line 11: Why do we need this here?

In `ApplicationController` we do
```
before_action :set_current_user, unless: :devise_controller?
```

Do we want to change that?

app/controllers/schools/dashboard_controller.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class Schools::DashboardController < Schools::BaseController
+  skip_after_action :verify_authorized
+  skip_after_action :verify_policy_scoped
```
kicferk1, line 5: Are we ever using those in a controller inheriting from `Schools::BaseController`? :D

app/helpers/application_helper.rb
```diff
@@ -2,10 +2,22 @@
...
+      induction_coordinator_dashboard_url(user)
     else
       dashboard_url
     end
   end
+
+private
+
+  def induction_coordinator_dashboard_url(user)
+    school = user.induction_coordinator_profile.schools.first
```
kicferk1, line 17: Do we have a ticket about one induction coordinator having multiple schools? Is that even on our radar?

app/helpers/application_helper.rb
```diff
@@ -2,10 +2,22 @@
...
+
+    return schools_choose_programme_url unless school.chosen_programme?(Cohort.current)
+
+    schools_dashboard_url
```
kicferk1, line 21: Are we sure this is the right behaviour? If there are multiple tasks, maybe showing them all on dashboard would be neater?

Ech, something to chase with designers I guess.

app/views/schools/choose_programme/show.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+  <h1 class="govuk-heading-xl">Welcome</h1>
+  <%= form_for @induction_choice_form, url: schools_choose_programme_path, method: :post do |f| %>
```
kicferk1, line 4: Some indentation woes.

app/views/schools/choose_programme/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <%= f.govuk_collection_radio_buttons(
+          :programme_choice, @induction_choice_form.programme_choices, :id, :name,
+          legend: { text: "How do you want to run your induction in the 2021/22 academic year?", size: "m" },
```
kicferk1, line 7: I prefer one argument per line, it's a bit hard to red when you have to guess how many arguments there are per line (or look for commas)

app/views/schools/choose_programme/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+          hint: -> { tag.p do
+            concat(govuk_link_to("See our guidance on programmes", "https://www.gov.uk/government/publications/early-career-framework-reforms-overview/early-career-framework-reforms-overview", target: "_blank", rel: "noopener noreferrer"))
```
kicferk1, line 9: `early-career-framework-reforms-overview` appears twice in the url - is this intentional?

app/views/schools/dashboard/show.html.erb
```diff
@@ -0,0 +1 @@
+Dashboard goes here
```
kicferk1, line 1: "I promise, it will" :D

config/routes.rb
```diff
@@ -112,6 +111,11 @@
     resources :induction_coordinators, only: %i[index edit update], path: "induction-coordinators"
   end
 
+  namespace :schools do
+    resource :dashboard, controller: :dashboard, only: :show, path: "/"
+    resource :choose_programme, controller: :choose_programme, only: %i[show create], path: "choose-programme"
```
kicferk1, line 116: Do we need controller names if they match resource name?

db/schema.rb
```diff
@@ -200,7 +200,7 @@
   end
 
   create_table "school_cohorts", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
-    t.string "induction_programme_choice", default: "not_yet_known", null: false
+    t.string "induction_programme_choice", null: false
```
kicferk1, line 203: Is that gonna work? Can we make it nullable?

spec/cypress/integration/schools/ChooseProgramme.feature
```diff
@@ -0,0 +1,25 @@
...
+  Scenario: Choosing Core Induction Programme
+    Given I am logged in as an "induction_coordinator"
+    Then I should be on "choose programme" page
+    
+    When I click on "accredited materials" label
+    And I click the submit button
+    Then I should be on "schools" page
+
+    When I am on "choose programme" page
+    Then I should be on "schools" page
```
kicferk1, line 17: That reads a bit weird, probably because immediate redirects aren't something we wrote our cucumber steps for :/

spec/cypress/integration/schools/ChooseProgramme.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Induction tutors choosing programmes
+  Induction tutors should be able to choose between the Full and Core
+  induction programmes for their school
```
kicferk1, line 3: I would like a few "And the page should be accessible" all over this spec :)

spec/requests/users/confirmations_spec.rb
```diff
@@ -4,8 +4,8 @@
 
 RSpec.describe "Users::Confirmations", type: :request do
   let(:confirmation_token) { "soC7rF8i_BgYdUxafFcP" }
-  let(:school) { create(:school) }
   let(:induction_coordinator) { user.induction_coordinator_profile }
+  let(:school) { induction_coordinator&.schools&.first }
```
kicferk1, line 8: Do we need the null safety checks here?

kicferk1 (commented):
A bunch of small comments, and funnily enough I left them for myself :P 

Have a read through them before you jump into fixing any, just so that I get my questions answered before you run off for holidays :D

app/controllers/schools/choose_programme_controller.rb
ethanmills: I guess we could have `SchoolCohortPolicy`. It should specify you can only create for your own school

app/controllers/schools/dashboard_controller.rb
ethanmills: Ah but we will be!

app/controllers/users/confirmations_controller.rb
ethanmills: So that we can do [this](https://github.com/DFE-Digital/early-careers-framework/pull/182/files/63078a0de6e6197e916bf41831d5a4f3bca1a8dd#diff-8d1216620140287a4b4580989961295ee24e8381011a36ff54d69e79f7a846c6L6). We're not actually using confirmations of any kind, and I'm not sure we plan to

app/helpers/application_helper.rb
ethanmills: Alaric said it's unlikely but that the first thing would be a school picker screen if that were to happen. We're deliberately omitting it for now

app/helpers/application_helper.rb
ethanmills: The idea was you can't do anything until you've made this choice. I think in future when you have more than one cohort this might move under each of the cohorts on the dashboard (there's a post it on the wireframes for future iterations). But you can't see anything else until you've answered this, because the answer to this decides what you see

app/views/schools/choose_programme/show.html.erb
ethanmills: Agreed, this was mostly moved, not written from scratch

app/views/schools/choose_programme/show.html.erb
ethanmills: Yep. As for why... ask comms 😛

app/views/schools/dashboard/show.html.erb
ethanmills: That's Callum's job 😛

config/routes.rb
ethanmills: Singular vs plural confusion I think

db/schema.rb
ethanmills: Why wouldn't it work? The idea is you can't do anything else until you've decided this

ethanmills: If it's null, everything else breaks

spec/requests/users/confirmations_spec.rb
ethanmills: It's for `context "when user is not an induction coordinator" do`

db/schema.rb
kicferk1: I hope you don't read it for a while, but I think I got confused and thought `SchoolCohorts` would exist before the user selects anything - which would obviously break things. But I don't think they will.

spec/cypress/integration/schools/ChooseProgramme.feature
callumacrae: We could add a `I should have been redirected to` step which does the same thing (idk if there's a way to check it's actually redirected too)

app/controllers/schools/choose_programme_controller.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class Schools::ChooseProgrammeController < Schools::BaseController
+  skip_after_action :verify_authorized
+  skip_after_action :verify_policy_scoped
+
+  def show
+    school = @current_user.induction_coordinator_profile.schools.first
+    redirect_to helpers.profile_dashboard_url(@current_user) if school.chosen_programme?(Cohort.current)
```
amazimbe, line 9: I think this should be redirect_to and return

app/models/school.rb
```diff
@@ -84,6 +84,10 @@ def partially_registered?
       &.any?
   end
 
+  def chosen_programme?(cohort)
+    school_cohorts.find_by(cohort: cohort).present?
```
amazimbe, line 88: school_cohorts.exists?(cohort:  cohort) would probably be neater

spec/requests/users/confirmations_spec.rb
amazimbe: Not sure why this got updated. Are we reusing this in the new school journey? I expected it to be taken out in subsequent prs.

app/controllers/schools/choose_programme_controller.rb
kicferk1: I went with `if-else` statement, but you are right, anything that happens after `redirect_to` will still be fired off :/

app/controllers/users/confirmations_controller.rb
amazimbe, line 11: @current_user is definitely not necessary. Why can't we just call current_user whereever this is being used?

app/forms/induction_choice_form.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class InductionChoiceForm
```
amazimbe, line 3: I think this form is missing some unit tests

spec/requests/users/confirmations_spec.rb
kicferk1: It got updated because in one of the tests we test that an email was (or wasn't) sent to a particular school.

I think
```
expect(UserMailer).not_to receive(:primary_contact_notification).with(anything, anything)
```
is a lot broader than checking for an email from a particular school, so we shouldn't need `school` in that test.

app/controllers/schools/choose_programme_controller.rb
kicferk1: Decided it's not worth the time - we never let the user specify which school they are choosing for, so there is no need to restrict that.

app/controllers/application_controller.rb
```diff
@@ -4,7 +4,6 @@ class ApplicationController < ActionController::Base
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
   before_action :configure_permitted_parameters, if: :devise_controller?
-  before_action :set_current_user, unless: :devise_controller?
```
amazimbe, line 7: 👍

amazimbe (approved):

PR #183, opened 2021-03-25
Add auto tag component

callumacrae:
### Context

For the induction tutor onboarding stuff

Small wrapper around the govuk-components tag component which handles the colour for us

### Testing

we need percy!

### DIFF
--- a/app/components/auto_tag_component.rb
+++ b/app/components/auto_tag_component.rb
@@ -3,7 +3,7 @@
 class AutoTagComponent < ViewComponent::Base
   def initialize(text:)
     colours = {
-      "to do": "grey",
+      "to do": "yellow",
       "cannot start yet": "grey",
       "setup complete": "green",
       "done": "green",
--- a/app/components/auto_tag_component.rb
+++ b/app/components/auto_tag_component.rb
@@ -1,16 +1,16 @@
 # frozen_string_literal: true
 
 class AutoTagComponent < ViewComponent::Base
-  def initialize(text:)
-    colours = {
-      "to do": "yellow",
-      "cannot start yet": "grey",
-      "setup complete": "green",
-      "done": "green",
-    }
+  COLOURS = {
+    "to do": "yellow",
+    "cannot start yet": "grey",
+    "setup complete": "green",
+    "done": "green",
+  }.freeze
 
+  def initialize(text:)
     @text = text
-    @colour = colours[text.downcase.to_sym] || "grey"
+    @colour = COLOURS[text.downcase.to_sym] || "grey"
   end
 
   def call
--- a/app/components/auto_tag_component.rb
+++ b/app/components/auto_tag_component.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class AutoTagComponent < ViewComponent::Base
+  def initialize(text:)
+    colours = {
+      "to do": "grey",
+      "cannot start yet": "grey",
+      "setup complete": "green",
+      "done": "green",
+    }
+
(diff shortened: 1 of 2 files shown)


app/components/auto_tag_component.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class AutoTagComponent < ViewComponent::Base
+  def initialize(text:)
+    colours = {
+      "to do": "grey",
+      "cannot start yet": "grey",
+      "setup complete": "green",
+      "done": "green",
```
ethanmills, line 9: These look like the wrong colours to me. Please reference the prototype, not the wireframes, as the source of truth

callumacrae: They're not all on the prototype (at least, the school preparing to do cip+fip prototypes), but I changed the one I could see which was wrong

app/components/auto_tag_component.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class AutoTagComponent < ViewComponent::Base
+  def initialize(text:)
+    colours = {
+      "to do": "yellow",
+      "cannot start yet": "grey",
+      "setup complete": "green",
+      "done": "green",
+    }
```
kicferk1, line 10: Can we move that to a constant outside `initialize` method? We don't need to make a copy for each instance of the component :)

app/components/auto_tag_component.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    @text = text
+    @colour = colours[text.downcase.to_sym] || "grey"
+  end
+
+  def call
+    render GovukComponent::Tag.new(text: @text, colour: @colour)
+  end
```
kicferk1, line 18: Does that replace the need for erb template?

app/components/auto_tag_component.rb
callumacrae: gooood point, i put it in there to debug and then forgot about it 😆

app/components/auto_tag_component.rb
callumacrae: yes

kicferk1 (approved):

amazimbe (approved):

app/components/auto_tag_component.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class AutoTagComponent < ViewComponent::Base
+  COLOURS = {
+    "to do": "yellow",
+    "cannot start yet": "grey",
+    "setup complete": "green",
+    "done": "green",
+  }.freeze
```
amazimbe, line 9: You don't need this freeze by the way . The # frozen_string_literal: true takes care of it

callumacrae: The linter added it, not me

kicferk1: Surely this is freezing the hash, not the strings in it?

Or so I hope by reading it anyway. If so, the comment won't be enough to freeze the hash, I guess.

amazimbe: It's a bit odd to use hash keys that have spaces in them. Pretty sure you can just use symbols

amazimbe: Yes, it is freezing the hash ☹️

callumacrae: they have spaces in because they're not primarily keys, they're the text that will be displayed to users - the hash is just for looking up what colour each text should have

PR #186, opened 2021-03-26
ECF-RP-411: Map routes to apps

ethanmills:
### Context
We have domains, we need to link them up to the apps for staging and prod

Draw inspiration from teaching vacancies

### Testing
ran a `terraform plan` for staging locally and it looked good

### DIFF
--- a/.github/workflows/deploy_to_dev.yml
+++ b/.github/workflows/deploy_to_dev.yml
@@ -8,9 +8,8 @@ on:
       - 'documentation/**'
 
 jobs:
-  turnstyle:
+  deploy:
     runs-on: ubuntu-20.04
-    timeout-minutes: 20
     steps:
       - uses: softprops/turnstyle@v1
         name: Check workflow concurrency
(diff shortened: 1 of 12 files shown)


ethanmills:
I've also bundled in a couple of other small changes, passing through the SHA to the app properly and putting the turnstyle in the right place

.github/workflows/deploy_to_dev.yml
```diff
@@ -8,9 +8,8 @@ on:
       - 'documentation/**'
 
 jobs:
-  turnstyle:
```
kicferk1, line 11: do we hate turnstyle? I know we copied it from somewhere else, is it not doing anything useful?

kicferk1 (approved):
Looks good to me

.github/workflows/deploy_to_dev.yml
kicferk1: Oh, sorry, I misread the changes, we still use it only elsewhere. Git diffs can be confusing.

PR #191, opened 2021-03-26
Update notify API key to production (live)

ethanmills:
### Context
I'm confident we won't email any schools by accident, the prod db won't have any schools in it


cfq (approved):

kicferk1 (approved):
very tidily written secret, I like lower letter start :D 
(no, I haven't actually checked it out, I am taking the piss)

PR #192, opened 2021-03-26
Set up statuscake monitoring for production

ethanmills:
### Context
Statuscake provides uptime monitoring by polling your site. When we have production, we want to make sure we keep production up...

This was lifted almost verbatim from Teaching Vacancies (thanks!). We're only monitoring prod, but we need to pass the API keys in for all envs to keep terraform happy. It's only uptime monitoring, so I'm not too fussed about that. 

There are a couple of steps that need to be done manually here:
- @cfq is sorting out a slack hook for alerting
- I think we will need to go in to statuscake and give the tests sensible names once this is deployed. @csutter sorry for the ping, I noticed your test names were totally different to what's in the code. Am I right in thinking you manually updated?


amazimbe (approved):
If it's good enough for teaching vacancies it must be good enough for us 😉

PR #194, opened 2021-03-29
Update rails

kicferk1:
### Context
Internet (Rails) was broken due to licensing issues. It should now work with new patch of Rails. 

Revert https://github.com/DFE-Digital/early-careers-framework/pull/178 which was a temporary fix anyway, update Rails to the newest shiny version (https://github.com/rails/rails/releases), and hope it works.

### Guidance to review
Commit by commit. 

### Testing
If it builds, it's test enough for me.


Gemfile
```diff
@@ -69,7 +58,7 @@ gem "tzinfo-data"
 # serialization gem that offers more features than active model serializer
 gem "blueprinter"
 
-gem "govuk-components", ">= 1.1.5"
+gem "govuk-components", ">= 1.1.3"
```
seb-siem-softwire, line 61: Are we downplaying gov-uk components?

seb-siem-softwire:
Just the question about gov uk components - otherwise its fine

seb-siem-softwire (approved):

Gemfile
kicferk1: Oops, didn't notice that, good spot

PR #190, opened 2021-03-26
CIP + FIP next steps pages

callumacrae:
### Context

Branched off ECF-RP-396_choose-programme, will point this PR at develop once that is merged.

### Changes proposed in this pull request

Adds generic "Manage your training" and "Next steps" pages that are used for both CIP and FIP.

Also adds some kind of unrelated step definitions allowing us to create simple factories directly from our cucumber specs.

### Guidance to review

`SchoolCohort.first.destroy` will undo the CIP/FIP selection when testing.

### Testing

Not sure there's any point adding tests until the statuses can change / the user journey is a bit more complete.

PR #197, opened 2021-03-30
Add placeholder pages for legal + add participants pages

callumacrae:
### Context

https://trello.com/c/ewjKniov/408-placeholders-for-add-participants-and-gdpr

Branched from https://github.com/DFE-Digital/early-careers-framework/pull/190

- Adds "Read and accept privacy and data policy" placeholder page
- Adds "Add teachers and mentors" placeholder page

### Testing

These will need visual tests when we have them, but as they're placeholder pages for now I'm not sure anything else is necessary.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -133,7 +133,9 @@ Rails.application.routes.draw do
     resource :choose_programme, controller: :choose_programme, only: %i[show create], path: "choose-programme"
     resources :cohorts do
       member do
-        get "/:start_year", action: :show
+        get "show"
+        get "legal"
+        get "add_participants"
       end
     end
   end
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -116,7 +116,9 @@
     resource :choose_programme, controller: :choose_programme, only: %i[show create], path: "choose-programme"
     resources :cohorts do
       member do
-        get "/:start_year", action: :show
+        get "show"
```
amazimbe, line 119: This one you probably don't need to define. When you have a resource like cohorts above rails automaticallys creates /cohorts/:id and you can use that for the show

amazimbe (approved):

kicferk1 (approved):

PR #198, opened 2021-03-30
[ECF-1-405] Nomination confirmation email

amazimbe:
### Context
Induction coordinator email notifying them they have been nominated as the induction tutor at a school.

https://trello.com/c/fAKWBldv/405-induction-coordinator-receives-an-email-notifying-them-they-have-been-nominated

### DIFF
--- a/app/mailers/school_mailer.rb
+++ b/app/mailers/school_mailer.rb
@@ -2,6 +2,7 @@
 
 class SchoolMailer < ApplicationMailer
   NOMINATION_EMAIL_TEMPLATE = "a7cc4d19-c0cb-4187-a71b-1b1ea029924f"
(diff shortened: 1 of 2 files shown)


kicferk1 (approved):
> LGTM

> It dawns on me that our `nomination_email` is called `Nomination invite email` in notify.

> I wonder if we want to name them something like "school nomination email" and "tutor nomination email"? 

> Also, do we want to use the new added email in the `ResendNominationForm`'s save method? 

Nomination email was ok when we had just one but now that we have 2 similar ones I'm happy to try your suggestion.

amazimbe:
@kicferk1 I think ResendNominationForm's ought to use the school nomination email and not this one

PR #201, opened 2021-03-31
Docs - add link to cpd-adr

timabell:
### Context

We discussed issues of testing across all the CPD projects and created a separate ADR repo to cover multiple projects.

* Docs - add link to cpd-adr

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,3 +123,6 @@ to execute this delayed job in the background.
 1. Add an option `on_duplicate_key_ignore` to lead providers, think carefully which ones from seed dump are needed.
 1. Commit, push, run seeding job from above in the deployed app.
 
+## Architecture Decision Records
+
+* There is a cross-project ADR at <https://github.com/DFE-Digital/cpd-adr>


amazimbe (approved):
LGTM!

kicferk1:
If one place for architecture documentation is our goal, there is a bunch of old documentation on https://dfedigital.atlassian.net/wiki/spaces/ECF/overview - some relating to early architecture decisions.

We should probably not have multiple places documenting stuff.

Also, as you will probably note, very little actually gets documented in markdown files in practice - and we tend to keep documentation on what to do with the current setup, rather than what we did months ago. Adding a programme-wide ADR probably means we need to agree what should go into one, and stick to it, or it will turn into a mess.

ethanmills:
@timabell Can we close this in favour of linking to tech docs when they're ready?

timabell:
Note from @mattonfoot via slack

> " this middleman site should do, we can tweak the content because its just lifted from the google doc so the auth bit is wrong now. https://github.com/DFE-Digital/early-careers-framework/tree/api-tech-docs/docs "

PR #202, opened 2021-03-31
Validate accessible paths

callumacrae:
To help us keep track of which pages are being tested for accessibility in the cypress tests.

This will add an additional step when adding new accessibility tests locally - they'll either have to modify the array manually or run `yarn cypress:run` to update it automatically (which takes about 2 minutes to run currently). Then we'll be able to see the changed array in the pull request.

I'm also looking for comments on whether we actually want to do this!


spec/cypress/support/accessible-paths.json
```diff
@@ -0,0 +1,37 @@
+[
```
kicferk1, line 1: Can we retrieve a list of paths from Rails at test runtime? 

https://guides.rubyonrails.org/routing.html#inspecting-and-testing-routes has some interesting ideas

I really don't want to re-create our routes file in json...

spec/cypress/plugins/index.js
```diff
@@ -12,11 +12,56 @@
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
 
+const fs = require("fs");
+const path = require("path");
 const cucumber = require("cypress-cucumber-preprocessor").default;
+const diff = require("fast-array-diff");
+
+/* eslint-disable no-console */
```
kicferk1, line 20: Do we need to enable it later?

kicferk1:
I would rather enforce `And the page should be accessible` on literally every cucumber test than have this setup. 

I am certain sooner or later (or continuously) people will forget to run a command that adds something to a json array, which is used in an obscure check on something that requires a different setup to run. 

Sorry I'm so negative xD
It just looks to me like one of those things that gets forgotten about, and someone needs to come clean it up every now and then - that someone most likely being you.

spec/cypress/support/accessible-paths.json
callumacrae: We could maybe replace it with that some day, but at the moment, not all routes are tested/accessible.

callumacrae:
> If we did end up using it, can we have an action that recreates the json file, and throws a fit if it differs from the one in code? I dislike it still, but at least it means we won't forget about it when Callum goes on holidays :P

Great idea - and it's what this PR is already doing 😆

kicferk1 (commented):
Had anther look. Left a few comments around readability, one question around running a subset of tests.

I personally think this gives us a bit too little to justify the need to re-run all cypress tests locally on any change to what routes are tested. Cause as I see it, what it gives us is a warning if we accidentally remove an accessibility check from one of the pages that were checked in the past - I hope it won't happen often, but in most of these cases it would also require cutting out an accessibility check in cucumber, which hopefully the reviewers would notice.

I think we could turn it into a check that for every page we visit in tests, we make an accessibility check. Callum found a lovely `url:changed` event, which would let us do that. 

We should probably discuss the details on it on dev catchup.

callumacrae:
Converted to draft until discussed further!

ethanmills:
@callumacrae can we close this or discuss it?

PR #203, opened 2021-03-31
ECF-1-407 Adds tasks to school_cohorts plus refactoring

amazimbe:
I've added a new model, task, that keeps track of what an induction tutor needs to do when they have selected the cip or the fip for their school. The rest is just refactoring. 
Screenshot 2021-03-31 at 23 39 02

### DIFF
--- a/app/models/school_cohort.rb
+++ b/app/models/school_cohort.rb
@@ -8,57 +8,49 @@ class SchoolCohort < ApplicationRecord
     not_yet_known: "not_yet_known",
   }
 
+  CIP_TASKS = [
+    { name: "choose_materials", description: "Choose your training materials" },
+    { name: "add_participants", description: "Add teachers and mentors", status: "on_hold" },
+  ].freeze
+
+  FIP_TASKS = [
+    { name: "add_estimates", description: "Add estimated numbers of teachers and mentors" },
+    { name: "add_provider", description: "Sign up with a training provider" },
+    { name: "accept_legals", description: "Read and accept privacy and data policy" },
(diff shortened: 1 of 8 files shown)


callumacrae:
This is great!

kicferk1 (commented):
Is this a good idea? We will know whether the user completed those tasks through other things stored in the database. 

For example, we will know if user estimated how many ECTs and mentors because those values won't be null on a school cohort. We will know if they chose a provider, because that relationship will exist in our database.

And with this, we will need to worry about that data getting out of sync. Is this something we want to worry about?

callumacrae:
> Is this a good idea? We will know whether the user completed those tasks through other things stored in the database.
> 
> For example, we will know if user estimated how many ECTs and mentors because those values won't be null on a school cohort. We will know if they chose a provider, because that relationship will exist in our database.
> 
> And with this, we will need to worry about that data getting out of sync. Is this something we want to worry about?

I thought about this a bit, but for stuff like accepting GDPR

callumacrae:
sorry wrong button!

amazimbe:
> Is this a good idea? We will know whether the user completed those tasks through other things stored in the database.
> 
> For example, we will know if user estimated how many ECTs and mentors because those values won't be null on a school cohort. We will know if they chose a provider, because that relationship will exist in our database.
> 
> And with this, we will need to worry about that data getting out of sync. Is this something we want to worry about?

Initially, I thought it was but I'm changing my mind. I'll release the pr that prompted this and then see if we need it or not. Probably just need methods but not schema. I'll put this on hold for now

PR #204, opened 2021-04-01
Estimate participants

goodviber:
### Context
Trello ticket 406 - we need to enable users to estimate ECTs and mentors
Added SchoolCohortForm Object 
### Guidance to review
Form object not validating as expected when initialized with SchoolCohort. attribute values injected into edit form directly, not ideal
### Testing
There  is  some overlap with ChooseProgramme.feature, perhaps we could have a background that allows to start from /schools/cohort/2021 or just extend the ChooseProgramme.feature ?


app/controllers/schools/estimate_participants_controller.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class Schools::EstimateParticipantsController < Schools::BaseController
+  # who has permissions to view this ?
```
amazimbe, line 4: School induction tutors. That's being checked in the base controller which this controller extends

app/controllers/schools/estimate_participants_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      redirect_to "/schools/cohorts/2021" # TODO: this needs changing?
+    else
+      render :edit
+    end
+  end
+
+private
+
+  def get_school_cohort
+    @school_cohort = SchoolCohort.find(params[:id])
```
callumacrae, line 27: You'll need to select the school_cohort for the correct school, too - see here: https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/controllers/schools/cohorts_controller.rb#L7-L19

app/views/schools/estimate_participants/_form.html.erb
```diff
@@ -0,0 +1,34 @@
+Which heading do we want ?
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+  <!-- have we a helper for cohort year, yes -->
+    <h2 class="govuk-heading-l">Estimated numbers of participants starting in 2021</h2>
```
amazimbe, line 5: You can use the cohort.start_year rather than hardcoding 2021

app/controllers/schools/estimate_participants_controller.rb
amazimbe: It depends what params[:id] came from. If it is the id of the school cohort then this will work

app/views/schools/estimate_participants/_form.html.erb
goodviber: Yes, I found it

app/controllers/schools/estimate_participants_controller.rb
goodviber: I am getting the id param from the previous screen, /schools/cohort/2021, would that not identify the correct SchoolCohort?

callumacrae: ahh, that's true - the other URLs are using the start_year, not the ID

goodviber: That is what I am doing, the id is something like adsd-afafsg-435345-43645654 .  Could we not slug the /schools/cohort/:start_year instead of using a member route?

config/locales/en.yml
```diff
@@ -39,3 +39,16 @@ en:
...
+        school_cohort_form:
+          attributes:
+            estimated_mentor_count:
+              not_a_number: "Enter a number between 0 and 1000"
+            estimated_teacher_count:
+              not_a_number:
+                not_a_number: "Enter a number between 0 and 1000"
+
+
+
```
kicferk1, line 54: No extra whitespace please

config/locales/en.yml
```diff
@@ -39,3 +39,16 @@ en:
               blank: "Enter an email"
               taken: "This email address is already in use"
               invalid: "Enter a valid email address"
+  activemodel:
+    errors:
+      models:
+        school_cohort_form:
```
kicferk1, line 45: I think we don't need to repeat `activerecord` and `errors` and `models` - specifying it once in lines 33-35 should be anough.

kicferk1 (commented):
A few small things

app/views/schools/estimate_participants/_form.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+    <%= form_with(model: @school_cohort_form, url: schools_estimate_participant_path(@school_cohort), method: :patch) do |f| %>
+      <%= f.govuk_error_summary %>
+      <%= f.govuk_text_field(
+          :estimated_teacher_count,
+          width: 20,
+          link_errors: true,
+          value: @school_cohort.estimated_teacher_count,
```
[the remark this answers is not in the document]
goodviber: The controller is set with the school_cohort_form values, which are not persisted. If you visit the form for the first time, the fields will  be blank, then you set the values. If you visit the form again, without a value, they are blank again, so the user cannot immediately see what the current values are. Unless I am missing something?

app/views/schools/estimate_participants/_form.html.erb
```diff
@@ -0,0 +1,34 @@
+Which heading do we want ?
```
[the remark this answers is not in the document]
goodviber: It is a reference to the content, am checking with Alaric

app/controllers/schools/estimate_participants_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  def edit
+    @school_cohort_form = SchoolCohortForm.new
+  end
+
+  def update
+    @school_cohort_form = SchoolCohortForm.new(form_params)
+
+    if @school_cohort_form.valid?
+      @school_cohort.update!(form_params)
+      redirect_to "/schools/cohorts/2021" # TODO: this needs changing?
```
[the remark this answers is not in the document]
goodviber: What is the task list, atm it is returning to 
Screenshot 2021-04-01 at 14 40 14

app/controllers/schools/estimate_participants_controller.rb
callumacrae: > That is what I am doing, the id is something like adsd-afafsg-435345-43645654 . Could we not slug the /schools/cohort/:start_year instead of using a member route?

i was hoping someone else would reply to this but nobody else has 😆 what do you mean? I don't really understand

app/views/schools/estimate_participants/_form.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+      <%= f.govuk_text_field(
+          :estimated_mentor_count,
+          width: 20,
+          link_errors: true,
+          value: @school_cohort.estimated_mentor_count,
+          label: {
+            text: "Number of mentors",
+            class: "govuk-label govuk-label--m"
+          }
```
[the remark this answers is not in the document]
goodviber: We need the labels, the class bolds the font. See the difference here. I think link_errors defaults to true, but don't quote me.
Screenshot 2021-04-01 at 14 56 55

app/views/schools/estimate_participants/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for :before_content, govuk_back_link(
+  text: "Back",
+  href: root_path) # TODO: get back path
+%>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l govuk-!-margin-bottom-3">How many early career teachers and mentors do you expect to have in your programme this year?</h1>
+    <p class="govuk-body">These can be rough estimates.</p>
+
+    <%= render 'form' %>
```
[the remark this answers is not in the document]
goodviber: There is only one _form within the estimate_participants folder, so we are unlikely to get confused ?

app/controllers/schools/estimate_participants_controller.rb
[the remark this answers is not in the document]
amazimbe: That's the task list but you need to generate the link to redirect to dynamically. Probably use schools_cohorts_path(start_year) or something along those lines

app/controllers/schools/estimate_participants_controller.rb
goodviber: https://medium.com/@hartaniyassir/creating-slug-urls-in-rails-without-gems-c693e0eeec8a

app/controllers/schools/estimate_participants_controller.rb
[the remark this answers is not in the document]
goodviber: I don't think there is a dynamic path for the current member route. I have raised the issue above, I suggested slugging the url param instead

app/controllers/schools/estimate_participants_controller.rb
callumacrae: ah cool, didn't know that was a thing! yeah, that seems like a good idea

app/controllers/schools/estimate_participants_controller.rb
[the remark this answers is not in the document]
goodviber: You can do schools_cohort_path(@school_cohort.cohort.start_year)

goodviber:
[Estimate participants https://github.com/DFE-Digital/early-careers-framework/pull/204](https://trello.com/c/HJ7QemF3/406-estimate-participants-https-githubcom-dfe-digital-early-careers-framework-pull-204)

app/views/schools/estimate_participants/edit.html.erb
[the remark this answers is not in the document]
kicferk1: Depends what we are doing. If we are searching for all uses of a particular partial (because we removed a field or two and want them inlined), I would hate all of our forms to live in files named `form`.

And I think in general we will have some folders that will contain more than one form - especially multi page flows - so it's unlikely naming things as `_form` will stick as a convention.

app/views/schools/estimate_participants/_form.html.erb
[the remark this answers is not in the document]
kicferk1: In your controller you do
```
  def edit
    @school_cohort_form = SchoolCohortForm.new
  end
```
You could change it to `SchoolCohortForm.from_cohort(@ school_cohort)`, add a `from_cohort` method to `SchoolCohortForm`, and then we get the correct behaviour without specifying value explicitly.

Whether it's worth the trouble, well, up to you :)

amazimbe (approved):
LGTM make sure you do the cypress tests though

config/locales/en.yml
kicferk1: There is still some here

spec/requests/schools/estimate_participants_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+        school_cohort_form: {
+          estimated_mentor_count: 20,
+          estimated_teacher_count: 30,
+        },
+      }
+      patch schools_estimate_participant_path(school_cohort), params: form_params
+      school_cohort.reload
+      expect(school_cohort.estimated_mentor_count).to eq 20
+      expect(school_cohort.estimated_teacher_count).to eq 30
+      # do we need this here, should really be part of cypress tests ?
```
kicferk1, line 35: I think checking for redirects in unit tests is fine. Cypress tests are a bit unwieldy when it comes to checking the exact url.

app/forms/school_cohort_form.rb
```diff
@@ -7,6 +7,6 @@ class SchoolCohortForm
 
   validates :estimated_mentor_count, presence: { message: "Enter your expected number of mentors" }
   validates :estimated_teacher_count, presence: { message: "Enter your expected number of ECTs" }
-  validates :estimated_mentor_count, numericality: { greater_than_or_equal_to: 0 }
-  validates :estimated_teacher_count, numericality: { greater_than_or_equal_to: 0 }
+  validates :estimated_mentor_count, numericality: { greater_than_or_equal_to: 0, less_than: 1000 }
+  validates :estimated_teacher_count, numericality: { greater_than_or_equal_to: 0, less_than: 1000 }
```
ethanmills, line 11: Probably want 
```suggestion
  validates :estimated_mentor_count, numericality: { greater_than_or_equal_to: 0, less_than: 1000, only_integer: true }
  validates :estimated_teacher_count, numericality: { greater_than_or_equal_to: 0, less_than: 1000, only_integer: true }
```

config/routes.rb
```diff
@@ -136,6 +136,7 @@
         get "/:start_year", action: :show
       end
     end
+    resources :estimate_participants, only: %i[edit update], path: "estimate-participants"
```
ethanmills, line 139: This should probably be nested inside `member` on `cohorts`, right?

spec/models/school_cohort_spec.rb
```diff
@@ -7,4 +7,22 @@
     it { is_expected.to belong_to(:cohort) }
     it { is_expected.to belong_to(:school) }
   end
+
+  it {
+    is_expected.to define_enum_for(:induction_programme_choice).with_values(
+      full_induction_programme: "full_induction_programme",
+      core_induction_programme: "core_induction_programme",
+      design_our_own: "design_our_own",
+      not_yet_known: "not_yet_known",
```
ethanmills, line 16: Do we use these two anywhere?

config/routes.rb
goodviber: That would create something like `/schools/cohorts/:id/estimate-participants/:id/edit` ?

config/locales/en.yml
goodviber: Its like a curse...

app/models/school_cohort.rb
```diff
@@ -12,7 +12,11 @@ class SchoolCohort < ApplicationRecord
   belongs_to :school
 
   def number_of_participants_status
-    "cannot start yet"
+    if !estimated_teacher_count.nil? && !estimated_mentor_count.nil?
```
kicferk1, line 15:
```
if estimated_teacher_count.present? && estimated_mentor_count.present?
```
might be a bit easier on the brain

app/views/schools/estimate_participants/edit.html.erb
```diff
@@ -0,0 +1,5 @@
+<% content_for :before_content, govuk_back_link(
+  text: "Back",
+  href: schools_cohort_path(@school_cohort.cohort.start_year))
+%>
+  <%= render 'school_cohort_form' %>
```
kicferk1, line 5: No need to indent here

kicferk1 (approved):
Two nitpicky comments from me, the whole thing looks good :)

PR #208, opened 2021-04-01
Update deploy_to_dev.yml

kicferk1:
Hopefully it's just a weird caching issue

### Context
Our app in dev environment is O L D - like 3 days old. The image it's using lives in `prod` space. I hope it's a caching issue and moving the cache back to dev is going to fix it, but who knows.


callumacrae (approved):
approving because u told me to

kicferk1:
> approving because u told me to

Thanks, really makes me confident it will work xD

For anyone wondering, I don't think there is a quick way of testing it other than merging and seeing if a new image gets deployed.

PR #209, opened 2021-04-01
Remove default school scope

kicferk1:
### Context
Default school scope that we chose a while ago, namely schools that are eligible, just isn't working for us anymore. There are many surprising places it gets in, like school import. 

1. Remove default school scope. 
2. Add `eligible` to all places making school queries without a scope. 
3. Remove `unscoped` from all school queries using it. 

### Guidance to review
If you think I missed something do let me know. 

### Testing
School import won't load more than 3k schools without it, or so I think. So if there are more in review app, it probably worked.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,3 +5,7 @@
 ### Guidance to review
 
 ### Testing
+
+### Review Checks
+- [ ] All pages have automated accessibility checks via cypress
+- [ ] All `School` queries are correctly scoped - `eligible` when they need to be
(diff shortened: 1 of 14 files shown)


amazimbe (approved):
Default scopes are evil! Never again

ethanmills (approved):
Looks good! Just need to remember to be careful and look out for `School` in review. Would it be worth adding a checkbox to the PR template asking if you've got your schools right?

kicferk1:
> School import won't load more than 3k schools without it, or so I think. So if there are more in review app, it probably worked.

I checked, there is a lot more schools than 3k.

PR #218, opened 2021-04-06
Only run webpacker once for tests

ethanmills:
### Context
Turns out if you run webpacker once up-front for all the tests things are faster than compiling every test. On my machine `rspec` goes from about 2min30 to around 17 seconds. Cypress will see a small improvement but less noticeable

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -158,6 +158,9 @@ jobs:
       - name: Set up test database
         run: bin/rails db:create db:schema:load
 
+      - name: Webpacker
+        run: bin/webpack
+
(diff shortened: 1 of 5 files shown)


ethanmills:
@callumacrae this probably means you will need to run `bin/webpack-dev-server` in the background while developing. Your sacrifice is greatly appreciated :P

kicferk1 (approved):

ethanmills:
I've also disabled screenshots and video in cypress when calling from `yarn cypress:run`, which should save a few seconds

callumacrae (approved):

PR #219, opened 2021-04-07
Add a unique index on cohorts start year

amazimbe:
### Context
I had an exciting time debugging locally and then realised that the cause of the weirdness was that I had 2 cohorts with start year 2021!

In addition, we are doing a lot of look ups on cohorts using the start year. Actually more than we do with the cohort id.

Add a unique index on the cohort start year.


amazimbe:
[Bug: make start year on a cohort unique](https://trello.com/c/1D2w1omL/457-bug-make-start-year-on-a-cohort-unique)

ethanmills (approved):

PR #220, opened 2021-04-07
Remove registration

ethanmills:
### Context
I think it's time to remove the registration journey

### Review Checks


kicferk1 (approved):
One thing I like about this journey is how easy it is to burn it away.

app/controllers/registrations/user_profile_controller.rb
```diff
@@ -1,36 +0,0 @@
...
-    @user = User.new(user_params)
-
-    ActiveRecord::Base.transaction do
-      @user.save!
-      @school.induction_coordinator_profiles.destroy_all
-      InductionCoordinatorProfile.create!(user: @user, schools: [@school])
-    end
-
-    session.delete(:school_id)
-    redirect_to :registrations_verification_sent
```
kicferk1, line 21: Do we have some cypress helpers / mailer stuff that needs to go too?

And maybe some templates in notify?

amazimbe (approved):
LGTM!

app/controllers/registrations/user_profile_controller.rb
ethanmills: Good shout

ethanmills: That's a big change, I'll remove in another PR

PR #221, opened 2021-04-07
Remove confirmations from users

ethanmills:
### Context
We don't use this anywhere, and it adds a load of boilerplate and confusion

### Review Checks


amazimbe (approved):

app/models/user.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class User < ApplicationRecord
-  devise :registerable, :trackable, :confirmable, :passwordless_authenticatable
+  devise :registerable, :trackable, :passwordless_authenticatable
```
kicferk1, line 4: soon we will just have our own `passwordless_authenticatable` here :D

kicferk1 (approved):

PR #224, opened 2021-04-08
Add a db reset task for dev environments

kicferk1:
### Context
We want a way to re-create the database for testing and re-testing purposes.

Use `db:reset` task, https://stackoverflow.com/a/10302357
Wrap it in an environment check, so the task we use won't work on staging or prod.

### Guidance to review
I went with a rake task, because the last thing I want to worry about is foreign key constraints. I sure hope our paas user lets us do that.
This PR adds a db reset on a rake task, and on deploy on dev and review apps.

### Testing
I want a PR app to see it working in practice. But that might be hard to do since our dockerhub is malfunctioning atm...

### Review Checks

### DIFF
--- a/lib/tasks/safe_reset.rake
+++ b/lib/tasks/safe_reset.rake
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+namespace :db do
+  desc "Drop, create, migrate then seed the development database"
+  task safe_reset: :environment do
+    if Rails.env.development? || Rails.env.deployed_development?
+      Rake::Task["db:reset"].invoke
+      puts "Reseeded the database!"
+    else
+      puts "You should think twice before recreating staging or production database!"
+      puts "Use 'db:reset' task to do it if you really want to."
+    end
+  end
+end
--- a/lib/tasks/safe_reset.rake
+++ b/lib/tasks/safe_reset.rake
@@ -12,7 +12,7 @@ namespace :db do
       tables.delete "schema_migrations"
       tables.each { |table| connection.execute("TRUNCATE #{table} CASCADE;") }
       Rake::Task["db:seed"].invoke
-      puts "Reseeded the database!"
+      puts "Re-seeded the database!"
     else
       puts "You should think twice before recreating staging or production database!"
       puts "Use 'db:reset' task to do it if you really want to."
--- a/lib/tasks/safe_reset.rake
+++ b/lib/tasks/safe_reset.rake
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 namespace :db do
-  desc "Drop, create, migrate then seed the development database"
+  desc "Remove all database content, then seed it. Only for dev environments."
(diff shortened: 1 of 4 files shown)


kicferk1:
Alright, I checked that a delivery partner was wiped on the review app after re-deploying. That should do the trick for our dev environment too.

lib/tasks/safe_reset.rake
```diff
@@ -0,0 +1,21 @@
...
+namespace :db do
+  desc "Drop, create, migrate then seed the development database"
+  task safe_reset: :environment do
+    if Rails.env.development? || Rails.env.deployed_development?
+      connection = ActiveRecord::Base.connection
+      tables = connection
+                   .execute("SELECT * FROM pg_catalog.pg_tables;")
+                   .filter { |row| row["schemaname"].include?("public") }
+                   .map { |row| row["tablename"] }
+      tables.delete "schema_migrations"
```
amazimbe, line 12: why do we need to delete this?

lib/tasks/safe_reset.rake
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+namespace :db do
+  desc "Drop, create, migrate then seed the development database"
```
amazimbe, line 4: We are more like resetting than dropping, creating and migrating

lib/tasks/safe_reset.rake
ethanmills: Looks like this removes that table from the list, since we don't want to run all the migrations again since the tables themselves aren't changing, just being truncated. Therefore we don't want to truncate `schema_migrations`. This looks correct to me

lib/tasks/safe_reset.rake
```diff
@@ -0,0 +1,21 @@
...
+                   .execute("SELECT * FROM pg_catalog.pg_tables;")
+                   .filter { |row| row["schemaname"].include?("public") }
+                   .map { |row| row["tablename"] }
+      tables.delete "schema_migrations"
+      tables.each { |table| connection.execute("TRUNCATE #{table} CASCADE;") }
+      Rake::Task["db:seed"].invoke
+      puts "Re-seeded the database!"
+    else
+      puts "You should think twice before recreating staging or production database!"
+      puts "Use 'db:reset' task to do it if you really want to."
```
ethanmills, line 18: This won't work, will it? I thought you tried a reset first and it didn't work?

kicferk1: Ah yes, I wrote this comment before I tested it and forgot to change it 😂

lib/tasks/safe_reset.rake
kicferk1: Ethan is right. We want to remove all contents of our tables, but `migrations` are a bit 'meta'. Not really our table.

amazimbe (approved):

ethanmills (approved):

PR #226, opened 2021-04-08
ECF-RP-404: nominate induction coordinator

ethanmills:
## Context
This is a rework of Seb's PR to nominate an induction coordinator

#### Request specs
- Check the correct redirects happen
- Check the correct templates are rendered
- High level sanity checks on model creation
#### Form specs
-Validation checks
- Detailed model creation checks
#### Model specs
- Check model is created correctly
- Mailer checks
#### Cypress tests
- Valid nomination (success journey)
- Expired link
- Already nominated
- Email in use

### Manual testing
I'm going to write some seed data for this and other scenarios, but that will be a separate PR 

## Review Checks

### DIFF
--- a/app/controllers/nominations/nominate_induction_coordinator_controller.rb
+++ b/app/controllers/nominations/nominate_induction_coordinator_controller.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+class Nominations::NominateInductionCoordinatorController < ApplicationController
+  def new
+    @token = params[:token]
+    @nomination_email = NominationEmail.find_by(token: @token)
+
+    if @nomination_email.nil?
+      redirect_to link_invalid_nominate_induction_coordinator_path
+    elsif @nomination_email.expired?
+      redirect_to link_expired_nominate_induction_coordinator_path(school_id: @nomination_email.school_id)
+    elsif @nomination_email.school.fully_registered?
+      redirect_to already_nominated_request_nomination_invite_path
+    else
+      load_nominate_induction_tutor_form
+      @nominate_induction_tutor_form.token = @token
+    end
+  end
+
+  def create
+    load_nominate_induction_tutor_form
+
+    render :new and return unless @nominate_induction_tutor_form.valid?
+
+    @nominate_induction_tutor_form.save!
+    redirect_to nominate_school_lead_success_nominate_induction_coordinator_path
+  rescue UserExistsError
+    redirect_to email_used_nominate_induction_coordinator_path(token: @nominate_induction_tutor_form.token)
+  end
+
+  def link_expired
+    @school_id = params[:school_id]
+  end
+
+  def resend_email_after_link_expired
+    nomination_request_form = build_nomination_request_form
+    nomination_request_form.save!
+
+    redirect_to success_request_nomination_invite_path
+  rescue TooManyEmailsError
+    redirect_to limit_reached_request_nomination_invite_path
+  end
+
+  def email_used
+    @token = params[:token]
+  end
+
+  def nominate_school_lead_success; end
+
+private
+
+  def load_nominate_induction_tutor_form
+    @nominate_induction_tutor_form = ::NominateInductionTutorForm.new
+    @nominate_induction_tutor_form.assign_attributes(nominate_induction_tutor_form_params)
+  end
+
+  def nominate_induction_tutor_form_params
+    return {} unless params.key?(:nominate_induction_tutor_form)
+
+    params.require(:nominate_induction_tutor_form).permit(:full_name, :email, :token)
+  end
+
+  def build_nomination_request_form
+    school = School.find(params[:resend_email_after_link_expired][:school_id])
+    NominationRequestForm.new(
+      local_authority_id: school.local_authority.id,
+      school_id: school.id,
+    )
+  end
+end
--- a/app/controllers/nominations/nominate_induction_coordinator_controller.rb
+++ b/app/controllers/nominations/nominate_induction_coordinator_controller.rb
@@ -3,13 +3,13 @@
 class Nominations::NominateInductionCoordinatorController < ApplicationController
   def new
     @token = params[:token]
-    @nomination_email = NominationEmail.find_by(token: @token)
+    nomination_email = NominationEmail.find_by(token: @token)
 
-    if @nomination_email.nil?
+    if nomination_email.nil?
       redirect_to link_invalid_nominate_induction_coordinator_path
-    elsif @nomination_email.expired?
-      redirect_to link_expired_nominate_induction_coordinator_path(school_id: @nomination_email.school_id)
-    elsif @nomination_email.school.registered?
+    elsif nomination_email.expired?
+      redirect_to link_expired_nominate_induction_coordinator_path(school_id: nomination_email.school_id)
+    elsif nomination_email.school.registered?
       redirect_to already_nominated_request_nomination_invite_path
     else
       load_nominate_induction_tutor_form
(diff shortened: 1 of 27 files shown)


ethanmills:
[User can nominate an induction coordinator for their school by providing their name and email address on the service](https://trello.com/c/elONU0Ua/404-user-can-nominate-an-induction-coordinator-for-their-school-by-providing-their-name-and-email-address-on-the-service)

app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+  end
+
+  def create
+    load_nominate_induction_tutor_form
+
+    render :new and return unless @nominate_induction_tutor_form.valid?
+
+    @nominate_induction_tutor_form.save!
+    redirect_to nominate_school_lead_success_nominate_induction_coordinator_path
```
amazimbe, line 26: Can't we just call this success_nominate_induction_coordinator_path ?

app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+class Nominations::NominateInductionCoordinatorController < ApplicationController
+  def new
+    @token = params[:token]
+    @nomination_email = NominationEmail.find_by(token: @token)
+
+    if @nomination_email.nil?
+      redirect_to link_invalid_nominate_induction_coordinator_path
+    elsif @nomination_email.expired?
+      redirect_to link_expired_nominate_induction_coordinator_path(school_id: @nomination_email.school_id)
+    elsif @nomination_email.school.fully_registered?
```
amazimbe, line 12: Is it worth changing this to registered? fully_registered? was good when we had partially_registered? now it just raises unnecessary questions.

app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    nomination_request_form = build_nomination_request_form
+    nomination_request_form.save!
+
+    redirect_to success_request_nomination_invite_path
+  rescue TooManyEmailsError
+    redirect_to limit_reached_request_nomination_invite_path
+  end
+
+  def email_used
+    @token = params[:token]
```
amazimbe, line 45: The method name doesn't correlate with what it does. I'm I missing something?

app/controllers/nominations/nominate_induction_coordinator_controller.rb
ethanmills: There's already a `nominations/success` from requesting a nomination invite email.

amazimbe (approved):

app/views/nominations/nominate_induction_coordinator/nominate_school_lead_success.html.erb
```diff
@@ -0,0 +1,14 @@
+<% content_for :title, "Nomination complete" %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_panel title: "Induction tutor nominated", body: nil, classes: "govuk-!-margin-bottom-8", html_attributes: { "data-test": "success-panel" } %>
```
kicferk1, line 5: I am sad every time I need to pass a `nil` as an argument :P

spec/requests/nominations/nominate_induction_coordinator_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+        before do
+          school = nomination_email.school
+          create(:user, :induction_coordinator, schools: [school])
+        end
+
+        it "redirects to already-nominated" do
+          get "/nominations/new?token=#{token}"
+
+          expect(response).to redirect_to("/nominations/already-nominated")
+          follow_redirect!
```
kicferk1, line 37: Huh, interesting. Good to know the helper is there :)

kicferk1 (approved):
Tiny comment about variable scope, cause it's a pet peeve of mine.

app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    if @nomination_email.nil?
+      redirect_to link_invalid_nominate_induction_coordinator_path
+    elsif @nomination_email.expired?
+      redirect_to link_expired_nominate_induction_coordinator_path(school_id: @nomination_email.school_id)
+    elsif @nomination_email.school.registered?
+      redirect_to already_nominated_request_nomination_invite_path
+    else
+      load_nominate_induction_tutor_form
+      @nominate_induction_tutor_form.token = @token
+    end
```
BroiSatse, line 17: why not:

```
    redirect_to link_invalid_nominate_induction_coordinator_path  if @nomination_email.nil?
    redirect_to link_expired_nominate_induction_coordinator_path(school_id: @nomination_email.school_id) if @nomination_email.expired?
    redirect_to already_nominated_request_nomination_invite_path if @nomination_email.school.registered?
    
    @nominate_induction_tutor_form.token = @token
```

app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+class Nominations::NominateInductionCoordinatorController < ApplicationController
+  def new
+    @token = params[:token]
+    @nomination_email = NominationEmail.find_by(token: @token)
```
BroiSatse, line 6: Out of interest - should we encrypt the token in the database? It feels a bit like we're storing plain-text passwords here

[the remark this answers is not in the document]
ethanmills: Yep, we get `school` from it in views

amazimbe: I think that would be overkill and I don't think it would be any more secure than what we currently have. The issue would still be that anyone who gets the email and clicks on the link can claim the school even if we add encryption.

app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+class Nominations::NominateInductionCoordinatorController < ApplicationController
+  def new
+    @token = params[:token]
+    nomination_email = NominationEmail.find_by(token: @token)
```
BroiSatse, line 6: This feels a bit as staroing plain-text password. Should we bcrypt the token?

ethanmills: I take your point, but I'm happy to leave this as is for now, since it's ephemeral. @cfq any thoughts?

ethanmills: Whatever we do here we should also do with magiclink sign in tokens

ethanmills: I think there's also the possibility of timing attacks, since find_by isn't constant time.

cfq: Yes, and only in dev environments.

BroiSatse: Ah, no definitely should be tackled, if at all, in another PR

app/controllers/nominations/nominate_induction_coordinator_controller.rb
BroiSatse: The problem is when attacker manages in any way to read the database. In this case, as the tokens are not hashed, he automatically can impersonate anyone we send those tokens to. Naturally it is not as bad as having un-hashed passwords (or, as mentioned in the link below - password reset tokens) as it will only give an access to limited part of the site. But, hackers are freakishly clever. Hashing one-time tokens is pretty much a security standard now-a-days, Devise does it automatically for all their token-led emails (reset, confirmation, invitations etc)

https://security.stackexchange.com/questions/86913/should-password-reset-tokens-be-hashed-when-stored-in-a-database

PR #227, opened 2021-04-09
ECF-1-402 Added notify callback

amazimbe:
### Context
We want to track the status of the nomination emails we send to schools.

I have explored 2 approaches. The first, https://docs.notifications.service.gov.uk/ruby.html#get-message-status, involves making API calls to Notify via the Notify client. The call would give us the latest 250 emails sent in the last 7 days. Based on the oldest of those records we would then need to fetch the next 250 records older than the oldest in the previous batch. It would be more difficult to get the status of all our nomination emails as there could be thousands of them. We would also need a rake task to run as a cron job to keep updating the statuses of our emails.

The other approach is to let Notify let us know the status of our emails via a callback and this is what has been implemented in this pr. https://docs.notifications.service.gov.uk/ruby.html#callbacks

### DIFF
--- a/app/controllers/nominations/notify_callbacks_controller.rb
+++ b/app/controllers/nominations/notify_callbacks_controller.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class Nominations::NotifyCallbacksController < ActionController::API
+  def create
+    email = NominationEmail.find_by(token: params[:reference])
+    email.update!(notify_status: params[:status])
+
+    head :no_content
+  end
+end
--- a/app/controllers/nominations/notify_callbacks_controller.rb
+++ b/app/controllers/nominations/notify_callbacks_controller.rb
@@ -2,9 +2,27 @@
 
 class Nominations::NotifyCallbacksController < ActionController::API
   def create
+    return head :no_content unless params[:reference]
+
     email = NominationEmail.find_by(token: params[:reference])
-    email.update!(notify_status: params[:status])
+    email.update!(notify_status: params[:status]) if email
+    log_email if failed_email?(email) 
 
     head :no_content
+  end
+
+  private
+
+  def log_email
+    Rails.logger.info "Email could not be sent"
+    Rails.logger.info "notify_id: #{params[:id]}"
+    Rails.logger.info "reference: #{params[:reference]}"
+    Rails.logger.info "template_id: #{params[:template_id]}"
+  end
+
+  def failed_email?(email)
+    return false if email.nil?
+  
+    params[:status] != "sending" && params[:status] != "delivered"
   end
 end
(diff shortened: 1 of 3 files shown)


amazimbe:
[Retrieve bounced emails from Notify](https://trello.com/c/1olAaPrS/402-retrieve-bounced-emails-from-notify)

ethanmills:
There are lots of emails that don't have a reference (e.g. sign in emails) Will this cause any trouble? It might be nice to log failures of those other types of emails

ethanmills:
From the docs, the statuses are delivered, permanent-failure, temporary-failure or technical-failure, so logging anything that isn't delivered would be good

kicferk1:
Is there a way to add more than one callback in notify? 

image

is what I am seeing, and it does not fill me with hope that we will be able to have multiple callbacks 😢 

And I really don't want to store email info cross-environments.
Will we need separate notify instances for prod, staging and dev? @cfq

amazimbe:
> There are lots of emails that don't have a reference (e.g. sign in emails) Will this cause any trouble? It might be nice to log failures of those other types of emails

Yes, I just remembered that and it needs to be taken care of at least to make sure it doesn't cause any errors

amazimbe:
> Is there a way to add more than one callback in notify?
> 
> image
> 
> is what I am seeing, and it does not fill me with hope that we will be able to have multiple callbacks 😢
> 
> And I really don't want to store email info cross-environments.
> Will we need separate notify instances for prod, staging and dev? @cfq

@kicferk1 no, we can only have one callback and the more I think about it the more it makes sense to me. I don't think we need to track if emails are delivered in dev or staging.

kicferk1:
> @kicferk1 no, we can only have one callback and the more I think about it the more it makes sense to me. I don't think we need to track if emails are delivered in dev or staging but we could setup separate notifys for them if necessary

Well, but it makes it hard for us to test that email tracking works. Which is (I imagine) why you set it to point at dev environment for time being - to test that integration is working as expected.

Overall, I think notify does not believe in the concept of environments, or at least doesn't think about them too hard. So I wonder if it would be a good idea to have two other notify services, both non-live I imagine, one for dev and one for staging.

app/controllers/nominations/notify_callbacks_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    head :no_content
+  end
+
+private
+
+  def log_email
+    Rails.logger.info "Email could not be sent"
+    Rails.logger.info "notify_id: #{params[:id]}"
+    Rails.logger.info "reference: #{params[:reference]}"
+    Rails.logger.info "template_id: #{params[:template_id]}"
```
ethanmills, line 20: Would be nice to have this as one log, to make filtering the logs easier. What do you think?

ethanmills (approved):

ethanmills (approved):
Looks good!

PR #229, opened 2021-04-09
Exclude more tables from deletion

kicferk1:
### Context
I just broke dev because it is subtly different to review apps - there is another important table there :/


lib/tasks/safe_reset.rake
```diff
@@ -10,6 +10,7 @@ namespace :db do
                    .filter { |row| row["schemaname"].include?("public") }
```
ethanmills, line 10: While we're here, why not just `SELECT * FROM pg_catalog.pg_tables WHERE schemaname = 'public';`?

ethanmills (approved):
1 optional change

amazimbe (approved):

PR #234, opened 2021-04-09
Nomination email link tracking

BroiSatse:
### Context

When school user has received an email for nomination induction coordinator for their school, we want to track the time they have started the nomination process.

### Changes proposed in this pull request.

When link is clicked, user will be directed to a start page with a single `start` button. Once the button is clicked, notification_email#opened_at column is updated when user is presented with the nomination form.

### Review notes
This PR is depending on https://github.com/DFE-Digital/early-careers-framework/pull/226

### Testing
1) Receive school invite email with nomination link (via `rake schools:send_invites[school_urn]`)
1) Click on the nomination link
1) You should see a nomination start page, showing school's details
1) Check the `opened_at` field of the nomination_email record matching the token from url
1) Click on start button
1) You should see nomination form
1) Check the `opened_at` field has been populated with current time
1) Click the same nomination link in the email again and follow through the start button again
1) Check that the `opened_at` field has not changed

### Review Checks

### DIFF
--- a/app/controllers/nominations/nominate_induction_coordinator_controller.rb
+++ b/app/controllers/nominations/nominate_induction_coordinator_controller.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 class Nominations::NominateInductionCoordinatorController < ApplicationController
-  def new
-    @token = params[:token]
-    nomination_email = NominationEmail.find_by(token: @token)
+  def start
+    token = params[:token]
+    nomination_email = NominationEmail.find_by(token: token)
 
     if nomination_email.nil?
       redirect_to link_invalid_nominate_induction_coordinator_path
@@ -13,8 +13,14 @@ class Nominations::NominateInductionCoordinatorController < ApplicationControlle
       redirect_to already_nominated_request_nomination_invite_path
     else
       load_nominate_induction_tutor_form
-      @nominate_induction_tutor_form.token = @token
+      @nominate_induction_tutor_form.token = token
+      session[:nominate_induction_tutor_form] = @nominate_induction_tutor_form
     end
+  end
+
+  def new
+    load_nominate_induction_tutor_form
+    record_nomination_email_opened
   end
 
   def create
@@ -25,7 +31,7 @@ class Nominations::NominateInductionCoordinatorController < ApplicationControlle
     @nominate_induction_tutor_form.save!
     redirect_to nominate_school_lead_success_nominate_induction_coordinator_path
   rescue UserExistsError
-    redirect_to email_used_nominate_induction_coordinator_path(token: @nominate_induction_tutor_form.token)
+    redirect_to email_used_nominate_induction_coordinator_path
   end
 
   def link_expired
@@ -50,7 +56,7 @@ class Nominations::NominateInductionCoordinatorController < ApplicationControlle
 private
 
   def load_nominate_induction_tutor_form
-    @nominate_induction_tutor_form = ::NominateInductionTutorForm.new
+    @nominate_induction_tutor_form = ::NominateInductionTutorForm.new(session[:nominate_induction_tutor_form])
     @nominate_induction_tutor_form.assign_attributes(nominate_induction_tutor_form_params)
   end
 
@@ -58,6 +64,12 @@ private
     return {} unless params.key?(:nominate_induction_tutor_form)
 
     params.require(:nominate_induction_tutor_form).permit(:full_name, :email, :token)
+  end
+
+  def record_nomination_email_opened
+    NominationEmail
+      .where(token: @nominate_induction_tutor_form.token, opened_at: nil)
+      .update_all(opened_at: Time.zone.now)
   end
 
   def build_nomination_request_form
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -24,11 +24,11 @@ class School < ApplicationRecord
   scope :eligible, -> { open.eligible_establishment_type.in_england }
 
   scope :with_name_like, lambda { |search_key|
-    School.eligible.where("name ILIKE ?", "%#{search_key}%")
+    where("name ILIKE ?", "%#{search_key}%")
   }
 
   scope :with_urn_like, lambda { |search_key|
-    School.eligible.where("urn ILIKE ?", "%#{search_key}%")
+    where("urn ILIKE ?", "%#{search_key}%")
   }
 
   scope :search_by_name_or_urn, lambda { |search_key|
@@ -41,15 +41,15 @@ class School < ApplicationRecord
   }
 
   scope :partnered, lambda { |year|
-    where(id: Partnership.joins(:cohort).where(cohorts: { start_year: year }).pluck(:school_id))
+    where(id: Partnership.joins(:cohort).where(cohorts: { start_year: year }).select(:school_id))
   }
 
   scope :partnered_with_lead_provider, lambda { |lead_provider_id|
-    where(id: Partnership.where(lead_provider_id: lead_provider_id).pluck(:school_id))
+    where(id: Partnership.where(lead_provider_id: lead_provider_id).select(:school_id))
   }
 
   scope :unpartnered, lambda { |year|
-    where.not(id: Partnership.joins(:cohort).where(cohorts: { start_year: year }).pluck(:school_id))
+    where.not(id: Partnership.joins(:cohort).where(cohorts: { start_year: year }).select(:school_id))
   }
 
   def lead_provider(year)
@@ -119,7 +119,7 @@ private
   end
 
   def in_england?
-    administrative_district_code.match?(/^[Ee].*/)
+    administrative_district_code.match?(/^[Ee]/)
   end
 
   scope :open, -> { where(school_status_code: ELIGIBLE_STATUS_CODES) }
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -4,7 +4,6 @@ class InviteSchools
   EMAIL_COOLDOWN_PERIOD = 24.hours
 
   def run(school_urns)
-    logger = Rails.logger
     logger.info "Emailing schools"
 
     school_urns.each do |urn|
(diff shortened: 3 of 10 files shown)


app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -1,22 +1,28 @@
...
-      @nominate_induction_tutor_form.token = @token
+      @nominate_induction_tutor_form.token = token
+      session[:nominate_induction_tutor_form] = @nominate_induction_tutor_form
     end
   end
 
+  def new
+    load_nominate_induction_tutor_form
+    record_nomination_email_opened
+  end
```
BroiSatse, line 24: Another approach would be to have another `receive_start` action which does the update and redirects to new. It is a trade-off between 1) a single (potentially) no-op db call on each `new` call and 2) additional action and route

app/models/school.rb
```diff
@@ -24,11 +24,11 @@ class School < ApplicationRecord
   scope :eligible, -> { open.eligible_establishment_type.in_england }
 
   scope :with_name_like, lambda { |search_key|
-    School.eligible.where("name ILIKE ?", "%#{search_key}%")
+    where("name ILIKE ?", "%#{search_key}%")
```
amazimbe, line 27: Pretty sure we need to scope all school queries with eligible

BroiSatse: Very much yes, but this is not a query but a scope definition. I would not expect `School.with_name_like('CoE')` to do any additional, surprising filtering, but I agree it should always be called together with `eligible` scope.

ethanmills: Are you confident that this does not introduce a behaviour change? Are you confident that everywhere that calls this scope also calls eligible? I agree with the change, I just want to make sure we're not introducing regressions

app/models/school.rb
```diff
@@ -41,15 +41,15 @@ class School < ApplicationRecord
   }
 
   scope :partnered, lambda { |year|
-    where(id: Partnership.joins(:cohort).where(cohorts: { start_year: year }).pluck(:school_id))
+    where(id: Partnership.joins(:cohort).where(cohorts: { start_year: year }).select(:school_id))
```
ethanmills, line 44: Why?

app/controllers/nominations/nominate_induction_coordinator_controller.rb
ethanmills: I think I prefer this

app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -1,22 +1,28 @@
...
-    elsif @nomination_email.school.registered?
+    elsif nomination_email.expired?
+      redirect_to link_expired_nominate_induction_coordinator_path(school_id: nomination_email.school_id)
+    elsif nomination_email.school.registered?
       redirect_to already_nominated_request_nomination_invite_path
     else
       load_nominate_induction_tutor_form
-      @nominate_induction_tutor_form.token = @token
+      @nominate_induction_tutor_form.token = token
+      session[:nominate_induction_tutor_form] = @nominate_induction_tutor_form
```
ethanmills, line 17: If you're saving the form in the session, I think we should delete it on success, just for completeness. Also, we wouldn't need to pass parameters to `#email_used`, we could just load the form and use the info from there

app/models/school.rb
BroiSatse: Not-related minor improvement. `where(id: other_scope.pluck(:id)` fires two sql queries. First one `SELECT id FROM <subscope>` and second `SELECT * FROM <scope> WHERE id IN (<massive list of ids fetched from first qury>)`. Using `select` AR performs a single query with subquery: `SELECT * FROM <scope> WHERE id in (SELECT id from <subscope>)`. I had some really nasty experience with AR cache memory leaking due to caching every single variant of `WHERE id in (<list of ids>)`.

spec/requests/nominations/nominate_induction_coordinator_spec.rb
```diff
@@ -45,13 +45,46 @@
...
-        get "/nominations/new?token=#{token}"
+        get "/nominations/start?token=#{token}"
 
         expect(response).to redirect_to("/nominations/link-expired?school_id=#{nomination_email.school.id}")
       end
     end
   end
 
+  describe "GET /nominations/new" do
+    let(:now) { rand(-100..100).hours.ago }
```
ethanmills, line 56: Why? Daylight savings?

app/models/school.rb
ethanmills: Nice, good to know. Thanks!

app/controllers/nominations/nominate_induction_coordinator_controller.rb
BroiSatse: Good point! Will do.

spec/requests/nominations/nominate_induction_coordinator_spec.rb
BroiSatse: This is to decouple the test from the actual time to avoid potential time-related issues (both positive and negatives). I normally like to introduce a bit of randomness into the tests as it forces me to write the tests in a more resilient way. Additionally, they do sometimes catch some edge cases, which is an extra bonus.

ethanmills: Fine by me, just like hearing rationale to learn new things 🙂

app/models/school.rb
BroiSatse: Those scopes are both only used within `search_by_name_or_urn` which in turn is only used in SchoolSearchController together with `eligible` scope. Naturally there is never 100% confidence (especially after being 3 days on the new project :P), but I am pretty sure it should not introduce behavioural change.

ethanmills: FWIW I think the answer is "yes" and this is fine :)

ethanmills (approved):
Happy with this once the "#email_used" change is in

ethanmills:
When you're ready let's merge this into ECF-RP-404_nominate-induction-coordinator and they can go through QA and POR together on the review app for that

app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -60,6 +66,12 @@ def nominate_induction_tutor_form_params
     params.require(:nominate_induction_tutor_form).permit(:full_name, :email, :token)
   end
 
+  def record_nomination_email_opened
+    NominationEmail
+      .where(token: @nominate_induction_tutor_form.token, opened_at: nil)
```
amazimbe, line 71: Reading this gives the impression that you are updating multiple records which is not the case. Why not just use find_by with update?

app/services/invite_schools.rb
```diff
@@ -46,23 +46,24 @@ def send_nomination_email(recipient, token, school_name)
   end
 
   def recipient_email(school)
-    return school.secondary_contact_email unless school.primary_contact_email
-
-    school.primary_contact_email
+    school.primary_contact_email || school.secondary_contact_email
```
amazimbe, line 49: 👍

amazimbe:
I think you'd be better off cherry picking your commits on to a fresh branch rather than merging

app/controllers/nominations/nominate_induction_coordinator_controller.rb
BroiSatse: Yeah, I was on the fence here. Since I only do the update here, I do not need to instantiate AR object at all and can just o a single `UPDATE` query (instead of two: SELECT and UPDATE). Additionally, as written it will automatically bypass all the model validations, which is, surprisingly, correct in this context (I don't want to miss this write only because we added new validation on the model?) - this will probably never be a case though.

BroiSatse:
Good point. This is why I am normally against rewriting git history. @ethanmills Given that target branch is yours I'd think it would be best either for you to cherry pick the only commit from here or we could pair-up for conflict resolution

app/controllers/nominations/nominate_induction_coordinator_controller.rb
BroiSatse: @ethanmills I believe this should now be done :)

app/controllers/nominations/nominate_induction_coordinator_controller.rb
kicferk1: So the part about number of queries is interesting. I would probably err on the side of code readability and expressiveness over database optimisations that save 1 query. Because premature optimisation is evil - I imagine we will have load tests, and I am 99% sure that this is not gonna be the place that cracks first.
Saving N+1 queries, well, that would be more compelling. 

As for bypassing validation, it looks like an evil side effect of inner Rails machinations. We should have tests for this flow (both in cucumber and request tests) one of which should test whether this email was updated.

app/views/nominations/nominate_induction_coordinator/start.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1>Nominate an induction tutor</h1>
+
+<%= render partial: 'nominations/school_details', locals: { school: @nominate_induction_tutor_form.school, title: "Your school" } %>
+
+This is someone who is responsible for running induction at your school.
+
+They will:
+<ul>
```
kicferk1, line 8: Do we want some govuk design system class on it, like `govuk-list`?

kicferk1 (approved):
Looks good to me

app/controllers/nominations/nominate_induction_coordinator_controller.rb
amazimbe: I completely agree with @kicferk1 In addition, the moment someone adds validations to this model - which is very likely - that will be the first thing to break.

amazimbe (approved):

ethanmills:
[Track when the unique link has been visited](https://trello.com/c/bW1VdhG5/403-track-when-the-unique-link-has-been-visited)

PR #238, opened 2021-04-12
back link from cookies page

goodviber:
### Context
Trello #485
We need to return to the previous page after visiting the"/cookies" page and viewing/updating our preferences
Introduces an application controller method to set session[:return_to] in all get actions apart from under CookiesController. Can also be extended to privacy policy if needed ?
### Guidance to review
There are several ways of doing this...not all cover all use cases.
### Testing
Noted some of the CookieFeature cypress tests are failing locally, as @consent_updated is not being set...
### Review Checks

### DIFF
--- a/app/controllers/cookies_controller.rb
+++ b/app/controllers/cookies_controller.rb
@@ -16,8 +16,8 @@ class CookiesController < ApplicationController
     respond_to do |format|
       format.html do
         set_cookie_form
-        @consent_updated = true # TODO: remove this and style notice
-        redirect_to cookies_path, notice: "You’ve set your cookie preferences."
+        set_success_message(content: "You’ve set your cookie preferences.", title: "Success")
+        redirect_to cookies_path
       end
 
       format.json do
--- a/app/controllers/cookies_controller.rb
+++ b/app/controllers/cookies_controller.rb
@@ -16,8 +16,8 @@ class CookiesController < ApplicationController
     respond_to do |format|
       format.html do
         set_cookie_form
-        @consent_updated = true
-        redirect_to cookies_path, notice: "Cookie preferences upated successfully"
+        @consent_updated = true # TODO: remove this and style notice
+        redirect_to cookies_path, notice: "You’ve set your cookie preferences."
       end
 
       format.json do
--- a/app/controllers/cookies_controller.rb
+++ b/app/controllers/cookies_controller.rb
@@ -3,7 +3,9 @@
 class CookiesController < ApplicationController
   before_action :set_cookie_form, only: :show
 
-  def show; end
+  def show
+    @backlink = session[:return_to]
+  end
 
   def update
     analytics_consent = params[:cookies_form][:analytics_consent]
@@ -15,7 +17,7 @@ class CookiesController < ApplicationController
       format.html do
         set_cookie_form
         @consent_updated = true
-        render :show
+        redirect_to cookies_path, notice: "Cookie preferences upated successfully"
       end
 
       format.json do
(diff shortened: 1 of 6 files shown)


callumacrae:
What's this required for? Unless the user doesn't have JavaScript, they will not be taken to the cookie page anyway.

app/views/cookies/show.html.erb
```diff
@@ -1,3 +1,8 @@
+<%= render GovukComponent::BackLink.new(
+      text: 'Back',
+      href: @backlink,
+    )
+%>
```
ethanmills, line 5: Please follow how this is done elsewhere, e.g.
`<% content_for :before_content, govuk_back_link(
  text: "Back",
  href: @backlink)
%>`

app/views/privacy_policies/show.html.erb
```diff
@@ -1,3 +1,8 @@
+<%= render GovukComponent::BackLink.new(
+      text: 'Back',
+      href: :back,
```
ethanmills, line 3: Why is this one different?

app/views/cookies/show.html.erb
goodviber: It does leave a lot of space until the next element ?

ethanmills: Please see this: https://design-system.service.gov.uk/components/back-link/#how-it-works:~:text=Always%20place%20back%20links%20at%20the,navigation%20links%2C%20including%20the%20back%20link

app/views/privacy_policies/show.html.erb
ethanmills: Would it not be simpler to set request.referer on the session as part of show? Rather than running this before every action on every controller?

goodviber:
[- it's done when a user on the cookies page can click a back button which takes them to the page they were last on before accessing the cookies page](https://trello.com/c/uATxNEYY/485-its-done-when-a-user-on-the-cookies-page-can-click-a-back-button-which-takes-them-to-the-page-they-were-last-on-before-accessin)

app/views/privacy_policies/show.html.erb
goodviber: That was my first thought, but it doesn't fit every use case. When we update the form, the request.referer becomes "/cookies" ...

amazimbe (approved):

app/controllers/cookies_controller.rb
```diff
@@ -14,8 +16,8 @@ def update
     respond_to do |format|
       format.html do
         set_cookie_form
-        @consent_updated = true
-        render :show
+        @consent_updated = true # TODO: remove this and style notice
```
ethanmills, line 19: We have components for this

ethanmills: Search for `set_success_message`

app/views/privacy_policies/show.html.erb
ethanmills: I think personally I would prefer containing this all to the cookies controller (by not updating the session if the referer is `/cookies`) but I don't feel strongly enough to insist you change this

kicferk1 (commented):
The behaviour seems good to me. Let's name our action a little better, and let's address your own TODOs :)

app/views/privacy_policies/show.html.erb
goodviber: I don't mind changing it, but it seems more complicated than it looks to contain it all in the cookies controller. I  you have a working suggestion that will fit all use cases, happy to implement?

ethanmills (approved):

PR #239, opened 2021-04-12
Run a rake task on an app in the dev space

ethanmills:
### Context
This should make it easier to run tasks in the dev cf space. I see this being used mostly for resetting the database on dev and review apps.

### DIFF
--- a/.github/workflows/run_task_dev.yml
+++ b/.github/workflows/run_task_dev.yml
@@ -1,0 +1,27 @@
+name: Run task in dev space
+
+on:
+  workflow_dispatch:
+    inputs:
+      task:
+        description: Rake task to run
+        required: true
+        default: db:safe_reset
+      env:
+        description: App to run the task on
+        required: true
+        default: ecf-dev
+
+jobs:
+  run-task:
+    runs-on: ubuntu-20.04
+    steps:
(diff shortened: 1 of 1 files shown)


kicferk1 (approved):
Hmm, this might be a little late, given that we dispatch plenty of workflows already, but who can dispatch our workflows exactly? Repository admins? People with write access? I sure hope not everyone with read access :P

cfq:
> Hmm, this might be a little late

You're right Michal, we might actually rethink our workflows after April. I need a quick way to reset a dev database right now.

kicferk1:
> You're right Michal, we might actually rethink our workflows after April. I need a quick way to reset a dev database right now.

Hey, judging by teaching-vacancies repo, where I can't dispatch any workflows, it is pretty restricted. Soo, I think this is good to go. It's just that I haven't done much github actions before, and it's something that would be good to know :)

ethanmills:
@kicferk1 Looks like everyone with write access https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event:~:text=GitHub%20Apps%20must%20have%20the%20actions%3Awrite%20permission%20to%20use%20this%20endpoint,-.%20For

PR #240, opened 2021-04-12
Apparently you can't start your command with cf

ethanmills:
DevOps never works on the first go...


kicferk1 (approved):

kicferk1:
I guess that kind of makes sense, the action has to log in first, so it helpfully gives you `cf` hand ...

Arghhhh!!!

amazimbe (approved):

PR #243, opened 2021-04-13
ECF-RP-492: admin manage schools

ethanmills:
### Context
This is the page an admin lands on when they first sign in. It is a list of all eligible schools. It is searchable by name and urn

### Testing
- Unit testing for school_policy
- Request specs for searching
- Cypress tests for a11y, searching

### Review Checks


ethanmills:
[DfE Admin: Manage Schools](https://trello.com/c/ax7bQFjn/492-dfe-admin-manage-schools)

ethanmills:
Thanks @callumacrae for sorting out my styling mess!

spec/cypress/integration/accessibility.js
```diff
@@ -49,7 +49,7 @@ describe("Accessibility", () => {
 
   it("Admin suppliers should be accessible", () => {
     cy.login("admin");
-    cy.get("h1").should("contain", "Suppliers");
+    cy.get("h1").should("contain", "Schools");
     cy.checkA11y();
   });
```
kicferk1, line 54: That test name is probably wrong.

spec/requests/users/sessions_spec.rb
```diff
@@ -125,7 +125,7 @@
 
       it "redirects to sign_in page" do
         post "/users/sign_in_with_token", params: { login_token: user.login_token }
-        expect(response).to redirect_to(admin_suppliers_path)
+        expect(response).to redirect_to(admin_schools_path)
```
kicferk1, line 128: I know it was here, but test name is wrong I think

app/controllers/admin/schools_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Admin
+  class SchoolsController < Admin::BaseController
+    skip_after_action :verify_authorized, only: :index
```
kicferk1, line 5: Do we need to skip it? Does `policy_scope` not authorize?

app/controllers/admin/schools_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+
+    def index
+      @query = params[:query]
+      @schools = policy_scope(School).order(:name).page(params[:page]).per(20)
+      if @query.present?
+        @schools = @schools.search_by_name_or_urn(@query)
```
amazimbe, line 11: Can we include local_authority and induction_coordinators when fetching?

spec/cypress/app_commands/scenarios/admin/schools.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+school = FactoryBot.create(:school, name: "Include this school", urn: 1234)
```
amazimbe, line 3: Can we make that urn 6 digits just to stay consistent with the actual data.

amazimbe (approved):
Couple of minor comments otherwise looks good to me

spec/cypress/integration/accessibility.js
ethanmills: Probably

app/controllers/admin/schools_controller.rb
ethanmills: Nope, there are two, verify_authorized and verify_policy_scoped

spec/cypress/integration/accessibility.js
ethanmills: I'm going to remove this, since we're checking this anyway

spec/cypress/support/step_definitions/common-interaction.js
```diff
@@ -54,6 +56,10 @@ When("I type {string} into {string}", (value, element) => {
   get(element).type(value);
 });
 
+When("I type {string} into {string} and press enter", (value, element) => {
+  get(element).type(`${value}{enter}`);
+});
```
callumacrae, line 61: For clearing, we have it as a separate step and I feel like it should be here too - so it'll be:

```
When I type "include" into "search box"
And I press enter in "search box"
```

By combining them in the same statement we end up with issues when we add new ones - e.g. if clearing was in the same statement, you'd have had to add two new rules for pressing enter, not one - and when another one is added in the future it'd require four new step definitions

spec/cypress/support/step_definitions/common-interaction.js
```diff
@@ -128,6 +134,10 @@ Then("{string} label should be unchecked", (text) => {
     });
 });
 
+Then("the table should have {int} row(s)", (number) => {
+  cy.get("tbody").find("tr").should("have.length", number);
+});
```
callumacrae, line 139: Is there a way of making this more generic? E.g. `there should be {int} instances of {string}`. wording feels clunky though

app/controllers/admin/schools_controller.rb
ethanmills: We're still N+1 on local authority, because of the way we fetch it. Open to suggestions

kicferk1: Can you do something like
```
@schools = @schools.search_by_name_or_urn(@query).joins(:local_authorities).includes(:local_authorities)
```
Is Rails clever enough to work the `through` relation into its query?

spec/cypress/support/step_definitions/common-interaction.js
ethanmills: I'd like to leave this for now and commonise if and when we find a need. I think if I try to do it now, it won't be quite right when we come to use it

ethanmills: Also, here we specifically need tbody > tr, since there is a tr in the header

ethanmills: Although I can do that with css selectors...

callumacrae: My problem with that is that it assumes the next person who needs to do something similar is aware of this conversation!

I'm not massively fussed either way :)

app/controllers/admin/schools_controller.rb
ethanmills: Nope, since we call `local_authority` on School, which is 
```
def local_authority
    school_local_authorities.latest.first&.local_authority
  end
```

spec/cypress/support/step_definitions/common-interaction.js
ethanmills: I shall endeavour to make them aware through review 😛

app/controllers/admin/schools_controller.rb
kicferk1: ufff

Yeah, I don't think we will find a way to get rid of n+1s without storing some more things in the database :/ Like `current_local_authority` on a school model.

ethanmills: This is an admin thing, and there are only a few of them. We also paginate to 20. Are you happy to leave this as-is for now?

kicferk1: Yes, I think the level of annoyance to avoid is it high enough to keep it for now and pick it up again when we know for sure that it's a problem worth fixing.

kicferk1 (approved):

PR #245, opened 2021-04-14
Code I think we no longer need

amazimbe:
### Context
Code I think we no longer need either because the designs have changed or was speculative in the first place. As we are going live I'd prefer to take out what isn't necessary.


kicferk1:
In some cypress test you get
> 2021-04-14 07:43:17 +0000 Rack app ("GET /schools/cohorts/2021" - (127.0.0.1)): #<ActionView::Template::Error: undefined method `edit_schools_estimate_participant_path' for #<ActionView::Base:0x0000000001ab80>>

Are we still using some path helpers that we remove here? `edit_schools_estimate_participant_path` in particular?

ethanmills (approved):

kicferk1 (approved):

PR #247, opened 2021-04-15
Privacy policy on first sign in

BroiSatse:
### Context

Implements: https://trello.com/c/yZsndlKq/460-add-privacy-policy-interstitial-page-to-first-sign-in-journey-for-induction-tutor

This PR ensures user must read (or rather be presented an opportunity to read) privacy policy before accessing any of the page. This satisfies acceptance criteria of user being presented Privacy Policy after the first sign up, but also prevents user from skipping the acceptance and provides a hook for re-acceptance on policy change.

PP Acceptance data is stored on new `privacy_policy_acceptance` column. I've decided on json for now, as there is no point of creating two new columns, especially given the changes in the next PP task: https://trello.com/c/MwoYUPCi/507-store-privacypolicy-versions-in-the-database.

### Guidance to review

The check is implemented in global before_action, which caused a bit of friction with our passwordless authentication method, as `current_user` throwing exceptions (`SignInIncomplete`)in the `session_controller` which is why I've decided to skip that `before_action` in this controller. I've attempted into fixing `current_user` to return `nil` instead but this has somehow broken the warden strategy and decided not to follow the white rabbit on this occasion.

NOTE: I have confirmed that the copy on the card (attached png file) is not a copy designed for this system but example of similar page implemented on another service. The copy displayed should be identical to our current `/privacy_policy` page.

### Testing

Log in as any user and try accessing any path - you will be redirected to `/privacy_policy`. Upon clicking `continue` your user record will be updated with acceptance information and you will be allowed to see any other page you'd normally expect to have access to. The `/privacy_policy` should not display the `continue` button once the policy is approved.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,13 @@ class ApplicationController < ActionController::Base
 
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :previous_url_for_cookies_page
+  before_action :check_privacy_policy_accepted
+
+  def check
+    render json: { status: "OK", version: release_version, sha: ENV["SHA"], environment: Rails.env }, status: :ok
+  end
+
+private
 
   def previous_url_for_cookies_page
     if request.get? && controller_name == "cookies"
@@ -12,10 +19,6 @@ class ApplicationController < ActionController::Base
     elsif request.get?
       session[:return_to] = request.original_url
     end
-  end
-
-  def check
-    render json: { status: "OK", version: release_version, sha: ENV["SHA"], environment: Rails.env }, status: :ok
   end
 
   def after_sign_in_path_for(user)
@@ -34,5 +37,12 @@ protected
 
   def configure_permitted_parameters
     devise_parameter_sanitizer.permit(:sign_up, keys: %i[email full_name])
+  end
+
+  def check_privacy_policy_accepted
+    return if !current_user || current_user.admin? || current_user.privacy_policy_acceptance.present?
+
+    session[:original_path] = request.fullpath
+    redirect_to privacy_policy_path
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,7 +40,9 @@ protected
   end
 
   def check_privacy_policy_accepted
-    return if !current_user || current_user.admin? || current_user.privacy_policy_acceptance.present?
+    return if !current_user || current_user.admin?
+    return unless current_user.induction_coordinator_profile
+    return if current_user.privacy_policy_acceptance.present?
 
     session[:original_path] = request.fullpath
     redirect_to privacy_policy_path
--- a/app/views/privacy_policies/show.html.erb
+++ b/app/views/privacy_policies/show.html.erb
@@ -1,4 +1,6 @@
-<% content_for :before_content, govuk_back_link( text: "Back", href: :back) %>
+<% if !current_user || current_user.privacy_policy_acceptance %>
+  <% content_for :before_content, govuk_back_link( text: "Back", href: :back) %>
+<% end %>
 <h1 class="govuk-heading-l">How we look after your personal data</h1>
   <h2 class="govuk-heading-m">Department for Education (DfE)</h2>
     <p class="govuk-body">Mentions of "us" and "we" mean DfE and "you" means anyone using this service.</p>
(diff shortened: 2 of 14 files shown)


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,9 @@ class ApplicationController < ActionController::Base
 
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :previous_url_for_cookies_page
+  before_action :check_privacy_policy_accepted
+
+  private
```
ethanmills, line 10: This cannot go here. `check` needs to be public

ethanmills: This is why your deploy is failing

app/controllers/application_controller.rb
```diff
@@ -35,4 +38,11 @@ def release_version
   def configure_permitted_parameters
     devise_parameter_sanitizer.permit(:sign_up, keys: %i[email full_name])
   end
+
+  def check_privacy_policy_accepted
+    return if !current_user || current_user.admin? || current_user.privacy_policy_acceptance.present?
```
ethanmills, line 43: Should be be showing this to lead provider users?

spec/tasks/manage_admins_spec.rb
```diff
@@ -10,6 +10,6 @@
     url = "http://www.example.com/users/sign_in"
     expect(AdminProfile).to receive(:create_admin).with(name, email, url)
 
-    task.execute(to_task_arguments(name, email))
+    capture_output { task.execute(to_task_arguments(name, email)) }
```
ethanmills, line 13: Nice 😄

ethanmills:
Looks like you have lint errors. Do you have the precommit hooks set up?

app/controllers/application_controller.rb
BroiSatse: That is an excellent question for Tom. I think we might have two privacy policies here.

app/controllers/application_controller.rb
BroiSatse: Doh! Fixed!

ethanmills (commented):
All looks good, last thing is to change so we only show the policy to Induction Coordinators

kicferk1 (commented):
I dislike using `json` in a db field, especially if all it is saving us is a single field.

Can we have two fields, `privacy_policy_accepted_at` and `privacy_policy_accepted_version`?

Which brings us neatly to the second point. The check for showing privacy policy is currently working, but as soon as we add a 'v2' it will break - cause it doesn't actually check the version that user accepted. I guess we could just 'remember to add a better check then', but if that's the case, do we need to add the version now?

config/routes.rb
```diff
@@ -17,7 +17,7 @@
   get "/check-account", to: "check_account#show"
 
   resource :cookies, only: %i[show update]
-  resource :privacy_policy, only: %i[show]
+  resource :privacy_policy, only: %i[show update]
```
kicferk1, line 20: I know it was here before, but can we use kebab-case for it?

app/views/privacy_policies/show.html.erb
```diff
@@ -97,4 +97,12 @@
     <p class="govuk-body">For further information or to raise a concern, visit the <%= govuk_link_to "Information Commissioner's Office.", "https://ico.org.uk/" %></p>
 
   <h2 class="govuk-heading-m">Keeping our privacy policy up to date</h2>
-  <p class="govuk-body">We reserve the right to update this privacy notice at any time, and we will provide you with a new privacy notice when we make any substantial updates. We will also notify you in other ways from time to time about the processing of your personal information.</p>
```
ethanmills, line 100: No back link at the top of this page please

kicferk1:
> Can we have two fields, `privacy_policy_accepted_at` and `privacy_policy_accepted_version`?
> 
> Which brings us neatly to the second point. The check for showing privacy policy is currently working, but as soon as we add a 'v2' it will break - cause it doesn't actually check the version that user accepted. I guess we could just 'remember to add a better check then', but if that's the case, do we need to add the version now?

Ah, I haven't read up on the next ticket, I guess the better check can come in that PR.

I still think json in a db is unnecessary - one more field on user is tiny, and it specifies the data integrity vastly better than a json. And it lets us query privacy policy info efficiently.

ethanmills:
The page in the footer should not be changed here, we don't want the form added there. How about a shared partial?

app/views/privacy_policies/show.html.erb
ethanmills: I've just noticed you're reusing the footer page

BroiSatse:
@ethanmills - I though about a partial and two separate views, but the only difference between those two pages is the single-button form at the bottom of the policy. I've made this form only to display when current user exists and has not yet accepted PP, so any normal visit via the link in the footer will have no visible changes to its previous content.

kicferk1 (approved):

ethanmills:
> @ethanmills - I though about a partial and two separate views, but the only difference between those two pages is the single-button form at the bottom of the policy. I've made this form only to display when current user exists and has not yet accepted PP, so any normal visit via the link in the footer will have no visible changes to its previous content.

What about when you're not logged in?

BroiSatse:
@ethanmills Ah, this is the bug! Nice catch! :)

ethanmills (approved):

PR #248, opened 2021-04-15
ECF-RP-501: Service to send partnership notification emails to schools

ethanmills:
### Context
We need to email schools when they are added to a partnership. This is the service that does that work. We need a unique token in these emails. We should also record the status of these emails. We need to send different emails to induction coordinator than to the school primary contact email

- Add mailer methods to send the emails
- Add a model to record the emails
- Add a service to create the record and call the mailer
- Generalise the notify callback endpoint to accept these emails

### Guidance to review
Probably commitwise

### Testing
- Manually tested
- Unit tests

### Review Checks
- ~[ ] All pages have automated accessibility checks via cypress~
- ~[ ] All `School` queries are correctly scoped - `eligible` when they need to be~


ethanmills:
[School receives email of new partnership](https://trello.com/c/VbbJB2Jt/501-school-receives-email-of-new-partnership)

app/controllers/api/notify_callbacks_controller.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
```
ethanmills, line 1: This was moved from /nominations, and modified a bit

app/controllers/api/notify_callbacks_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+class Api::NotifyCallbacksController < Api::ApiController
+  def create
+    return head :no_content unless params[:reference] || params[:id]
+
+    if params[:reference]
+      email = NominationEmail.find_by(token: params[:reference])
+      email.update!(notify_status: params[:status]) if email
+    else
+      email = PartnershipNotificationEmail.find_by(notify_id: params[:id])
+      email.update!(notify_status: params[:status], delivered_at: params[:sent_at]&.to_datetime) if email
```
kicferk1, line 12: This logic looks like a mess. I think it would be nice to have a reference that's decoupled from the things in the email. Kind of like the partnership emails do it.

Can we do it for Nomination emails too? As in, have a notify id?

Can we have delivered_at on both? 

In fact, do we want to have some inheritance between them?

app/mailers/school_mailer.rb
```diff
@@ -30,4 +32,58 @@ def nomination_confirmation_email(user:, school:, start_url:)
...
+      personalisation: {
+        provider_name: provider_name,
+        cohort: cohort,
+        start_url: start_url,
+        challenge_url: challenge_url,
+      },
+    )
+  end
+
+  def self.send_school_partnership_notification_email(recipient:, provider_name:, cohort:, nominate_url:, challenge_url:)
```
amazimbe, line 66: I'd put these 2 methods where you are calling them from to keep the mailers as conventional as possible.

app/controllers/api/notify_callbacks_controller.rb
amazimbe: I agree best to make the two emails work in a similar way and using the notify_id as a reference is definitely better. I'd be happy to have that done in another pr though.

app/mailers/school_mailer.rb
ethanmills: Will move, just to explain my reasoning: I put them here so as not to expose the internals of the mailer. I wanted to encapsulate all the notify weirdness, so there's only one place to change if we change the notify gem

app/models/nomination_email.rb
```diff
@@ -2,14 +2,41 @@
 
 class NominationEmail < ApplicationRecord
   belongs_to :school
+  belongs_to :partnership_notification_email, optional: true
```
amazimbe, line 5: Can we add the association boilerplate spec for this as well, please.

app/models/partnership.rb
```diff
@@ -5,6 +5,7 @@ class Partnership < ApplicationRecord
   belongs_to :lead_provider
   belongs_to :cohort
   belongs_to :delivery_partner
+  has_many :partnership_notification_emails
```
amazimbe, line 8: ditto

db/schema.rb
```diff
@@ -175,10 +175,26 @@
...
+    t.index ["partnership_notification_email_id"], name: "index_nomination_emails_on_partnership_notification_email_id"
     t.index ["school_id"], name: "index_nomination_emails_on_school_id"
     t.index ["token"], name: "index_nomination_emails_on_token", unique: true
   end
 
+  create_table "partnership_notification_emails", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "token", null: false
+    t.string "sent_to", null: false
+    t.string "email_type", null: false
+    t.string "notify_id"
```
amazimbe, line 188: I think you need to index this column as the callback from notify will lookup using the notify id

spec/services/partnership_notification_service_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      lead_provider: @lead_provider,
+      delivery_partner: @delivery_partner,
+      cohort: @cohort,
+      school: school,
+    )
+  end
+  let(:partnership_notification_email) { partnership.partnership_notification_emails.last }
+  let(:notify_id) { Faker::Alphanumeric.alphanumeric(number: 16) }
+
+  describe "#run" do
```
amazimbe, line 25: You called the method #notify and not #run

app/models/nomination_email.rb
amazimbe: And i haven't quite understood why we need this association.

ethanmills: @amazimbe I added this association for analytics purposes. We want to be able to link a clicked nominate link back to the partnership email, and a partnership email notify status through to a nomination email record

ethanmills: Since some of the partnership notification emails generate a nomination email record without actually sending a separate nomination email (We need the token for a nomination)

app/controllers/api/notify_callbacks_controller.rb
ethanmills: Will pick up as part of this ticket, but in a new PR

ethanmills: see https://github.com/DFE-Digital/early-careers-framework/pull/259

kicferk1 (approved):

amazimbe (approved):

PR #253, opened 2021-04-16
Adding instructions for reviewers on PR template

cfq:
We need to make it easy for non-developer reviewers (for example POs or URs) to
review the changes in PRs.

This last section can include a link to the exact place where the change can be
viewed and also instructions on how to get to that page with the right test
email or any other change/selection you have to do to view it.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,3 +9,5 @@
 ### Review Checks
 - [ ] All pages have automated accessibility checks via cypress
 - [ ] All `School` queries are correctly scoped - `eligible` when they need to be
+
+### How can I view this in a review app?


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -9,3 +9,5 @@
 ### Review Checks
 - [ ] All pages have automated accessibility checks via cypress
 - [ ] All `School` queries are correctly scoped - `eligible` when they need to be
+
+### How can I view this in a review app?
```
amazimbe, line 13: Do we need to add a checklist, like: - [ ] Have you seeded the db with the right data etc

cfq: It's a good question. I think it could be included in the review checks. I presume the developer writing out the instructions will have tried it in the preview app to see if they have the right data in place though.

ethanmills (approved):

amazimbe (approved):

PR #259, opened 2021-04-19
ECF-RP-501: Use notify id for callbacks on nomination emails

ethanmills:
### Context
Per this comment: https://github.com/DFE-Digital/early-careers-framework/pull/248#discussion_r615936569

- Change nomination emails callback to use notify id
- Record delivered_at for nomination emails

### Testing
Unit tests

### Review Checks

N/A

### DIFF
--- a/app/controllers/api/notify_callbacks_controller.rb
+++ b/app/controllers/api/notify_callbacks_controller.rb
@@ -2,17 +2,12 @@
 
 class Api::NotifyCallbacksController < Api::ApiController
   def create
-    return head :no_content unless params[:reference] || params[:id]
+    return head :no_content unless params[:id]
 
-    if params[:reference]
-      email = NominationEmail.find_by(token: params[:reference])
-      email.update!(notify_status: params[:status]) if email
-    else
-      email = PartnershipNotificationEmail.find_by(notify_id: params[:id])
-      email.update!(notify_status: params[:status], delivered_at: params[:sent_at]&.to_datetime) if email
-    end
+    log_email if failed_email?
 
-    log_email if failed_email?(email)
+    email = PartnershipNotificationEmail.find_by(notify_id: params[:id]) || NominationEmail.find_by(notify_id: params[:id])
+    email.update!(notify_status: params[:status], delivered_at: params[:sent_at]&.to_datetime) if email
 
     head :no_content
   end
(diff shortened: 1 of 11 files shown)


app/controllers/api/notify_callbacks_controller.rb
```diff
@@ -2,30 +2,23 @@
...
-      email.update!(notify_status: params[:status]) if email
-    else
-      email = PartnershipNotificationEmail.find_by(notify_id: params[:id])
-      email.update!(notify_status: params[:status], delivered_at: params[:sent_at]&.to_datetime) if email
-    end
+    log_email if failed_email?
 
-    log_email if failed_email?(email)
+    email = PartnershipNotificationEmail.find_by(notify_id: params[:id]) || NominationEmail.find_by(notify_id: params[:id])
+    email.update!(notify_status: params[:status], delivered_at: params[:sent_at]&.to_datetime) if email
```
kicferk1, line 10: `email&.update!(...)`?

db/migrate/20210415163231_add_notify_fields_to_nomination_email.rb
```diff
@@ -0,0 +1,12 @@
...
+class AddNotifyFieldsToNominationEmail < ActiveRecord::Migration[6.1]
+  def change
+    change_table :nomination_emails, bulk: true do
+      add_column :nomination_emails, :notify_id, :string
+      add_column :nomination_emails, :delivered_at, :datetime
+    end
+
+    add_index :nomination_emails, :notify_id
+  end
+end
```
kicferk1, line 12: In a dream world we would do single table inheritance on them, buuut I think that's probably alright.

app/controllers/api/notify_callbacks_controller.rb
```diff
@@ -2,30 +2,23 @@
...
-      email = NominationEmail.find_by(token: params[:reference])
-      email.update!(notify_status: params[:status]) if email
-    else
-      email = PartnershipNotificationEmail.find_by(notify_id: params[:id])
-      email.update!(notify_status: params[:status], delivered_at: params[:sent_at]&.to_datetime) if email
-    end
+    log_email if failed_email?
 
-    log_email if failed_email?(email)
+    email = PartnershipNotificationEmail.find_by(notify_id: params[:id]) || NominationEmail.find_by(notify_id: params[:id])
```
kicferk1, line 9: Do we also need to update `notify_id` on NominationEmails somewhere that's sending them?

db/migrate/20210415163231_add_notify_fields_to_nomination_email.rb
amazimbe: Just what I was thinking too but we can do that in future when things have calmed down

app/controllers/api/notify_callbacks_controller.rb
amazimbe: I think we can remove the reference param in the mailer too

ethanmills: Good spot both

kicferk1 (approved):

amazimbe (approved):

PR #260, opened 2021-04-19
Ecf rp 507 database privacy policy

BroiSatse:
### Context

We need better tracking of PrivacyPolicy versions accepted by users

This PR moves PrivacyPolicy text into a database table and separates compliance and User model. Creating a new PrivacyPolicy major version will automatically trigger re-acceptance for all the users. 

### Exclusions

The policy can only be updated via the console. Given the size of the PrivacyPolicy, it might be a bit cumbersome and error-prone, probably best if we could come up with some form of admin policy update.

### Review Checks


Gemfile
```diff
@@ -108,6 +108,7 @@ group :development do
 end
 
 group :test do
+  gem "database_cleaner"
```
BroiSatse, line 111: Cypress tests can leave behind quite a few records in the database, ruining all further rspec tests.

BroiSatse:
Closing to reset the deployment image after migration change

PR #269, opened 2021-04-20
Add percy everywhere

callumacrae:
### Context

🕺

### Review Checks

### DIFF
--- a/spec/cypress/integration/admin/SchoolManagment.feature
+++ b/spec/cypress/integration/admin/SchoolManagment.feature
@@ -9,6 +9,7 @@ Feature: Admin user managing schools
   Scenario: Viewing a list of schools
     Then the table should have 20 rows
     And the page should be accessible
+    And percy should be sent snapshot
 
     When I type "include" into "search box"
     And I press enter in "search box"
--- a/spec/cypress/integration/schools/ViewPartnerships.feature
+++ b/spec/cypress/integration/schools/ViewPartnerships.feature
@@ -2,29 +2,23 @@ Feature: Induction tutors viewing partnerships
   Induction tutors should be able to view details of their chosen induction programme
 
   Background:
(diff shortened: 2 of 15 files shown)


spec/cypress/integration/schools/ViewPartnerships.feature
```diff
@@ -2,29 +2,23 @@ Feature: Induction tutors viewing partnerships
...
     And I click the submit button
     Then I should be on "schools" page
-    And the page should be accessible
 
     When I click on "link" containing "2021"
     Then I am on "2021 school cohorts" page
-    And "page body" should contain "Choose your induction programme"
-    And "page body" should contain "Confirm your training provider"
-    And "page body" should contain "Add early career teachers and mentors"
-    And the page should be accessible
```
callumacrae, line 25: This was all duplicate from the Choose Programme spec. Ideally it would all be done using factory_bot in the the background like how ChooseMaterials.feature does it, but that felt outside the scope of this PR.

ethanmills:
Such context... Also, could you update the PR template to add a percy box next to the A11y one please?

spec/cypress/integration/admin/SchoolManagment.feature
```diff
@@ -9,6 +9,7 @@ Feature: Admin user managing schools
   Scenario: Viewing a list of schools
     Then the table should have 20 rows
     And the page should be accessible
+    And percy should be sent snapshot
```
ethanmills, line 12: What happens when you don't name them? What does the name do?

callumacrae: Defaults to the scenario+feature name. Basically we only need to specify the name if there's more than one snapshot per feature.

kicferk1 (approved):

PR #272, opened 2021-04-20
Ecf rp 507 database privacy policy

BroiSatse:
### Context

We need better tracking of PrivacyPolicy versions accepted by users

This PR moves PrivacyPolicy text into a database table and separates compliance and User model. Creating a new PrivacyPolicy major version will automatically trigger re-acceptance for all the users. 

### Exclusions

The policy can only be updated via the console. Given the size of the PrivacyPolicy, it might be a bit cumbersome and error-prone, probably best if we could come up with some form of admin policy update.

### Review Checks


app/models/privacy_policy.rb
```diff
@@ -0,0 +1,36 @@
...
+    belongs_to :privacy_policy
+  end
+
+  def self.current
+    order(:major_version, :minor_version).last
+  end
+
+  def acceptance_required?(user)
+    return false if !user || user.admin?
+    return false unless user.induction_coordinator_profile
```
ethanmills, line 17: Does this simplify to `return false unless user&.induction_coordinator?`?

spec/support/databse_cleaner.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+RSpec.configure do |rspec|
+  rspec.before(:suite) { DatabaseCleaner.clean_with :truncation }
```
ethanmills, line 4: Why do we need this?

ethanmills (approved):
Looks great!

spec/support/databse_cleaner.rb
BroiSatse: Cypress is cleaning up only before each test, meaning that rspec running after cypress starts with a few records in the database from the last cypress test. This was causing quite a few unexpected failures

app/models/privacy_policy.rb
BroiSatse: Theoretically, but induction coordinator can in theory be an admin (at least db structure does not prohibit that). Happy to update if we are certain this will never be the case

ethanmills:
[Store Privacy Policy versions in the database](https://trello.com/c/MwoYUPCi/507-store-privacy-policy-versions-in-the-database)

app/models/privacy_policy.rb
```diff
@@ -0,0 +1,36 @@
...
+      .where("privacy_policies.major_version >= ?", major_version)
+      .exists?
+  end
+
+  def version
+    [major_version, minor_version].join(".")
+  end
+
+  def accept!(user)
+    Acceptance.create(
```
amazimbe, line 31: I think we want Acceptance.create! here. I'm surprised rubocop didn't moan

app/models/privacy_policy.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def acceptance_required?(user)
+    return false if !user || user.admin?
+    return false unless user.induction_coordinator_profile
+
+    !Acceptance
+      .joins(:privacy_policy)
+      .where(user: user)
+      .where("privacy_policies.major_version >= ?", major_version)
+      .exists?
```
amazimbe, line 23: Instead of !exists why not just Acceptance.joins...... .none?

app/models/privacy_policy.rb
BroiSatse: Good catch! It is interesting. My **guess** is that that cop only runs in the controller?

app/models/privacy_policy.rb
BroiSatse: Awesome, it seems `none?` is also executing `EXIST` query (I was under impression it will load all the records and delegate `none?` to the @results array)

db/migrate/20210419142255_create_privacy_policy_acceptances.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+class CreatePrivacyPolicyAcceptances < ActiveRecord::Migration[6.1]
+  class PrivacyPolicy < ApplicationRecord
+  end
+
+  class User < ApplicationRecord
+  end
+
```
amazimbe, line 9: Why can't we just require these classes in here as we have already defined them in the models?

BroiSatse: @amazimbe Migrations should always be independent from the main application. Honestly, we should not even depend on the ApplicationRecord (but rubocop requires it...). The reason is that application code changes all the time and those changes should not affect migration. If at any point we rename User (I know, it is unlikely) this would stop new developers from running migrations.

db/seeds/privacy_policy_1.0.html
```diff
@@ -0,0 +1,126 @@
...
+<p class="govuk-body">If you’re an early career teacher using Manage Training For Early Career Teachers we may collect
+ the following information about you:</p>
+<ul class="govuk-list govuk-list--bullet">
+ <li>name</li>
+ <li>address</li>
+ <li>date of birth</li>
+ <li>phone number</li>
+ <li>email address</li>
+ <li>TRN number</li>
+ <li>any other information you choose to share during the applicaiton process - for example, you can choose to share
```
ethanmills, line 28:
```suggestion
 <li>any other information you choose to share during the application process - for example, you can choose to share
```

ethanmills (commented):
Spotted a typo 🙂

db/seeds/privacy_policy_1.0.html
```diff
@@ -0,0 +1,126 @@
...
+<ul class="govuk-list govuk-list--bullet">
+ <li>the Department for Education, and/or</li>
+ <li>the teacher training providers the candidate applies to</li>
+</ul>
+<p class="govuk-body">This means we’re joint data controllers, as well as data processors, for your information (defined
+ by data protection law).</p>
+<p class="govuk-body">Manage Training for Early Career Teachers is a new service. We’re confident that it has met our
+ security standards, but we’ll continue to run security tests as we improve the service.</p>
+
+<h2 class="govuk-heading-m">What personal data we collect</h2>
```
ethanmills, line 15: Legislation section missing, PM'd

ethanmills:
[Add cookies policy and privacy policy pages link to footer](https://trello.com/c/FzDFlecv/443-add-cookies-policy-and-privacy-policy-pages-link-to-footer)

PR #273, opened 2021-04-20
Remove all non-deterministic data from visual tests

callumacrae:
### Context

Will fix the percy errors in every PR 😬

### DIFF
--- a/spec/cypress/app_commands/scenarios/admin/schools.rb
+++ b/spec/cypress/app_commands/scenarios/admin/schools.rb
@@ -3,4 +3,7 @@
 school = FactoryBot.create(:school, name: "Include this school", urn: 123_456)
 FactoryBot.create(:user, :induction_coordinator, schools: [school])
 
+Faker::UniqueGenerator.clear
+Faker::Config.random = Random.new(42)
 FactoryBot.create_list(:school, 20)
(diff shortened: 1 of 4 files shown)


spec/cypress/app_commands/scenarios/admin/schools.rb
```diff
@@ -3,4 +3,7 @@
 school = FactoryBot.create(:school, name: "Include this school", urn: 123_456)
 FactoryBot.create(:user, :induction_coordinator, schools: [school])
 
+Faker::UniqueGenerator.clear
+Faker::Config.random = Random.new(42)
```
ethanmills, line 7: The most random of numbers

kicferk1:
Well, this is probably a good makeshift solution, but I really think in the long run we should re-consider having any randomness to begin with in cypress, so that we don't re-introduce this problem in the future :/

ethanmills (approved):

callumacrae:
> Well, this is probably a good makeshift solution, but I really think in the long run we should re-consider having any randomness to begin with in cypress, so that we don't re-introduce this problem in the future :/

Yep, it's definitely something to look out for in future PRs. Not sure of a way of automatically doing it short of removing faker from all the factories, though

PR #268, opened 2021-04-20
ECF-RP-523: Report schools

kicferk1:
### Context
We want the start page to the 'report schools' lead provider journey.

### Changes proposed in this pull request
Add a static page, controller, route, tests. Mostly as foundations for another multi-page flow. 

### Guidance to review
Lots of stuff is copy-pasted, so double check that naming makes sense. I'm also open to suggestions on how you would define the routes in this flow. 

### Testing
Added simple request specs and cucumber tests.

### Review Checks
- [x] All pages have automated accessibility checks via cypress
- [x] All `School` queries are correctly scoped - `eligible` when they need to be

### How can I view this in a review app?
Log in as `lead-provider@example.com`. Click 'Find and add schools'. You should see the start page. The button should do nothing.

PR #274, opened 2021-04-20
[ECF-1-497] Lead provider dashboard layout

amazimbe:
### Context
The dashboard layout for lead providers has changed from a two column / block layout to a simple heading/paragraph layout

Also, it's using the correct navigation component now (after checking with Alaric)—and the tabs component has been renamed to subnav to prevent this from happening again.

### Review Checks

Login as lead-provider@example.com and check the dashboard.

Screenshot 2021-04-20 at 18 29 03


amazimbe:
[Update Lead Provider Dashboard](https://trello.com/c/05POg6Ow/497-update-lead-provider-dashboard)

amazimbe:
[Update Lead Provider Dashboard](https://trello.com/c/05POg6Ow/497-update-lead-provider-dashboard)

spec/cypress/integration/Dashboard.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Dashboard page
+  Background:
+    Given cohort was created with start_year "2021"
+  Scenario: Visiting the dashboard
+    Given I am logged in as a "lead_provider"
+    Then I should be on "dashboard" page
+    And "page body" should contain "Confirm your schools"
+    And "page body" should contain "Check your schools"
+    And the page should be accessible
```
kicferk1, line 9: And a Percy shapshot?

spec/cypress/support/step_definitions/common-navigation.js
```diff
@@ -19,6 +19,7 @@ const pagePaths = {
   start: "/",
   privacy: "/privacy-policy",
   accessibility: "/accessibility-statement",
+  dashboard: "/dashboard",
```
kicferk1, line 22: Ech, I anticipate conflicts with https://github.com/DFE-Digital/early-careers-framework/pull/268 :/

spec/requests/dashboard_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Dashboard", type: :request do
+  let(:lead_provider) { create(:user, :lead_provider) }
+
+  before do
+    sign_in lead_provider
+    Cohort.create!(start_year: "2021")
```
BroiSatse, line 10: I think this is a bit dangerous here. 2021 seems to be hardcoded quite a lot in the test and we'll spend the first week(s) after Christmas fixing test failures. Could we have `create :cohort, :current` instead or use time travel?

kicferk1 (approved):

app/views/dashboard/_navigation.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render TabsComponent.new do |component| %>
+  <%= component.tab(path: dashboard_path) do %>
+    <strong>Overview</strong>
+  <% end %>
+  <%= component.tab(path: admin_supplier_users_path) do %>
```
kicferk1, line 5: Uhh, does that path look wrong to you too?

amazimbe: It does but then it doesn't really matter. I need to rebase once the report schools pr has been merged and that will be corrected then. Or perhaps I can copy your routes into mine and worry about the conflicts later

app/views/dashboard/show.html.erb
```diff
@@ -2,22 +2,17 @@
...
+        <%= render partial: "navigation" %>
+        <h1 class="govuk-heading-xl"><%= @lead_provider.name %></h1>
+        <h2 class="govuk-heading-m">
+          <!-- TODO Fix after Ethan code merge for lead providers-->
+          <%= govuk_link_to "Confirm your schools", nil %>
+        </h2>
+        <p>Tell the Department for Education (DfE) which schools you have confirmed an agreement with.</p>
+        <h2 class="govuk-heading-m">
+          <%= govuk_link_to "Check your schools", lead_providers_your_schools_path(selected_cohort_id: Cohort.current.id) %>
+        </h2>
```
ethanmills, line 14: I think the ticket says not to have this?

amazimbe: That's right

app/views/dashboard/_navigation.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render TabsComponent.new do |component| %>
+  <%= component.tab(path: dashboard_path) do %>
+    <strong>Overview</strong>
+  <% end %>
+  <%= component.tab(path: start_lead_providers_report_schools_path) do %>
+    <strong>Schools</strong>
+  <% end %>
+<% end %>
```
kicferk1, line 8: I might have added the same navigation inline in the report page. Could you reuse this version there?

kicferk1 (approved):

app/views/dashboard/_navigation.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render TabsComponent.new do |component| %>
+  <%= component.tab(path: dashboard_path) do %>
+    <strong>Overview</strong>
```
callumacrae, line 3: Why are these being styled bold here in this way? They should either be styled bold for all sub navigation across the site using CSS, or they shouldn't be bold.

callumacrae (changes requested):

app/views/dashboard/_navigation.html.erb
amazimbe: All the designs I have seen had them in bold. I agree that should be done in css though

callumacrae: Oh, right, I just checked the design and I understand now.

In that case, I think this is using the wrong component - the wireframes for the admin flow have the same thing and it looks like this on the real site:

image

amazimbe: @callumacrae from what Alaric said there's just one component that does this but it needs to support bold font if at the top and normal font if inside the page. The component this is based on is  https://hmcts-design-system.herokuapp.com/components/sub-navigation . What I have done is correct and if I can't see any benefit from styling this in css in this case.

callumacrae: image

For reference, this is what the page I just posted a screenshot of above looked like in the prototypes.

In our implementation currently, there's more than one component that does this! Per the HMCTS design system, that component is for _sub_ navigation, but we're dealing with primary navigation here (I don't know why it was renamed to Tabs when it was copied into this repo - it shouldn't have been).

If we're going to go with the design in this PR, we should still use the primary navigation component, but we should change it where it's already used, too. We 100% shouldn't have inconsistent navigation between pages!

app/webpacker/styles/subnav.scss
```diff
@@ -0,0 +1,118 @@
+/* ==========================================================================
```
callumacrae, line 1: This is just page-tabs.scss but actually in SCSS format

callumacrae (approved):

db/schema.rb
```diff
@@ -205,6 +205,10 @@
     t.uuid "school_id", null: false
     t.uuid "lead_provider_id", null: false
     t.uuid "cohort_id", null: false
+    t.string "status", default: "pending", null: false
+    t.string "reason_for_rejection"
+    t.datetime "accepted_at"
+    t.datetime "rejected_at"
```
kicferk1, line 211: That seems... surprising

callumacrae: dammit i'll revert this

kicferk1 (approved):

ethanmills (approved):
Super minor thing, but I think primary nav is supposed to have font weight 700 @callumacrae

app/views/admin/schools/_navigation.html.erb
```diff
@@ -1,11 +1,11 @@
-<%= render TabsComponent.new do |component| %>
-  <%= component.tab(path: admin_school_path(@school)) do %>
+<%= render SubnavComponent.new do |component| %>
+  <%= component.nav_item(path: admin_school_path(@school)) do %>
     <strong>Overview</strong>
```
ethanmills, line 3:
```suggestion
    Overview
```

app/views/admin/schools/_navigation.html.erb
```diff
@@ -1,11 +1,11 @@
...
   <% end %>
-  <%= component.tab(path: "#") do %>
+  <%= component.nav_item(path: "#") do %>
     <strong>Participants</strong>
```
ethanmills, line 6:
```suggestion
    Participants
```

app/views/admin/schools/_navigation.html.erb
```diff
@@ -1,11 +1,11 @@
...
   <% end %>
-  <%= component.tab(path: "#") do %>
+  <%= component.nav_item(path: "#") do %>
     <strong>Cohorts</strong>
```
ethanmills, line 9:
```suggestion
    Cohorts
```

ethanmills (commented):
Sorry... One last thing?

app/views/admin/schools/_navigation.html.erb
callumacrae: I thought I got rid of these!

callumacrae: oh wait this isn't part of this PR. i'll change anyway 😆

ethanmills: Thanks!

PR #275, opened 2021-04-21
Bump cloudfoundry to 0.14.1, bring back blue-green deploys

kicferk1:
### Context
We switched to standard deployment strategy a while back, because blue-green wasn't updating docker images. This PR is to test if that is no longer needed, since there is a new version with a fix.

When review app builds I will take note of docker image we have for it in paas. Then I will push a silly change. Then I will wait for it to re-deploy and see if the image is updated. And if it is, it's good to go.

Pin cloudfoundry to a new version. Switch to using blue-green deployment strategy.

You can't.


kicferk1:
Well, looks like we need to wait a bit longer before it's available on https://registry.terraform.io/providers/cloudfoundry-community/cloudfoundry/latest.

kicferk1:
Before dummy update image is 
```
dfedigital/early-careers-framework-dev:384708559acc5a94fe5369c2440c32e1e5b288a4
```

kicferk1:
After dummy update it is
```
dfedigital/early-careers-framework-dev:7825dd80b485e3ab708ff8571e8fe175e2358815
```
Which is different! I think we can probably merge it and see if it works on dev.

ethanmills (approved):
🎉

PR #277, opened 2021-04-21
Copy updates

ethanmills:
### Context
Another round of copy updates. Sat down with Shani, went through the changes, and made them live.

Lots of copy updates. Most significant changes are extra params to emails. This will need to be updated on notify after this has made it to staging

### Guidance to review
Do your best? Any obvious typos or horriblness, but Shani has confirmed the correctness of this copy

### Testing
Updated relevant tests

### Review Checks

Follow existing scenarios. These are wide ranging changes, touching most pages

### DIFF
--- a/app/forms/nominate_induction_tutor_form.rb
+++ b/app/forms/nominate_induction_tutor_form.rb
@@ -6,7 +6,7 @@ class NominateInductionTutorForm
   attr_accessor :full_name, :email, :token
 
   validates :full_name, presence: { message: "Enter a full name" }
-  validates :email, presence: { message: "Enter email" }
+  validates :email, presence: { message: "Enter email" }, format: { with: Devise.email_regexp, message: "Enter an email address in the correct format, like name@example.com" }
 
   def school
     NominationEmail.find_by(token: token).school
--- a/app/forms/supplier_user_form.rb
+++ b/app/forms/supplier_user_form.rb
@@ -7,7 +7,10 @@ class SupplierUserForm
   attr_accessor :full_name, :email, :supplier
   validates :supplier, presence: { message: "Select one" }, on: :supplier
   validates :full_name, presence: { message: "Enter a name" }, on: :details
-  validates :email, presence: { message: "Enter email" }, on: :details
+  validates :email,
+            presence: { message: "Enter email" },
+            format: { with: Devise.email_regexp, message: "Enter an email address in the correct format, like name@example.com" },
+            on: :details
   validate :email_not_taken, on: :details
 
   def attributes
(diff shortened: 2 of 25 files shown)


ethanmills:
[Additional copy updates](https://trello.com/c/PtmXEqsa/539-additional-copy-updates)

ethanmills:
[Bug #3: validation error messages and system error pages](https://trello.com/c/8oACxyyQ/511-bug-3-validation-error-messages-and-system-error-pages)

kicferk1 (approved):

app/forms/supplier_user_form.rb
```diff
@@ -7,7 +7,10 @@ class SupplierUserForm
   attr_accessor :full_name, :email, :supplier
   validates :supplier, presence: { message: "Select one" }, on: :supplier
   validates :full_name, presence: { message: "Enter a name" }, on: :details
-  validates :email, presence: { message: "Enter email" }, on: :details
+  validates :email,
+            presence: { message: "Enter email" },
+            format: { with: Devise.email_regexp, message: "Enter an email address in the correct format, like name@example.com" },
```
BroiSatse, line 12: I think it will display two error messages for blank email. We need `allow_blank: true` here

app/forms/nominate_induction_tutor_form.rb
```diff
@@ -6,7 +6,7 @@ class NominateInductionTutorForm
   attr_accessor :full_name, :email, :token
 
   validates :full_name, presence: { message: "Enter a full name" }
-  validates :email, presence: { message: "Enter email" }
+  validates :email, presence: { message: "Enter email" }, format: { with: Devise.email_regexp, message: "Enter an email address in the correct format, like name@example.com" }
```
BroiSatse, line 9: We need `allow_blank: true` for format validator.

app/mailers/user_mailer.rb
```diff
@@ -3,7 +3,7 @@
 class UserMailer < ApplicationMailer
   SIGN_IN_EMAIL_TEMPLATE = "7ab8db5b-9842-4bc3-8dbb-f590a3198d9e"
 
-  def sign_in_email(user, url)
+  def sign_in_email(user, url, token_expiry)
```
BroiSatse, line 6: Should we convert this into keywords? It is hard to remember the correct order :P

app/forms/supplier_user_form.rb
ethanmills: Doesn't seem to be the case 
image

BroiSatse: Interesting. So it seems controller is casting `""` into `nil` which is why fromat_validator ignores the check. It'd still happen in the console though:

image

Happy to leave it, it is really more of the nit-pick than an issue :P

ethanmills: Huh, ok. Just merged, will update in future if I remember

PR #280, opened 2021-04-21
ECF-RP-548 pentest fixes

kicferk1:
### Context
We want to fix some vulnerabilities. We won't get a full report until a while in the future, but these fixes are very likely to be needed.

1. Remove some `xxx_url` helpers - they create full url, as opposed to path helpers, which create relative path. We can't remove all of them, namely the ones that provide links in emails. 
2. Added rack-attack with some basic config
3. Set the session cookie to expire, be http only
4. Set all cookies on deployed envs to be secure, require https in them too
5. Sneaky cypress tests improvement - clean db after cypress run. I got some data from cypress tests interfering with rspecs :/ 

### Guidance to review
Feel free to suggest better / more sophisticated config. For some of the checks I will need the review app to be sure they work. 

### Testing
I am not sure how automatically testable those are.

### Review Checks

1. Uhh, not sure if you even want to test that.
2. Try logging in more than 5 times in 20 seconds (or whatever the config we set). You can use multiple tabs to make it easier.
3. Check the settings of your cookie, I hope the new settings should be there.
4. Try accessing review app through http, it should fail.
5. You can't.

### DIFF
--- a/config/environments/deployed_development.rb
+++ b/config/environments/deployed_development.rb
@@ -40,7 +40,8 @@ Rails.application.configure do
   config.active_storage.service = :local
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
+  config.session_store :cookie_store, key: "_early_career_framework_session", secure: true, expire_after: 2.weeks
 
   # Use a different cache store in production.
   # config.cache_store = :mem_cache_store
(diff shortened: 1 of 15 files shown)


config/environments/deployed_development.rb
```diff
@@ -40,14 +40,15 @@
   config.active_storage.service = :local
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
```
ethanmills, line 43: Clearly this _does_ work, but I'm surprised. The deployed app communicates with a CDN/router/load balancer/something over port 8080 using HTTP, not HTTPS. 🤷 so long as it works

ethanmills (approved):
LGTM!

amazimbe (approved):

PR #285, opened 2021-04-22
ECF-RP-562: Ensure emails are sent to schools without primary contact

ethanmills:
### Context
Currently trying to send an invite email to a school with only a secondary contact address fails. Empty emails are "", not nil in the db. This is truthy. 

- Change to a presence check
- Fix a flakey test

### Testing
Updated unit test

### Review Checks

I have updated "ZZ Test School 2" to only have a secondary email.
Go through the resend email journey starting at https://ecf-review-pr-285.london.cloudapps.digital/nominations/choose-location
You should receive the email


ethanmills:
[Bug: invite emails do not send to schools with only a secondary contact email](https://trello.com/c/LO4hfNgE/562-bug-invite-emails-do-not-send-to-schools-with-only-a-secondary-contact-email)

spec/requests/admin/schools/induction_coordinators_spec.rb
```diff
@@ -28,9 +28,11 @@
...
       }
       post admin_school_induction_coordinators_path(school.id), params: form_params
-      expect(User.last.full_name).to eq "jo"
-      expect(User.last.email).to eq "jo@example.com"
-      expect(User.last.induction_coordinator?).to be_truthy
+
+      created_user = User.order(:created_at).last
+      expect(created_user.full_name).to eq "jo"
+      expect(created_user.email).to eq "jo@example.com"
+      expect(created_user.induction_coordinator?).to be_truthy
```
ethanmills, line 35: FYI @goodviber

kicferk1 (approved):

amazimbe (approved):

app/services/invite_schools.rb
```diff
@@ -48,7 +48,7 @@ def send_nomination_email(recipient, token, school_name, email_expiry_date)
   end
 
   def recipient_email(school)
-    school.primary_contact_email || school.secondary_contact_email
+    school.primary_contact_email.presence || school.secondary_contact_email
```
BroiSatse, line 51: Should we prevent blank values from being written into the db?

kicferk1: I guess the only place we are currently doing it is school importer, and the empty strings come from a quirk of how the `csv` gem handles empty values in a row.

I am not sure we want to handle it in both places, because it could lead to inconsistencies in the usage of the values - some places checking for `presence`, others relying on the earlier check.

Personally I think making the school importer insert `nil`s instead of empty strings in the values it doesn't receive would be nice to have, precisely because it would prevent accidents like this in other fields.

ethanmills: Will add tech debt ticket

PR #286, opened 2021-04-22
Lead provider school confirmation

BroiSatse:
### Context

Once lead provider went through either CSV data upload or school search process for reporting recruited schools they enter 
 school confirmation. They are presented with the list of schools resulting from the previous steps, can remove the schools from that list and, when they're happy with the selection, confirm to proceed with the recruitment process.

This PR adds school confirmation process for lead providers. This does not implement any of the preceding journeys so there are a number of extra steps included here that should be removed once the setup journeys are completed.

### Testing

Sign in as a lead provider and navigate to /lead-providers/report-schools/confirm/start. This is a temporary route that will set up a random school list and redirect you to the actual production page at /lead-providers/report-schools/confirm. Refreshing the page should not affect the list of schools. You can remove the schools from the list as per requirements and you can reset the selection by clicking an extra `Restart` button at the bottom of the page.

Once Confirm button is clicked, you'll be presented with a success banner. Actual Partnership creation will be handled in another story.

As this story does not represent user journey, there are no cypress tests.

### Review Checks


app/controllers/lead_providers/confirm_schools_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  class ConfirmSchoolsController < ApplicationController
+    before_action :load_form
+
+    def show
+      @schools = School.where(id: @form.school_ids).order(:name)
+      @delivery_partner = DeliveryPartner.find(@form.delivery_partner_id)
+    end
+
+    def remove
+      school = School.find @form.school_ids.delete(params[:remove][:school_id])
```
kicferk1, line 13: I got lost somewhere between third and fourth thing that this line is doing :P 
Can we split it into a few? Maybe one to get the id, another one to find school, another one to remove the school id from the form?

app/controllers/lead_providers/confirm_schools_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+      set_success_message heading: "The list of schools has been confirmed", content: "You will be redirected to success page in another story"
+      redirect_to action: :show
+    end
+
+  private
+
+    def load_form
+      # TODO: remove when CVS/search journey is added
+      redirect_to(action: :start) unless session[:add_schools_form]
+      @form = SelectSchoolsForm.new(session[:add_schools_form])
```
kicferk1, line 39: Can we call the variable `@select_schools_form`?

app/forms/select_schools_form.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class SelectSchoolsForm
+  include ActiveModel::Model
+  include ActiveModel::Serialization
+
+  attr_accessor :school_ids
+  attr_accessor :delivery_partner_id
+  attr_accessor :source
```
kicferk1, line 9: We can define them all in one line, and we do it in other forms.

app/views/lead_providers/confirm_schools/show.html.erb
```diff
@@ -0,0 +1,45 @@
...
+      <th scope="col" class="govuk-table__header">Delivery partner</th>
+      <th scope="col"></th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+    <% @schools.each do |school| %>
+      <tr class="govuk-table__row">
+        <td class="govuk-table__cell"><strong><%= school.name %></strong></td>
+        <td class="govuk-table__cell"><%= school.urn %></td>
+        <td class="govuk-table__cell"><%= school.local_authority&.name %></td>
```
kicferk1, line 23: Uhh, N+1 queries here we come :( 
Not easily fixable either. 

@ethanmills do we have a tech debt ticket for looking into it? That's not the first place I see it in I think.

config/routes.rb
```diff
@@ -64,6 +64,11 @@
 
     resource :report_schools, path: "report-schools", only: [] do
       get "start", action: :start
+
+      resource :confirm_schools, only: %i[show update], path: "confirm" do
+        get :start
+        post :remove
+      end
```
kicferk1, line 71: I am not convinced we want a separate controller here, but I'm happy for that to get sorted when we have more parts of the flow.

spec/support/request.rb
```diff
@@ -0,0 +1,25 @@
...
+      routes.disable_clear_and_finalize = false
+    end
+
+    def set_session(key, value)
+      post "/__session", params: { key: key, value: value }
+    end
+
+    RSpec.configure do |rspec|
+      rspec.include self, type: :request
+    end
```
kicferk1, line 23: I think in some other request specs we have a different way of using session - namely by mocking it.
```
before do
      allow_any_instance_of(Nominations::RequestNominationInviteController)
        .to receive(:session)
              .and_return(session)
    end
```
I don't mind either solution, but I want our approach to be consistent across the tests.

app/controllers/lead_providers/confirm_schools_controller.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module LeadProviders
+  class ConfirmSchoolsController < ApplicationController
+    before_action :load_form
+
+    def show
+      @schools = School.where(id: @form.school_ids).order(:name)
+      @delivery_partner = DeliveryPartner.find(@form.delivery_partner_id)
```
kicferk1, line 9: Maybe we want the form to have `delivery_partner` and `schools` methods?

app/views/lead_providers/confirm_schools/show.html.erb
BroiSatse: It should be quite straight forward to fix though :)

kicferk1:
```
class School < ApplicationRecord
...
  def local_authority
    school_local_authorities.latest.first&.local_authority
  end
...
end
```
This doesn't look straightforward to preload to me :P But then again, I haven't spent too long in Rails query world - feel free to give it a go :)

app/controllers/lead_providers/confirm_schools_controller.rb
BroiSatse: I am not certain about this as it couples form to both models. I think i prefer the form to be just a simple data structure.

app/views/lead_providers/confirm_schools/show.html.erb
BroiSatse: @kicferk1 Ugh, I see! Yeah, that is not really straight forward! :P

spec/support/request.rb
BroiSatse: I might be biased but I'd vote for the above. :P Mocking the session has a major downside of using a hash instead of a session object - they might have compatible interface, but their behaviour is slightly different. It also reads better:
```
before do
  set_session :nomination_request_form, { school_id: school.id }
end
```

config/routes.rb
BroiSatse: Yeah, I like splitting everything into a smaller chunks whenever possible. Please challenge me on this a lot, otherwise we'll end up with 3-line classes everywhere :P

In this case I went for a separate controller as this confirmation process is common for two separate user journeys (both csv and school search path). Another option would be to place it on `report_schools_controller`, but that did not exist at the time of writing this.

app/forms/select_schools_form.rb
BroiSatse: Old habbit from reform/dry-validations :)

app/models/school.rb
```diff
@@ -53,6 +53,9 @@ class School < ApplicationRecord
     where.not(id: Partnership.joins(:cohort).where(cohorts: { start_year: year }).select(:school_id))
   }
 
+  has_one :latest_school_authority, -> { latest }, class_name: "SchoolLocalAuthority"
+  delegate :local_authority, to: :latest_school_authority, allow_nil: true
+
```
kicferk1, line 58: Nice!

I think we could use this in `app/views/admin/schools` - mind sneaking it in there? :)

kicferk1 (approved):

config/routes.rb
kicferk1: I think small controllers are great, so long as the same session is not shared between multiple controllers - I shudder to think about that :D 
But there are no controllers sharing that same session yet, so... temporarily there is no problem ;)

app/models/school.rb
BroiSatse: @kicferk1 Actually, we could even do it better. Code updated and admin/schools fixed as well.

ethanmills:
[Confirm schools (csv)](https://trello.com/c/4seSXswK/526-confirm-schools-csv)

app/controllers/application_controller.rb
```diff
@@ -4,7 +4,7 @@ class ApplicationController < ActionController::Base
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
   before_action :configure_permitted_parameters, if: :devise_controller?
-  before_action :previous_url_for_cookies_page
+  before_action :previous_url_for_cookies_page, except: :check
```
ethanmills, line 7: Good idea

app/controllers/lead_providers/confirm_schools_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    before_action :load_form
+
+    def show
+      @schools = School.where(id: @confirm_schools_form.school_ids).order(:name).includes(:local_authority)
+      @delivery_partner = DeliveryPartner.find(@confirm_schools_form.delivery_partner_id)
+    end
+
+    def remove
+      school_id = params[:remove][:school_id]
+      @confirm_schools_form.school_ids.delete(school_id)
```
ethanmills, line 14: Do we need to save this back to the session?

app/controllers/lead_providers/confirm_schools_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+      session[:confirm_schools_form] = {
+        source: :csv,
+        school_ids: School.order(Arel.sql("RANDOM()")).limit(10).pluck(:id),
+        delivery_partner_id: DeliveryPartner.order(Arel.sql("RANDOM()")).first.id,
+      }
+      redirect_to action: :show
+    end
+
+    def update
+      set_success_message heading: "The list of schools has been confirmed", content: "You will be redirected to success page in another story"
```
ethanmills, line 32: TODO here please?

app/controllers/lead_providers/confirm_schools_controller.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module LeadProviders
+  class ConfirmSchoolsController < ApplicationController
+    before_action :load_form
+
+    def show
+      @schools = School.where(id: @confirm_schools_form.school_ids).order(:name).includes(:local_authority)
+      @delivery_partner = DeliveryPartner.find(@confirm_schools_form.delivery_partner_id)
```
ethanmills, line 9: How do you feel about moving this logic into the form? Having a `schools` and `delivery_partner` method that returns these?

app/views/lead_providers/confirm_schools/show.html.erb
```diff
@@ -0,0 +1,45 @@
...
+</table>
+
+<p>Once you've confirmed, schools will be notified</p>
+
+<%= govuk_button_to "Confirm", { action: :update }, method: :put, class: "govuk-button govuk-!-margin-bottom-3" %>
+<% if @confirm_schools_form.source == "csv" %>
+  <%= govuk_link_to "Re upload CSV", dashboard_path, class: "govuk-button govuk-button--secondary" %>
+<% end %>
+
+<%# This is for presentation only %>
```
ethanmills, line 43: TODO?

app/controllers/lead_providers/confirm_schools_controller.rb
BroiSatse: @ethanmills Yes and no. :)

Yes, as it should be just a good practice to do this and would protect us from any changes.

No, as `@confirm_schools_form.school_ids` and `session[:confirm_schools_form][:school_ids]` both references the same object, so any modification to `@confirm_schools_form` modifies session as well. This is a bit too magical for my liking and would much rather prefer having explicit session update but I've let it slide for now. If we were to change this line to `@confirm_schools_form.school_ids -= [schoolId]` or sth similar, this would introduce behaviour change - it is covered with the test though.

To do it properly, I think we should stop using `Form.new(session[:key_to_session_hash]` or to modify our form objects to assign mutable objects with `dup`

app/controllers/lead_providers/confirm_schools_controller.rb
BroiSatse: So yes, it seems to be repeating now. The only thing I'd be very here though is additional coupling. As written, the form is completely oblivious to what uuids represent, and as such is pretty much just a silly data structure encapsulating absolutely no behaviour, leaving it to a controller to actually interpret that data. But, the benefit of this is pretty low, so I think I'll just make it consistent with other forms

app/controllers/lead_providers/confirm_schools_controller.rb
ethanmills: Oh weird.... Ok, I didn't know references to the session worked like that! I agree `dup` is probably the way forward. I much prefer things to be immutable

app/controllers/lead_providers/confirm_schools_controller.rb
ethanmills: Should make it testable too, right?

ethanmills: I personally prefer taking logic out of the controller where possible. How does that fit with your design patterns?

BroiSatse: Ideally I'd move pretty much all the logic out of controller and split it onto domain-specific dervice objects and web-specific details. I think we'd need to make a decision here though - form is supposed to be a simple data structure for validating input. Once we start adding more business logic and business transformations to it than they are no longer forms, but domain processes (which seems like you are shifting towards as well!)

kicferk1: Forms can be whatever we want them to be. 

In my mind, basic Rails thinks models should do a lot. They should validate data. They should be used to create forms in views. Among other things, like creating / updating entries in our database, or so that we can access some fields on the model in views from a single `@` variable, rather than many.

I think the reason we use forms is that we have use cases which go a bit outside what Rails wants - either validation that is not meant to be on the model, or manipulating data that doesn't belong on a single model. And so our forms are really a layer of data manipulation that's more flexible than models. That's why in a lot of places we have methods like `save` (with a transaction), or loading data in the form - because really, they are replicating what Rails models are capable of.

I would be weary of restricting our idea of the form to something more narrow, precisely because it will raise the need for something that will load the data, or save things to commonise what controllers are doing. Kind of like what's happening here. 

Also, in several places we still use Rails models instead of forms - whether we use a form is kind of dictated by whether our use case is possible to achieve using models or not. In that way, models and forms are kind of interchangeable when it comes to their place in their capabilities - and truth be told I don't really see the need to change it.

kicferk1: I guess the main thing is 'what do we call them, and is `form` a bad name?' :D 

Maybe? `form_with @confirm_schools_domain_object` doesn't roll off the tongue.

BroiSatse: @kicferk1 Absolutely, Rails way is wrong on so many leveles. Which is why I always suggest a trailblazer-style coding: https://github.com/trailblazer/trailblazer. There is a concept there called `Operation` which should contain all the business logic (so - empty controllers). `Forms` are only responsible for validations and they are used within the operations. I'd absolutely love to write few spikes using these concepts (however I would not use trailblazer implementations as they are also wrong in quite a feww places :P)

kicferk1 (approved):

PR #287, opened 2021-04-22
Fixed intermittently failing test

BroiSatse:
### Context

`User.last` orders records by id and in case of UUID gives a pretty random results. In this case, `User.last` has 50% chances of picking the admin user instead of newly created one.

It fixes it.

### Testing

yup!

You can't


BroiSatse:
Fix already introduced in https://github.com/DFE-Digital/early-careers-framework/pull/285#discussion_r618536247

PR #288, opened 2021-04-23
ECF-RP-480: challenge partnerships

ethanmills:
### Context
This allows schools to challenge partnerships from either an emailed link, or from a link on their dashboard. They have 14 days to do so.

The sending of the email is out of scope here  (happens when the partnership is created). There are separate tickets for the 7 day reminder email and the notification banner, as well as making changes if no issues are reported after 14 days.

 - Add screens to report issues with partnership
 - Add concept of challenge to partnerships
 - Update dashboard to reflect challenge status

### Testing
Cypress tests for the main cases
Unit tests for the edge cases around dashboard display

### Review Checks

Either:
- https://ecf-review-pr-288.london.cloudapps.digital/report-incorrect-partnership?token=abc123
Or:
- Sign in as cpd-test+tutor-2@digital.education.gov.uk
- Select 2021 cohort
- Select "Sign up with a training provider"
- Follow report link


spec/cypress/support/commands.js
```diff
@@ -22,6 +22,17 @@ Cypress.Commands.add("login", (...traits) => {
   cy.get('[action="/users/sign_in_with_token"] [name="commit"]').click();
 });
 
+Cypress.Commands.add("loginWithEmail", (email) => {
+  cy.visit("/users/sign_in");
+  cy.get("input[name*=email]").type(email);
+  cy.clickCommitButton();
+
+  cy.appEval(`User.find_by(email: "${email}")`).then((user) => {
```
callumacrae, line 30: This won't work on accounts that don't have login tokens (such as the one in the test case on my PR)

app/controllers/challenge_partnerships_controller.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+class ChallengePartnershipsController < ApplicationController
+  before_action :set_form, only: :create
+
+  def show
+    set_form
```
kicferk1, line 7: why not put this `set_form` in `before_action` too?

app/controllers/challenge_partnerships_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  rescue TokenExpiredError
+    redirect_to link_expired_challenge_partnership_path
+  rescue AlreadyChallengedError
+    redirect_to already_challenged_challenge_partnership_path(school_name: @school_name)
+  end
```
kicferk1, line 12: should we worry about these exceptions in `create` too?

app/controllers/challenge_partnerships_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def set_form
+    token = params[:token] || params.dig(:challenge_partnership_form, :token)
+    notification_email = PartnershipNotificationEmail.find_by(token: token)
+    raise ActionController::RoutingError, "Not Found" if notification_email.blank?
+    raise TokenExpiredError if notification_email.token_expired?
+
+    @partnership = notification_email.partnership
+    @school_name = @partnership.school.name
+    raise AlreadyChallengedError if @partnership.challenged?
```
kicferk1, line 39: double checking, do school name and partnership need to be `@` variables? We will have them on the form too.

spec/cypress/integration/ChallengePartnership.feature
```diff
@@ -0,0 +1,41 @@
...
+    Then I should be on "challenge partnership (any token)" page
+
+    When I click on "I do not recognise this training provider" label
+    And I click the submit button
+    Then I should be on "challenge partnership success" page
+
+    When I am on "2021 school partnerships" page
+    Then "page body" should contain "Have you signed up with a training provider?"
+
+
```
kicferk1, line 41: One of these empty lines can go ;)

spec/requests/challenge_partnership_spec.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Challenging a partnership", type: :request do
+  let(:partnership_notification_email) { create(:partnership_notification_email) }
+
+  describe "GET /report-incorrect-partnership?token=:token" do
+    it "renders the correct template" do
```
kicferk1, line 9: Let's specify which template is correct here and in other tests like it?

app/controllers/challenge_partnerships_controller.rb
ethanmills: So that we can rescue in `show`

app/controllers/challenge_partnerships_controller.rb
ethanmills: I don't think so. The odds of something changing between the two is so slim that I think a 500 is fine here

app/controllers/challenge_partnerships_controller.rb
ethanmills: Yup, we use them in some of the rescues, before we have the form

ethanmills:
[School induction coordinator can challenge a partnership and report reasoning why](https://trello.com/c/EHgR7mQA/480-school-induction-coordinator-can-challenge-a-partnership-and-report-reasoning-why)

spec/cypress/integration/ChallengePartnership.feature
```diff
@@ -0,0 +1,41 @@
...
+    Given I am logged in as user with email "test-subject@example.com"
+    And I am on "2021 school partnerships" page
+    When I click on "link" containing "report that your school has been confirmed incorrectly"
+    Then I should be on "challenge partnership (any token)" page
+
+    When I click on "I do not recognise this training provider" label
+    And I click the submit button
+    Then I should be on "challenge partnership success" page
+
+    When I am on "2021 school partnerships" page
```
callumacrae, line 38: This should be a `Given` or a `When I navigate to` - "I am on" isn't an action a user can take. I'm actually not sure why it's working!

app/controllers/challenge_partnerships_controller.rb
kicferk1: I guess if we had redirects in `set_form` method, we could maybe simplify this significantly - no need to raise our custom errors if we redirect straightaway. Question is, will it work as expected?

spec/cypress/integration/ChallengePartnership.feature
ethanmills: I'm using this, which I think is fine?
```
Given("I am on {string} page", (page) => {
  const path = pagePaths[page];
  cy.visit(path);
});
```

Are you thinking of 
```
Then("I should be on {string} page", (page) => {
  assertOnPage(page);
});
```

kicferk1 (approved):

app/controllers/challenge_partnerships_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+
+class ChallengePartnershipsController < ApplicationController
+  before_action :set_form, only: %i[show create]
+
+  def show; end
+
+  def create
+    render :show and return unless @challenge_partnership_form.valid?
+
+    @challenge_partnership_form.challenge!
```
BroiSatse, line 11: Ah, that is interesting! Basically you've converted the form object into a domain service object, which is responsible for both validation, data storage and business logic. I very much love this idea - this is my preferable way of structuring the project. :)

app/controllers/schools/partnerships_controller.rb
```diff
@@ -6,7 +6,15 @@ class Schools::PartnershipsController < Schools::BaseController
...
-    @partnership = @school.partnerships.find_by(cohort: cohort)
+    @partnership = @school.partnerships.unchallenged.find_by(cohort: cohort)
+
+    if @partnership
+      email = @partnership.partnership_notification_emails.order(:created_at).first
+      if email.present? && !email.token_expired?
+        @report_mistake_link = challenge_partnership_path(token: email.token)
+        @mistake_link_expiry = email.token_expiry.strftime("%d/%m/%Y")
+      end
+    end
```
BroiSatse, line 17: I'm probably too late here - I really don't think going back to the email record is a good option here. Email represents, well, an email that has been sent and should not take a responsibility of deciding the deadline for partnership challange - this is strictly partnership model responsibility. Token is only used as a mean of authorisation and should not be required for a signed-in user.

app/models/school_cohort.rb
```diff
@@ -13,7 +13,7 @@ class SchoolCohort < ApplicationRecord
   belongs_to :core_induction_programme, optional: true
 
   def training_provider_status
-    school.partnerships&.exists?(cohort: cohort) ? "Done" : "To do"
+    school.partnerships&.unchallenged&.exists?(cohort: cohort) ? "Done" : "To do"
```
BroiSatse, line 16: No need for `&` here - `partnerships` cannot be nil

spec/cypress/support/commands.js
```diff
@@ -74,6 +74,7 @@ Cypress.Commands.add("chooseNameAndEmailForUser", (name, email) => {
 Cypress.Commands.add("titleShouldEqual", (title) => {
   cy.title().should(
     "equal",
+
```
callumacrae, line 77: 🤔

app/controllers/challenge_partnerships_controller.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class ChallengePartnershipsController < ApplicationController
+  include Pundit
```
BroiSatse, line 4: Isn't pundit included by default?

app/controllers/challenge_partnerships_controller.rb
```diff
@@ -24,14 +25,23 @@ def success; end
...
+      notification_email = PartnershipNotificationEmail.find_by(token: token)
+      raise ActionController::RoutingError, "Not Found" if notification_email.blank?
+
+      @partnership = notification_email.partnership
+    elsif partnership_id.present?
+      @partnership = Partnership.find(partnership_id)
+      authorize(@partnership, :update?)
+    else
+      raise ActionController::RoutingError, "Not Found"
+    end
```
BroiSatse, line 39: This is getting pretty long now. Could we extract it into `@partnership = authorize_partnership` or even into a 
before_action? (then we don't need to use `and return` on redirections like on lines 42-43)

app/models/partnership.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Partnership < ApplicationRecord
+  CHALLENGE_WINDOW = 14.days
```
BroiSatse, line 4: Ultra minor - feels a bit odd not to freeze a constant, but I don't really know any way of modifying Duration object.

app/models/partnership.rb
```diff
@@ -28,4 +30,12 @@ def challenge!(reason)
 
     update!(challenge_reason: reason, challenged_at: Time.zone.now)
   end
+
+  def challenge_deadline
+    created_at + CHALLENGE_WINDOW
+  end
```
BroiSatse, line 36: I think we should store challenge_deadline in the separate column - there is a business question of "What do we want to happen to already existing Partnerships when the decision is made to shorten or extend the CHALLENGE_WINDOW".  In general I try avoiding any business logic depending on `created_at` column - it feels more like a database-level metadata than the domain data. Common issue would be importing pre-existing data - we would like to know both information - when the record was created (db-metadata) and when the partnership started.

app/policies/partnership_policy.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class PartnershipPolicy < ApplicationPolicy
+  def update?
+    user&.induction_coordinator? && user&.schools&.include?(record.school)
+  end
+
+  class Scope < Scope
+    def resolve
+      scope.all
```
BroiSatse, line 10: TODO?

app/policies/partnership_policy.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class PartnershipPolicy < ApplicationPolicy
+  def update?
+    user&.induction_coordinator? && user&.schools&.include?(record.school)
```
BroiSatse, line 5:
```
return false unless user&.induction_coordinator?
user.schools.include?(record.school)
```

What about admin users?

spec/models/partnership_spec.rb
```diff
@@ -63,4 +63,31 @@
       }.to raise_error ArgumentError
     end
   end
+
+  describe "#challenge_deadline" do
+    it "returns the correct value" do
+      travel_to Time.utc(2020, 1, 1, 13, 0, 0)
```
BroiSatse, line 69: Probably personal preference, but:
```
Time.parse("01/01/2020 13:00")
```
might be a little easier to read

app/controllers/challenge_partnerships_controller.rb
ethanmills: Not from the application controller, only the base controllers of each site section

app/policies/partnership_policy.rb
ethanmills: Until we have a way to do that, I think this is fine. There might some logic around what admins are able to do, and that would be part of implementing their functionality

ethanmills: Lead providers will also need to be able to create, but same applied

BroiSatse (approved):

PR #290, opened 2021-04-23
ECF-RP-512: Add programme choice confirmation and success pages

kicferk1:
### Context
https://trello.com/c/7ZwF2Isr/512-bug-12-additional-confirmation-and-success-page-when-choosing-a-programme
We want some confirmation and success pages for selecting the type of induction a school will run.

Add a confirmation page.
Add a new 'save programme' endpoint
Add a success page

### Guidance to review
Miro suggests having different sets of pages for FIP and CIP, but Shani thinks those pages are going to be very similar, and suggested commonising them - which I did and I like. It means we need to store the choice - which I do in session.

### Testing
I updated request specs and cypress tests.

### Review Checks

Log in as `school-leader@example.com`, follow your nose:
- continue
- pick a programme + continue
- should be on the confirm page (to test)
- confirm
- should be on the success page (to test)

If you want to re-test, I guess we would need to reset the db :/

### DIFF
--- a/app/views/schools/choose_programme/success.html.erb
+++ b/app/views/schools/choose_programme/success.html.erb
@@ -1,3 +1,4 @@
+<% content_for :title, "Programme confirmed" %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= govuk_panel(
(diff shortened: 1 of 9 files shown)


app/views/schools/choose_programme/success.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="govuk-grid-row">
```
ethanmills, line 1: page title

ethanmills (approved):
I think just a page title missing 🙂

kicferk1:
[Bug #12: additional confirmation and success page when choosing a programme](https://trello.com/c/7ZwF2Isr/512-bug-12-additional-confirmation-and-success-page-when-choosing-a-programme)

PR #291, opened 2021-04-23
Update canonical url host with domain name

ethanmills:
### Context
This was something we got from the boilerplate, and should have been updated when we got our domain name

Check the link tag in the head of any page

### DIFF
--- a/config/initializers/canonical_rails.rb
+++ b/config/initializers/canonical_rails.rb
@@ -9,8 +9,7 @@ CanonicalRails.setup do |config|
 
   # This is the main host, not just the TLD, omit slashes and protocol. If you have more than one, pick the one you want to rank in search results.
 
-  # TODO: update when we have domain
-  config.host = "www.education.gov.uk"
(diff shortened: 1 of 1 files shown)


config/initializers/canonical_rails.rb
```diff
@@ -9,8 +9,7 @@
 
   # This is the main host, not just the TLD, omit slashes and protocol. If you have more than one, pick the one you want to rank in search results.
 
-  # TODO: update when we have domain
-  config.host = "www.education.gov.uk"
+  config.host = "manage-training-for-early-career-teachers.education.gov.uk"
```
kicferk1, line 12: For my understanding, is this (or should it be) environment-dependant?

kicferk1 (approved):

config/initializers/canonical_rails.rb
ethanmills: I really don't know to be honest. I think having it wrong in prod is a _bad thing_, but having it wrong elsewhere isn't an issue. We don't want staging to appear in search results

amazimbe (approved):

PR #292, opened 2021-04-23
ecf-1-524 Lead providers can choose delivery partners for a school

amazimbe:
### Context
DfE need Lead Providers to specify which delivery partner is assigned to which school so that schools don't mistakenly challenge the partnership because they only know about the delivery partner.

A form with radio buttons that select the delivery partner for a school

### Review Checks

Login as lead-provider@example.com. From the dashboard click on find schools
image (2)

### DIFF
--- a/app/controllers/lead_providers/report_schools_controller.rb
+++ b/app/controllers/lead_providers/report_schools_controller.rb
@@ -2,7 +2,9 @@
 
 module LeadProviders
   class ReportSchoolsController < ::LeadProviders::BaseController
-    def start; end
+    def start
+      session.delete(:delivery_partner_id)
+    end
 
     def choose_delivery_partner
       @delivery_partner_form = LeadProviderDeliveryPartnerForm.new
--- a/app/controllers/lead_providers/report_schools_controller.rb
+++ b/app/controllers/lead_providers/report_schools_controller.rb
@@ -3,5 +3,24 @@
 module LeadProviders
   class ReportSchoolsController < ::LeadProviders::BaseController
     def start; end
+
+    def choose_delivery_partner
+      @delivery_partner_form = LeadProviderDeliveryPartnerForm.new
+      @delivery_partners = current_user.lead_provider.delivery_partners
+    end
+
+    def check_delivery_partner
+      @delivery_partner_form = LeadProviderDeliveryPartnerForm.new(
+        params.require(:lead_provider_delivery_partner_form).permit(:delivery_partner_id),
+      )
+
+      if @delivery_partner_form.valid?
+        session[:delivery_partner_id] = @delivery_partner_form.delivery_partner_id
+        redirect_to action: :choose_delivery_partner
+      else
+        @delivery_partners = current_user.lead_provider.delivery_partners
+        render :choose_delivery_partner
+      end
+    end
   end
 end
--- a/app/views/lead_providers/report_schools/choose_delivery_partner.html.erb
+++ b/app/views/lead_providers/report_schools/choose_delivery_partner.html.erb
@@ -1,0 +1,17 @@
+<% content_for :title, "Choose delivery partner" %>
+<% content_for :before_content, govuk_back_link( text: "Back", href: start_lead_providers_report_schools_path) %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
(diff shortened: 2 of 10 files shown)


app/controllers/lead_providers/report_schools_controller.rb
```diff
@@ -3,5 +3,24 @@
...
+      @delivery_partners = current_user.lead_provider.delivery_partners
+    end
+
+    def check_delivery_partner
+      @delivery_partner_form = LeadProviderDeliveryPartnerForm.new(
+        params.require(:lead_provider_delivery_partner_form).permit(:delivery_partner_id),
+      )
+
+      if @delivery_partner_form.valid?
+        session[:delivery_partner_id] = @delivery_partner_form.delivery_partner_id
```
kicferk1, line 18: I would be tempted to make the form serializable and do `@delivery_partner_form.serializable_hash` here.

kicferk1: Also, should we wipe the session in `start` action?

app/views/lead_providers/report_schools/choose_delivery_partner.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Choose the delivery partner</h1>
+    <p>Select the delivery partner that these schools have confirmed an agreement with.</p>
+    <%= form_with model: @delivery_partner_form, url: check_delivery_partner_lead_providers_report_schools_path, method: :post do |f| %>
+      <%= f.govuk_collection_radio_buttons(
+        :delivery_partner_id,
+        @delivery_partners,
+        :id,
+        :name,
+        legend: nil) %>
```
kicferk1, line 13: Hmm, I wonder if we should make the paragraph above into the `legend`.

spec/cypress/integration/lead_providers/ReportSchools.feature
```diff
@@ -10,3 +10,14 @@ Feature: Report Schools flow
     And "page body" should contain "2021"
     And the page should be accessible
     And percy should be sent snapshot called "Lead provider report schools start page"
+
+  Scenario: Selecting a delivery partner
+    Given scenario "lead_provider_with_delivery_partners" has been run
```
kicferk1, line 15: I think that's something I would put in `Background` part - hopefully all tests in this spec will need a setup like that.

kicferk1 (commented):
A few small comments

app/controllers/lead_providers/report_schools_controller.rb
amazimbe: Definitely agree with the comment about the session. I can't see any advantage to making the form serializable though. If there were many fields and / or I was using this to initialise an object then may be.

app/views/lead_providers/report_schools/choose_delivery_partner.html.erb
amazimbe: I could but these legends use a header tag by default and I find what we've been doung to use a p tag quite fiddly and difficult to understand when reading the code without offering any clear benefit.

app/controllers/lead_providers/report_schools_controller.rb
ethanmills: My expectation is that we will use a single form model for this step and the next step of selecting schools, either via CSV or UI. The changes that requires would be part of the later work though

spec/cypress/app_commands/scenarios/lead_provider_with_delivery_partners.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+profile = LeadProviderProfile.last
```
ethanmills, line 3: We need some ordering here

app/controllers/lead_providers/report_schools_controller.rb
amazimbe: @ethanmills I don't see how we can use a single form model for this part and the csv upload part. Those two things are just not related logically.

amazimbe:
[Select a delivery partner](https://trello.com/c/3fW2dDz7/524-select-a-delivery-partner)

kicferk1 (approved):

kicferk1:
Some conflicts with Ethan's error pages that just went in I think

PR #293, opened 2021-04-23
Update error copy, check a11y

ethanmills:
### Context
Update the copy on the 403 page to match 404. Get rid of the double `<main>` and do some a11y tests

### Review Checks

Go to `/403`, `/404`, `/500`

### DIFF
--- a/app/views/errors/forbidden.html.erb
+++ b/app/views/errors/forbidden.html.erb
@@ -1,1 +1,9 @@
-<h1>Forbidden</h1>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Page not found</h1>
+    <p class="govuk-body">If you typed the web address, check it is correct.</p>
+    <p class="govuk-body">If you pasted the web address, check you copied the entire address.</p>
+    <p class="govuk-body">If the web address is correct or you selected a link or button,
+      contact: <%= render MailToSupportComponent.new %></p>
+  </div>
+</div>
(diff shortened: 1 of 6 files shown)


app/views/errors/not_found.html.erb
```diff
@@ -7,4 +6,3 @@
       <p class="govuk-body">If the web address is correct or you selected a link or button, contact: <%= render MailToSupportComponent.new %></p>
     </div>
   </div>
-</main>
```
kicferk1, line 10: Do we want to un-indent the html inside?

app/views/errors/forbidden.html.erb
```diff
@@ -1 +1,9 @@
-<h1>Forbidden</h1>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Page not found</h1>
+    <p class="govuk-body">If you typed the web address, check it is correct.</p>
+    <p class="govuk-body">If you pasted the web address, check you copied the entire address.</p>
+    <p class="govuk-body">If the web address is correct or you selected a link or button,
+      contact: <%= render MailToSupportComponent.new %></p>
```
kicferk1, line 7: That's a funny message for a 403 page, but I guess we don't want do let people map our site by giving them information that a page is 'real but not allowed'.

kicferk1 (approved):

app/views/errors/forbidden.html.erb
ethanmills: Alaric said that's a common pattern for Gov, that to a user perspective there's no difference between "you can't see this" and "this doesn't exist"

kicferk1: Sure there is a difference, if it doesn't exist, you think that maybe you got a dead url. But if it exists and you can't see it, well, you might need to ask for some permissions. So for a determined user, different messages point them to different first course of action.

Whether having incorrect permissions is a common thing in government is a common problem or not, eh, who knows.

PR #294, opened 2021-04-23
Cucumber changes from other repo

callumacrae:
Adds:

- The ability to log in as an existing user (e.g., one created in a scenario)
- The ability to load seed data
- A bit of reorganisation

The seed data stuff isn't working at the moment - there are no schools loaded, but there appear to be schools created in the seed files.


kicferk1:
So, Rspec tests are unhappy due to an intermittent failure that was on develop when you merged off. Rebasing on the latest develop should fix that.

I have no idea why cypress tests are failing though :/

callumacrae:
> I have no idea why cypress tests are failing though :/

 Cypress was failing because this was WIP and I pushed it to get help :)

db/seeds.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-require_relative "seeds/initial_seed"
+Dir[Rails.root.join("db/seeds/initial_seed.rb")].each { |seed| load seed }
```
kicferk1, line 3: So we make a directory object containing a single file, and for each file in that directory we load it? Eh.

How about
```
load Rails.root.join("db/seeds/initial_seed.rb")
```
and similar below?

spec/cypress/support/step_definitions/factory-bot.js
```diff
@@ -51,12 +51,30 @@ const login = (traits, args) => {
...
+  const argsStrRails = Object.entries(args)
+    .map(([key, value]) => `${key}: "${value}"`)
+    .join(", ");
+
+  cy.appEval(
+    `User.find_by(${argsStrRails}).update(
+      login_token: "abcdefghij",
+      login_token_valid_until: 60.minutes.from_now)`
+  );
+  cy.visit(`/users/confirm_sign_in?login_token=abcdefghij`);
```
kicferk1, line 77: Doesn't `appEval` fire off a request (and so is async)? 

Do we need to do `then(...)` to be able to log in with that token?

Or is cypress nice enough to wait for it by itself?

db/seeds.rb
callumacrae: I honestly don't really understand what's really going on here - I copied it from the other repo. Every time I tried to make a change like what you're suggesting, it broke something 😂

spec/cypress/support/step_definitions/factory-bot.js
callumacrae: Everything cypress does is async! `.then()` is for when you need to refer to returned data, not for waiting, which is handled automatically.

db/seeds.rb
kicferk1: Good point, I guess I should fix other repo 😂

kicferk1: https://github.com/DFE-Digital/ecf-engage-and-learn/pull/78 is what I _hope_ will work - it worked when I tried seeding the db locally

spec/cypress/support/step_definitions/factory-bot.js
kicferk1: I guess since tests pass, cypress nicely waits for it :)

db/seeds.rb
kicferk1: Oh, well, it didn't. Uhh, I don't know why. Some side effect of enclosing a file in a dictionary I guess 👎 
Fine, this works, so it can probably stay.

kicferk1 (approved):

spec/cypress/support/step_definitions/factory-bot.js
```diff
@@ -51,12 +51,30 @@ const login = (traits, args) => {
...
+
+  const argsStrRails = Object.entries(args)
+    .map(([key, value]) => `${key}: "${value}"`)
+    .join(", ");
+
+  cy.appEval(
+    `User.find_by(${argsStrRails}).update(
+      login_token: "abcdefghij",
+      login_token_valid_until: 60.minutes.from_now)`
+  );
```
ethanmills, line 76: Ugh, I just wrote something like this 😢

ethanmills: I would say this probably shouldn't be in `factory-bot.js` though? Doesn't use factorybot at all does it?

callumacrae: It doesn't, but it's closely related to the stuff above it in the file so I think it's still a good place for it.

callumacrae: It doesn't, but it's closely related to the stuff above it in the file so I think it's still a good place for it.

callumacrae: It doesn't, but it's closely related to the stuff above it in the file so I think it's still a good place for it - it's databasey and none of the stuff in other files is.

PR #299, opened 2021-04-26
Lead provider successful school confirmation

BroiSatse:
### Context

Once lead provider confirms his selection of schools for a given delivery provider we want to create Partnership records and display success message.

In addition to required business logic, this PR also introduces changes to testing the delayed job enqueuing. ActiveJob test adapter does not respect database transactions and simply storing all the jobs in an array, severely lowering the confidence in those tests. Instead, we should be testing Delayed::Job directly. 

### Testing
Got to `/lead-provider/report-schools/confirm/start`. This will emulate CSV / schools search up to the confirmation stage.


BroiSatse:
This PR depends on https://github.com/DFE-Digital/early-careers-framework/pull/286, so it is best if we wait with the review until #286 is merged.

ethanmills:
The diff here doesn't seem right, so I'm not going to review until the other is merged/diff is sorted

app/controllers/lead_providers/confirm_schools_controller.rb
```diff
@@ -28,14 +28,30 @@ def start
       redirect_to action: :show
     end
 
+    # TODO: This is temporary behaviour and will be replaced with Partnership creation
     def update
```
ethanmills, line 32: Would this not be `create`?

app/controllers/lead_providers/confirm_schools_controller.rb
```diff
@@ -28,14 +28,30 @@ def start
...
+          partnership = Partnership.create!(
+            school_id: school_id,
+            cohort: cohort,
+            delivery_partner_id: @confirm_schools_form.delivery_partner_id,
+            lead_provider: current_user.lead_provider,
+          )
+
+          PartnershipNotificationService.new.delay.notify(partnership)
+        end
+      end
```
ethanmills, line 46: We've typically been doing this as  a `@confirm_schools_form.save!` or similar

spec/rails_helper.rb
```diff
@@ -90,10 +90,6 @@
   config.include ViewComponent::TestHelpers, type: :component
   config.include Rails.application.routes.url_helpers
 
-  config.before :each do
-    clear_enqueued_jobs
-  end
-
```
ethanmills, line 96: Why not?

spec/requests/lead_providers/report_schools_spec.rb
```diff
@@ -4,9 +4,9 @@
 
 RSpec.describe "Report schools spec", type: :request do
   let(:user) { create(:user, :lead_provider) }
+  let!(:cohort) { create(:cohort, :current) }
 
   before do
-    create(:cohort, start_year: 2021)
```
ethanmills, line 9: Is there not and advantage to doing this once, rather than before each test?

spec/rails_helper.rb
BroiSatse: Unnecessary - the way ActiveJob works in test is that it creates a brand new TestQueue adapter before each test with a blank `enqueued jobs` list. No point in clearing it.

spec/requests/lead_providers/report_schools_spec.rb
BroiSatse: No advantage really, those two are equivalent. I have a habit of getting all model creations out of `before` blocks, just feels cleaner but there is really no difference (expecte for the reference)

app/controllers/lead_providers/confirm_schools_controller.rb
BroiSatse: Good question! I'd say it is neither - it is `confirm`. CRUD/resourcefullness  is another part of Rails philosophy that makes very little sense to me:P

spec/factories/cohorts.rb
```diff
@@ -3,5 +3,9 @@
 FactoryBot.define do
   factory :cohort do
     start_year { Faker::Number.unique.between(from: 2021, to: 2100) }
+
+    trait :current do
+      start_year { 2021 }
+    end
```
BroiSatse, line 9: This is to match definition of `Cohort.current` which is also hard-coded to 2021 at the moment.

app/controllers/lead_providers/confirm_schools_controller.rb
ethanmills: Yes, this being a `PUT` feels wrong somehow

Procfile
```diff
@@ -1 +0,0 @@
-web: bundle exec puma -p $PORT
```
kicferk1, line 1: Did we inherit that from the boilerplate? :O

app/controllers/lead_providers/confirm_schools_controller.rb
```diff
@@ -24,20 +24,22 @@ def start
...
       redirect_to action: :show
     end
 
-    # TODO: This is temporary behaviour and will be replaced with Partnership creation
     def update
-      set_success_message heading: "The list of schools has been confirmed", content: "You will be redirected to success page in another story"
-      redirect_to action: :show
+      @confirm_schools_form.save!
+
+      redirect_to success_lead_providers_report_schools_path
```
kicferk1, line 36: What happens to the session after submission? Should it be deleted?

spec/requests/lead_providers/confirm_schools_spec.rb
```diff
@@ -46,4 +50,65 @@
...
+      end
+
+      it "is not creating any Partnerships" do
+        expect { confirm! }.not_to change(Partnership, :count)
+      end
+
+      it "schedules no partnership notifications" do
+        confirm!
+
+        expect(PartnershipNotificationService.new).not_to delay_execution_of(:notify)
```
kicferk1, line 110: Why do we need to be so specific? 

Can we do something like https://stackoverflow.com/a/27639157? Do we expect any other delayed jobs to be scheduled here?

spec/support/delayed_job.rb
```diff
@@ -15,3 +15,28 @@ class TestCronJob < CronJob
...
+
+      jobs.any? do |job|
+        @arguments.args_match?(*job.payload_object.args)
+      end
+    end
+
+    chain :with do |*args|
+      @arguments = RSpec::Mocks::ArgumentListMatcher.new(*args)
+    end
+  end
```
kicferk1, line 37: This is all pretty hard to understand to me - and I imagine to anyone who doesn't go into the details of classes we extend. Obvious questions being 'why do we ned the first three lines of handler in the query?', 'where do @arguments come from', 'what the hell is chain for', 'is handler the best place to check it', and I'm sure there are others.

Why do we need this level of power and precision in our tests? Are there any tests where we need to differentiate between the names of the jobs? Do we ever need to check the arguments of scheduled jobs? Is there no way to do this level of precision with specific stubs, rather than generic matchers?

spec/requests/lead_providers/confirm_schools_spec.rb
BroiSatse: @kicferk1 So the problem with `enqueued_jobs` as described on SO is an ActiveJob creation. Basicaly how it works is ActiveJob is creting a new queue adapter per each test, simply recording all the enqueued jobs in an array. This means two things: 
1. `object.delay.method()` will not be recorded - this is pure DelayedJob syntax, going around ActiveJob
2. Even if we wrap execution in ActiveJob object, ActiveJob::TestQueueAdapter will not be aware of database transactions (which is really one of few advantages of delayed job over sidekiq). This test would give false negative

spec/support/delayed_job.rb
BroiSatse: @kicferk1 Absolutely agree, this is not the easiest code. This is to deal with a limitation of ActiveJob abstraction. What we need to test is that given job/process has been enqueued, and there is no way of testing this using neither ActiveJob test helpers nor stubs as they are both unaware of DelayedJob transactional behaviour. 

On the upside - this is strictly technical domain. There is absolutely no business logic in this code and as such it is the best place for complexity to live in, while at the same time it makes the test code much simpler and easy to follow.

To answer other question: `chain` is part of RSpec::Matchers::DSL and is pretty common tool for defining new matchers. It defines a new method on the matcher which executes the code in the block and then returns itself. It is an equivalent of:

```
def with(*args)
  @arguments = ...
  self
end
```

and is used so the matcher can be used like this:

```
expect(object).to my_matcher_name.with(....)
```

`@arguments` are defined within optional `with` call.

BroiSatse: **Are there any tests where we need to differentiate between the names of the jobs?**
We need to be confident we are en-queuing the correct job. ActiveJob test helpers work in a very similar manner - `enqueued_jobs` is just an array of hashes containg the name ofthe job and its arguments.

spec/requests/lead_providers/confirm_schools_spec.rb
kicferk1: Can we do something like:
```
expect(Delayed::Job.count).to equal(1)
expect(Delayed::Job.first.handler.job_class).to equal(:notify)
```
As I understand it this gets rid of ActiveJob problems because it goes straight for the actual implementation.

spec/support/delayed_job.rb
kicferk1: Ech, this looks like classic `YAGNI` to me (You Aint Gonna Need It). Particular the part for defining `.with` method. Then again, the hard work for it has been done, so we might as well use it...

But if we do, I want it to be more readable.

I would put `chain` definition above `match` definition, to signify which one is happening first (cause I think `chain` precedes `match`, right?). 

Then inside `match`, let's make the line 
```
jobs = Delayed::Job.where("handler LIKE ?", handler.lines.first(3).join + "%")
```
a bit less magic - let's define the query more explicitly, and name the first three lines of our handler. 

I wonder if we need defining an in-memory handler too, maybe we could fetch the handlers from database using the method names we already know. But that should become more apparent when we have a more explicit query in place.

ethanmills:
[Success: schools have been reported](https://trello.com/c/saKFfYG8/529-success-schools-have-been-reported)

spec/support/delayed_job.rb
BroiSatse: I'd argue `YAGNI` does not apply here (and in does not apply in 95% of places when it is being used). There is an actual test which could not be written without this matcher - namely that no job is scheduled if a single partnership object fails to save. Also `with` is used to ensure that the partnership mailer has been enqueued with the correct arguments.

I am not sure how changing the order of match and chain would help, both those calls are DSL methods to define methods on a matcher and their execution does not matter. Morover, `with` call is optional. This is a standard matcher definition as per Rspec documentaiton: https://relishapp.com/rspec/rspec-expectations/v/3-10/docs/custom-matchers/define-matcher-with-fluent-interface.

As per the line suggested - it is not possible to write it with less magic. DelayedJob is serializing objects into yaml, of which the first 3 lines define the class of the object receiving the call. The goal of this query is to limit a number of jobs received to a reasonable amount - alternatively we could just loop over all the jobs, serialize all the payload_objects in search for an exact match. Another option is to configure DelayedJobs to use json columns instead and use json operators in the query.

kicferk1: > As per the line suggested - it is not possible to write it with less magic. DelayedJob is serializing objects into yaml, of which the first 3 lines define the class of the object receiving the call.

Fine, let's literally have three variables, one for each of those lines. So someone reading the code knows _why_ we care about first three lines, and what they represent. So they don't think 'hmm, I wonder if I can reduce it to first two lines, or just one line out of those three'.

BroiSatse: @kicferk1 There's really no way to name those variable which would make much sense. This is an example of a handler as DJ stores it in the database:

```
--- !ruby/object:Delayed::PerformableMethod
object: !ruby/object:SchoolDataImporter
  logger: !ruby/object:ActiveSupport::Logger
    level: 0
    progname: 
    default_formatter: !ruby/object:Logger::Formatter
      datetime_format: 
    formatter: !ruby/object:ActiveSupport::Logger::SimpleFormatter
      datetime_format: 
      thread_key: activesupport_tagged_logging_tags:17500
    logdev: !ruby/object:Logger::LogDevice
      shift_period_suffix: "%Y%m%d"
      shift_size: 1048576
      shift_age: 0
      filename: "/home/broisatse/work/dfe/early-careers-framework/log/development.log"
      dev: !ruby/object:File {}
      binmode: false
      mon_data: !ruby/object:Monitor {}
      mon_data_owner_object_id: 6640
  start_year: 2021
method_name: :run
args: []
```

PerformableObject is a object defined by DelayedJob, which represents `.delay` call. It encapsulates 3 piece of information, receiving object, method to delay and arguments. In ideal world we would grab yaml representation of the entire `object` attribute, however it might be time dependent which is why I have only grabbed the first 3 lines.

I really see no value in assigning it into 3 separate variables - the whole goal of this query is to get all the jobs that has been enqueued for objects of the same class as object given in `expect`. This code cannot be understood without knowing the internals of DelayedJob, even with additional variables.

I'll defenitely add some comments explaining how this behaves, however if anyone was to apporach this in the future, short `byebug` session would make everything clear.

kicferk1: Right, so the `3` is arbitrary. I got the impression that the number was specifically important.

I wonder if the real thing we are looking for is the line `object: !ruby/object:XXX`? Maybe we could call it something like `handler_object_class_line` and get objects based off that? Is that line 2?

app/controllers/lead_providers/confirm_schools_controller.rb
BroiSatse: @kicferk1 It is being deleted on the success page

app/views/lead_providers/confirm_schools/show.html.erb
```diff
@@ -35,13 +33,11 @@
   </tbody>
 </table>
 
-<p>Once you've confirmed, schools will be notified by email.</p>
-<p>Schools will have the option to notify the Department for Education and the training provider if they think this is incorrect.</p>
+<p>Once you've confirmed, schools will be notified</p>
 
 <%= govuk_button_to "Confirm", { action: :update }, method: :put, class: "govuk-button govuk-!-margin-bottom-3" %>
 <% if @confirm_schools_form.source == "csv" %>
-  <%# TODO: This should point to a csv upload once it is implemented %>
```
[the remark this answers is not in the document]
BroiSatse: Eh, silent merge conflicts!

app/controllers/lead_providers/report_schools_controller.rb
```diff
@@ -24,5 +24,9 @@ def check_delivery_partner
         render :choose_delivery_partner
       end
     end
+
+    def success
+      @confirm_schools_form = ConfirmSchoolsForm.new(session.delete(:confirm_schools_form))
+    end
```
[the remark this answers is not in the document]
BroiSatse: Sessions are not really controller-scoped, but absolutely agree it does not feel right here.

Basically the process we're modelling here is:

1. Start reporting schools
2. Create list of schools (either by search or csv upload)
3. Confirm / remove schools from the list
4. Create partnerships

Ideally, the entire process would use the exact same form (meaning that the form I've created is just a temporary  placeholder) which would be then shared by all the controllers used in the process. That somehow feels more correct and I'd tackle this in the finalisation of the joruney.

spec/requests/lead_providers/confirm_schools_spec.rb
```diff
@@ -46,4 +50,65 @@
       expect(flash[:success]).to be_present
     end
   end
+
+  describe "PUT /lead-providers/report-schools/confirm" do
+    let!(:cohort) { create :cohort, :current }
+
+    it "creates new partnerships" do
```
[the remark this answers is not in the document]
BroiSatse: Yup, the only action that spark joy here is abandoning CRUD

app/views/lead_providers/confirm_schools/show.html.erb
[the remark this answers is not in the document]
BroiSatse: Actually that view should not be modified almost at all! :D

kicferk1 (approved):

PR #301, opened 2021-04-27
ECF-RP-564: Worker app to run delayed jobs

ethanmills:
### Context
This is PR 1/3 for this ticket.
Inspired by TVS, we want to run background jobs in a different instance. We do this in staging and prod. Review and dev have the background jobs running in the web container already.

### Testing
Added the worker app to review in the first commit, checked it worked:
image

N/A

### DIFF
--- a/terraform/app/modules/paas/main.tf
+++ b/terraform/app/modules/paas/main.tf
@@ -11,6 +11,7 @@ resource cloudfoundry_app web_app {
   docker_image = var.app_docker_image
   health_check_type = "http"
   health_check_http_endpoint = "/check"
+  health_check_timeout = 60
   instances = var.web_app_instances
   memory = var.web_app_memory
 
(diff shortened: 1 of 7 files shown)


terraform/app/modules/paas/main.tf
```diff
@@ -11,6 +11,7 @@ resource cloudfoundry_app web_app {
   docker_image = var.app_docker_image
   health_check_type = "http"
   health_check_http_endpoint = "/check"
+  health_check_timeout = 60
```
ethanmills, line 14: This is from TVS, to stop PaaS from killing our app: https://github.com/DFE-Digital/teaching-vacancies/pull/2927

terraform/app/modules/paas/main.tf
```diff
@@ -38,20 +39,42 @@ resource cloudfoundry_app web_app {
...
 resource cloudfoundry_route web_app_route_gov_uk {
   for_each = toset(var.govuk_hostnames)
-  domain   = data.cloudfoundry_domain.education_gov_uk.id
-  space    = data.cloudfoundry_space.space.id
+  domain = data.cloudfoundry_domain.education_gov_uk.id
+  space = data.cloudfoundry_space.space.id
   hostname = each.value
 }
 
+resource "cloudfoundry_app" "worker_app" {
```
ethanmills, line 54: Most changes here are inspired by TVS

terraform/app/variables.tf
```diff
@@ -65,15 +65,19 @@ variable paas_web_app_start_command {
...
+  default = 0
 }
 
 variable paas_worker_app_memory {
-  default = 512
+  default = 2048
+}
+
+variable paas_worker_app_start_command {
+  default = ""
```
kicferk1, line 80: I wonder, do we need the default both here and in `terraform/app/modules/paas/variables.tf`?

terraform/app/modules/paas/main.tf
kicferk1: Is this 60s? What is the default? 1s? Do we need 60 times the boost?

ethanmills:
[7 day reminder emails for schools within 14 days of partnership being reported to DfE who have not reported an incorrect sign up](https://trello.com/c/sRQPCQD3/564-7-day-reminder-emails-for-schools-within-14-days-of-partnership-being-reported-to-dfe-who-have-not-reported-an-incorrect-sign-up)

terraform/app/modules/paas/main.tf
ethanmills: This is 60s, yes. There's a different between startup and running. Current running is 1s. Looking at the TVS PR, I believe this is set to 60s because it sets both startup and running, and we need more than e.g. 10s for startup

terraform/app/variables.tf
ethanmills: No, meant to remove this (removed the other duplicated and created new ones 🙄 )

kicferk1 (approved):

PR #305, opened 2021-04-30
add bucket for csv uploads

goodviber:
### Context
We need to add a store for partnership_csv_uploads
Add s3 bucket to terraform build
### Testing
aws-s3-bucket config should be available in cf $VCAP_SERVICES
### Review Checks

### DIFF
--- a/terraform/app/modules/paas/variables.tf
+++ b/terraform/app/modules/paas/variables.tf
@@ -61,6 +61,7 @@ locals {
   )
   app_cloudfoundry_service_instances = [
     cloudfoundry_service_instance.postgres_instance.id,
+    cloudfoundry_service_instance.csv_bucket.id,
     cloudfoundry_user_provided_service.logging.id,
   ]
   app_service_bindings = concat(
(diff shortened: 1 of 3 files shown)


ethanmills (approved):
Looks very familiar, probably best to wait for another review

kicferk1 (approved):

terraform/app/modules/paas/variables.tf
```diff
@@ -69,4 +70,5 @@ locals {
   logging_service_name     = "${var.service_name}-logit-${var.environment}"
   postgres_service_name    = "${var.service_name}-postgres-${var.environment}"
   web_app_name             = "${var.service_name}-${var.environment}"
+  csv_bucket_name          = "${var.service_name}-csv-${var.environment}"
```
kicferk1, line 73: Maybe include `bucket` in the name - all other services include the answer to 'what are you?' in their names.

kicferk1:
> Looks very familiar, probably best to wait for another review

So machiavellian, asks to wait and merges something that causes conflicts right after 😂

PR #306, opened 2021-04-30
Add swagger doc generation rake task

plisovin:
### Context

Rake task: **rswag:specs:swaggerize**

This was mainly borrowed from the [teacher-training-api](https://github.com/DFE-Digital/teacher-training-api) project

### Guidance to review

Place you spec request files into spec/docs directory
Run the task ```rswag:specs:swaggerize```
The open api spec will be written into the ```swagger``` directory

https://ecf-review-pr-306.london.cloudapps.digital/api-docs/index.html

Selection_20210430-01-10bd710bd710bd7

(operation definitions will come in other PRs)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -181,5 +181,8 @@ Rails.application.routes.draw do
   get "/422", to: "errors#unprocessable_entity", via: :all
   get "/500", to: "errors#internal_server_error", via: :all
 
+  mount OpenApi::Rswag::Ui::Engine => "/api-docs"
+  mount OpenApi::Rswag::Api::Engine => "/api-docs"
+
   resource :school_search, only: %i[show create], path: "school-search", controller: :school_search
 end
--- a/lib/tasks/swaggerize.rake
+++ b/lib/tasks/swaggerize.rake
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+if defined?(RSpec)
+  require "rspec/core/rake_task"
+
+  Rake::Task["rswag:specs:swaggerize"].clear
+
+  namespace :rswag do
+    namespace :specs do
+      desc "Generate Swagger JSON files from integration specs"
+      RSpec::Core::RakeTask.new("swaggerize") do |t|
+        t.pattern = "spec/docs/**/*_spec.rb"
+
+        t.rspec_opts = ["--format OpenApi::Rswag::Specs::SwaggerFormatter", "--order defined"]
+      end
+    end
+  end
+end
--- a/spec/swagger_helper.rb
+++ b/spec/swagger_helper.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.configure do |config|
+  config.swagger_root = Rails.root.join("swagger").to_s
+  config.swagger_docs = {
+    "v1/api_spec.json" => {
+      swagger: "2.0",
+      info: {
+        title: "API documentation",
+        version: "v1",
+        description: "Auto generated doc",
+        definitions: Dir[Rails.root.join("spec/docs/schemas/*.yaml")]
+                        .map { |f| { File.basename(f, ".yaml").titleize.gsub(/\W+/, "") => YAML.safe_load(File.read(f)) } }
+                        .reduce(&:merge),
+      },
+      basePath: "/api/v1",
+    },
+  }
+
+  config.swagger_format = :json
+end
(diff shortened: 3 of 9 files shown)


config/routes.rb
```diff
@@ -174,5 +174,8 @@
   get "/422", to: "errors#unprocessable_entity", via: :all
   get "/500", to: "errors#internal_server_error", via: :all
 
+  mount OpenApi::Rswag::Ui::Engine => "/api-docs"
```
plisovin, line 177: _We could always stick the route behind `unless Rails.env.production?`_ (quote by Michal from slack) i.e.:

```diff
+ unless Rails.env.production?
+  mount OpenApi::Rswag::Ui::Engine => "/api-docs"
+  mount OpenApi::Rswag::Api::Engine => "/api-docs"
+ end
```

spec/swagger_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+
+require "rails_helper"
+
+RSpec.configure do |config|
+  config.swagger_root = Rails.root.join("swagger").to_s
+  config.swagger_docs = {
+    "v1/api_spec.json" => {
+      swagger: "2.0",
+      info: {
+        title: "Provider Events API",
```
timabell, line 11:
```suggestion
        title: "API documentation",
```
Sounds like this might end up being more than just the events interface

config/routes.rb
timabell: Good point. We should probably think about whether we want to feature-flag this for production (and staging perhaps) or not while we're iterating on it. I completely can't make my mind up.

On the one hand we should hold it back until it's a bit more final.
On the other hand no-one will be looking at it in production and it won't be wrong just incomplete, so no consequence of shipping it.

timabell:
The commit message should link to the inspiration over in teacher-training-api https://github.com/DFE-Digital/teacher-training-api/tree/592c38f24a80566ba299d07d206d4193fce1af46

lib/tasks/swaggerize.rake
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+if defined?(RSpec)
```
blackrat, line 3: Do we just not want it to blow up if RSpec isn't installed, for example on a production only system?

blackrat (approved):

lib/tasks/swaggerize.rake
```diff
@@ -0,0 +1,18 @@
...
+  require "rspec/core/rake_task"
+
+  Rake::Task["rswag:specs:swaggerize"].clear
```
kicferk1, line 6: Can we add a note to `Readme` about generating and re-generating swagger docs?

config/routes.rb
mattonfoot: I would suggest doing that because the jQuery dependencies in that UI is massive behind the times and always get pulled out on a ITHC. Also even better to use the GOV UK Tech docs for consistency

lib/tasks/swaggerize.rake
kicferk1: Uhh, when does this file get loaded?

I don't know, but chances are the answer is "more often than when you try to run that task". And if that's the case, we could accidentally load it in production as part of legitimate step and... BOOM

I prefer not adding task silently over blowing up :)

config/routes.rb
kicferk1: I think that's a decision to be made when we add the first api. Trade-offs should become more apparent then.

kicferk1 (commented):
Looks great, I think a small section in readme about API docs will save us time in future :)

config/routes.rb
kicferk1: But Matt raises good points, we don't want to cause R&P more trouble in their technical architecture review :P

ethanmills: Where is the jQuery dependency sorry?

ethanmills: Where was the jQuery dependency?

lib/tasks/swaggerize.rake
plisovin: Added here 2fb193f

config/routes.rb
mattonfoot: looks like the UI is now React, so as long as the package pulls in the latest UI version we should be alright.

kicferk1 (approved):

lib/tasks/swaggerize.rake
blackrat: I was asking if that was the reason, rather than suggesting it. Thought that it might be better to define that it was only loaded during test (and possibly development) explicitly, rather than just a magic dependency.

timabell (approved):

PR #308, opened 2021-04-30
Clear location field when changing school in resend nomination flow

callumacrae:
### Context

https://trello.com/c/cm5YvYr9

### Testing

yes

### Review Checks

not sure - i worked in cypress

### DIFF
--- a/app/views/nominations/request_nomination_invite/review.html.erb
+++ b/app/views/nominations/request_nomination_invite/review.html.erb
@@ -13,7 +13,7 @@
 
     <div class="govuk-!-margin-bottom-9">
       <%= render partial: 'nominations/school_details', locals: { school: @nomination_request_form.school, title: "Your school" } %>
-      <%= govuk_link_to "Change school", choose_location_request_nomination_invite_path(continue: true), class: "govuk-link--no-visited-state" %>
+      <%= govuk_link_to "Change school", choose_location_request_nomination_invite_path, class: "govuk-link--no-visited-state", "data-test": "change-school" %>
     </div>
 
     <%= form_for :confirmation, url: review_request_nomination_invite_path, method: :post do |f| %>
--- a/spec/cypress/integration/nominations/ResendNominations.feature
+++ b/spec/cypress/integration/nominations/ResendNominations.feature
@@ -21,6 +21,18 @@ Feature: Resend nominations flow
     And the page should be accessible
     And percy should be sent snapshot called "Resend nominations review page"
 
+    # Clicking change school link should reset location input
+    When I click on "change school link"
+    Then "location input" should have value ""
+
+    When I type "test" into "location input"
+    And I click on "autocomplete dropdown item" containing "Test"
+    And I click the submit button
+    And I type "test" into "school input"
+    And I click on "autocomplete dropdown item" containing "Test"
+    And I click the submit button
+    Then I am on "resend nominations review" page
+
     When I click the submit button
     Then I am on "resend nominations success" page
     And the page should be accessible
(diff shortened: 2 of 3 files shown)


callumacrae:
[Bug #8: Remove previous data from fields when "changing school"](https://trello.com/c/cm5YvYr9/517-bug-8-remove-previous-data-from-fields-when-changing-school)

app/views/nominations/request_nomination_invite/review.html.erb
```diff
@@ -13,7 +13,7 @@
 
     <div class="govuk-!-margin-bottom-9">
       <%= render partial: 'nominations/school_details', locals: { school: @nomination_request_form.school, title: "Your school" } %>
-      <%= govuk_link_to "Change school", choose_location_request_nomination_invite_path(continue: true), class: "govuk-link--no-visited-state" %>
+      <%= govuk_link_to "Change school", choose_location_request_nomination_invite_path, class: "govuk-link--no-visited-state", "data-test": "change-school" %>
```
kicferk1, line 16: There is some controller code that needs removing too...
```
class Nominations::RequestNominationInviteController < ApplicationController
  before_action :load_nomination_request_form, except: %i[choose_location]

  def choose_location
    @local_authorities = LocalAuthority.all
    unless params[:continue]
      session.delete(:nomination_request_form)
    end
    load_nomination_request_form
  end
```

kicferk1: Actually, what I wrote is only true if we also want to remove that behaviour from `app/views/nominations/request_nomination_invite/choose_school.html.erb` (back link). I imagine we do, right? 

In that case let's remove it there too, and the resulting code would be something like
```
class Nominations::RequestNominationInviteController < ApplicationController
  before_action :load_nomination_request_form, except: %i[choose_location]

  def choose_location
    @local_authorities = LocalAuthority.all
    session.delete(:nomination_request_form)
    load_nomination_request_form
  end
```

callumacrae: It's still used when pressing the back button during the initial flow so is still required

callumacrae: > I imagine we do, right?

Nope! Spoke to Ethan before making this change and he said that that behaviour shouldn't change (design system guidelines).

ethanmills: I don't think we should change the behaviour of the back button. The design system states [Make sure the link takes users to the previous page they were on, in the state they last saw it](https://design-system.service.gov.uk/components/back-link/#:~:text=make%20sure%20the%20link%20takes%20users%20to%20the%20previous%20page%20they%20were%20on%2C%20in%20the%20state%20they%20last%20saw%20it)

ethanmills: I wish github loaded comments in realtime

kicferk1: I struggle to see how we think it's a good UX when we go 1 step back, but horrible when we go 2 steps back. Is that because on review we show the address, so the user has a legitimate 'oh, the name was right but location is not' moment?

callumacrae: > I struggle to see how we think it's a good UX when we go 1 step back, but horrible when we go 2 steps back. Is that because on review we show the address, so the user has a legitimate 'oh, the name was right but location is not' moment?

This doesn't really seem clear to me either, but it's what they've asked for

ethanmills (approved):
LGTM

kicferk1 (approved):

PR #311, opened 2021-05-04
ECF-RP-564: partnership reminder email

ethanmills:
### Context
7 days after we send a notification to a school that a Lead Provider has declared a partnership with them, we want to send a reminder email if no error has been reported for the partnership

- Change activejob to use delayed_job in most environments
- Add partnership reminder job
- Add fields to partnership notification emails

### Review Checks

I'm really not sure how best to do this.


jamesrose (approved):
Looks good to me!

spec/services/partnership_notification_service_spec.rb
```diff
@@ -88,4 +114,71 @@
...
+        create(:user, :induction_coordinator, schools: [school], email: contact_email)
+      end
+
+      before(:all) do
+        RSpec::Mocks.configuration.verify_partial_doubles = false
+      end
+
+      after(:all) do
+        RSpec::Mocks.configuration.verify_partial_doubles = true
+      end
```
kicferk1, line 163: What does this do?

spec/support/delayed_job.rb
```diff
@@ -49,29 +49,34 @@ module DelayedJobMatchers
...
 
   define :delay_execution_of do |method_name|
     match do |actual|
       jobs = query_jobs(actual, method_name)
-      return jobs.any? unless @arguments
+      return jobs.any? unless @arguments || @at
 
       jobs.any? do |job|
         @arguments.args_match?(*job.payload_object.args)
+        job.run_at == @at if @at.present?
```
kicferk1, line 68: Uhh, will that now ignore the args check? Ruby likes returning the last line, and I think we used to return the results of whether our arguments match the real arguments. But now we return whether the job was run at the correct time?

jamesrose (approved):

kicferk1 (approved):

PR #314, opened 2021-05-05
CPDEL-385: Add very basic users endpoint

kicferk1:
### Context
E&L will need to know the list of users from R&P. Coupled with https://github.com/DFE-Digital/ecf-engage-and-learn/pull/79 to actually hit the client.

Add Users controller to our API. Generate docs for it. Hide it in any environments with real data.

### Guidance to review
What that list exactly is, what fields they will need, which subset of users depending on time, whether it is paginated, and so on are all great questions, but they can be tackled one by one in later PRs. 

### Testing
I checked that I receive and deserialize the data correctly in E&L app, when running both things locally. Other than that, not sure how much there is to test here. I guess I should add a unit test in rspec for it. 

### Review Checks

Hit `/api-docs` and look at our nice swagger doc.

### DIFF
--- a/spec/docs/users_spec.rb
+++ b/spec/docs/users_spec.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require "swagger_helper"
+
+describe "API", type: :request do
+  path "/api/v1/users" do
+    get "Returns all users" do
+      operationId :public_api_v1_user_index
+      tags "user"
+      produces "application/json"
+
+      response "200", "Collection of users." do
+        schema type: :object,
+               properties: {
+                 users: {
+                   type: :array,
+                   items: {
+                     properties: {
+                       id: { type: :string },
+                       email: { type: :string },
+                       full_name: { type: :string },
+                     },
+                     required: %w[id email full_name],
+                   },
+                 },
+               }
+        run_test!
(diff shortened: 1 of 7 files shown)


spec/docs/users_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require "swagger_helper"
+
+describe "API", type: :request do
+  path "/api/v1/users" do
```
ethanmills, line 6: This needs to just be `/users` I think? It's generating `curl -X GET "https://ecf-review-pr-314.london.cloudapps.digital/api/v1/api/v1/users" -H "accept: application/json"`

ethanmills (approved):
I think I'd like to see a request spec for this, and the endpoint needs changing for swagger. Happy other than that. Have we thought about how we're doing authn in future?

kicferk1:
> I think I'd like to see a request spec for this, and the endpoint needs changing for swagger. Happy other than that. Have we thought about how we're doing authn in future?

I haven't thought of it in great depth, but bearer tokens seem fine to me - or whatever we plan to make lead providers use. Once we generate them we can stick them in rails secrets in E&L app.

PR #316, opened 2021-05-05
Setup and readme for automerge of schema.rb

timabell:
### Context

The schema.rb will conflict often, this aims to smooth this.

https://github.com/pocke/merge_db_schema

It requires some local setup to work, documented in the readme.

I'm not yet 100% sure if this is an improvement as it's a little tricky to test. If anyone has the opportunity to try it that would be good. I don't want to rush this one in in case it turns out to be worse rather than better.

* Enable merge_db_schema
* Document usage in readme under developer setup

### Guidance to review

* See if pulling this in and *not* doing the setup causes any problems (I don't want to make dev setup harder)
* Find a conflicting branch with schema changes and try a merge, see if this makes it any easier.

Please report back in the PR comments.

We should also consider whether https://gist.github.com/jeromedalbert/afab24e80102b41d75d5f5538f1459c4 is a better solution. I'm wondering if it will be as it doesn't require any local modification of `.git/config` which this PR does.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -106,6 +106,8 @@ group :development, :test do
   gem "multi_json"
   gem "open_api-rswag-specs"
   gem "rswag"
+
+  gem "merge_db_schema"
 end
 
 group :development do
--- a/Gemfile
+++ b/Gemfile
@@ -106,8 +106,6 @@ group :development, :test do
   gem "multi_json"
   gem "open_api-rswag-specs"
   gem "rswag"
-
-  gem "merge_db_schema"
 end
 
 group :development do
@@ -117,6 +115,8 @@ group :development do
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem "spring"
   gem "spring-watcher-listen", "~> 2.0.0"
+
+  gem "merge_db_schema"
 end
 
 group :test do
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -106,6 +106,8 @@ group :development, :test do
   gem "multi_json"
   gem "open_api-rswag-specs"
   gem "rswag"
+
+  gem "merge_db_schema"
 end
```
blackrat, line 111: Minor comment. Should it be in development only, rather than development and test?

blackrat (changes requested):

Gemfile
timabell: I actually have no idea, cough.

timabell: But yes, moved to dev only as that makes more sense to me.

jamesrose:
How much of an issue is this? I'd be a little hesitant introducing a new Gem for such a straight-forward conflict.

timabell:
> How much of an issue is this? I'd be a little hesitant introducing a new Gem for such a straight-forward conflict.

I think reducing developer friction is a worth cause. I'm not hugely keen on this approach though. I think I'll try the magic attribute and see if that's tidier. I'll leave this open for a bit in case anyone has other thoughts.

timabell:
Leaving this for now as I'm not sufficiently confident it's a good idea. We can always come back to it.

PR #318, opened 2021-05-05
Ecf rp 563 notification banner

BroiSatse:
### Context

When induction coordinator selected CIP for give cohort and, simultaneously, some provider has reported recruitment of given school for FIP, we are now displaying a warning message on all the pages related for this cohort within partnership challenge window togehter with a challenge link.

### Testing

That requires some more though as for now there is no way of reporting partnerships on the site yet. I'm hoping the [partnership creating PR](https://github.com/DFE-Digital/early-careers-framework/pull/299) will be merged before this goes into QA.


app/components/school_recruited_transition_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= govuk_notification_banner(title: "Important") do |banner| %>
+  <% banner.add_heading(text: "#{delivery_partner.name} has confirmed they will be delivering an induction programme to your early career teachers, starting in #{cohort.display_name}.") %>
+  <%# TODO: Add a correct link! %>
+  <p>If this is a mistake, <%= govuk_link_to "report it now", challenge_partnership_path(partnership: partnership.id) %></p>
+  <p>If you do not report this by <%= partnership.challenge_deadline.strftime("%-d %B %Y, %l.%M %p") %>, your programme and next steps will change</p>
```
ethanmills, line 5: I've been formatting the deadline as `"%d/%m/%Y"`. Either way, we should make it consistent 🙂

app/views/schools/cohorts/add_participants.html.erb
```diff
@@ -5,6 +5,8 @@
   href: schools_cohort_path(@cohort.display_name))
 %>
 
+<%= render SchoolRecruitedTransitionComponent.new(school_cohort: @school_cohort) %>
+
```
ethanmills, line 9: Would it not be better to use a layout for this, like https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/views/layouts/admin/base.html.erb ?

spec/components/school_recruited_transition_component_spec.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe SchoolRecruitedTransitionComponent, type: :component do
+  subject(:component) { described_class.new school_cohort: school_cohort }
+  # subject(:rendered) { Capybara.string(render_inline(component).to_html) }
```
ethanmills, line 7: remove?

spec/components/school_recruited_transition_component_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  let(:school_cohort) { create :school_cohort, induction_programme_choice: induction_choice }
+  let(:school) { school_cohort.school }
+  let(:cohort) { school_cohort.cohort }
+
+  context "when school did not choos cip" do
```
ethanmills, line 13: choose

spec/components/school_recruited_transition_component_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  #   rendered_component = render_inline(described_class.new(text: "Done")).to_html
+  #   expect(rendered_component).to include("Done")
+  #   expect(rendered_component).to include("green")
+  # end
+  #
+  # it "renders grey tag for unknown" do
+  #   rendered_component = render_inline(described_class.new(text: "Test string")).to_html
+  #   expect(rendered_component).to include("Test string")
+  #   expect(rendered_component).to include("grey")
+  # end
```
ethanmills, line 62: ?

ethanmills:
Would be good to have a percy snapshot for this component. @callumacrae any advice?

app/views/schools/cohorts/add_participants.html.erb
BroiSatse: Yeah, was thinking about this as well. I though adding a single component is not enough to warrant new layout. If I were to use a layout, I'd need to create new base controller as well and change all the controllers ancestry (which sort of make sense).

Aaaaand, I just realized that such a base controller already exists! :D Yup, doing it then!

spec/components/school_recruited_transition_component_spec.rb
BroiSatse: Don't judge me, it was late :P

BroiSatse:
@ethanmills Yes, i think it makes perfect sense to use percy for components. I think the standart way of doing this would be to have an extra endpoint (in test and development) rendering just a single components so that they can be percied in isolation. As mentioned earlier - using percy in feature test does not spark joy :P

app/components/school_recruited_transition_component.html.erb
BroiSatse: I agree, probably question for Al. In any case we should store the common formatting somewhere - either I18n,  Time::DATE_FORMATS (however this would not work for dates AFAIR) or just a helper (I hate helpers!)

ethanmills: Shani is the person here, this is definitely content 🙂

jamesrose: It looks like we already have the [GOV.UK date formats](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates) in `config/initializers/date_formats.rb`.

BroiSatse: @jamesrose - Good find! It looks it only contains date formats though. I'll add some for the time as well (That is after I confirm we want to display time here :P)

BroiSatse:
[Notification banner - school preparing to do a CIP but recruited by a training provider and in 14 day transition period](https://trello.com/c/QiB9Dxj7/563-notification-banner-school-preparing-to-do-a-cip-but-recruited-by-a-training-provider-and-in-14-day-transition-period)

Gemfile
```diff
@@ -118,6 +118,7 @@ group :development do
 end
 
 group :test do
+  gem "capybara"
```
kicferk1, line 121: Why do we need capybara?

jamesrose: Echo'ing this - seems like a lot for just this content matcher?  https://github.com/DFE-Digital/early-careers-framework/pull/318/files#diff-27bfa81d2f2f9f521f25ada1cab4cd1d875238941cabbbf6b91127de926883a9R37

BroiSatse: @jamesrose Agree - removed it for now. However I'd love to add this at later stage so that we can test our component content in isolation.

kicferk1 (approved):

spec/support/components.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Support
+  module ComponentsTests
+    extend RSpec::Matchers::DSL
+
+    define :render do
+      match(&:render?)
+    end
```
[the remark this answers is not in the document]
BroiSatse: @kicferk1 `render?` is part ov ViewComponent library: https://viewcomponent.org/guide/conditional_rendering.html. And yes, this super-small matcher is just to enable single-line tests to avoid superfluous spec titles. Also `it { is_epxected.to render }` is just easier to understand (IMO).

PR #321, opened 2021-05-06
ECF-RP-545: csv validation

ethanmills:
### Context
Once we have  a CSV of school urns uploaded, we need to parse it and find any errors.

- Display errors in uploaded CSV
- Redirect to confirm page if no errors
- Connect data with the confirm page
- Remove temporary pages

### Guidance to review
I'm a little concerned that we're downloading the recently upload CSV from S3, I think several times. I think I'd like to improve this in a separate ticket. Thoughts?

### Testing
- Unit tests for each kind of CSV error
- Request tests for pages and redirects
- Cypress test for end-to-end journey

### Review Checks

Log in as lead-provider@example.com
Use a CSV containing the following URNs:
000001
000002
000005
99
101377
100773


ethanmills:
[CSV has been processed - errors](https://trello.com/c/TAjUCSoI/545-csv-has-been-processed-errors)

app/controllers/lead_providers/confirm_schools_controller.rb
```diff
@@ -46,8 +33,24 @@ def confirm
...
+      if session[:partnership_csv_upload_id].present?
+        load_partnership_csv_upload
+      elsif session[:confirm_schools_form].present?
+        @confirm_schools_form = ConfirmSchoolsForm.new(session[:confirm_schools_form])
+      end
+    end
+
+    def load_partnership_csv_upload
+      partnership_csv_upload = PartnershipCsvUpload.find(session[:partnership_csv_upload_id])
+      session.delete(:partnership_csv_upload_id)
```
kicferk1, line 45: I would delete it after we create the form, but that's just paranoid me

app/controllers/lead_providers/partnership_csv_uploads_controller.rb
```diff
@@ -7,26 +7,37 @@ def new
...
-      @partnership_csv_upload = PartnershipCsvUpload.new(upload_params.merge(
-                                                           lead_provider_id: current_user.lead_provider_profile.lead_provider.id,
-                                                           delivery_partner_id: session[:delivery_partner_id],
-                                                         ))
+      @partnership_csv_upload = PartnershipCsvUpload.new(
+        upload_params.merge(
+          lead_provider_id: current_user.lead_provider_profile.lead_provider.id,
+          delivery_partner_id: session[:delivery_partner_id],
+        ),
+      )
```
kicferk1, line 21: I asked this on another PR and I still don't get it. Why does `upload_params` have parameters that we are always overriding here with `merge`?

app/controllers/lead_providers/partnership_csv_uploads_controller.rb
```diff
@@ -7,26 +7,37 @@ def new
...
+          redirect_to errors_lead_providers_report_schools_partnership_csv_uploads_path
+        end
       else
         render :new
       end
     end
 
-    def error_page; end
+    def errors
+      partnership_csv_upload = PartnershipCsvUpload.find(session[:partnership_csv_upload_id])
```
kicferk1, line 36: I am not too thrilled about having multiple session parameters in the flow. It makes tracking when we delete them hard :/ Can we make them into a serializable form? Or have a single hash for this flow?

app/controllers/lead_providers/report_schools_controller.rb
```diff
@@ -27,6 +27,7 @@ def check_delivery_partner
 
     def success
       @confirm_schools_form = ConfirmSchoolsForm.new(session.delete(:confirm_schools_form))
+      session.delete(:delivery_partner_id)
```
kicferk1, line 30: Do we want to delete `partnership_csv_upload_id` from session too?

app/views/lead_providers/partnership_csv_uploads/errors.html.erb
```diff
@@ -0,0 +1,53 @@
...
+    <li><%= @errors.count %> row has errors</li>
+  <% else %>
+    <li><%= @errors.count %> rows have errors</li>
+  <% end %>
+
+  <% if (@valid_schools.count) == 1 %>
+    <li><%= @valid_schools.count %> row matches available schools</li>
+  <% else %>
+    <li><%= @valid_schools.count %> rows match available schools</li>
+  <% end %>
```
kicferk1, line 17: Rails likes to pretend it knows how English works, is there a nicer way to pluralise those?

app/views/lead_providers/partnership_csv_uploads/errors.html.erb
```diff
@@ -0,0 +1,53 @@
+<% content_for :title, "CSV results" %>
+<% content_for :before_content, govuk_back_link(text: "Back", href: new_lead_providers_report_schools_partnership_csv_uploads_path) %>
+<h1 class="govuk-heading-xl">Your CSV has errors</h1>
```
kicferk1, line 3: Do we want the content do be in a grid?

app/views/lead_providers/partnership_csv_uploads/errors.html.erb
```diff
@@ -0,0 +1,53 @@
...
+        <td class="govuk-table__cell"><%= error[:school_name] %></td>
+        <td class="govuk-table__cell error-message"><%= error[:message] %></td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
+
+<div class="govuk-button-group">
+  <%= govuk_link_to(
+        "Continue with #{(@valid_schools.count)} #{"school".pluralize((@valid_schools.count))}",
```
kicferk1, line 45: Like that, this is much better than pluralising things ourselves!

spec/requests/lead_providers/confirm_schools_spec.rb
```diff
@@ -78,7 +78,8 @@
       )
     end
 
-    it "schedules partnership notifications" do
+    # TODO: ECF-RP-564: reenable
+    xit "schedules partnership notifications" do
```
kicferk1, line 82: Why are we disabling it?

app/controllers/lead_providers/partnership_csv_uploads_controller.rb
ethanmills: Yeah really not sure here. I'll see if I can simplify it

kicferk1:
> I'm a little concerned that we're downloading the recently upload CSV from S3, I think several times. I think I'd like to improve this in a separate ticket. Thoughts?

Eh, this is a file of schools for a single delivery partner, right? How big do we expect it to get (2MB seems like a vast over-estimation to me)? And how much traction do we expect this endpoint to get from our 6 lead providers? 

I'm happy for you to look into it more in a new ticket, but I am not sure how important it really is.

app/controllers/lead_providers/partnership_csv_uploads_controller.rb
ethanmills: The thing is, we really want to delete them at different times. We want to delete this one as soon as we've processed it, but keep the delivery partner ID for a while (through CSV reuploads etc.)

app/controllers/lead_providers/report_schools_controller.rb
ethanmills: That's deleted earlier, in confirm

app/views/lead_providers/partnership_csv_uploads/errors.html.erb
ethanmills: I took a look, but could only find pluralizing single words

app/views/lead_providers/partnership_csv_uploads/errors.html.erb
ethanmills: 🤷 not sure. I went without because it's quite a wide table. Haven't been given any guidance on this

app/views/lead_providers/partnership_csv_uploads/errors.html.erb
ethanmills: Rails gets confused by the whole "has" vs "have"

spec/requests/lead_providers/confirm_schools_spec.rb
ethanmills: Because I've changed the parameters in the notify template for the other ticket, and this fails

ethanmills: Enabled here https://github.com/DFE-Digital/early-careers-framework/pull/311/commits/b004dbf5e36d92c28412b6e70980d5c1e1d827bb#diff-82fbcf218cd13c64d2f09e7d683d8cbce1d962969a11de99c3647c07c401e784R81

app/controllers/lead_providers/partnership_csv_uploads_controller.rb
ethanmills: I removed LP and DP from `upload_params`, since they're not in the form

app/views/lead_providers/partnership_csv_uploads/errors.html.erb
kicferk1: Arghh, and this is the time when using i18n would actually make our lives easier. https://stackoverflow.com/questions/8790093/how-to-pluralize-there-is-are-n-object-objects

Fine, let's keep it the way it is, sad but I don't think it's easily fixable. Unless someone experienced with Rails has a quick solution.

app/views/lead_providers/partnership_csv_uploads/errors.html.erb
ethanmills: I'll get some design review for this whole journey and apply any minor markups like this separately

kicferk1 (approved):

spec/models/partnership_csv_upload_spec.rb
```diff
@@ -33,4 +38,106 @@
...
+      given_the_csv_contains_urns([valid_school.urn, closed_school.urn])
+
+      expect(@subject.valid_schools).to contain_exactly(valid_school)
+    end
+  end
+
+private
+
+  def given_the_csv_contains_urns(urns)
+    file = Tempfile.new
```
goodviber, line 136: Any reason to not use a fixture rather than a Tempfile?

ethanmills: Because it's easier to use generated schools than to hardcode it all

app/models/partnership_csv_upload.rb
```diff
@@ -17,7 +17,7 @@ def invalid_schools
   def valid_schools
     invalid_urns = invalid_schools.map { |error| error[:urn] }
     valid_urns = urns - invalid_urns
-    School.where(urn: valid_urns)
+    valid_urns.map { |urn| School.find_by(urn: urn) } # We need to preserve order uploaded
```
kicferk1, line 20: Ech, how about
```
schools = School.where(urn: valid_urns)
valid_urns.map { |urn| schools.find { |school| school.urn == urn} }
```
to reduce n+1 queries here?

ethanmills: They're all cached by this point from `find_school_errors` anyway, don't think it will make much difference

kicferk1: Is Rails clever enough to know that we loaded them before?

We don't operate on a scope we got from before, we operate on `School` :/

PR #325, opened 2021-05-06
Case insensitive user email field

BroiSatse:
### Context

User could not log in by entering their email address with any capital letter.

As devise is already down-casing all email I've decided to use citext for users email column - this way case insenstive search is now enabled on email column by default (in case we'd ever need to do sth similar in other places)

### DIFF
--- a/db/migrate/20210506161800_case_insensitive_user_emails.rb
+++ b/db/migrate/20210506161800_case_insensitive_user_emails.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class CaseInsensitiveUserEmails < ActiveRecord::Migration[6.1]
+  def up
+    enable_extension("citext")
+
+    change_column :users, :email, :citext, null: false
+  end
+
+  def down
+    change_column :users, :email, :string, null: false, default: ""
+
+    # GovUK PaaS has "citext" extension enabled by default which cannot be disabled.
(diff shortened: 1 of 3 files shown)


BroiSatse:
[BUG: make the sign-in page case not sensitive to capital letters](https://trello.com/c/b017utHV/601-bug-make-the-sign-in-page-case-not-sensitive-to-capital-letters)

db/migrate/20210506161800_case_insensitive_user_emails.rb
```diff
@@ -0,0 +1,15 @@
...
+  def up
+    enable_extension("citext")
+
+    change_column :users, :email, :citext, null: false
+  end
+
+  def down
+    change_column :users, :email, :string, null: false, default: ""
+
+    disable_extension("citext")
```
ethanmills, line 13: I feel this might cause trouble. See https://docs.cloud.service.gov.uk/deploying_services/postgresql/#:~:text=you%20cannot%20remove%20these%20mandatory%20extensions%20from%20a%20postgresql%20service%20instance

BroiSatse: Ah, interesting! Good catch :)

BroiSatse: OK, left the comment with the link. I still need to enable the extension so it works locally - `enable_extension` is nicely idempotent though so no risk for production db.

jamesrose (approved):

PR #328, opened 2021-05-07
ECF-RP-564: Create partnership request

ethanmills:
### Context
When a lead provider declares a partnership with a school who said they were doing CIP, we essentially ignore it for two weeks to give the school time to challenge. This introduces a data structure to represent that

- Create a new PartnershipRequest, which inherits from a common base with Partnership
- Create mechanisms for a PartnershipRequest to become a Partnership, or be challenged
- Create PartnershipRequests when a CIP school is reported by an LP

### Review Checks

### DIFF
--- a/app/jobs/partnership_finalisation_job.rb
+++ b/app/jobs/partnership_finalisation_job.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class PartnershipFinalisationJob < ApplicationJob
+  discard_on ActiveJob::DeserializationError
+
   def perform(partnership_request)
     return if partnership_request.reload.blank?
 
--- a/app/jobs/partnership_finalisation_job.rb
+++ b/app/jobs/partnership_finalisation_job.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class PartnershipFinalisationJob < ApplicationJob
+  def perform(partnership_request)
+    return if partnership_request.reload.blank?
+
+    originator = partnership_request.versions.first.whodunnit
+    PaperTrail.request(whodunnit: originator) do
+      partnership_request.finalise!
+    end
+  end
+end
--- a/app/models/base_partnership.rb
+++ b/app/models/base_partnership.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class BasePartnership < ApplicationRecord
+  self.abstract_class = true
+
+  CHALLENGE_WINDOW = 14.days.freeze
+
+  before_create :set_challenge_deadline
+
+  belongs_to :school
+  belongs_to :lead_provider
+  belongs_to :cohort
+  belongs_to :delivery_partner
+
+  has_paper_trail
+
+  def in_challenge_window?
+    challenge_deadline > Time.zone.now
+  end
+
+protected
+
+  def set_challenge_deadline
+    self.challenge_deadline ||= Time.zone.now + CHALLENGE_WINDOW
+  end
+end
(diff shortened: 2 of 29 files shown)


app/jobs/partnership_finalisation_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class PartnershipFinalisationJob < ApplicationJob
+  discard_on ActiveJob::DeserializationError
+
+  def perform(partnership_request)
+    return if partnership_request.reload.blank?
```
kicferk1, line 7: When does this happen?

app/jobs/partnership_finalisation_job.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    originator = partnership_request.versions.first.whodunnit
+    PaperTrail.request(whodunnit: originator) do
+      partnership_request.finalise!
+    end
```
kicferk1, line 12: What does the `first` version mean? First ever created? Last ever created? First bu alphabetical order of our uids?

app/models/partnership_csv_upload.rb
```diff
@@ -66,4 +66,14 @@ def find_school_errors
...
+
+  def partnered_with_current_lead_provider?(school)
+    school.lead_provider(Cohort.current.start_year) == lead_provider ||
+      PartnershipRequest.where(lead_provider: lead_provider, school: school, cohort: Cohort.current).any?
+  end
+
+  def partnered_with_other_lead_provider?(school)
+    school.lead_provider(Cohort.current.start_year).present? ||
+      PartnershipRequest.where(school: school, cohort: Cohort.current).any?
+  end
```
kicferk1, line 78: Ehhh, n+1 queries here we go 😢 
I don't see an easy fix though.

app/models/base_partnership.rb
```diff
@@ -0,0 +1,31 @@
...
+  def challenged?
+    false
+  end
+
+protected
+
+  def set_challenge_deadline
+    self.challenge_deadline ||= Time.zone.now + CHALLENGE_WINDOW
+  end
+end
```
kicferk1, line 31: Can this be private? On one hand it's only called in this file, on the other, who knows with dynamic typing...

app/models/partnership_request.rb
```diff
@@ -0,0 +1,52 @@
...
+        school: school,
+        cohort: cohort,
+        challenge_deadline: Time.zone.now,
+      )
+      school_cohort = SchoolCohort.find_by!(school_id: school_id, cohort_id: cohort_id)
+      school_cohort.update!(induction_programme_choice: "full_induction_programme", core_induction_programme: nil)
+
+      update_notification_emails(partnership)
+      destroy!
+    end
```
kicferk1, line 20: Question one: Why do we change challenge deadline? 

Question two, much much bigger: Why do we need a separate model to represent a partnership request? 
Do we expect multiple lead providers to request to partner with one school? If so do we want to destroy all other requests when we finalise this one? 
Maybe a less complex (polymorphic) solution would be to create the partnership when LP requests to partner, and have a status on it that tells us whether it was accepted, rejected or pending? Then you can accomplish all the changes here with a status change.

app/jobs/partnership_finalisation_job.rb
ethanmills: papertrail orders its versions by created_at, something it calls "timestamp order"

ethanmills:
Replaced by #368

PR #333, opened 2021-05-10
CPDTP-120 - Add a new endpoint for receiving "participant started" events

timabell:
https://dfedigital.atlassian.net/browse/CPDTP-120

### Context

This patch is a joint effort with Pavel, Tim and Paul from the Track-and-Pay team.

From the jira ticket:

> As a provider 
> 
> I want to submit an event indicating that a participant has started their course 
> 
> so that I can receive appropriate payment

* New endpoint at `/api/v1/early_career_teacher_participants` that allows POSTing a single "participant started" event and stores that in `participant_event` table by use of the `paper_trail` gem.
* The state of a participant is managed with the AASM gem (state machine) and stored in the `participant_profiles` table.
* The new endpoint uses bearer-token based authentication of lead providers with tokens stored in new table `lead_provider_api_tokens`.
* New rake task for creating access tokens (second commit)

### Guidance to review

`rspec-default_http_header` - this gem has been added which allows not to repeat the auth headers in the specs by way of `DefaultHttpHeader` in `spec/rails_helper.rb`

`ParticipantProfiles` - this is separate to the existing `EarlyCareerTeacherProfile` because: 1) participants includes mentors (no model for these yet) and 2) separtation of concers.

Bearer tokens vs JWT - Other services within Teacher Services are already using bearer token auth in like this. JWT was also considered but there was nothing in what we are doing here that required it.

The `aasm` state machine for the participant is allowing us to track the real-world state of a participant with consistency, the history of the transitions is then kept by `paper_trail`.

participant_profile-aasm diagram
Diagram generated with [aasm-diagram](https://github.com/Katee/aasm-diagram)

### Testing

You can see the new endpoint defined here: https://ecf-review-pr-333.london.cloudapps.digital/api-docs/index.html

#### Locally

1. Run seeds with `bundle exec rake db:seed` to get a provider
2. Generate a new access token for local testing with `bundle exec rake lead_provider:generate_token "Capita"`
3. POST to the endpoint with curl or swagger http://localhost:3000/api-docs/index.html 

e.g. 

```
$ bundle exec rake db:seed
$ bundle exiec rake lead_provider:generate_token "Capita"
Generated API token for lead provider (6a375564-d1cb-4171-b1f9-ab890f8e1229): bTNKpkANRdtsHkwcj-ED
$ curl -iX POST "http://localhost:3000/api/v1/early-career-teacher-participants" -H  "accept: application/json" -H  "Authorization: Bearer bTNKpkANRdtsHkwcj-ED"
```

### Review Checks
- [ ] All pages have automated accessibility checks via cypress - n/a
- [ ] All pages have visual tests via cypress + percy - n/a
- [ ] All `School` queries are correctly scoped - `eligible` when they need to be - n/a

https://ecf-review-pr-333.london.cloudapps.digital/api-docs/

Screenshot 2021-05-13 at 15 57 30


spec/docs/early_career_teacher_participation_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+require "swagger_helper"
+
+RSpec.describe "Early Career Teacher Participation", type: :request do
+  let(:user) { create(:user) }
+  let(:lead_provider) { create(:lead_provider) }
+  let(:token) { LeadProviderApiToken.create_with_random_token!(lead_provider: lead_provider) }
+  let(:bearer_token) { "Bearer #{token}" }
+
+  before do
+    default_headers[:Accept] = "application/vnd.dfe.gov.uk.v1"
```
blackrat, line 12: See previous comment about harm vs necessity. This one can be removed too.

spec/docs/early_career_teacher_participation_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    default_headers[:Authorization] = bearer_token
+  end
+
+  path "/early-career-teacher-participants" do
+    post "Add Early Career Teacher to Course" do
+      operationId :api_v1_create_ect_participant
+      tags "ect_participant"
+      consumes "application/json"
+      parameter name: "Authorization", in: :header, required: false, type: :string, description: "The bearer token associated with a lead provider"
+      parameter name: "Accept", in: :header, required: false, type: :string, description: "The version of the current API"
```
blackrat, line 22: This isn't needed anymore.

app/controllers/api/v1/early_career_teacher_participants_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+        return head :no_content unless params[:id]
+
+        begin
+          user = User.find(params[:id])
+          early_career_teacher = user.early_career_teacher_profile
+          ParticipantAddService.call(early_career_teacher)
+          head :created
+        rescue ActiveRecord::RecordNotFound
+          head :unprocessable_entity
+        end
```
kicferk1, line 18: I wonder if it needs to be in `begin rescue end` pattern. Maybe we can do it by using `def rescue end`?

app/controllers/concerns/lead_provider_authenticatable.rb
```diff
@@ -0,0 +1,24 @@
...
+
+module LeadProviderAuthenticatable
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :authenticate
+  end
+
+  def authenticate
+    authenticate_or_request_with_http_token do |unhashed_token|
```
kicferk1, line 11: Where does this `authenticate_or_request_with_http_token` come from?

app/services/participant_add_service.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class ParticipantAddService
+  attr_accessor :participant_profile
+
+  class << self
+    def call(early_career_teacher_profile)
+      new(early_career_teacher_profile).call
+    end
+  end
```
kicferk1, line 10: Is this the same as
```
def self.call(early_career_teacher_profile)
  new(early_career_teacher_profile).call
end
```
? If so, could we stick with that syntax?

app/services/participant_add_service.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def call
+    participant_profile.join!
+  end
```
kicferk1, line 14: What's the difference between `join!` and `save!`?

db/migrate/20210510130434_create_participant_events.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class CreateParticipantEvents < ActiveRecord::Migration[6.1]
+  def change
+    create_table :participant_events, id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
kicferk1, line 5: Are we sure we want a random uid to be the default here? :/

db/migrate/20210510130434_create_participant_events.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.string :item_type, null: false
+      t.string :event, null: false
+      t.string :whodunnit
+      t.json :object
+      t.json :object_changes
+      t.datetime :created_at
+      t.uuid :item_id, default: -> { "gen_random_uuid()" }, null: false
```
kicferk1, line 12: what is `item`? And do we really want a random uid as a default?

db/migrate/20210510130434_create_participant_events.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class CreateParticipantEvents < ActiveRecord::Migration[6.1]
+  def change
+    create_table :participant_events, id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+      t.string :item_type, null: false
+      t.string :event, null: false
+      t.string :whodunnit
+      t.json :object
+      t.json :object_changes
```
kicferk1, line 10: What is `object` exactly?

lib/tasks/lead_provider_token.rake
```diff
@@ -0,0 +1,18 @@
...
+namespace :lead_provider do
+  desc "creates API token for lead provider ID parameter"
+  task generate_token: :environment do
+    begin
+      lead_provider = LeadProvider.find(ARGV[1])
+    rescue StandardError
+      lead_provider = LeadProvider.find_by(name: ARGV[1])
+    end
+    token = LeadProviderApiToken.create_with_random_token!(lead_provider: lead_provider)
+    puts "Generated API token for lead provider (#{lead_provider.id}): #{token}"
```
kicferk1, line 12: Maybe use Rails logger for it, and... Will that get logged in our logs? Are we sure this is the safest way of retrieving the token?

mattonfoot:
We should probably define the API security in line with the specifications for OAS https://swagger.io/docs/specification/authentication/bearer-authentication/

timabell:
I'm not sure about returning 204 when not enough data is sent, seems like providers might think they've got it working when they haven't.

app/controllers/api/v1/early_career_teacher_participants_controller.rb
mattonfoot: I think it should return 404 for RecordNotFound?

timabell:
Todo: add early career teacher to seeds and expand usage example in PR to show sending that user's uuid

db/migrate/20210510130434_create_participant_events.rb
mattonfoot: Should we lean on postgreSQL to give us the UUIDs https://pawelurbanek.com/uuid-order-rails

app/services/participant_add_service.rb
blackrat: It's not exactly the same. class << self is more flexible. For one it defines a block for the singleton class and allows methods within it to be scoped as public, protected or private. You have to just through hoops to do that with self. defined class instance methods.

db/migrate/20210510130434_create_participant_events.rb
blackrat: > Are we sure we want a random uid to be the default here? :/

Standard for paper_trail (as also used in the "versions" table that it was based on.

db/migrate/20210510130434_create_participant_events.rb
blackrat: > What is `object` exactly?

Standard for paper_trail (as also used in the "versions" table that it was based on.
https://github.com/paper-trail-gem/paper_trail

spec/models/participant_profile_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    @object = create(:participant_profile)
+  end
+  let(:ect_profile) { create(:early_career_teacher_profile) }
+
+  describe "state transitions" do
+    it "should have an initial state of assigned" do
+      expect(@object).to have_state(:assigned)
+    end
+
+    it "should only have expected transitions from the initial state" do
```
mattonfoot, line 16: How do state changes relate to the start event?

spec/swagger_helper.rb
```diff
@@ -10,7 +10,7 @@
       info: {
         title: "API documentation",
         version: "v1",
-        description: "Auto generated doc",
+        description: "Auto generated document, run `bundle exec rake rswag` to regenerate",
```
mattonfoot, line 13: This is normally a description of the API to help API clients understand the purpose of the API rather than how to manage the API

config/routes.rb
```diff
@@ -29,14 +29,13 @@
...
-  namespace :api do
+  namespace :api, defaults: { format: "json" } do
     resources :school_search, only: %i[index]
     resource :notify_callback, only: :create, path: "notify-callback"
 
     namespace :v1 do
-      unless Rails.env.staging? || Rails.env.production?
-        resources :users, only: :index
-      end
+      resources :early_career_teacher_participants, only: %i[create], path: "early-career-teacher-participants"
```
ethanmills, line 37: Not convinced by the naming here. Maybe `participants/early-career-teacher`?

db/migrate/20210510130434_create_participant_events.rb
ethanmills: You should be able to omit all of this. We have a gem that sorts this for us: https://github.com/DFE-Digital/early-careers-framework/blob/develop/Gemfile#L30

cf the other migrations

spec/models/lead_provider_api_token_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+require "rails_helper"
+
+RSpec.describe LeadProviderApiToken, type: :model do
+  it "generates a hashed token that can be used" do
+    unhashed_token = LeadProviderApiToken.create_with_random_token!(lead_provider: create(:lead_provider))
+
+    expect(
+      LeadProviderApiToken.find_by_unhashed_token(unhashed_token),
+    ).to eql(LeadProviderApiToken.last)
```
ethanmills, line 11: We should have an `order(:created_at)` here

app/models/participant_profile.rb
```diff
@@ -0,0 +1,41 @@
...
+    class_name: "ParticipantEvent",
+  }
+  belongs_to :early_career_teacher_profile
+
+  include AASM
+
+  aasm column: :state
+
+  aasm do
+    state :assigned, initial: true
```
ethanmills, line 14: What does this mean?

app/models/participant_profile.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class ParticipantProfile < ApplicationRecord
```
ethanmills, line 3: Would you consider calling this something like `ParticipationRecord`? We already have the idea of a profile, and use it to mean something else

ethanmills:
As far as I can see, there is nothing here to stop a lead provider from sending events for ECTs working with other providers

spec/docs/early_career_teacher_participation_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      response 201, "Successful" do
+        let(:params) { { "id" => user.id } }
+        run_test!
+      end
+
+      response "204", "No Content" do
+        run_test!
+      end
+
+      response 422, "Unprocessable Entity" do
```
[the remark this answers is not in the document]
blackrat: > I would expect 404 not 422
Noted. Possibly a wider discussion on the use of 422. The thought that I had originally was to use them for invalid state transitions and things that were syntactically correct, but semantically wrong. This may be a definition too far though.

spec/models/participant_profile_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+require "rails_helper"
+
+RSpec.describe ParticipantProfile, type: :model do
+  before :each do
+    @object = create(:participant_profile)
+  end
+  let(:ect_profile) { create(:early_career_teacher_profile) }
+
+  describe "state transitions" do
+    it "should have an initial state of assigned" do
```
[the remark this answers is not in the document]
blackrat: It's the default preliminary state that a user has been added to the state transitions and is available to "Join" a course and become "active". "assigned" rather than "entered", since the user has already been entered via registration and exists, but the join event has not yet been triggered. That's almost certainly going to happen immediately from that state, but it doesn't seem good practice to initialise them in the "joined" state.

app/controllers/api/v1/early_career_teacher_participants_controller.rb
blackrat: Ah reading further, you probably were more referring to the 422 than the 204. The thought was that if it got that far, then the call was syntactically correct, but created an error. It was something that I'd used, and seen used for API calls with correct data structure, but bad data. I'm open to changing it.

app/controllers/api/api_controller.rb
```diff
@@ -1,4 +1,7 @@
 # frozen_string_literal: true
 
-class Api::ApiController < ActionController::API
+module Api
+  class ApiController < ActionController::API
+    include Pundit
```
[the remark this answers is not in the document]
blackrat: I don't think we are. I removed it when I added in the rescue_from clause from James' suggestion and it had no noticeable effect.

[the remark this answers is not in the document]
blackrat: > Where are you using Pundit?

I don't think we are. I removed it when I added in the rescue_from clause from James' suggestion and it had no noticeable effect.

spec/models/participant_profile_spec.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ParticipantProfile, type: :model do
+  before :each do
+    @object = create(:participant_profile)
```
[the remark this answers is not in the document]
blackrat: > descriptive names
Done

spec/models/participant_profile_spec.rb
blackrat: These are atomic participant states. For the "start" event, the expectation for the participant is to set them up initially and to make them "active". I can run you through the logic tomorrow if you want, but it's probably a bit deeper than PR comments can convey correctly.

app/models/participant_profile.rb
blackrat: Ok. Maybe we don't fully grok your use of Profile.

app/services/participant_add_service.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class ParticipantAddService
```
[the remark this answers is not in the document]
mattonfoot: maybe `ParticipantInductionProgressRecordService`

spec/docs/early_career_teacher_participation_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    default_headers[:Authorization] = bearer_token
+  end
+
+  path "/early-career-teacher-participants" do
+    post "Add Early Career Teacher to Course" do
+      operationId :api_v1_create_ect_participant
+      tags "ect_participant"
+      consumes "application/json"
+      parameter name: "Authorization", in: :header, required: false, type: :string, description: "The bearer token associated with a lead provider"
+      parameter name: :params, in: :body, required: false, type: :string, description: "The unique id of the participant"
```
mattonfoot, line 21: Would this not be `in: :path` ?

mattonfoot: Also seems there is a validation issue with the generated swagger https://validator.swagger.io/validator/debug?url=https://ecf-review-pr-333.london.cloudapps.digital/api-docs/v1/api_spec.json

app/controllers/concerns/lead_provider_authenticatable.rb
```diff
@@ -0,0 +1,25 @@
...
+
+module LeadProviderAuthenticatable
+  extend ActiveSupport::Concern
+  include ActionController::HttpAuthentication::Token::ControllerMethods
+
+  included do
+    before_action :authenticate
+  end
+
+  def authenticate
```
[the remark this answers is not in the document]
blackrat: Done

spec/docs/early_career_teacher_participation_spec.rb
blackrat: > Would this not be `in: :path` ?

Needs to be in body. First pass thin slice is single participant, later we will expand to multiple.

app/services/participant_add_service.rb
kicferk1: Right, and are we using any of this flexibility here?

I would rather have one consistent syntax for defining class methods, and we use `def self.method_name` in a bunch of places already. Plus, as I see it, this method is meant to be public, so using syntax that only lets you define public methods is fine. Or am I missing something and it is, in fact, private?

db/migrate/20210510130434_create_participant_events.rb
kicferk1: Well, I imagine the `uid` in there comes from 
```
class ChangeItemIdTypeToUuid < ActiveRecord::Migration[6.1]
  def change
    remove_column :versions, :item_id, :bigint
    remove_index :versions, column: %i[item_type item_id], if_exists: true
    add_column :versions, :item_id, :uuid, default: "gen_random_uuid()", null: false
    add_index :versions, %i[item_type item_id]
  end
end
```
In that migration we were changing a field on an existing table, so not having a default was hard. In your case you are adding a new table, so not having a default should be easy, or so I hope.

db/migrate/20210510130434_create_participant_events.rb
kicferk1: Could we visually separate the parts of migration that are 'ours' and the parts that are for papertrail? Maybe with a comment for the latter saying "Fields for paper_trail"?

plisovin:
> We should probably define the API security in line with the specifications for OAS https://swagger.io/docs/specification/authentication/bearer-authentication/

Added in 220f65a

app/services/participant_add_service.rb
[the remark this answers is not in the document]
blackrat: `InductParticipant`? It's only adding the participant to the Participation Record (currently still called ParticipantProfile)

app/services/participant_add_service.rb
timabell: I prefer the `class << self` style over the `self.xxx` style just on the basis that it makes the grouping of singleton object instance methods vs normal instance methods clearer and keeps them all together. That preference is just stylistic though. Personally I'd defer to MaK's superior knowledge of this stuff.

db/migrate/20210510130434_create_participant_events.rb
timabell:
```
add_column :versions, :item_id, :uuid, default: "gen_random_uuid()", null: false
```
This doesn't seem right to me; the `item_id` for paper_trail is as I understand it a reference to the primary key field of whichever entity is being tracked. As such having a randomly generated guid would mean the data would be corrupt because it wouldn't point to any valid primary key. I'm inclined to ignore the migration you have quoted and just make our own one correct.

kicferk1: I would also be inclined to make your migration correct. 

I mentioned that migration because that's how I think the random uid got into schema, which you copied to get your migration in the first place. Which I really think you shouldn't need or do.

app/services/participant_add_service.rb
jamesrose: I don't think this is big deal either way - we know clearly what both mean.

db/migrate/20210510130434_create_participant_events.rb
blackrat: Sure. Dilemma that I had was that it was originally a bigint and we seem to have moved away from those. It won't ever take the default, since the value will be filled in on creation in any case. 

Something that I think is way more important is that it does seem to suggest that the production papertrail may have been broken by that migration which I think is a much bigger issue that how we define a default (that is never actually going to be used). Unless you're not actually using papertrail for anything, or it wasn't used for anything prior to the migration. I'd be very concerned about the data integrity if it's being relied on for anything.

I'll remove the default here, but keep the uid column type.

spec/models/lead_provider_api_token_spec.rb
plisovin: Actioned here c2f792f5

timabell:
ADR documenting the decision to version the API in URL instead of as a header: https://github.com/DFE-Digital/cpd-service-manual/blob/c43864282fb2eaae6064621cf1a8927028bf1cad/source/adr/0003-use-versioning-strings-in-api-urls.md
Cheers @mattonfoot

blackrat (dismissed):

spec/docs/early_career_teacher_participation_spec.rb
plisovin: Now it's fixed -> https://ecf-review-pr-333.london.cloudapps.digital/api-docs
Split into public API doc and secure API doc, drop down list on top:

Screenshot 2021-05-11 at 16 01 38

app/controllers/concerns/lead_provider_authenticatable.rb
```diff
@@ -0,0 +1,25 @@
...
+    authenticate_or_request_with_http_token do |unhashed_token|
+      @current_lead_provider_api_token = LeadProviderApiToken.find_by_unhashed_token(unhashed_token)
```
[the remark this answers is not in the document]
plisovin: This method belongs to the model, it's not a db column ( we don't store unhashed tokens, remember? ). We can rename it to avoid confusion though

lib/tasks/lead_provider_token.rake
timabell: We don't know how this is going to work for production yet so this is just a starting point that we can iterate on.

spec/swagger_helper.rb
timabell: Good spot, yes that's not too helpful is it.

Selection_20210512-01-655fd655fd655fd

@mattonfoot  I'll do something simple for now but if you have suggestions for something better let us know or do a PR when this is in. As I understand it this swagger UI will be seen by lead provider tech people.

config/routes.rb
plisovin: Renamed to ```participants/induction-progress``` in 5416057
https://ecf-review-pr-333.london.cloudapps.digital/api-docs/index.html?urls.primaryName=API%20V1%20Docs

app/controllers/api/public/v1/users_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  module Public
+    module V1
+      class UsersController < Api::ApiController
+        def index
+          render json: { users: User.all.as_json(only: %i[id email full_name]) }
+        end
+      end
+    end
+  end
+end
```
kicferk1, line 13: What is this doing here? Why was it moved to `Public`?
Please revert that change, it will cause conflicts with https://github.com/DFE-Digital/early-careers-framework/pull/341 and more importantly, the users api is not going to be public. We will add authentication to it.

config/routes.rb
```diff
@@ -29,13 +29,19 @@
...
-        resources :users, only: :index
+      namespace :participants do
+        resources :induction_progress, only: %i[create], path: "induction-progress"
+      end
+    end
+
+    namespace :public do
+      namespace :v1 do
+        resources :users, only: :index unless Rails.env.staging? || Rails.env.production?
       end
```
kicferk1, line 45: As mentioned above, I really don't think `users` api will end up as `public` - I don't think returning everyone's emails and full names should be in that namespace. We will add auth to it to make it non-public before we release to staging and live.

spec/docs/users_spec.rb
```diff
@@ -2,8 +2,8 @@
 
 require "swagger_helper"
 
-describe "API", type: :request do
-  path "/users" do
+describe "API", type: :request, swagger_doc: "public_v1/api_spec.json" do
+  path "/api/public/v1/users" do
```
kicferk1, line 6: Ditto

kicferk1 (changes requested):
Not a big fan of touching users api, and I really don't think it should be moved to `public` namespace.

app/controllers/api/public/v1/users_controller.rb
plisovin: Well, it was moved to public for that reason, no authentication)

config/routes.rb
plisovin: Then you can move to to non-public before releasing to staging and live, simple

kicferk1: Look, users api is a work-in-progress, which is why it only lives in non-real environments. I chose not to include authentication in the API in the first place because I plan to reuse a lot of your work for it, and because E&L doesn't have a proper live environment yet.

Moving the api between namespaces makes our life harder - we want to iterate and improve it, but if it changes namespace while we are doing it, we will have to re-write all of our work. We know where it will eventually end up, so why would we move it around before releasing it?

And on a completely different note: If you strongly feel it should live elsewhere, that change should live in its own PR. This one already has enough changes in it.

plisovin: Ok, how about if we move it back to non-public namespace and disable authentication with a non-live flag, so all swagger docs and rspec tests will pass?

kicferk1 (approved):

jamesrose (approved):
Nice work! 🚢

timabell:
> As far as I can see, there is nothing here to stop a lead provider from sending events for ECTs working with other providers

Noted on the backlog for future work, cheers @ethanmills
https://dfedigital.atlassian.net/browse/CPDTP-167

app/controllers/concerns/lead_provider_authenticatable.rb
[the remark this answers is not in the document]
blackrat: I think it's probably because of the way that RSpec and Papertrail interact. Checked against the existing papertrail implementation and it's the same there. As we're not actually doing anything with it now, I'm happy to leave this as is for now. We probably won't need whodunnit when we get to the point of making sure that the lead_provider who created the record is the lead_provider who is triggering that state and making sure that the participant is counted for the correct lead_provider.

app/models/participant_profile.rb
blackrat: Done

timabell:
squashed & rebased with an updated commit message in preparation for merging

PR #323, opened 2021-05-06
Add Your schools to lead provider

tonyheadford:
### Context
[Trello card](https://trello.com/c/n7Yoih9i/580-view-and-manage-schools-school-overview-your-schools)
Adds __Your schools__ view for lead providers

### Changes proposed in this pull request
Lead provider view of __Your schools__

### Guidance to review
Given you are signed in as a Lead Provider that has some schools
Visit `/lead-providers/your-schools` (there is no direct link to the page currently)
You should see a list of your schools and be able to search the list by school name/urn or delivery partner name

#### Issues:
* In-place nav at the top as this isn't available yet
* Hard-coded '0' for __Mentors__ column in the table
* No __Group/network__ column in the table

### Review Checks
- [ ] All pages have automated accessibility checks via cypress
- [ ] All pages have visual tests via cypress + percy
- [ ] All `School` queries are correctly scoped - `eligible` when they need to be

### How can I view this in a review app?

When the lead provider has no schools as yet for the cohort (removed Recruitment target box):
image
When the search returns no results:
image

PR #334, opened 2021-05-10
Add school detail view for Lead providers

tonyheadford:
### Context
[Trello card](https://trello.com/c/oLsN0lH9/581-view-and-manage-schools-school-detail)
Builds upon the __Your schools__ functionality from PR #323 

Add school detail view

### Guidance to review
When signed-in as a lead provider, visit `/lead-providers/your-schools` (there is no direct link to the page currently) then clicking on one of the school links in the table of schools will show the school details page for that school.

### Review Checks

### DIFF
--- a/app/components/primary_nav_component.rb
+++ b/app/components/primary_nav_component.rb
@@ -8,12 +8,13 @@ class PrimaryNavComponent < ViewComponent::Base
   class NavItemComponent < ViewComponent::Base
     attr_reader :path
 
-    def initialize(path:)
+    def initialize(path:, selected: false)
       @path = path
+      @selected = selected
     end
 
     def current_section?(path)
-      request.path.include?(path)
+      @selected || request.path.include?(path)
     end
   end
 end
--- a/app/controllers/lead_providers/school_details_controller.rb
+++ b/app/controllers/lead_providers/school_details_controller.rb
@@ -3,8 +3,22 @@
 module LeadProviders
   class SchoolDetailsController < ::LeadProviders::BaseController
     def show
-      @school = School.eligible.find(params[:id])
-      @selected_cohort = Cohort.find(params[:selected_cohort_id])
+      @school = find_school
+      @selected_cohort = selected_cohort_or_current
+    end
+
+  private
+
+    def selected_cohort_or_current
+      Cohort.find_by(id: params[:selected_cohort_id]) || Cohort.current
+    end
+
+    def find_school
+      year = selected_cohort_or_current.start_year
+      search_scope = School.eligible
+
+      search_scope = search_scope.partnered_with_lead_provider(current_user.lead_provider.id, year) if current_user.lead_provider?
+      search_scope.find(params[:id])
     end
   end
 end
--- a/app/controllers/lead_providers/school_details_controller.rb
+++ b/app/controllers/lead_providers/school_details_controller.rb
@@ -15,10 +15,7 @@ module LeadProviders
 
     def find_school
       year = selected_cohort_or_current.start_year
-      search_scope = School.eligible
-
-      search_scope = search_scope.partnered_with_lead_provider(current_user.lead_provider.id, year) if current_user.lead_provider?
-      search_scope.find(params[:id])
+      School.eligible.partnered_with_lead_provider(current_user.lead_provider.id, year).find(params[:id])
     end
   end
 end
--- a/app/controllers/lead_providers/your_schools_controller.rb
+++ b/app/controllers/lead_providers/your_schools_controller.rb
@@ -12,7 +12,7 @@ module LeadProviders
                            @cohorts.find_by(start_year: Time.zone.today.year)
                          end
 
-      @schools = School.partnered_with_lead_provider(@lead_provider.id, @selected_cohort.start_year)
+      @schools = School.eligible.partnered_with_lead_provider(@lead_provider.id, @selected_cohort.start_year)
         .includes(:early_career_teachers)
         .order(:name)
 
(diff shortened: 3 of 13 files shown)


kicferk1:
I think an interactive rebase that drops all commits before and including `7c38cdd` is in order

kicferk1:
Or maybe one that just removes the merged commits, I am not quite sure what happened in this git history

app/views/lead_providers/school_details/show.html.erb
```diff
@@ -1,3 +1,104 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h2 class="govuk-heading-m"><%= @selected_cohort.start_year %> participants</h2>
+  </div>
+</div>
+
+<div class="govuk-grid-row govuk-!-margin-bottom-6">
+  <div class="govuk-grid-column-one-third">
+    <div class="dashboard-numbers key-number">
+      <p class="govuk-heading-xl">0</p>
```
ethanmills, line 34: Why 0 here and the correct number in the previous screen?

app/controllers/lead_providers/your_schools_controller.rb
```diff
@@ -12,7 +12,7 @@ def index
                            @cohorts.find_by(start_year: Time.zone.today.year)
                          end
 
-      @schools = School.partnered_with_lead_provider(@lead_provider.id, @selected_cohort.start_year)
+      @schools = School.eligible.partnered_with_lead_provider(@lead_provider.id, @selected_cohort.start_year)
```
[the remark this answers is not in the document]
tonyheadford: Yes, in my testing I had unwittingly added a school that wasn't eligible to a partnership and it was showing in my list of schools. Whether that would happen in real-life or whether an eligible school could close/merge/split/academise or otherwise change so it is no longer eligible is a good question.

app/views/lead_providers/school_details/show.html.erb
tonyheadford: Doh!

tonyheadford:
> Or maybe one that just removes the merged commits, I am not quite sure what happened in this git history

I started this branch off of the up-to-date `develop` and then after a while realised I needed a few things from my other PR so rebased off of that.

app/components/primary_nav_component.rb
```diff
@@ -8,12 +8,13 @@ class PrimaryNavComponent < ViewComponent::Base
...
 
-    def initialize(path:)
+    def initialize(path:, selected: false)
       @path = path
+      @selected = selected
     end
 
     def current_section?(path)
-      request.path.include?(path)
+      @selected || request.path.include?(path)
```
kicferk1, line 17: That looks a bit scary. I think there should only ever be one `current_section` element - but with this change, there could be many. Well, technically there could be many beforehand too, if you got very unlucky with paths.

I wonder if we should change the logic so that `path` is only used for navigation, and `selected` is the only thing responsible for selecting current section. Then you could still have multiple current sections, but at least it would be pretty obvious from reading template code that it's happening.

app/controllers/lead_providers/school_details_controller.rb
```diff
@@ -3,8 +3,22 @@
...
+
+    def selected_cohort_or_current
+      Cohort.find_by(id: params[:selected_cohort_id]) || Cohort.current
+    end
+
+    def find_school
+      year = selected_cohort_or_current.start_year
+      search_scope = School.eligible
+
+      search_scope = search_scope.partnered_with_lead_provider(current_user.lead_provider.id, year) if current_user.lead_provider?
```
kicferk1, line 20: Is the `curret_user` ever not going to be a lead provider? Given that we inherit from `::LeadProviders::BaseController`?

app/views/lead_providers/school_details/show.html.erb
```diff
@@ -1,3 +1,106 @@
...
+        </dd>
+      </div>
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          Induction tutor
+        </dt>
+        <dd class="govuk-summary-list__value">
+          <% if @school.induction_coordinators.any? %>
+            <%= mail_to @school.induction_coordinators.first.email %>
+          <% end -%>
```
kicferk1, line 101: Do we need the `-` in `-%>`?

app/views/lead_providers/your_schools/index.html.erb
```diff
@@ -1,9 +1,22 @@
...
+    <%= component.nav_item(path: lead_providers_your_schools_path) do %>
+      Schools
+    <% end %>
+  <% end %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Your schools</h1>
+    <%= govuk_link_to 'Confirm more schools', start_lead_providers_report_schools_path, button: true %>
```
kicferk1, line 17: We don't have a linter for that, but we try to keep a convention of using double quotes in views.

app/components/primary_nav_component.rb
tonyheadford: Yes, I think this would need some more thought as it looks like this is normally used as content for a layout where it's not known which item is selected? Feels like refactoring that shouldn't be in this PR?

app/controllers/lead_providers/school_details_controller.rb
tonyheadford: All the tests pass, I'll remove the admin check in the base controller then and simplify that find school method

app/components/primary_nav_component.rb
tonyheadford: Good idea, Tech-debt ticket created

kicferk1 (approved):

jamesrose (approved):

PR #341, opened 2021-05-12
/api/v1/users returns valid jsonapi

asmega:
### Context

- I gather we want to conform to jsonapi spec defined at https://jsonapi.org

- Change existing `/api/v1/users` endpoint so it conforms to jsonapi spec
- Added `jsonapi-serializer` it seems far better maintained compared to https://github.com/jsonapi-rb/jsonapi-rails and https://github.com/jsonapi-rb/jsonapi-rb despite the fact it only deals with the serialisation part I feel this is the better option
- Uncommented an rspec config so developers can focus on tests

### Guidance to review

- This is a breaking change to the API but as far as I'm aware is not in use so should be safe to change

### Review Checks
- [X] All pages have automated accessibility checks via cypress
- [X] All pages have visual tests via cypress + percy
- [X] All `School` queries are correctly scoped - `eligible` when they need to be

- curl -I https://ecf-review-pr-341.london.cloudapps.digital/api/v1/users
- check `Content-Type` header
- should match spec defined at https://jsonapi.org/format/#content-negotiation-servers
- curl https://ecf-review-pr-341.london.cloudapps.digital/api/v1/users
- check reponse body
- should match spec defined at https://jsonapi.org/format/#document-structure

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -4,7 +4,7 @@ module Api
   module V1
     class UsersController < Api::ApiController
       def index
-        render json: { users: User.all.as_json(only: %i[id email full_name]) }
+        render json: UserSerializer.new(User.all).serializable_hash.to_json
       end
     end
   end
--- a/Gemfile
+++ b/Gemfile
@@ -76,6 +76,8 @@ gem "delayed_job_active_record"
 # Acts as State Machine for participant states
 gem "aasm"
 
+gem "jsonapi-serializer"
+
 # OpenApi Swagger
 gem "open_api-rswag-api", ">= 0.1.0"
 gem "open_api-rswag-ui", ">= 0.1.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -177,6 +177,8 @@ GEM
     jsonapi-rspec (0.0.11)
       rspec-core
       rspec-expectations
+    jsonapi-serializer (2.2.0)
+      activesupport (>= 4.2)
     jwt (2.2.2)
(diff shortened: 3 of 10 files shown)


kicferk1 (commented):
Do we need to update `spec/docs/users_spec.rb` and run `bundle exec rake rswag:specs:swaggerize` to make `/api-docs` route return up to date documentation? 

I presume yes, since the response is a bit different than what it was. Definitely a tomorrow problem though, if by any chance you are reading this on Wednesday :)

app/controllers/api/v1/users_controller.rb
```diff
@@ -2,9 +2,9 @@
 
 module Api
   module V1
-    class Api::V1::UsersController < Api::ApiController
+    class Api::V1::UsersController < BaseController
```
blackrat, line 5: Shouldn't have the Api::V1 prefixing the name inside the namespace. This translates to Api::V1::Api::V1::UsersController, which will confuse the autoloader.

blackrat (changes requested):

app/controllers/api/v1/users_controller.rb
asmega: good spot. copy paste fail

asmega:
> Do we need to update `spec/docs/users_spec.rb` and run `bundle exec rake rswag:specs:swaggerize` to make `/api-docs` route return up to date documentation?

@kicferk1 Good spot. I wasn't aware of the existence of this. I've updated the docs and generated the new output

jamesrose (approved):

blackrat (approved):

PR #349, opened 2021-05-13
Switching to database-backed session store for larger session limit

BroiSatse:
### Context

Currently we're using cookie session, which is limited to mere 4kb. This is problematic as certain journeys require storing much larger amount of information (like lists of school ids) within a session.

Switching to ActiveRecord session store virtually removes limits from session size. 

### Notes

There are some other potenital solution to this problem. One of them would be creating another table in the database representing a list of schools and only storing list id in the session. I am absolutely open for discussion which one is better. I have gone for alternative session store as we are using sessions a lot and it was quicker to implement.

Also it is worth mentioning that AR session store is much better in serialization than cookie store to the extend that it does not require manual serialization/deserialization. This has actually broken a number of tests as we were relying on implicit serialization of the cookie store.


app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -14,7 +14,7 @@ def start
     else
       load_nominate_induction_tutor_form
       @nominate_induction_tutor_form.token = token
-      session[:nominate_induction_tutor_form] = @nominate_induction_tutor_form
+      session[:nominate_induction_tutor_form] = @nominate_induction_tutor_form.as_json
```
BroiSatse, line 17: This is interesting - AR storage actually deserialized the entire from, so `session[:nominate...]` was an instance of the form object in the subsequent requests. I've maintained the original logic, but potentially we could just use this.

app/jobs/session_trim_job.rb
```diff
@@ -0,0 +1,17 @@
...
+  def perform
+    Rails.logger.info "Trimming session store..."
+
+    cutoff_period = ENV.fetch("SESSION_DAYS_TRIM_THRESHOLD", 30).to_i.days.ago
+
+    ActiveRecord::SessionStore::Session
+      .where("updated_at < ?", cutoff_period)
+      .delete_all
+  end
+end
```
BroiSatse, line 17: This job is to run daily to remove stale session objects. We might need to consider how this works with our session timeout - do we even have any?

db/migrate/20210513103622_add_sessions_table.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddSessionsTable < ActiveRecord::Migration[6.1]
+  def change
+    create_table :sessions do |t|
+      t.string :session_id, null: false
+      t.jsonb :data
```
BroiSatse, line 7: 1GB session limit!

spec/support/request.rb
```diff
@@ -6,7 +6,9 @@ class SessionHelperController < ApplicationController
       respond_to :json
 
       def update
-        session[params[:key]] = params[:value]
+        value = params.require(:value)
+        value = value.to_unsafe_hash if value.respond_to?(:to_unsafe_hash)
+        session[params[:key]] = value
```
BroiSatse, line 11: These changes are required due to the serialization capabilities. `params[:value]` was returning an instance of ActionController::Parameters and it was stored and returned from the session.

app/jobs/session_trim_job.rb
kicferk1: I am pretty sure we have an expiration on all cookies, including session cookies. I don't remember exactly how long they last, a week?

app/jobs/session_trim_job.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class SchoolDataImporterJob < CronJob
```
kicferk1, line 3: That name doesn't sound right ;)

kicferk1: Also, is it actually triggered? The school import job, which this is clearly copied from, is not actually triggered at any point, because we don't want to worry about school updates yet. But this sounds like something we would like to run.

app/controllers/nominations/nominate_induction_coordinator_controller.rb
kicferk1: I prefer `as_json`, I think storing the form was a bit of an oversight.

app/jobs/session_trim_job.rb
```diff
@@ -0,0 +1,17 @@
...
+  queue_as :session_trim
+
+  def perform
+    Rails.logger.info "Trimming session store..."
+
+    cutoff_period = ENV.fetch("SESSION_DAYS_TRIM_THRESHOLD", 30).to_i.days.ago
+
+    ActiveRecord::SessionStore::Session
+      .where("updated_at < ?", cutoff_period)
+      .delete_all
```
kicferk1, line 15: Is this the best logic? Do we have access to sessions here? I don't know how Rails handles them (never really looked into that), but I think we could do something like 'fetch all session store data that have id pointing at an expired or deleted session, and drop them'. Which would mean we would not need to worry about timing - we would simply defer that logic to session logic.

That said, it is a bit surprising to me that the gem doesn't automatically delete session store when session expires / is deleted. Thoughts?

kicferk1 (commented):
I wonder if we can make a more elegant solution for removing session data.

One is 'use session logic for it'. The other worth considering is 'store it against a user, only ever have one per user, avoid deleting them'. 

Thoughts?

app/controllers/nominations/nominate_induction_coordinator_controller.rb
BroiSatse: Definitely. Cookie store was implicitly casting the form onto json. I think I'd prefer `attributes` (from ActiveModel::Attributes) here as `as_json` is not that well defined. But it is close enough! :)

app/jobs/session_trim_job.rb
BroiSatse: Yeah, that was something I was actually about to ask you (and clearly forgot) - how are our Cron jobs scheduled? I can clearly see we have cron pattern defined here so I think i assumed this will be scheduled automatically on deploy with eager-load.

app/jobs/session_trim_job.rb
BroiSatse: So the gem actually defines a rake task `db:session:trim` for cleaning up the session, this logic was copied from that task (Wasn't sure if we should load the task into the web server logic). Alternative was to invoke rake via a system call or, ideally add rake task call to the cron schedule (do we need rufus-scheduler for that?)

So the way sessions works is that we still keep session id within a cookie (with expiry date), but session record itself has no expiration date, which is why they suggest running that rake regularly to prevent uncontrolled session table growth. I agree, we could do it better - we can replace the default Session record definition with our custom one, but I think for now the default is good enough.

kicferk1 (approved):

db/migrate/20210513103622_add_sessions_table.rb
ethanmills: We'll need to keep an eye on how this goes. Our prod db is 100GB currently

app/jobs/session_trim_job.rb
ethanmills: Two weeks

db/migrate/20210513103622_add_sessions_table.rb
kicferk1: I guess we can limit it to 1MB if we want?

app/jobs/session_trim_job.rb
ethanmills: This is not going to do anything by itself. We need to go back to where Michal lifted this from, and get the final piece for this to work. When we do that (now?), we need to prevent the _actual_ SchoolDataImporterJob from running.

config/environments/deployed_development.rb
```diff
@@ -47,7 +47,6 @@
   # config.cache_store = :mem_cache_store
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
-  config.active_job.queue_adapter = :delayed_job
   # config.active_job.queue_name_prefix = "early_career_framework_production"
```
ethanmills, line 50: Could you remove the lonely comments too please?

app/jobs/session_trim_job.rb
kicferk1: Someone's looking at an old version ;)

ethanmills:
We should consider finishing ripping off the cron stuff from claim: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/c6e8f78dc49414c5b2d9f2dcd6326b8e0d804843/app/models/cron_job_scheduler.rb

Just need to make sure we don't schedule the school import job

db/migrate/20210513103622_add_sessions_table.rb
BroiSatse: I mean if we are anything close to this numbers than we're doing sth wrong. At the moment we put hard limit of 2MB files for CSV upload so 1MB session limit would not work.

app/jobs/session_trim_job.rb
BroiSatse: I have added `cron:schedule` rake task which re-schedules selected jobs :)

ethanmills (approved):

PR #358, opened 2021-05-14
Updated 'what happens next'

jamesrose:
### Context

Superseding old PR https://github.com/DFE-Digital/early-careers-framework/pull/351

### DIFF
--- a/app/views/nominations/request_nomination_invite/success.html.erb
+++ b/app/views/nominations/request_nomination_invite/success.html.erb
@@ -6,16 +6,10 @@
 
     <h2 class="govuk-heading-m">What happens next</h2>
     <ul class="govuk-list govuk-list--bullet">
-      <li>your school will receive an email with a link to nominate an induction tutor</li>
-      <li>this link will expire in 21 days</li>
+      <li>sent to the email address your school has listed on our register, <%= govuk_link_to "Get information about schools", "https://www.get-information-schools.service.gov.uk/" %></li>
       <li>
-        we cannot disclose the email address. It will be an administration email, for example, office@, admin@ or head@
+        your school must follow this link to nominate an induction lead or tutor
       </li>
-      <li>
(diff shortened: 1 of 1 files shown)


app/views/nominations/request_nomination_invite/success.html.erb
```diff
@@ -6,16 +6,10 @@
 
     <h2 class="govuk-heading-m">What happens next</h2>
     <ul class="govuk-list govuk-list--bullet">
-      <li>your school will receive an email with a link to nominate an induction tutor</li>
-      <li>this link will expire in 21 days</li>
+      <li>sent to the email address your school has listed on our register, <%= govuk_link_to "Get information about schools", "https://www.get-information-schools.service.gov.uk/" %></li
```
blackrat, line 9: "www" not required on the url

app/views/nominations/request_nomination_invite/success.html.erb
```diff
@@ -6,16 +6,10 @@
...
       <li>
-        we cannot disclose the email address. It will be an administration email, for example, office@, admin@ or head@
+        your school must follow this link to nominate an induction lead or tutor
       </li>
-      <li>
-        if your school does not receive this email, ensure your school email address is correct on our register,
-        <%= govuk_link_to "Get information about schools", "https://www.get-information-schools.service.gov.uk/" %>
```
blackrat, line 16: although it's not actually your change that had it there. :)

app/views/nominations/request_nomination_invite/success.html.erb
ethanmills, line 9: Looks like we dropped a `>` on the closing tag here. I suspect that what Cypress is annoyed about

jamesrose: @ethanmills Ah, great spot! I must have copied it over from the old PR incorrect.

ethanmills (approved):

PR #366, opened 2021-05-17
Add missing env for database.yml and secret key base

plisovin:
Add secret key base to encrypted credentials for sandbox


config/secrets.yml
```diff
@@ -23,4 +23,7 @@ development:
 test:
   secret_key_base: a5cf81cfe4f58f3b177301038681edc63bb0ae35912ed7e6d35ba911275e53efdf05fb05d5c87095519bb09021c39e3b561bd242fe7d6a75982d8f0150cdca60
 
+sandbox:
+  secret_key_base: a5cf81cfe4f58f3b177301038681edc63bb0ae35912ed7e6d35ba911275e53efdf05fb05d5c87095519bb09021c39e3b561bd242fe7d6a75982d8f0150cdca60
+
```
ethanmills, line 28: This should not be here. It lives in an encrypted credentials file, much like staging. Please cycle the key now.

ethanmills (changes requested):

config/secrets.yml
jamesrose: @ethanmills They are aware. This is a dup of the test key.

ethanmills: But why...? This is what happens when the PR has no description 😛

ethanmills (approved):

jamesrose (approved):

PR #368, opened 2021-05-17
Pending partnerships

ethanmills:
### Context
Replaces #328 as a method for having pending partnerships

When a lead provider declares a partnership with a school who has told us they are doing CIP, we wait 2 weeks before doing anything about it.

- Add pending boolean to partnerships
- Add job to convert pending partnerships

### Review Checks

### DIFF
--- a/app/jobs/partnership_activation_job.rb
+++ b/app/jobs/partnership_activation_job.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class PartnershipActivationJob < ApplicationJob
+  discard_on ActiveJob::DeserializationError
+
+  def perform(partnership)
+    return if partnership.challenged?
+    return unless partnership.pending
+
+    ActiveRecord::Base.transaction do
+      partnership.update!(pending: false)
+      school_cohort = SchoolCohort.find_by!(school_id: partnership.school_id, cohort_id: partnership.cohort_id)
+      school_cohort.update!(induction_programme_choice: "full_induction_programme", core_induction_programme: nil)
+    end
+  end
+end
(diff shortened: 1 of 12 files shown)


kicferk1 (approved):
Small commonization comment, but other than that it looks good to me. Much clearer (in my personal opinion) than polymorphism :)

app/jobs/partnership_activation_job.rb
```diff
@@ -0,0 +1,16 @@
...
+  discard_on ActiveJob::DeserializationError
+
+  def perform(partnership)
+    return if partnership.challenged?
+    return unless partnership.pending
+
+    ActiveRecord::Base.transaction do
+      partnership.update!(pending: false)
+      school_cohort = SchoolCohort.find_by!(school_id: partnership.school_id, cohort_id: partnership.cohort_id)
+      school_cohort.update!(induction_programme_choice: "full_induction_programme", core_induction_programme: nil)
```
blackrat, line 13: Still looks like it should be using the ObserverPattern to me.

ethanmills: Still?

ethanmills: By this you mean a single call to Partnership, which then updates the SchoolCohort? I don't think I agree

BroiSatse:
I wonder if we really need Partnership activation Job at all. How about if, instead of boolean `pending` column we had datetime `active_on` that is set on partnership creation? Then the `active` scope is just `where("active_on < ?", Time.zone.now)`. `active_on` would be either equal to `created_at` or `challenge_deadline`.

blackrat (approved):

PR #372, opened 2021-05-17
Add a dummy landing page for sandbox

plisovin:
Add root for sandbox environment
At the moment there is no content, but will be added once we have it

### DIFF
--- a/app/controllers/sandbox_controller.rb
+++ b/app/controllers/sandbox_controller.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class SandboxController < ApplicationController
+  def show; end
+end
--- a/app/views/sandbox/show.html.erb
+++ b/app/views/sandbox/show.html.erb
@@ -1,4 +1,4 @@
-<%= content_for :title, t('page_titles.sandbox') %>
+<% content_for :title, "API sandbox" %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
--- a/app/views/sandbox/show.html.erb
+++ b/app/views/sandbox/show.html.erb
@@ -1,0 +1,17 @@
+<%= content_for :title, t('page_titles.sandbox') %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t('page_titles.sandbox') %></h1>
+
+    <p class="govuk-body">The API sandbox is a sandbox environment which allows you to experience the participant events.</p>
+
+    <p class="govuk-body">You can:</p>
+
+    <ul class="govuk-list govuk-list--bullet">
+      <li>Authenticate</li>
+      <li>Test the API <%= govuk_link_to 'API integrations', '/api-docs' %></li>
+    </ul>
+
+    </div>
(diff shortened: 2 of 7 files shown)


app/controllers/sandbox_controller.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class SandboxController < ApplicationController
```
blackrat, line 3: Picky point, but controllers should really be plural.

blackrat (approved):

config/routes.rb
```diff
@@ -13,7 +13,12 @@
   get "/pages/:page", to: "pages#show", as: :page
   get "check" => "application#check"
 
-  root "start#index"
+  if Rails.env.sandbox?
```
jamesrose, line 16: I'd test this, at minimum.

app/views/sandbox/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+
+    <p class="govuk-body">You can:</p>
+
+    <ul class="govuk-list govuk-list--bullet">
+      <li>Authenticate</li>
+      <li>Test the API <%= govuk_link_to 'API integrations', '/api-docs' %></li>
+    </ul>
+
+    </div>
+</div>
```
kicferk1, line 17: Couple of general points:
1. We try to use double quotes in views where possible.
2. Either all of it should be internationalised or none of it should imo.
3. Second-to-last `div` is too indented :)

config/routes.rb
plisovin: Tested it locally

app/views/sandbox/show.html.erb
plisovin: Ok, but this will completely change once we get right content. This is just to put a landing page for sandbox web app.

blackrat: I find the first point odd here. Single quotes parse faster, since Ruby is expecting interpolation if double quotes are used. It does appear that Rubocop has been set up that way, which still gets me every time.

kicferk1: I guess the speed bump is negligible compared to everything else that a web app is doing, and we use interpolation quite a lot. And having both types of quotes looks messy.

config/routes.rb
plisovin: Added test in ede93d0

kicferk1 (commented):
Still not all that thrilled about internationalising page title, but that's probably something we will talk over on a dev retro, when we get one.

In the meantime... You can't see it on a dev app, but when I fired it off locally and hacked it to show sandbox behaviour, the page title (on tab hover) was saying something about `translation_missing`...

plisovin:
> Still not all that thrilled about internationalising page title, but that's probably something we will talk over on a dev retro, when we get one.
> 
> In the meantime... You can't see it on a dev app, but when I fired it off locally and hacked it to show sandbox behaviour, the page title (on tab hover) was saying something about `translation_missing`...

Just replaced with static string

kicferk1 (approved):

config/routes.rb
jamesrose: Awesome - thanks.

PR #377, opened 2021-05-18
CPDRP-308: sign out page

ethanmills:
### Context
Add a sign out page

image

### Review Checks

Sign in as admin@example.com
Sign out

### DIFF
--- a/spec/cypress/integration/accessibility.js
+++ b/spec/cypress/integration/accessibility.js
@@ -48,5 +48,9 @@ describe("Accessibility", () => {
     cy.get('[action="/users/sign_in_with_token"] [name="commit"]').click();
     cy.get("h1").should("contain", "User dashboard");
     cy.checkA11y();
+
+    cy.logout();
+    cy.checkA11y();
+    cy.percySnapshot("Logout page");
   });
 });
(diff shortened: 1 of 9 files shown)


spec/cypress/integration/accessibility.js
```diff
@@ -48,5 +48,9 @@ describe("Accessibility", () => {
     cy.get('[action="/users/sign_in_with_token"] [name="commit"]').click();
     cy.get("h1").should("contain", "User dashboard");
     cy.checkA11y();
+
+    cy.logout();
+    cy.checkA11y();
+    cy.percySnapshot("Logout page");
```
kicferk1, line 54: I wonder if we should do that in cucumber

kicferk1 (approved):
Code looks good, but I think a cucumber test might be a better place than cypress test for checking percy and accessibility on logout page.

spec/cypress/integration/accessibility.js
ethanmills: I wanted to keep it with the tests for the login page

ethanmills: (and didn't want to rewrite those in cucumber)

callumacrae (approved):

PR #379, opened 2021-05-18
CPDRP-311: Format sign in form errors correctly

ethanmills:
### Context
Replace the flash alert with some proper design system error messages when the sign in email is blank

Before:
image

After:
image

### Review Checks

try to sign in with a blank email

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -10,6 +10,16 @@ class Users::SessionsController < Devise::SessionsController
   before_action :redirect_to_dashboard, only: %i[sign_in_with_token redirect_from_magic_link]
   before_action :ensure_login_token_valid, only: %i[sign_in_with_token redirect_from_magic_link]
 
+  def new
+    super do
+      if flash.present?
+        flash.clear
+        resource.valid?
+        resource.errors.delete(:full_name)
+      end
+    end
+  end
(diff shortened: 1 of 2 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -10,6 +10,16 @@ class Users::SessionsController < Devise::SessionsController
...
   before_action :ensure_login_token_valid, only: %i[sign_in_with_token redirect_from_magic_link]
 
+  def new
+    super do
+      if flash.present?
+        flash.clear
+        resource.valid?
+        resource.errors.delete(:full_name)
+      end
+    end
```
kicferk1, line 20: What does this syntax do? Call the body first and then the `new` method in the parent?

kicferk1 (commented):
Is this something we want to test?

app/controllers/users/sessions_controller.rb
ethanmills: The method definition in super is 
```
  def new
    self.resource = resource_class.new(sign_in_params)
    clean_up_passwords(resource)
    yield resource if block_given?
    respond_with(resource, serialize_options(resource))
  end
```

So it just allows us to inject a bit of our own code in the middle there

kicferk1 (approved):

PR #381, opened 2021-05-18
CPDEL-438: Add authentication to users endpoint

kicferk1:
### Context
We want to hide users api behind authentication.

Copy-paste the approach of Track and Pay.

### Guidance to review
I went for the simplest approach - making a carbon copy of lead provider tokens and authentication in parallel. The problem is, I am not convinced it is a good idea - I think at least in principle LPs will want to have access to the same APIs that E&L will have access to. And perhaps vice versa.

Of course the difficulty with that is that E&L is not _really_ a lead provider - and the current lead provider tokens belong to lead providers.

I would be tempted to make a single table inheritance, where we get a `ApiToken` as the parent class, and `LeadProviderApiToken` and `EngageAndLearnApiToken` are child classes of that, but it looked like some work and potentially controversial decision, so I opted for talking it through first.

### Testing
Added unit tests, updated docs. 

Try making unauthenticated request to the users endpoint (you can go to `/api-docs` and `execute` as a shortcut). You should get a 401.
Follow the readme to get a token. Make a request with curl or postman, with `Authentication` header being `Bearer your-token`. You should get a list of users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,3 +135,8 @@ bundle exec rake lead_provider:generate_token "name or id"
 ```
 
 Where `"name or id"` is a name or id from the `lead_providers` table.
+
+## Generating a token for E&L api
+1. Get into Rails console for the environment you want to generate the token for.
+2. Run `EngageAndLearnApiToken.create_with_random_token!`
+3. Rails console should output a string, that's your unhashed token, you can keep it and use it to access E&L endpoints.
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -3,6 +3,8 @@
 module Api
   module V1
     class UsersController < Api::ApiController
+      include EngageAndLearnAuthenticatable
+
       def index
         render json: UserSerializer.new(User.all).serializable_hash.to_json
       end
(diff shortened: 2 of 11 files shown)


ethanmills (approved):
Looks good. Do we want to set papertrail whodunnit anywhere?

jamesrose:
I'm in favour of the STI option. Do you want to do it as part of this PR, or later?

kicferk1:
> I'm in favour of the STI option. Do you want to do it as part of this PR, or later?

I will do it here, I think migrations should be nicer that way.

db/migrate/20210518143026_add_token_inheritance.rb
```diff
@@ -0,0 +1,10 @@
+class AddTokenInheritance < ActiveRecord::Migration[6.1]
+  def change
+    change_table :lead_provider_api_tokens, bulk: true do |t|
+      t.change_null :lead_provider_id, true
```
kicferk1, line 4: If we want STI this needs to be nullable - cause other token classes will not necessarily have a lead provider for them (say, engage and learn token)

jamesrose (approved):
Looks good!

PR #382, opened 2021-05-18
Hard delete users, add paper trails

tonyheadford:
### Context
[Jira ticket](https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-290)
Hard delete user and associated profile, remove use of `discard` for these models and add `has_paper_trail`

Replace `discard`ing users with `destroy`ing
Change `has_and_belongs_to_many` relationship to `has_many` and `has_many through` to enable tracking in `paper_trail`
Added `dependent: :destroy` to associations

### Guidance to review
Destroying a User should destroy any associated profile and records of such recorded in `paper_trail`

### Review Checks


app/models/admin_profile.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
-class AdminProfile < BaseProfile
+class AdminProfile < ApplicationRecord
```
kicferk1, line 3: Why are we changing this?

kicferk1: Oh, because we are removing base profile, that makes sense :)

app/models/induction_coordinator_profiles_school.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class InductionCoordinatorProfilesSchool < ApplicationRecord
+  has_paper_trail
```
kicferk1, line 4: I wonder if there is some way to tell Rails that we want to have paper trail with all the default models it creates? 

Oh, I am already scared to day it, is this something we want in our `ApplicationRecord`? Or is that too monkey-patchy for our taste?

ethanmills (approved):

spec/requests/admin/lead_providers/lead_provider_users_spec.rb
```diff
@@ -61,19 +61,22 @@
     end
   end
 
-  describe "DELETE /admin/suppliers/lead_providers/users/:id/" do
-    it "marks the lead_provider profile as deleted" do
+  describe "DELETE /admin/suppliers/lead_providers/users/:id/", versioning: true do
```
kicferk1, line 64: What does `versioning: true` do?

tonyheadford: It enables `PaperTrail` for that block, it's not enabled in specs by default

app/models/induction_coordinator_profiles_school.rb
tonyheadford: Would we always want to have it enabled on every model?  I could envisage some kind of base class or concern that was `Auditable` or something like that

kicferk1: Ehh, always is such a strong word 😂 

On E&L certainly not. But this is not E&L land. Thoughts @ethanmills? @jamesrose?

I think `has_paper_trail` is that concern that you would end up devising. Base class might be more like it, cause then we avoid having `has_paper_trail` all over the place. 

And I guess if we went all the way and changed application record we would not need to add our own models for the default many-to-many relations. But it's also changing the global behaviour of application record, so it sounds a bit scary.

ethanmills: Eh, I think this is fine. There are lots of things we don't particularly care about versioning

kicferk1 (approved):

PR #385, opened 2021-05-18
Add payment calculator

timabell:
### Context

This was previously a separate repository but the wider team have agreed that we don't currently have a need to split this out into separate services.

1. Add payment calculator in separate directory (for clear history of the move)
2. Move everything to the right places in another commit

This is a copy paste from https://github.com/DFE-Digital/cpd-payment-calculations/tree/fd30b3ed074cb9029126963984b49452fe38459d

### Guidance to review

See the new specs mingled with some existing specs with:

```bash
bundle exec rspec -fd spec/lib spec/features
```


timabell:
fixed linter error and rebased

documentation/adr/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,19 @@
+# 1. Record architecture decisions
```
jamesrose, line 1: Should these be in https://github.com/DFE-Digital/cpd-adr?

lib/services/ecf/output_payment_aggregator.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Services
```
jamesrose, line 3:
```suggestion
module PaymentCalculator
```

Can we update the namespace to something that better describes this area?

spec/spec_helper.rb
```diff
@@ -16,7 +16,14 @@
 #
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 #
+require "require_all"
+
 require "aasm/rspec"
+
+# payment calculation code
+require_relative "../lib/initialize_with_config"
+require_all "lib/services"
```
jamesrose, line 25: I don't think this warrants a gem. Can we autoload these?

lib/meta.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Meta
```
jamesrose, line 3: Should this be in the calculator namespace?

lib/initialize_with_config.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require_relative "meta"
+require "active_support"
+require "active_support/core_ext"
+
+module InitializeWithConfig
```
jamesrose, line 7: Should this be in the calculator namespace?

documentation/adr/0001-record-architecture-decisions.md
timabell: I don't think we're going to be using that repo.... @mattonfoot ?

lib/meta.rb
timabell: I think @blackrat wanted to delete it entirely

lib/initialize_with_config.rb
timabell: Seems sensible now it's part of a bigger thing

lib/meta.rb
blackrat: Yes. `InitializeWithConfig` has been changed to make Meta unnecessary.

lib/initialize_with_config.rb
blackrat: It's not necessarily just for the calculator. I'd saw leave it in lib.

lib/services/ecf/output_payment_aggregator.rb
timabell: done, good call

timabell: (specs still to do)

spec/spec_helper.rb
blackrat: @jamesrose Hmm. Do you mean autoload?  Wouldn't it be better to require them explicitly if we don't like that testing gem. 

Setting the config.autoload_paths array seems a little heavy? I'm not even sure that's documented in the file anymore. I've got some experience of doing that, and it's normally quite hard for people to find unless they know how Rails internals work.

blackrat: Or do you mean to do it at the application level the same way that it's being done to force reloading of decorators and serializers? Scoped for test in application.rb?

blackrat: Ah. Lightbulb moment. It's just for Rspec, so I'm guessing you can do that just in the .rspec file.

timabell: @jamesrose Are you happier with 3a593d9ec88a4954e1afe7b3399cb9d87336c8ef by @blackrat that eliminates it?
It does mean that there are lots more require statements around, which maybe is fine.

jamesrose: @timabell @blackrat Yes, I think that's fine 👍

jamesrose (approved):
Nice work getting this in - one step closer to an E2E calculation. 🚢

spec/spec_helper.rb
```diff
@@ -16,7 +16,7 @@
 #
```
timabell, line 16: I'm not yet sure where the problem is, but we seem to now be running the cypress features with rspec which obviously doesn't work. They are then marked as pending with "no such step" hence not failing on CI :worried: 
This was introduced in the first two commits of the branch. Currently tracking this down. Don't merge till this is fixed.

timabell: Okay I think I see the problem, turnip seems to be hard-coded to load all feature files and is picking up the cypress feature files. Dang.

https://github.com/jnicklas/turnip/blob/00bd963dea49c5ff8f87edb1bcafcdb73f310c6a/lib/turnip/rspec.rb#L124

```ruby
  config.pattern += ',**/*.feature'
```

blackrat (changes requested):
Turnip gem is incorrectly picking up the cypress tests, so rejecting this to give us a chance to fix this before someone clicks merge.

spec/spec_helper.rb
timabell: Fixed by adding and configuring `rutabaga` gem instead of directly using `turnip`.

lib/services/ecf/output_payment_aggregator.rb
timabell: specs done

blackrat (approved):

PR #386, opened 2021-05-18
Sentry integration

ethanmills:
### Context
We now have a Sentry.io account for error tracking and performance monitoring. This is the integration for that

Add Sentry for rails and delayed_job in production (only)

### Guidance to review
Would we like sentry on staging? We have a limited number of ingest events
Would we like to send the user email to sentry? I assume we can't? Just sending user ID now

### Testing
I tested this locally, and saw events coming through both for requests and delayed jobs

### Review Checks

Not possible


config/initializers/sentry.rb
```diff
@@ -0,0 +1,37 @@
...
+        case transaction_name
+        when /check/
+          0.0 # ignore healthcheck requests
+        else
+          0.1
+        end
+      else
+        0.0 # We don't care about performance of other things
+      end
+    end
```
kicferk1, line 35: What does this do? Is that setting a chance that request will be sampled?

ethanmills: Yes, performance tracing. They suggest you start low and work up to get the correct volume

jamesrose:
Yes, we can't send the email address to Sentry, or any other PII.

jamesrose (approved):

app/controllers/application_controller.rb
```diff
@@ -5,7 +5,8 @@ class ApplicationController < ActionController::Base
 
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :previous_url_for_cookies_page, except: :check
-  before_action :check_privacy_policy_accepted
+  before_action :check_privacy_policy_accepted, except: :check
```
ethanmills, line 8: Unrelated change so we don't keep hitting the db for the heartbeat

spec/requests/admin/schools/induction_coordinators_spec.rb
```diff
@@ -14,7 +14,7 @@
     it "renders the new template" do
       get "/admin/schools/#{school.id}/induction-coordinators/new"
 
-      expect(response.body).to include("Induction tutor for #{school.name}")
+      expect(response.body).to include(CGI.escapeHTML("Induction tutor for #{school.name}"))
```
ethanmills, line 17: unrelated change to fix flakey test

kicferk1 (approved):

PR #392, opened 2021-05-19
Sending partnership challenged email notifications

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-281

### Context

We'd like to notify lead provider when school challenges reported by them partnership.

I have added another quite complicated matcher but, as previously, I strongly believe that it will increase our confidence in unit tests and lowers need for higher-level testing. Also, as the complexity is encapsulated within the matcher, it will allow keep our tests as simple as possible. Would be great if we could improve that matcher and use it everywhere where we can :)

[PENDING]


spec/support/delayed_job.rb
```diff
@@ -87,6 +82,55 @@ def query_jobs(object, method_name)
     end
   end
 
+  # Usage:
+  #    expect(Mailer).to delay_email_delivery_of(:some_email).with(arguments / argument_matchers)
```
kicferk1, line 86: So you are telling me that all of this could be avoided if we used `deliver_now` instead of `deliver_later`? 😂

spec/forms/challenge_partnership_form_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      expect { subject.challenge! }.to change { partnership.reload.challenged? }.to true
+    end
+
+    it "schedules partnership challenged emails" do
+      subject.challenge!
+
+      lead_provider_profiles.each do |lp_profile|
+        expect(LeadProviderMailer).to delay_email_delivery_of(:partnership_challenged_email)
+          .with(user: lp_profile.user, partnership: partnership)
+      end
```
kicferk1, line 29: We could also stub the mailer class like we do in sessions spec and check the email method is being called with the correct arguments. 

I guess what this does extra is it checks that we didn't forget to call `deliver` on the email, but maybe we could also check that by using `TestMailer`?

You know my opinion on complicated matchers that stem from delayed job :P

BroiSatse: There is one extra check that we do here - we actually check that the email is scheduled for delivery. Any stubbing/mocking can only check that the method `deliver` was called - this is not enough to be certain that the mail is scheduled. If the call is executed within a database transaction, then the job will not be committed and we have false positive. This matcher checks that the appropriate job has been committed to the database.

kicferk1: Right, and what is the non-engineered use case when that false positive happens? A dev adds some code after scheduling the email, and that code breaks the transaction, so the email is not scheduled but we think in this test that it is?

spec/support/delayed_job.rb
BroiSatse: Well, I am pretty sure 100% of this codebase could have been avoided by employing a call center with operators updating a giant google spreadsheet. :P But we descided to use background jobs, so we need reliable way of testing scheduled jobs.

spec/forms/challenge_partnership_form_spec.rb
BroiSatse: Commons scenario is that we schedule emails together with some database changes. Example would be partnership creation - we save a number of partnership in the database and schedule an email for each one of them. It was requested that the process is all or none - so we have to wrap the entire loop in the transaction. There was no way of testing that no emails will be sent without going directly to the databse.

What's also worth noting is that this code is much simpler than what ActiveJob is doing behind the curtain. This code is highly afferent and as such is "write it once, improve twice and never see it again".

kicferk1: What the new matcher is giving us is the ability to check that an email was delayed successfully with particular parameters, and it will not succeed when transaction breaks. 

But really, should we not have a separate test for the transaction breaking (and then we can check there are no new delayed jobs), and another test for the happy path, in which we can check that the mailer got called with the correct parameters? And even check that exactly one new job got scheduled?

I know you know it, but I am weary of delving into the depths of delayed job and how it stores things in the database. And to your point that this code is essentially forgettable - we hoped so about `delay_execution_of`, but it is actually [touched](https://github.com/DFE-Digital/early-careers-framework/pull/311/files#diff-dc4a9b45ec5529e4af2ff7db9553384327f1079f3765e96eaebd21a8f51d55b4R61-R67), cause we couldn't predict everything :/

BroiSatse: `But really, should we not have a separate test for the transaction breaking (and then we can check there are no new delayed jobs), and another test for the happy path, in which we can check that the mailer got called with the correct parameters? And even check that exactly one new job got scheduled?
`

That's exactly the point of this. If we use ActiveJob test queue adapter, no job would be ever written to DJ, regardless of the success. Naturally those are the new matchers and they have not been written to satisfy every single scenario yet - that's where "improve twice" comes from. It will eventually reach its stable form after which no further changes are needed.

But yes, absolutely agree it would be best implemented within the ActiveJob itself. I've really given up on suggesting new features/improvements to rails team and being shot down with "we don't need it in bootscamp" argument...

app/forms/challenge_partnership_form.rb
```diff
@@ -3,9 +3,11 @@
 class ChallengePartnershipForm
   include ActiveModel::Model
 
+  # TODO: Does this form really needs anything apart partnership and challenge reason?
```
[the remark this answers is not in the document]
BroiSatse: Yup, that's a bit interesting. So yes, token is used - by controller. This form has no point in knowing the token - it doesn't validate it nor doesn't use it in its only method.

The token itself is extremely interesting concept - it serves as an authorisation without authentication which I am not 100% how I feel about. I am 100% behind including authorisation within the form though (but would also strongly push on changing the name from form into DomainService or similar).

app/forms/challenge_partnership_form.rb
```diff
@@ -3,9 +3,11 @@
...
   attr_accessor :challenge_reason, :token, :school_name, :provider_name, :partnership
   validates :challenge_reason, presence: { message: "Select a reason why you think this confirmation is incorrect" }
 
+  # TODO: Should this be static?
```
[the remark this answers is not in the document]
BroiSatse: Actually, a constant would do.

spec/support/delayed_job.rb
```diff
@@ -87,6 +82,55 @@ def query_jobs(object, method_name)
...
+      return email_arguments.any? unless @arguments
+
+      email_arguments.any? do |args|
+        @arguments.args_match?(*args)
+      end
+    end
+
+    chain :with do |*args|
+      @arguments = RSpec::Mocks::ArgumentListMatcher.new(*args)
+    end
```
ethanmills, line 115: Is it worth adding an :at matcher here?

ethanmills (approved):

spec/support/delayed_job.rb
BroiSatse: I think that would be the point to apply YAGNI. There is no place that chain would be used now and I don't think any requirements for that are coming. :) Actually, I don't think I have ever had to schedule email delivery on its own.

PR #394, opened 2021-05-20
Use smart quotes consistently

fofr:
There are many uses of smart quotes already, but some inconsistencies in the privacy policy, accessibility statement and the cookie banner.


tonyheadford (approved):
LGTM!

jamesrose (approved):

PR #397, opened 2021-05-20
Email redirector

BroiSatse:
## Purely technical spike.

Currently, in all non-production environments we are sending emails using limited notify API keys that only allows emails ending to a small subset of email addresses. This is absolutely great as we can run the systems without worrying about sending live emails. For testing purposes, we have hardcoded `cpd-test+<alias>` email addresses in the seed data and added those addresses to our team in Notify. There are still a lot of emails that we are not able to see - for example, when testing partnership reporting by lead provider emails are sent to schools email addresses coming from SchoolDataImport which we have no visibility of in PR apps without some heavy DB setup (those emails will silently fail in notify).

This PR introduces an optional redirection for ALL emails sent in non-production environments. Once the environment variable SEND_EMAILS_TO is set, all emails will be sent to given email address instead of the original receiver, which should make QA testing of email journeys much simpler.

Additionally, each email will hold additional tags next to the subject specifying what environment it comes from and to whom it would be sent. However, due to notify limitations (it is not possible to send blank personalisation), this requires specifying subject within a mailer rather than on the template in notify and putting `((subject))` as a subject in the template. I have also tested that all current emails can still be sent (just they will be missing tags in non-production environment)

### Pros:

- All emails are visible during QA with no database and notify group tweaking. For all current and future emails.
- We can redirect emails to our own email address for a local environment - less noise for the team.
- Prevents `Notifications::Client::BadRequestError (BadRequestError: Can’t send to this recipient using a team-only API key)` errors in non-production environments.
- Allows us to turn `raise_notification_errors` in development, which in turn will make it more visible in the development when we are missing 

### Cons:

- Putting subject in the codebase rather than on the template.

## Testing the functionality

1. ssh to the PR box `cf ssh ecf-review-pr-397`
1. `cd ../app && /usr/local/bin/bundle exec rails c`
1. *(optional)* change `ENV['SEND_EMAILS_TO']` to your own `digital.education.gov.uk` address
1. Send a test email: `UserMailer.test_email(user: User.order("RANDOM()").first).deliver_now`
1. Check your box (or cpd-test box if you skipped step 3) - you should see an email with additional tags in the subject

Optionally, try sending any other email - those have not subject personalisation and as such will not have tags when sent, but will still be redirected to given email address.

## Issues to follow

It seems that gmail is still grouping the emails together, despite them having different subjects - causing subject tags to be unreadable/unusable in gmail. This would require more investigating, however good filter should resolve this.


kicferk1 (approved):
So do we need to change the templates in notify for this to work? Do we merge this change, promote it across all the environments, and then change notify templates to use the `subject` variable? Will that let us use the existing templates?

Also, do we need to let anyone know what the default subjects for those templates are?

BroiSatse:
@kicferk1 - There are couple of steps. All the emails will continue to work when this is merged. The only downside is lack of tags in non-prod environments. Once this approach is accepted, I'll add another commit adding subject personalisation to all of the existing emails (extra persoanlisations are silently ignored). Once this is deployed we can safely change subjects on templates.

lib/email_redirector.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module EmailRedirector
+  class << self
+    def delivering_email(mail)
+      return unless enabled?
+
+      overide_personalisation(mail, :subject) { |subject| "#{tags(mail)} #{subject}" }
+      mail.to = target_email if enabled?
```
ethanmills, line 9: redundant check?

terraform/workspace-variables/staging_app_env.yml
```diff
@@ -10,3 +10,4 @@ DATABASE_URL: postgres://postgres@localhost:5432
 DOMAIN: s-manage-training-for-early-career-teachers.education.gov.uk
 GOVUK_WEBSITE_ROOT: s-manage-training-for-early-career-teachers.education.gov.uk
 GOVUK_APP_DOMAIN: s-manage-training-for-early-career-teachers.education.gov.uk
+SEND_EMAILS_TO: cpd-test@digital.education.gov.uk
```
ethanmills, line 13: I'm not sure we want this in staging? We should check what the wider team want here.

Also, are we sure we want to introduce another difference between staging and production?

BroiSatse: That is a good question. I'd guess it all depends on what we use staging for. I'd imagine we'd be using it for stakeholders demo (instead of dev or PR app as we do now)?

BroiSatse: Right, so it seem that staging is supposed to be used for QA and PO acceptance purpose. If this is the case than we definitely want to redirect emails in staging. I don't think it introduces another difference between staging and production - we already use different Notify API key to prevent external emails from being sent so the behaviour is already different.

We might however want to send emails to different address than cpd-test, maybe cpd-staging?

PR #399, opened 2021-05-20
Change induction tutor

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-346)
Enable changing and replacing of induction tutors (coordinators)

Add admin journey to `Change` the induction tutor for a school
Change link on induction tutor leads to an interstitial page which asks you to choose between replacing the existing tutor or updating the existing tutors details
Edit/Update functionality for induction tutor
Replacing creates new user and deletes the existing one.

### Testing
As an admin user, select a school with an induction tutor (or add an induction tutor)
When viewing that school, click the `Change` link next to the induction tutor
Make a selection to replace or update
Replace will work in the same way as adding an induction tutor and email will be sent - the existing tutor will be deleted
Update enables changing the full name and/or email - no emails are sent in this case.

### Review Checks

### DIFF
--- a/app/controllers/admin/schools/induction_coordinators_controller.rb
+++ b/app/controllers/admin/schools/induction_coordinators_controller.rb
@@ -6,76 +6,45 @@ module Admin
     skip_after_action :verify_policy_scoped
     before_action :set_school
 
-    def email_used; end
-
     def new
-      @nominate_induction_tutor_form = NominateInductionTutorForm.new
+      @induction_tutor_form = NominateInductionTutorForm.new
     end
 
     def create
-      @nominate_induction_tutor_form = NominateInductionTutorForm.new(
-        tutor_form_params.merge(school_id: params[:school_id]),
-      )
+      @induction_tutor_form = NominateInductionTutorForm.new(tutor_form_params)
 
-      if @nominate_induction_tutor_form.valid?
-        if email_already_used?(@nominate_induction_tutor_form.email)
-          render_email_used_page(email: @nominate_induction_tutor_form.email,
-                                 action_path: new_admin_school_induction_coordinator_path(@school))
-        else
-          create_school_induction_tutor!(school: @school,
-                                         email: @nominate_induction_tutor_form.email,
-                                         full_name: @nominate_induction_tutor_form.full_name)
-          set_success_message(content: "New induction tutor added. They will get an email with next steps.", title: "Success")
-          redirect_to admin_school_path(@school)
-        end
+      if @induction_tutor_form.valid?
+        CreateInductionTutor.call(school: @school,
+                                  email: @induction_tutor_form.email,
+                                  full_name: @induction_tutor_form.full_name)
+
+        set_success_message(content: "New induction tutor added. They will get an email with next steps.", title: "Success")
+        redirect_to admin_school_path(@school)
+      elsif @induction_tutor_form.email_already_taken?
+        render_email_used_page(email: @induction_tutor_form.email,
+                               action_path: new_admin_school_induction_coordinator_path(@school))
       else
         render :new
       end
     end
 
-    def choose_replace_or_update
-      @replace_or_update_tutor_form = ReplaceOrUpdateTutorForm.new
-    end
-
-    def replace_or_update
-      @replace_or_update_tutor_form = ReplaceOrUpdateTutorForm.new(replace_or_update_params)
-      if @replace_or_update_tutor_form.valid?
-        if @replace_or_update_tutor_form.replace_tutor?
-          redirect_to new_admin_school_induction_coordinator_path(@school)
-        else
-          redirect_to edit_admin_school_induction_coordinator_path(@school, school_induction_tutor)
-        end
-      else
-        render "choose_replace_or_update"
-      end
-    end
-
     def edit
-      @induction_tutor = school_induction_tutor
+      @induction_tutor_form = NominateInductionTutorForm.new(school_induction_tutor_attributes)
     end
 
     def update
-      @induction_tutor = school_induction_tutor
-      @induction_tutor.assign_attributes(tutor_form_params)
-      if @induction_tutor.changed?
-        if @induction_tutor.email_changed?
-          user = User.find_by(email: @induction_tutor.email)
-          if user && user.id != @induction_tutor.id
-            render_email_used_page(email: @induction_tutor.email,
-                                   action_path: edit_admin_school_induction_coordinator_path(@school))
-            return
-          end
-        end
+      @induction_tutor_form = NominateInductionTutorForm.new(school_induction_tutor_attributes.merge(tutor_form_params))
 
-        if @induction_tutor.valid?
-          @induction_tutor.save!
-          set_success_message(content: "Induction tutor details updated", title: "Success")
-          redirect_to admin_school_path(@school)
-        else
-          render :edit
-        end
-      else
+      if @induction_tutor_form.valid?
+        school_induction_tutor.update!(full_name: @induction_tutor_form.full_name,
+                                       email: @induction_tutor_form.email)
+        set_success_message(content: "Induction tutor details updated", title: "Success")
         redirect_to admin_school_path(@school)
+      elsif @induction_tutor_form.email_already_taken?
+        render_email_used_page(email: @induction_tutor_form.email,
+                               action_path: edit_admin_school_induction_coordinator_path(@school, school_induction_tutor))
+      else
+        render :edit
       end
     end
 
@@ -89,23 +58,21 @@ module Admin
       @school.induction_coordinators.first
     end
 
-    def create_school_induction_tutor!(school:, email:, full_name:)
-      school.induction_coordinators.first.destroy! if school.induction_coordinators.first
-
-      InductionCoordinatorProfile.create_induction_coordinator(
-        full_name,
-        email,
-        school,
-        Rails.application.routes.url_helpers.root_url(host: Rails.application.config.domain),
-      )
+    def school_induction_tutor_attributes
+      user = @school.induction_coordinators.first
+      if user
+        {
+          user_id: user.id,
+          email: user.email,
+          full_name: user.full_name,
+        }
+      else
+        {}
+      end
     end
 
     def school_using_this_email(email)
       User.find_by(email: email).schools.first
-    end
-
-    def email_already_used?(email)
-      User.exists?(email: email)
     end
 
     def render_email_used_page(email:, action_path:)
@@ -117,10 +84,6 @@ module Admin
 
     def tutor_form_params
       params.require(:tutor_details).permit(:full_name, :email)
-    end
-
-    def replace_or_update_params
-      params.require(:replace_or_update_tutor_form).permit(:choice)
     end
   end
 end
--- a/app/controllers/admin/schools/induction_coordinators_controller.rb
+++ b/app/controllers/admin/schools/induction_coordinators_controller.rb
@@ -25,6 +25,23 @@ module Admin
       redirect_to email_used_admin_school_induction_coordinators_path
     end
 
+    def choose_replace_or_update
+      @replace_or_update_tutor_form = ReplaceOrUpdateTutorForm.new
+    end
+
+    def replace_or_update
+      @replace_or_update_tutor_form = ReplaceOrUpdateTutorForm.new(replace_or_update_params)
+      if @replace_or_update_tutor_form.valid?
+        if @replace_or_update_tutor_form.replace_tutor?
+          redirect_to new_admin_school_induction_coordinator_path(@school)
+        else
+          redirect_to edit_admin_school_induction_coordinator_path(id: @school.id)
+        end
+      else
+        render "choose_replace_or_update"
+      end
+    end
+
     def edit
       @induction_tutor = school_induction_tutor
     end
@@ -51,6 +68,10 @@ module Admin
 
     def form_params
       params.require(:tutor_details).permit(:full_name, :email)
+    end
+
+    def replace_or_update_params
+      params.require(:replace_or_update_tutor_form).permit(:choice)
     end
   end
 end
--- a/app/controllers/admin/schools/induction_coordinators_controller.rb
+++ b/app/controllers/admin/schools/induction_coordinators_controller.rb
@@ -25,14 +25,32 @@ module Admin
       redirect_to email_used_admin_school_induction_coordinators_path
     end
 
+    def edit
+      @induction_tutor = school_induction_tutor
+    end
+
+    def update
+      @induction_tutor = school_induction_tutor
+      if @induction_tutor.update(form_params)
+        set_success_message(content: "Induction tutor details updated", title: "Success")
+        redirect_to admin_school_path(@school)
+      else
+        render :edit
+      end
+    end
+
   private
 
     def set_school
       @school = School.find params[:school_id]
     end
 
+    def school_induction_tutor
+      @school.induction_coordinators.first
+    end
+
     def form_params
-      params.require(:nominate_induction_tutor_form).permit(:full_name, :email)
+      params.require(:tutor_details).permit(:full_name, :email)
     end
   end
 end
--- a/app/controllers/admin/schools/induction_coordinators_controller.rb
+++ b/app/controllers/admin/schools/induction_coordinators_controller.rb
@@ -36,13 +36,13 @@ module Admin
       @induction_tutor_form = NominateInductionTutorForm.new(school_induction_tutor_attributes.merge(tutor_form_params))
 
       if @induction_tutor_form.valid?
-        school_induction_tutor.update!(full_name: @induction_tutor_form.full_name,
+        @school.induction_tutor.update!(full_name: @induction_tutor_form.full_name,
                                        email: @induction_tutor_form.email)
         set_success_message(content: "Induction tutor details updated", title: "Success")
         redirect_to admin_school_path(@school)
       elsif @induction_tutor_form.email_already_taken?
         render_email_used_page(email: @induction_tutor_form.email,
-                               action_path: edit_admin_school_induction_coordinator_path(@school, school_induction_tutor))
+                               action_path: edit_admin_school_induction_coordinator_path(@school, @school.induction_tutor))
       else
         render :edit
       end
@@ -54,12 +54,8 @@ module Admin
       @school = School.find params[:school_id]
     end
 
-    def school_induction_tutor
-      @school.induction_coordinators.first
-    end
-
     def school_induction_tutor_attributes
-      user = @school.induction_coordinators.first
+      user = @school.induction_tutor
       if user
         {
           user_id: user.id,
--- a/app/controllers/admin/schools/induction_coordinators_controller.rb
+++ b/app/controllers/admin/schools/induction_coordinators_controller.rb
@@ -16,14 +16,50 @@ module Admin
       @nominate_induction_tutor_form = NominateInductionTutorForm.new(
         form_params.merge(school_id: params[:school_id]),
       )
-      render :new and return unless @nominate_induction_tutor_form.valid?
+      # render :new and return unless @nominate_induction_tutor_form.valid?
 
-      @nominate_induction_tutor_form.save!
-      set_success_message(content: "New induction tutor added. They will get an email with next steps.", title: "Success")
-      redirect_to admin_school_path(@school)
-    rescue UserExistsError
-      redirect_to email_used_admin_school_induction_coordinators_path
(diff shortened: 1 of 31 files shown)


app/controllers/admin/schools/induction_coordinators_controller.rb
```diff
@@ -16,13 +16,72 @@ def create
...
+                                         full_name: @nominate_induction_tutor_form.full_name)
+          set_success_message(content: "New induction tutor added. They will get an email with next steps.", title: "Success")
+          redirect_to admin_school_path(@school)
+        end
+      else
+        render :new
+      end
+    end
+
+    def create_school_induction_tutor!(school:, email:, full_name:)
```
kicferk1, line 37: Let's make as much stuff as possible `private`. I like the public methods in the controller to be just the actions - cause I don't think anything else needs to be.

I also don't like to have too much logic in the controller, and we happen to have a form object wandering around - maybe using that for some checks (like if the email is being used) would reduce the cognitive load on the controller?

app/views/admin/schools/induction_coordinators/choose_replace_or_update.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @replace_or_update_tutor_form, url: replace_or_update_admin_school_induction_coordinators_path(@school), method: :post do |f| %>
+      <%= f.govuk_error_summary %>
+      <h1 class="govuk-heading-xl"><%= title %></h1>
+      <%= f.govuk_collection_radio_buttons(
+            :choice,
+            @replace_or_update_tutor_form.choices,
+            :id,
+            :name,
+            legend: { text: "" },
```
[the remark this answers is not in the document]
tonyheadford: Nice one!

app/views/admin/schools/induction_coordinators/edit.html.erb
```diff
@@ -0,0 +1,14 @@
+<% title = "Update induction tutor for #{@school.name}" %>
+<% content_for :title, title %>
+<% content_for :before_content, govuk_back_link(text: "Back", href: admin_school_path(@school)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= title %></h1>
+    <%= render partial: 'induction_tutor_form', locals: {
```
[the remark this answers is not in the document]
tonyheadford: Good catch, same for `:new` template

app/controllers/admin/schools/induction_coordinators_controller.rb
tonyheadford: Another good catch, agree about the `private` stuff, I had meant to refactor that before submitting the PR.  Kind of agree about the controllers being lean but I wasn't too keen on the form object handling the save and raising an exception, I prefer to treat them like a simple view model. Additionally we need to pass additional info concerning where the email address is being used in the "error" page so needed to restructure it.

kicferk1 (commented):
I really think we should try our best to move majority of validation and saving logic out of the controller. `NominateInductionTutorForm` has a `save!` method which is now unused (I think) 😨

And the controller refers to `@nominate_induction_tutor_form.email` and `@nominate_induction_tutor_form.full_name` a lot in those validation / creation moments, which suggests to me that that part of logic wants to belong closer to where the data actually lives. 

I quite like treating our forms as a stand-in for model objects, in situations when data input doesn't quite match individual models. If the controller was operating on a single model, you wouldn't make a `save_my_model!` method on a controller, you would put it on the model. And I think it's nice to pretend forms are like models in that regard. Maybe `form` is not the best name for it, `domain object` sounds more like what they are, but it doesn't roll off the tongue 😂

app/controllers/admin/schools/induction_coordinators_controller.rb
```diff
@@ -16,13 +16,67 @@ def create
...
+      @replace_or_update_tutor_form = ReplaceOrUpdateTutorForm.new
+    end
+
+    def replace_or_update
+      @replace_or_update_tutor_form = ReplaceOrUpdateTutorForm.new(replace_or_update_params)
+      if @replace_or_update_tutor_form.valid?
+        if @replace_or_update_tutor_form.replace_tutor?
+          redirect_to new_admin_school_induction_coordinator_path(@school)
+        else
+          redirect_to edit_admin_school_induction_coordinator_path(id: @school.id)
```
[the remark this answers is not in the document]
tonyheadford: Typically I would probably look to push that logic into a service object, perhaps that's what you mean?.  I think "form" objects should just be means to pass info to and from forms, handling attribute validation is ok but I don't think they should be responsible for updating the database or controlling flow by raising errors? Blurred lines.  I could see that form object was used somewhere else so didn't want to change it.
I _think_ that path needed the `:id` to be explicit for some reason, will check...

app/controllers/admin/schools/induction_coordinators_controller.rb
```diff
@@ -16,13 +16,67 @@ def create
...
+        if @induction_tutor.valid?
+          @induction_tutor.save!
+          set_success_message(content: "Induction tutor details updated", title: "Success")
+          redirect_to admin_school_path(@school)
+        else
+          render :edit
+        end
+      else
+        redirect_to admin_school_path(@school)
+      end
```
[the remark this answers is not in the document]
tonyheadford: Not convinced that should be in a form object, if anything it should be removed from there if not used anywhere else.
A better thing might be to refactor the controller

spec/requests/admin/schools/induction_coordinators_spec.rb
```diff
@@ -34,6 +35,89 @@
...
+  end
+
+  describe "GET /admin/schools/:school_id/induction-coordinators/choose-replace-or-update" do
+    it "renders the choose replace or update template" do
+      get "/admin/schools/#{school.id}/induction-coordinators/choose-replace-or-update"
+
+      expect(response.body).to include("Are you replacing an induction tutor or updating their details?")
+      expect(response.body).to include("Replace induction tutor with someone new")
+      expect(response.body).to include("Update induction tutor’s details")
+      expect(response).to render_template("admin/schools/induction_coordinators/choose_replace_or_update")
```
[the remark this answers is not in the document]
tonyheadford: Radio button options do come from a component though

app/controllers/admin/schools/induction_coordinators_controller.rb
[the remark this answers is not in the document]
tonyheadford: Doh I'm an idiot - that path 😬

spec/cypress/support/step_definitions/common-interaction.js
```diff
@@ -20,6 +20,8 @@ const buttons = {
   "create supplier user button": '.govuk-button:contains("Add a new user")',
   "search button": "[data-test=search-button]",
   "remove button": ".govuk-button[value=Remove]",
+  "replace induction tutor": "input[value=replace].govuk-radios__input",
+  "update induction tutor": "input[value=update].govuk-radios__input",
```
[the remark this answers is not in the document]
tonyheadford: radio buttons ;) but moved 👍

spec/requests/admin/schools/induction_coordinators_spec.rb
```diff
@@ -14,15 +15,15 @@
     it "renders the new template" do
       get "/admin/schools/#{school.id}/induction-coordinators/new"
 
-      expect(response.body).to include(CGI.escapeHTML("Induction tutor for #{school.name}"))
+      expect(response.body).to include("New induction tutor for #{school.name}")
```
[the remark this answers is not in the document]
tonyheadford: Let's leave it out if we don't know why we're using it - can aways add it in again if we get flaky tests, I didn't think Faker would just randomly use unsafe chars in strings?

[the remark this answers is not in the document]
kicferk1: @ethanmills knows for sure, I just know that every time I see a 'fix flaky test' that deals with response body it uses this magic code to do it 😂 

Happy to leave it and wait for it to break.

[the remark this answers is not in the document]
tonyheadford: :joy: if it _has_ been put there for good reason then I'm happy for it to stay - sounds _really_ fishy though 😕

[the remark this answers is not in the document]
tonyheadford: Let's leave it out and see if we can get to the bottom of it if it breaks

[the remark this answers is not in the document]
tonyheadford: A-ha it failed (in a different place) - I think it's primarily because we're mis-using request specs to check page content

app/controllers/admin/schools/induction_coordinators_controller.rb
ethanmills: The approach we've been taking here is to lift the majority of the logic in to the `form`s, which makes things easier to test. @BroiSatse is likely to have view on this. 

I personally _much_ prefer the logic in the form to in the controller

ethanmills:
If you really don't like having the logic in the form then I'm open to suggestions as to where else it should go. But I think moving it into the controller is a big step in the wrong direction.

spec/requests/admin/schools/induction_coordinators_spec.rb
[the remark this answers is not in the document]
ethanmills: Please leave this in. It's here for a reason. Whenever generated content contains an apostrophe, tests will fail without it.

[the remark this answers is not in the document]
tonyheadford: I've removed that line, we shouldn't be testing content in request specs.

tonyheadford:
> If you really don't like having the logic in the form then I'm open to suggestions as to where else it should go. But I think moving it into the controller is a big step in the wrong direction.

I don't disagree about the controller, I've been moving some of this functionality into a service object to keep the controller leaner as suggested. I _really_ think having the current logic in the "form" object is a bad pattern - a form object should not be raising errors to control flow (that should be in the controller) or doing only doing some validation when you call `valid?` but other validation (that raises) when you call `save!` 😬

kicferk1:
> I _really_ think having the current logic in the "form" object is a bad pattern - a form object should not be raising errors to control flow (that should be in the controller) or only doing some validation when you call `valid?` but other validation (that raises) when you call `save!` 😬

We can rename the form object to something better if it will make you feel better. 

It feels very natural to me. We have some html form, that passes data to controller via params, that data then needs to be validated and saved in the database. If we were creating a user with no profiles (so, if the html form only sent data corresponding to a single Rails model) we could do something like
```
def create
  @user = User.new(user_params)
  if @user.save
    redirect_to somewhere
  else
    render :new
  end
end
```
When the html form is sending data that doesn't match a single db model, I really like the idea of writing
```
def create
  @user_form = UserForm.new(user_params)
  if @user_form.save
    redirect_to somewhere
  else
    render :new
  end
end
```
So the controller doesn't need to know, or bother about what the data really means for the database - all of that is encapsulated in the form. Think about it this way - if we had a single `InductionCoordinator` model in the database, you would move all the saving logic into the `InductionCoordinator` model, so why not let the form object abstract away database structure for the controller?

tonyheadford:
> > I _really_ think having the current logic in the "form" object is a bad pattern - a form object should not be raising errors to control flow (that should be in the controller) or only doing some validation when you call `valid?` but other validation (that raises) when you call `save!` 😬
> 
> We can rename the form object to something better if it will make you feel better.
> 
> It feels very natural to me. We have some html form, that passes data to controller via params, that data then needs to be validated and saved in the database. If we were creating a user with no profiles (so, if the html form only sent data corresponding to a single Rails model) we could do something like
> 
> ```
> def create
>   @user = User.new(user_params)
>   if @user.save
>     redirect_to somewhere
>   else
>     render :new
>   end
> end
> ```
> 
> When the html form is sending data that doesn't match a single db model, I really like the idea of writing
> 
> ```
> def create
>   @user_form = UserForm.new(user_params)
>   if @user_form.save
>     redirect_to somewhere
>   else
>     render :new
>   end
> end
> ```
> 
> So the controller doesn't need to know, or bother about what the data really means for the database - all of that is encapsulated in the form. Think about it this way - if we had a single `InductionCoordinator` model in the database, you would move all the saving logic into the `InductionCoordinator` model, so why not let the form object abstract away database structure for the controller?

I've reworked things, will push up my changes shortly

app/controllers/admin/schools/induction_coordinators_controller.rb
```diff
@@ -31,8 +54,36 @@ def set_school
...
     end
 
-    def form_params
-      params.require(:nominate_induction_tutor_form).permit(:full_name, :email)
+    def school_induction_tutor
+      @school.induction_coordinators.first
+    end
+
+    def school_induction_tutor_attributes
+      user = @school.induction_coordinators.first
```
kicferk1, line 62: `school_induction_tutor`?

app/controllers/admin/schools/induction_coordinators_controller.rb
```diff
@@ -31,8 +54,36 @@ def set_school
       @school = School.find params[:school_id]
     end
 
-    def form_params
-      params.require(:nominate_induction_tutor_form).permit(:full_name, :email)
+    def school_induction_tutor
+      @school.induction_coordinators.first
+    end
```
kicferk1, line 59: I wonder if this is something we would like on `school` model - so that we don't duplicate a logic of 'who is the induction tutor for a school'. And so that eventually it's easier to change it, if we choose to.

app/controllers/nominations/nominate_induction_coordinator_controller.rb
```diff
@@ -26,12 +26,15 @@ def new
...
-    render :new and return unless @nominate_induction_tutor_form.valid?
-
-    @nominate_induction_tutor_form.save!
-    redirect_to nominate_school_lead_success_nominate_induction_coordinator_path
-  rescue UserExistsError
-    redirect_to email_used_nominate_induction_coordinator_path
+    if @nominate_induction_tutor_form.valid?
+      CreateInductionTutor.call(school: @nominate_induction_tutor_form.school,
+                                email: @nominate_induction_tutor_form.email,
+                                full_name: @nominate_induction_tutor_form.full_name)
```
kicferk1, line 32: Eh, in my personal opinion passing all the data of one object to another object is a good indication that the functionality really wants to live on the first object.

And the `call` includes `save!` which could in principle throw an exception - so the form will need to be tightly logically coupled with the service - if we change what we do within the `call` so it throws with more cases, we will need to change the form validation too.

I guess all form-service separation has to deal with that problem, and maybe the cost is worth it. I'm not convinced, but I won't make you rewrite it :P

app/services/create_induction_tutor.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    ActiveRecord::Base.transaction do
+      school.induction_coordinators.first.destroy! if school.induction_coordinators.first
+
+      user = User.create!(full_name: full_name, email: email)
+      ic_profile = InductionCoordinatorProfile.create!(user: user, schools: [school])
+      SchoolMailer.nomination_confirmation_email(user: user, school: school, start_url: start_url).deliver_now
+    end
+
+    self.profile = ic_profile
```
kicferk1, line 23: Why do we not want to set the profile inside the transaction?

app/services/create_induction_tutor.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def start_url
+    @start_url ||= Rails.application.routes.url_helpers.root_url(host: Rails.application.config.domain)
```
kicferk1, line 27: Do we get anything out of caching it here?

app/controllers/admin/schools/induction_coordinators_controller.rb
tonyheadford: I can change it but isn't that how they are referred to in the UI?

app/controllers/admin/schools/induction_coordinators_controller.rb
tonyheadford: Yes, I thought the same but didn't want to do more changes! I'll make that change though as it will help

app/controllers/nominations/nominate_induction_coordinator_controller.rb
tonyheadford: Hmm I don't think they are tightly coupled, what we are doing is trying to separate the concerns so that the form object is concerned with validating the user input and the service object is concerned with all the all the tasks related to creating an induction tutor.  The service object can be used completely independently from the form object

app/controllers/admin/schools/induction_coordinators_controller.rb
kicferk1: Oh, I meant that we could reuse the method above to get the user here

app/controllers/nominations/nominate_induction_coordinator_controller.rb
kicferk1: It can be used completely independently, but it will throw an error whenever it gets data that it doesn't expect. And the form has to make sure that the data is what the service expects.

It might well be that that is the reason why Rails chose to do validation and saving in the same place, on a model :P

app/services/create_induction_tutor.rb
tonyheadford: Probably could do that, I think I was thinking if the transaction rolled back it would leave the `profile` with and invalid value if it was assigning to `self` inside the block... unlikely to happen perhaps as it would only be if the mailer failed?
BTW, do we not have background workers? - was wondering why the mailer was using `deliver_now` instead of `deliver_later`?

app/services/create_induction_tutor.rb
tonyheadford: I had jumped off into trying to make it easy to change in tests, but probably unecessary 👍

app/controllers/nominations/nominate_induction_coordinator_controller.rb
tonyheadford: Where's the trollface emoji?

app/services/create_induction_tutor.rb
kicferk1: I mean, unlikely failures are precisely the reason we do stuff in transactions.

We have a worker, although we added it relatively recently. You can see the first ever usage of `deliver_later` on https://github.com/DFE-Digital/early-careers-framework/pull/392. 

I am pretty sure we are using `deliver_now` because a developer who since left the project found `deliver_later` annoying, with it increasing the time you would need to wait for your email if the worker was busy with something heavy at the time. Or at least I think that was the worry when we discussed using it half a year ago for sending sign-in emails.

tonyheadford: I'll make it `deliver_later` if there are issues with workers then that's an infrastructure thing.

spec/services/create_induction_tutor_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    it "emails the new induction tutor" do
+      allow(SchoolMailer).to receive(:nomination_confirmation_email).and_call_original
+
+      service = CreateInductionTutor.new(school: school, email: email, full_name: name)
+      service.call
+
+      expect(SchoolMailer).to have_received(:nomination_confirmation_email)
+        .with(user: service.profile.user, school: school, start_url: service.start_url)
+    end
```
kicferk1, line 26: Rather than using `service.profile.user` (is that the only reason the service has a `profile` field?), I would be tempted to add `new_user = User.find_by(email: email)` after the `call` and before the `expect`, and check that that `user` is what the mailer received. 

Emails are unique per user, and it will mean we don't need to store the profile we just created on the service instance.

app/services/create_induction_tutor.rb
kicferk1: Oh, you are gonna love this. I think making it `deliver_later` means that it's not immediately available in the cypress tests (they use test mailer to figure out which emails were sent). Argh. Can you make it `deliver_now` and we can look into changing it later?

tonyheadford: Yes thanks, had a chat with @BroiSatse and we'll need to look at this later.

spec/services/create_induction_tutor_spec.rb
tonyheadford: Yes could do

kicferk1 (approved):
Eh, not particularly keen on extracting a service out of the form, but the code looks well enough and has good amount of tests.

PR #400, opened 2021-05-20
Add user_type and core_induction_programme to users api

kicferk1:
### Context
E&L will need to get some more user data than just their name and email. In particular we want to know if the user is an ECT or mentor (or someone else), and which CIP they are doing. 

1. Add two enum fields to user serializer.
2. Update docs.
3. Add some users with CIPs to dummy structures.
4. Update doc generation so that schema actually displays correctly in swagger - that took me way too long to figure out, given it's a one-liner :P 

### Guidance to review
I was thinking about sending over all of user profiles, but that seemed a bit excessive.

I haven't added types for any users other than ECTs, because E&L only cares about ECTs and mentors (which don't exist in R&P yet). I guess if we ever want to return more than these two, we might need to worry about the use case of 'what if the user is both an induction coordinator and a mentor', in which case, eh, a list of types? Or maybe we switch to sending full blown profiles then. Not sure. 

### Testing
Updated unit tests and docs.

Get an api token, hit the user API, see the response that includes the new fields.


app/serializers/user_serializer.rb
```diff
@@ -1,8 +1,45 @@
 # frozen_string_literal: true
 
+USER_TYPES = {
```
jamesrose, line 3: Why are these outside of the serializer?

kicferk1: Cause I didn't think about putting them inside. Yeah, I guess when we access them it's nicer to know where they come from.

jamesrose (approved):
🚢

PR #407, opened 2021-05-24
Add feature flag service

jamesrose:
### Context

Add feature flag service from [get-help-with-technology](https://github.com/DFE-Digital/get-help-with-tech/blob/main/app/services/feature_flag.rb) who themselves stole it from [apply-for-teacher-training](https://github.com/DFE-Digital/apply-for-teacher-training/blob/master/app/services/feature_flag.rb). It's very simple:

```ruby
if FeatureFlag.active?(:some_flag)
  # do something
end
```

```ruby
if FeatureFlag.inactive?(:some_flag)
  # do something else
end
```

Working locally you can:

```bash
FEATURES_rate_limiting=active bundle exec rails s
```

On PaaS you can:

```sh
cf set-env early-careers-framework-prod FEATURES_rate_limiting active
```

And you can temporarily enable or disable a flag in your specs with the around hook `with_feature_flags`. For example:

```ruby
it 'show feature flags statuses', with_feature_flags: { rate_limiting: 'active' } do
  render
  page.load(rendered)

  expect(page.table['rate_limiting'].status).to eql('Active')
  expect(page.table['display_sign_in_token_links'].status).to eql('Not active')
end
```


kicferk1 (approved):
How are we intending to change the feature flags? With a release? By typing `cf` commands?

Also, I imagine we plan to have some difference between environments with which features are on and off by default, so we don't need to set them each time we deploy (I am not 100% sure of it, but I think the environment resets when we deploy, might be worth testing on a review app).

`terraform/workspace-variables/dev_app_env.yml` (and similarly named files for other environments) are an easy way to pass environment variables to our application, so long as they are not secret. I don't think a list of feature flags is, so it might be worth thinking about it when we get to a point of wanting to release something that's feature flagged on in most envs, but off in staging and production. I guess we could do that with an initialiser checking the env as well, but hey, that sounds way too simple :D

jamesrose:
@kicferk1 Thanks for highlighting that. I would have expected the environment variables to be long-lived, but I will confirm. Assuming they are long-lived, and just for now, let's use `cf` to toggle them. I am eager not to need anything but a no-op deploy or restage to change the state of a feature.

BroiSatse:
@jamesrose @kicferk1 We might want to introduce another flag `ENABLE_ALL_FLAGS` to use in development. I'd also think we should enable all the flags by default in test and only have some (rare, only for long-living features) test with flags disabled (to check that some links are not present when flag is disabled).

BroiSatse (approved):

app/services/feature_flag.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+class FeatureFlag
+  # Long-lived settings that are often environment-specific
+  PERMANENT_SETTINGS = %i[
+  ].freeze
```
ethanmills, line 6: Is this something that would benefit from a permanent, env specific feature flag? https://github.com/DFE-Digital/early-careers-framework/blob/84270a33adbb917bbafbd0c31307a323c5778ebf/app/controllers/users/sessions_controller.rb#L9

jamesrose: This service gives us the flexibility to define the state of each flag outside of code, per environment. In my mind, feature flags should care about the state, not the environment.

PR #408, opened 2021-05-24
Move payment calculator readme content into main readme

timabell:
### Context

This used to be its own repo so had a full readme.

This doesn't make so much sense now so I've taken the bits that are still likely to be useful and put them on the end of the main readme.

### Guidance to review

:ship:


README.md
```diff
@@ -140,3 +140,22 @@ Where `"name or id"` is a name or id from the `lead_providers` table.
 1. Get into Rails console for the environment you want to generate the token for.
 2. Run `EngageAndLearnApiToken.create_with_random_token!`
 3. Rails console should output a string, that's your unhashed token, you can keep it and use it to access E&L endpoints.
+
+## About the payment_calculator
+
+This engine performs all payment calculations for both [ECFs (Early Career Framework)](https://www.early-career-framework.education.gov.uk/) and [the reformed NPQs (National Professional Qualification)](https://www.gov.uk/government/publications/national-professional-qualifications-frameworks-from-september-2021) so that training providers can be paid the correct amount.
```
kicferk1, line 146: > This engine

What engine? Can you point to it?

timabell: Good point, that's rather ambiguous. Iterated.

kicferk1 (approved):

README.md
```diff
@@ -140,3 +140,22 @@ Where `"name or id"` is a name or id from the `lead_providers` table.
 1. Get into Rails console for the environment you want to generate the token for.
 2. Run `EngageAndLearnApiToken.create_with_random_token!`
 3. Rails console should output a string, that's your unhashed token, you can keep it and use it to access E&L endpoints.
+
+## payment_calculator
+
+The code in [`lib/payment_calculator/`](lib/payment_calculator/) performs payment calculations for both [ECFs (Early Career Framework)](https://www.early-career-framework.education.gov.uk/) and [the reformed NPQs (National Professional Qualification)](https://www.gov.uk/government/publications/national-professional-qualifications-frameworks-from-september-2021) so that training providers can be paid the correct amount.
```
ethanmills, line 146: I'm reluctant to link to the dreaded wordpress site from the readme, especially as it's going offline soon-ish

README.md
```diff
@@ -140,3 +140,22 @@ Where `"name or id"` is a name or id from the `lead_providers` table.
...
+Here are the names we are using in the code and specs for the different concepts involved in the calculations by way of an example:
+
+> Per participant price £995 >>
+per participant service fee £398 (40%) >> monthly service fee £27k >> total service fee £796k
+>
+> Per participant price £995 >> per participant output payment £597 (60%) >> per participant output payment for a retention period £119 (20% of 60%) >> output payment subtotal for a retention period with 1900 retained participants £226k
+
+* "Participants" includes both teachers and mentors.
+* "Output payments" are payments made based on the performance of the training provider (i.e. their output).
+* "Payment type" for start/retention/completion output payments.
```
ethanmills, line 161: (optional nitpick) Maybe this could move to `/documentation` with a link from the readme? People are already reluctant to read it,  I imagine it will only get worse as it gets longer

README.md
timabell: I think it adds useful context personally, how strongly do you want rid it? If you really want rid of it suggest a change and I'll accept it.

README.md
timabell: How about we merge this as-is and then you can move it to the structure you'd like to see in another PR? Or you could push more commits to this PR if you like.

README.md
ethanmills: Not strongly at all, just a suggestion. We often link out to https://www.gov.uk/government/publications/early-career-framework-reforms-overview/early-career-framework-reforms-overview FWIW

kicferk1: I think it's alright to link there, before we delete it we (E&L) will go through all the places we know of that use it and change them, and R&P copy is one of those places.

PR #411, opened 2021-05-25
Use persisted feature flags

jamesrose:
### Context

Extending the work in https://github.com/DFE-Digital/early-careers-framework/pull/407

Our current infrastructure set-up made it non-trivial to persist feature flag environment variables across deploys. They would be reset each time. Instead, we can store them in the database. This will allow us to create an interface to display, create and edit them in the future.

We might want a way to seed states in development, but we can figure that out later.

(This was copied from [apply-for-teacher-training](https://github.com/DFE-Digital/apply-for-teacher-training).)

### DIFF
--- a/lib/tasks/safe_reset.rake
+++ b/lib/tasks/safe_reset.rake
@@ -11,6 +11,7 @@ namespace :db do
                    .map { |row| row["tablename"] }
       tables.delete "schema_migrations"
       tables.delete "spatial_ref_sys"
+      tables.delete "features"
       tables.each { |table| connection.execute("TRUNCATE #{table} CASCADE;") }
       Rake::Task["db:seed"].invoke
       puts "Re-seeded the database!"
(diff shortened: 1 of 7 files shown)


kicferk1 (approved):
This should be a lot more persistent than environment vars on sandbox, staging and prod. On dev and review apps, ehh, each re-deploy we wipe and re-create the db.

`lib/tasks/safe_reset.rake` has code for script doing it. If you remove the `features` table from the list of truncated tables in there, the features should persist across deployments on dev and review too.

jamesrose:
@kicferk1 Cheers!

lib/tasks/safe_reset.rake
```diff
@@ -11,6 +11,7 @@ namespace :db do
                    .map { |row| row["tablename"] }
       tables.delete "schema_migrations"
       tables.delete "spatial_ref_sys"
+      tables.delete "features"
```
ethanmills, line 14: Do we actually want this? I would have thought we wanted the flags to persist between deploys on dev?

ethanmills (approved):
Looks good, one question

PR #415, opened 2021-05-26
Update session trim job to run daily instead of monthly

ethanmills:
### Context
As discussed on slack, this corrects the cron schedule to run daily instead of monthly


kicferk1 (approved):
Does that run at 1am rather than midnight?

In any case, it should run daily, so it's a tick from me.

PR #416, opened 2021-05-26
Cpdrp 283 manage schools cohort

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-283)
View a school's induction choice by cohort

Adds view of cohorts to admin pages

### Testing
I have added cohorts and induction choices for 2022 so that we can test more than 1 cohort. The schools in the test data with URNs that are greater than "000100" and are even numbers, have choices for both 2021 and 2022.

### Review Checks

Given you are signed in as a DfE Admin user, select a school from the list of schools and then select "Cohorts" from the nav menu. You should see the induction for each cohort.


app/policies/school_cohort_policy.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class SchoolCohortPolicy < ApplicationPolicy
+  def index?
+    user.admin?
+  end
```
kicferk1, line 6: Can we have a test for it?

app/models/school_cohort.rb
```diff
@@ -18,6 +18,14 @@ def training_provider_status
     school.partnerships&.active&.exists?(cohort: cohort) ? "Done" : "To do"
   end
 
+  def lead_provider
+    school.lead_provider(cohort.start_year)
+  end
+
+  def delivery_partner
+    school.delivery_partner_for(cohort.start_year)
+  end
```
kicferk1, line 27: Can we unit test those?

app/controllers/admin/schools/cohorts_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    skip_after_action :verify_authorized
+    before_action :set_school
+
+    def index
+      @school_cohorts = policy_scope(SchoolCohort)
+        .joins(:school)
+        .where(school: @school)
+        .joins(:cohort)
+        .order("cohorts.start_year ASC")
+    end
```
kicferk1, line 14: Do we need to paginate it? There will be thousands of schools with a school cohort :/

kicferk1: I guess the second part is: Do we need this list to be searchable? Not necessarily in this PR, just whether it's taken into account in general.

tonyheadford: This is only the cohorts for a single school?

db/seeds/test_data.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 DOMAIN = "@digital.education.gov.uk" # Prevent low effort email scraping
+cohort_2022 = Cohort.find_or_create_by!(start_year: 2022)
```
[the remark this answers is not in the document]
tonyheadford: Maybe, I don't know the reason why this was split into initial and test seeds so put it in the test seeds - should that future cohort go in the initial seeds now?

app/policies/school_cohort_policy.rb
tonyheadford: Good spot, actually this is not needed now so I will remove it.

db/seeds/test_data.rb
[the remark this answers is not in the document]
kicferk1: So `initial_seed` is the common part between all environments. The other seed files are for more specific envs, with fake data that we don't want on 'real' versions.

I don't think we need the 2022 cohort in the initial seed file any time soon, but it is a piece of data that is correct and likely the same across all environments. Thoughts @ethanmills ?

kicferk1 (approved):

db/seeds/test_data.rb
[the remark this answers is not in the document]
ethanmills: Let's leave it here until we've thought about dealing with multiple cohorts properly. We don't need it any time soon on production, and I can't see any benefit in having it there atm

ethanmills (approved):

jamesrose (approved):

PR #417, opened 2021-05-26
Adjust deploy workflow so that master reflects production

jamesrose:
### Context

We currently update `master` to reflect what we are _attempting_ to deploy to staging. If the deploy to staging failed `master` would diverge from what was deployed to staging.

This PR adjusts the deploy workflow so that master reflects what has been _deployed_ to production. I think it is more useful if we have visibility of production than staging like this.

I also needed to relax the linear history constraint default as we had previously changed this protection rule off on develop.

It's worth noting that this is not tested because it needs to be merged first.

### DIFF
--- a/.github/workflows/deploy_to_production.yml
+++ b/.github/workflows/deploy_to_production.yml
@@ -40,3 +40,66 @@ jobs:
           cd terraform/app
           terraform init -reconfigure -input=false -backend-config="bucket=paas-s3-broker-prod-lon-037149fa-bf9a-4577-a066-f627d277d6c4"
           terraform apply -input=false -auto-approve -var-file ../workspace-variables/production.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_PRODUCTION}}", "RELEASE_VERSION":"${{github.event.inputs.version}}"}' -var 'logstash_url=${{secrets.SYSLOG_DRAIN_URL}}'
+
+      - name: Branch protection OFF
+        uses: octokit/request-action@v2.x
+        with:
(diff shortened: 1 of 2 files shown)


jamesrose:
We spoke today about changing the tag format. I'll return to that in another PR later.

.github/workflows/deploy_to_staging.yml
```diff
@@ -35,74 +35,11 @@ jobs:
...
-          git branch tmp-branch
-          git switch master
-          git reset --hard tmp-branch --
-          git branch --delete --force tmp-branch
           git tag --force ${{ github.event.inputs.version }}
-
           git push --force origin refs/tags/${{ github.event.inputs.version }}
-          git push --force origin master
-
-          echo "HEAD=$(git rev-parse ${{ github.event.inputs.version }})" >> $GITHUB_ENV
```
ethanmills, line 74: We still need this line

ethanmills: It's used in retagging the docker image

ethanmills: This is what turns e.g. `develop` into a sha that we have an image for

.github/workflows/deploy_to_production.yml
```diff
@@ -40,3 +40,68 @@ jobs:
...
+      - name: Update master
+        run: |
+          git fetch
+          git branch tmp-branch
+          git switch master
+          git reset --hard tmp-branch --
+          git branch --delete --force tmp-branch
+          git push --force origin master
+
+          echo "HEAD=$(git rev-parse ${{ github.event.inputs.version }})" >> $GITHUB_ENV
```
ethanmills, line 76: Don't need this line here, it should still be in the staging release

ethanmills (approved):

PR #418, opened 2021-05-26
Cpdrp 273 lead provider partnership challenged

BroiSatse:
### Context

https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-273

When the partnership is challenged by the school, we will:
* dijsplay appropriate tag naxt to the school name on the lead provider's list of schools
* display a notification banner with details about the challenge and next steps to take.

There are quite a few new things propsed in this PR apart from the business requirements:

1. **Isolated ViewComponents tests** When writing view-component test, we can now stub other components. Stubbed component will yield a random output, ommitting all of its implementation details. It comes together with a matcher `expect(view).to have_rendered_component(ComponentClass).with(component_initialize_arguments)` (also available as a simple `expect(view).to have_rendered(component_class)`. The matcher tests that the tested output contains the random output generated by stubbed component, so it will detect mistakes like `<% render` in place of `<%= render`. This allows each component to be well tested in isolation.
1. **Refactor of `LeadProvider - YourSchools` and `LeadProvider - SchoolDetails`**. Previosuly, both pages were based on the school lookup based on current user and given cohort (via optional selected_cohort GET param). While displaying the schools, we had to get a lot of information via separate, Partnership related queries. Refactoring these to query Partnerships has made it a bit simpler. Also, for this reason:
1. **Added ransack** - for easier search forms without odd scopes like `with_name_like`, `with_name_or_urn_like` etc.
1. **Added Capybara** - this is just for the Capybara matchers like `have_content`, `have_link` or `have_element` to be used within view-component specs.


app/controllers/admin/schools_controller.rb
```diff
@@ -9,13 +9,12 @@ class SchoolsController < Admin::BaseController
...
-                                     .order(:name)
-                                     .page(params[:page])
-                                     .per(20)
-      if @query.present?
-        @schools = @schools.search_by_name_or_urn(@query)
-      end
+      @schools = policy_scope(School)
+        .includes(:induction_coordinators, :local_authority)
+        .order(:name)
+        .ransack(name_or_urn_cont: @query).result
```
ethanmills, line 15: What happens is `@query` is nil here?

app/controllers/lead_providers/your_schools_controller.rb
```diff
@@ -12,18 +12,21 @@ def index
...
-      @schools = School.eligible.partnered_with_lead_provider(@lead_provider.id, @selected_cohort.start_year)
-        .includes(:early_career_teachers)
-        .order(:name)
+      @partnerships = Partnership
+        .includes(school: :early_career_teachers)
+        .order("schools.name")
+        .where(
+          cohort: @selected_cohort,
+          lead_provider: @lead_provider,
+        )
```
ethanmills, line 21: I might have misunderstood, but I think we're missing an edge-case here. 

1. LP reports partnership with school
2. School declares that partnership is incorrect
3. LP reports partnership with school again

We end up with two partnerships in this instance, one challenged and one not. Would the school appear twice in this list?

app/models/school.rb
```diff
@@ -31,27 +31,27 @@ class School < ApplicationRecord
...
+  # }
+  #
+  # scope :search_by_name_or_urn, lambda { |search_key|
+  #   with_name_like(search_key).or(with_urn_like(search_key))
+  # }
+  #
+  # scope :search_by_name_or_urn_or_delivery_partner_for_year, lambda { |search_key, year|
+  #   joins(partnerships: %i[delivery_partner cohort])
+  #     .with_name_like(search_key).or(with_urn_like(search_key)).or(with_delivery_partner_for_year_like(search_key, year))
+  # }
```
ethanmills, line 54: remove?

app/controllers/lead_providers/your_schools_controller.rb
BroiSatse: Ah, excellent question! I'd imagine we should just update the existing partnership instead of creating a new record?

app/controllers/admin/schools_controller.rb
BroiSatse: `ransack` ignores empty values :)

app/controllers/lead_providers/your_schools_controller.rb
ethanmills: No, we need to retain an easy record of the challenge

ethanmills: Otherwise answering a question like "how often are challenged partnerships reinstated" becomes quite tricky. And that's something we'll be interested in

ethanmills: Not sure what the best way to do this would be. If you used [active_record_distinct_on](https://github.com/alecdotninja/active_record_distinct_on) it could become `Partnership.distinct_on(:school_id).order(:school_id, challenged_at: :desc)`

ethanmills: Doesn't seem very robust though, `.count` wouldn't work

BroiSatse: Ok, I *think* it would be great for complexity here to decouple the current state of the partnership from the history of that partnership. If the partnership is reinstated after the challenge, I'd argue this is still the same challenge (not to say that a unique index on [:school_id, :lead_provider_id] would be quite a nice constraint to have). Once the school is reconfirmed, than the old, challenged partnership has no domain meaning. If we were to look at this with Event Sourcing, all the Partnership::Reported and Partnership::Challenged events would live in a single stream.

There are quite a few options here - one being having duplicated partnerships. But this makes things really complicated in terms of getting the list of partnerships. Honestly, it feels a bit strange that we are displaying the challenged partnership in the "your schools" list - it is not really their school when it is challenged.

Another option - papertrail the partnerships. We could add extra scopes to find the challenge versions so that the stats are visible.

And option 3 - database Log record. (Polymorphing) object belonging to any record in the system representing major changes in the history of that object. It is a bit like a reversed event sourcing, where we are still in monolithic state of "now" as a source of truth, but with some history points.

I'd vote for option 3!

ethanmills: I don't see how 3 differs from papertrail? Is that not a description of the Versions table?

ethanmills: OR we propose e.g. having a Your Schools that is _actually_ your schools, and a separate view for all partnership challenges? That is not incompatible with any of the other options here, but seems to make the most sense as an interface, right?

BroiSatse:
Closing to split the logic into separate, more coherent PRs.

PR #419, opened 2021-05-26
CPDRP-296: GIAS email updates

ethanmills:
### Context
https://dfedigital.atlassian.net/browse/CPDRP-296

Schools are updating their contact emails on GIAS (we're telling them to) but at the moment we're not listening to that change.

We can't update all details of the school yet because that's complicated, but we want the latest emails.

Daily task to update school emails

### Testing
Tested locally, unit tests

### Review Checks

Not sure...

### DIFF
--- a/app/jobs/school_email_update_job.rb
+++ b/app/jobs/school_email_update_job.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class SchoolEmailUpdateJob < CronJob
-  self.cron_expression = "0 2 * * *"
+  self.cron_expression = "30 6 * * *"
 
   queue_as :update_emails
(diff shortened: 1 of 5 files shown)


app/jobs/school_email_update_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class SchoolEmailUpdateJob < CronJob
+  self.cron_expression = "30 6 * * *"
```
ethanmills, line 4: I've chosen daily at 0630 since the source files seem to typically be generated by GIAS between 0530 and 0615

kicferk1 (approved):

PR #421, opened 2021-05-27
updated README.md to reflect what the new payment calculator is actually doing

blackrat:
The payment_calculator section in the readme file has been updated to reflect what is actually happening in the code.

Specifically, it now expands on usage and scope of the calculator, since that has now been updated to require a contract object to be passed in.


ethanmills (approved):
I'll trust you that this is correct 😛

PR #422, opened 2021-05-27
View component testing

BroiSatse:
### Context

At the moment, we are making very little use of ViewComponents. In general, they make it much easier to test the view logic of the application and as such I'd propose we use it more and more together with comprehensive view testing suite to increase our confidence in our unit tests and, in result, reduce required amount of feature tests.

While ViewComponents are already great improvement in testability, they are lacking an option to be tested in isolation. This approach is based on the testing approach used in Vue.js called shallow-mount. WHen rendering given component, developer can choose not to render one or more sub-components and instead just check if given subcomponent has been at all rendered as well to test the attributes given to the subcomponent.

Stubbing component isolates the test from the implementation of the subcomponent, preventing test failures in case of changes to the subcomponent.

### Usage

```
# my_component.html.erb
<%= render MyOtherComponent.new(value: some_value) %>

# spec/my_component_spec.rb
RSpec.describe MyComponent, type: :view_component do
  let(:component) { described_class.new } # component instance must be defined explicitly as a let

  it { is_expected.to have_rendered(MyOtherComponent).with(value: :some_value) }
end
```


callumacrae:
We actually had component tests before but decided to get rid of them in favour of running our components through Percy (which we never actually go around to adding) as the components can break without these tests breaking. Is there any advantage of this approach over the one we decided on before?

ethanmills:
> We actually had component tests before but decided to get rid of them in favour of running our components through Percy (which we never actually go around to adding) as the components can break without these tests breaking. Is there any advantage of this approach over the one we decided on before?

My 2cents on this is that this allows lots of edgecase testing. Percy is good for visual testing, this is good for view logic testing

callumacrae:
> My 2cents on this is that this allows lots of edgecase testing. Percy is good for visual testing, this is good for view logic testing

fair enough - but that's not what these tests are currently doing

BroiSatse:
@callumacrae - This PR has been extracted form the larger piece of work to make the changes a bit more readable. In the original change one of the test was chacking for the existance of the error tag in some scenarios. This PR only provides some tools to make those tests properly isolated and they will be put in use in further PRs.

As per testing philosophy - Percy is a visual tool and as such it lies even over the feature tests in testing pyramide and number of such test should be limited: https://sparkbox.com/foundry/design_system_visual_regression_testing. Visual tests cannot replace unit tests - their testing scope is completely different. Unit tests serve to cover every possible functional scenario for the given component. The goal here is to increase our confidence in unit tests to such a level that feature (and visual) tests are less needed.

app/components/lead_providers/your_schools/table_row.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module LeadProviders
+  module YourSchools
+    class TableRow < BaseComponent
+      include PaginationHelper
```
kicferk1, line 6: Does the row need this helper?

app/components/lead_providers/your_schools/table_row.rb
```diff
@@ -0,0 +1,20 @@
...
+      with_collection_parameter :school
+
+      def initialize(school:, cohort:)
+        @school = school
+        @cohort = cohort
+      end
+
+    private
+
+      attr_reader :school, :cohort
```
kicferk1, line 17: Are these attribute readers actually private?

And more importantly, do we want them to worry about this? I don't think any place ever attempts to do anything with view components other than render them, so having these attributes as public is not likely to break anything.

And I do like having this stuff at the top of the file, purely out of aesthetics.

spec/components/lead_providers/table_spec.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+RSpec.describe LeadProviders::YourSchools::Table, type: :view_component do
+  let(:schools) { Kaminari.paginate_array(Array.new(rand(5..10)) { |i| double("School #{i}") }).page(1) }
```
kicferk1, line 4: What is `double` doing here?
Also, I get a bit lost around second `{` here - can we do it with `do-end` syntax and some intermediate steps?

spec/support/view_component/context.rb
```diff
@@ -0,0 +1,25 @@
...
+        )
+      end
+
+      def output_for(instance)
+        components.find { |comp| comp.instance == instance }.output
+      end
+
+      def components
+        @components ||= []
+      end
```
kicferk1, line 22: When is this wiped? Every test? Every suite? Every test run?

spec/support/view_component/context.rb
```diff
@@ -0,0 +1,25 @@
...
+module Support
+  module ViewComponent
+    class Context
+      StubbedComponent = Struct.new(:instance, :args, :output)
+
+      def register_component(instance, args)
+        components << StubbedComponent.new(
+          instance,
+          args,
+          Faker::Lorem.paragraphs.join(" "),
```
kicferk1, line 12: Uhh, are we sure it's a good idea to have a bunch of lorem ipsum content in our stubbed components? Can we maybe do something like `Stubbed Component #{instance}`, or something to that effect? I don't know when this gets actually output, if at all.

callumacrae:
I agree with that, but given that components are inherently visual I'd also argue that you _do_ want visual tests for every possible state (at least, to the extent that you'd test them in unit tests)

spec/support/view_component/example_group.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Support
+  module ViewComponent
+    module ExampleGroup
+      extend ActiveSupport::Concern
+
+      included do
+        subject(:rendered) { render_inline component }
+        let(:_test_context) { Context.new }
```
kicferk1, line 10: Why are we putting `_` before the test name? Is it a new convention of some type?

spec/support/view_component/matchers.rb
```diff
@@ -0,0 +1,39 @@
...
+        chain :with do |*args|
+          @arguments = RSpec::Mocks::ArgumentListMatcher.new(*args)
+        end
+
+      private
+
+        def arguments
+          @arguments ||= RSpec::Mocks::ArgumentListMatcher.new(any_args)
+        end
+      end
```
kicferk1, line 26: Can we move this into the `match`? I find a `private` method for a `define` scope to be a bit of a mind bender.

spec/support/view_component/matchers.rb
```diff
@@ -0,0 +1,39 @@
...
+  module ViewComponent
+    module Matchers
+      extend RSpec::Matchers::DSL
+
+      define :have_rendered_component do |component_class|
+        match do |rendered|
+          @stubs = _test_context.components.select { |stubbed| stubbed.instance.is_a? component_class }
+          @matching = @stubs.select { |stubbed| arguments.args_match?(*stubbed.args) }
+          @matching.any? do |stubbed|
+            rendered.content.include?(stubbed.output)
```
kicferk1, line 13: Is that making use of the lorem ipsums above? Can it find a wrong stubbed component if their output is the same?

spec/support/view_component/matchers.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      private
+
+        def arguments
+          @arguments ||= RSpec::Mocks::ArgumentListMatcher.new(any_args)
+        end
+      end
+
+      def have_rendered(object, *args)
+        return super unless object.is_a?(Class) && object < ::ViewComponent::Base
```
kicferk1, line 29: What does `super` do here exactly?

spec/support/view_component/example_group.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Support
+  module ViewComponent
+    module ExampleGroup
```
kicferk1, line 5: Can we give this a better name? As far as I can tell it's used to add stubbing capability to places in test files that have some tests in them (contexts, describes). `ComponentStubHelper`?

kicferk1 (commented):
Alright, that is much easier to follow on its own than in another PR. Several small comments, but overall it looks useful :)

spec/support/view_component/matchers.rb
ethanmills, line 13: Is this checking that the correct lorem text is present in the parent?

app/components/lead_providers/your_schools/table_row.rb
BroiSatse: Ah, famous copy-paste artifact! :D

app/components/lead_providers/your_schools/table_row.rb
BroiSatse: Private/public split is really quite philosophical. Making all the methods public won't break anything. Some companies even imposes a public-only policy (and some other work with JS where they have no other option). I personally like making a distinction between public and private API - this is however more for other developers than the requirement imposed by the code.

As per localisation of that call - from my experience it is just a matter of getting used to it. I absolutely hated that when encounter this style some time ago, it is pretty hard to write it differently now. :)

spec/components/lead_providers/table_spec.rb
BroiSatse: `double` is a handy object meaning "I do not really care what it is in this test". All I care here is that whatever it is, it is being passed to a TableRow.

I think I have refactored it in the next branch, I'll bring it here. I also consider if I shouldn't just accept a scope of schools here and paginate it inside the component?

spec/support/view_component/context.rb
BroiSatse: Each test has it's own context. :)

spec/support/view_component/context.rb
BroiSatse: I'd like to enforce usage of our matchers instead of `expect(page).to have_content("Stubbed Component"), which is why I'd prefer random content. This content is being outputed and completely replaces the stubbed component output.

spec/support/view_component/example_group.rb
BroiSatse: Good question. TestContext is an implementation detail of the test setup and as such I wanted to hide it as much as possible to avoid accidental override. `_` is a convention I encounter in javascript to denote private/public interface of the objects. Alternatively I could come up with some sort of registry that creates a new example-specific context before the test and then removes in after hook, but this seems a bit over the top :P

spec/support/view_component/matchers.rb
BroiSatse: Yeah it is a bit odd. Block given to `define` is evaluated in the context of the new matcher class, so in fact `define` is very similar to `class` (but with a symbol rather than constant) which makes it slightly easier to read. We could move this, however this pattern will keep appearing from time to time - see spec/support/delayed_job.rb`

spec/support/view_component/matchers.rb
BroiSatse: Yes, context stores all the outputs and uses them in the matcher to find that the component has in fact been rendered. If we had two components that has the same output then it could, theoretically, find a wrong component in the output. However the chance of this happening is infinitesimal - I haven't got a single duplicate in 1,000,000 random paragraphs, and we're using `paragraphs` which by default uses 4 paragraphs. I'd say probability would be comparable with uuid colliding.

spec/support/view_component/matchers.rb
BroiSatse: There's already a matcher `have_rendered` in some contexts (views/requests) to which I'd like to extend those matchers as well. We only want to use our matcher when component is given.

spec/support/view_component/matchers.rb
BroiSatse: Yes :) This will catch errors like having <% instead of <%=

spec/support/view_component/example_group.rb
BroiSatse: One might argue that this file defined the entire `type: :view_component` type of specs which are usually addressed by ExampleGroup (like "RSpec::Rails::ModelExampleGroup".) This particular implementation also relies on existence of defined `component` which might not be required for other contexts stubbing might be required (like view/request). Maybe this module should be simplified then (without component) and we could create `ViewComponentExampleGroup` then outside of view_component folder?

jamesrose (approved):
Looks good to me :)

spec/support/view_component/example_group.rb
kicferk1: fine, just wanted to make sure that it is our convention and not some Ruby / Rails magic :)

spec/support/view_component/example_group.rb
kicferk1: I actually like the idea of stubbing components in a views a lot, but we can probably add that capability in a future PR (well, relatively close future I think, from the looks of the prior longer PR - I think about testing that a banner is shown in view in some scenarios).

spec/components/lead_providers/table_spec.rb
BroiSatse: We can pass the page as an attribute. Just wrote that and it looks pretty solid to me :)

spec/support/view_component/matchers.rb
BroiSatse: Absolutely, uuid will do the trick (should be also slightly faster). :)

spec/support/view_component/example_group.rb
BroiSatse: @kicferk1 - This is one thing that I keep thinking about actually. Who should be responsible to make the decision whether given component should render or not? It looks like it lies somewhere between parent and the component itself - on one hand I'd like to test in the view that banner under certain conditions, but it is a banner component that knows what is required for that component to be rendered.

ViewComponent allows to define `render?` method that stops the component from rendering. So, given ChallengedBanner component - it knows it makes no sense to be rendered when given partnership is not challenged, so it seems it should go into its `render?` logic. But this leaves us with unconditional component rendering in the parent view `<%= render ChallengeBanner.new(parnership: partnership) %>` which feels a bit odd? Thoughts on that?

spec/support/view_component/context.rb
BroiSatse: Ah, that argument works for me. Random output reduced to uuids now. :)

spec/support/view_component/example_group.rb
kicferk1: This is probably something to discuss in greater detail on the PR adding the banner, but I strongly believe it should be the parent's responsibility to decide whether to render a component. 

That's because if it's the child's, and after you get a page with several components, then the logic of what gets rendered on a page gets split out across many files and possibly duplicated. Which makes it harder to read in the parent, harder to refactor, etc.

It gets even harder to understand what will render in a particular scenario when you get nested components.

Now, there are cases when putting that logic in the parent is just not a good idea - when it is complex and depends on internal component state. That's imo the time to shine for `render?`.

PR #423, opened 2021-05-27
Add mentor profile + more API tests

callumacrae:
### Context

Part one of https://dfedigital.atlassian.net/browse/CPDRP-357 (and a couple other tickets)

This PR adds the relevant database bits for a few tickets this sprint including the one linked above. It'll need more adding to it probably, this is just adding the fundamentals so that we don't all try to do the same thing at the same time.

### Testing

yes

you cannot

### DIFF
--- a/app/models/mentor_profile.rb
+++ b/app/models/mentor_profile.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class MentorProfile < ApplicationRecord
+  has_paper_trail
+
+  belongs_to :user
+  belongs_to :school
+  belongs_to :core_induction_programme, optional: true
+  belongs_to :cohort, optional: true
+  has_many :early_career_teacher_profiles
+  has_many :early_career_teachers, through: :early_career_teacher_profiles, source: :user
+  # has_one :participation_record, dependent: :destroy
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -10,7 +10,7 @@ class User < ApplicationRecord
   has_one :lead_provider, through: :lead_provider_profile
   has_one :admin_profile, dependent: :destroy
   has_one :early_career_teacher_profile, dependent: :destroy
-  has_one :core_induction_programme, through: :early_career_teacher_profile
+  has_one :mentor_profile, dependent: :destroy
 
   validates :full_name, presence: true
   validates :email, presence: true, uniqueness: true, format: { with: Devise.email_regexp }
@@ -33,6 +33,15 @@ class User < ApplicationRecord
 
   def early_career_teacher?
     early_career_teacher_profile.present?
+  end
+
+  def mentor?
+    mentor_profile.present?
+  end
+
+  def core_induction_programme
+    return early_career_teacher_profile.core_induction_programme if early_career_teacher?
+    return mentor_profile.core_induction_programme if mentor?
   end
 
   scope :induction_coordinators, -> { joins(:induction_coordinator_profile) }
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_05_25_120431) do
+ActiveRecord::Schema.define(version: 2021_05_27_102930) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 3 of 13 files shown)


app/models/mentor_profile.rb
```diff
@@ -0,0 +1,13 @@
...
+class MentorProfile < ApplicationRecord
+  has_paper_trail
+
+  belongs_to :user
+  belongs_to :school
+  belongs_to :core_induction_programme, optional: true
+  belongs_to :cohort, optional: true
+  has_many :early_career_teacher_profiles
+  has_many :early_career_teachers, through: :early_career_teacher_profiles, source: :user
+  # has_one :participation_record, dependent: :destroy
```
BroiSatse, line 12: Delete

app/models/user.rb
```diff
@@ -35,6 +35,15 @@ def early_career_teacher?
...
   end
 
+  def mentor?
+    mentor_profile.present?
+  end
+
+  def core_induction_programme
+    return early_career_teacher_profile.core_induction_programme if early_career_teacher?
+    return mentor_profile.core_induction_programme if mentor?
+  end
```
BroiSatse, line 45: I don't think this should live on user class. User on its own is just an abstract persona using our service - it should not know about core_induction_programme. I''m relatively happy to add this for now, but I guess we should be striving to reduce use of this class for everything as we are heading for afferent/efferent combo here

spec/models/user_spec.rb
```diff
@@ -90,6 +90,20 @@
...
+
+      expect(user.mentor?).to be true
+    end
+
+    it "is expected to be false when the user does not have a mentor profile" do
+      user = create(:user)
+
+      expect(user.mentor?).to be false
+    end
+  end
```
BroiSatse, line 105: This here is a bit interesting - we are actually testing that our factory is working as expected, not the code itself.

BroiSatse (approved):

app/models/mentor_profile.rb
callumacrae: I left this in on purpose! It's to make it clear that it doesn't have it (but might need to in the future).

spec/models/user_spec.rb
callumacrae: pretty much half the model tests are basically pointless and will never break imo, but good to have in case the logic becomes more complicated in the future

app/models/mentor_profile.rb
kicferk1: Commented out code looks ugly :/ 
Can we add a comment with a TODO here? 
`# TODO: Add a link to participant_record if we need to`?

db/schema.rb
```diff
@@ -302,6 +317,10 @@
     t.uuid "school_id", null: false
     t.uuid "lead_provider_id", null: false
     t.uuid "cohort_id", null: false
+    t.string "status", default: "pending", null: false
+    t.string "reason_for_rejection"
+    t.datetime "accepted_at"
+    t.datetime "rejected_at"
```
kicferk1, line 323: That's funny. Can you re-generate schema from scratch to make sure these belong here?

spec/models/user_spec.rb
```diff
@@ -104,6 +118,29 @@
...
+
+    it "is expected to return ect cip for ect users" do
+      user = create(:user, :early_career_teacher)
+      cip = create(:core_induction_programme)
+      user.early_career_teacher_profile.core_induction_programme = cip
+
+      expect(user.core_induction_programme).to be cip
+    end
+
+    it "is expected to not blow up when no cip" do
```
kicferk1, line 138: `to return nil` please, `blow up` can mean 'become very expensive' :P

spec/requests/api/v1/users_spec.rb
```diff
@@ -10,7 +10,14 @@
     let(:parsed_response) { JSON.parse(response.body) }
 
     before :each do
-      3.times { create(:user) }
+      # Heads up, for some reason the stored CIP IDs don't match
```
kicferk1, line 13: Why would they not match, argh!

If you do `cip = CoreInductionProgramme.create!(name: "Teach First")` do they match? (I hope it's some funny behaviour by factory bot :/ )

spec/models/user_spec.rb
callumacrae: booooooo

spec/requests/api/v1/users_spec.rb
callumacrae: Dunno but it doesn't affect the tests or the current output of the API so I figured I'd just leave a note - don't want to spend time investigating it if it never affects anyone!

db/schema.rb
callumacrae: generates the same thing - not sure what's going on here 🤔

app/models/mentor_profile.rb
BroiSatse: @callumacrae Ah, I though that it is a testing leftover or something. In this case, yea, please add TODO

db/schema.rb
kicferk1: Someone forgot to commit their schema change I guess. Alright, if it's repeatable then it can stay imo.

BroiSatse: @callumacrae - Most likely you run a migration on another branch and those changes are still in your database. Schema is a dump of your current database state, hence the changes. You can either rollback the other migration, reset your whole database or delete those manually

app/models/mentor_profile.rb
callumacrae: done :)

PR #424, opened 2021-05-27
CPDRP-348: Import additional school emails from a CSV file

ethanmills:
### Context
CPDRP-348

Matt provided a CSV of emails scraped from school websites. This cleans and imports that list to a new table. When we come to import these emails in production we will transfer the CSV to the server via SCP and run the job there.

- Create model for additional emails for a school
- Define a ruleset for cleaning additional emails
- Add a task to import emails from a CSV

### Testing
Unit testing, manually tested

### Review Checks

This can be tested as part of CPDRP-374

### DIFF
--- a/app/services/additional_email_importer.rb
+++ b/app/services/additional_email_importer.rb
@@ -33,7 +33,7 @@ private
     email.gsub!(/[\/:]/, "")
     return unless email_good?(email)
 
-    AdditionalSchoolEmail.find_or_create_by!(school: school, email: email)
+    AdditionalSchoolEmail.find_or_create_by!(school: school, email_address: email)
   end
 
   def email_good?(email)
--- a/app/services/additional_email_importer.rb
+++ b/app/services/additional_email_importer.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+require "csv"
+
+class AdditionalEmailImporter
+  attr_reader :logger
+  attr_reader :source_file
+
+  def initialize(logger, source_file = nil)
+    @logger = logger
+    @source_file = source_file
+  end
+
+  def run
+    CSV.foreach(data_file, headers: true, encoding: "ISO-8859-1:UTF-8") do |row|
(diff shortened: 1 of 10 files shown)


db/migrate/20210527113042_create_additional_school_emails.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateAdditionalSchoolEmails < ActiveRecord::Migration[6.1]
+  def change
+    create_table :additional_school_emails do |t|
+      t.references :school, null: false, foreign_key: true, type: :uuid
+      t.string :email, null: false
+
+      t.timestamps
+    end
```
tonyheadford, line 10: Could we have a unique constraint here (email and school_id)?

tonyheadford (commented):
Looks good, I think there's an opportunity to add a unique constraint in the migration and model. Also I think `email_address` is clearer than `email` (which to me is the communication) but not a blocker especially if this naming is already established.

app/services/additional_email_importer.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def matches_keywords(email)
+    keywords = %w[head info enquiries office admin hello principal secretary reception]
+    keywords.any? { |keyword| email.include?(keyword) }
+  end
+
+  def matches_banned_words(email)
+    banned_words = %w[covid corona emergency admissions recruitment absences officer mailto]
+    banned_words.any? { |banned_word| email.include?(banned_word) }
+  end
```
kicferk1, line 56: Soo, why do we need to do those checks exactly? Do we not trust the file we are getting?

kicferk1: Oh, I guess it comes from scraping, so not really :P

ethanmills: Definitely not...

kicferk1 (approved):

PR #426, opened 2021-05-28
CPDTP-133 Add a new PaymentService that generates payment html

timabell:
### Context

This method of output is just enough to allow testing a minimum viable output with relevant end users.

It is intended to be called from the rails console and writes html to standard-out.

This intended to be connected to the event aggregation that is a work in progress on another branch to supply real numbers.

I've PR'd this piece to keep the overall PR sizes down rather than wait for the whole end to end journey.

https://dfedigital.atlassian.net/browse/CPDTP-133

Add a new PaymentService that generates payment html

### Guidance to review

```bash
bundle exec rails console
```

```ruby
PaymentService.new().generate_html({ service_fee: BigDecimal(123, 2), output_payment: BigDecimal(345, 2), total_payment: BigDecimal(567, 2) })
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,18 @@ Where `"name or id"` is a name or id from the `lead_providers` table.
 2. Run `EngageAndLearnApiToken.create_with_random_token!`
 3. Rails console should output a string, that's your unhashed token, you can keep it and use it to access E&L endpoints.
 
+## Generating html payment information
+
+```bash
+bundle exec rails console
+```
+
+```ruby
+PaymentService.new().generate_html({ service_fee: BigDecimal(123, 2), output_payment: BigDecimal(345, 2), total_payment: BigDecimal(567, 2) })
+```
+
+This will in future aggregate received events and do the payment calculations. Currently it just outputs the supplied numbers in html to "standard out" ready for copying into an email for finance etc. to review.
+
 ### Feature Flags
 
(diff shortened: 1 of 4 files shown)


kicferk1 (commented):
Soo, why would we not want this to be its own route with its own tiny controller that's available to admins for time being?

jamesrose:
I'm not sure we need to write any code that generates a table. There might be a `PaymentService` that runs in the foreground, eventually the background, that returns those numbers as Ruby objects. Where those numbers end up - an email, a UI, etc doesn't matter at this point.

timabell:
> Soo, why would we not want this to be its own route with its own tiny controller that's available to admins for time being?

That was what I had in mind, but @jamesrose wanted to go even more minimal than that. It also meant we would have had to have figured out auth questions. I don't have strong feelings either way.

timabell:
> I'm not sure we need to write any code that generates a table. There might be a `PaymentService` that runs in the foreground, eventually the background, that returns those numbers as Ruby objects. Where those numbers end up - an email, a UI, etc doesn't matter at this point.

Do you object to this going in as-is @jamesrose ? Could you suggest what you would like to see instead if anything?

If we don't do something along these lines then I guess it's going to just be instructions for running the forthcoming calculator thing from the console and looking at the default formatting of a ruby object or hash and we can just close this PR. .... Or on the other extreme as @kicferk1 mentions a full webpage.

jamesrose:
@timabell Yeah, I think instructions about how to run the calculator would be useful. I don't think we need to commit this though.

PR #427, opened 2021-05-28
Upgrade cypress to 7.4.0

timabell:
To get the fix for https://github.com/cypress-io/cypress/issues/15101 which is causing the cypress tests to flap in CI.


callumacrae (approved):

kicferk1 (approved):
Let's hope it's enough 😅

PR #429, opened 2021-05-28
Give admins the ability to monitor single school participants

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDRP-357

### Testing

yes

### Review Checks

- Sign in as admin
- Find a school with participants
- Click onto "Participants" tab

### DIFF
--- a/app/controllers/admin/participants_controller.rb
+++ b/app/controllers/admin/participants_controller.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Admin
+  class ParticipantsController < Admin::BaseController
+    skip_after_action :verify_policy_scoped, only: :show
+
+    before_action :load_participant, only: :show
+
+    def show
+      authorize @participant
+    end
+
+  private
+
+    def load_participant
+      @participant = User.is_participant.find(params[:id])
+    end
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -83,11 +83,15 @@ class User < ApplicationRecord
     end.order(:updated_at, :id)
   }
 
+  scope :includes_school, lambda {
+    includes(early_career_teacher_profile: %i[cohort school], mentor_profile: %i[cohort school])
+  }
+
   scope :is_participant, lambda {
-    where.not(early_career_teacher_profile: { id: nil }).or(User.where.not(mentor_profile: { id: nil }))
+    includes_school.where.not(early_career_teacher_profile: { id: nil }).or(User.where.not(mentor_profile: { id: nil }))
   }
 
   scope :in_school, lambda { |school_id|
-    where(early_career_teacher_profile: { school_id: school_id }).or(User.where(mentor_profile: { school_id: school_id }))
+    includes_school.where(early_career_teacher_profile: { school_id: school_id }).or(User.where(mentor_profile: { school_id: school_id }))
   }
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -60,6 +60,16 @@ class User < ApplicationRecord
     end
   end
 
+  def cohort
+    return early_career_teacher_profile.cohort if early_career_teacher?
+    return mentor_profile.cohort if mentor?
+  end
+
+  def school
+    return early_career_teacher_profile.school if early_career_teacher?
+    return mentor_profile.school if mentor?
+  end
+
   scope :induction_coordinators, -> { joins(:induction_coordinator_profile) }
   scope :for_lead_provider, -> { includes(:lead_provider).joins(:lead_provider) }
   scope :admins, -> { joins(:admin_profile) }
@@ -71,5 +81,13 @@ class User < ApplicationRecord
     else
       where("updated_at is not null")
     end.order(:updated_at, :id)
+  }
+
+  scope :is_participant, lambda {
+    where.not(early_career_teacher_profile: { id: nil }).or(User.where.not(mentor_profile: { id: nil }))
+  }
+
+  scope :in_school, lambda { |school_id|
+    where(early_career_teacher_profile: { school_id: school_id }).or(User.where(mentor_profile: { school_id: school_id }))
   }
 end
--- a/app/views/admin/participants/show.html.erb
+++ b/app/views/admin/participants/show.html.erb
@@ -1,0 +1,46 @@
+<% content_for :title, "Participant detail" %>
+
+<h1 class="govuk-heading-l"><%= @participant.full_name %></h1>
+
+<h2 class="govuk-heading-m">Details</h2>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-half">
+    <table class="govuk-table">
+      <tbody class="govuk-table__body">
+        <tr class="govuk-table__row">
+          <th scope="row" class="govuk-table__header">Full name</th>
+          <td class="govuk-table__cell"><%= @participant.full_name %></td>
+        </tr>
+        <tr class="govuk-table__row">
+          <th scope="row" class="govuk-table__header">Email address</th>
+          <td class="govuk-table__cell"><%= @participant.email %></td>
+        </tr>
+        <tr class="govuk-table__row">
+          <th scope="row" class="govuk-table__header">Participant type</th>
+          <td class="govuk-table__cell"><%= @participant.user_description %></td>
+        </tr>
+      </tbody>
+    </table>
+  </div>
+  <div class="govuk-grid-column-one-half">
+    <table class="govuk-table">
+      <tbody class="govuk-table__body">
+        <% if @participant.early_career_teacher? %>
+          <tr class="govuk-table__row">
+            <th scope="row" class="govuk-table__header">Mentor</th>
+            <td class="govuk-table__cell"><%= @participant.early_career_teacher_profile.mentor&.full_name&.presence || 'Not yet assigned' %></td>
+          </tr>
+        <% end %>
+        <tr class="govuk-table__row">
+          <th scope="row" class="govuk-table__header">School</th>
+          <td class="govuk-table__cell"><%= @participant.school.name %></td>
+        </tr>
+        <tr class="govuk-table__row">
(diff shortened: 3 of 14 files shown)


spec/requests/admin/schools/participants_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      expect(response).to render_template("admin/schools/participants/show")
+    end
+
+    it "only displays school's participants" do
+      get "/admin/schools/#{school.id}/participants"
+
+      expect(response.body).not_to include("No participants found for this school.")
+      expect(response.body).to include(mentor_user.full_name)
+      expect(response.body).not_to include(unrelated_user.full_name)
```
kicferk1, line 41: Can we use https://gist.github.com/tomas-stefano/6652111#capybara-matchers for it?

Or better yet, use something like
```
      expect(assigns(:participants)).not_to include unrelated_user
      expect(assigns(:participants)).to include mentor_user
```

app/views/admin/schools/participants/show.html.erb
```diff
@@ -0,0 +1,33 @@
...
+        <th scope="col" class="govuk-table__header">Full name</th>
+        <th scope="col" class="govuk-table__header">Type</th>
+        <th scope="col" class="govuk-table__header">Cohort</th>
+        <th scope="col" class="govuk-table__header">Validation status</th>
+      </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+      <% @participants.each do |participant| %>
+        <tr class="govuk-table__row">
+          <th scope="row" class="govuk-table__header"><%= govuk_link_to participant.full_name, "#" %></th>
```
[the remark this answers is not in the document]
callumacrae: it will go somewhere and it turns out it's in the same ticket! working on it now

spec/cypress/integration/admin/ViewSchoolParticipants.feature
```diff
@@ -0,0 +1,15 @@
+Feature: Admin users viewing school participants
+  As a… DfE Admin 
+  I need… to oversee the process of school induction tutors adding participant details
+  So that… I can view which schools have added their participants and
```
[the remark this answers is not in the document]
callumacrae: i just copied and pasted this from the ticket - cucumber (or at least the parser we're using) doesn't care

spec/requests/admin/schools/participants_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    user.mentor_profile.update!(school: school)
+    user
+  end
+  let(:unrelated_user) { create(:user, :mentor) }
+
+  before do
+    # Initiate lazy-evaluated users
+    ect_user
+    mentor_user
+    unrelated_user
```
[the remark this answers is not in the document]
callumacrae: good to know!

spec/requests/admin/schools/participants_spec.rb
callumacrae: We're not in capybara so we can't use matchers, right?

I don't know what `assigns` does and the docs aren't useful (so are the non-API docs):

image

what does it do? 😅

kicferk1: Uhh, the way I've seen it used in a few other places suggests they check what the value of `@` variable with specified name would be.

spec/cypress/integration/admin/ViewSchoolParticipants.feature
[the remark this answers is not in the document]
kicferk1: Yeah, cucumber doesn't care, but I care 😂 That would be the first description with three dots in our codebase, I think. 

And more importantly, I am not sure why it's on the ticket too, what's the point of having the dots? To signify that the ticket description came from the same template? To let your brain filter out the initial words which are always the same? I don't really know, but I don't think we get the benefits of that by sticking it in code.

And yes, I know it's a conversation about grammar and punctuation in a code review :P

app/controllers/admin/participants_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Admin
+  class ParticipantsController < Admin::BaseController
+    # TODO: probably add this back
```
kicferk1, line 5: I agree, add them in `load_participant` - `authorize @participant` would be great there.

If you want it on a separate PR, cause we don't want to bother adding a policy here, feel free to do it there, maybe with a ticket number? Although this PR feels like it makes sense.

app/views/admin/participants/show.html.erb
```diff
@@ -0,0 +1,44 @@
...
+        </tr>
+      </tbody>
+    </table>
+  </div>
+  <div class="govuk-grid-column-one-half">
+    <table class="govuk-table">
+      <tbody class="govuk-table__body">
+        <tr class="govuk-table__row">
+          <th scope="row" class="govuk-table__header">Mentor</th>
+          <td class="govuk-table__cell"><%= @participant.early_career_teacher? ? @participant.early_career_teacher_profile.mentor&.full_name&.presence || 'Not yet assigned' : '-' %></td>
```
kicferk1, line 31: What if participant is a mentor?

config/routes.rb
```diff
@@ -108,8 +108,11 @@
       get "/replace-or-update-induction-tutor", to: "schools/replace_or_update_induction_tutor#show"
       post "/replace-or-update-induction-tutor", to: "schools/replace_or_update_induction_tutor#choose"
       resources :cohorts, controller: "schools/cohorts", only: :index
+      resources :participants, controller: "schools/participants", only: :index
     end
 
+    resources :participants, only: :show
```
kicferk1, line 114: These should probably be a single resource

spec/cypress/support/on-rails.js
```diff
@@ -54,7 +54,7 @@ beforeEach(() => {
 });
 
 Cypress.on("after:run", () => {
-  reset();
+  // reset();
```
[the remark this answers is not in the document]
kicferk1: You really love dangling state 😂

app/controllers/admin/participants_controller.rb
callumacrae: ooops this was supposed be a _pre-commit_ todo, not one that gets pushed to github 😆

spec/cypress/support/on-rails.js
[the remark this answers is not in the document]
callumacrae: should probably be behind a flag or something so that the people who want sensible cypress behaviour don't have to comment this out all the time 😅

app/models/user.rb
```diff
@@ -60,6 +60,16 @@ def user_description
...
 
+  def cohort
+    return early_career_teacher_profile.cohort if early_career_teacher?
+    return mentor_profile.cohort if mentor?
+  end
+
+  def school
+    return early_career_teacher_profile.school if early_career_teacher?
+    return mentor_profile.school if mentor?
+  end
```
ethanmills, line 71: @BroiSatse is there a better way to delegate this?

app/views/admin/participants/show.html.erb
ethanmills, line 31: Not sure how accessible that `-` is for screen reader users, do we need to consider adding some visually hidden helper text?

config/routes.rb
callumacrae: Why? There is more than one participant - am I misunderstanding resource vs resources?

app/models/user.rb
callumacrae: there's another one for core_induction_programme just above which is basically the same too. it's not great but i didn't want to worry about it too much as some of them will change soon anyway when mentors can have more than one core induction programme

app/views/admin/participants/show.html.erb
```diff
@@ -0,0 +1,46 @@
...
+            <td class="govuk-table__cell"><%= @participant.early_career_teacher_profile.mentor&.full_name&.presence || 'Not yet assigned' %></td>
+          </tr>
+        <% end %>
+        <tr class="govuk-table__row">
+          <th scope="row" class="govuk-table__header">School</th>
+          <td class="govuk-table__cell"><%= @participant.school.name %></td>
+        </tr>
+        <tr class="govuk-table__row">
+          <th scope="row" class="govuk-table__header">Cohort</th>
+          <td class="govuk-table__cell"><%= @participant.cohort&.start_year || '-' %></td>
```
callumacrae, line 41: The accessibility comment applies too here, but can a participant ever _not_ have a cohort outside of automated testing?

kicferk1: Very unlikely. Still, since it's such a fringe case, do we have any designs for it? If not, maybe `No cohort` is a quick fix here?

app/models/user.rb
kicferk1: I am almost certain a better way to delegate this would involve a polymorphic relationship of some kind, or maybe a `ParticipantProfile` that encompases both ECTs and mentors. Either of those would probably be best done in a separate PR.

app/views/admin/participants/show.html.erb
callumacrae: no design and I don't think it's worth asking for one - let's go with "No cohort"

kicferk1 (approved):

app/models/user.rb
BroiSatse: @ethanmills @kicferk1 The longer i think about this the more it feels like we are missing a Participant model (or PariticpantProfile) maybe MentorProfile and EarlyCareerTeacherProfiles are just a subclass of ParticipantProfile (with STI or MTI?)

kicferk1: I am not sure about adding `ParticipantProfile` that is inherited by `EarlyCareerTeacherProfile` and `MentorProfile`. Will the user need to have a link to the `participant_profile`, and that's how we get the common stuff out of them?

I guess it could work, so long as we are sure no teacher will ever simultaneously be a mentor and a mentee 😂 Are we sure of that?

I think we can give it a go and see if things look nicer, but definitely in a separate PR.

PR #430, opened 2021-05-28
Refactor LeadProvider/YourSchool

BroiSatse:
### Context

As written, lead_providers/your_schools queried schools that has partnership with current lead_provider only to query partnership-related data for each school later on.

This PR brings no behaviour change, but makes the controller query the partnerships instead of school. This fixes a  umber of N+1 issues as it makes all the partnership data easy to access.

I have also added ransack for more streamlined search forms. This made a number of scopes on school obsolete.

### DIFF
--- a/app/views/lead_providers/your_schools/index.html.erb
+++ b/app/views/lead_providers/your_schools/index.html.erb
@@ -51,7 +51,7 @@
     <% end %>
   </div>
 
-  <% if @schools.count.zero? %>
+  <% if @partnerships.count.zero? %>
     <h2 class="govuk-heading-s govuk-!-margin-top-4">There are no matching results</h2>
     <p class="govuk-body">Improve your search results by:</p>
     <ul class="govuk-list govuk-list--bullet">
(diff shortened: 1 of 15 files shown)


app/views/lead_providers/your_schools/index.html.erb
```diff
@@ -42,7 +42,7 @@
         <div class="school-search__input">
           <%= f.govuk_text_field(
             :query,
-            value: @query,
+            value: params[:query],
```
kicferk1, line 45: Are we sure it's a good idea to refer to params in views? I quite like the convention of the controller being responsible for handling params.

kicferk1 (approved):
Looks good overall. A bit surprising that there were no changes in cypress, but I guess that goes to show that the refactor is not changing the actual behaviour.

app/views/lead_providers/your_schools/index.html.erb
BroiSatse: Yup, I was on the fence here as well :) Will correct

kicferk1 (approved):

PR #432, opened 2021-06-01
Changed lead-providers/school-details to lead-provider/partnerships

BroiSatse:
### Context

This extends #430. For now, `lead-partnerships/school-details` is based on school record with an additional cohort information and then it queries additional, partnership information. 

This PR changes `/lead-providers/school-details/:school_id?selected_cohort=:cohort_id` to `/lead-providers/partnerships/:partnership_id`. Partnership already has an access to both school and cohort which makes all the page logic simpler.


app/components/lead_providers/your_schools/table_row.html.erb
```diff
@@ -1,6 +1,6 @@
 <tr class="govuk-table__row">
-  <td class="govuk-table__cell" >
-    <%= govuk_link_to school.name, lead_providers_school_detail_path(school, selected_cohort: cohort.id) %>
+  <td class="school-name-cell govuk-table__cell" >
+    <%= govuk_link_to school.name, lead_providers_partnership_path(partnership) %>
```
kicferk1, line 3: What does `school-name-cell` do? Is the css for it gonna come later?

spec/requests/lead_providers/partnerships_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    context "when the user is the lead provider for the school" do
+      it "should show the partnership page" do
+        get lead_providers_partnership_path(partnership)
+
+        expect(response).to render_template :show
+        expect(assigns(:partnership)).to eq partnership
+      end
+    end
+  end
```
kicferk1, line 24: What about the context from the test file that this replaces,
```
    context "when school is not in a partnership with the lead provider" do
      it "should return not found" do
        expect {
          get lead_providers_school_detail_path(school)
        }.to raise_error(ActiveRecord::RecordNotFound)
      end
    end
```
Is this never going to happen by default, because partnership always has a school and a lead provider? What if a lead provider tries to access someone else's partnership?

config/routes.rb
```diff
@@ -84,7 +84,7 @@
 
   namespace :lead_providers, path: "lead-providers" do
     resources :your_schools, path: "/your-schools", only: %i[index create]
-    resources :school_details, path: "school-details", only: %i[show]
+    resources :partnerships, only: %i[show]
```
kicferk1, line 87: Do we need to change any cypress tests for that? I kind of would expect that there would be a test saying 'given something, when I go to partnership page, then I am on a page with path matching xxx'.

kicferk1 (commented):
Not sure why rspec failed :(

BroiSatse:
@kicferk1 - Ha, it failed because of a random scenario data - There are somewhere between 20-30 partnership and we render aither first or 2nd page. There is no 2nd page for 20 partnership, so this test has currently 1/22 chances of failing. Will fix it in a separate PR.

spec/requests/lead_providers/partnerships_spec.rb
BroiSatse: Yup, that's a  bit odd - to rewrite the test I'd need to test that when partnership_id is completely random then it fails with 404, which is kind of given. That test was previously required as an extra authorisation layer which is no longer needed...

And now when I wrote that it just occurred to me that you're right - we need to test lead provider is unable to access other providers partnerships, not only when the partnership is missing. For now I'll just add the test, but it would be nice to have partnership policy here.

config/routes.rb
BroiSatse: Soooo, that's quite interesting discussion - should feature tests at all care about urls? Do users care?

The current test for `your-schools` does not check for the urls. Since this refactor does not change the user experience, I'd vote that we should not check for the urls in the tests. I think the only reason we got sort of used to check the urls is the concurrent nature of capybara when we used to check for url to ensure that page has loaded after form submission...

app/components/lead_providers/your_schools/table_row.html.erb
BroiSatse: Yup, that should be part of another PR. Will remove for now

config/routes.rb
kicferk1: I think feature tests care about urls because it's a relatively easy and resilient way of checking that 'when I click a button I go to the page that that button should lead to'. If we don't have a test like that for the page where you can see a partnership, well, no need to worry about it.

kicferk1 (approved):

PR #434, opened 2021-06-01
Sneaky fixes

kicferk1:
### Context
1. I really should have made `ApiToken` an abstract class when I made it, it's not obvious that it shouldn't be instantiated by itself. 
2. Are we running any cron jobs on dev or review envs? If not, is it intentional? If yes, how do they get scheduled? @BroiSatse?

1. Make ApiToken an abstract class
2. Add cron scheduling to review and dev app commands

### Guidance to review
N/A

### Testing
I guess I will ssh into review app and see if it has any jobs scheduled...

You can't.

### DIFF
--- a/app/models/api_token.rb
+++ b/app/models/api_token.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 class ApiToken < ApplicationRecord
-  self.abstract_class = true
+  # This is meant to be an abstract class
+  # Since it is a base class for a STI, we can't actually make it abstract (not backed by a table)
 
   def self.create_with_random_token!(**options)
     unhashed_token, hashed_token = Devise.token_generator.generate(ApiToken, :hashed_token)
--- a/app/models/api_token.rb
+++ b/app/models/api_token.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class ApiToken < ApplicationRecord
+  self.abstract_class = true
+
   def self.create_with_random_token!(**options)
     unhashed_token, hashed_token = Devise.token_generator.generate(ApiToken, :hashed_token)
     create!(hashed_token: hashed_token, **options)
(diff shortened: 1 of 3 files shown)


app/models/api_token.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class ApiToken < ApplicationRecord
+  # This is meant to be an abstract class
+  # Since it is a base class for a STI, we can't actually make it abstract (not backed by a table)
```
ethanmills, line 5: I think `validates_presence_of :type` here would prevent storing an instance of ApiToken. See https://stackoverflow.com/questions/2850418/rails-sti-prevent-base-class-from-instantiation/4834082#4834082

Happy to leave this as is if you want though

ethanmills (approved):

app/models/api_token.rb
kicferk1: I tried that actually.

The PR that introduced `ApiToken` and STI on it had a problem: how do we migrate existing lead provider tokens into new api tokens? And I chose the easy way out - I added a default value to `type`, namely `LeadProviderApiToken`. So that check won't work 😢 

I guess I could write a migration to undo that, but that feels like significantly more effort than it's worth.

ethanmills: Ah... fair enough 🙂

terraform/workspace-variables/dev.tfvars
```diff
@@ -11,4 +11,4 @@ paas_app_stopped = false
 paas_web_app_deployment_strategy = "blue-green-v2"
 paas_web_app_instances = 1
 paas_web_app_memory = 512
-paas_web_app_start_command = "bundle exec rake cf:on_first_instance db:migrate && rails db:safe_reset && /app/bin/delayed_job start && rails s"
+paas_web_app_start_command = "bundle exec rake cf:on_first_instance db:migrate && /app/bin/delayed_job start && rails db:safe_reset && rails cron:schedule && rails s"
```
kicferk1, line 14: It took me way too long to figure out that the previous version, 
```
"bundle exec rake cf:on_first_instance db:migrate cron:schedule && rails db:safe_reset && /app/bin/delayed_job start && rails s"
```
was failing because we were doing a safe reset after we schedule our jobs, which are in our db 😂 

I thought about not deleting the jobs in safe reset, but decided against it - after all, they could refer to other objects that were deleted in the soft reset.

PR #435, opened 2021-06-01
Add API endpoint for DQT data

asmega:
### Context

- NPQ Registration requires a hard TRN validation check which can be resolved by fetching data from DQT

- Builds upon @ethanmills spike work
- Adds api endpoint which allows clients to fetch data from DQT
- Minor tweak to fix rspec deprecation warning

### Guidance to review

- `ParticipantValidationService` was introduced in an earlier commit but may not fit our needs as we do not collect NI number at the moment but may change in the future
- As a result the API endpoint just proxies the data from DQT so clients can meet whatever needs they may have

### Testing

- Setup DQT api keys so it can call into the sandbox
- Call endpoint with `curl -H "Authorization: Bearer UNHASHED_TOKEN" localhost:3001/api/v1/dqt-records/VALID_TRN`
- See spreadsheet of valid TRNs present in dataset
- Should return a json response with some data

### Review Checks

- NOTE: not sure if review apps are hooked up to DQT sandbox? if not this won't work
- Call endpoint with `curl -H "Authorization: Bearer UNHASHED_TOKEN" localhost:3001/api/v1/dqt-records/VALID_TRN`
- See spreadsheet of valid TRNs present in dataset
- Should return a json response with some data


.env.template
```diff
@@ -7,4 +7,7 @@ GOVUK_WEBSITE_ROOT=http://localhost:3000
 GIAS_API_SCHEMA=https://ea-edubase-api-prod.azurewebsites.net/edubase/schema/service.wsdl
 GIAS_EXTRACT_ID=13904
 GIAS_API_USER=ecftech
-GIAS_API_PASSWORD=
\ No newline at end of file
+GIAS_API_PASSWORD=
+DQT_CLIENT_API_KEY=apikey
+DQT_CLIENT_HOST=http://dqt.com
+DQT_CLIENT_PARAMS={"param": "value"}
```
kicferk1, line 13: Are we sure that environment is the best place to store API parameters?

config/environments/deployed_development.rb
```diff
@@ -57,6 +57,10 @@
   config.gias_api_user = ENV["GIAS_API_USER"]
   config.gias_api_password = Rails.application.credentials.GIAS_API_PASSWORD
 
+  config.dqt_client_api_key = Rails.application.credentials.DQT_CLIENT_API_KEY
+  config.dqt_client_host = Rails.application.credentials.DQT_CLIENT_HOST
+  config.dqt_client_params = Rails.application.credentials.DQT_CLIENT_PARAMS
```
kicferk1, line 62: Are we sure all three of them belong in credentials? I would think only the key belongs there, everything else can live in unencrypted terraform files (`terraform/workspace-variables/review_app_env.yml` for example)

config/initializers/dqt.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "dqt"
+
+Dqt.configure do |config|
+  def json_to_hash(json)
+    JSON.parse(json.gsub("=>", ":"), symbolize_names: true)
+  end
```
kicferk1, line 8: If we don't store the params in an environment, we don't need this method, I think.

config/initializers/dqt.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  config.client.headers = { Authorization: Rails.application.config.dqt_client_api_key } if Rails.application.config.respond_to?(:dqt_client_api_key)
+  config.client.host = Rails.application.config.dqt_client_host if Rails.application.config.respond_to?(:dqt_client_host)
+  config.client.params = json_to_hash(Rails.application.config.dqt_client_params) if Rails.application.config.respond_to?(:dqt_client_params) && Rails.application.config.dqt_client_params.present?
```
kicferk1, line 12: Should we check that we have the right config first, and set everything later? Does having partial config give us anything?

Also, just to make sure, this `config.client` only affects the dqt client, right? Won't do anything to our other clients?

lib/dqt.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Dqt
+  def self.configuration
+    @configuration ||= Configuration.new
+  end
+
+  def self.configure
+    yield(configuration) if block_given?
+  end
```
kicferk1, line 10: @blackrat does this look like something that we can use `InitializeWithConfig` for? I am just going off of name similarity, I haven't checked the capabilities or use cases :P

lib/dqt/api.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def v1
+      @v1 ||= V1.new(client: client) # rubocop:disable Naming/VariableNumber
+    end
+
+  private
+
+    attr_accessor :client
+    attr_writer :v1
+  end
```
kicferk1, line 20: What's the purpose of this class? Do we expect different versions of api client? Do we expect to need them in our codebase at the same time? Is now the time to add support for it?

lib/dqt/api/v1.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      def dqt_record
+        @dqt_record ||= DQTRecord.new(client: client)
+      end
+
+    private
+
+      attr_accessor :client
+    end
+  end
```
kicferk1, line 20: What is the purpose of this? So far as I can see, the purpose is caching the response of `dqt_record`, cause we pass `client` through initialization firther down to DQTRecord. Is that something we need a separate class for?

lib/dqt/api/v1/dqt_record.rb
```diff
@@ -0,0 +1,44 @@
...
+            national_insurance_number: first_item[:niNumber],
+            qts_date: Date.parse(first_item[:qtsAwardDate]),
+            active_alert: first_item[:activeAlert],
+          }
+        rescue Dqt::Client::ResponseError => e
+          if e.response.code == 404
+            nil
+          else
+            raise e
+          end
```
kicferk1, line 35: Let's use full words for variables please, `error` > `e`, `response_error` > `error`.

lib/dqt/client.rb
```diff
@@ -0,0 +1,48 @@
...
+  class Client
+    def initialize(
+      headers: Dqt.configuration.client.headers,
+      host: Dqt.configuration.client.host,
+      params: Dqt.configuration.client.params
+    )
+      self.headers = headers
+      self.host = host
+      self.params = params
+    end
```
kicferk1, line 13: Do we ever need to override these for our client?

lib/dqt/client.rb
```diff
@@ -0,0 +1,48 @@
...
+    def api
+      @api ||= Api.new(client: self)
+    end
+
+    def get(path: "/", params: {})
+      http_client = HTTPClient.new
+
+      headers = {
+        'Content-Type': "application/json",
+      }.merge(send(:headers))
```
kicferk1, line 24: Uhhh, `send` feels dirty. Can we not use `self.headers` here? Ditto about all other `send`s in this file.

lib/dqt/client.rb
```diff
@@ -0,0 +1,48 @@
...
+      params = params.merge(send(:params))
+
+      response = Response.new(
+        response: http_client.get(
+          url(path),
+          header: headers,
+          query: params,
+        ),
+      )
+      raise ResponseError.new("DQT request failed", response) if [*0..199, *300..599].include? response.code
```
kicferk1, line 35: 3xx can potentially not be a failure, at least in http world, right?

lib/dqt/client/response.rb
```diff
@@ -0,0 +1,22 @@
...
+  class Client::Response
+    delegate :code, to: :response
+
+    def initialize(response:)
+      @body = nil
+      self.response = response
+    end
+
+    def body
+      return @body if !@body.nil? || (@body.nil? && @response.body.empty?)
```
kicferk1, line 13:
```
if @body.present? || (@body.nil? && @response.body.empty?)
```

kicferk1 (commented):
I didn't get the chance to look at the tests yet.

I left a bunch of comments, which I guess could have been asked to whoever wrote the original code in Claim service.

The main theme I felt when reading this code was 'hmm, is this a solved problem? could we use some request making gem and save ourselves the trouble of defining what a response is?'

And I guess that is the main question that super-seeds all the questions on details of how the dqt lib works. Do we not want use something like https://github.com/lostisland/faraday instead of rolling out our own tool? I appreciate it has been written already, but, well :P

config/environments/deployed_development.rb
ethanmills: Client params is definitely secret, contains a key. This is the pattern we followed for GIAS. It just seems polite not to leak the API endpoint, it's not public anywhere currently I don't think

config/initializers/dqt.rb
ethanmills: We do in dev

config/initializers/dqt.rb
ethanmills: > Also, just to make sure, this config.client only affects the dqt client, right? Won't do anything to our other clients?
correct

lib/dqt/api.rb
ethanmills: 🤷 Copied from Claim. Check commitwise for what I actually changed. It seemed prudent to use their client, as they wrote the API

ethanmills: If they have versioned the client, I expect they plan to version the API

lib/dqt/api/v1/dqt_record.rb
ethanmills: Rubocop disagrees with you

ethanmills: https://www.rubydoc.info/gems/rubocop/0.75.0/RuboCop/Cop/Naming/RescuedExceptionsVariableName

lib/dqt/client.rb
ethanmills: For this client, it's a failure. No data is a fail

app/controllers/api/v1/dqt_records_controller.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class DqtRecordsController < Api::ApiController
+      include ApiTokenAuthenticatable
+
+      def show
+        hash = Dqt::Client.new.api.dqt_record.show(params: { teacher_reference_number: params[:id].to_i })
```
ethanmills, line 9: Are we sure TRNs can't have leading zeros? What do we gain by `to_i`

spec/routing/sandbox_controller_spec.rb
```diff
@@ -4,7 +4,7 @@
   describe "Based on rails environment routes" do
     context "when it is not sandbox environment" do
       before do
-        Rails.env.stub(production?: true)
+        allow(Rails.env).to receive(:production?).and_return(true)
```
ethanmills, line 7: 👍

ethanmills (approved):

app/controllers/api/v1/dqt_records_controller.rb
asmega: from the sample of data I have I don't see any with leading zero but that may be a false assumption. i just thought it would minimise the chance of someone putting in dodgee data. I will remove it

PR #438, opened 2021-06-01
CPDRP-374: send chaser emails

ethanmills:
### Context
CPDRP-374

We want to send chaser emails to school who have not yet nominated a tutor. We send these to every email we have available for the school

Add a task to email all eligible schools with no nominated tutor

This will need updating once the opt out journey is finished to not email opted out schools

Drop me a message, we can go through this, the additional email CSV (CPDRP-348), and the GIAS email updates (CPDRP-296) all together

### DIFF
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -7,7 +7,6 @@ class InviteSchools
     logger.info "Emailing schools"
 
     school_urns.each do |urn|
-      rate_limit
       school = School.eligible.find_by(urn: urn)
 
       if school.nil?
@@ -21,6 +20,9 @@ class InviteSchools
         school: school,
       )
 
+      send_nomination_email(nomination_email)
+    rescue Notifications::Client::RateLimitError
+      sleep(1)
       send_nomination_email(nomination_email)
     rescue StandardError
       logger.info "Error emailing school, urn: #{urn} ... skipping"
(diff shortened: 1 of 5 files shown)


app/services/invite_schools.rb
```diff
@@ -48,6 +76,21 @@ def email_expiry_date
     NominationEmail::NOMINATION_EXPIRY_TIME.from_now.strftime("%d/%m/%Y")
   end
 
+  # Notify gives us a 3000/min rate limit. Limit this to 1800/min to allow for other calls and leeway
+  def rate_limit
```
kicferk1, line 80: We use `deliver_now`, which I think is synchronous. Do we ever actually get close to 30 requests per second this way?

kicferk1: And secondly, do we need to test it? Maybe less sophisticated `sleep(0.1.second)` would do the trick too and be much simpler to understand?

ethanmills: Not sure what the realised throughput is on Notify, I suspect the answer is "it depends". I went for this over a simple `sleep` since it's going to take _aaages_ already and I wanted to keep it as quick as possible. Happy to remove if you think this is particularly unclear though.

kicferk1: How long does it take per email? I would think for each call, we make an api call to notify, we wait for it to finish, and then we move on to the next one.

And I find it really unlikely that we could do 50 api calls to notify synchronously in a single second. Which is probably why it is taking ages already.

kicferk1: I think if we were ever gonna hit the rate limit on notify we would have when we sent school invite emails - or did we do it in batches smaller than 3k at all times?

ethanmills: @kicferk1 how does that look now?

kicferk1 (approved):

app/services/invite_schools.rb
kicferk1: Much better, thanks :)

PR #440, opened 2021-06-01
CPDRP-392: Safelist notify API callbacks

ethanmills:
Don't throttle Notify API callbacks, as they've been hitting the rate limit

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -2,6 +2,10 @@
 
 # Throttle general requests by IP
 class Rack::Attack
+  safelist("Allow notify callbacks at any rate") do |request|
+    request.path == "/api/notify-callback" && request.post?
+  end
+
   throttle("General requests by ip", limit: 300, period: 5.minutes, &:ip)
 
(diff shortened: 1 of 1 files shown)


tonyheadford (commented):
Is there any way to identify the request as coming from Notify rather than just opening that endpoint up?

kicferk1:
> Is there any way to identify the request as coming from Notify rather than just opening that endpoint up?

It seems there is a `bearer token` option that we can specify on https://www.notifications.service.gov.uk/services/d1207ebf-ac0c-47b8-b1bf-16dc899e0923/api/callbacks/delivery-status-callback - so we probably can tell if the request came from notify.

ethanmills:
Asked if the callbacks come from a fixed set of IPs, so we could just safelist those instead

ethanmills:
> Asked if the callbacks come from a fixed set of IPs, so we could just safelist those instead

The answer is no...

ethanmills:
Will implement bearer auth in separate PR

tonyheadford (approved):
As discussed secure this later

PR #441, opened 2021-06-01
Cpdrp 273 reporting challenged partnership

BroiSatse:
### Context

We currently allow multiple partnership records per cohort between same lead provider and school. While this maintains history of partnership challenges, it makes reading most recent partnership much more complex than required.

To use a single partnership record and allow multiple report/challenge events. To maintain same level of history inspection, event log has been added.

This also fixes the current bug displaying the same school twice in the lead-provider list of schools when the challenged partnership is re-submitted.


app/models/event_log.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class EventLog < ApplicationRecord
+  belongs_to :owner, polymorphic: true
+end
```
kicferk1, line 5: Do we expect to show the events to anyone, like in a timeline or something? At the moment I don't think we are doing anything with them, other than storing them, and for storing object versions over time papertrail is likely to do a much better job than we are.

I guess that is really the question: are we sure we care about the history of partnership statuses enough to create our own events for them?

BroiSatse: I agree that papertrail is great in what it's doing - it maintains a list of versions of given database table. Event log on the other hand does not track the state of the records and only records explicit domain events forming a subset of transaction log. The advantage is that we are in a full control of what events we want to store. Additionally events are easier to query than versions - I believe the only reason we decided to create new partnership instead of updating the existing one in case of reporting previously challenged partnership is for us to have a visibility of how often those challenges happen. While it would still be possible to get this information with papertrail, it would not be as simple as `EventLog.where(owner_type: "Partnership", event: :challenged)`.

Additional benefit of event log is that event log might contain arbitrary information - we can add additional context of the given event - for reporting we could for example include whether it was added via csv upload or school search. Papertrail is concerned with a database record rather than domain object.

In short - papertrail is great when we want to see the state given object was at given time. Event log is a win when we want to query the domain model history.

It is however a good question whether we care enough about this. I think the answer is "not sure", in which case I think the cheap solution like this even log is quite well suited, as we can either expand it scrap it :)

kicferk1: I think if the answer is "not sure" then a much cheaper solution is to ask people whose job it is to be sure about (or chase after) those things. I guess Tom might be that person?

Cause it might be that the answer is "we don't care at all from product standpoint, we just want to have an audit trail in case someone tries to cheat". In which case it would be a pity that some minimal work for it `EventLog` has been done, but at least we wouldn't add something we don't need to our codebase. Which truth be told is growing bigger by the day, with a lot of people working on it.

kicferk1: Alternatively the answer might be 'yeah, it's great, and we will want a feature like that for admins and analytics people later down the line', and it will stay.

BroiSatse: @kicferk1 - Very much agree here. Which is basically why I didn't want that audit trail to affect the database design (as we did alter the "simple" structure of unique partnerships to allow auditing). I have found using event logs extremely useful in the past though, especially for support queries when we need to figure out not only what the current state is (or was) but also how and why we got into that state. I think that the transparency given by the log massively outweights the cost of development - it is really just a small addition to this PR, which "meat" really lies in the alteration to the "report_school" form process.

app/forms/lead_providers/report_schools_form.rb
```diff
@@ -17,13 +17,20 @@ class ReportSchoolsForm
...
             school_id: school_id,
             cohort_id: cohort_id,
-            delivery_partner_id: delivery_partner_id,
             lead_provider_id: lead_provider_id,
           )
 
+          partnership.challenge_reason = partnership.challenged_at = nil
+          partnership.delivery_partner_id = delivery_partner_id
+          partnership.save!
+
```
ethanmills, line 29: I think we also need to update the challenge deadline here, otherwise it will not be possible to challenge the re-reported partnership. It's currently handled in a `before_create`

ethanmills: There's something very strange going on here. We have both a `report_schools_form` and a `confirm_schools_form`, only one of which deals with pending partnerships

ethanmills: In any case, we're missing the `after_create :schedule_activation` here for the following edge case:

1. Partnership reported for CIP school
2. School challenges within 14 days, remains CIP
3. Partnership re-reported at least 14 days after the first one

In this case, the school would never be transitioned from CIP to FIP

app/forms/lead_providers/report_schools_form.rb
```diff
@@ -17,13 +17,20 @@ class ReportSchoolsForm
     def save!
       ActiveRecord::Base.transaction do
         school_ids.each do |school_id|
-          partnership = Partnership.create!(
+          partnership = Partnership.find_or_initialize_by(
```
ethanmills, line 20: We're missing lots of the functionality from confirm_schools_form here. What about schools without a SchoolCohort?

app/forms/lead_providers/report_schools_form.rb
BroiSatse: Good catch! :)

app/forms/lead_providers/report_schools_form.rb
BroiSatse: Ugh, I think the problem here is much bigger - `confirm_schools_form` is not used any more. It seems we (most likely me) missed quite a few silent merge conflicts...

app/models/event_log.rb
kicferk1: The cost of adding an event log in my eyes has little to do with the code change in this PR.

I think the real cost is:
1. We would have two systems that do very similar things (papertrail doesn't deal with 'how' or 'why', sure, but it knows which user got the resource into a particular state, and based on user type we can reverse engineer potential ways into it). 
2. We would need to remember to add relevant event logs everywhere we wanted to track. That adds an extra dimension to code complexity, which needs to be unit tested, which needs to exist in either controllers or services, or forms. One big advantage of papertrail is that it works without the need to add much code, or much tests - but if we want to rely on our EventLog then we probably would need those.
3. Some models will have EventLog, some will have papertrail, some will probably have both, some will have neither. Which imo leads to more developer errors down the line.

Personally, I think these costs outweigh the usefulness, unless we have a product use case for adding them. Bear in mind that when you are the one in control of developing the logging system, you run into a problem of "what was the version of the code when those events got created? And did that version of code log the events I am after?", which is not always easy, and is even harder if we add it to a live system mid-flight as it were. We can maybe do it for the partnerships, but if we try to add them to any other objects, who knows if we can.

Have we ran into a problem (so far) that would be made easy to analyse with EventLogs that was hard to analyse with papertrail? I don't deal with support tickets on R&P so I wouldn't know.

BroiSatse: > We would have two systems that do very similar things

Actually, the difference between is quite big. Papertrail is a database versioning tool, closely coupled with a database system and as such should be only use for technical purposes. Unfortunately, rails team made at least controversial decision to use ActiveRecord pattern as a domain mode and that decision lead to a number of horrific constructs. While there is nothing wrong with ActiveRecord on its own, using it as a domain object is a commonly analysed anti-pattern (https://softwareengineering.stackexchange.com/questions/119352/does-the-activerecord-pattern-follow-encourage-the-solid-design-principles). Similarly, using papertrail for a domain log can be seen as an anti-pattern. Papertrail is only aware of the state changes, but has no understading of the nature of the change - which is also an artefact of rails resourcefulness. To understand the history of the object, one would need to analyse papertrail of many database tables. On the other hand events are closer to the business logic - we are not updating the record in the database, we are reporting a partnership, which is different to challenging a partnership

> We would need to remember to add relevant event logs everywhere we wanted to track.

This can be seen as a win, as we are in the control of defining the lifecycle of our domain object, decoupled from their database representation.

Few more notes - we do not need a full event log for every object in the system. We are still in the monolithic application and only use the logs to mark important event s in the object lifecycle - shift to a fully blown event driven system would be much more time consuming (however this simple event log would make that shift easier). It is still the database that is the source of truth for a current state. You are correct that we have already lost some of the events and there is no way of retrieving them, which is a shame, but also it means that we should start collecting that data. It is easier to discard data than reconstructing it.

Sorry for being a bit defensive here - there are some very subtle, but important differences here and I guess a lot of personal opinions about the RailsWay of writing the code. Rails has made a number of quite odd design decisions which has triggered an avalanche of problematic patterns and it is quit difficult to get back to "standard way" once you worked outside of those patterns.

I am also not to claim that event log pattern is 100% correct here - the only reason it was introduced was to easily check number of challenges in unit of time as it was explained to me this was the reason for previous, more complex database structure. Introducing event log allows the remaining data structure refactor, without loosing said information. Introducing it here is cheap enough and easy to remove, but I do not want to wait for a business decision before the refactor is done (as it is best when this is merged and deployed before the first challenge occurs).

app/services/partnerships/report.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def initialize(cohort_id:, school_id:, lead_provider_id:, delivery_partner_id:)
+      @cohort_id = cohort_id
+      @school_id = school_id
+      @lead_provider_id = lead_provider_id
+      @delivery_partner_id = delivery_partner_id
+    end
+
+    def call
+      Partnership.transaction do
```
kicferk1, line 20: Why not `ActiveRecord::Base.transaction`? We operate on quite a few more things than `Partnership`, will that `transaction` only roll back `Partnership` changes?

app/services/partnerships/report.rb
```diff
@@ -0,0 +1,78 @@
...
+          partnership: partnership,
+          report_id: partnership.report_id,
+        )
+
+        if partnership.pending?
+          PartnershipActivationJob.new.delay(run_at: partnership.challenge_deadline).perform(
+            partnership: partnership,
+            report_id: partnership.report_id,
+          )
+        end
```
kicferk1, line 49: Instead of using `report_id` to make the jobs not do their job when report id changes, can we wipe the future jobs for the partnership when we re-report the partnership?

spec/requests/schools/partnerships_spec.rb
```diff
@@ -46,20 +56,19 @@
...
-            sent_to: user.email,
-            partnership: @partnership,
-            email_type: PartnershipNotificationEmail.email_types[:induction_coordinator_email],
-          )
+        let!(:partnership) do
+          create(:partnership, :in_challenge_window,
+                 cohort: cohort,
+                 lead_provider: lead_provider,
+                 school: school,
+                 delivery_partner: delivery_partner1)
```
kicferk1, line 64: I know you haven't done it, but is this context _inside_ `when the school is in a partnership` context? is our `let!(:partnership)` here overriding the one above? Can we split them up?

app/services/partnerships/report.rb
```diff
@@ -0,0 +1,78 @@
...
+
+        partnership
```
kicferk1, line 51: Do we need to return it for anything other than testing? If not, can we retrieve it differently in testing? Say, a database call fetching a partnership between a school and a lead provider?

spec/services/partnerships/report_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+  subject(:service_instance) do
+    described_class.new(
+      school_id: school.id,
+      cohort_id: cohort.id,
+      lead_provider_id: lead_provider.id,
+      delivery_partner_id: delivery_partner.id,
+    )
+  end
+
+  subject(:result) { service_instance.call }
```
kicferk1, line 18: I think we are conflating the execution of the call with the resulting object. Can we have something like `execute` from one of the other tests to run the service call and a `let` to retrieve the resulting partnership?

spec/support/delayed_job.rb
```diff
@@ -81,7 +81,7 @@ def query_jobs(object, method_name)
         .where("handler LIKE ?", "%\nmethod_name: :#{method_name}\n%")
 
       scope.select do |job|
-        return object == job.payload_object.object unless object.respond_to?(:matches?)
+        next object == job.payload_object.object unless object.respond_to?(:matches?)
```
kicferk1, line 84: What does this do?

.rubocop.yml
```diff
@@ -17,6 +17,10 @@ Style/FrozenStringLiteralComment:
 Layout/EndOfLine:
   EnforcedStyle: lf
 
+Lint/AmbiguousBlockAssociation:
+  Exclude:
+    - 'spec/**/*_spec.rb'
+
```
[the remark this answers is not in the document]
ethanmills: I'm guessing it's for `expect { ... }.to change { ... }.by

app/services/partnerships/report.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
```
ethanmills, line 1: Nice!

app/services/partnerships/report.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      @school_cohort ||= SchoolCohort.create!(
+        school_id: school_id,
+        cohort_id: cohort_id,
+        induction_programme_choice: "full_induction_programme",
+      )
+    end
+
+    def partnership_pending?
+      school_cohort.core_induction_programme?
```
ethanmills, line 75: We need to make sure this gets updated with Tony's opt out work. There will be more statuses here

ethanmills: FYI @tonyheadford

db/migrate/20210601181158_unique_partnerships.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class UniquePartnerships < ActiveRecord::Migration[6.1]
+  def change
+    add_index :partnerships, %i[school_id lead_provider_id cohort_id], unique: true, name: "unique_partnerships"
+  end
```
ethanmills, line 6: I think this would fail if we had data, right? We don't yet in production, worth checking staging

spec/jobs/partnership_activation_job_spec.rb
```diff
@@ -2,36 +2,44 @@
...
     end
 
-    it "does nothing when the partnership request has been challenged" do
-      challenged_partnership = create(:partnership, :pending, :challenged)
-      school_cohort = SchoolCohort.create!(
-        school: challenged_partnership.school,
-        cohort: challenged_partnership.cohort,
-        induction_programme_choice: "core_induction_programme",
-      )
+    context "when aprtnership has been challenged" do
```
ethanmills, line 29: Typo 😛

ethanmills (approved):

.rubocop.yml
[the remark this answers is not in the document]
BroiSatse: Otherwise rubocop flags `expect { sth }.to change { sth }.by 1`, demanding `expect {sth}.to(change {sth}.by(1))` which drives me nuts. :P Happy to discuss though

db/migrate/20210601181158_unique_partnerships.rb
BroiSatse: Yup, I was not very keen on writing consolidation migration hoping we have no data. I've checked both production and staging, but few days ago. I'll double check...

spec/jobs/partnership_activation_job_spec.rb
BroiSatse: "You don't know what aprtnership means??" :P

app/services/partnerships/report.rb
BroiSatse: Yeah, it was nightmare to write a test for the whole iterating form! :P

spec/support/delayed_job.rb
BroiSatse: `return` was terminating the whole method and returning a boolean value as a AR scope. `next` terminates just a single block invocation and yields said boolean to the iterating method (`any?`) allowing further iteration.

app/services/partnerships/report.rb
BroiSatse: TBH, not sure why :) There is no real difference, `transaction` is a method defined on `ActiveRecord::Base` class and I'll never understand why class methods are inherited in ruby :P. I **guess** I went for Partnership here because it is a) a bit shorter to type or 2) we are dealing with partnership domain model object. I probably just swap it back to ActiveRecord::Base for consistency.

Or... shall we be using `ApplicationRecord` instead?

app/services/partnerships/report.rb
BroiSatse: That would be ideal, but is not that straight forward - delayed job is not made for querying. I am happy to hack around with that in test (remember my DJ matchers?), but I would not touch this in production code if the simpler solution is available. Delayed job `delay` does yield the DJ object so we could in theory store the reference, but scheduling AJ job does not. The problematic thing would be to hold a list of job ids to delete associated with partnership record - this would potentially require either a new table (database driven active job!) or a new column with an array of job ids to remove on re-reporting. Will need to think about this a bit more, maybe there is (pretty much for sure) a solution I'm missing

app/services/partnerships/report.rb
BroiSatse: I think the question is really "what would you expect a service called `Partnerhsip::Report` return" rather than `where is the return value being used`. This service should have no knowledge of how and when it is being used and instead it should try to be as useful as possible. As it's only job is to create a new partnership, returning that partnership feels as a natural result of the execution (however I'd much rather prefer to have Success/Failure monad returned with some associated data).

app/services/partnerships/report.rb
BroiSatse: Ah, technically each model can define its own connection as they can live in a separate dabases (which then would work as a "partial" transaction. In this case I'd say connection on ApplicationRecord is probably the most suitable one... However, it really makes no difference in our case and I don't see any value in doing this, so I'll just change this to ActiveRecord::Base (there's a lot of value in consistency!)

spec/requests/schools/partnerships_spec.rb
BroiSatse: I think this is one of the examples where I'd just love to split the view into components and check them being rendered instead of testing the whole html response with all the possible cases. But that's for later.

I have renamed this inner context for now - this is really a special case of the outer context (the partnership being still in the challenge window). As per splitting, for me it is an everlasting battle between well defined contexts and too-many-nested levels (and I am on both sides at the same time). I am still reusing quite a lot of the parent context, so I'd leave it as is (with the inner context renamed), but I'll need to read through those tests at some point as the setup seems a bit much (why do we need 5 delivery partners, what's the other partnership needed for here, why do we need ProviderRelationship on the school partnerships page?)

spec/services/partnerships/report_spec.rb
BroiSatse: Yup, rspec is pretty much noun-oriented, which is tricky when you're testing a verb... Sometimes I do:

```
subject(:execution) { -> { do_sth }`

it { is_expected.to change ...` }
```

Hmmmm, actually now that I think about this, we could make it work out of the box by defining `to_proc` method on the service. Then the service itself would act like a verb in the test while result can be defined as process output

.rubocop.yml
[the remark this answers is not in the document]
kicferk1: Truth be told I just put the second scope in parenthesis when rubocop shouted at me in E&L tests :D 

That's a good enough reason to disable it there imo.

app/services/partnerships/report.rb
kicferk1: Oh yeah, I did forget that the partnership would only be retrievable from some horribly serialized delayed job. Ugh. Alright, I think I prefer having `report_id` over that.

PR #443, opened 2021-06-02
DQT record api returns 404 if no record is found

asmega:
### Context

- API endpoint currently returns data if no DQT record is found
- the client would need to parse this to figure out there is no record

- API now returns 404 if no DQT record is found
- This makes it easier for clients to handle such cases

### Guidance to review

- n/a

### Testing

- Setup DQT api keys so it can call into the sandbox
- Call endpoint with `curl -H "Authorization: Bearer UNHASHED_TOKEN" localhost:3001/api/v1/dqt-records/100`
- Should return body-less 404 response

### Review Checks
- [X] All pages have automated accessibility checks via cypress
- [X] All pages have visual tests via cypress + percy
- [X] All `School` queries are correctly scoped - `eligible` when they need to be

- NOTE: not sure if review apps are hooked up to DQT sandbox? if not this won't work
- Call endpoint with `curl -H "Authorization: Bearer UNHASHED_TOKEN" localhost:3001/api/v1/dqt-records/100`
- Should return body-less 404 response

### DIFF
--- a/app/controllers/api/v1/dqt_records_controller.rb
+++ b/app/controllers/api/v1/dqt_records_controller.rb
@@ -8,7 +8,11 @@ module Api
       def show
         hash = Dqt::Client.new.api.dqt_record.show(params: { teacher_reference_number: params[:id] })
 
-        render json: DqtRecordSerializer.new(OpenStruct.new(hash)).serializable_hash.to_json
+        if hash.present?
+          render json: DqtRecordSerializer.new(OpenStruct.new(hash)).serializable_hash.to_json
+        else
+          head :not_found
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


kicferk1 (approved):
The change looks good to me.

Do we have any swagger docs for the dqt endpoint? I guess not, since they don't get updated in this PR, and they should if we had them. Should we add them? Maybe not in this PR though.

asmega:
> The change looks good to me.
> 
> Do we have any swagger docs for the dqt endpoint? I guess not, since they don't get updated in this PR, and they should if we had them. Should we add them? Maybe not in this PR though.

Good point, I'll get those after this one

PR #451, opened 2021-06-02
Add rake task to run payment calculator to get payment breakdown for a lead provider

plisovin:
Depends on [this branch](https://github.com/DFE-Digital/early-careers-framework/tree/CPDTP-131)

```
rake payment_calculation:breakdown Capita
```

Result:

```
{
  "service_fees": {
    "service_fee_per_participant": "316.67",
    "service_fee_total": "633339.0",
    "service_fee_monthly": "21839.28"
  },
  "output_payment": {
    "per_participant": "587.4",
    "start": {
      "retained_participants": 0,
      "per_participant": "117.48",
      "subtotal": "0.0"
    }
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,10 +134,10 @@ bundle exec rake 'schools:send_invites[urn1 urn2 ...]'
 bundle exec rake lead_provider:generate_token "name or id"
 ```
 
-## Run payment calculator for a given lead provider to generate the payment breakdown
+## Run payment calculator for a given lead provider with an optional number of participants (default is 2,000) to generate the payment breakdown
 
 ```bash
-bundle exec rake payment_calculation:breakdown "name or id"
+bundle exec rake payment_calculation:breakdown "<name or id>" "<number of participants>"
 ```
 
 Where `"name or id"` is a name or id from the `lead_providers` table.
--- a/README.md
+++ b/README.md
@@ -134,6 +134,12 @@ bundle exec rake 'schools:send_invites[urn1 urn2 ...]'
 bundle exec rake lead_provider:generate_token "name or id"
 ```
 
+## Run payment calculator for a given lead provider to generate the payment breakdown
+
+```bash
+bundle exec rake payment_calculation:breakdown "name or id"
+```
+
 Where `"name or id"` is a name or id from the `lead_providers` table.
 
 ## Generating a token for E&L api
@@ -151,7 +157,7 @@ The available flags are listed in `app/services/feature_flag.rb`, and available 
 
 The code in[`lib/payment_calculator/ecf/`](lib/payment_calculator/ecf/)performs payment calculations for [ECFs (Early Career Framework)](https://www.early-career-framework.education.gov.uk/) using commercial information so that training providers can be paid the correct amount.
 
-The calculator can generate each intermediary step in the calculation so that any questions over how the final totals were reached can be answered by interested parties. 
+The calculator can generate each intermediary step in the calculation so that any questions over how the final totals were reached can be answered by interested parties.
 
 Output from `PaymentCalculation.new(contract: <ContractObject>)` will instantiate a calculator for that specific contract. This can then be called, passing in the retention event type, and total number of participants to calculate for. (There is also a class level call shortcut for this.) which means that for a one off calculation you can call `PaymentCalculation.new(contract: <ContractObject>).call(event_type:, total_participants:)`, or `PaymentCalculation.call({contract: <ContractObject>}, event_type:, total_participants:)`. (Note the brackets around the first hash. In this call format, that is what determines what is passed to the initializer and what goes to the call.)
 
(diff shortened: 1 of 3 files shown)


db/seeds/sandbox_data.rb
```diff
@@ -18,15 +18,16 @@ def generate_provider_token(lead_provider, cohort, logger)
...
       user = User.create!(full_name: Faker::Name.name, email: Faker::Internet.email)
 
-      school = School.create!(
-        urn: Faker::Internet.uuid,
-        name: Faker::Company.name,
-        address_line1: Faker::Address.street_address,
-        postcode: Faker::Address.postcode,
-      )
+      school = School.find_or_create_by!(
+        urn: sprintf("%06d", (10_000 + i)),
```
kicferk1, line 25: What does `sprintf` do here? Make sure we have 6 digits in a string?

lib/tasks/payment_breakdown.rake
```diff
@@ -0,0 +1,24 @@
...
+namespace :payment_calculation do
+  desc "run payment calculator for a given lead provider"
+  task breakdown: :environment do
+    logger = Logger.new($stdout)
+    logger.formatter = proc do |_severity, _datetime, _progname, msg|
+      "#{msg}\n"
+    end
+
+    begin
+      lead_provider = LeadProvider.find(ARGV[1])
```
kicferk1, line 12: Instead of using `ARGV`, could we do 
```
task breakdown:, [:lead_provider_name] => :environment do |_task, args|
  lead_provider_name = args.lead_provider_name
```
or something similar?

lib/tasks/payment_breakdown.rake
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+namespace :payment_calculation do
+  desc "run payment calculator for a given lead provider"
+  task breakdown: :environment do
+    logger = Logger.new($stdout)
+    logger.formatter = proc do |_severity, _datetime, _progname, msg|
+      "#{msg}\n"
+    end
```
kicferk1, line 9: The more I see loggers and puts all over the place the more I wonder if we should have a common way of doing them. Oh well, something for later.

lib/tasks/payment_breakdown.rake
```diff
@@ -0,0 +1,24 @@
...
+    rescue StandardError
+      lead_provider = LeadProvider.find_by(name: ARGV[1])
+    end
+
+    breakdown = CalculationOrchestrator.call({ lead_provider: lead_provider }, event_type: :start)
+    logger.info JSON.pretty_generate(breakdown)
+  rescue StandardError
+    logger.info "Lead provider for '#{ARGV[1]}' not found"
+  ensure
+    exit(0)
```
kicferk1, line 22: Do we need this? What happens if we don't have it?

db/seeds/sandbox_data.rb
plisovin: yep, just adding some trailing zeros

lib/tasks/payment_breakdown.rake
plisovin: I personally prefer the ARGV as otherwise you will need to run it using the following:

```
rake "payment_calculation:breakdown[Capita]"
```

instead of a more intuitive:

```
rake payment_calculation:breakdown Capita
```

lib/tasks/payment_breakdown.rake
plisovin: It will treat the command line parameter as a task name and will print an error although after successful execution

lib/tasks/payment_breakdown.rake
kicferk1: Alright.

In that case, could we add a note to readme, or a comment in this file about running this task? Seeing how (again) some tasks use another parameter format?

plisovin: Add here 17f91d9b

db/seeds/sandbox_data.rb
```diff
@@ -18,15 +18,16 @@ def generate_provider_token(lead_provider, cohort, logger)
...
+      ) do |s|
+        s.name = Faker::Company.name
+        s.address_line1 = Faker::Address.street_address
+        s.postcode = Faker::Address.postcode
+      end
```
blackrat, line 30: Can we split this into a separate PR? The seed_data update needs to happen with or without this ticket, and in looking at migrations, we may be missing entries in SchoolCohort to correctly pick up the lead_provider for the update.

blackrat (changes requested):

db/seeds/sandbox_data.rb
plisovin: Ok, this is just a small fix to school urn to have a correct format

plisovin: @blackrat Rolled back

README.md
```diff
@@ -134,6 +134,12 @@ bundle exec rake 'schools:send_invites[urn1 urn2 ...]'
 bundle exec rake lead_provider:generate_token "name or id"
 ```
 
+## Run payment calculator for a given lead provider to generate the payment breakdown
+
+```bash
+bundle exec rake payment_calculation:breakdown "name or id"
```
kicferk1, line 140: Probably warrants an update now that we added number of participants

plisovin: Thanks, done in c6e947e

kicferk1 (approved):

PR #461, opened 2021-06-03
Banding entries with nil sort as the highest values.

blackrat:
### Context

PostgreSQL orders with NULL values being compared as higher than any other numeric.
Ref: https://stackoverflow.com/questions/5826210/rails-order-with-nulls-last/7055259

Force NULLS to be treated first as a sort for band ordering

### Testing
Validate in Console with the Banding model. CalloffContract model test has been updated to use unnamed bands and force the lookup.

### Review Checks

You can't

### DIFF
--- a/app/models/call_off_contract.rb
+++ b/app/models/call_off_contract.rb
@@ -5,6 +5,12 @@ class CallOffContract < ApplicationRecord
   belongs_to :lead_provider
 
   def band_a
-    participant_bands.order(:min).first
+    bands.first
+  end
+
+private
+
+  def bands
+    participant_bands.min_nulls_first
   end
 end
--- a/app/models/participant_band.rb
+++ b/app/models/participant_band.rb
@@ -2,6 +2,7 @@
 
 class ParticipantBand < ApplicationRecord
   belongs_to :call_off_contract
+  scope :min_nulls_first, -> { order("min asc nulls first") }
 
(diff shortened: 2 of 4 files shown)


app/models/call_off_contract.rb
```diff
@@ -5,6 +5,6 @@ class CallOffContract < ApplicationRecord
   belongs_to :lead_provider
 
   def band_a
-    participant_bands.order(:min).first
+    participant_bands.order("min ASC NULLS FIRST").first
```
plisovin, line 8: We could also use scope on the ```ParticipantBand```  

i.e. ```scope :nulls_first, -> { order("min asc nulls first") }```

and then do the following:

```
def band_a
  participant_bands.nulls_first.first
end
```

so in future could be reused for band_b, band_c etc

```
def band_b
  participant_bands.nulls_first.second
end
```

blackrat: Yep. Scope is a good idea. I've modified to something similar to your suggestion.

kicferk1 (commented):
Instead of adding a scope, can we add a `default: 0` to the `min` column in a migration? And make it non-nullable?

plisovin (approved):

blackrat:
> Instead of adding a scope, can we add a `default: 0` to the `min` column in a migration? And make it non-nullable?

That's definitely an approach that would work and is probably more database agnostic.

PR #463, opened 2021-06-03
School induction tutor managing/viewing participants in a cohort

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDRP-360

### Review Checks

### DIFF
--- a/app/controllers/schools/cohorts_controller.rb
+++ b/app/controllers/schools/cohorts_controller.rb
@@ -8,5 +8,7 @@ class Schools::CohortsController < Schools::BaseController
 
   def show; end
 
-  def add_participants; end
+  def add_participants
+    redirect_to schools_cohort_participants_path(@cohort.start_year) if FeatureFlag.active?(:induction_tutor_manage_participants)
+  end
 end
--- a/app/controllers/schools/participants_controller.rb
+++ b/app/controllers/schools/participants_controller.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+class Schools::ParticipantsController < Schools::BaseController
+  before_action :set_school_cohort
+  before_action :set_participant, except: :index
+
+  def index
+    @participants = User.order(:full_name).is_participant.in_school(@school.id)
+    authorize @participants, policy_class: ParticipantPolicy
+  end
+
+  def show
+    @mentor = @participant.early_career_teacher_profile&.mentor
+  end
+
+  def edit_details; end
+
+  def edit_mentor
+    @mentor_form = ParticipantMentorForm.new(
+      mentor_id: @participant.early_career_teacher_profile.mentor&.id,
+      school_id: @school.id,
+    )
+  end
+
+  def update_mentor
+    @mentor_form = ParticipantMentorForm.new(participant_mentor_form_params.merge(school_id: @school.id))
+
+    if @mentor_form.valid?
+      @participant.early_career_teacher_profile.update!(mentor_profile: @mentor_form.mentor ? @mentor_form.mentor.mentor_profile : nil)
+
+      flash[:success] = { title: "Success", heading: "The mentor for this participant has been updated" }
+      redirect_to schools_cohort_participant_path(@cohort.start_year, @participant)
+    else
+      render :edit_mentor
+    end
+  end
+
+private
+
+  def set_participant
+    @participant = User.find(params[:participant_id] || params[:id])
+    authorize @participant, policy_class: ParticipantPolicy
+  end
+
+  def participant_mentor_form_params
+    params.require(:participant_mentor_form).permit(:mentor_id)
+  end
+end
--- a/app/forms/participant_mentor_form.rb
+++ b/app/forms/participant_mentor_form.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class ParticipantMentorForm
+  include ActiveModel::Model
+
+  attr_accessor :mentor_id, :school_id, :user_id
+
+  validate :mentor_exists
+
+  def mentor
+    mentor_id == "later" ? nil : User.find(mentor_id)
+  end
+
+  def available_mentors
+    school.mentors.order(:full_name)
+  end
+
+private
+
+  def school
+    @school ||= School.find(school_id)
+  end
+
+  def mentor_exists
+    if mentor && available_mentors.exclude?(mentor)
+      errors.add(:mentor_id, :not_authorized)
+    end
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -39,6 +39,10 @@ class User < ApplicationRecord
     mentor_profile.present?
   end
 
+  def participant?
+    early_career_teacher? || mentor?
+  end
+
   def core_induction_programme
     return early_career_teacher_profile.core_induction_programme if early_career_teacher?
     return mentor_profile.core_induction_programme if mentor?
@@ -68,12 +72,14 @@ class User < ApplicationRecord
   def school
     return early_career_teacher_profile.school if early_career_teacher?
     return mentor_profile.school if mentor?
+    return induction_coordinator_profile.schools.first if induction_coordinator?
   end
 
   scope :induction_coordinators, -> { joins(:induction_coordinator_profile) }
   scope :for_lead_provider, -> { includes(:lead_provider).joins(:lead_provider) }
   scope :admins, -> { joins(:admin_profile) }
   scope :early_career_teachers, -> { joins(:early_career_teacher_profile).includes(:early_career_teacher_profile) }
+  scope :mentors, -> { joins(:mentor_profile).includes(:mentor_profile) }
 
   scope :changed_since, lambda { |timestamp|
     if timestamp.present?
--- a/app/policies/participant_policy.rb
+++ b/app/policies/participant_policy.rb
@@ -1,11 +1,26 @@
 # frozen_string_literal: true
 
-class ParticipantPolicy < ApplicationPolicy
+class ParticipantPolicy < UserPolicy
   def index?
-    admin_only
+    return true if admin_only
+    return @record.all? { |participant| user_can_access?(participant) } if user.induction_coordinator?
   end
 
   def show?
-    admin_only
+    return true if admin_only
+    return user_can_access?(@record) if user.induction_coordinator?
+  end
+
+  alias_method :edit_details?, :show?
+  alias_method :update_details?, :show?
+  alias_method :edit_mentor?, :show?
+  alias_method :update_mentor?, :show?
+
+private
+
+  def user_can_access?(participant)
+    return false unless participant.participant?
+
+    user.school == participant.school
   end
 end
--- a/app/views/schools/participants/edit_mentor.html.erb
+++ b/app/views/schools/participants/edit_mentor.html.erb
@@ -1,0 +1,24 @@
+<% content_for :title, "Change mentor" %>
+
+<% content_for :before_content, govuk_back_link(
+  text: "Back",
+  href: schools_cohort_participant_path(@cohort.start_year, @participant))
+%>
+
+<%= form_with model: @mentor_form, url: schools_cohort_participant_update_mentor_path(@cohort.start_year, @participant), method: :put, id: "school-cohort-participant-update-mentor" do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <%= f.govuk_radio_buttons_fieldset :mentor_id,
+    legend: { text: "Who is the mentor for #{@participant.full_name}?", tag: 'h1', size: 'xl' },
+    hint: { text: "If you don't know, or haven't added them yet you can do this later." } do %>
+    <% @mentor_form.available_mentors.each_with_index do |mentor, i| %>
+      <%= f.govuk_radio_button :mentor_id, mentor.id, label: { text: mentor.full_name }, link_errors: i == 0 %>
+    <% end %>
+
+    <%= f.govuk_radio_divider %>
+
+    <%= f.govuk_radio_button :mentor_id, "later", label: { text: 'Assign mentor later' } %>
+  <% end %>
+
+  <%= f.govuk_submit %>
+<% end %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -186,6 +186,12 @@ Rails.application.routes.draw do
       resources :partnerships, only: :index
       resource :programme, only: %i[edit], controller: "choose_programme"
 
+      resources :participants, only: %i[index show], constraints: ->(_request) { FeatureFlag.active?(:induction_tutor_manage_participants) } do
+        get :edit_details, path: "edit-details"
+        get :edit_mentor, path: "edit-mentor"
+        put :update_mentor, path: "update-mentor"
+      end
+
       namespace :core_programme, path: "core-programme" do
         resource :materials, only: %i[edit update show] do
           get :info
--- a/spec/requests/schools/participants_request_spec.rb
+++ b/spec/requests/schools/participants_request_spec.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Schools::Participants", type: :request do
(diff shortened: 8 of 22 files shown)


app/views/schools/cohorts/add_participants.html.erb
```diff
@@ -1,22 +0,0 @@
...
-    <p>We will let you know when you can get started.</p>
-    <p>For each early career teacher and mentor, you will need:</p>
-    <ul class="govuk-list govuk-list--bullet">
-      <li>first name</li>
-      <li>last name</li>
-      <li>teacher reference number (TRN)</li>
-      <li>date of birth</li>
-    </ul>
-  </div>
-</div>
```
ethanmills, line 22: You don't want to delete this yet, I think we're releasing this one under a feature flag, so we're still going to need this

config/routes.rb
```diff
@@ -186,6 +186,11 @@
       resources :partnerships, only: :index
       resource :programme, only: %i[edit], controller: "choose_programme"
 
+      resources :participants, only: %i[index show update], constraints: ->(_request) { FeatureFlag.active?(:induction_tutor_manage_participants) } do
+        get :edit_details, path: "edit-details"
+        get :edit_mentor, path: "edit-mentor"
+      end
```
kicferk1, line 192: `constraints` are fancy, but really, we could also brute force it with 
```
if FeatureFlag.active?(:induction_tutor_manage_participants)
  resources :participants ...
end
```
The fact that you aren't using `request` kind of tells me `constraints` was designed to be a little more sophisticated.

app/controllers/schools/participants_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    @mentors = @school.mentors.order(:full_name)
+  end
+
+  def update
+    @mentor_form = ParticipantForm.new(participant_form_params)
+
+    # mentor_id can be nil
+    if @mentor_form.mentor_id
+      mentor = @mentor_form.mentor_id != "later" && User.find(@mentor_form.mentor_id)
```
BroiSatse, line 33: Could we move it inside the form. Would be great to contain `"later"` trick within a single object.

spec/requests/schools/participants_request_spec.rb
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Schools::Participants", type: :request do
+  let(:user) { create(:user, :induction_coordinator) }
+  let(:school) { user.induction_coordinator_profile.schools.first }
+  # TODO: ?
```
kicferk1, line 8: Hmmm, what could it mean?

app/controllers/schools/participants_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      if mentor && @school.mentors.exclude?(mentor)
+        raise Pundit::NotAuthorizedError, "No access to this mentor"
```
BroiSatse, line 36: Could this be rather validation on the form?

spec/requests/schools/participants_request_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+  end
+
+  describe "PUT /schools/cohorts/:start_year/participants/:id" do
+    it "updates mentor" do
+      params = { participant_form: { mentor_id: mentor_user_2.id } }
+      put "/schools/cohorts/#{cohort.start_year}/participants/#{ect_user.id}", params: params
+
+      expect(response).to redirect_to(schools_cohort_participant_path(cohort.start_year, ect_user))
+      expect(flash[:success][:title]).to eq("Success")
+      # expect(ect_user.early_career_teacher_profile.mentor).to eq(mentor_user_2)
```
kicferk1, line 102: ummmm, does that not work?

app/controllers/schools/participants_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+private
+
+  def set_school_and_cohort
+    @school = current_user.induction_coordinator_profile.schools.first
+    @cohort = Cohort.find_by(start_year: params[:cohort_id])
+  end
+
+  def set_participant
+    @participant = User.is_participant.in_school(@school.id).find(params[:participant_id] || params[:id])
+  end
```
BroiSatse, line 55: This is something that is bogging me for a while. So we have `User` class which might have  anumber of profiles, which is fine. But why do we need to touch `User` record here at all when all we do is operating on their profile? Why instead not query for the profile? I'm fine with this as is, but it's only becuase the proper split of concerns raises a massive amount of questions :D

app/models/user.rb
```diff
@@ -74,6 +74,7 @@ def school
   scope :for_lead_provider, -> { includes(:lead_provider).joins(:lead_provider) }
   scope :admins, -> { joins(:admin_profile) }
   scope :early_career_teachers, -> { joins(:early_career_teacher_profile).includes(:early_career_teacher_profile) }
+  scope :mentors, -> { joins(:mentor_profile).includes(:mentor_profile) }
```
BroiSatse, line 77: As mentioned earlier - this type of scope is only required because we query User in the controller instead of profile. I'd see a User as a record used for authentication (and email), but apart of that it should be pretty much logic-less.

app/views/schools/participants/edit_mentor.html.erb
```diff
@@ -0,0 +1,22 @@
...
+<% content_for :before_content, govuk_back_link(
+  text: "Back",
+  href: schools_cohort_participant_path(@cohort.start_year, @participant))
+%>
+
+<%= form_for @mentor_form, url: schools_cohort_participant_path(@cohort.start_year, @participant), method: :put do |f| %>
+  <%= f.govuk_radio_buttons_fieldset :mentor_id,
+    legend: { text: "Who is the mentor for #{@participant.full_name}?", tag: 'h1', size: 'xl' },
+    hint: { text: "If you don't know, or haven't added them yet you can do this later." } do %>
+    <% @mentors.each do |mentor, i| %>
```
BroiSatse, line 12: Wouldn't `i` be nil here?

config/routes.rb
BroiSatse, line 192: I think I'd avoid using FeatureFlags in the routes. This file is only loaded once so enabling the feature would require server reset.

BroiSatse (approved):

config/routes.rb
BroiSatse: @kicferk1 - Constraint block is eveluated per each reuqest, while hardcoded condition will be run once on application start. The latter requires application restart when the flag is enabled.

app/controllers/schools/participants_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      mentor_id: @participant.early_career_teacher_profile.mentor&.id,
+    )
+
+    @mentors = @school.mentors.order(:full_name)
+  end
+
+  def update
+    @mentor_form = ParticipantForm.new(participant_form_params)
+
+    # mentor_id can be nil
```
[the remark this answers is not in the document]
callumacrae: It sets the ECT to having no mentor (this hopefully addresses your comment above too)

app/forms/participant_form.rb
```diff
@@ -0,0 +1,24 @@
...
+  validates :email, format: { with: Devise.email_regexp }
+  validate :email_is_not_in_use
+
+  def email_already_taken?
+    if user_id
+      User.where.not(id: user_id).exists?(email: email)
+    else
+      User.exists?(email: email)
+    end
+  end
```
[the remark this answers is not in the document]
callumacrae: probably! it was copied and pasted from an existing form in this repo and i haven't actually touched or tested this code, just the mentor part

app/views/schools/participants/edit_mentor.html.erb
```diff
@@ -0,0 +1,22 @@
...
+      <%= f.govuk_radio_button :mentor_id, mentor.id, label: { text: mentor.full_name }, link_errors: i == 0 %>
```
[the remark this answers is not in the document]
callumacrae: oof

config/routes.rb
callumacrae: originally did it like that, but it meant that the server required a reboot after updating a feature flag otherwise a bunch of pages would error - spoke about it with ethan and went with this approach instead

spec/requests/schools/participants_request_spec.rb
callumacrae: lool thanks for reminding me - i started this test file by copying it from another file which was setting up `cohort` literally just to get the `start_year` from it (which is specified here). was trying to figure out if there was a reason `2021` wasn't just hardcoded in the test instead

app/controllers/schools/cohorts_controller.rb
```diff
@@ -8,5 +8,7 @@ class Schools::CohortsController < Schools::BaseController
 
   def show; end
 
-  def add_participants; end
+  def add_participants
+    redirect_to schools_cohort_participants_path(@cohort.start_year) if FeatureFlag.active?(:induction_tutor_manage_participants)
+  end
```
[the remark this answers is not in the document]
callumacrae: If they've bookmarked the page or clicking on a link someone else sent them, this is indeed necessary!

app/controllers/schools/participants_controller.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class Schools::ParticipantsController < Schools::BaseController
+  skip_after_action :verify_authorized
+  skip_after_action :verify_policy_scoped
```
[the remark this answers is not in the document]
callumacrae: We definitely shouldn't remove `verify_authorized` from the base controller and I definitely need to add authentication to this so that it doesn't need to be skipped 😬 

I'm not sure about the other one, though

config/routes.rb
kicferk1: Ah, good point, we won't do a re-stage when changing feature flags :)

app/controllers/schools/participants_controller.rb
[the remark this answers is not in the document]
kicferk1: I agree, every time I see that `skip_after_action` I think that having them in base controller is a mistake :P

app/controllers/schools/participants_controller.rb
callumacrae: I'm not sure I follow your comment. We need the user, not the profile. Why would we query the profile?

Also, more specifically, "participant" is two profiles, ECT or mentor - my question above is for the more general case.

app/views/schools/participants/edit_mentor.html.erb
kicferk1: Heh, is this a js pattern? In our Ruby code? 😂 
Good old javascript passing index as second argument to `forEach` :P

app/controllers/schools/participants_controller.rb
kicferk1: I am not sure it's a silver bullet to operate on profiles rather than on users. In many cases the two are very closely related - for example, creating a new mentor / ect requires both. 

I guess to set a mentor of an ect would be simpler if we gave the induction coordinator the list of mentor profiles to select from (cause that's what is happening under the surface anyway), but I am not convinced the same is true in general.

In this particular case, I think we actively want to handle ects and mentors at the same time - we want to display a list of both ordered by full name, we want to show their details on a very similar page, so it kind of makes sense to me to operate on `User` model. And I think that pattern, of us operating on user model when we want participants is going to be quite prevalent across the app.

app/forms/participant_form.rb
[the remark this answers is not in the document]
callumacrae: It isn't private because there's a separate page for when an email address is already in use: https://github.com/DFE-Digital/early-careers-framework/blob/induction-tutor-manage-participants/app/controllers/admin/schools/induction_coordinators_controller.rb#L43

It looks like this flow is going to follow a similar pattern, so I'll keep it public.

app/views/schools/participants/edit_mentor.html.erb
callumacrae: at least i didn't write `===` anywhere in this PR!

app/controllers/schools/participants_controller.rb
BroiSatse: As mentioned, there are a lot of questions here and a massive potential for a domain modal collapse. For example, we are now allowing updating the email address for a participant/mentor. This makes perfect sense, but in fact, we are updating an underlying user. There's already a discussion about a single user taking multiple roles, so we are at risk for an induction coordinator being able to change an email of another induction coordinator or even, potentially, an admin user? I'd say IC should only be able to edit p[articipant data, and User record, while related to participant, is not a participant data.

I don't think I am able to explain what I am thinking here as it is not 100% clarified in my head neither - I'm just saying that the  data model we're using around User does not translate onto the domain very well and it will, eventually, cause some pain. But it is definitely out of the scope of this PR :D

> creating a new mentor / ect requires both

Absolutely, creating a new mentor is a domain operation and it is absolutely fine for domain operation to create multiple domain models. What about all the cases when we create participant that does not need to log in? Will we create a third type of participant then?

> we want to display a list of both ordered by full name, we want to show their details on a very similar page

This might be flagging that we are just looking at a single Participant record between a school cohort and a profile. In any case, we should not drive the design of the domain by the required UX

app/forms/participant_mentor_form.rb
```diff
@@ -0,0 +1,29 @@
...
+class ParticipantMentorForm
+  include ActiveModel::Model
+
+  attr_accessor :mentor_id, :school_id, :user_id
+
+  validate :mentor_exists
+
+  def mentor
+    mentor_id != "later" && User.find(mentor_id)
+  end
```
BroiSatse, line 12: Minor but could we make it return `nil` instead of `false`?

app/forms/participant_mentor_form.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def school
+    School.find(school_id)
+  end
```
BroiSatse, line 16: Should it be memoized?

app/policies/participant_policy.rb
```diff
@@ -1,11 +1,26 @@
 # frozen_string_literal: true
 
-class ParticipantPolicy < ApplicationPolicy
+class ParticipantPolicy < UserPolicy
```
BroiSatse, line 3: So we have ParticipantPolicy but no Participant... We'll need to revisit that later...

app/policies/participant_policy.rb
```diff
@@ -1,11 +1,26 @@
...
   def index?
-    admin_only
+    return true if admin_only
+    return @record.all? { |participant| user_can_access?(participant) } if user.induction_coordinator?
   end
```
BroiSatse, line 7: Ah, the beauty of Pundit. So we authorize `index` by passing collection of records as a record and then for each of this record we're loading its school (so we must rely on the policy internals being leaked outside of the policy) and we reject the entire view when a single record is wrong?

Again, this is not something to be addresses here (expecially given a massive amount of PR pending on this one), but  ideally we would not perform per-record authorisation here and make use of a Scope object inside the controller.

BroiSatse: On another note - could we have `return false unless user.induction_coordinator?` instead

BroiSatse (approved):

kicferk1 (commented):
Are the percy changes random?

app/policies/participant_policy.rb
callumacrae: To be fair I have absolutely no idea if this is the best way to do this - the docs seem very unclear on what to do with indexes and I couldn't find any existing examples in our codebase

callumacrae:
> Are the percy changes random?

Yes, but not as a result of this PR - will fix it in another PR

app/policies/participant_policy.rb
kicferk1: For a list of admins we do
```
      def index
        @administrators = policy_scope(User).admins.order(:full_name)
      end
```
which could probably work in here too with some tweaking.

In any case, I think we should go with something simple that works here - and worry about improving it in a separate PR that doesn't block further work :)

PR #464, opened 2021-06-03
Cpdrp 344 participant data api spike

ethanmills:
### Review Checks


app/controllers/api/v1/participants_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+        end
+      end
+
+    private
+
+      def ensure_lead_provider
+        head :forbidden unless current_user.class == LeadProvider
+      end
+
+      def to_csv(hash)
```
jamesrose, line 26: We'll have a better time with `CSV.generate` for the polished PR.

swagger/v1/api_spec.json
```diff
@@ -213,6 +213,205 @@
...
+            "example": {
+              "updated_since": "2020-11-13T11:21:55Z"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Collection of participants.",
+            "content": {
+              "application/json": {
```
ethanmills, line 402: This is wrong (in other places also) but I can't figure out how to set it. It still works in the interactive usage

db/seeds/sandbox_data.rb
```diff
@@ -15,14 +15,17 @@
...
-      user = User.create!(full_name: Faker::Name.name, email: Faker::Internet.email)
+      mentor = User.create!(full_name: Faker::Name.name, email: Faker::Internet.email)
+      mentor_profile = MentorProfile.create!(user: mentor, school: school, cohort: cohort)
+      uuids << mentor.id
 
-      EarlyCareerTeacherProfile.create!(user: user, school: school, cohort: cohort)
-      uuids << user.id
+      ect = User.create!(full_name: Faker::Name.name, email: Faker::Internet.email)
+      EarlyCareerTeacherProfile.create!(user: ect, school: school, cohort: cohort, mentor_profile: mentor_profile)
+      uuids << ect.id
```
kicferk1, line 28: This won't actually run, right? Cause `existing_user_uuids` is not gonna be empty, cause we don't wipe the sandbox db. @timabell thoughts on that?

spec/docs/participants_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+                      example: "2021-05-13T11:21:55Z",
+                    },
+                  },
+                },
+                type: :object,
+                style: :deepObject,
+                explode: true,
+                required: false,
+                description: "Refine participants to return.",
+                example: { updated_since: "2020-11-13T11:21:55Z" }
```
kicferk1, line 36: It would be really nice to find some way of reusing this stuff, although definitely not on this PR.

ethanmills: Yeah, totally agree. I'd also like to commonise the pagination to the API controller, but will do that separately

app/controllers/api/v1/participants_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+        return "" if hash[:data].empty?
+
+        headers = %w[id]
+        attributes = hash[:data].first[:attributes].keys
```
jamesrose, line 36: It feels a bit messy having to do this despite having _a_ serializer, but I can't think of an alternative. Can you? 🤷

app/serializers/participant_serializer.rb
```diff
@@ -0,0 +1,38 @@
...
+class ParticipantSerializer
+  include JSONAPI::Serializer
+
+  set_id :id
+  attributes :email, :full_name
+
+  attributes :mentor_id do |user|
+    if user.early_career_teacher?
+      user.early_career_teacher_profile.mentor&.id
+    end
```
jamesrose, line 12:
```suggestion
    user&.early_career_teacher_profile&.mentor&.id
```

The condition essentially does the same thing.

jamesrose (approved):

app/controllers/api/v1/participants_controller.rb
jamesrose: Like, I feel like we should just be able to `to_csv` something.

ethanmills: I had a quick look & think, but couldn't come up with anything. Reluctant to spend too much time on this given we're hoping to rip this out soon

kicferk1 (approved):

swagger/v1/api_spec.json
kicferk1: It's a bit unfortunate, but I am tempted to say that figuring out why that's happening might be worth more trouble than we want to sink into it at the moment :/ Can we have a tech debt ticket for it?

PR #466, opened 2021-06-04
Add mailer for ministerial letter

ethanmills:
### Context
I've been asked to send lots of emails to an unspecified group of schools on Monday. Sending the chasers was pretty painful, and would have failed if my ssh connection dropped. This aims to improve that.

- Add everything to send ministerial letter emails
- host ministerial letter and leaflet
- Add more worker instances (process things faster, stop a long running task from blocking)
- Bump review app memory to 1GB, we were hitting 100% utilisation

### Testing
Tested the change to the worker app with the review app, then reverted. Manually tested the emails

### Review Checks


kicferk1 (approved):
So, we move to scheduling email sending via console, but the sending of those emails will be done by workers without you needing to ssh into the container?

ethanmills:
> So, we move to scheduling email sending via console, but the sending of those emails will be done by workers without you needing to ssh into the container?

Yes, that pretty much sums it up. You SSH in and schedule emails to be sent (very fast, maybe 500/second), then a worker actually sends them. That means that your SSH session only needs to last a few seconds, and we get all the retry logic and failure tracking of DelayedJob

PR #469, opened 2021-06-04
CPDTP-188 Record events as a declaration event stream

blackrat:
### Context

We want to start recording the events coming in as a declaration event stream with an updated schema with more data (declaration_type, declaration_date and the lead_provider the request comes from)

Create the new participant_declaration resource
Tie the ParticipantDeclarations controller to this via RecordParticipant adding in some validation of the incoming parameters and enable it to also store the raw event message.
Update responses to show missing required parameters in the json body returned for 422 response types.
Create scopes to allow the events to be used for aggregation for the CalculatorOrchestration to only consider unique early_career_teacher_profiles and "started" declarations.
Update the ParticipantEventAggregator to use this rather than the ParticipationRecord as default recorder
Added in more scopes to enable us to compare dates with declared as well as created_at dates.
Change default_config for InitializeWithConfig to use HashWithIndifferentAccess to enable it to work better with params as well as hash object.
Update swagger to enable api documentation to reflect the new parameters and returns.

### Guidance to review

Plan is to switch to using the events as the storage. There's not a great deal of processing on them yet, but the calculation aggregator should be switchable between the event and state based approaches. The state machine is currently not used, but this doesn't preclude it being called as a post-validation optimisation since a call from the RecordParticipant (post creation) will optionally enable it again.

### Testing

Most of the endpoints are either API or runnable from Rake tasks or the Ruby console.

### Review Checks

You can only use this with the API, so the only visible changes will be updated swagger documentation.


app/models/concerns/participant_declaration_event_recorder.rb
```diff
@@ -0,0 +1,18 @@
...
+  module ParticipantDeclarationEventRecorder
+    extend ActiveSupport::Concern
+    included do
+      scope :unique_early_career_teacher_profile_id, -> { select(:early_career_teacher_profile_id).distinct }
+      scope :active, -> { where(declaration_type: "Start").order(declaration_date: "desc").unique_early_career_teacher_profile_id }
+      scope :declared_as_between, ->(start_date, end_date) { where(declaration_date: start_date..end_date) }
+      scope :submitted_between, ->(start_date, end_date) { where(created_at: start_date..end_date) }
+    end
+  end
+end
```
kicferk1, line 18: What do we get out of making it a concern? I can't think of it being reused, and `ParticipantDeclaration` doesn't really strike me as a long file in need of splitting,

app/models/participant_declaration.rb
```diff
@@ -0,0 +1,12 @@
...
+
+class ParticipantDeclaration < ApplicationRecord
+  include Concerns::ParticipantDeclarationEventRecorder
+  belongs_to :lead_provider
+  belongs_to :early_career_teacher_profile
+
+  scope :for_lead_provider, ->(lead_provider) { where(lead_provider: lead_provider) }
+  scope :active_for_lead_provider, ->(lead_provider) { active.for_lead_provider(lead_provider) }
+  scope :count_active_for_lead_provider, ->(lead_provider) { active_for_lead_provider(lead_provider).count }
+  scope :count_active_for_lead_provider_between, ->(lead_provider, start_date, end_date) { declared_as_between(start_date, end_date).count_active_for_lead_provider(lead_provider) }
```
kicferk1, line 11: Are we sure we need all of them to be individual scopes? If some are just stepping stones for further scopes, and won't be used outside of the model, I would be tempted to combine them into fewer scopes - at least until we need the intermediate scopes.

app/services/record_participant_event.rb
```diff
@@ -0,0 +1,50 @@
...
+      %i[participant_id lead_provider declaration_type declaration_date raw_event]
+    end
+  end
+
+  def call
+    (set_config_ect_profile || :not_found) &&
+      (create_record || :unprocessable_entity) &&
+      (invalid_provider || :unprocessable_entity) &&
+      :no_content
+  end
```
kicferk1, line 16: What am I looking at here? If we have no profile, what exactly is going to happen?

Can we make use of some early returns here? And explicit if statements?

lib/initialize_with_config.rb
```diff
@@ -64,7 +64,7 @@ def call(*)
   end
 
   def default_config
-    {}
+    HashWithIndifferentAccess.new
```
kicferk1, line 67: What's the difference between the two?

spec/services/participant_event_aggregator_service_spec.rb
```diff
@@ -17,25 +17,32 @@ def count_something_else(*)
...
+
+      it "can be injected with a different recorder" do
+        expect(described_class.call({ recorder: DummyRecorder, lead_provider: lead_provider })).to eq(100)
+      end
 
-    it "can be injected with a different recorder and different scope" do
-      expect(described_class.call({ recorder: DummyRecorder, start: :count_something_else, lead_provider: lead_provider })).to eq(40)
+      it "can be injected with a different recorder and different scope" do
+        expect(described_class.call({ recorder: DummyRecorder, start: :count_something_else, lead_provider: lead_provider })).to eq(40)
+      end
```
kicferk1, line 45: I am missing some context here.
I think `.call` should be in a `describe` block, and it should have one `context` block for each type of recorder that we can give it. I struggle to understand the difference between the first two tests from test descriptions.

app/models/concerns/participant_declaration_event_recorder.rb
blackrat: Reason is that if we swap between the two different mechanisms for either event or cached state holding the "active" in two different concerns means that there is less to change.

lib/initialize_with_config.rb
blackrat: HashWithIndifferentAccess doesn't care if the key is a symbol or a string. {} does.

app/models/participant_declaration.rb
blackrat: It makes it easier to double check things in the console and we're going to have different queries. (I actually have used all of them separately in the console to check on different scenarios. As we've not pinned down everything yet, keeping them distinct make modelling and checking edge cases a lot easier.)

app/models/concerns/participant_declaration_event_recorder.rb
kicferk1: "If we swap"? Are we not sure? Do we plan to have both mechanisms in our codebase for extended period of time?

app/services/record_participant_event.rb
blackrat: Are you having problems associating this logical flow without branches?
is:

```
return :not_found unless set_config_ect_profile
return :unprocessable_entity unless create_record
return :unprocessable_entity unless invalid_provider
:no_content`

```
really clearer?

(I deliberately added in the `(invalid_provider || :unprocessable_entity) &&` as a separate line in a live session with Matt to make sure that we could map the required return states to the condition that generated the error.

kicferk1: I find it much easier to read the second version.

In the second version, is the order correct? Should we check for the invalid provider before we try to create a record?

kicferk1: also
```
return :unprocessable_entity unless invalid_provider
```
reads a bit surprising - I would expect
```
return :unprocessable_entity if invalid_provider
```
Cause if the provider is not invalid, then why would the entity not be processable?

app/models/concerns/participant_declaration_event_recorder.rb
blackrat: Yes. We'll probably end up using both. There are use cases where one is more appropriate as cached information and another where we need all information as it is.

app/services/record_participant_event.rb
blackrat: > In the second version, is the order correct? Should we check for the invalid provider before we try to create a record?

Yes. The logic was that we would want to create the record anyway, but flag that it was for a different provider.

blackrat: `invalid_provider` was incorrect. Fixed.

app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -3,19 +3,37 @@
...
+        required_params - config.keys
+      end
+
+      def check_config(config)
+        missing_params(config).empty?
+      end
 
-        head :no_content
+      def declaration_params
+        params.permit(:participant_id, :declaration_date, :declaration_type)
```
jamesrose, line 36: Did you intend to use `permitted_params` and `required_params` here?

spec/docs/participant_declarations_spec.rb
```diff
@@ -20,37 +21,104 @@
           "schema": {
             "type": "object",
             "properties": {
-              "name": "id",
-              "type": "string",
+              "participant_id": {
+                "type": "string",
+              },
+              "declaration_type": {
+                "enum": %w[Start],
```
jamesrose, line 28: Should these be capitalised? 😬

app/controllers/api/v1/participant_declarations_controller.rb
blackrat: Enabling StrongParameters didn't work as expected, and the incoming API wasn't wrapping in such a way that `required_params` could be used. I may be missing something from that, so if you can enlighten....

spec/docs/participant_declarations_spec.rb
blackrat: No they shouldn't. We also changed the name during the session on Friday. They need to be updated.

app/services/record_participant_event.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class RecordParticipantEvent
```
kicferk1, line 3: Do we have unit tests for this file? If not, could we?

spec/services/participant_event_aggregator_service_spec.rb
blackrat: OK. Done as per offline conversation.

app/services/record_participant_event.rb
blackrat: Integration tests only for this service. It doesn't really stand well in isolation without lots of supporting database structure.

kicferk1 (approved):

timabell:
I'd like a more meaningful PR title please :pray:

PR #473, opened 2021-06-07
Remove randomness from percy snapshots

callumacrae:
### Context

it was annoying and bad

### Guidance to review

percy will error in this PR, but once merged to develop should start working!

you can't

### DIFF
--- a/spec/cypress/app_commands/scenarios/admin/schools.rb
+++ b/spec/cypress/app_commands/scenarios/admin/schools.rb
@@ -3,7 +3,9 @@
 school = FactoryBot.create(:school, name: "Include this school", urn: 123_456)
 local_authority = FactoryBot.create(:local_authority)
 SchoolLocalAuthority.create!(school: school, local_authority: local_authority, start_year: 2021)
-FactoryBot.create(:user, :induction_coordinator, full_name: "Sarah Smith", email: "sarah.smith@example.com", schools: [school])
+coordinator = FactoryBot.create(:user, :induction_coordinator, full_name: "Sarah Smith", email: "sarah.smith@example.com")
+coordinator.induction_coordinator_profile.schools.first.destroy!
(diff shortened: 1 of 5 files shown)


spec/cypress/app_commands/scenarios/admin/schools.rb
```diff
@@ -3,7 +3,9 @@
 school = FactoryBot.create(:school, name: "Include this school", urn: 123_456)
 local_authority = FactoryBot.create(:local_authority)
 SchoolLocalAuthority.create!(school: school, local_authority: local_authority, start_year: 2021)
-FactoryBot.create(:user, :induction_coordinator, full_name: "Sarah Smith", email: "sarah.smith@example.com", schools: [school])
```
callumacrae, line 6: This was an interesting one - creating a user with an induction_coordinator profile would create a school which had a constant name but a random URN.

kicferk1 (approved):
I would ask "is that all?", but I guess there is no easy way of finding out until we merge this, right?

callumacrae:
> I would ask "is that all?", but I guess there is no easy way of finding out until we merge this, right?

See percy on this PR - there's only two snapshots containing randomness! So yes, this should be all :)

spec/cypress/integration/admin/SchoolManagment.feature
```diff
@@ -26,8 +26,9 @@ Feature: Admin user managing schools
     And the page should be accessible
     And percy should be sent snapshot
 
+  @focus
   Scenario: Viewing a list of schools
-    Then the table should have 20 rows
+    Then the table should have 12 rows
```
callumacrae, line 31: This isn't directly related to this PR, but I was concerned that it was testing for 20 rows when there were actually 23 schools and it was just testing the pagination.

PR #474, opened 2021-06-07
First step of adding participant journey

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-356

This PR contains the seed for the entire InductorCoordinator journey of adding a new participant.

### DIFF
--- a/app/controllers/schools/add_participants/base_controller.rb
+++ b/app/controllers/schools/add_participants/base_controller.rb
@@ -7,8 +7,6 @@ module Schools
       FORM_PARAM_KEY = :schools_add_participant_form
 
       skip_after_action :verify_authorized
-      skip_after_action :verify_policy_scoped
-
       before_action :set_school_cohort
 
       helper_method :add_participant_form
--- a/app/controllers/schools/add_participants/base_controller.rb
+++ b/app/controllers/schools/add_participants/base_controller.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module Schools
+  module AddParticipants
+    class BaseController < ::Schools::BaseController
+      FORM_SESSION_KEY = :add_participant_form
+      FORM_PARAM_KEY = :schools_add_participant_form
+
+      skip_after_action :verify_authorized
+      skip_after_action :verify_policy_scoped
+
+      before_action :set_school_cohort
+
+      helper_method :add_participant_form
+
+      def start
+        session.delete(FORM_SESSION_KEY)
+        redirect_to schools_cohort_add_participants_type_path
+      end
+
+    private
+
+      def add_participant_form
+        return @add_participant_form if defined?(@add_participant_form)
+
+        @add_participant_form = AddParticipantForm.new(session[FORM_SESSION_KEY])
+        @add_participant_form.assign_attributes(params.require(FORM_PARAM_KEY).permit!) if params[FORM_PARAM_KEY]
+
+        @add_participant_form
+      end
+
+      def store_form_in_session
+        session[FORM_SESSION_KEY] = add_participant_form.attributes
+      end
+    end
+  end
+end
--- a/app/controllers/schools/add_participants/details_controller.rb
+++ b/app/controllers/schools/add_participants/details_controller.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module Schools::AddParticipants
+  class DetailsController < BaseController
+    def show; end
+  end
+end
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Schools
+  class AddParticipantForm
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+    include ActiveModel::Serialization
+
+    TYPE_OPTIONS = {
+      ect: "Early Career Teacher",
+      mentor: "Mentor",
+    }.freeze
+
+    attribute :type
+
+    validates :type,
+              on: :type,
+              presence: { message: "Please select type of the new participant" },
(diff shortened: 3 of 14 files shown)


spec/services/feature_flag_spec.rb
```diff
@@ -3,6 +3,10 @@
 require "rails_helper"
 
 RSpec.describe FeatureFlag do
+  before do
+    stub_const("FeatureFlag::FEATURES", { add_participants: described_class.new(name: :add_participants) })
```
jamesrose, line 7: 👍

jamesrose (approved):

app/controllers/schools/add_participants/details_controller.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Schools::AddParticipants
+  class DetailsController < BaseController
+    def show; end
+  end
```
kicferk1, line 6: Is it a good idea to inherit from a controller that has a `start` action? Will that action get inherited? 

Same for `TypesController`.

BroiSatse: Good point. It is definitely inherited, but without the route it will stay unused. As this is the first draft of this feature I think I'd get it as it is and will clean it up a bit later (most likely separate start / finalize, both inheiriting completley abstract base?)

app/controllers/schools/add_participants/base_controller.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Schools
+  module AddParticipants
+    class BaseController < ::Schools::BaseController
+      FORM_SESSION_KEY = :add_participant_form
+      FORM_PARAM_KEY = :schools_add_participant_form
+
+      skip_after_action :verify_authorized
+      skip_after_action :verify_policy_scoped
```
callumacrae, line 10: These shouldn't be skipped (and you're using `set_school_cohort` so the second one doesn't even need to be)

app/forms/schools/add_participant_form.rb
```diff
@@ -0,0 +1,25 @@
...
+module Schools
+  class AddParticipantForm
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+    include ActiveModel::Serialization
+
+    TYPE_OPTIONS = {
+      ect: "Early Career Teacher",
+      mentor: "Mentor",
+    }.freeze
```
callumacrae, line 12: Possible duplicate of `user.user_description` although I'm not 100% sure what it's actually doing

app/controllers/schools/add_participants/details_controller.rb
kicferk1: Uhh, this reads like a multi-page creation flow with a shared session store. Is that what it is?

If so, can we have a single controller for it? It will mean we can keep the convention of 'the only place that changes the session with a key is a controller that defines this key' - and well, `TypesController` and `DetailsController` are pretty tiny as is.

app/forms/schools/add_participant_form.rb
BroiSatse: `User#user_description` is an abomination and should be killed with fire! :D

But actually it returns much more options than what we need here, including admin, or lead provider users (have i mentioned that User knows waaaaay too much about the application already... :P)

app/controllers/schools/add_participants/base_controller.rb
BroiSatse: I think we can have a longer ping-pong like discussion about those :D IMO, they not only should be skipped, but they should not exist at all. Authorisation is way too complex and important part of the system to be guarded by those terribly hacked methods. As mentioned earlier, they are very easy to be walked around. I prefer no safety net over a safety net that doesn't work.

app/forms/schools/add_participant_form.rb
callumacrae: My main concern around this is that defining user definitions from keys in multiple places leads to possible variance - in this case it's just the capitalisation, in the future it could be more.

app/controllers/schools/add_participants/base_controller.rb
callumacrae: My actual point was that there is no authorization being done in this controller.

app/controllers/schools/add_participants/details_controller.rb
BroiSatse: @kicferk1 - Yup, actually got annoyed by this as well yesterday and this is now corrected in the next PR.

app/forms/schools/add_participant_form.rb
BroiSatse: So this constant does not define possible user types but rather possible options on the form. Ideally, I'd had those types translated in a single translation scope (and that will be dona later if time allows).

app/controllers/schools/add_participants/details_controller.rb
kicferk1: Alright, sooo, what exactly should I check in this PR? 😂 

What's the split between this and next one exactly?

kicferk1 (approved):
Seeing how the controller inheritance will get refactored, I take it that this PR is mostly about adding a form, some views, and tests. And those are fine by me.

app/controllers/schools/add_participants/details_controller.rb
BroiSatse: @kicferk1 Yup, that's exactly my question - how do we divide the comments between multiple PRs? After all, the feature is still in progress. I wonder if maybe it will be better to have one larger PR, but with well defined, coherent commits so it can be reviewed commit by commit instead?

kicferk1: I think in the future, I would prefer a single PR that tries to be pretty minimal in how it attempts to implement the feature. If it means we need to add a ticket of tech debt here or there, with a todo comment in code, so be it.

PR #476, opened 2021-06-07
CPDRP-343: Track CSV errors in GA data layer

ethanmills:
### Context
CPDRP-343
Our data analyst needs information on the errors displayed to users to measure the performance of new features.

- Track number and type of errors on CSV rows
- Track CSV file errors

### Review Checks

Sign in as lead-provider@example.com, upload a CSV with errors, check the datalayer using dev tools

### DIFF
--- a/app/views/lead_providers/report_schools/csv/show.html.erb
+++ b/app/views/lead_providers/report_schools/csv/show.html.erb
@@ -1,5 +1,9 @@
 <% content_for :title, "Upload partnership csv" %>
-<% content_for :before_content, govuk_back_link( text: "Back", href: lead_providers_report_schools_delivery_partner_path ) %>
+<% content_for :before_content, govuk_back_link(text: "Back", href: lead_providers_report_schools_delivery_partner_path) %>
+
+<% data_layer.add(
+  csv_file_errors: @partnership_csv_upload.errors.map { |error| error.type.parameterize.underscore }
+) if @partnership_csv_upload.errors.any? %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
@@ -21,11 +25,11 @@
           </h2>
           <div class="govuk-error-summary__body">
             <ul class="govuk-list govuk-error-summary__list">
-            <% @partnership_csv_upload.errors[:base].each do |error| %>
-              <li>
-                <a href="#csv-error"><%= error %></a>
-              </li>
-            <% end %>
+              <% @partnership_csv_upload.errors[:base].each do |error| %>
+                <li>
+                  <a href="#csv-error"><%= error %></a>
+                </li>
+              <% end %>
             </ul>
           </div>
         </div>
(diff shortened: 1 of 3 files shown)


app/views/lead_providers/report_schools/csv/show.html.erb
```diff
@@ -1,5 +1,9 @@
 <% content_for :title, "Upload partnership csv" %>
-<% content_for :before_content, govuk_back_link( text: "Back", href: lead_providers_report_schools_delivery_partner_path ) %>
+<% content_for :before_content, govuk_back_link(text: "Back", href: lead_providers_report_schools_delivery_partner_path) %>
+
+<% data_layer.add(
+  csv_file_errors: @partnership_csv_upload.errors.map { |error| error.type.parameterize.underscore }
+) if @partnership_csv_upload.errors.any? %>
```
kicferk1, line 6: Are we sure we need this? I thought it would include errors like 'the file is too large', and the actual errors would be on the other page.

If we care about validation errors, should we, hmm, make the error summary do it? And maybe use the error summary here?

kicferk1: Hmm, the error summary comes from DfE controlled formbuilder gem (right?). And the maintainers for that gem are pretty responsive. We could try to open a PR to do it.

Alternatively we could monkey patch `GOVUKDesignSystemFormBuilder::Elements::ErrorSummary`, 
```
      def html
        return nil unless object_has_errors?

        tag.div(class: summary_class, **summary_options) do
          safe_join([title, summary])
        end
      end
```
seems like the place to add the html for the data layer, namely the line with `safe_join`.

Thoughts?

ethanmills: That doesn't feel right to me. That component is all about presentation of errors, it seems odd that we would modify that component for a one-off google analytics tracking thing

ethanmills: If anything, I would create a new component

ethanmills: Looking at it, this doesn't generate any markup. Having a component for it feels wrong, it should stay a helper. I can add another layer of helpers to abstract it away a little more, or leave it as is

kicferk1: A new component will work, I guess, but it will also need to be added to every single form, which sounds annoying. Error reporting doesn't seem like a one off to me, although I guess it might be rare enough that we don't want it in a govuk design inspired component.

kicferk1: > Looking at it, this doesn't generate any markup. Having a component for it feels wrong, it should stay a helper. I can add another layer of helpers to abstract it away a little more, or leave it as is

Ah, that makes sense

ethanmills: I think this one is a bit of a rarity actually. In future, Chris will be able to pick out those errors when he has access to BigQuery. Right now though, just using GA, he needs info pushed to the datalayer. Everything else can wait, but people are really keen to see how this one performs

kicferk1 (approved):

PR #480, opened 2021-06-08
Added migration to replace integer id with uuid

tonyheadford:
### Context
[Jira ticket](https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-393)
Fix issue with integer primary key

The `induction_coordinator_profiles_schools` is a join table that had an `id` added so that paper trail could record it. We are using a gem `ar-uuid` which is supposed to automagically make primary keys uuids but for reasons unknown it didn't pick up the `primary_key` statement in that migration and so it was created as a `bigint`.

Worth noting is that the `versions` table managed by paper trail was configured to expect `item_id` to be a `uuid` but this also has a default to automatically generate a uuid so any existing paper trails for `induction_coordinator_profiles_schools` will not be easily reconciled.

### Guidance to review
Look at the migration steps

Adds a migration to replace the current integer `id` with a `uuid` one.

### Review Checks

You can't really

### DIFF
--- a/db/migrate/20210608084207_change_primary_key_to_uuid_on_induction_coordinator_profiles_schools.rb
+++ b/db/migrate/20210608084207_change_primary_key_to_uuid_on_induction_coordinator_profiles_schools.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class ChangePrimaryKeyToUuidOnInductionCoordinatorProfilesSchools < ActiveRecord::Migration[6.1]
   def up
     add_column :induction_coordinator_profiles_schools, :uuid_id, :uuid, default: "gen_random_uuid()", null: false
--- a/db/migrate/20210608084207_change_primary_key_to_uuid_on_induction_coordinator_profiles_schools.rb
+++ b/db/migrate/20210608084207_change_primary_key_to_uuid_on_induction_coordinator_profiles_schools.rb
@@ -1,0 +1,23 @@
+class ChangePrimaryKeyToUuidOnInductionCoordinatorProfilesSchools < ActiveRecord::Migration[6.1]
+  def up
+    add_column :induction_coordinator_profiles_schools, :uuid_id, :uuid, default: "gen_random_uuid()", null: false
+    rename_column :induction_coordinator_profiles_schools, :id, :integer_id
+    rename_column :induction_coordinator_profiles_schools, :uuid_id, :id
+    execute "ALTER TABLE induction_coordinator_profiles_schools DROP CONSTRAINT induction_coordinator_profiles_schools_pkey;"
+    execute "ALTER TABLE induction_coordinator_profiles_schools ADD PRIMARY KEY (id);"
+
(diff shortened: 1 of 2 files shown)


db/migrate/20210608084207_change_primary_key_to_uuid_on_induction_coordinator_profiles_schools.rb
```diff
@@ -0,0 +1,23 @@
...
+  def up
+    add_column :induction_coordinator_profiles_schools, :uuid_id, :uuid, default: "gen_random_uuid()", null: false
+    rename_column :induction_coordinator_profiles_schools, :id, :integer_id
+    rename_column :induction_coordinator_profiles_schools, :uuid_id, :id
+    execute "ALTER TABLE induction_coordinator_profiles_schools DROP CONSTRAINT induction_coordinator_profiles_schools_pkey;"
+    execute "ALTER TABLE induction_coordinator_profiles_schools ADD PRIMARY KEY (id);"
+
+    remove_column :induction_coordinator_profiles_schools, :integer_id
+    execute "DROP SEQUENCE IF EXISTS induction_coordinator_profiles_schools_id_seq;"
+  end
```
kicferk1, line 13: So just so I'm sure I understand, I'm gonna write the steps in english.

We add a new column for our uuids.
We rename old integer id column to something else.
We name uuid column 'id'.

We throw away old primary key constraint.
We add a new primary key constraint on our uuid column.

We delete the integer id column and any sequencing related to it.


Is this equivalent to
Removing primary key from id (integer)
Removing id (integer)
Removing any sequencing relating to id (integer)

Adding uuid `id` column
Adding primary key constraint to the newly added column?

As in, why do we bother with renaming?

tonyheadford: Sounds feasible, but I found this solution online and it works. I tried a few different options and not all worked as expected but didn't try just doing what you suggest.

tonyheadford: Actually what happened was I intended to keep `integer_id` and try and reconcile the paper trails `versions` entries after the migration so this was written with that in mind. Once I found out that all the `item_id`s had been replaced with random uuids anyway for this table I removed the `integer_id` column in the migration but didn't revisit anything else.

ethanmills (approved):

db/migrate/20210608084207_change_primary_key_to_uuid_on_induction_coordinator_profiles_schools.rb
kicferk1: Right, I guess this can go in, cause at worst it's a bit cumbersome. The perfectionist in me would prefer to axe renaming, but if we want the fix in quick, just merge it :)

kicferk1 (approved):

PR #482, opened 2021-06-08
CPDRP-418: Notify email validation

ethanmills:
### Context
CPDRP-418
Users have been entering invalid emails and are presented with an error
page instead of a validation message. This is because our validation
rules do not match Notify's. This is an attempt to bring them in line.

I asked the Notify team if they'd be interested in adding the logic to the client SDKs, but they said no: https://ukgovernmentdigital.slack.com/archives/C0E1ADVPC/p1623151272077400

- Add NotifyEmailValidator
- Use the validator where we accept emails

### Testing
Copied the tests from Notify, with some caveats

Try nominating someone with email "a@b,com"

### DIFF
--- a/app/forms/nominate_induction_tutor_form.rb
+++ b/app/forms/nominate_induction_tutor_form.rb
@@ -6,7 +6,7 @@ class NominateInductionTutorForm
   attr_accessor :full_name, :email, :token, :school_id, :user_id
 
   validates :full_name, presence: true
-  validates :email, presence: true, format: { with: Devise.email_regexp }
+  validates :email, presence: true, notify_email: true
   validate :email_is_not_in_use
 
   def school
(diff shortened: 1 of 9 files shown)


app/forms/nominate_induction_tutor_form.rb
```diff
@@ -6,7 +6,7 @@ class NominateInductionTutorForm
   attr_accessor :full_name, :email, :token, :school_id, :user_id
 
   validates :full_name, presence: true
-  validates :email, presence: true, format: { with: Devise.email_regexp }
+  validates :email, presence: true, notify_email: true
```
kicferk1, line 9: How does this work? How does it get wired and told to use `NotifyEmailValidator`?

ethanmills: Rails magic of course! It's just based on the name, and that it inherits from `ActiveModel::EachValidator`

ethanmills: Your favourite!

kicferk1: Aaaaarghhhhhhh, my head has just blown up.

kicferk1 (approved):

PR #483, opened 2021-06-08
Remove another bit of randomness from percy

callumacrae:
This one snuck through because I only tested on the same day!

You can't

### DIFF
--- a/spec/cypress/app_commands/scenarios/confirm_cip.rb
+++ b/spec/cypress/app_commands/scenarios/confirm_cip.rb
@@ -25,6 +25,6 @@ FactoryBot.create(
   cohort: cohort,
   delivery_partner: delivery_partner,
   lead_provider: lead_provider,
-  created_at: 2.days.ago,
-  challenge_deadline: 6.days.from_now,
+  created_at: Date.new(2021, 6, 7),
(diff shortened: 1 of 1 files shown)


spec/cypress/app_commands/scenarios/confirm_cip.rb
```diff
@@ -25,6 +25,6 @@
   cohort: cohort,
   delivery_partner: delivery_partner,
   lead_provider: lead_provider,
-  created_at: 2.days.ago,
-  challenge_deadline: 6.days.from_now,
+  created_at: Date.new(2021, 6, 7),
+  challenge_deadline: Date.new(2025, 6, 7),
```
kicferk1, line 29: Is that because it needs to be in the future? 😂 

Maybe let's add another 100 years to that or something.

callumacrae: Ideally we'd just be able to hardcode the current date so that we could carry on doing `6.days.from_now` and that date wouldn't change, but I dunno if that's possible with cypress+rails.

I wanted to keep the date somewhat realistic! I don't think having to update a test every four years is that unreasonable.

kicferk1 (approved):

PR #486, opened 2021-06-09
ApiTokens and endpoints can have private access

asmega:
### Context

- We want certain api endpoint to be private. e.g. for internal use only and not for public consumption
- For example the `DqtRecordsController` api endpoint should only be accessible for internal use

- On `ApiToken` introduce `private_api_access` to determine if a token can access private api endpoints
- On api controllers one can now define `access_scope` to limit api token to a certain subset
- By default all existing endpoints will continue to be accessible to existing tokens
- The `DqtRecordsController` endpoint is now only accessible to tokens marked with `private_api_access` set to `true`
- By default new `NpqRegistrationApiToken` will have access to private api endpoints
- When adding new endpoints if developers do not want to use this feature there should be no extra burden to them

#### Future suggestions

- No changes has been made to the structure of api endpoints though it may be a good idea to separate private endpoints into a different namespace
- Documentation for public/private api endpoints may also want to be separated

### Guidance to review

- Existing api endpoints should continue to work with no breaking changes except `DqtRecordsController`
- Existing api tokens should continue to work with no break changes except `NpqRegistrationApiToken`
- DQT api endpoint should now have limited access
- The mechanism should be fairly flexible should we want add more constraints/limitation to tokens

### Testing

- Test via review app

### Review Checks

- Call endpoint with `curl -I -H "Authorization: Bearer UNHASHED_TOKEN" https://ecf-review-pr-486.london.cloudapps.digital/api/v1/dqt-records/100`
- Should return body-less 404 response
- Create another api token with `EngageAndLearnApiToken.create_with_random_token!`
- Call endpoint with `curl -I -H "Authorization: Bearer UNHASHED_TOKEN" https://ecf-review-pr-486.london.cloudapps.digital/api/v1/dqt-records/100`
- Should get 401

### DIFF
--- a/app/controllers/api/v1/dqt_records_controller.rb
+++ b/app/controllers/api/v1/dqt_records_controller.rb
@@ -14,6 +14,12 @@ module Api
           head :not_found
         end
       end
+
+    private
+
+      def access_scope
+        ApiToken.where(private_api_access: true)
+      end
     end
   end
 end
--- a/app/models/npq_registration_api_token.rb
+++ b/app/models/npq_registration_api_token.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class NpqRegistrationApiToken < ApiToken
+  attribute :private_api_access, default: true
+
   def owner
     "npq_registration_application"
   end
--- a/spec/requests/api/v1/dqt_records_spec.rb
+++ b/spec/requests/api/v1/dqt_records_spec.rb
@@ -64,5 +64,15 @@ RSpec.describe "DQT records api endpoint", type: :request do
         expect(response.status).to eq 401
       end
     end
+
+    context "using token from different scope" do
+      let(:other_token) { EngageAndLearnApiToken.create_with_random_token! }
+
+      it "returns 401 for invalid bearer token" do
+        default_headers[:Authorization] = "Bearer #{other_token}"
+        get "/api/v1/dqt-records/#{trn}"
+        expect(response.status).to eq 401
+      end
+    end
   end
 end
--- a/app/controllers/concerns/api_token_authenticatable.rb
+++ b/app/controllers/concerns/api_token_authenticatable.rb
@@ -10,7 +10,7 @@ module ApiTokenAuthenticatable
 
   def authenticate
     authenticate_or_request_with_http_token do |unhashed_token|
-      @current_api_token = ApiToken.find_by_unhashed_token(unhashed_token)
+      @current_api_token = ApiToken.merge(access_scope).find_by_unhashed_token(unhashed_token)
       if @current_api_token
         @current_api_token.update!(
           last_used_at: Time.zone.now,
@@ -21,5 +21,11 @@ module ApiTokenAuthenticatable
 
   def current_user
     @current_api_token.owner
+  end
+
+private
+
+  def access_scope
+    ApiToken.all
   end
 end
--- a/db/migrate/20210609094438_add_access_booleans_to_token.rb
+++ b/db/migrate/20210609094438_add_access_booleans_to_token.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddAccessBooleansToToken < ActiveRecord::Migration[6.1]
+  def change
+    change_table :api_tokens, bulk: true do |t|
+      t.boolean :private_api_access, default: false
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_06_08_084207) do
+ActiveRecord::Schema.define(version: 2021_06_09_094438) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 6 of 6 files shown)


spec/requests/api/v1/dqt_records_spec.rb
```diff
@@ -64,5 +64,15 @@
         expect(response.status).to eq 401
       end
     end
+
+    context "using token from different scope" do
+      let(:other_token) { EngageAndLearnApiToken.create_with_random_token! }
```
ethanmills, line 69: I find it surprising that E&L is not considered internal

ethanmills (approved):
Looks good, one comment

app/models/npq_registration_api_token.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class NpqRegistrationApiToken < ApiToken
+  attribute :private_api_access, default: true
+
```
kicferk1, line 5: What does this do?

app/controllers/api/v1/dqt_records_controller.rb
```diff
@@ -14,6 +14,12 @@ def show
           head :not_found
         end
       end
+
+    private
+
+      def access_scope
+        ApiToken.where(private_api_access: true)
+      end
```
kicferk1, line 22: I don't know which level of granularity we want really. Is it enough for our use case to have these permissions on the level of STI?

So in here, you could do something like `NpqRegistrationApiToken.all`?

Do we expect some tokens of particular type to have more access than other tokens of particular type? Like, having a live E&L token that can access DQT, and test E&L tokens which can't? Likewise, do we ever expect to give some LP tokens the right to use this endpoint, but not all?

asmega: > So in here, you could do something like NpqRegistrationApiToken.all?

We can do that but then you've coupled the entire controller to one type of token. No other token can be used. Whereas the current implementation doesn't care what token it gets as long as the flag is true.

 > Do we expect some tokens of particular type to have more access than other tokens of particular type? Like, having a live E&L token that can access DQT, and test E&L tokens which can't? Likewise, do we ever expect to give some LP tokens the right to use this endpoint, but not all?

With the mechanism I have suggested you'd be able to set certain tokens access to private apis if you so wished. This gives us full flexibility. If we scope to `NpqRegistrationApiToken.all` you've dug yourself a hole.

kicferk1: > With the mechanism I have suggested you'd be able to set certain tokens access to private apis if you so wished.

Yeah, you would be able to. Would we want to? 

Are we confident that there is an existing use case for it? If not, should we postpone adding the functionality until there is?

app/models/npq_registration_api_token.rb
asmega: I actual only found this earlier today and mind was blown. its documented here https://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html#method-i-attribute

so it looks like they (rails/activerecord) added a mechanism where it allows for fine grain control between AR and the DB for type casting and general handling of serialisation/unserialisation of types.

in this particular case i'm not using any of that but setting the default value of `private_api_access` to `true` for `NpqRegistrationApiToken` as it defaults to false as defined by the database default value on `ApiToken`

kicferk1: So is that only going to impact new npq tokens, or any existing ones too? Mostly a theoretical question, I don't think many of them will survive database wipes.

spec/requests/api/v1/dqt_records_spec.rb
asmega: I don't know the engage and learn api endpoints so not sure if we want it to have private access. if it gets changed it will break this test. I can't actually change to `ApiToken.create_with_random_token!` as I get the following error
```
ActiveRecord::RecordInvalid:
       Validation failed: Lead provider must exist
```
I don't really want to write code to workaround that and would prefer to use a token I know at this moment in time has `private_api_access` set to false

app/controllers/api/v1/dqt_records_controller.rb
asmega: > Yeah, you would be able to. Would we want to?

The answer to that is I don't know. I think none one does. But if we needed to, we can now do that easily.

> Are we confident that there is an existing use case for it? If not, should we postpone adding the functionality until there is?

The suggested change currently meets our need to be able to make certain api endpoints private and for those private apis only be accessible to certain tokens. I don't see any benefit here of postponing.

app/models/npq_registration_api_token.rb
asmega: Existing ones will use whatever is persisted in the database. The new ones will take `true`. from the docs:
```
The default value to use when no value is provided. If this option is not passed, the previous default value (if any) will be used. Otherwise, the default will be nil.
```

app/controllers/api/v1/dqt_records_controller.rb
kicferk1: Fine, I am worried that we will eventually get some concrete requirements that conflict with this setup, but then again, removing a boolean is pretty trivial.

kicferk1 (approved):

tonyheadford (approved):
LGTM

PR #491, opened 2021-06-09
Fix randomness in percy snapshot for admin view school participants

callumacrae:
my bad


kicferk1 (approved):
Other users in this file don't matter? I guess we don't show them at the moment...

PR #494, opened 2021-06-09
Cpdrp 356 ic adding participants personal details

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-356

Second step of the IC add participant journey.

Featuring an abstract MultistepForm implementation - I am quite tempted to extract it to a module at one point so that we could easily reuse it in the future (would also apply to a LeadProvider partnership reporting etc)


app/forms/schools/add_participant_form.rb
```diff
@@ -6,20 +6,55 @@ class AddParticipantForm
...
+
+      completed_steps[completed_steps.index(current_step) - 1]
+    end
+
+    def next_step(step = completed_steps.last)
+      case step
+      when :type then :details
+      else :not_implemented
+      end
+    end
```
BroiSatse, line 50: This method defines multistep form flow - it is very simple but will get more and more complex when we add more steps and possible forks. I **think** it is fine for now, however I am aware it violates open-closed principle. Ideally this would be relaying on polymorphism of the step objects, so it would be nice to revisit it / refactor when the MultistepForm is extracted.

app/forms/schools/add_participant_form.rb
```diff
@@ -6,20 +6,52 @@ class AddParticipantForm
...
+
     def type_options
       TYPE_OPTIONS
     end
+
+    def previous_step(current_step)
+      return completed_steps.last unless current_step.in?(completed_steps)
+
+      completed_steps[completed_steps.index(current_step) - 1]
+    end
```
callumacrae, line 43: There's a few issues with this which I suspect would be uncovered by splitting it out so other forms can reuse it and properly testing it

The main ones I've noticed is that on the first page it just goes to the page you're already on, and if you press back twice you end up in a two page loop

app/controllers/schools/add_participants_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+module Schools
+  class AddParticipantsController < ::Schools::BaseController
+    FORM_SESSION_KEY = :add_participant_form
+    FORM_PARAM_KEY = :schools_add_participant_form
+
+    skip_after_action :verify_authorized
+    before_action :set_school_cohort
+
+    helper_method :add_participant_form
```
kicferk1, line 11: I wonder if it's nicer to do it before relevant actions. I don't like the idea of views creating @ variables shared with controller when they render. 

I don't think it does anything bad here though, just a code style preference.

app/controllers/schools/add_participants_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      session[FORM_SESSION_KEY] = add_participant_form.attributes
+    end
+
+    def current_step
+      params[:step].tr("-", "_").to_sym
+    end
+
+    def step_param(step)
+      step.to_s.tr("_", "-")
+    end
```
kicferk1, line 53: You won't be shocked when I say I don't particularly like this. There are gems that perhaps do a better job at it, but more fundamentally, I don't like to have a single `update` for many steps. If a controller is meant to handle three update pages, maybe it should have three similar actions.

app/forms/schools/add_participant_form.rb
```diff
@@ -6,20 +6,54 @@ class AddParticipantForm
     include ActiveModel::Attributes
     include ActiveModel::Serialization
 
+    STEPS = %i[type details not_implemented].freeze
```
kicferk1, line 9: I am not sure a `form` is a good place to define steps. I think `wicked` did it in a controller (if I remember correctly), which feels somewhat better. But more importantly...

I much prefer having an explicit action in controller and explicit method on a form for handling a step. It means if steps have a degree of difference between them it is really easy to tell.

app/forms/schools/add_participant_form.rb
kicferk1: I don't think we want to spend time on writing a form wizard. If we think we need it let's use an existing solution - there are quite a few gems that do it around.

app/views/schools/add_participants/details.html.erb
```diff
@@ -0,0 +1,17 @@
+<% content_for :title, "Add early career teacher or mentor - participant details" %>
+
+<% content_for :before_content, govuk_back_link(text: "Back", href: back_link_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-l"><%= @school.name %>, <%= @cohort.display_name %> cohort</span>
+    <h1 class="govuk-heading-l">Enter their personal details</h1>
+
+    <%= form_for add_participant_form, url: { action: :update }, method: :post do |f| %>
```
kicferk1, line 10: action `update` method `post`? That sounds a bit surprising.

config/routes.rb
```diff
@@ -193,11 +193,19 @@
...
+            controller: "add_participants",
+            path: "add",
+            constraints: {
+              step: Regexp.union(
+                *Schools::AddParticipantForm::STEPS.map { |step| step.to_s.tr("_", "-") },
+              ),
+            },
+          ) do
+            get ":step", action: :show
+            post ":step", action: :update
```
kicferk1, line 208: How many steps do we expect to have there?

If it's less than 5 I think having a pair of `get` and `put/post` endpoints is a solution that takes similar amount of space, but a lot less brain power to process...

app/controllers/schools/add_participants_controller.rb
BroiSatse: Hmmm, I think I'd m,uch rather prefer a helper method than sharing an instance variable between two separate objects. I am aware that's how rails work by default, but it feels, well, dirty - it's a result of adaptation of MVC pattern that has been created for a GUI applications and really doesn't fit web server architecture...

app/controllers/schools/add_participants_controller.rb
BroiSatse: This is again a result of Rails convention of resource-oriented thinking messing with a non-resourcefull logic. There is no resourcefull "update" here, we simply populate the form/domai operation object. No domain logic is being executed until the last step (that will have a separate action). Defining multiple steps would be really cumbersome.

app/forms/schools/add_participant_form.rb
BroiSatse: You're correct - wicked does it in the controller. And I always find it extremely weird and, well, incorrect. Wicked AFAIR is based on a fixed step order which will not work in this case.

The form needs to perform step-related validations, so it would need to reach to the controller to know what step it is on. It should be the form that defines its own flow and order of assignments. Even when we speak about this, we say "mutlistep form", not "multistep controllers" so I'd argue it is form's responsibility to define its steps.

As per explicit actions in the controller - each step is only to assign the attributes to the form. There is absolutely no business logic there - I had those actions as a separate actions before and they were just repeating same code all the time.

app/views/schools/add_participants/details.html.erb
BroiSatse: For some reason cannot respond to precious comment: `I don't think we want to spend time on writing a form wizard.` - I've worked with wicked before and I found fighting with it much more than the time spent on writing this, especially around dynamic step ordering. It was a while ago, so maybe it would be better this way though...

config/routes.rb
BroiSatse: Something between 6 or 9 depending on decision if certain scenarios display the same step with conditionals or completely different steps.

I agree this might seem a bit complicated, but as befere - this code extracts a technical code in order to make business logic easier to follow. Ideally I'd extract this later into a separate file(s), reducing the whole controller into just few configuration lines.

app/forms/schools/add_participant_form.rb
kicferk1: When actions repeat a lot of code, I prefer to extract the common part into a private method. That makes the differences obvious - say, where do we go after a step fails or succeeds.

If I want to know from this code where we go after a particular step I need to understand what the next step method does (which for now is easy, but it doesn't always need to be). 

And as a general solution, I think it centralises the routing logic in a pretty scary way. Think about our flow for nominating induction tutors - there are several branches that lead into dead ends when particular conditions are met. And the `next_step` method needs to know about those, right?

I remember wicked from a long time ago, and my experience was pretty similar - it did the simple thing we wanted, but we struggled a lot around edge cases. I think if we try to apply this solution to the wider class of problems that we face across this project we will have a very similar experience. And if we don't apply it to any other solution, well, it seems excessively complicated for one flow :( 

For what it's worth... If we really want to go down this road, can we replicate what `apply-for-teacher-training` are doing? They have two wizards (from what I can tell), and it looks like they spent a long time thinking about this problem. 
`ReconfirmDeferredOfferWizard` is probably closer in scope to what we want. They set up a controller action for each step, have a list of steps in the form (well, they call it a wizard, but you know, naming), but the rest looks pretty similar to what you have. The larger one is `OfferWizard` - which has its own `new create edit update` controller per step. A bit much for our need I think.

app/views/schools/add_participants/confirm.html.erb
```diff
@@ -0,0 +1,65 @@
...
+      <%#         Not yet assigned %>
+      <%#       <% end %1> %>
+      <%#     </dd> %>
+      <%#     <dd class="govuk-summary-list__actions"> %>
+      <%#       <%= govuk_link_to({ action: :show, step: "mentor" }) do %1> %>
+      <%#         Change <span class="govuk-visually-hidden">mentor</span> %>
+      <%#       <% end %1> %>
+      <%#     </dd> %>
+      <%#   </div> %>
+      <%# <% end %1> %>
```
kicferk1, line 58: Somehow I don't think it should be here 😂

callumacrae: it should it's just waiting for the mentor changes stan is making! at the moment all the variables it's referring to are undefined

config/environments/development.rb
```diff
@@ -11,7 +11,7 @@
   config.cache_classes = false
 
   # Do not eager load code on boot.
-  config.eager_load = false
+  config.eager_load = true
```
callumacrae, line 14: The comment was just _asking_ for this to happen

```
# Eager load code on boot?
```

or delete imo

BroiSatse: Ah, that was just for order-dependent loading issues debugging :D Definitely deleting

app/views/schools/add_participants/confirm.html.erb
BroiSatse: @kicferk1 - Yup, this is getting a bit odd to have multiple PRs for a single journey. Not really sure what the best solution is - maybe viewing a single PR by a well defined commit history?

kicferk1: Fine, let's add the remaining part of the feature here.

I know dev retro happened after a lot of this code was written. Buut, in future I would expect to see a PR adding this feature in as simple a way as possible, and then, once that goes in, another PR that adds an abstraction of multi-page form. Cause this is exactly the thing that makes reviewing things harder.

BroiSatse: I think the problem lies in the definition of `as simple a way as possible`. I don't think writing this in a RailsWay(tm) would make it simple at all. There would be a massive amount of repetitions - not only around all the separate action but also around the `back` link options which would require writing conditions for both going back and forward depending on the state (not to mention a number of edge cases when the conditions are time-sensitive). The form flow would be heavily obfuscated by the amount of boilerplate code and, quite likely, it would take more time to write it this way to handle all the odd cases.

So I'd say our approach is pretty much the same, but we differ in definition of `as simple as possible` - I think this is pretty simple - there is no metaprogramming involved, no custom DSL, the whole logic is split in two files and views. Refactor is definitely pending with some extra edge case handling - like we probably should reset all the step information when user navigates back and changes some details (but only in some cases!).

app/forms/schools/add_participant_form.rb
```diff
@@ -6,20 +6,114 @@ class AddParticipantForm
...
+    def record_completed_step(step)
+      if completed_steps.include? step
+        self.completed_steps = completed_steps[0..completed_steps.index(step)]
+      else
+        completed_steps << step
+      end
+    end
+
+    def completed_steps=(value)
+      super(value.map(&:to_sym))
```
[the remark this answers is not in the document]
kicferk1: And I guess, if there is a `super` method for it, can we just make sure that we always pass an array of symbols to it?

app/forms/schools/add_participant_form.rb
```diff
@@ -6,20 +6,114 @@ class AddParticipantForm
...
+    end
+
+    def email_already_taken?
+      User.exists?(email: email)
+    end
+
+    def type=(value)
+      super(value&.to_sym)
+    end
```
kicferk1, line 98: Ditto, when do we pass `type` that is a string?

app/forms/schools/add_participant_form.rb
```diff
@@ -6,20 +6,114 @@ class AddParticipantForm
...
+        user = User.create!(full_name: full_name, email: email)
+
+        profile_class = type == :ect ? EarlyCareerTeacherProfile : MentorProfile
+        profile = profile_class.new(
+          user: user,
+          school_id: school_cohort.school_id,
+          cohort_id: school_cohort.cohort_id,
+        )
+        profile.mentor_profile = mentor&.mentor_profile if type == :ect
+        profile.tap(&:save!)
```
kicferk1, line 115: What's the difference between `provile.save!` and `profile.tap(&:save!)`?

app/views/schools/add_participants/choose_mentor.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <%= f.govuk_radio_button :mentor_id, mentor.id, label: { text: mentor.full_name }, link_errors: i == 0 %>
+  <% end %>
+
+  <%= f.govuk_radio_divider %>
+
+  <%= f.govuk_radio_button :mentor_id, "later", label: { text: 'Assign mentor later' } %>
+<% end %>
+
+  <%= f.govuk_submit "Continue" %>
+<% end %>
```
kicferk1, line 22: Something wrong with whitespace here.

app/views/schools/add_participants/confirm.html.erb
```diff
@@ -0,0 +1,71 @@
+<% content_for :title, "Confirm details" %>
+
+<% content_for :before_content, govuk_back_link(text: "Back", href: back_link_path) %>
+
+<%= add_participant_form.attributes.inspect %>
```
kicferk1, line 5: Uhh, what is this?

app/views/schools/add_participants/email_taken.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% if email_used_in_the_same_school? %>
+      <h1 class="govuk-heading-xl govuk-!-margin-bottom-7">This email has already been added</h1>
+
+      <p class="govuk-body">This email address has already been used to add an early career teacher or mentor.</p>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li>if they are an early career teacher, see if they have been added to another cohort</li>
+        <li>if they are a mentor, you just need <%= govuk_link_to "add them to this cohort", "#" %></li>
```
kicferk1, line 14: Can we have a todo comment here (link to work in future ticket)

config/routes.rb
```diff
@@ -203,11 +203,20 @@
...
+            path: "add",
+            constraints: {
+              step: Regexp.union(
+                *Schools::AddParticipantForm::STEPS.map { |step| step.to_s.tr("_", "-") },
+              ),
+            },
+          ) do
+            get ":step", action: :show
+            patch ":step", action: :update
+            post :complete
```
kicferk1, line 219: So we can `get`, `patch` and `post` for each of the steps?

spec/cypress/integration/schools/AddParticipants.feature
```diff
@@ -6,13 +6,91 @@ Feature: School leaders should be able to manage participants
...
+    Then I should be on "2021 school participant confirm" page
+    And the page should be accessible
+    And percy should be sent snapshot called "school ect participant confirmation page"
+
+    When I click the submit button
+    Then "page body" should contain "James Bond has been added as an ECT to the 2021 cohort"
+    And the page should be accessible
+    And percy should be sent snapshot called "school ect participant added"
+
+
```
kicferk1, line 48: Too much whitespace

kicferk1 (commented):
Are we missing some rspecs?

app/forms/schools/add_participant_form.rb
```diff
@@ -6,20 +6,114 @@ class AddParticipantForm
...
+      @mentor = (User.find(mentor_id) if mentor_id.present? && mentor_id != "later")
+    end
+
+    def previous_step(current_step)
+      step_index = completed_steps.index(current_step)
+      return completed_steps.last if step_index.nil?
+      return if step_index.zero?
+
+      completed_steps[step_index - 1]
+    end
```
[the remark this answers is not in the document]
callumacrae: You can rejoin the journey halfway through by clicking the "Change" links on the confirmation page - unsure what the desired behaviour of the back button would even be in that case though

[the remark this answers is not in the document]
kicferk1: Yeah, but then we will have the previous steps from the session. You can't enter the journey in the middle without having session with starting step, because if you try you get redirected elsewhere (controller makes sure of it)

app/views/schools/add_participants/confirm.html.erb
BroiSatse: Uhh, cypress dubigging! :D

config/routes.rb
BroiSatse: No, `get` and `patch` for each step and special `post` to complete.

app/forms/schools/add_participant_form.rb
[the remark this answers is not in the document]
BroiSatse: Oh bay, we will disagree a lot. :P

I am not certain I understand what you mean by leaky. When it return nil then It flags to the controller there is no previous step and it is p to controller to decide what to do in this case. Tthat's the opposite of leaky as it encapusltes the form-related logic without need to know the controller-level logic. If this was to return the back path for the controller to navigate back to, then we can talk about leaked abstraction.

> more generic than it needs to be
Abstract mean generic, so it has to be as generic as possible, with as little assumptions as possible to avoid leaking abstraction. We should not avoid abstraction neither, they are one of the OOD pillars.

Whether we want to start in the middle is only up to the controller and form object should (probably) not really care about this. Even if we decide that form should enforce initial step, than this would only further complcate this PR (happy to do this in refactor as an optional feature). The way this concrete controller is written it is impossible to start the journey midstep for this journey but I might not be the case for some other journeys.

app/views/schools/add_participants/complete.html.erb
```diff
@@ -0,0 +1,14 @@
+<% content_for :title, "Core induction materials selected" %>
```
ethanmills, line 1: I don't think this is right?

ethanmills (commented):
I think I'm with Michal on this one. If we want a wizard we should probably use a gem. This looks like something I'm going to misunderstand and introduce bugs

app/views/schools/add_participants/choose_mentor.html.erb
```diff
@@ -0,0 +1,22 @@
+<% content_for :title, "Add early career teacher or mentor - participant type" %>
```
ethanmills, line 1: Wrong title?

app/views/schools/add_participants/email_taken.html.erb
ethanmills: I don't think we want to show this this year, right? Just confusing

spec/cypress/integration/schools/AddParticipants.feature
```diff
@@ -6,13 +6,91 @@ Feature: School leaders should be able to manage participants
...
+    And I type "james.bond.007@secret.gov.uk" into field labelled "Email"
+    And I click the submit button
+    Then I should be on "2021 school choose etc mentor" page
+    And the page should be accessible
+    And percy should be sent snapshot called "school participant choose mentor page"
+
+    When I click the submit button
+    Then "page body" should contain "can't be blank"
+
+    When I set "new participant mentor radio" to "51223b41-a562-4d94-b50c-0ce59a8bb34d"
```
ethanmills, line 37: Really minor, but in future I think I would prefer using the :name over the :value. We're trying to keep this human readable

app/controllers/schools/add_participants_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+      { action: :show, step: step_param(previous_step) }
+    end
+
+    def add_participant_form_params
+      params.require(FORM_PARAM_KEY).permit(:type, :full_name, :email, :mentor_id)
+    end
+
+    def email_used_in_the_same_school?
+      User.find_by(email: add_participant_form.email).school == add_participant_form.school_cohort.school
+    end
```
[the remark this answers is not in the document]
BroiSatse: Yeah, I was on edge with that one. Ideally, I'd convert our step views into view-components and then that's where this code should be then. For now, neither form nor controller seems to be a great place for it. Alternatively, we could split the `email_taken` view into two separate view, and then have this check within forms `next_step` logic.

Actually there are many alternatives - previously @callumacrae has wrote this check as an additional validation on the form and some extra code in the controller to pick up that specific error and display the email_taken page instead. I've changed this into having `email_taken` as a step instead as, to me, at the point that page is displayed the details step has been completed, resulting in the premature termination of the flow. 

Yet another option would be to keep it as a validation, but make it strict and rescue the exception in the controller. This would however require a special treatmant of `previous_step` as it will preven `details` step from completion which does not feel right.

config/routes.rb
```diff
@@ -193,11 +193,19 @@
...
-            resource :type, only: %i[show update]
-            resource :details, only: %i[show update]
+          get :add, to: "add_participants#start"
+
+          scope(
+            controller: "add_participants",
+            path: "add",
+            constraints: {
+              step: Regexp.union(
+                *Schools::AddParticipantForm::STEPS.map { |step| step.to_s.tr("_", "-") },
```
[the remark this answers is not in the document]
BroiSatse: And I was looking for `kebabize` :D

spec/cypress/integration/schools/AddParticipants.feature
BroiSatse: Very much agree - actually I'd prefer to write it as `When I click on <name>` to properly simulate user interaction...

app/forms/schools/add_participant_form.rb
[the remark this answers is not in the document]
kicferk1: > The way this concrete controller is written it is impossible to start the journey midstep for this journey but I might not be the case for some other journeys.

Other journeys using the same form? I guess this goes back to 'ideally, generic multistep form should be in its own PR' so it's easier to understand it without the baggage of feature changes clogging my head :P

app/controllers/schools/add_participants_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+    def step_param(step)
+      step.to_s.tr("_", "-")
+    end
+
+    def back_link_path
+      previous_step = add_participant_form.previous_step(current_step)
+      return schools_cohort_participants_path unless previous_step
+
+      { action: :show, step: step_param(previous_step) }
+    end
```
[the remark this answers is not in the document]
BroiSatse: Under the hood, all the url_helpers exists only to generate a hash that is then passed to `url_for`. The minimal url hash requires at least controller and action key, with any additional params defined in the routes. Additionally, controller keeps the hash for the current route in `url_options` and uses this as a base for any future route generations, so that we only need to specify the keys that are different from the current route. Hash can be used wherever one would use the url_helper.

BTW, this is also why in the context of any schools/base controller we can freely us `school_cohort_path` instead of `school_cohort_path(@cohort.start_year)`. Rails will automatically use the cohort data from the current route.

app/forms/schools/add_participant_form.rb
```diff
@@ -6,20 +6,114 @@ class AddParticipantForm
     include ActiveModel::Attributes
     include ActiveModel::Serialization
 
+    STEPS = %i[type details choose_mentor email_taken confirm not_implemented].freeze
```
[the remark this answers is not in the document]
BroiSatse: I'm a bit on the edge here as well - email_taken indeed does not update the form, but it is still a step in the process - user can navigate back to the previous step which makes it a bit of a blind alley of the journey. It is the form though that decides that user should go there and as such I think it should still be listed as a step.

Confirm however is definitely a step - ideally, I'd add `on: [:complete]` to every single validation as well to revalidate the entire form before processing.

app/forms/schools/add_participant_form.rb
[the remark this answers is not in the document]
BroiSatse: > Other journeys using the same form?

No, other forms built using the same abstraction. The problem here is that I didn't draw the abstraction bounderies very well -  a lot of decision here were taken based on an implementation of abstract, multistep form and ensuring this abstraction is not leaking, however abstraction itself has not been extracted from the class yet.

app/forms/schools/add_participant_form.rb
BroiSatse: The problem with existing solutions is that they do not allow dynamic step configuration (at least all the ones I've seen so far, maybe there is a diamond somewhere that hasn't pick much popularity yet).

app/views/schools/add_participants/confirm.html.erb
kicferk1: We have a bunch of multipage flows, and they use one action per page (and one url per page). It might be that we pushed that pattern quite far, and that it would become cumbersome to use for this feature. 

But this feature would need to be a large PR (if we want a single PR for logically cohesive feature) in any case - loads of new content, loads of new logic, loads of tests. I imagine easily over 500 lines. This level of complexity is exactly something that we don't want to complicate further, by adding new and complicated systems to our setup.

So I guess my definition of `as simple as possible` = `as little new cognitive load as possible`. I kind of see introducing the generic multipage form as a refactor of the old setup with one action and one route per page. And refactors are great on small PRs, and not great on large PRs :/ The bigger the PR the scarier it is to include a complex refactor in it.

app/forms/schools/add_participant_form.rb
```diff
@@ -6,20 +6,114 @@ class AddParticipantForm
...
+        end
+      when :choose_mentor then :confirm
+      else :not_implemented
+      end
+    end
+
+    def record_completed_step(step)
+      if completed_steps.include? step
+        self.completed_steps = completed_steps[0..completed_steps.index(step)]
+      else
```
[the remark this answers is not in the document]
BroiSatse: Yes, but only on the completion of the previous step. Step completion can change the further step order, whcih would seriously iompact the `previous_step` logic. 

imagine user going through type (etc) -> details -> confirmation (completed steps are then `%i[type details]`, then navigating back to `type` and changing selection to `mentor`. In the full journey, it will then be presented with another step to choose whether they want to add an existing mentor from other cohort or the completely new mentor. If they select to create a new mentor they are taken to the details page again. If we didn't trim the `completed steps` on update of `type`, `completed steps` would have value of `%i[type details mentor_options]`, so the back_button on the `confirm` page would lead him to the `mentor_options` instead of `details`.

Ideally, we also should wipe the data provided in the steps we're trimming out

app/forms/schools/add_participant_form.rb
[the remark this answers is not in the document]
BroiSatse: `super` comes from the `attribute :complete_steps`. We could ensure the correct format of the input, but then it would need to happen in the controller, as we read those values from the session whcih converts them to string) and the whole functionality will be broken if the user of this class forgets to do that. I think it is alwayes best to limit any future user of this class from knowing too much about how it works.

But again, I'd prefer this to be done declaratively. Again, probably will be added in the refactor:

```
attribute :complete_steps, transform: ->(value) { value.map(&:to_sym) }
attribute :type, transform: :to_sym
```

app/forms/schools/add_participant_form.rb
BroiSatse: `type` comes from two sources - either form representation stored in the session or from the user's input, which is fine since ruby 2.2.

app/views/schools/add_participants/email_taken.html.erb
BroiSatse: It is even more confusing as they could have been already added to this cohort :)

app/controllers/schools/add_participants_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+        add_participant_form.record_completed_step current_step
+        store_form_in_session
+        redirect_to action: :show, step: step_param(add_participant_form.next_step(current_step))
+      else
+        render current_step
+      end
+    end
+
+    def complete
+      @participant = add_participant_form.save!
```
kicferk1, line 37: So this is actually a `participant_profile`, right? I think I would like to stick to naming convention of 
`participant` = `user who is a mentor or ect`
`participant_profile` = `ect_profile or mentor_profile`, at least until we work out a neater way of handling them.

ethanmills (approved):

PR #496, opened 2021-06-09
CPDRP-260 Add API for Data Studio

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/browse/CPDRP-260)
Create API endpoint that Data Studio can use to fetch analytics data for school rollout
This is the first step in getting an automatic feed of data into Data Studio.

Add endpoint to expose school rollout data used by Data Studio
Add `DataStudioApiToken` as a private auth token

### Guidance to review
You need an API key.
The endpoint `/api/v1/data-studio/school-rollout` should return ~12Mb of school related data

### Review Checks


app/controllers/api/v1/data_studio/school_rollout_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+          render json: ::DataStudio::SchoolRolloutSerializer.new(paginate(rollout_data))
+            .serializable_hash.to_json
+        end
+
+      private
+
+        def rollout_data
+          School
+            .eligible
+            .joins(:nomination_emails)
```
ethanmills, line 20: Once we start updating schools from GIAS, we will have eligible schools who have not been sent a nomination email

app/controllers/api/v1/data_studio/school_rollout_controller.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module DataStudio
+      class SchoolRolloutController < Api::ApiController
```
ethanmills, line 6: I think we need to stop lead provider tokens from working with this endpoint. This includes details of schools who have opted out, which we don't want to share with them. I'm also not sure we're allowed to share which providers have recruited which schools, which we will be adding very soon

tonyheadford: Yes, was going to add the private endpoint code that @asmega wrote

app/controllers/api/v1/data_studio/school_rollout_controller.rb
tonyheadford: Thanks, good to know so that needs to be a left outer join then

PR #497, opened 2021-06-09
Create json schemas for participant_declarations and contracts

blackrat:
This is in preparation for json schema validation
- providers v0.1 and v0.2 (existing shapes for each iteration)
- contract data to be loaded from secure storage as v1.0, which we believe is the correct format and currently what the calculation engine and database uses.

### Context

Data pushed in should be tested against a json schema to ensure that it's in the correct format.

Add in boilerplate schemas to be used for participant events and contracts

### Guidance to review

This is purely to check the shape of things and whether we have the correct structure for the incoming json objects.

### Review Checks

You can't


jamesrose (approved):

ethanmills (approved):

etc/schema/0.1/ecf/participant_declarations/create/request.json
```diff
@@ -0,0 +1,3 @@
+{
```
kicferk1, line 1: This file (and all others) lives in `etc` directory. Is that a typo? If not, what does it stand for?

PR #498, opened 2021-06-10
CPDTP-196 JSON Schema URI to Local Filename Converter and reader

blackrat:
### Context

incoming json can be validated against a schema document to ensure that the content and format are as expected before passing bad data to the underlying code.
This puts in place the ability to use the JSON::Schema gem to validate against a set of connected json schema documents and to have those local file based and translate http uris to file uris.
In addition, it has been proposed that we allow current API version + 2 ancestors, and this will help ensure that versioning is managed correctly by comparing incoming json documents against the correctly versioned schema.

- Slight enhancement to InitializeWithConfig to allow it to raise more verbose error messages when required configuration items (Dependency Injection) are required and support for classes which include them to optionally declare them as such.
- Versioned schema reader. JSON::Schema will use this as the reader for when it tries to pull in files.
- Versioned event file name. Set up the baseline filenames for the schema files with everything overridable, but mainly (version=>schema version and event=>create) for the incoming json document.
- Slight change to the request_schema.json files to add in $refs where they were missing or point to too high a level.

### Guidance to review

Run and read the tests. They describe how things work and can be changed.

### Testing

Tests are supplied for the version_event_file_name and the versioned_schema_reader which show how names are constructed and converted from the uri.
InitializeWithConfig tests have been enhanced to show the new errors coming back when required_config is used.

### Review Checks

You can't

### DIFF
--- a/lib/initialize_with_config.rb
+++ b/lib/initialize_with_config.rb
@@ -49,6 +49,8 @@ require "active_support"
 require "active_support/core_ext"
 
 module InitializeWithConfig
+  class MissingRequiredArguments < RuntimeError; end
+
   attr_accessor :config
 
   class << self
@@ -64,17 +66,31 @@ module InitializeWithConfig
   end
 
   def default_config
-    HashWithIndifferentAccess.new
+    {}
+  end
+
+  def required_config
+    self.class.required_config
   end
 
 private
 
   # @param [Hash] config
   def initialize(config)
-    self.config = config.respond_to?(:[]) ? default_config.merge(config) : default_config
+    self.class.externally_injected ||= []
+    self.config = HashWithIndifferentAccess.new(config.respond_to?(:[]) ? default_config.merge(config) : default_config)
     self.config.each do |key, value|
       auto_define(key) { value } unless allow_override? && respond_to?(key)
     end
+    check_required
+  end
+
+  def check_required
+    raise ::InitializeWithConfig::MissingRequiredArguments, "missing required dependency injected items #{missing_required_params} in class #{self.class.name}" unless missing_required_params.empty?
+  end
+
+  def missing_required_params
+    self.class.externally_injected.reject { |name| respond_to?(name) }
   end
 
   def auto_define(key, &value)
@@ -87,10 +103,16 @@ private
 
   module InitialiseClassConfig
     include ActiveSupport
-    attr_accessor :prevent_override
+    attr_accessor :prevent_override, :externally_injected
 
     def prevent_local_override
       self.prevent_override = true
+    end
+
+    def required_config(*args)
+      return @externally_injected || [] if args.nil?
+
+      self.externally_injected = args
     end
 
     # @param [Hash] config
--- a/lib/json_schema/version_event_file_name.rb
+++ b/lib/json_schema/version_event_file_name.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require "initialize_with_config"
+
+module JsonSchema
+  class VersionEventFileName
+    include InitializeWithConfig
+    required_config :version, :event
+
+    def call
+      Rails.root.join(schema_root, version.to_s, schema_path, event.to_s, schema_file)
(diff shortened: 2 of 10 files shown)


lib/json_schema/versioned_schema_reader.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+  private
+
+    attr_reader :version
+
+    def initialize(version: 1.0, options: {})
+      @version = version
+      super(options)
+    end
```
kicferk1, line 16: Is this meant to be private?

How is this class used exactly?

lib/json_schema/version_event_file_name.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def default_config
+      {
+        schema_root: "etc/schema",
+        schema_path: "ecf/participant_declarations",
+        schema_file: "request_schema.json",
+      }
+    end
+  end
+end
```
kicferk1, line 24: Soo, where do `version` and `event` live? Where do they need to live?

And what does `call` do exactly?

lib/initialize_with_config.rb
```diff
@@ -64,17 +66,31 @@ def call(*)
...
+  def required_config
+    self.class.required_config
   end
 
 private
 
   # @param [Hash] config
   def initialize(config)
-    self.config = config.respond_to?(:[]) ? default_config.merge(config) : default_config
+    self.class.externally_injected ||= []
```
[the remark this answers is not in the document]
blackrat: InitializeWithConfig is a Dependency injector. externally_injected is a private array which holds items (methods, classes, variables etc.), which have been explicitly declared as required to be present in the config. The main use of this is to compare what has been passed in with what is required and to raise an appropriate error message if anything is missing. They are held against the including class themselves, since that's where they are defined.

lib/json_schema/versioned_schema_reader.rb
blackrat: https://www.rubydoc.info/gems/json-schema/2.6.2/JSON/Schema/Reader

It's passed as an option into JSON::Schema.fully_validate to allow it to read the included files when they are declared as references or schemas.

lib/json_schema/version_event_file_name.rb
blackrat: Have a look at InitializeWithConfig. It's doing dependency injection (hence the required_config), via the configuration object passed from the calling services. Call in this instance is just creating a path to the schema file.

lib/initialize_with_config.rb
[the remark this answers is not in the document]
kicferk1: Right. I've read through a bunch of code, and what I have in my head is:

When you have `SomeClass` include `InitializeWithConfig`, then...

At the moment you make a new instance of `SomeClass`, you add a method to the new instance for each of the config elements. And to `SomeClass` instance, those methods are `externally_injected`?

lib/json_schema/version_event_file_name.rb
kicferk1: Right, so they live on `VersionEventFileName`, so long as it's initialised with `version` and `event` arguments.

kicferk1 (approved):

lib/json_schema/version_event_file_name.rb
blackrat: Correct. The class level call with the appropriate parameters set as a configuration hash creates version and event methods which the instance level call uses.

lib/initialize_with_config.rb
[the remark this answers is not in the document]
blackrat: The ones that you have declared in the class with `required_config` are the ones that are compared against the `externally_injected` array. It's mainly to clean up the error messages since it's more meaningful to raise: `::InitializeWithConfig::MissingRequiredArguments` and show which values are missing on instantiation and for which class than to let the Ruby standard ArgumentMissing raise when they are used.

PR #501, opened 2021-06-10
CPDTP-196 Adding in the schema_validator to the API

blackrat:
### Context

Schema validation should happen to make sure that the data coming in matches the expectations.
This is performed by the JSON::Schema validator gem. Most of the code for this is already committed, but it needs to be connected to the actual API handling code for ParticipantDeclarations.

- Add in the schema validation as a first step to the RecordParticipantEvent dispatcher service.
- Update errors being sent back to the lead_provider to give a more detailed description of the problem
- Fixes to the schema docs in etc to remove the incorrect references. These are used internally to walk through trees of referenced files, but everything needed is currently included in the top level since there are no separate definitions for properties (as yet).
- Create a proxy and adapter for the schema validator
- Update tests with the correct messages
- Generate swagger docs to reflect the changes

### Guidance to review

JSON Schema validation has been written as generic solution for incoming json. Currently the schemas that it knows, and for which the handlers are included in this PR are the participant_declarations and the contracts.
Most of the library code is to enable file based schema documents lookups and references to be used since they are hosted in the /etc directory alongside the rest of the application.

### Testing

Full test suite for the participant_declaration_validation libraries and testing via the endpoint can be done within the `/api-docs/index.html` or directly to the `/api/v1/participant-declarations`.
`contract_validation` is more of a placeholder to allow it to be integrated with the secure vault and validate content pulled from there.

### Review Checks

API is viewable from the Swagger UI detailed as the endpoints above.


app/services/record_participant_event.rb
```diff
@@ -20,11 +18,19 @@ def call
...
+  def schema_validation
+    errors = participant_event_validator.call(config, body: raw_event)
+    raise ActionController::ParameterMissing, errors unless errors.empty?
+
+    true
+  end
+
   def default_config
-    HashWithIndifferentAccess.new(
+    {
```
blackrat, line 29: Switched to doing this in the merge rather than needing to create it as one.

app/services/record_participant_event.rb
```diff
@@ -20,11 +18,19 @@ def call
 
 private
 
+  def schema_validation
+    errors = participant_event_validator.call(config, body: raw_event)
```
jamesrose, line 22: Does this read the schema file on _every_ request?

blackrat: Currently, yes. That can be optimised if it's an issue though. The validator will work with an in-memory schema object.

app/services/json_schema/base.rb
```diff
@@ -0,0 +1,30 @@
...
+    def call(body:)
+      config[:schema] = schema
+      error_checker.call(config, body: body)
+    end
+
+  private
+
+    def schema
+      JSON.parse(File.read(json_schema_file_location.call(config)))
+    end
```
kicferk1, line 19: Ehhh, this seems... convoluted. In `call` we change the `config` so that it includes `schema` field. and we get `schema` by calling some method to which we pass the old `config`? I sure hope that method doesn't need `schema` in it...

Can we minimise the size of config with which we call `json_schema_file_location`? A new hash would probably make it a lot easier to read.

app/services/json_schema/validate_participant_event.rb
```diff
@@ -0,0 +1,16 @@
...
+  class ValidateParticipantEvent < Base
+    def default_config
+      super.merge(
+        {
+          version: "0.2",
+          schema_path: "ecf/participant_declarations",
+        },
+      )
+    end
+  end
```
kicferk1, line 15: That looks quite different to `ValidateContract`. Is that difference intentional?

lib/json_schema/validation.rb
```diff
@@ -0,0 +1,27 @@
...
+require "json_schema/validator_adapter"
+require "initialize_with_config"
+
+module JsonSchema
+  class Validation
+    include InitializeWithConfig
+    required_config :schema, :schema_reader, :schema_validator, :version
+
+    def call(body:)
+      schema_validator.fully_validate(schema, body, schema_reader: schema_reader.new(version: version))
```
kicferk1, line 14: What other types of validations are there? `half_ass_validate`? 😂 
Asking more out of curiosity :)

spec/lib/json_schema/validator_adapter_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+          "b" => { "type" => "integer" },
+        },
+        "additionalProperties": false,
+      }
+    end
+
+    it "validates a json body against a schema" do
+      expect(described_class.fully_validate(schema, body, schema_reader: JSON::Schema::Reader.new)).to eq []
+      expect(*described_class.fully_validate(schema, invalid_body, schema_reader: JSON::Schema::Reader.new)).to include("did not contain a required property of 'a'").once
+      expect(*described_class.fully_validate(schema, extra_attributes, schema_reader: JSON::Schema::Reader.new)).to include("contains additional properties [\"extra_stuff\"] outside of the schema when none are allowed in schema").once
```
kicferk1, line 26: Is there some matcher that can help us avoid stringifying the output?

Something like `expect(described_class.fully_validate(schema, invalid_body, schema_reader: JSON::Schema::Reader.new).to contain_exactly("did not contain a required property of 'a'")`?

app/services/json_schema/validate_participant_event.rb
blackrat: Yes. You are correct. That should be private and the InitializeWithConfig isn't directly required since it's inheritance will take care of that.

app/services/record_participant_event.rb
```diff
@@ -1,16 +1,14 @@
...
   include InitializeWithConfig
-  class << self
-    def required_params
-      %i[participant_id lead_provider declaration_type declaration_date raw_event]
-    end
-  end
+  required_config :participant_id, :lead_provider, :declaration_type, :declaration_date, :raw_event
 
   def call
+    return :unprocessable_entity unless schema_validation
```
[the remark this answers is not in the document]
blackrat: Ok. Good point. It raises to set the exception error, which means that it will only hit this line if successful. I was considering making that schema based validation a service on its own which is called before this one. I flip-flopped on the "validation" and "it's part of the pipeline" a couple of times in my head.

lib/json_schema/validation.rb
blackrat: From the validation Readme at: https://github.com/ruby-json-schema/json-schema

1. validate: returns a boolean on whether a validation attempt passes
2. validate!: throws a JSON::Schema::ValidationError with an appropriate message/trace on where the validation failed
3. fully_validate: builds an array of validation errors return when validation is complete

spec/lib/json_schema/validator_adapter_spec.rb
blackrat: I considered using a regex since matching is harder for this one since it generates a schema uuid as part of the original message:
`The property '#/' did not contain a required property of 'a' in schema 78d55aeb-99b3-565f-b95e-fdb2e1014a93"`
As a later step, I want to parse the string to extract what exactly was missing and feed that into the error message that's presented back, but for the test's stringifying and doing an include was the simplest thing that could work.

app/services/json_schema/base.rb
blackrat: Yes. Making this an orchestration to pass in a schema as a param to call (and on down to error_checker), would make this cleaner. It strikes me that it's not as SOLID as it should be too and caching the schema is also something that may be required.

config/environments/development.rb
```diff
@@ -8,10 +8,10 @@
   # In the development environment your application's code is reloaded any time
   # it changes. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = false
+  config.cache_classes = true
 
   # Do not eager load code on boot.
-  config.eager_load = false
+  config.eager_load = true
```
kicferk1, line 14: Are we sure about that?

callumacrae: the comment!! 😢

spec/lib/json_schema/validator_adapter_spec.rb
kicferk1: Alright, that makes sense. Happy to keep string matching in that case :)

app/services/json_schema/base.rb
blackrat: Changed. This doesn't actually need to call the JSON parser. More solid if it is just passed the schema and the body to validate. Moved to lib as well.

app/services/json_schema/validate_participant_event.rb
blackrat: And now gone. The Contract and ParticipantEvent are now passed through by the calling service to a lib class to perform the validation

config/environments/development.rb
blackrat: Oops. No we're not. That slipped through from my testing to ensure that we don't have an autoload issue.

lib/has_di_parameters.rb
```diff
@@ -48,10 +48,10 @@
 require "active_support"
 require "active_support/core_ext"
 
-module InitializeWithConfig
+module HasDIParameters
```
jamesrose, line 51: I'm still struggling to see the value of this vs. basic keywords args with defaults and accessors. I think the name is a smell - we essentially have a module called 'has parameters' !

lib/has_di_parameters.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 #
-# The InitializeWithConfig module is used to allow for a hash of key/value pairs to be passed in for initialization
+# The HasDIParameters module is used to allow for a hash of key/value pairs to be passed in for initialization
```
kicferk1, line 4: There are some more mentions of old class name around in the comment documenting it.

What does `DI` stand for?

app/services/record_participant_event.rb
```diff
@@ -20,23 +24,40 @@ def call
 
 private
 
-  def default_config
-    HashWithIndifferentAccess.new(
+  def initialize(params)
+    inject_params(params)
+  end
```
kicferk1, line 29: That looks like we don't need it, do we?

lib/json_schema/version_event_file_name.rb
```diff
@@ -1,24 +1,23 @@
...
-      {
-        schema_root: "etc/schema",
-        schema_path: "ecf/participant_declarations",
-        schema_file: "request_schema.json",
-      }
+    def initialize(schema_root: "etc/schema", schema_path: "ecf/participant_declarations", schema_file: "request_schema.json")
+      @schema_root = schema_root
+      @schema_path = schema_path
+      @schema_file = schema_file
     end
```
kicferk1, line 21: I quite like having a list of all parameters that you can pass when creating an object in `initialize` which lives on the class. It helps ensure you are not passing something that's unused, and it makes it easier to spot if the class is missing some info.

If we want to have a re-usable way of handling 'missing parameter' errors, we could do it with a helper method, say `validate_parameters_exist`, called at the end of `initialize` with the parameters we want to enforce. I guess it would need to be aware of whether some `@` variables are defined, but I hope Ruby lets you check that.

lib/has_di_parameters.rb
blackrat: I've switched in the areas where it didn't make sense, but left it where we want to change things from the top level. I changed the name to make it more obvious that we're using something more akin to the way that "params" work when passed through from the controller.

lib/has_di_parameters.rb
blackrat: "Dependency Injection"

app/services/record_participant_event.rb
blackrat: We don't. One of the comments was that if someone overrides the `initialize` and doesn't use `super` was that it wasn't clear what it was doing. I wanted to show it, but I was pretty sure that I'd get a comment like yours.

lib/json_schema/version_event_file_name.rb
blackrat: It's a bit counter to the "Pass no more than four parameters into a method. Hash options are parameters." from the DfE Technical guidelines link. Personally, I prefer to pass an object which responds (and can be delegated to), which adheres more to the Liskov Substitution Principle.

kicferk1 (approved):

PR #504, opened 2021-06-11
CPDTP-166 Add bands to service fees and tweak payment breakdown rake task

plisovin:
### Context

Previously only Band A was used in calculation, this change includes all the banding and generates payment breakdown as the following:

```
+----------------------------------------------------+
|                 Breakdown Payments                 |
+---------+------------------+-----------------------+
| Banding | Service fee (PP) | Service fee (monthly) |
+---------+------------------+-----------------------+
| Band A  |     £558      |         £38,448          |
| Band B  |     £520      |         £35,862          |
| Band C  |      £480     |         £8,276           |
+---------+------------------+-----------------------+
Output payment (started) £756,000
Based on 4,500 participants and £1400 per participant in Band A, £1300 per participant in Band B, £1200 per participant in Band C
```

Run the rake task ```rake payment_calculation:breakdown Capita 4500``` with the contract as per [jira ticket](https://dfedigital.atlassian.net/browse/CPDTP-166):

```
{
  "uplift_target": 0,
  "uplift_amount": 0,
  "recruitment_target": 4500,
  "set_up_fee": 5_000,
  "band_a": {
    "max": 2000,
    "per_participant": 1400,
  },
  "band_b": {
    "min": 2001,
    "max": 4000,
    "per_participant": 1300,
  },
  "band_c": {
    "min": 4001,
    "per_participant": 1200,
  },
}
```

### DIFF
--- a/lib/payment_calculator/ecf/service_fees.rb
+++ b/lib/payment_calculator/ecf/service_fees.rb
@@ -1,18 +1,17 @@
 # frozen_string_literal: true
 
-require "initialize_with_config"
 require "payment_calculator/ecf/service_fees_for_band"
 
 module PaymentCalculator
   module Ecf
     class ServiceFees
-      include InitializeWithConfig
+      include PaymentCalculator::Ecf::Contract::ServiceFeeCalculations
       # required_config :contract # TODO: Uncomment this when the updated InitializeWithConfig is added from the other PR
       delegate :bands, to: :contract
 
       def call
         bands.map do |band|
-          Ecf::ServiceFeesForBand.call(config, band: band)
+          Ecf::ServiceFeesForBand.call(params, band: band)
         end
       end
     end
--- a/lib/payment_calculator/ecf/service_fees.rb
+++ b/lib/payment_calculator/ecf/service_fees.rb
@@ -6,7 +6,6 @@ module PaymentCalculator
   module Ecf
     class ServiceFees
       include PaymentCalculator::Ecf::Contract::ServiceFeeCalculations
-      # required_config :contract # TODO: Uncomment this when the updated InitializeWithConfig is added from the other PR
       delegate :bands, to: :contract
 
       def call
--- a/lib/payment_calculator/ecf/service_fees.rb
+++ b/lib/payment_calculator/ecf/service_fees.rb
@@ -1,18 +1,19 @@
 # frozen_string_literal: true
 
-require "payment_calculator/ecf/contract/service_fee_calculations"
+require "initialize_with_config"
+require "payment_calculator/ecf/service_fees_for_band"
 
 module PaymentCalculator
   module Ecf
     class ServiceFees
-      include PaymentCalculator::Ecf::Contract::ServiceFeeCalculations
+      include InitializeWithConfig
+      # required_config :contract # TODO: Uncomment this when the updated InitializeWithConfig is added from the other PR
+      delegate :bands, to: :contract
 
       def call
-        {
-          service_fee_per_participant: service_fee_per_participant.round(2),
-          service_fee_total: service_fee_total.round(2),
-          service_fee_monthly: service_fee_monthly.round(2),
-        }
+        bands.map do |band|
+          Ecf::ServiceFeesForBand.call(config, band: band)
+        end
       end
     end
   end
--- a/lib/payment_calculator/ecf/service_fees_for_band.rb
+++ b/lib/payment_calculator/ecf/service_fees_for_band.rb
@@ -9,9 +9,9 @@ module PaymentCalculator
 
       def call(band:)
         {
-          service_fee_total: service_fee_total(band).round(0),
-          service_fee_per_participant: service_fee_per_participant(band).round(0),
           service_fee_monthly: service_fee_monthly(band).round(0),
+          service_fee_per_participant: service_fee_per_participant(band).round(0),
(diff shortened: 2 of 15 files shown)


lib/payment_calculator/ecf/service_fees.rb
```diff
@@ -1,18 +1,19 @@
...
-require "payment_calculator/ecf/contract/service_fee_calculations"
+require "initialize_with_config"
+require "payment_calculator/ecf/service_fees_for_band"
 
 module PaymentCalculator
   module Ecf
     class ServiceFees
-      include PaymentCalculator::Ecf::Contract::ServiceFeeCalculations
+      include InitializeWithConfig
+      # required_config :contract # TODO: Uncomment this when the updated InitializeWithConfig is added from the other PR
```
kicferk1, line 10: I think that's merged now

lib/payment_calculator/ecf/service_fees_for_band.rb
```diff
@@ -0,0 +1,19 @@
...
+module PaymentCalculator
+  module Ecf
+    class ServiceFeesForBand
+      include Ecf::Contract::ServiceFeeCalculations
+
+      def call(band:)
+        {
+          service_fee_monthly: service_fee_monthly(band).round(0),
+          service_fee_per_participant: service_fee_per_participant(band).round(0),
+          service_fee_total: service_fee_total(band).round(0),
```
kicferk1, line 14: `round(0)`? Are we sure about that? I think I saw some `round(2)`s floating around...

lib/tasks/payment_breakdown.rake
```diff
@@ -28,23 +30,36 @@ namespace :payment_calculation do
...
+      rows: service_fees,
+    )
+    table.style = { alignment: :center }
+
+    output = <<~RESULT
+      Output payment (started) £#{output_payment}
+      Based on #{number_to_delimited(total_participants.to_i)} participants and #{per_participant_in_bands}
     RESULT
 
+    logger.info table
```
kicferk1, line 54: Alright, that is getting a little silly. If we want it to look presentable, maybe let's use html? I know we went for a minimum effort console approach first, but adding a new gem that lets you style things in console really doesn't sound that minimal for me. @jamesrose thoughts? 

Maybe two-step controller, step one being a simple form for inputting what you want and step two a view displaying something comparable to this? We can only have this route in non-real environments if it makes our lives easier...

lib/payment_calculator/ecf/service_fees_for_band.rb
plisovin: The requirement is to round to the nearest pound

lib/tasks/payment_breakdown.rake
plisovin: Could use html and 2-step controller of course

spec/lib/payment_calculator/ecf/contract/service_fee_calculations_spec.rb
```diff
@@ -8,12 +8,14 @@ class DummyClass
...
+    band_a = double("Band Double", per_participant: 996.00,
+                                   number_of_participants_in_this_band: 2000,
+                                   deduction_for_setup?: true)
     contract = double("Contract Double", recruitment_target: 2000, set_up_fee: 149_651.00, band_a: band_a)
     call_off_contract = DummyClass.new({ contract: contract })
 
-    expect(call_off_contract.service_fee_total.round(2)).to eq(647_149.00)
-    expect(call_off_contract.service_fee_monthly.round(0)).to eq(22_315.00)
-    expect(call_off_contract.service_fee_per_participant.round(0)).to eq(324.00)
+    expect(call_off_contract.service_fee_total(band_a).round(0)).to eq(647_149.00)
```
jamesrose, line 17: Why do you round the output here?

lib/tasks/payment_breakdown.rake
jamesrose: @kicferk1 This is all throw-away for now. As long as this task is just _formatting_ data returned by `PaymentCalculator::Ecf::PaymentCalculation` then I'm content. I'm keen to hold off on a UI until we know how it would be used.

Saying that, if there are things here that you wouldn't want to delete _today_ I would look to move them into the calculator. One day very soon this file won't exist.

spec/lib/payment_calculator/ecf/contract/service_fee_calculations_spec.rb
plisovin: We round as per requirement in https://docs.google.com/spreadsheets/d/1wGrFuwoyZJcQlyjnFxQD-5H8cWljSfbc/edit#gid=1504585790 

```Where a price is required, the figure will be rounded to the nearest pound```

Comment (jamesrose) [reply spec/lib/payment_calculator/ecf/contract/service_fee_calculations_spec.rb:17]: @plisovin Should that be in the calculator itself?

Comment (plisovin) [reply spec/lib/payment_calculator/ecf/contract/service_fee_calculations_spec.rb:17]: It's outside of the scope of this ticket, there is another one https://dfedigital.atlassian.net/browse/CPDTP-106

**Review by kicferk1 — APPROVED** [review]
We should probably get rid of 
```
# required_config :contract # TODO: Uncomment this when the updated InitializeWithConfig is added from the other PR
```
line, but other than that it looks good to me.

PR #506, opened 2021-06-11
Fix issue where cookie banner couldn't be accepted from privacy page

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDRP-407

### Guidance to review

To replicate the original issue:

- Go to e.g. https://ecf-review-pr-493.london.cloudapps.digital/
- DON'T TOUCH the cookies
- log in as school-leader@example.com
- On priv policy page, try to accept cookies

### Testing

Unsure if this needs request specs. I can't see how this would change in the future without invalidating the tests.

see pr app below

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -50,7 +50,7 @@ protected
     policy = PrivacyPolicy.current
     return if policy.nil?
 
-    return if request.format == "application/json"
+    return if request.format.json?
 
     return unless policy.acceptance_required?(current_user)
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -51,7 +51,6 @@ protected
     return if policy.nil?
 
     return if request.format == "application/json"
-    return if controller_name == "cookies"
 
     return unless policy.acceptance_required?(current_user)
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -50,6 +50,9 @@ protected
     policy = PrivacyPolicy.current
     return if policy.nil?
 
+    return if request.format == "application/json"
+    return if controller_name == "cookies"
+
     return unless policy.acceptance_required?(current_user)
 
     session[:original_path] = request.fullpath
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -50,6 +50,9 @@ def check_privacy_policy_accepted
     policy = PrivacyPolicy.current
     return if policy.nil?
 
+    return if request.format == "application/json"
+    return if controller_name == "cookies"
```
kicferk1, line 54: That feels a bit dirty.

Can we have a protected `requires_privacy_policy` method on this controller, have it return `true` by default and then in cookies controller we override that method to return `false`? 

And in this place we `return if requires_privacy_policy`?

callumacrae: good shout 👍

callumacrae: found an even nicer solution!

app/controllers/application_controller.rb
```diff
@@ -50,6 +50,8 @@ def check_privacy_policy_accepted
     policy = PrivacyPolicy.current
     return if policy.nil?
 
+    return if request.format == "application/json"
```
callumacrae, line 53: this is technically unrelated to this PR, but it feels like it's probably a good idea

app/controllers/cookies_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class CookiesController < ApplicationController
+  skip_before_action :check_privacy_policy_accepted
+
   before_action :set_cookie_form, only: :show
```
kicferk1, line 6: oh yeah, it is nicer <3

kicferk1 (approved):

app/controllers/application_controller.rb
ethanmills, line 53:
```suggestion
    return if request.format.json?
```

Does this work? We use a few mime types for JSON, and this should capture them all

ethanmills (approved):
One suggestion to try, don't forget to squash that squashable 🙂

PR #508, opened 2021-06-11
api/v1/users controller now requires private token

asmega:
### Context

- NPQ requires access to api/v1/users endpoint
- It will need to be able to create users
- Currently this is not possible as endpoint has been restricted to `EngageAndLearnApiToken` tokens only

- Valid api tokens but with incorrect permissions based off the `access_scope` now get a 403 instead of a 401
- `app/controllers/api/v1/users_controller.rb` now requires a token with `private_api_access` set to `true`
- Default `type` on `ApiToken` is now `ApiToken` not `LeadProviderApiToken`
- New `EngageAndLearnApiToken` tokens will have have `private_api_access` with `true` and there is migration to update any existing tokens to this state

### Guidance to review

- Change of default on `type` on `ApiToken` should not cause issues
- Access to api endpoints should continue to work

### Testing

- Test via review app

### Review Checks

- Call endpoint with `curl -I -H "Authorization: Bearer UNHASHED_TOKEN" https://ecf-review-pr-508.london.cloudapps.digital/api/v1/users`
- Should return 200 response
- Create another api token with `ApiToken.create_with_random_token!`
- Call endpoint with `curl -I -H "Authorization: Bearer UNHASHED_TOKEN" https://ecf-review-pr-508.london.cloudapps.digital/api/v1/users`
- Should get 403
- Call endpoint with `curl -I -H "Authorization: Bearer TOKEN_THAT_DOES_NOT_EXIST" https://ecf-review-pr-508.london.cloudapps.digital/api/v1/users`
- Should get 401

### DIFF
--- a/app/models/engage_and_learn_api_token.rb
+++ b/app/models/engage_and_learn_api_token.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class EngageAndLearnApiToken < ApiToken
+  attribute :private_api_access, default: true
+
   def owner
     "Engage and learn application"
   end
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -6,16 +6,14 @@ module Api
       include ApiTokenAuthenticatable
       include Pagy::Backend
 
-      before_action :ensure_engage_and_learn
-
       def index
         render json: UserSerializer.new(paginate(users)).serializable_hash.to_json
       end
 
     private
 
-      def ensure_engage_and_learn
-        head :forbidden unless current_user == "Engage and learn application"
+      def access_scope
+        ApiToken.where(private_api_access: true)
       end
 
       def updated_since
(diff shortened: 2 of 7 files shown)


app/models/engage_and_learn_api_token.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class EngageAndLearnApiToken < ApiToken
+  attribute :private_api_access, default: true
```
kicferk1, line 4: Can we also add a migration to change existing E&L tokens to have private access? I just added a token to staging that should last for a while, and if things go smoothly I wanna add another one to production soon.

asmega: Sure. Therefore I'm assuming we're happy doing migrations that change data.

kicferk1: I sure am

kicferk1: Definitely wrote some on E&L, pretty sure I saw some on R&P, in particular one that didn't work so well on the sandbox environment recently 😂 
But in principle, yes, I think they are too useful a tool to avoid :)

asmega: done

ethanmills (approved):
Thanks for sorting this, looks great!

kicferk1 (approved):

PR #512, opened 2021-06-14
Handle pagy overflow

ethanmills:
Couple of related things here:
- Change Pagy to return an empty page when requesting a page with no contents
- Update the swagger docs to match the data on sandbox

Previously Pagy would throw an exception when it ran out of pages. Sandbox has 10 ppts per lead provider, so we were seeing exceptions in the swagger examples

### DIFF
--- a/config/initializers/pagy.rb
+++ b/config/initializers/pagy.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+require "pagy/extras/overflow"
+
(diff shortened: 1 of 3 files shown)


kicferk1 (approved):
Sweet, this looks like it will work for `users` api too, right? (it's a global pagy setting rather than endpoint by endpoint)

ethanmills:
> Sweet, this looks like it will work for `users` api too, right? (it's a global pagy setting rather than endpoint by endpoint)

Yep!

PR #514, opened 2021-06-14
Multistep forms

BroiSatse:
This PR extracts an abstract Multistep::Form object, which together with `Multistep::Controller` (I hope) will allow us to build the future mutli-step journeys quite effortlessly.


app/lib/multistep/controller.rb
```diff
@@ -0,0 +1,137 @@
...
+        form.record_completed_step current_step
+        store_form_in_session
+        redirect_to action: :show, step: step_param(form.next_step)
+      else
+        render current_step
+      end
+    end
+
+    def complete
+      @result = add_participant_form.save!
```
kicferk1, line 80: Hmm, I would not expect `add_participant_form` mentioned here.

app/lib/multistep/controller.rb
```diff
@@ -0,0 +1,137 @@
...
+      session[self.class.session_key] = form.attributes
+    end
+
+    def current_step
+      params[:step].underscore.to_sym
+    end
+
+    def setup_form(form)
+      # noop
+    end
```
kicferk1, line 107: Uhh, what?

app/lib/multistep/controller.rb
```diff
@@ -0,0 +1,137 @@
...
+
+    def back_link_path
+      previous_step = add_participant_form.previous_step(from: current_step)
```
kicferk1, line 110: `add_participant_form` probably should be something else.

app/lib/multistep/controller.rb
```diff
@@ -0,0 +1,137 @@
...
+
+      def params_key(value = nil)
+        if value
+          @params_key = value
+        else
+          @params_key ||= @form && @form.model_name.param_key
+        end
+      end
+
+      def escape_route(&block)
```
kicferk1, line 36: Can we call it `default_back_route` or something like that? `escape` can mean many many things.

app/lib/multistep/controller.rb
```diff
@@ -0,0 +1,137 @@
...
+      params[self.class.params_key]&.permit(*attributes) || {}
+    end
+
+    def ensure_form_present
+      redirect_to escape_route unless session.key?(self.class.session_key)
+    end
+
+    def escape_route
+      instance_exec(&self.class.escape_route)
+    end
```
kicferk1, line 127: We already have this method in `public` scope - except there it takes an argument. Is that intentional? Also, what is `instance_exec` exactly?

app/lib/multistep/form.rb
```diff
@@ -0,0 +1,72 @@
...
+        builder.instance_exec(&block) if block
+
+        steps[step_name] = builder.to_step
+      end
+
+      def steps
+        @steps ||= {}
+      end
+
+      def step_route_constraint
```
kicferk1, line 30: Is `form` the place that this should live?

app/views/schools/add_participants/choose_mentor.html.erb
```diff
@@ -5,9 +5,10 @@
 <span class="govuk-caption-l"><%= @school.name %>, <%= @cohort.display_name %> cohort</span>
 
 <%= form_for add_participant_form, url: { action: :update }, method: :patch do |f| %>
-<%= f.govuk_radio_buttons_fieldset :mentor_id,
-  legend: { text: "Who is the mentor for this Early Career Teacher?", tag: 'h1', size: 'xl' },
-  hint: { text: "If you don't know, or haven't added them yet you can do this later." } do %>
+  <%= f.hidden_field :mentor_id, value: '' %>
```
kicferk1, line 8: Why do we need this hidden field?

kicferk1 (commented):
I will probably give it a more thorough look tomorrow, but I would really appreciate some request specs for the add participant flow - did I miss those and they already exist somewhere?

They would be useful to show that this setup works as expected in practice.

app/lib/multistep/controller.rb
BroiSatse: Definitely should not! :) Should be `form`

app/lib/multistep/controller.rb
BroiSatse: Definitely `form`

app/lib/multistep/controller.rb
BroiSatse: Yeah, I wasn't happy with the name, `default_back_route` sounds much better.

app/lib/multistep/controller.rb
BroiSatse: So yes, there are two `escape_routes` but in two different context. The first exists on the class, forming part of the public API to configure the behaviour of the controller. The other call is defined on the instance of the controller and executes the block given in the configuration in the context of the controller instance. The reason for this to always be a block is that neither url_helpers nor url_options are not available in the class context. Additionally, the back link can be dynamic (and there's already another PR that makes it dynamic for the add_participant journey.

`instance_exec` is an awesome method defined on `Object` used pretty much for meta-programming only - it executes given block in the context of the receiver (so changes the `self` inside the block), giving the block full access to the object internals, including access to its instance variables. As always, it needs to be used sparingly and only in certain scenarios.  I think it is applicable here as the block is to be defined within the same class, so it doesn't violate object encapsulation.

There's another similar method `instance_eval` which works very similarly with two important differences:
1) when executed with lambda instances passed as a block, `instance_eval` expects lambda to have arity of 1(yields object), while `instance_exec` expects arity of 0.
2). `instance_exec` additionally allows yielding more data into the block.

In our case, we're capturing the block resulting in a proc object, so 1) is not an issue, we also do not use 2) so we could use both. I went for `instance_exec` as I see no point of optional yield of self into the block.

Other options here would be:
1) Provide an overridable method. This however expects developer to know which method to override on the instance and requires the abstract, default implementation (in this case noop or to raise an exception). I used this pattern for `setup_form` but I think I should probably use declarative form there as well.
2) instead of instance_exec, we could do `yield self`. This would require declaration to be: escape_route { |controller| controller.url_helper }` which is a bit cumbersome (especially for more complicated logic) given that pretty much every single call would be prefixed with `controller.` and we'd loose access to private methods and variables (sort-of, we can always do `send` :P)

app/lib/multistep/form.rb
BroiSatse: Yup, probably not! I think I'd prefer another module `Multistep::Route` that is responsible for drawing all the routes with a single call, making this method its private implementation detail. I'll get this added.

app/views/schools/add_participants/choose_mentor.html.erb
BroiSatse: Without a hidden field, no params are sent on the form submission when no option is selected. This caused some issues with strong params (`school_add_participant_form param missing or empty` exception). Even if we skip strong params in those cases, there was a bit of an edge case when user comes back to this page (after submitting it previously), un-selects the radio and submits the form: since it subitted empty params the form is not changed, remains valid and proceeds forward without any change. It is really an edge case as it is not that easy to unselect a radio, but I thought it is probably best to have this fixed as well. If we were to use `govuk_radiu_buttons_collection` it would add that field automatically.

app/lib/multistep/controller.rb
kicferk1: Right. So it's like a hook that happens after the form is created but before it is used?

I think it's fine as is, but the comment should say something similar to what you said in the comment - what is this method for.

app/controllers/schools/add_participants_controller.rb
```diff
@@ -2,87 +2,25 @@
...
-    end
-
-    def complete
-      @participant_profile = add_participant_form.save!
-      @type = add_participant_form.type
-
-      session.delete(FORM_SESSION_KEY)
-    end
+    form AddParticipantForm, as: :add_participant_form
+    escape_route { schools_cohort_participants_path }
```
ethanmills, line 11: Does this not reintroduce the issue in #516 ?

app/forms/schools/add_participant_form.rb
```diff
@@ -98,7 +80,7 @@ def type=(value)
     end
 
     def school_cohort
-      @school_cohort ||= SchoolCohort.find(school_cohort_id)
+      @school_cohort ||= SchoolCohort.find_by(id: school_cohort_id)
```
ethanmills, line 83: Just curiosity now, why this change?

ethanmills (commented):
Looks nice! Just a couple of comments, then some things Michal asked for

app/controllers/schools/add_participants_controller.rb
BroiSatse: Yes, there will be a merge conflict to resolve on that. I'd leave it as is for now and will update the PR when the other one is merged.

app/forms/schools/add_participant_form.rb
BroiSatse: I've made the change which allows you to call `form.valid?` to execute all the steps validation. `find` was raising exception on validating a new form.

BroiSatse:
@kicferk1 As per request specs - again, I am on the fence here. We already have feature tests and I am not certain what request tests would give us. One thing that I'm definitely missing here are tests for `Multistep::Controller` which are pretty crucial, however rails is making it really tricky to test that and I hope to get them in a separate PR once the time allows.

kicferk1:
> @kicferk1 As per request specs - again, I am on the fence here. We already have feature tests and I am not certain what request tests would give us. One thing that I'm definitely missing here are tests for `Multistep::Controller` which are pretty crucial, however rails is making it really tricky to test that and I hope to get them in a separate PR once the time allows.

Request specs for the flow that uses the multistep form lets you test that `Multistep::Controller` behaves reasonably when used by an existing flow. We could also use those request tests as a baseline for when we want to add rspecs for `Multistep::Controller`. 

More importantly, we should really test what's happening in the controller when user tries to reach an endpoint without a session, or maybe without being an induction coordinator. Or as induction coordinator who told us they have no teachers this year. Or any other edge case we run into, which is too annoying and expensive to test in cucumber.

app/lib/multistep/controller.rb
```diff
@@ -0,0 +1,147 @@
+# frozen_string_literal: true
+
+module Multistep
+  module Controller
+    extend ActiveSupport::Concern
+
+    class_methods do
+      def form(value = nil, as: nil) # rubocop:disable Naming/MethodParameterName
```
kicferk1, line 8: What is Rubocop unhappy about exactly? And can we avoid it somehow without disabling it?

BroiSatse: He's unhappy with `as` being too short. I could replace it with something like `alias:` but I think `as` is used often enough to be pretty standard.

app/views/schools/add_participants/complete.html.erb
```diff
@@ -1,12 +1,12 @@
...
 
-<%= govuk_panel title: "You have added #{@participant_profile.user.full_name} to the #{@cohort.display_name} cohort", body: nil, classes: "govuk-!-margin-bottom-8" %>
+<%= govuk_panel title: "You have added #{participant_profile.user.full_name} to the #{@cohort.display_name} cohort", body: nil, classes: "govuk-!-margin-bottom-8" %>
 
 <h2 class="govuk-heading-m">What happens next</h2>
 <ul class="govuk-list govuk-list--bullet">
   <li>we will contact you if we need any more details from you such as their Teacher Reference Number (TRN)</li>
   <li>we will also update you on their status</li>
-  <li>you can view your early career teachers (ECTs) and mentors to see their status</li>
+  <li>you can view your early career teachers (ECTs) adnd mentors to see their status</li>
```
kicferk1, line 9: Hehe, I liked the old copy more

BroiSatse: I think... I mean... yeah...

app/controllers/schools/add_participants_controller.rb
BroiSatse: It is now corrected :)

app/controllers/schools/add_participants_controller.rb
```diff
@@ -5,7 +5,7 @@ class AddParticipantsController < ::Schools::BaseController
     include Multistep::Controller
 
     skip_after_action :verify_authorized
-    before_action :set_school_cohort
+    prepend_before_action :set_school_cohort
```
kicferk1, line 8: What's the difference?

BroiSatse: So that's the bit I am not very happy with - there is a before_action defined in `Multiform::Controller` that checks the sesssion is in correct state and redirects to the `abandon_journey_path` it is not. `abandon_journey_path` requires @school_cohort variable to be set to work - and this one is set in `:set_school_cohort`. Sooo, we need `set_school_cohort` to run before the `before_action` from `Multiform::Controller` and prepend is enforcing this.

Now, what's odd is that rails have `append_before_action` which I'd love to use for `ensure_form_present` check, the problem is that it is just an alias for `before_action` - rails is storing all the filters in a single array, so you cannot force for the filter to be last.

BroiSatse: Oh, actually one potential fix to this would be to move this redirection from before_action to the top of show and update actions. This is much better as it will allow other actions to be defined on the controller without triggering the filter.

kicferk1: Well, that's a bit unfortunate. We could (I guess) decouple `abandon_journey_path` from the `set_school_cohort` setup, but, ehhh, I am not sure I like it.

I will leave it up to your judgement :)

kicferk1 (approved):

app/controllers/schools/add_participants_controller.rb
kicferk1: Oooh, your solution sounds much better. Let's go for that!

PR #515, opened 2021-06-14
CPDTP-196 Fix error message returned by schema validator.

blackrat:
### Context

Error messages returned by the schema validator include "in schema <GUID_FOR_SCHEMA>" which contains more information than the caller actually needs.
Also, it has been decided to return a 422 if we don't have a valid participant_id (uuid) for the API call.

- Update errors returned by the validate_schema! to remove the superfluous part of the message.
- Change ect_profiles_params and valid_provider to ! versions and raise 422 with error messages for their failing calls.
- Use I18n to display the new participant_id error message
- Update swagger to reflect the updated 422 only messages on error and correct a mistake in the date parameter

### Guidance to review

add_ect_profile_params! and valid_provider! should now raise a 422 with the participant_id error message.
validate_schema! should iterate through all of the returned errors and truncate the messages
The error message for "invalid_participant" has been chosen to be the same as the format now returned by the schema validator.

### Review Checks

API only.

### DIFF
--- a/app/services/record_participant_event.rb
+++ b/app/services/record_participant_event.rb
@@ -20,7 +20,7 @@ class RecordParticipantEvent
     add_ect_profile_params!
     return :unprocessable_entity unless create_record
 
-    valid_provider!
+    validate_provider!
 
     :no_content
   end
@@ -58,7 +58,7 @@ private
     SchoolCohort.find_by(school: early_career_teacher_profile.school, cohort: early_career_teacher_profile.cohort)&.lead_provider
   end
 
-  def valid_provider!
+  def validate_provider!
     raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || lead_provider == actual_lead_provider
   end
 
--- a/app/services/record_participant_event.rb
+++ b/app/services/record_participant_event.rb
@@ -17,9 +17,10 @@ class RecordParticipantEvent
 
   def call
     validate_schema!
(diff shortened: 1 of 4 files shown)


app/services/record_participant_event.rb
```diff
@@ -17,9 +17,10 @@ def required_params
 
   def call
     validate_schema!
-    return :not_found unless add_ect_profile_params
+    add_ect_profile_params!
     return :unprocessable_entity unless create_record
-    return :not_found unless valid_provider
+
+    valid_provider!
```
kicferk1, line 23: Let's name it `validate_provider!` - it's no longer a method that returns a boolean, it's a method that sometimes raises an exception, so a verb over an adjective feels more fitting.

spec/docs/participant_declarations_spec.rb
```diff
@@ -25,7 +25,7 @@
                 "type": "string",
               },
               "declaration_type": {
-                "enum": %w[started],
+                "enum": %w[started retained_1 retained_2 retained_3 retained_4 completed],
```
kicferk1, line 28: Are we sure we want to add those to swagger schema?

blackrat: Only for consistency to show the applicable types since they also appear in the error messages when type is incorrect.

kicferk1 (approved):

timabell:
@blackrat what was the reason for "Use I18n to display the new participant_id error message" on this one?

PR #516, opened 2021-06-14
Fixed back link on the first add-participant step when there are no participants

BroiSatse:
When there are no participants, participant list page automatically redirects to the first step of the add-participant journey. Previously, link was fixed to return user to the participant index page, so it would be redirected back to the first step.

### DIFF
--- a/app/controllers/schools/add_participants_controller.rb
+++ b/app/controllers/schools/add_participants_controller.rb
@@ -64,10 +64,12 @@ module Schools
     end
 
     def back_link_path
-      previous_step = add_participant_form.previous_step(current_step)
-      return schools_cohort_participants_path unless previous_step
-
-      { action: :show, step: step_param(previous_step) }
+      if (previous_step = add_participant_form.previous_step(current_step))
+        { action: :show, step: step_param(previous_step) }
+      else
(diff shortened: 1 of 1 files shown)


app/controllers/schools/add_participants_controller.rb
```diff
@@ -64,10 +64,12 @@ def step_param(step)
...
 
     def back_link_path
-      previous_step = add_participant_form.previous_step(current_step)
-      return schools_cohort_participants_path unless previous_step
-
-      { action: :show, step: step_param(previous_step) }
+      if (previous_step = add_participant_form.previous_step(current_step))
+        { action: :show, step: step_param(previous_step) }
+      else
+        participants = User.order(:full_name).is_participant.in_school(@school.id)
```
ethanmills, line 70: Why the `order`?

ethanmills (approved):

app/controllers/schools/add_participants_controller.rb
BroiSatse: Because I just copied it from the participants_controller without really looking what I'm copying :P

BroiSatse: I'll merge this in for now (as it has literally no effect on anything) and remove when resolving with multistep form changes

PR #517, opened 2021-06-14
Nicer URLs for cohorts + schools

callumacrae:
### Context

Related to https://dfedigital.atlassian.net/browse/CPDRP-468 but might conflict with @ethanmills's current work

- Cohorts already used nice URLs but I've overridden `to_param` and removed `.start_year` from where the URLs were generated
- Schools now have slugs
- Some extremely minor visual changes

This PR will need to be half reverted for deployment (hence the commented out code in the middle) and done in two deploys, but ignore that for now when reviewing.

### Guidance to review

If reviewing locally, be aware that the migration will take like five minutes to run if you have all the schools downloaded

### Testing

I'm genuinely worried about how few changes this required to the tests - is it a good idea for them to be using the rails path helper functions? It feels to me like they shouldn't be and this should have broken a lot more tests.

Follow link in comment below

### DIFF
--- a/app/controllers/lead_providers/your_schools_controller.rb
+++ b/app/controllers/lead_providers/your_schools_controller.rb
@@ -6,11 +6,14 @@ module LeadProviders
 
     def index
       @cohorts ||= @lead_provider.cohorts
-      @selected_cohort = if params[:selected_cohort_id]
-                           @cohorts.find(params[:selected_cohort_id])
-                         else
-                           @cohorts.find_by(start_year: Time.zone.today.year)
-                         end
+
+      # Don't break old URLs
+      if params[:selected_cohort_id]
+        redirect_to cohort: params[:selected_cohort_id]
+        return
+      end
+
+      @selected_cohort = params[:cohort] ? @cohorts.find_by(start_year: params[:cohort]) : Cohort.current
 
       @partnerships = Partnership
         .includes(school: :early_career_teachers)
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -1,9 +1,13 @@
 # frozen_string_literal: true
 
 class School < ApplicationRecord
+  extend FriendlyId
+
   ELIGIBLE_TYPE_CODES = [1, 2, 3, 5, 6, 7, 8, 12, 14, 15, 18, 28, 31, 32, 33, 34, 35, 36, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48].freeze
   CIP_ONLY_TYPE_CODES = [10, 11, 30, 37].freeze
   ELIGIBLE_STATUS_CODES = [1, 3].freeze
+
+  friendly_id :slug_candidates
 
   belongs_to :network, optional: true
 
@@ -143,6 +147,19 @@ class School < ApplicationRecord
   def induction_tutor
     induction_coordinators.first
   end
+
+  def slug_candidates
+    [
+      :name,
+      %i[urn name],
+    ]
+  end
+
+# This temporary code is temporarily commented out during development
+# # This is temporary so that nothing breaks until the slugs are generated
+# def to_param
+#   id
+# end
 
 private
 
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -150,7 +150,6 @@ class School < ApplicationRecord
 
   def slug_candidates
     [
-      :name,
       %i[urn name],
     ]
   end
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -155,11 +155,10 @@ class School < ApplicationRecord
     ]
   end
 
-# This temporary code is temporarily commented out during development
-# # This is temporary so that nothing breaks until the slugs are generated
-# def to_param
-#   id
-# end
+  # This is temporary so that nothing breaks until the slugs are generated
+  def to_param
+    id
+  end
 
 private
 
--- a/app/views/lead_providers/your_schools/index.html.erb
+++ b/app/views/lead_providers/your_schools/index.html.erb
@@ -18,10 +18,13 @@
   </div>
 </div>
 
-<%= render SubnavComponent.new do |component| %>
-  <% @cohorts.each do |cohort| %>
-    <%= component.nav_item(path: lead_providers_your_schools_path(selected_cohort_id: cohort.id)) do %>
-      <%= cohort.start_year %> cohort
+<% if @cohorts.length > 1 %>
+  <%= render SubnavComponent.new do |component| %>
+    <% @cohorts.each do |cohort| %>
+      <%# Use a query param to access cohorts other than the current one %>
+      <%= component.nav_item(path: lead_providers_your_schools_path(cohort: cohort == Cohort.current ? nil : cohort)) do %>
+        <%= cohort.start_year %> cohort
+      <% end %>
     <% end %>
   <% end %>
 <% end %>
--- a/config/initializers/friendly_id.rb
+++ b/config/initializers/friendly_id.rb
@@ -1,0 +1,119 @@
+# frozen_string_literal: true
+
+# FriendlyId Global Configuration
+#
+# Use this to set up shared configuration options for your entire application.
+# Any of the configuration options shown here can also be applied to single
+# models by passing arguments to the `friendly_id` class method or defining
+# methods in your model.
+#
+# To learn more, check out the guide:
+#
+# http://norman.github.io/friendly_id/file.Guide.html
+
+FriendlyId.defaults do |config|
+  # ## Reserved Words
+  #
+  # Some words could conflict with Rails's routes when used as slugs, or are
+  # undesirable to allow as slugs. Edit this list as needed for your app.
+  config.use :reserved
+
+  config.reserved_words = %w[new
+                             edit
+                             index
+                             session
+                             login
+                             logout
+                             users
+                             admin
+                             stylesheets
+                             assets
+                             javascripts
+                             images]
+
+  # This adds an option to treat reserved words as conflicts rather than exceptions.
+  # When there is no good candidate, a UUID will be appended, matching the existing
+  # conflict behavior.
+
+  # config.treat_reserved_as_conflict = true
+
+  #  ## Friendly Finders
+  #
+  # Uncomment this to use friendly finders in all models. By default, if
+  # you wish to find a record by its friendly id, you must do:
+  #
+  #    MyModel.friendly.find('foo')
+  #
+  # If you uncomment this, you can do:
+  #
+  #    MyModel.find('foo')
+  #
+  # This is significantly more convenient but may not be appropriate for
+  # all applications, so you must explicity opt-in to this behavior. You can
+  # always also configure it on a per-model basis if you prefer.
+  #
+  # Something else to consider is that using the :finders addon boosts
+  # performance because it will avoid Rails-internal code that makes runtime
+  # calls to `Module.extend`.
+  #
+  # config.use :finders
+  #
+  # ## Slugs
+  #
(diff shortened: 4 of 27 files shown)


app/models/school.rb
```diff
@@ -144,6 +148,19 @@ def induction_tutor
...
+      :name,
+      %i[name urn],
+    ]
+  end
+
+# This temporary code is temporarily commented out during development
+# # This is temporary so that nothing breaks until the slugs are generated
+# def to_param
+#   id
+# end
```
kicferk1, line 162: mhmmmm.

Development is over, do we uncomment or remove it?

app/models/school.rb
```diff
@@ -144,6 +148,19 @@ def induction_tutor
     induction_coordinators.first
   end
 
+  def slug_candidates
+    [
+      :name,
+      %i[name urn],
+    ]
+  end
```
kicferk1, line 156: So what can our urls look like? Are we sure school name is something we want in the url?

app/views/lead_providers/your_schools/index.html.erb
```diff
@@ -18,10 +18,13 @@
...
 </div>
 
-<%= render SubnavComponent.new do |component| %>
-  <% @cohorts.each do |cohort| %>
-    <%= component.nav_item(path: lead_providers_your_schools_path(selected_cohort_id: cohort.id)) do %>
-      <%= cohort.start_year %> cohort
+<% if @cohorts.length > 1 %>
+  <%= render SubnavComponent.new do |component| %>
+    <% @cohorts.each do |cohort| %>
+      <%# The link to the current cohort doesn't have it in the URL %>
```
kicferk1, line 24: Mhmmm, if we care about it enough to make a cryptic comment about it, should we add an automated test for it (cypress?)

app/views/lead_providers/your_schools/index.html.erb
```diff
@@ -18,10 +18,13 @@
...
+      <%= component.nav_item(path: lead_providers_your_schools_path(cohort: cohort.start_year === Time.zone.today.year ? nil : cohort)) do %>
```
kicferk1, line 25: One day I will add something that crashes the server if it sees `===` outside of js. Until that day, I will simply comment saying that this is not js land, we should use `==` 😂

db/migrate/20210614145351_create_friendly_id_slugs.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+MIGRATION_CLASS =
+  if ActiveRecord::VERSION::MAJOR >= 5
+    ActiveRecord::Migration["#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}"]
+  else
+    ActiveRecord::Migration
+  end
```
kicferk1, line 8: We know which version we want, just copy it from another migration instead relying on what friendly brought with it :)

db/migrate/20210614145351_create_friendly_id_slugs.rb
```diff
@@ -0,0 +1,32 @@
...
+    add_index :friendly_id_slugs, %i[slug sluggable_type scope], length: { slug: 70, sluggable_type: 50, scope: 70 }, unique: true
+
+    add_column :schools, :slug, :string
+    add_index :schools, :slug, unique: true
+
+    # This takes > 4 minutes to run which fails deploys - run manually on
+    # deployed environments
+    if Rails.env.development? || Rails.env.test?
+      School.find_each(&:save)
+    end
```
kicferk1, line 30: uhhhhhhh, I would really prefer not to have to run it manually. Can we bump the time limit? What is failing exactly? Our github action? paas?

app/models/school.rb
callumacrae: this will be uncommented (and half the code removed) before merge but after review

kicferk1: Wait, I haven't read it properly.

Will we need another PR removing this method after we have generated slugs in all environments?

app/models/school.rb
callumacrae: two example urls:

/admin/schools/buckingham-primary-school
/admin/schools/buckingham-primary-school-110329

why would we not want school name in URL?

app/views/lead_providers/your_schools/index.html.erb
callumacrae: is the comment cryptic? 😅

app/views/lead_providers/your_schools/index.html.erb
callumacrae: :lolsob:

db/migrate/20210614145351_create_friendly_id_slugs.rb
callumacrae: dunno, this was decided after conversations with @ethanmills and @jamesrose. I was way out of my depth in them, so this is probably a question that's best for them

app/models/school.rb
kicferk1: I thought names are long, annoying, can have special characters in them. But if we are sure we want to have names in the url, then it's alright.

app/controllers/lead_providers/your_schools_controller.rb
```diff
@@ -6,11 +6,14 @@ class YourSchoolsController < ::LeadProviders::BaseController
...
-                           @cohorts.find_by(start_year: Time.zone.today.year)
-                         end
+
+      # Don't break old URLs
+      if params[:selected_cohort_id]
+        redirect_to cohort: params[:selected_cohort_id]
+        return
+      end
+
+      @selected_cohort = @cohorts.find_by(start_year: params[:cohort] || Time.zone.today.year)
```
ethanmills, line 16: I know I know, it wasn't you... Still can we make use of `Cohort.current` instead of the current year? This will break in January

app/views/lead_providers/your_schools/index.html.erb
ethanmills, line 25: What happens 1st Jan 2022?

ethanmills: Also, why do we not include it for the current cohort?

config/initializers/friendly_id.rb
```diff
@@ -0,0 +1,119 @@
...
+  # config.use :slugged
+  #
+  # By default, FriendlyId's :slugged addon expects the slug column to be named
+  # 'slug', but you can change it if you wish.
+  #
+  # config.slug_column = 'slug'
+  #
+  # By default, slug has no size limit, but you can change it if you wish.
+  #
+  # config.slug_limit = 255
```
ethanmills, line 75: I think we're going to want a size limit. Some schools have veeeeery long names

ethanmills: Wyvil Primary School and Resource Bases for Speech, Language and Communication Needs, and Autism

db/migrate/20210614145351_create_friendly_id_slugs.rb
ethanmills: What's wrong with running this manually? 60s is the limit on a deploy going green - has to 200 from /check within this time

app/models/school.rb
callumacrae: the library we're using deals with special characters, but yeah, they can be pretty long

app/models/school.rb
callumacrae: yep!

app/views/lead_providers/your_schools/index.html.erb
callumacrae: it breaks! will fix

I changed it to not include it to fix the bug where no tab would be selected when initially viewing the page:

image

alternative fix would have been to always ensure the cohort is in the URL, but I felt this was a nicer solution

db/migrate/20210614145351_create_friendly_id_slugs.rb
kicferk1: Anything that needs to be ran manually is not ran at some point.

Although I guess for all new environments, the post create hook (which this is triggering) should run as part of the school import job, so we don't need this in the migration. Is that correct?

If so, it sounds like bumping the time for paas is not worth the effort, so let's keep it as is and make sure we remember to run slugs on dev, staging, sandbox and prod.

callumacrae: It _only_ needs running on production, staging, and probably sandbox. The develop branches are reset not migrated, and this will run on people's local machines.

app/views/lead_providers/your_schools/index.html.erb
kicferk1: A little bit. Doesn't have it in the path param? Doesn't have it as a query param? Is that documenting a bug or intended outcome? 

Those are the first few questions I have when reading it :D

db/migrate/20210614145351_create_friendly_id_slugs.rb
callumacrae: It will also cause the site to fail pretty catastrophically if it isn't ran, so it's not like it will be causing subtle issues for a while that won't get noticed

app/views/lead_providers/your_schools/index.html.erb
callumacrae: it's a query param - i'll make the comment a bit clearer

kicferk1 (approved):
LGTM. Is it time to uncomment the thing that makes it work without running the 4-min long after-save hooks on each school?

ethanmills (approved):

BroiSatse (approved):

tonyheadford (commented):
Looks good except - not sure why we need a gem to handle this?  Have we considered what happens when a school changes name or URN ?  We just used URN for the slug on GHWT without any uniqueness issues.  @mattonfoot had some ideas about managing changes to URNs when the topic was discussed briefly on slack a while ago.

callumacrae:
> Looks good except - not sure why we need a gem to handle this?

i'll come back to this…

> Have we considered what happens when a school changes name or URN ?

friendly_id has some history functionality that presumably can be used to change slugs and handle redirects from the old ones, i just haven't done anything with it in this PR - this isn't the full work, though, just a PR-sized chunk of it

stuff like this is why we're using a gem :)

> We just used URN for the slug on GHWT without any uniqueness issues. @mattonfoot had some ideas about managing changes to URNs when the topic was discussed briefly on slack a while ago.

There aren't any uniqueness issues! We use the URN when the (potentially truncated) school name is already used as slug - imo it's better to use english words over numbers in URLs wherever possible

tonyheadford:
> friendly_id has some history functionality that presumably can be used to change slugs and handle redirects from the old ones, i just haven't done anything with it in this PR - this isn't the full work, though, just a PR-sized chunk of it
> 
Ah OK 👍 
> There aren't any uniqueness issues! We use the URN when the (potentially truncated) school name is already used as slug - imo it's better to use english words over numbers in URLs wherever possible
Just a quick look in prod, out of the eligible schools (25417) there are 23533 with unique names so ~1000 with duplicate school names. Also if names are truncated then that's not great for a resourceful url, URN on its own would be far better imo.

ethanmills:
IMO a URN is about as human friendly as a GUID. I'm for using an intelligible portion of the name

tonyheadford:
> IMO a URN is about as human friendly as a GUID. I'm for using an intelligible portion of the name

It's a 6 digit number?  There will be no consistency if you are munging or truncating school names will there?

callumacrae:
after discussion on hangouts, i'm going to switch this to always use urn and then school name, and then add the ability to just put the urn in the url in another PR (to get the conflicts with ethan's PR out of the way asap)

PR #518, opened 2021-06-15
Retrieve each user's school induction programme choice.

xswoodsx:
### Context
Engage and Learn want to know whether a user is doing fip or cip

Adding a induction programme choice attribute to the user data sent to E&L
Additional users in dummy structures

### Guidance to review
I used the the participant method on user model because it's doing what I wanted, although i don't know why it's called participant. 

I'm not 100% school_cohort is retrieving exactly what we're after but SchoolCohort.find_by(school: user.school, cohort: user.cohort) kept returning nil class. 

### Testing
Updated tests

### Review Checks

You can't but run the app and use insomnia or postman to check that the api is responding with the correct attributes.


app/serializers/user_serializer.rb
```diff
@@ -44,4 +50,18 @@ class UserSerializer
       CIP_TYPES[:none]
     end
   end
+
+  attributes :induction_programme_choice do |user|
+    if user.participant?
+      school_cohort = SchoolCohort.find_by(school: user.school)
```
kicferk1, line 56: I think we should specify `cohort` here. You should be able to get it via 
```
  def cohort
    return early_career_teacher_profile.cohort if early_career_teacher?
    return mentor_profile.cohort if mentor?
  end
```
on the `user` model. If the `cohort` ends up `nil` it probably means our seeds don't specify cohort for a particular ECT or mentor, and we should fix that.

app/serializers/user_serializer.rb
```diff
@@ -44,4 +50,18 @@ class UserSerializer
...
+      case school_cohort.induction_programme_choice
+      when "full_induction_programme"
+        PROGRAMME_TYPES[:full_induction_programme]
+      when "core_induction_programme"
+        PROGRAMME_TYPES[:core_induction_programme]
+      when "not_yet_known"
```
kicferk1, line 62: Let's make it `else`?

db/seeds/dummy_structures.rb
```diff
@@ -44,13 +59,27 @@
   end
 end
 
+User.find_or_create_by!(email: "rp-mentor-edt@example.com") do |user|
+  user.update!(full_name: "Jane Doe")
+  MentorProfile.find_or_create_by!(user: user) do |profile|
+    profile.update!(school: school_two, core_induction_programme: CoreInductionProgramme.find_by(name: "Education Development Trust"))
+  end
```
kicferk1, line 66: That's where you should add the cohort :)

db/seeds/dummy_structures.rb
```diff
@@ -44,13 +59,27 @@
...
   EarlyCareerTeacherProfile.find_or_create_by!(user: user) do |profile|
     profile.update!(school: school, core_induction_programme: CoreInductionProgramme.find_by(name: "Ambition Institute"), mentor_profile: mentor.mentor_profile)
   end
 end
 
+User.find_or_create_by!(email: "rp-ect-edt@example.com") do |user|
+  user.update!(full_name: "John Doe")
+  EarlyCareerTeacherProfile.find_or_create_by!(user: user) do |profile|
+    profile.update!(school: school_two, core_induction_programme: CoreInductionProgramme.find_by(name: "Education Development Trust"), mentor_profile: mentor.mentor_profile)
+  end
```
kicferk1, line 80: And here, and in any other place in seed files that don't do it.

app/serializers/user_serializer.rb
```diff
@@ -17,6 +17,12 @@ class UserSerializer
     none: "none",
   }.freeze
 
+  PROGRAMME_TYPES = {
+    core_induction_programme: "core_induction_programme",
+    full_induction_programme: "full_induction_programme",
+    not_yet_known: "not_yet_known",
+  }.freeze
```
ethanmills, line 24: Is there a reason not just to use `SchoolCohort.induction_programme_choices`?

app/serializers/user_serializer.rb
```diff
@@ -44,4 +50,18 @@ class UserSerializer
...
+      case school_cohort.induction_programme_choice
+      when "full_induction_programme"
+        PROGRAMME_TYPES[:full_induction_programme]
+      when "core_induction_programme"
+        PROGRAMME_TYPES[:core_induction_programme]
+      when "not_yet_known"
+        PROGRAMME_TYPES[:not_yet_known]
+      end
+    end
+  end
```
ethanmills, line 66: This is a noop, right?

spec/requests/api/v1/users_spec.rb
```diff
@@ -12,11 +12,16 @@
     before :each do
       # Heads up, for some reason the stored CIP IDs don't match
       cip = create(:core_induction_programme, name: "Teach First")
+      cohort = create(:cohort)
```
ethanmills, line 15: You don't need this, the school_cohort factory will create it's own cohort if you don't provide one

db/seeds/sandbox_data.rb
```diff
@@ -38,14 +38,14 @@ def create_school_and_associations(lead_provider, cohort, index)
     s.postcode = Faker::Address.postcode
   end
 
-  Partnership.create!(
+  Partnership.find_or_create_by!(
```
ethanmills, line 41: This is never going to match, because the delivery partner is randomly generated, right?

app/serializers/user_serializer.rb
kicferk1: Is it? I think the `case when` statement will return some string (symbol?), then if statement will return what the case statement returns, and the block will return what if returns. I hope. We could probably add an explicit `return`, but I am not sure rubocop would be happy with it.

ethanmills: My point was this was the same as returning `school_cohort.induction_programme_choice`. Sorry that wasn't clear

ethanmills: The Case specifically is an identity mapping

kicferk1: Oh, right. Well, not quite, I think this case will return some `nil`s when `induction_programme_choice` doesn't, but I agree we can probably just return `school_cohort .induction_programme_choice` here.

Although in principle that means changing the values of `induction_programme_choice` enum will be harder, I don't think it's likely to ever happen.

ethanmills: I think we would add this back in if we changed it, but it's just odd to have it here doing nothing

db/seeds/sandbox_data.rb
```diff
@@ -45,10 +45,11 @@ def create_school_and_associations(lead_provider, cohort, index)
     school: school,
     lead_provider: lead_provider,
     cohort: cohort,
-    delivery_partner: DeliveryPartner.find_or_create_by!(name: Faker::Company.name),
-  )
+  ) do |partnership|
+    partnership.delivery_partner = DeliveryPartner.create!(name: Faker::Company.name)
+  end
```
kicferk1, line 50: Heh, are these meaningfully different? maybe just revert to the original change?

kicferk1 (approved):
LGTM, one small comment about the seed file change

db/seeds/sandbox_data.rb
kicferk1: Ah, they are, delivery partner will be different each time so we don't want to try to find things by it. Makes sense.

ethanmills: Yes, these are different. This block only runs for the create part

PR #519, opened 2021-06-15
Copy changes for participants flow

callumacrae:
### Context

Made these changes with Shani earlier!

click link in comment below

### DIFF
--- a/app/views/schools/participants/show.html.erb
+++ b/app/views/schools/participants/show.html.erb
@@ -1,4 +1,4 @@
-<% content_for :title, "Early career teachers and mentors" %>
+<% content_for :title, @participant.full_name %>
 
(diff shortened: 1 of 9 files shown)


kicferk1 (approved):

app/views/schools/participants/show.html.erb
```diff
@@ -1,4 +1,4 @@
-<% content_for :title, "Early career teachers and mentors" %>
+<% content_for :title, @participant.full_name %>
```
ethanmills, line 1: Do we need to worry about sending PPI to GA?

ethanmills (approved):
One question

app/views/schools/participants/show.html.erb
callumacrae: good catch!

PR #520, opened 2021-06-15
API endpoint to create users

asmega:
### Context

- NPQ registration collects user data which must be pushed into "CPD manage"
- There will be subsequent pull requests to add attach NPQ profiles and data to these users

- Add a barebones api endpoint to create a user with name and email
- We use email as the "identifier" here

### Guidance to review

- Existing api points should not be affected
- Should required token with private api access permissions
- Should conform to jsonapi spec

### Testing

- Test in the review app

### Review Checks

- Note down the token
- create a user with `curl -H "Content-Type: application/vnd.api+json" -H "Authorization: Bearer TOKEN" -X POST -d '{ "data": { "type": "user", "attributes": { "full_name": "John Doe", "email": "john.doe@example.com" } } }' https://ecf-review-pr-520.london.cloudapps.digital/api/v1/users`
- response should be 201
- response should have user guid so client can associate with the created user record

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -10,6 +10,32 @@ module Api
         render json: UserSerializer.new(paginate(users)).serializable_hash.to_json
       end
 
+      def create
+        user = User.find_by(email: params[:data][:attributes][:email])
+
+        if user.present?
+          user.errors.add(:email, :taken)
+
+          hash = {
+            errors: [{
+              status: "409",
+              title: user.errors.full_messages.join(", "),
+            }],
+          }
+
+          render json: hash, status: :conflict and return
+        end
+
+        user = User.create!(
+          email: params[:data][:attributes][:email],
+          full_name: params[:data][:attributes][:full_name],
+        )
+
+        hash = UserSerializer.new(user).serializable_hash
+
+        render json: hash, status: :created
+      end
+
     private
 
       def access_scope
--- a/spec/requests/api/v1/users_spec.rb
+++ b/spec/requests/api/v1/users_spec.rb
@@ -52,17 +52,9 @@ RSpec.describe "API Users", type: :request do
 
       it "returns correct user types" do
         get "/api/v1/users"
-        mentors = 0
-        ects = 0
 
-        parsed_response["data"].each do |user|
-          user_type = user["attributes"]["user_type"]
-          if user_type == "mentor"
-            mentors += 1
-          elsif user_type == "early_career_teacher"
-            ects += 1
-          end
-        end
+        mentors = parsed_response["data"].count { |hash| hash["attributes"]["user_type"] == "mentor" }
+        ects = parsed_response["data"].count { |hash| hash["attributes"]["user_type"] == "early_career_teacher" }
 
         expect(mentors).to eql(1)
         expect(ects).to eql(2)
--- a/spec/requests/api/v1/users_spec.rb
+++ b/spec/requests/api/v1/users_spec.rb
@@ -3,12 +3,11 @@
 require "rails_helper"
 
 RSpec.describe "API Users", type: :request do
+  let(:parsed_response) { JSON.parse(response.body) }
+  let(:token) { EngageAndLearnApiToken.create_with_random_token! }
+  let(:bearer_token) { "Bearer #{token}" }
+
   describe "#index" do
-    let(:token) { EngageAndLearnApiToken.create_with_random_token! }
-    let(:bearer_token) { "Bearer #{token}" }
-
-    let(:parsed_response) { JSON.parse(response.body) }
-
     before :each do
       # Heads up, for some reason the stored CIP IDs don't match
       cip = create(:core_induction_programme, name: "Teach First")
@@ -71,12 +70,14 @@ RSpec.describe "API Users", type: :request do
         expect(parsed_response["data"].size).to eql(2)
       end
 
-      it "returns different users for each page" do
+      it "returns different users for first page" do
         get "/api/v1/users", params: { page: { per_page: 2, page: 1 } }
         expect(parsed_response["data"].size).to eql(2)
+      end
 
+      it "returns different users for second page" do
         get "/api/v1/users", params: { page: { per_page: 2, page: 2 } }
-        expect(JSON.parse(response.body)["data"].size).to eql(1)
+        expect(parsed_response["data"].size).to eql(1)
       end
 
       it "returns users changed since a particular time, if given a changed_since parameter" do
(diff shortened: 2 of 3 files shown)


db/migrate/20210615101632_add_dob_and_trn_to_users.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddDobAndTrnToUsers < ActiveRecord::Migration[6.1]
+  def change
+    change_table(:users, bulk: true) do |t|
+      t.date :date_of_birth, null: true
+      t.text :teacher_reference_number, null: true
+    end
+  end
```
kicferk1, line 9: I don't think those should live on `users` table - we have admins, induction coordinators, lead providers there.

Should we make a new profile for this?

db/migrate/20210615103220_add_trn_verified_to_users.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddTrnVerifiedToUsers < ActiveRecord::Migration[6.1]
+  def change
+    add_column :users, :teacher_reference_number_verified, :boolean, null: false, default: false
```
kicferk1, line 5: Likewise, `users` is a much bigger space than we likely want for this boolean.

db/migrate/20210615101632_add_dob_and_trn_to_users.rb
asmega: ok, i'll remove all the extra stuff. the endpoint will be just to create a user. i'll do another PR to be able to create a npq profile with all the stuff i just removed (plus there will be other stuff not present here)

db/migrate/20210615103220_add_trn_verified_to_users.rb
asmega: this is gone now and will be in a different PR

spec/requests/api/v1/users_spec.rb
```diff
@@ -105,11 +98,64 @@
...
+      let(:json) do
+        {
+          data: {
+            type: "user", # is this singular or plural??? # spec is agnostic # i chose singular
+            attributes: {
+              full_name: "John Doe",
+              email: "john.doe@example.com",
+            },
+          },
+        }.to_json
```
kicferk1, line 125: That's what we need to send to the api?

I don't really understand why we would not go with something like
```
{
  full_name: "John Doe",
  email: "john.doe@example.com"
}
```
Everything else seems to be, well, a bit excessive. In particular, why do we send a `user` type to an api with `users` in the path - will the api not know that it expects user data? Will it do something else if we send it `school` type? 

`participant-declarations` api, at least judging by their documentation, goes with the minimalist approach. Their swagger example is
```
{
  "participant_id": "db3a7848-7308-4879-942a-c4a70ced400a",
  "declaration_date": "2021-05-31T02:21:32Z"
}
```
Thoughts?

asmega: my understanding is we are conforming to the jsonapi spec. which documents here https://jsonapi.org/format/#crud-creating for the shape of what creating a resource is. my change should conform and the change you are suggesting would make it not conform. or are we not conforming to jsonapi spec?

kicferk1: It's our internal API, we can choose to make it whatever we want it to be. Is conforming to the `jsonapi` spec making life easier for us? In this PR it probably is not, cause you need to do
```
        user = User.find_or_create_by!(email: params[:data][:attributes][:email]) do |u|
          u.full_name = params[:data][:attributes][:full_name]
        end
```
to get the relevant data. And the users controller happily ignores the `type` parameter that you are sending its way, which makes me think the spec was written for something much more generic than our use case - say an endpoint that needs to handle creating many, many types of objects.

If we really think that conforming to that spec is important, then we should probably install some gems that do it for us, like what `teacher-training-api` are doing. And probably make some changes to participant details api :P 

Something to discuss on dev catchup perhaps?

app/controllers/api/v1/users_controller.rb
```diff
@@ -10,6 +10,16 @@ def index
...
 
+      def create
+        user = User.find_or_create_by!(email: params[:data][:attributes][:email]) do |u|
+          u.full_name = params[:data][:attributes][:full_name]
+        end
+
+        hash = UserSerializer.new(user).serializable_hash
+
+        render json: hash, status: :created
+      end
```
ethanmills, line 21: I don't know what your requirements are, but I find this surprising that we respond with 201 created when we match an existing record. I'm not sure what would be better here, perhaps just 200, or maybe 409?

spec/requests/api/v1/users_spec.rb
```diff
@@ -105,11 +98,64 @@
...
+        post "/api/v1/users.json", params: json
+        expect(response.headers["Content-Type"]).to eql("application/vnd.api+json")
+      end
+
+      it "creates a user record" do
+        expect {
+          post "/api/v1/users", params: json
+        }.to change(User, :count).by(1)
+
+        user = User.last
```
ethanmills, line 143:
```suggestion
        user = User.order(:created_at).last
```

spec/requests/api/v1/users_spec.rb
```diff
@@ -105,11 +98,64 @@
...
+
+        expect(user.full_name).to eql("John Doe")
+        expect(user.email).to eql("john.doe@example.com")
+      end
+
+      it "returns the created user resource" do
+        post "/api/v1/users.json", params: json
+
+        user = User.last
```
ethanmills, line 152:
```suggestion
        user = User.order(:created_at).last
```

app/controllers/api/v1/users_controller.rb
asmega: i can do the extra check and return 201 if it is new and 200 if it already exists

spec/requests/api/v1/users_spec.rb
kicferk1: Alright, so to recap, we will do our best to conform to the jsonapi spec. It means there will be some unwieldiness, like here, but hopefully the consistency will be worth it.

app/controllers/api/v1/users_controller.rb
ethanmills: Only if you think it would be useful to distinguish, just wanted to make sure it was considered 🙂

spec/requests/api/v1/users_spec.rb
asmega: done

spec/requests/api/v1/users_spec.rb
asmega: done

spec/requests/api/v1/users_spec.rb
```diff
@@ -105,11 +98,81 @@
...
+    context "when authorized" do
+      before do
+        default_headers[:Authorization] = bearer_token
+        default_headers["Content-Type"] = "application/vnd.api+json"
+      end
+
+      let(:json) do
+        {
+          data: {
+            type: "user", # is this singular or plural??? # spec is agnostic # i chose singular
```
kicferk1, line 119: I would go for singular, and remove the comment,

app/controllers/api/v1/users_controller.rb
asmega: - ok i changed my mind. so i kind of purposefully glossed over this particular implementation detail as I know it would open a can of worms.
- `409` works better i feel because the `full_name` field is also passed over
- I feel there is ambiguity if a 200 be returned whether `full_name` gets updated or not
- so instead we simply return a `409` if a user is found with the email address given. what do you think? 
- the current implementation lacks but i think is fine for now. it is missing error info and should have but will implement as a later date the ability to link to the conflicting record so the client can fetch that existing record instead. we do not have such an api endpoint so this also would not be possible.

spec/requests/api/v1/users_spec.rb
asmega: yep sorry, I chopped of that commit when i was removing the migrations

kicferk1 (approved):

ethanmills (approved):

PR #523, opened 2021-06-15
CPDRP-345: Add school ID to urls

ethanmills:
I just really want merge conflicts with @callumacrae 

### Context
Believe it or not, this is actually me splitting up my PRs. This is a big change, but I don't really see how it can be smaller. I'm adding the school ID to all the school routes in preparation for induction coordinators having more than one school (future PR)

 - Add school id/slug segment to all relevant routes
 - Get school from path, not induction coordinator's first
 - Slug those routes
 - Use breadcrumbs instead of back links

### Guidance to review
Get a coffee? Sorry

### Testing
Lots of unit tests and cypress tests updated here. I've probably missed a few, since `School.friendly.find` accepts ID as well as slug...

### Review Checks


callumacrae:
image
 
my stuff isn't even merged yet!

ethanmills:
I've just seen that I'm exposing some random test data. Will fix

app/views/schools/participants/index.html.erb
```diff
@@ -1,14 +1,13 @@
 <% content_for :title, "Early career teachers and mentors" %>
 
-<% content_for :before_content, govuk_back_link(
-  text: "Back",
-  href: schools_cohort_path(@cohort.display_name))
-%>
+<% content_for :before_content, govuk_breadcrumbs(
+  breadcrumbs: { @school.name => schools_dashboard_path, "#{@cohort.display_name} cohort" => schools_cohort_path }
```
kicferk1, line 4: The breadcrumbs seem to repeat a lot. On E&L we made a helper to generate them for CIP materials. Maybe not worth doing here, but something to think about when we touch them next.

config/routes.rb
```diff
@@ -184,41 +184,45 @@
...
+          resource :programme, only: %i[edit], controller: "choose_programme"
+
+          resources :participants, only: %i[index show] do
+            get :edit_details, path: "edit-details"
+            get :edit_mentor, path: "edit-mentor"
+            put :update_mentor, path: "update-mentor"
+
+            collection do
+              multistep_form :add,
+                             Schools::AddParticipantForm, controller: :add_participants
```
kicferk1, line 212: If we do multiline, maybe we want brackets around it?

spec/cypress/integration/ChallengePartnership.feature
```diff
@@ -27,20 +27,21 @@ Feature: Reporting an error with a partnership
...
 
-    When I navigate to "2021 school partnerships" page
+    When I navigate to "2021 school partnerships" page with id "0000bd75-31d0-4eb3-8df3-07f866e41d51"
     Then "page body" should contain "Signing up with a training provider"
 
+    @focus
   Scenario: A logged in induction tutor for CIP school challenges partnership
     Given I am logged in as existing user with email "test-subject2@example.com"
-    And I am on "2021 school cohorts" page
+    And I am on "2021 school cohorts" page with id "00041221-d612-46a8-a096-87ad63ff3a7d"
```
kicferk1, line 44: Hmm, could we use pretty url slugs here instead of ids?

kicferk1: I guess Callum was adding that change, and you will have a lot of conflicts with him 😂

config/routes.rb
ethanmills: I'm happy to inline this again, snuck in in a merge

kicferk1 (commented):
Looks like still some more randomness.
```
<span class="govuk-caption-l"><%= @school.name %></span>
```
in the dashboard page is one. 
`Choose no early career teachers success` snapshot has another of those I think.

app/views/schools/participants/index.html.erb
ethanmills: Will take a look in the next one, I'm doing some breadcrumb logic there anyway

kicferk1 (approved):
A step 
```
And I am on "2021 school participants" page with id "00041221-d612-46a8-a096-87ad63ff3a7d"
```
can probably be replaced with a slug one, other than that lgtm

PR #524, opened 2021-06-15
Selective ff selected objects

BroiSatse:
Our FeatureFlags are currently global, but we want to be able to enable some features for selected schools.

This PR introduces a generic way of handling this type of requests by connecting feature to any record in the database. To enable given feature per record, simply pass a record via `for:` to `FeatureFlag.activate(:feature, for: <record>)` and later on check the flag is enabled for a given record `FeatureFlag.active?(:feature, for: <record>)`. `active?` will ignore `for` argument if the feature is enabled globally.


app/services/feature_flag.rb
```diff
@@ -26,22 +26,32 @@ def feature
...
-    sync_with_database(feature_name, true)
+    if opts.key?(:for).present?
+      sync_with_database_with_object(feature_name, opts[:for], true)
+    else
+      sync_with_database(feature_name, true)
+    end
   end
 
-  def self.deactivate(feature_name)
+  def self.deactivate(feature_name, **opts)
```
ethanmills, line 39: Do we need a way to easily deactivate for all? E.g. if we start rollout of a feature that is so broken we want to turn it off

ethanmills (approved):

ethanmills:
We should update the specs to test this I think

app/services/feature_flag.rb
BroiSatse: Good question. Another one I had in mind is if we would like to globally activate the feature with just a few exceptions? But that is much more edgy.

For now I'll add `for: :all` option that destroys all the selected objects as well.

app/services/feature_flag.rb
```diff
@@ -50,6 +60,17 @@ def self.sync_with_database(feature_name, active)
...
+  def self.sync_with_database_with_object(feature_name, object, active)
+    ActiveRecord::Base.transaction do
+      feature = Feature.find_or_create_by!(name: feature_name)
+      if active
+        feature.selected_objects.find_or_create_by!(object: object)
+      else
+        feature.selected_objects.where(object: object).destroy_all
+      end
+    end
+  end
```
kicferk1, line 72: Right, so an existence of the selected object is equivalent with the flag being on for whatever it is pointing at?

config/routes.rb
```diff
@@ -198,7 +198,7 @@
       resources :partnerships, only: :index
       resource :programme, only: %i[edit], controller: "choose_programme"
 
-      resources :participants, only: %i[index show], constraints: ->(_request) { FeatureFlag.active?(:induction_tutor_manage_participants) } do
+      resources(:participants, only: %i[index show]) do
```
kicferk1, line 201: Can we use spaces instead of brackets (like in other places in this file?)

BroiSatse: Yup, I was trying to preserve the constraint which forced me to split the call into multiple lines, hence brackets. They need to go though.

app/services/feature_flag.rb
BroiSatse: In the first iteration, yes. I'd guess later on we could put some more data on that (and make it for example an exception of otherwisee activated feature) but for now its presence is simply taken as activation.

app/controllers/schools/participants_controller.rb
```diff
@@ -18,8 +19,6 @@ def show
     @mentor = @participant.early_career_teacher_profile&.mentor
   end
 
-  def edit_details; end
-
```
kicferk1, line 22: Is this unused?

kicferk1 (approved):

spec/services/feature_flag_spec.rb
```diff
@@ -3,40 +3,131 @@
...
+      it "does not deactivate the feature for other objects" do
+        expect { described_class.deactivate(feature_name, for: object) }
+          .not_to change { described_class.active?(feature_name, for: other_object) }
+      end
+    end
+  end
+
+  def random_record
+    # TODO: WHY ARE SOME FACTORIES BROKEN?
+    create FactoryBot.factories.map(&:name).without(:participant_band, :pupil_premium).sample
```
kicferk1, line 131: Ummm...
I want some context xD
What is broken exactly?

kicferk1: I guess some factories are never meant to be used by themselves, just as auxiliary things for other factories to use. Mostly out of general convenience.

Is it likely that we will want feature flags against anything other than users and schools? Maybe lead providers? Maybe this should be 'opt in' for factories and not 'opt out'?

app/controllers/schools/participants_controller.rb
BroiSatse: It was used to display "Not implemented" page

spec/services/feature_flag_spec.rb
BroiSatse: So `participant_band` and `pupil_premium` do not have enough attributes defined to be valid on creation. They do however look like a valid domain objects and we probably should fix those factories.

TBH, Ideally I would not be using any exisitng record in this test and use `with_model` or similar gem to mock AR class. The point is that this is to work with any AR object and I wanted to make that obvious in the test.

jamesrose (approved):
This is nice, thanks!

PR #527, opened 2021-06-15
Add support for named params in Cucumber

callumacrae:
Will be used to rename :id to :slug in #517 but will also be useful in the future if we ever want to specify more than one param


ethanmills (approved):
I need this now! 😛

callumacrae:
> I need this now! 😛

Let's put this one in first then! :)

kicferk1 (approved):

PR #528, opened 2021-06-15
Improved healthcheck

ethanmills:
We don't use the existing healthcheck endpoint, and the current gem causes a nasty deprecation warning when you run tests.

I'm not sure we want all of these features, think of it as a set of options to choose from. I'd also like to hear suggestions on what might be useful.


spec/requests/health_check_spec.rb
```diff
@@ -1,27 +0,0 @@
-# frozen_string_literal: true
-
-require "rails_helper"
```
ethanmills, line 3: I can add any specs we need once we've decided on the features

Gemfile
```diff
@@ -21,9 +21,6 @@ gem "sentry-ruby", ">= 4.5.1"
 # Pagination
 gem "kaminari", ">= 1.2.0"
 
-# Adds health check functionality
-gem "health_check", github: "ianheggie/health_check", ref: "0b799ead604f900ed50685e9b2d469cd2befba5b"
-
```
kicferk1, line 26: yesss

app/controllers/application_controller.rb
```diff
@@ -9,7 +9,7 @@ class ApplicationController < ActionController::Base
   before_action :set_sentry_user, except: :check, unless: :devise_controller?
 
   def check
-    render json: { status: "OK", version: release_version, sha: ENV["SHA"], environment: Rails.env }, status: :ok
+    head :ok
```
kicferk1, line 12: Is this what paas uses to decide whether to let our app live for another few moments?

config/initializers/health_check.rb
```diff
@@ -1,70 +0,0 @@
-# frozen_string_literal: true
-
-HealthCheck.setup do |config|
```
kicferk1, line 3: I am so glad we are axing this

app/controllers/healthcheck_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+    status_response = HTTPClient.get(NOTIFY_STATUS_API)
+    return "Status request failed" unless status_response.status == 200
+
+    JSON.parse(status_response.body)["status"]["indicator"]
+  rescue StandardError
+    "FAIL"
+  end
+
+  def puma_stats
+    JSON.parse(Puma.stats)
```
kicferk1, line 76: What is in there exactly?

kicferk1 (approved):
Looks good to me. How do we plan to use this exactly? Have a dev ping the endpoint every now and then to see how things are going? Or do we have some automated tool for that? Statuscake?

app/controllers/application_controller.rb
ethanmills: Yep, hence the simple, speedy response. "Please don't kill me!"

app/controllers/healthcheck_controller.rb
ethanmills: https://ecf-review-pr-528.london.cloudapps.digital/healthcheck

```json
{
version: "-",
sha: "45c7f8310f93bc96889477ebe7f1c9a3416839bf",
environment: "deployed_development",
database: {
connected: true,
migration_version: 20210611141524
},
delayed_job: {
job_count: 2,
errors: 0,
failed: 0,
last_failure: null
},
notify: {
incident_status: "none"
},
puma: {
started_at: "2021-06-15T18:07:38Z",
backlog: 0,
running: 5,
pool_capacity: 4,
max_threads: 5,
requests_count: 1706
}
}
```

app/controllers/healthcheck_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      delayed_job: {
+        job_count: delayed_jobs,
+        errors: jobs_with_errors,
+        failed: failed_jobs,
+        last_failure: last_job_failure,
+      },
+      notify: {
+        incident_status: notify_incident,
+      },
+      puma: puma_stats,
```
jamesrose, line 24: We'll likely want the DQT API status here at some point.

jamesrose (approved):

PR #532, opened 2021-06-16
Part two of "Nicer URLs for cohorts + schools"

callumacrae:
### Context

Part one: https://github.com/DFE-Digital/early-careers-framework/pull/517

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -154,10 +154,11 @@ class School < ApplicationRecord
     ]
   end
 
-  # This is temporary so that nothing breaks until the slugs are generated
-  def to_param
-    id
-  end
+# This temporary code is temporarily commented out during development
+# # This is temporary so that nothing breaks until the slugs are generated
+# def to_param
+#   id
+# end
 
 private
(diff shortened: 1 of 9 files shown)


ethanmills (approved):

app/models/school.rb
```diff
@@ -154,11 +154,6 @@ def slug_candidates
     ]
   end
 
-  # This is temporary so that nothing breaks until the slugs are generated
-  def to_param
-    id
-  end
-
```
kicferk1, line 161: So we can't merge this PR until we do a prod release and run the script for slugifying urls for all schools, right?

kicferk1 (approved):
Looks good to me, probably should not go in until you are sure it will work :)

app/models/school.rb
callumacrae: it'll be merged once i've done the _staging_ release and script, develop will become undeployable for a bit, then i'll deploy the first release to production, run the script, then deploy the second release (and develop will become deployable again)

PR #534, opened 2021-06-17
Fix calculation of band's subtotal in output payment

plisovin:
### Context

The bug in the calculation when total number of participants was used for each band's subtotal

```rake payment_calculation:breakdown Capita 4500```

Should return correct result as per jira [ticket](https://dfedigital.atlassian.net/browse/CPDTP-110)

### DIFF
--- a/lib/payment_calculator/ecf/contract/output_payment_calculations.rb
+++ b/lib/payment_calculator/ecf/contract/output_payment_calculations.rb
@@ -19,7 +19,7 @@ module PaymentCalculator
         end
 
         def output_payment_for_event(event_type:, total_participants:, band:)
-          total_participants * output_payment_per_participant_for_event(event_type: event_type, band: band)
+          [band.number_of_participants_in_this_band(total_participants), total_participants].min * output_payment_per_participant_for_event(event_type: event_type, band: band)
         end
 
         def output_payment_per_participant_for_event(event_type:, band:)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lib/payment_calculator/ecf/contract/output_payment_calculations.rb [hunk 1]
```diff
@@ -19,7 +19,7 @@ def output_payment_per_participant(band)
         end
 
         def output_payment_for_event(event_type:, total_participants:, band:)
-          total_participants * output_payment_per_participant_for_event(event_type: event_type, band: band)
+          [band.number_of_participants_in_this_band(total_participants), total_participants].min * output_payment_per_participant_for_event(event_type: event_type, band: band)
```
kicferk1, line 22: Can we add a unit test in `spec/lib/payment_calculator/ecf/contract/output_payment_calculations_spec.rb` that will prevent anyone reintroducing this bug?

plisovin: Ok, I will add it, although it's covered by the retention table feature test

jamesrose: @plisovin Did that test fail?

plisovin: yeah, and I need to fix the way this feature test is written as it's using Double rather than expected method

plisovin: Added in c501e66

kicferk1 (approved):

PR #536, opened 2021-06-17
add api endpoint to create npq profiles

asmega:
### Context

- NPQ registration collects data on applications of NPQ courses
- We need to push these applications into ECF with profiles (`NpqProfile`)
- A user can have multiple applications therefore are able to have multiple `NpqProfile`s

-  add `jsonapi-resources` gem which does all the heavy lifting of ingesting and producing jsonapi compliant requests and responses
- I have been able to call this endpoint successfully with `json_api_client`

### Guidance to review

- Of all the libraries looked at, this ones seems the most mature and ticked most of the boxes I feel for our needs
- I can revisit other api endpoints I have recently created and swap out to use this framework instead at a later PR
- I did attempt load the seeds file at before the test suite ran but this cause many failures. Is this something we want to consider since the app seems to depends on a load of static data why don't we load these seeds before the test suite kicks in?

### Testing

- note down the token
- create a user with

```
curl -H "Content-Type: application/vnd.api+json" -H "Authorization: Bearer TOKEN" -X POST -d '{ "data": { "type": "user", "attributes": { "full_name": "John Doe", "email": "john.doe@example.com" } } }' http://localhost:3001/api/v1/users
```

- note down the id of the newly created user

```
curl -H "Content-Type: application/vnd.api+json" -H "Authorization: Bearer TOKEN" -X POST -d '{"data":{"type":"npq_profiles","attributes":{"teacher_reference_number":"1234567","teacher_reference_number_verified":true,"date_of_birth":"1990-12-13","school_urn":"123456","headteacher_status":"no"},"relationships":{"user":{"data":{"type":"users","id":"USER_ID_HERE"}},"npq_lead_provider":{"data":{"type":"npq_lead_providers","id":"NPQ_LEAD_PROVIDER_ID"}},"npq_course":{"data":{"type":"npq_courses","id":"NPQ_COURSE_ID"}}}}}' http://localhost:3001/api/v1/npq-profiles
```

### Review Checks

- cant test in review app as i modified a migration and forced push and now db is out of sync. i have no idea how to teardown and setup review app again to fix this. just wasted 30 minutes figuring this out

### DIFF
--- a/app/controllers/api/v1/npq_profiles_controller.rb
+++ b/app/controllers/api/v1/npq_profiles_controller.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class NpqProfilesController < Api::ApiController
+      include ApiTokenAuthenticatable
+      include JSONAPI::ActsAsResourceController
+
+    private
+
+      def access_scope
+        ApiToken.where(private_api_access: true)
+      end
+    end
+  end
+end
--- a/app/models/npq_profile.rb
+++ b/app/models/npq_profile.rb
@@ -4,4 +4,10 @@ class NpqProfile < ApplicationRecord
   belongs_to :user
   belongs_to :npq_lead_provider
   belongs_to :npq_course
+
+  enum headteacher_status: {
+    no: "no",
+    yes_in_first_two_years: "yes_in_first_two_years",
+    yes_over_two_years: "yes_over_two_years",
+  }
 end
--- a/app/models/npq_profile.rb
+++ b/app/models/npq_profile.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class NpqProfile < ApplicationRecord
+  belongs_to :user
+  belongs_to :npq_lead_provider
+  belongs_to :npq_course
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,6 +11,7 @@ class User < ApplicationRecord
   has_one :admin_profile, dependent: :destroy
   has_one :early_career_teacher_profile, dependent: :destroy
   has_one :mentor_profile, dependent: :destroy
+  has_many :npq_profiles, dependent: :destroy
 
   validates :full_name, presence: true
   validates :email, presence: true, uniqueness: true, notify_email: true
--- a/app/resources/api/v1/npq_profile_resource.rb
+++ b/app/resources/api/v1/npq_profile_resource.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class NpqProfileResource < JSONAPI::Resource
+      attributes :date_of_birth,
+                 :teacher_reference_number,
+                 :teacher_reference_number_verified,
+                 :school_urn,
+                 :headteacher_status
+
+      has_one :user
+      has_one :npq_lead_provider
+      has_one :npq_course
+    end
+  end
+end
--- a/db/migrate/20210616103631_create_npq_profiles.rb
+++ b/db/migrate/20210616103631_create_npq_profiles.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class CreateNpqProfiles < ActiveRecord::Migration[6.1]
+  def change
+    create_table :npq_profiles do |t|
+      t.references :user
+      t.references :npq_lead_provider
+      t.references :npq_course
+
+      t.date :date_of_birth
+      t.text :teacher_reference_number
+      t.boolean :teacher_reference_number_verified, default: false
+      t.text :school_urn
+      t.text :headteacher_status # no, yes_in_first_two_years, yes_over_two_years
+
+      t.timestamps
+    end
+  end
+end
--- a/spec/requests/api/v1/npq_profiles_spec.rb
+++ b/spec/requests/api/v1/npq_profiles_spec.rb
@@ -1,0 +1,121 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "NPQ profiles api endpoint", type: :request do
+  let(:token) { NpqRegistrationApiToken.create_with_random_token! }
+  let(:bearer_token) { "Bearer #{token}" }
+  let(:parsed_response) { JSON.parse(response.body) }
+
+  describe "#create" do
+    let(:user) { create(:user) }
+    let(:npq_lead_provider) { create(:npq_lead_provider) }
+    let(:npq_course) { create(:npq_course) }
+
+    context "when authorized" do
+      before do
+        default_headers[:Authorization] = bearer_token
+        default_headers["Content-Type"] = "application/vnd.api+json"
+      end
+
+      let(:json) do
+        {
+          data: {
+            type: "npq_profiles",
+            attributes: {
+              teacher_reference_number: "1234567",
+              teacher_reference_number_verified: true,
+              date_of_birth: "1990-12-13",
+              school_urn: "123456",
+              headteacher_status: "no",
+            },
+            relationships: {
+              user: {
+                data: {
+                  type: "users",
+                  id: user.id,
+                },
+              },
+              npq_lead_provider: {
+                data: {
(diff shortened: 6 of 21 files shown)


spec/requests/api/v1/npq_profiles_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+              teacher_reference_number: "1234567",
+              teacher_reference_number_verified: true,
+              date_of_birth: "1990-12-13",
+              school_urn: "123456",
+              headteacher_status: "no",
+            },
+            relationships: {
+              user: {
+                data: {
+                  type: "users",
```
asmega, line 35: the only downside of `jsonapi-resources` is it only allows for plural `type` this is not configurable and cannot be changed.

app/resources/api/v1/npq_profile_resource.rb
```diff
@@ -0,0 +1,17 @@
...
+    class NpqProfileResource < JSONAPI::Resource
+      attributes :date_of_birth,
+                 :teacher_reference_number,
+                 :teacher_reference_number_verified,
+                 :school_urn,
+                 :headteacher_status
+
+      has_one :user
+      has_one :npq_lead_provider
+      has_one :npq_course
```
asmega, line 14: this is a `belongs_to` but `jsonapi-resources` wants a `has_one`

app/controllers/api/v1/npq_profiles_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class NpqProfilesController < Api::ApiController
+      include ApiTokenAuthenticatable
+      include JSONAPI::ActsAsResourceController
```
asmega, line 7: controller action all dealt with by `jsonapi-resources` out of the box. which is why there is `def create`

db/migrate/20210616103631_create_npq_profiles.rb
```diff
@@ -0,0 +1,19 @@
...
+class CreateNpqProfiles < ActiveRecord::Migration[6.1]
+  def change
+    create_table :npq_profiles do |t|
+      t.references :user
+      t.references :npq_lead_provider
+      t.references :npq_course
+
+      t.date :date_of_birth
+      t.text :teacher_reference_number
+      t.boolean :teacher_reference_number_verified, default: false
```
asmega, line 12: this will be the marker used to determine if application goes into the "holding pen"

db/migrate/20210616103631_create_npq_profiles.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.text :school_urn
```
asmega, line 13: I decided not to link up with actual `School` record. I don't feel it would give us any gain. Also I'm not sure if potentially ECF has a smaller dataset than NPQ??? We can link up via the `urn` at a later date if we need to

app/models/user.rb
```diff
@@ -11,6 +11,7 @@ class User < ApplicationRecord
   has_one :admin_profile, dependent: :destroy
   has_one :early_career_teacher_profile, dependent: :destroy
   has_one :mentor_profile, dependent: :destroy
+  has_many :npq_profiles, dependent: :destroy
```
asmega, line 14: differs from other profile types as this one is not singleton

app/controllers/api/v1/npq_profiles_controller.rb
kicferk1: I guess it adds a bunch of other actions, right? But hopefully they can't be accessed cause we control the routes?

app/models/npq_profile.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class NpqProfile < ApplicationRecord
+  belongs_to :user
+  belongs_to :npq_lead_provider
+  belongs_to :npq_course
```
kicferk1, line 6: Hm, I wonder if `npq_course` should have `npq_lead_provider` in it, and whether `belongs_to :npq_lead_provider` should be `through`. Thoughts?

app/resources/api/v1/npq_profile_resource.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class NpqProfileResource < JSONAPI::Resource
```
kicferk1, line 5: How do these resources work?

Does jsonapi know to use this one on `NpqProfilesController` by use of clever naming convention?

app/models/npq_profile.rb
asmega: My understanding is all providers offer all the courses. You don't select a provider first then get a provider specific course list.

app/controllers/api/v1/npq_profiles_controller.rb
asmega: at routing level you can disable any routes you don't want like a normal rails `resources` with `only` see routes file

app/models/user.rb
jamesrose: How come?

app/resources/api/v1/npq_profile_resource.rb
asmega: it seems to have default look up paths where it goes to look in a certain order. so this looks to be done through convention. I don't see an option to specify this by hand.

app/models/user.rb
asmega: a user can register for multiple NPQs so we need this as a has many

db/migrate/20210616103631_create_npq_profiles.rb
ethanmills: We have the full dataset, including closed schools. Happy for this to stay as-is though

db/migrate/20210616103631_create_npq_profiles.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.text :headteacher_status # no, yes_in_first_two_years, yes_over_two_years
```
ethanmills, line 14: Can this be defined as an enum on the model?

asmega: I've added the enum in, i've kept them as strings

app/resources/api/v1/npq_profile_resource.rb
kicferk1: Scary stuff. Alright, well, I guess it matches Rails so it is not too bad. Thanks for digging into it.

kicferk1:
> cant test in review app as i modified a migration and forced push and now db is out of sync. i have no idea how to teardown and setup review app again to fix this. just wasted 30 minutes figuring this out

Close the PR. Wait 10 minutes. Re-open the PR.

kicferk1:
> I did attempt load the seeds file at before the test suite ran but this cause many failures. Is this something we want to consider since the app seems to depends on a load of static data why don't we load these seeds before the test suite kicks in?

Which test suite? Rspec or cypress?

I think our unit tests should not load seed data as a general rule - it might make testing edge cases harder, and the setup more convoluted. But it would be useful to be able to load seed for a single test, or a single suite, to show how 'live' setup works.

I imagine most failures is factory bot (or any other test setup) trying to create things that are already created by seeds - say a cohort for year 2021.

asmega:
> Close the PR. Wait 10 minutes. Re-open the PR.

Thanks, noted for next time, don't want to close this one for now. 

> Which test suite? Rspec or cypress?

rspec

> I think our unit tests should not load seed data as a general rule - it might make testing edge cases harder, and the setup more convoluted. But it would be useful to be able to load seed for a single test, or a single suite, to show how 'live' setup works.

ok agreed. not really fussed and makes sense. you can run it in a `before` block but as it runs the script multiple times it raises warning as it redefines constants.

kicferk1 (approved):

ethanmills (approved):

jamesrose (approved):
Nice one!

timabell:
Having read this PR and the commits therein I don't understand the reason for the introduction of the `NpqLeadProvider` model/table. @asmega could you enlighten me?

PR #540, opened 2021-06-17
Cpdrp 388 update school information

BroiSatse:
https://dfedigital.atlassian.net/browse/CPDRP-388

This PR modifies SchoolDataImport to ignore certain changes in school information for existing schools. This is a temporary change until we can handle those problematic changes.

Additionally I have applied few performance tweaks, reducing the  execution time on the empty database from 1118s to 465s and each other run from 562s to 36s (however removing LA and LAD association handling had quite a bit impact here as well). These are:
1) Loading schools in batches of 1000
2) Caching LocalAuthority and LocalAuthorityDistrict records

Some further speed up could bo achieved by:
1) Creating a thread pool and handling schools concurrenty (requires thread safe cache object)
2) Performing batch inserts


app/services/school_data_importer.rb
```diff
@@ -46,77 +41,119 @@ def school_data_files
...
+      @new_school
+    end
+
+    def record_school_details
+      school.name = data.fetch("EstablishmentName")
+      school.address_line1 = data.fetch("Street")
+      school.address_line2 = data.fetch("Locality")
+      school.address_line3 = data.fetch("Town")
+      school.postcode = data.fetch("Postcode")
+      school.ukprn = data.fetch("UKPRN")
```
jamesrose, line 80: I'd exclude UKPRN as well if they're not new, just to be safe.

app/services/school_data_importer.rb
```diff
@@ -46,77 +41,119 @@ def school_data_files
...
+      school.school_phase_type = data.fetch("PhaseOfEducation (code)").to_i
+      school.school_phase_name = data.fetch("PhaseOfEducation (name)")
+      school.school_website = data.fetch("SchoolWebsite")
+      school.primary_contact_email = data.fetch("MainEmail").presence
+      school.secondary_contact_email = data.fetch("AlternativeEmail").presence
+
+      # Changes to properties below carries major consequences and must be avoided until we know how to handle them
```
jamesrose, line 87: 👍

app/services/school_data_importer.rb
```diff
@@ -46,77 +41,119 @@ def school_data_files
...
+      )
+    end
+
+    def link_school_to_local_district_authority
+      SchoolLocalAuthorityDistrict.create!(
+        school: school,
+        local_authority_district: local_district_authority,
+        start_year: start_year,
+      )
+    end
```
kicferk1, line 157: Do we always want to `create!`? I know we only do it for new schools, but maybe `find_or_create_by!` is safer?

BroiSatse: I am not certain about this. At the moment this executes only for new schools, so there should never be any `SchoolLocalAuthority` record for that school. I agree it is slightly safer in case of future changes to this service, but for now this would only add an extra db query. Additionally, if we assume that there might be a record already, then we need to handle it better in case the district changes (which is the main reason we only do this for new schools at the moment - to prevent LA changes to propagate to our db until we know how to handle it better)

app/services/school_data_importer.rb
BroiSatse: Done. :)

app/services/school_data_importer.rb
kicferk1: Fine. I imagine we already have a test which updates a school that has the same LA, right? So if we end up changing the service, and forget about this line, that test would break?

BroiSatse: Yup, it's here: https://github.com/DFE-Digital/early-careers-framework/pull/540/files#diff-5681fb24e488348c6c99cddc5acd1f5461796e35d49eea62178959f9d0eed6aeR69

kicferk1 (approved):

app/services/school_data_importer.rb
ethanmills: This doesn't look right to me. One LA has many schools

ethanmills (changes requested):

ethanmills:
Eugh, totally misunderstood the code. Ignore my request for changes, still reading

app/services/school_data_importer.rb
ethanmills: It would be good to get a unique index on this combination I think

app/services/school_data_importer.rb
```diff
@@ -46,77 +41,120 @@ def school_data_files
...
-      school_local_authority.start_year = start_year
-      SchoolLocalAuthority.latest.find_by(school: school)&.update!(end_year: start_year)
+        @local_authority = @la_cache[code] = local_authority
+      end
     end
 
-    school_local_authority.save!
-  end
+    def local_district_authority
+      return @local_district_authority if @local_district_authority
```
ethanmills, line 124: Under what circumstances would this be set?

app/services/school_data_importer.rb
```diff
@@ -46,77 +41,120 @@ def school_data_files
...
+      @la_cache.fetch(code) do
+        local_authority = LocalAuthority.find_or_initialize_by(code: code)
+        new_name = data.fetch("LA (name)")
 
-  def link_school_to_local_authority(school, local_authority)
-    school_local_authority = SchoolLocalAuthority.find_or_initialize_by(school: school, local_authority: local_authority)
+        if local_authority.name != new_name
+          if local_authority.persisted?
+            logger.info "LA name changed in school import. Old name: #{local_authority.name}, New name: #{new_name}"
+          end
```
ethanmills, line 114: The `else` here is that GIAS is inconsistent in it's naming, right? I think this is interesting enough to log

ethanmills (approved):
Couple of questions, nothing major

app/services/school_data_importer.rb
BroiSatse: If we call this twice. In the current code probably never, but it feels like something that should be memoized - without memoization this method (in my head) changes from a more complicated getter (noun) into a processing object (verb) which woudl probably require a name change (like `find_local_district_authority` or sth). It is purely opinionated though and I cannot really justify it by anything else than semantic.

app/services/school_data_importer.rb
BroiSatse: That's a bit where it is a bit shaky - the else is when we have no local authority in the database at all (in which case name is `nil`). This code is assuming that the CSV file we recceive is internally correct, so it does not contain two different names for the same LA codes) to such an extend that the name is only checked on the first encounter of LA code in the file - each next encounter of the same code will fetch LA record from the provided cache (`fetch` only executes the block when given key is not present).

ethanmills: Understood, thanks

jamesrose (approved):

PR #542, opened 2021-06-17
CPDRP-345: Induction coordinators have multiple schools

ethanmills:
### Context
CPDRP-345
These are the changes to allow an induction tutor to be appointed for - and manage - more than one school.
At present, we are not allowing this change if the provided name does not match the existing SIT name. We'll monitor this, and make changes as required. There is possibly a bigger piece to come around names.

- Allow an existing SIT to be nominated for a new school
- Allow an existing SIT to be added to another school by admin
- Add a school switcher page to allow managing multiple schools
- Update the breadcrumbs for correct nav if the user has multiple schools

### Guidance to review
Commitwise should work well here

### Testing
Pretty extensive unit and cypress tests. Let me know if there's anything I've missed!

### Review Checks


app/views/admin/schools/induction_coordinators/name_different.html.erb
```diff
@@ -0,0 +1,14 @@
+<% content_for :title, "Error: name different" %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-header-xl">
+      The name does not match
+    </h1>
+
+      <p class="govuk-body">An induction coordinator with that email address is already registered as <%= @existing_name %>.</p>
```
kicferk1, line 9: whitespace is off

app/views/admin/schools/induction_coordinators/name_different.html.erb
```diff
@@ -0,0 +1,14 @@
...
+
+    <p>Either change the name to match, or contact the Register and Partner team</p>
```
kicferk1, line 11: That copy feels wrong. Do we call ourselves `R&P` to the users? Should this be a support email link?

spec/requests/nominations/nominate_induction_coordinator_spec.rb
```diff
@@ -131,11 +132,43 @@
...
+
+      it "adds the schools to their list of schools" do
+        expect {
+          post "/nominations", params: { nominate_induction_tutor_form: {
+            full_name: existing_induction_coordinator.full_name,
+            email: email,
+            token: token,
+          } }
+        }
+          .not_to(change { User.count })
```
kicferk1, line 146: Some funny whitespace here and in a test below

app/controllers/schools/base_controller.rb
```diff
@@ -9,6 +9,8 @@ class Schools::BaseController < ApplicationController
   after_action :verify_authorized
   after_action :verify_policy_scoped
 
+  helper_method :breadcrumbs
```
kicferk1, line 12: Should we unit test this? If so, should we extract it to a helper to make our lives easier in tests?

kicferk1 (commented):
Looks pretty good, some whitespace and a question on testing breadcrumbs helper method

app/views/admin/schools/induction_coordinators/name_different.html.erb
ethanmills: We do to the frontline support staff. This is the admin interface remember

kicferk1: Ah, right. Ehh, frustrating that we have many names, but nothing that we can easily do about it.

app/controllers/schools/base_controller.rb
ethanmills: Not quite sure how we would extract this, as it calls `active_school` and `active_cohort`

kicferk1: Mhmm, and the difficulty of extraction means we don't want to test it too? :P 

If I were to extract it I would make a helper with breadcrumbs method that takes a school parameter and a cohort parameter, and builds breadcrumb content from that. And then in the controller all you do is pass the active school and/or cohort in. Or maybe in the view you pass the relevant school / cohort into the helper method. And then you can test the logic in helper easily.

If it sounds like too much work, I guess we can leave it untested, or maybe 'tested implicitly via percy'.

spec/cypress/app_commands/scenarios/schools/multiple_schools.rb
```diff
@@ -0,0 +1,7 @@
+cohort = FactoryBot.create(:cohort, :current)
+first_school = FactoryBot.create(:school, name: "Test School 1", slug: "111111-test-school-1")
+FactoryBot.create(:school_cohort, school: first_school, cohort: cohort)
+second_school = FactoryBot.create(:school, name: "Test School 2", slug: "111112-test-school-2")
+FactoryBot.create(:school_cohort, school: second_school, cohort: cohort)
```
callumacrae, line 5: Maybe it would be worth adding a third school that isn't added to the user to check that it _isn't_ displayed?

spec/cypress/integration/schools/MultipleSchools.feature
```diff
@@ -0,0 +1,27 @@
...
+  Background:
+    Given scenario "schools/multiple_schools" has been run
+    And I am logged in as existing user with email "school-leader@example.com"
+
+  Scenario: Selecting and changing schools
+    Given I am on "schools" page
+    Then "page body" should contain "Test School 1"
+    And "page body" should contain "Test School 2"
+    And the page should be accessible
+    And percy should be sent snapshot called "select school page"
```
callumacrae, line 12: snapshot name seems a bit vague, maybe worth adding "induction coordinator" to it?

ethanmills: Ah I thought it got prefixed with the scenario name. Apparently not

app/controllers/schools/base_controller.rb
ethanmills: No, it just means that I don't know how to extract it. Thanks for the suggestion, I'll take a look

kicferk1 (approved):

PR #546, opened 2021-06-18
Remove n+1 queries from users api

kicferk1:
### Context
I played with the users api for other reasons, and noticed n+1 queries

Cache school cohorts in the user serialiser. Use existing scope to get user schools alongside their profiles.

### Guidance to review

If there is a way to accomplish adding a `school_cohort` to the relevant profiles in a railsy association, I would love to know about it. We have a school, we have a cohort, but I don't know how to tell Rails 'and by the way, that defines a school_cohort'.

And that would be imo a neater solution - cause then we could just include `school_cohort` in our query for users.

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -47,7 +47,7 @@ module Api
       end
 
       def users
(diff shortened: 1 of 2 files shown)


ethanmills (approved):
Happy with this, would also love to know if there is a better way

PR #548, opened 2021-06-18
hardcode NPQ ids for resources

asmega:
### Context

- database is constantly teared down so identifiers keep changing in certain environments
- which means external references to these resources will not be in sync

- hardcode ids on NPQ resources so integration environments do not break 
- Might cause dupes in staging and prod but I can resolve those by hand

### Guidance to review

- Seeding process should continue to work

### Testing

- Should be able to run seeds multiple times

### Review Checks

- Log onto box
- Seeded data should match that of seeds file

### DIFF
--- a/db/seeds/initial_seed.rb
+++ b/db/seeds/initial_seed.rb
@@ -36,26 +36,26 @@ PrivacyPolicy.find_or_initialize_by(major_version: 1, minor_version: 0)
   .save!
 
 [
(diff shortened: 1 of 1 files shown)


kicferk1 (approved):
LGTM, if you end up having to do something like that a lot then it might be worth considering a gem we use on E&L: https://github.com/zdennis/activerecord-import

PR #549, opened 2021-06-18
CPDRP-498 Fix issue with Change link for mentor

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-498)
Fix wrong link action and remove link when no mentor available to select.  Also add workaround for `govuk_link_to` [issue](https://github.com/DFE-Digital/govuk-components/issues/193)

Make Change mentor link point to the mentor selection step.
Remove Change mentor link when no mentors available.
Workaround for `govuk_link_to` issue by adding `button: false` to method params so it can correctly handle `url_options`.

### Guidance to review
As an induction tutor, for a school with no mentors, add an ECT.
On the confirmation step, there should be no "Change" link for the Mentor.
Repeat once a mentor has been added and the Change link should be visible.
Clicking the Change link takes you to the mentor selection step.

### Review Checks

### DIFF
--- a/app/views/schools/add_participants/confirm.html.erb
+++ b/app/views/schools/add_participants/confirm.html.erb
@@ -57,7 +57,7 @@
             </div>
           </dd>
           <dd class="govuk-summary-list__actions">
-            <% unless add_participant_form.type == :self || add_participant_form.mentor_options.blank? %>
+            <% if add_participant_form.type != :self && add_participant_form.mentor_options.any? %>
               <%# govuk_link_to issue: https://github.com/DFE-Digital/govuk-components/issues/193 %>
               <%= govuk_link_to({ action: :show, step: "choose-mentor" }, button: false) do %>
                 Change <span class="govuk-visually-hidden">mentor</span>
--- a/app/views/schools/add_participants/confirm.html.erb
+++ b/app/views/schools/add_participants/confirm.html.erb
@@ -17,6 +17,7 @@
         </dd>
         <dd class="govuk-summary-list__actions">
           <% unless add_participant_form.type == :self %>
+            <%# govuk_link_to issue: https://github.com/DFE-Digital/govuk-components/issues/193 %>
             <%= govuk_link_to({ action: :show, step: "details" }, button: false) do %>
               Change <span class="govuk-visually-hidden">personal details</span>
             <% end %>
@@ -33,6 +34,7 @@
         </dd>
         <dd class="govuk-summary-list__actions">
           <% unless add_participant_form.type == :self %>
+            <%# govuk_link_to issue: https://github.com/DFE-Digital/govuk-components/issues/193 %>
             <%= govuk_link_to({ action: :show, step: "type" }, button: false) do %>
               Change <span class="govuk-visually-hidden">participant type</span>
             <% end %>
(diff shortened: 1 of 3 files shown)


app/views/schools/add_participants/confirm.html.erb
```diff
@@ -16,7 +16,7 @@
           <div><%= add_participant_form.email %></div>
         </dd>
         <dd class="govuk-summary-list__actions">
-          <%= govuk_link_to({ action: :show, step: "details" }) do %>
+          <%= govuk_link_to({ action: :show, step: "details" }, button: false) do %>
```
kicferk1, line 19: Can we add a comment linking to the issue please?

app/views/schools/add_participants/confirm.html.erb
```diff
@@ -30,7 +30,7 @@
           <%= add_participant_form.type_options[add_participant_form.type.to_sym] %>
         </dd>
         <dd class="govuk-summary-list__actions">
-          <%= govuk_link_to({ action: :show, step: "type" }) do %>
+          <%= govuk_link_to({ action: :show, step: "type" }, button: false) do %>
```
kicferk1, line 33: Ditto here

spec/cypress/app_commands/scenarios/no_mentors_for_participants.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+school = School.find_by(name: "Hogwarts Academy")
+if school.present?
+  school.mentor_profiles.each do |profile|
+    EarlyCareerTeacherProfile.where(mentor_profile: profile).update_all(mentor_profile_id: nil)
+    profile.user.destroy!
+  end
+  school.reload
+end
```
kicferk1, line 10: What happens if the school is not present?

Surely we can make sure it exists, this is our test setup after all.

kicferk1: Oh, I see, we are relying on another file being run before this one. A bit scary, but not too bad I guess.

I would still hope that the school was always present - otherwise the test cannot possibly work, right?

tonyheadford: Yes, just wanted a test to prove this was doing the right thing

tonyheadford: Comment added to file

app/views/schools/add_participants/confirm.html.erb
```diff
@@ -55,9 +57,10 @@
             </div>
           </dd>
           <dd class="govuk-summary-list__actions">
-            <% unless add_participant_form.type == :self %>
-              <%= govuk_link_to({ action: :show, step: "details" }) do %>
-                Change <span class="govuk-visually-hidden">personal details</span>
+            <% unless add_participant_form.type == :self || add_participant_form.mentor_options.blank? %>
```
BroiSatse, line 60: Could we change it to `if` - negated negation are a bit tricky to read. Also, could we replace `blank?` with `any?` or `none?` if negated (this will trigger `EXISTS` instead of `SELECT` and is slightly better semantically)

spec/cypress/integration/schools/AddParticipants.feature
```diff
@@ -118,3 +119,22 @@ Feature: School leaders should be able to manage participants
...
+    Then I should be on "2021 school participant details" page
+
+    When I type "Arthur Hall" into field labelled "Full name"
+    And I type "arthur.hall@school.gov.uk" into field labelled "Email"
+    And I click the submit button
+
+    Then I should be on "2021 school participant confirm" page
+    And "page body" should not contain "Change mentor"
+    And the page should be accessible
+    And percy should be sent snapshot called "school ect participant without mentor confirmation page"
```
BroiSatse, line 140: Note: This is example of scenario that should be tested in the test view. Ignore for now as we have no view tests yet...

BroiSatse (approved):

kicferk1 (approved):

PR #550, opened 2021-06-18
Replacing user description with participant type on participant list page

BroiSatse:
https://dfedigital.atlassian.net/browse/CPDRP-496 - issue mentioned in the comments section.

I might have said this already `user.user_description` should be burnt with fire :P


kicferk1 (approved):

kicferk1:
No visual changes - is that because the text was the same in both cases?

BroiSatse:
@kicferk1 Most likely, this was pretty edge case really, probably worth to be tested in the view spec.

PR #551, opened 2021-06-21
CPDTP-109 Display ECF uplift payments

blackrat:
### Context

The calculator requires uplift payments to be added into the display and calculation.

**Calculator**:
- Take the number of approved uplift participants and add in the uplift based on this number.
- Update the calculation orchestrator to call and retrieve this number from the aggregator (Note this is a temporary value based on the expected uplift number since there is a separate ticket to perform the actual aggregation which needs to be added to this.)
- Update the calculator to be the PaymentCalculator::Ecf::PaymentCalculation directly and switch this to take a contract rather than a lead provider, since the contract determines the calculation and the only thing that was done with the lead provider was to look this up.
- Update the PaymentCalculator to take an uplift number for calculation of the new section.
- Create the UpliftCalculation to return the correct amount for the `started` event as a per_participant and subtotal based on the number of uplift participants passed in.
- Update the rake task to call with and display the uplift participant values.
- Update integration tests to show the uplift and fix the number of participants displayed for each band.

**Bug fix**:
Change to the OutputPaymentRetentionEvent to return only the number of participants in this band rather than the total number in each band.

### Guidance to review

Mainly calculation code. 
Ignore the aggregation, since that's going to be added by the next PR.

### Review Checks

You can only view the results by running the rake task:

```rake payment_calculation:breakdown <lead_provider> <total_participants> <uplift_participants>```

### DIFF
--- a/app/services/calculation_orchestrator.rb
+++ b/app/services/calculation_orchestrator.rb
@@ -1,25 +1,20 @@
 # frozen_string_literal: true
 
 require "participant_event_aggregator"
-require "contract_event_payment_calculator"
+require "payment_calculator/ecf/payment_calculation"
 
 class CalculationOrchestrator
   class << self
-    def call(lead_provider:, aggregator: ::ParticipantEventAggregator, calculator: ::ContractEventPaymentCalculator, event_type: :started)
-      new(lead_provider: lead_provider, aggregator: aggregator, calculator: calculator).call(event_type: event_type)
+    def call(lead_provider:,
+             contract:,
+             aggregator: ::ParticipantEventAggregator,
+             # TODO: Add in aggregator for Uplift as uplift_aggregator: ::ParticipantUpliftAggregator (for example)
+             calculator: ::PaymentCalculator::Ecf::PaymentCalculation,
+             event_type: :started)
+      total_participants = aggregator.call({ lead_provider: lead_provider }, event_type: event_type)
+      uplift_participants = total_participants / 3 # Temporary code to fake the 33% target which will be removed when the aggregation is pulled in.
+      # TODO: uplift_participants = uplift_aggregator.call(leod_provider: event_type:) etc..
+      calculator.call(contract: contract, total_participants: total_participants, uplift_participants: uplift_participants, event_type: event_type)
     end
-  end
-
-  def call(event_type:)
-    total_participants = @aggregator.call({ lead_provider: @lead_provider }, event_type: event_type)
-    @calculator.call(lead_provider: @lead_provider, total_participants: total_participants, event_type: event_type)
-  end
-
-private
-
-  def initialize(lead_provider:, aggregator: ::ParticipantEventAggregator, calculator: ::ContractEventPaymentCalculator)
-    @lead_provider = lead_provider
-    @aggregator = aggregator
-    @calculator = calculator
   end
 end
--- a/app/services/calculation_orchestrator.rb
+++ b/app/services/calculation_orchestrator.rb
@@ -9,10 +9,12 @@ class CalculationOrchestrator
              contract:,
              aggregator: ::ParticipantEventAggregator,
              # TODO: Add in aggregator for Uplift as uplift_aggregator: ::ParticipantUpliftAggregator (for example)
+             # This is currently defined in Jira as https://dfedigital.atlassian.net/browse/CPDTP-195
              calculator: ::PaymentCalculator::Ecf::PaymentCalculation,
              event_type: :started)
       total_participants = aggregator.call({ lead_provider: lead_provider }, event_type: event_type)
       uplift_participants = total_participants / 3 # Temporary code to fake the 33% target which will be removed when the aggregation is pulled in.
+      # This is currently defined in Jira as https://dfedigital.atlassian.net/browse/CPDTP-195
       # TODO: uplift_participants = uplift_aggregator.call(leod_provider: event_type:) etc..
       calculator.call(contract: contract, total_participants: total_participants, uplift_participants: uplift_participants, event_type: event_type)
     end
--- a/lib/payment_calculator/ecf/contract/uplift_payment_calculations.rb
+++ b/lib/payment_calculator/ecf/contract/uplift_payment_calculations.rb
@@ -22,8 +22,8 @@ module PaymentCalculator
           event_type == :started ? uplift_payment_per_participant : 0
         end
 
-        def uplift_payment_for_event(total_participants:, event_type:)
-          total_participants * uplift_payment_per_participant_for_event(event_type: event_type)
+        def uplift_payment_for_event(uplift_participants:, event_type:)
+          uplift_participants * uplift_payment_per_participant_for_event(event_type: event_type)
         end
       end
     end
--- a/lib/payment_calculator/ecf/contract/uplift_payment_calculations.rb
+++ b/lib/payment_calculator/ecf/contract/uplift_payment_calculations.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require "has_di_parameters"
+
+module PaymentCalculator
+  module Ecf
+    module Contract
+      module UpliftPaymentCalculations
+        extend ActiveSupport::Concern
+
+        included do
+          include HasDIParameters
+        end
+
+        delegate :uplift_amount, to: :contract
+
+        def uplift_payment_per_participant
+          uplift_amount
+        end
+
+        def uplift_payment_per_participant_for_event(event_type:)
+          event_type == :started ? uplift_payment_per_participant : 0
+        end
+
+        def uplift_payment_for_event(total_participants:, event_type:)
+          total_participants * uplift_payment_per_participant_for_event(event_type: event_type)
+        end
+      end
+    end
+  end
+end
--- a/lib/payment_calculator/ecf/output_payment_retention_event.rb
+++ b/lib/payment_calculator/ecf/output_payment_retention_event.rb
@@ -9,7 +9,7 @@ module PaymentCalculator
 
       def call(total_participants:, event_type:, band:)
         {
-          retained_participants: total_participants,
+          retained_participants: band.number_of_participants_in_this_band(total_participants),
           per_participant: output_payment_per_participant_for_event(event_type: event_type, band: band).round(0),
           subtotal: output_payment_for_event(total_participants: total_participants, event_type: event_type, band: band).round(0),
         }
--- a/lib/payment_calculator/ecf/uplift_calculation.rb
+++ b/lib/payment_calculator/ecf/uplift_calculation.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+require "initialize_with_config"
+
+module PaymentCalculator
+  module Ecf
+    class UpliftCalculation
+      include Ecf::Contract::UpliftPaymentCalculations
+
+      # @param [Symbol] event_type
+      # @param [Integer] total_participants
+      # This is end number of participants who will be used to make the payment calculation.
+      # All invalid users will have already been filtered out before this number is generated and passed here.
+      def call(total_participants: 0, event_type: :started)
+        return nil unless event_type == :started
+
+        config[:contract] ||= lead_provider.call_off_contract
+        {
+          per_participant: uplift_payment_per_participant,
+          monthly: uplift_payment_for_event(event_type: event_type, total_participants: total_participants),
+        }
+      end
+    end
+  end
+end
--- a/lib/payment_calculator/ecf/uplift_calculation.rb
+++ b/lib/payment_calculator/ecf/uplift_calculation.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "initialize_with_config"
+require "payment_calculator/ecf/contract/uplift_payment_calculations"
 
 module PaymentCalculator
   module Ecf
@@ -8,16 +8,15 @@ module PaymentCalculator
       include Ecf::Contract::UpliftPaymentCalculations
 
       # @param [Symbol] event_type
-      # @param [Integer] total_participants
+      # @param [Integer] uplift_participants
       # This is end number of participants who will be used to make the payment calculation.
       # All invalid users will have already been filtered out before this number is generated and passed here.
-      def call(total_participants: 0, event_type: :started)
+      def call(uplift_participants: 0, event_type: :started)
         return nil unless event_type == :started
 
-        config[:contract] ||= lead_provider.call_off_contract
         {
           per_participant: uplift_payment_per_participant,
-          monthly: uplift_payment_for_event(event_type: event_type, total_participants: total_participants),
+          sub_total: uplift_payment_for_event(event_type: event_type, uplift_participants: uplift_participants),
         }
       end
     end
--- a/lib/tasks/payment_breakdown.rake
+++ b/lib/tasks/payment_breakdown.rake
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "payment_calculator/ecf/payment_calculation"
+require "payment_calculator/ecf/uplift_calculation"
 require "terminal-table"
 
 include ActiveSupport::NumberHelper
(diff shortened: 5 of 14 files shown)

### CODE REVIEW

File: app/services/calculation_orchestrator.rb [hunk 1]
```diff
@@ -1,25 +1,20 @@
...
-      new(lead_provider: lead_provider, aggregator: aggregator, calculator: calculator).call(event_type: event_type)
+    def call(lead_provider:,
+             contract:,
+             aggregator: ::ParticipantEventAggregator,
+             # TODO: Add in aggregator for Uplift as uplift_aggregator: ::ParticipantUpliftAggregator (for example)
+             calculator: ::PaymentCalculator::Ecf::PaymentCalculation,
+             event_type: :started)
+      total_participants = aggregator.call({ lead_provider: lead_provider }, event_type: event_type)
+      uplift_participants = total_participants / 3 # Temporary code to fake the 33% target which will be removed when the aggregation is pulled in.
+      # TODO: uplift_participants = uplift_aggregator.call(leod_provider: event_type:) etc..
```
kicferk1, line 16: Is there a ticket we could mention here?

lib/payment_calculator/ecf/contract/uplift_payment_calculations.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require "has_di_parameters"
+
+module PaymentCalculator
+  module Ecf
+    module Contract
+      module UpliftPaymentCalculations
+        extend ActiveSupport::Concern
```
kicferk1, line 9: Why do we want it to be a concern? At the moment it is used in a single place, so we could just inline it?

lib/payment_calculator/ecf/uplift_calculation.rb
```diff
@@ -0,0 +1,24 @@
...
+module PaymentCalculator
+  module Ecf
+    class UpliftCalculation
+      include Ecf::Contract::UpliftPaymentCalculations
+
+      # @param [Symbol] event_type
+      # @param [Integer] uplift_participants
+      # This is end number of participants who will be used to make the payment calculation.
+      # All invalid users will have already been filtered out before this number is generated and passed here.
+      def call(uplift_participants: 0, event_type: :started)
```
kicferk1, line 14: `Ecf::Contract::UpliftPaymentCalculations` has `HasDIParameters` included in it, and I vaguely remember that that module has its own idea about what a `call` of a class should do. How does it interact with this `call`?

And, well, should we maybe not mix the two approaches?

lib/tasks/payment_breakdown.rake
```diff
@@ -54,15 +63,20 @@ namespace :payment_calculation do
...
       Based on #{number_to_delimited(total_participants.to_i)} participants and #{per_participant_in_bands}
     RESULT
 
     logger.info table
     logger.info output
-  rescue StandardError
-    logger.info "Lead provider for '#{ARGV[1]}' not found"
+  rescue StandardError => e
+    puts e.message
+    puts e.backtrace
```
kicferk1, line 79: I know we print things in different ways across different files, but I definitely want consistency in a single file. This one already uses `logger`. Let's fo that please.

lib/tasks/payment_breakdown.rake
```diff
@@ -20,14 +21,18 @@ namespace :payment_calculation do
...
 
     breakdown = PaymentCalculator::Ecf::PaymentCalculation.call(
-      lead_provider: lead_provider,
+      contract: lead_provider.call_off_contract,
       total_participants: total_participants,
+      uplift_participants: uplift_participants,
       event_type: :started,
     )
 
+    puts "*** breakdown #{JSON.pretty_generate(breakdown)}"
```
timabell, line 34: This looks like local debug code, is it supposed to be here?

spec/lib/payment_calculator/ecf/uplift_calculation_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+        expect(result.dig(:sub_total)).to be_a(BigDecimal)
+      end
+
+      @combined_results ||= { uplift_payment: result }
+    end
+
+    @combined_results[:uplift_payment].each do |_, value|
+      expect(value).to be_a(BigDecimal)
+    end
+  end
```
kicferk1, line 38: Can we test the same things as we are testing here without using `combined_results`?
```
  it "returns the expected types for all outputs" do 
    retained_events.each do |key|
      result = described_class.call(contract: contract, event_type: key, total_participants: total_participants_eligible)
      expect(result.dig(:per_participant)).to be_a(BigDecimal)
      expect(result.dig(:sub_total)).to be_a(BigDecimal)
    end
  end
```
Are we missing anything here?

app/services/calculation_orchestrator.rb
```diff
@@ -1,25 +1,20 @@
...
+require "payment_calculator/ecf/payment_calculation"
 
 class CalculationOrchestrator
   class << self
-    def call(lead_provider:, aggregator: ::ParticipantEventAggregator, calculator: ::ContractEventPaymentCalculator, event_type: :started)
-      new(lead_provider: lead_provider, aggregator: aggregator, calculator: calculator).call(event_type: event_type)
+    def call(lead_provider:,
+             contract:,
+             aggregator: ::ParticipantEventAggregator,
+             # TODO: Add in aggregator for Uplift as uplift_aggregator: ::ParticipantUpliftAggregator (for example)
```
timabell, line 11: I suggest adding the jira ticket link to the TODO if this is referring to https://dfedigital.atlassian.net/browse/CPDTP-195
That way it's clear how this todo will be resolved.

app/services/calculation_orchestrator.rb
timabell, line 16: I suggest adding the jira ticket link to the TODO if this is referring to https://dfedigital.atlassian.net/browse/CPDTP-195
That way it's clear how this todo will be resolved.

lib/payment_calculator/ecf/contract/uplift_payment_calculations.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        included do
+          include HasDIParameters
```
timabell, line 12: Do we need to use this?

lib/payment_calculator/ecf/output_payment_retention_event.rb
```diff
@@ -9,7 +9,7 @@ class OutputPaymentRetentionEvent
 
       def call(total_participants:, event_type:, band:)
         {
-          retained_participants: total_participants,
+          retained_participants: band.number_of_participants_in_this_band(total_participants),
```
timabell, line 12: it doesn't really matter but I'd probably have done this bugfix as a separate PR as it's unrelated to uplift I think

app/services/calculation_orchestrator.rb
```diff
@@ -1,25 +1,20 @@
...
+             calculator: ::PaymentCalculator::Ecf::PaymentCalculation,
+             event_type: :started)
+      total_participants = aggregator.call({ lead_provider: lead_provider }, event_type: event_type)
```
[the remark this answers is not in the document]
blackrat: That's the optional config block used in the aggregator. I could change it, but that's a much bigger change than I wanted in this PR.

app/services/calculation_orchestrator.rb
blackrat: Yes. I'll add that in.

lib/payment_calculator/ecf/uplift_calculation.rb
blackrat: The class level call was designed to work in this way. The `uplift_participants` and `event_type` will be passed to this instance call since they are named parameters in the class level call.

lib/tasks/payment_breakdown.rake
timabell, line 79: If we're just spitting out the message and trace do we need to `rescue` at all?

lib/payment_calculator/ecf/contract/uplift_payment_calculations.rb
blackrat: It's a module as are most parts of the calculator. Keeping them as separated mix-ins keeps the similar functionality grouped together. Extending it as a concern gives us access to the `included do` rather than having to write `def self.included(base); base.class_eval do....` etc.

lib/tasks/payment_breakdown.rake
```diff
@@ -44,6 +49,10 @@ namespace :payment_calculation do
       ]
     end
 
+    uplift_payment = breakdown[:uplift].each_with_object({}) do |(k, v), h|
```
timabell, line 52: There's a fair amount of code in this rake task, if we change it do any tests break?

lib/tasks/payment_breakdown.rake
```diff
@@ -44,6 +49,10 @@ namespace :payment_calculation do
...
+      h[k] = "£#{number_to_delimited(v.to_i)}"
+    end
```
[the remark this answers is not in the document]
plisovin: Alternatively, could use one liner:

```
uplift_payment = breakdown[:uplift].tap{|hash| hash.transform_values! { |val| "£#{number_to_delimited(val.to_i)}" } }
```

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require "payment_calculator/ecf/contract/uplift_payment_calculations"
+
+class DummyClass
```
timabell, line 5: It took me a little while to work out that `DummyClass` only exists to make the `module` testable. I wonder if it's worth a short comment saying as much or whether this would be obvious to other devs more used to this way of testing ruby modules?

timabell: I think the name perhaps could be clearer in intent. `ModuleTestHarness` or something

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+end
+
+describe ::PaymentCalculator::Ecf::Contract::UpliftPaymentCalculations do
+  it "returns the expected types for uplift" do
+    contract = double("Contract Double", uplift_amount: BigDecimal(100.00, 2))
+    call_off_contract = DummyClass.new({ contract: contract })
+    event_type = :started
+
+    if @combined_results.nil?
+      expect(call_off_contract.uplift_payment_per_participant.round(2)).to be_a(BigDecimal)
```
timabell, line 16: why the round?

lib/tasks/payment_breakdown.rake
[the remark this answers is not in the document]
blackrat: Ok. 

It was originally using `inject`, but Rubocop changed it to `each_with_object`

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
timabell: I'd probably put the type checks for a single next to the value checks....

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  it "performs calculations of the uplift payments" do
+    contract = double("Contract Double", uplift_amount: 100)
+    call_off_contract = DummyClass.new({ contract: contract })
+    total_participants = 330
+
+    expect(call_off_contract.uplift_payment_per_participant.round(0)).to eq(100.00)
+
+    %i[started].each do |event_type|
+      expect(call_off_contract.uplift_payment_per_participant_for_event(event_type: event_type).round(2)).to eq(100.00)
```
timabell, line 30: I'd probably put the type checks for a single next to the value checks....
... like this:

```suggestion
      expect(call_off_contract.uplift_payment_per_participant_for_event(event_type: event_type).round(2)).to eq(100.00)
      expect(call_off_contract.uplift_payment_per_participant).to be_a(BigDecimal)
```

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      expect(call_off_contract.uplift_payment_for_event(event_type: event_type, uplift_participants: total_participants).round(2)).to eq(33_000.00)
```
timabell, line 31: We are doing turnip features for calculation outputs, I don't see a diff on them in this PR but you have implemented the calculations. I think we need them before this can be merged.

timabell (changes requested):

spec/lib/payment_calculator/turnip_steps/ecf/service_fees_steps.rb
```diff
@@ -13,11 +14,11 @@ class DummyClass
                       set_up_fee: @set_up_fee,
                       band_a: @band_a,
                       bands: [@band_a],
+                      uplift_amount: 100,
```
timabell, line 17: If it's not been specified in the steps wouldn't zero be a better default? I can see this causing confusion later.

lib/tasks/payment_breakdown.rake
blackrat: It was spitting out the wrong message before. Telling that the lead provider could not be found.

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
blackrat: The features were the wrong shape for the calculations required and are adding less and less value and requiring more and more maintenance. I'd suggest that we drop them since the same tests are being run in different ways and the more dynamic rake tasks are being used to generate output in the correct format.

lib/payment_calculator/ecf/contract/uplift_payment_calculations.rb
blackrat: Yes for now at least. It's keeps the 3 different blocks consistent with each other. May be removed in a PR later, but for now it's much harder to change just one.

lib/tasks/payment_breakdown.rake
timabell: I can see that was the case in the removed code, but do we need the replacement code, or can we just delete the rescue entirely?

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
timabell: I'm hesitant to drop the business readable specs, but equally I'm not keen on cargo-culting it if it's just work for no value. What do you think?

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
blackrat: It's fairly standard naming afaik. I've seen it called "DummyClass" and "DummyTestClass".

https://semaphoreci.com/community/tutorials/testing-mixins-in-isolation-with-minitest-and-rspec
https://mixandgo.com/learn/how-to-test-ruby-modules-with-rspec

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
blackrat: It takes longer to change the features that any other part of the coding. Adds overhead for no real value. It's also one reason that extracting the DIParameters has become so tricky since it's expecting a baseline to be created which it then calls with the totals. (Maybe that's a justification as to why we still need the new().call() structure, but I don't think it should be the only reason.)

spec/lib/payment_calculator/turnip_steps/ecf/service_fees_steps.rb
blackrat: That does seem to be the suitable default and if we were to expand the steps to use this dummy contract, I'd prefer a fast fail if for some reason it were checking uplift against declarations that didn't require uplift (e.g, Retained_1).

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
blackrat: That type is only applicable for the first one, not for every event type

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
blackrat: Done in the aggregation PR, so not required here too.

plisovin: It will be done in my branch

plisovin (approved):

kicferk1 (approved):

lib/tasks/payment_breakdown.rake
blackrat: No. Tests are using the same calls as the rake task, but not the rake task itself.

PR #554, opened 2021-06-21
Run percy config:migrate

ethanmills:
It told me to


kicferk1 (approved):
That one percy snapshot that surfaces is getting annoying, but this PR looks good to me.

PR #555, opened 2021-06-21
Add induction tutor email to school search

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/browse/CPDRP-501)
Enable admins to search search for schools by induction tutor email

Change the school search to additionally search against induction tutor email
Add hint text to search box to indicate what criteria should be supplied.

### Guidance to review
As an admin user, on the `/admin/schools` page the search function will also match against induction tutor email address and return schools that have a matching induction tutor (in addition to matching against school name and urn).

### Review Checks

### DIFF
--- a/app/controllers/admin/schools_controller.rb
+++ b/app/controllers/admin/schools_controller.rb
@@ -12,7 +12,7 @@ module Admin
 
       @schools = policy_scope(School)
         .includes(:induction_coordinators, :local_authority)
-        .yield_self(&method(:search_query))
+        .ransack(induction_coordinators_email_or_urn_or_name_cont: @query).result
         .order(:name)
         .page(params[:page])
         .per(20)
@@ -27,20 +27,6 @@ module Admin
 
     def load_school
       @school = School.eligible.friendly.find(params[:id])
-    end
-
-    def search_query(relation)
-      if @query.present?
-        search_value = "%#{@query}%"
-
-        relation
-          .left_joins(:induction_coordinators)
-          .where(School.arel_table[:name].matches(search_value)
-            .or(School.arel_table[:urn].matches(search_value))
-            .or(User.arel_table[:email].matches(search_value)))
-      else
-        relation
-      end
     end
   end
 end
--- a/app/controllers/admin/schools_controller.rb
+++ b/app/controllers/admin/schools_controller.rb
@@ -9,10 +9,11 @@ module Admin
 
     def index
       @query = params[:query]
+
       @schools = policy_scope(School)
         .includes(:induction_coordinators, :local_authority)
+        .yield_self(&method(:search_query))
         .order(:name)
-        .ransack(name_or_urn_cont: @query).result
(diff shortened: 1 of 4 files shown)


app/controllers/admin/schools_controller.rb
```diff
@@ -9,10 +9,11 @@ class SchoolsController < Admin::BaseController
 
     def index
       @query = params[:query]
+
       @schools = policy_scope(School)
         .includes(:induction_coordinators, :local_authority)
+        .yield_self(&method(:search_query))
```
ethanmills, line 15: Could this just be `.ransack(induction_coordinators_email_or_urn_or_name_cont: @query).result`?

app/controllers/admin/schools_controller.rb
```diff
@@ -27,5 +28,19 @@ def show
...
       @school = School.eligible.friendly.find(params[:id])
     end
+
+    def search_query(relation)
+      if @query.present?
+        search_value = "%#{@query}%"
+
+        relation
+          .left_joins(:induction_coordinators)
+          .where(School.arel_table[:name].matches(search_value)
```
kicferk1, line 38: What does `arel_table` do exactly?

BroiSatse: @kicferk1 ActiveRecord querying language is build on top of arel: https://github.com/rails/arel. `arel_table` returns instance of `Arel::Table`, `School.arel_table[:name]` returns `Arel::Attributes::Attributes` and `.matches` returns `Arel::Nodes::Matches` (which is a subclass of `Arel::Nodes::NodeExpression`) and `mathches(str).to_sql` returns nice `"schools"."name" ILIKE 'hello'`. Arel is much more generic than hash and allows constructing all sort of crazy sql than standard AR doesn't. 

However, I am not sure if it work here. I#d also prefer using ransack for searching.

app/controllers/admin/schools_controller.rb
BroiSatse: Yes, that should have the same result.

tonyheadford: OK, I will take a look, I'm not familiar with ransack and didn't twig it could cope with searching relations

app/controllers/admin/schools_controller.rb
tonyheadford: Yes, what @BroiSatse said ☝️ and yes it does work here.

app/controllers/admin/schools_controller.rb
tonyheadford: Yes, seems to work 👍

kicferk1 (approved):

app/controllers/admin/schools_controller.rb
kicferk1: cool, good to know :)

ethanmills (approved):

PR #556, opened 2021-06-21
CPDTP-195 Aggregate number of participants eligible for uplift payment

plisovin:
### Context

Aggregator to return number of participants eligible for uplift payment

- New scope for ParticipantDeclaration
- Add new columns `sparsity_uplift` and `pupil_premium_uplift` to profiles
- Update rspec and feature tests

### Testing

Run calculation orchestrator i.e.

```
lead_provider = LeadProvider.find_by(name: 'Capita')

CalculationOrchestrator.call(
  lead_provider: lead_provider,
  contract: lead_provider.call_off_contract,
  aggregator: ::ParticipantEventAggregator,
  uplift_aggregator: ::ParticipantUpliftAggregator,
  calculator: ::PaymentCalculator::Ecf::PaymentCalculation,
  event_type: :started
)

```

### Review Checks

### DIFF
--- a/app/models/participant_declaration.rb
+++ b/app/models/participant_declaration.rb
@@ -8,9 +8,14 @@ class ParticipantDeclaration < ApplicationRecord
   scope :active_for_lead_provider, ->(lead_provider) { active.for_lead_provider(lead_provider) }
   scope :count_active_for_lead_provider, ->(lead_provider) { active_for_lead_provider(lead_provider).count }
   scope :count_active_for_lead_provider_between, ->(lead_provider, start_date, end_date) { declared_as_between(start_date, end_date).count_active_for_lead_provider(lead_provider) }
+  scope :count_active_uplift_for_lead_provider, ->(lead_provider) { active_uplift_for_lead_provider(lead_provider).count }
 
   scope :unique_early_career_teacher_profile_id, -> { select(:early_career_teacher_profile_id).distinct }
   scope :active, -> { where(declaration_type: "started").order(declaration_date: "desc").unique_early_career_teacher_profile_id }
   scope :declared_as_between, ->(start_date, end_date) { where(declaration_date: start_date..end_date) }
   scope :submitted_between, ->(start_date, end_date) { where(created_at: start_date..end_date) }
+  scope :active_uplift_for_lead_provider, lambda { |lead_provider|
+    active_for_lead_provider(lead_provider).joins(:early_career_teacher_profile)
+      .where("early_career_teacher_profiles.sparsity_uplift OR early_career_teacher_profiles.pupil_premium_uplift")
+  }
 end
--- a/app/services/participant_uplift_aggregator.rb
+++ b/app/services/participant_uplift_aggregator.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+require "has_di_parameters"
+
+class ParticipantUpliftAggregator
+  include HasDIParameters
+
+  def call(event_type: :started)
+    recorder.send(params[event_type], lead_provider)
+  end
+
+private
+
+  def default_params
+    {
+      recorder: ParticipantDeclaration,
+      started: :count_active_uplift_for_lead_provider,
+    }
+  end
+end
--- a/db/migrate/20210610090328_add_uplift_to_early_career_teacher_profile.rb
+++ b/db/migrate/20210610090328_add_uplift_to_early_career_teacher_profile.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddUpliftToEarlyCareerTeacherProfile < ActiveRecord::Migration[6.1]
+  def change
+    change_table(:early_career_teacher_profiles, bulk: true) do |table|
+      table.column :sparsity_uplift, :boolean, null: false, default: false
+      table.column :pupil_premium_uplift, :boolean, null: false, default: false
+    end
+  end
+end
--- a/spec/factories/early_career_teacher_profiles.rb
+++ b/spec/factories/early_career_teacher_profiles.rb
@@ -4,5 +4,9 @@ FactoryBot.define do
   factory :early_career_teacher_profile do
     user
     school
+
+    trait :sparsity_uplift do
+      sparsity_uplift { true }
+    end
   end
 end
--- a/spec/factories/early_career_teacher_profiles.rb
+++ b/spec/factories/early_career_teacher_profiles.rb
@@ -8,5 +8,14 @@ FactoryBot.define do
     trait :sparsity_uplift do
       sparsity_uplift { true }
     end
+
+    trait :pupil_premium_uplift do
+      pupil_premium_uplift { true }
+    end
+
+    trait :uplift_flags do
+      sparsity_uplift
+      pupil_premium_uplift
+    end
   end
 end
--- a/spec/lib/payment_calculator/turnip_features/ecf/uplift_payment.feature
+++ b/spec/lib/payment_calculator/turnip_features/ecf/uplift_payment.feature
@@ -3,6 +3,6 @@ Feature: ECF contract calculations
 
   Scenario: Uplift payment
     Given An uplift per participant of £100
-    And there are 300 participants who have started that are eligible for the uplift payment
+    And there are 100 sparsity and 200 pupil premium participants who have started that are eligible for the uplift payment
     When I setup the contract with uplift payment
     Then the total uplift payment should be £30,000
--- a/spec/lib/payment_calculator/turnip_features/ecf/uplift_payment.feature
+++ b/spec/lib/payment_calculator/turnip_features/ecf/uplift_payment.feature
@@ -3,6 +3,7 @@ Feature: ECF contract calculations
 
   Scenario: Uplift payment
     Given An uplift per participant of £100
-    And there are 100 sparsity and 200 pupil premium participants who have started that are eligible for the uplift payment
+    And there are 100 sparsity participants who have started that are eligible for the uplift payment
+    And there are 200 pupil premium participants who have started that are eligible for the uplift payment
     When I setup the contract with uplift payment
     Then the total uplift payment should be £30,000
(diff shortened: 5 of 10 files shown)


db/migrate/20210610090328_add_uplift_to_early_career_teacher_profile.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddUpliftToEarlyCareerTeacherProfile < ActiveRecord::Migration[6.1]
+  def change
+    change_table(:early_career_teacher_profiles, bulk: true) do |table|
+      table.column :sparsity_uplift, :boolean, null: false, default: false
+      table.column :pupil_premium_uplift, :boolean, null: false, default: false
```
kicferk1, line 7: How are we planning to populate this stuff?

plisovin: It's in this commit 42ef1cc

kicferk1: Well, some of it is. What about the profiles that don't exist yet, but will exist in a few days or weeks, when the ECTs are added by induction coordinators?

I imagine you want to add something adding it to `save!` method of `AddParticipantForm`.

plisovin: yeah, Paul is working on this at the moment actually, it's exactly there or in the model callback

kicferk1: Also, perhaps a more important question. If we store this data against the school, are we sure we want to duplicate it on the ECT profiles? 
Are we doing it for efficiency reason? Are we doing it because when user switches school we want to remember it? Are we doing it because we are worried that over time this stuff will change?

timabell: > Also, perhaps a more important question. If we store this data against the school, are we sure we want to duplicate it on the ECT profiles?
> Are we doing it for efficiency reasons? Are we doing it because when user switches school we want to remember it? Are we doing it because we are worried that over time this stuff will change?

"Are we doing it because when user switches school we want to remember it?" - Yes, because they can in theory change schools between the point of registration and (months later) when we calculate the payment. We suspect these school changes may be done via support initially. The business rules we were given are that it is the uplift at the time they are registered that matters, not the uplift at the point we calculate.

timabell (changes requested):
We can't merge the backfill migration until the code to populate uplift for new registrations is in place otherwise we'll end up with registrations that never got the uplift data.

timabell:
Backfill migration commits removed from this PR so that we can ship this without having to wait for the code to populate the new columns.

kicferk1 (commented):
There seems to be a change on `.githooks/pre-commit` which I think changes whether it's treated as executable. Could we revert that?

Looks good other than that.

app/models/participant_declaration.rb
```diff
@@ -8,9 +8,15 @@ class ParticipantDeclaration < ApplicationRecord
...
   scope :count_active_for_lead_provider_between, ->(lead_provider, start_date, end_date) { declared_as_between(start_date, end_date).count_active_for_lead_provider(lead_provider) }
+  scope :count_active_uplift_for_lead_provider, ->(lead_provider) { active_uplift_for_lead_provider(lead_provider).count }
 
   scope :unique_early_career_teacher_profile_id, -> { select(:early_career_teacher_profile_id).distinct }
   scope :active, -> { where(declaration_type: "started").order(declaration_date: "desc").unique_early_career_teacher_profile_id }
   scope :declared_as_between, ->(start_date, end_date) { where(declaration_date: start_date..end_date) }
   scope :submitted_between, ->(start_date, end_date) { where(created_at: start_date..end_date) }
+  scope :active_uplift_for_lead_provider, lambda { |lead_provider|
+    active_for_lead_provider(lead_provider).joins(:early_career_teacher_profile)
+      .where(lead_provider: lead_provider)
```
timabell, line 19: Redundant line it seems

spec/factories/early_career_teacher_profiles.rb
```diff
@@ -4,5 +4,9 @@
   factory :early_career_teacher_profile do
     user
     school
+
+    trait :sparsity_uplift do
+      sparsity_uplift { true }
+    end
```
blackrat, line 10: Would be good to do this for `pupil_premium_uplift` too.

timabell (dismissed):
This commit dff517dd1150be9b9d3fbffc9de458c2bd3a2864 doesn't make a lot of sense at the moment. Aside from the pre-commit noise, the message doesn't match the patch

spec/services/calculation_orchestrator_spec.rb
```diff
@@ -50,15 +50,15 @@
...
         per_participant: 100.0,
+        sub_total: 1000.0,
       },
     }
   end
 
   before do
     10.times do
-      participant_declaration = create(:participant_declaration, lead_provider: call_off_contract.lead_provider)
+      participant_declaration = create(:participant_declaration, :sparsity_uplift, lead_provider: call_off_contract.lead_provider)
```
blackrat, line 61: Also would test with pupil_premium_uplift and a combination of the two.

blackrat (dismissed):

app/models/participant_declaration.rb
```diff
@@ -8,9 +8,15 @@ class ParticipantDeclaration < ApplicationRecord
   scope :active_for_lead_provider, ->(lead_provider) { active.for_lead_provider(lead_provider) }
   scope :count_active_for_lead_provider, ->(lead_provider) { active_for_lead_provider(lead_provider).count }
   scope :count_active_for_lead_provider_between, ->(lead_provider, start_date, end_date) { declared_as_between(start_date, end_date).count_active_for_lead_provider(lead_provider) }
+  scope :count_active_uplift_for_lead_provider, ->(lead_provider) { active_uplift_for_lead_provider(lead_provider).count }
```
timabell, line 11: Suggestion to call `.count` from calling code instead of having another scope here

app/services/participant_uplift_aggregator.rb
```diff
@@ -0,0 +1,20 @@
...
+  def call(event_type: :started)
+    recorder.send(params[event_type], lead_provider)
+  end
+
+private
+
+  def default_params
+    {
+      recorder: ParticipantDeclaration,
+      started: :count_active_uplift_for_lead_provider,
```
timabell, line 17: I don't think we need this configurability now and we should simplify this class down to something smaller and easier to follow

spec/lib/payment_calculator/turnip_features/ecf/uplift_payment.feature
```diff
@@ -0,0 +1,8 @@
+@ecf
+Feature: ECF contract calculations
+
+  Scenario: Uplift payment
+    Given An uplift per participant of £100
+    And there are 300 participants who have started that are eligible for the uplift payment
```
timabell, line 6: maybe we should split this line into 100 sparsity and 200 pupil premium

spec/lib/payment_calculator/turnip_steps/ecf/uplift_payment_steps.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module UpliftPaymentsSteps
+  class DummyClass
```
timabell, line 4: I think we should rename this to something more meaningful

```suggestion
  class ModuleTestHarness
```

spec/services/calculation_orchestrator_spec.rb
timabell: Might be okay as-is if we expand the unit test (the above turnip feature) to test both.

app/models/participant_declaration.rb
plisovin: Fixed

spec/factories/early_career_teacher_profiles.rb
plisovin: Added

spec/lib/payment_calculator/turnip_features/ecf/uplift_payment.feature
plisovin: Added in the dfdc589

spec/lib/payment_calculator/turnip_features/ecf/uplift_payment.feature
```diff
@@ -0,0 +1,8 @@
+@ecf
+Feature: ECF contract calculations
+
+  Scenario: Uplift payment
+    Given An uplift per participant of £100
+    And there are 100 sparsity and 200 pupil premium participants who have started that are eligible for the uplift payment
```
timabell, line 6: This would make more sense to me as two steps

```suggestion
    And there are 100 sparsity participants who have started that are eligible for the uplift payment
    And there are 200 pupil premium participants who have started that are eligible for the uplift payment
```

plisovin:
> This commit [dff517d](https://github.com/DFE-Digital/early-careers-framework/commit/dff517dd1150be9b9d3fbffc9de458c2bd3a2864) doesn't make a lot of sense at the moment. Aside from the pre-commit noise, the message doesn't match the patch

Was it deleted already?

spec/lib/payment_calculator/turnip_features/ecf/uplift_payment.feature
plisovin: Done in a5f3779

blackrat (approved):

spec/services/calculation_orchestrator_spec.rb
blackrat: I'd do it in the spec explicitly. Turnip scenarios are more suitable for end-to-end integration tests as and such are much more fragile than testing just the participant_declaration.

app/services/participant_uplift_aggregator.rb
blackrat: Now as in "not yet", or now as in there's something that can replace it that is equally as good or better?

PR #560, opened 2021-06-22
Displaying cohort opt-out confirmation before saving the choice

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-446

### DIFF
--- a/app/forms/induction_choice_form.rb
+++ b/app/forms/induction_choice_form.rb
@@ -4,28 +4,38 @@ class InductionChoiceForm
   include ActiveModel::Model
   include ActiveModel::Serialization
 
-  attr_accessor :programme_choice
+  attr_reader :programme_choice
+
+  PROGRAMME_OPTIONS = %i[full_induction_programme core_induction_programme design_our_own no_early_career_teachers].freeze
 
   def attributes
     { programme_choice: nil }
   end
 
-  validates :programme_choice, presence: { message: "Select how you want to run your induction" }
+  validates :programme_choice, presence: { message: "Select how you want to run your induction" }, inclusion: { in: PROGRAMME_OPTIONS }
 
   def programme_choices
-    [
-      OpenStruct.new(id: "full_induction_programme", name: "Use a training provider, funded by the DfE (full induction programme)"),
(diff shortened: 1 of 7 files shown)


app/forms/induction_choice_form.rb
```diff
@@ -21,6 +21,10 @@ def programme_choices
     ]
   end
 
+  def programme_choice_words
+    programme_choices.find { |opt| opt.id == programme_choice }&.name
```
kicferk1, line 25: Let's use full words, `option` > `opt`

app/views/schools/choose_programme/confirm_programme.html.erb
```diff
@@ -9,11 +9,7 @@
     <p class="govuk-body">You have chosen to:</p>
 
     <ul class="govuk-list govuk-list--bullet">
-      <% if @induction_choice_form.programme_choice == "full_induction_programme" %>
-        <li>Use a training provider, funded by the DfE</li>
-      <% else %>
-        <li>Deliver your own programme using DfE accredited materials</li>
-      <% end %>
+      <li><%= @induction_choice_form.programme_choice_words %></li>
```
kicferk1, line 12: Is this the same? I think the programme_choice have `(full induction programme)` and `(core induction programme)` in them in addition to those options.

kicferk1: And percy noticed that :O 
Also, looks like cypress tests are broken 😢

BroiSatse: Yup, there are small changes. I was hoping those are all right given that FIP and CIP are a technical jargon. Content files are not too helpfull there (I don't know what does it mean when the text in brackets is stroke orange?) and prototype doesn't show these neither.

Cypress fixed.

kicferk1 (approved):

kicferk1:
One of the new pages says 
> You have chosen to
> We don't expect new teachers...

Not sure if it is the right wording :P

kicferk1 (approved):

PR #561, opened 2021-06-22
Serialise CIP from cohorts as well as from profiles

kicferk1:
### Context
E&L wants users CIPs. For private beta users, I think we need to get them from school cohort model. Up till now we were getting it from the profile. This fixes it.

If user has no CIP on their profile, fetch it from their cohort in the serialiser.

### Guidance to review
I had a passing thought of recording user's CIP when we create them and backfilling them. Thoughts on that?

### Review Checks

You can't

### DIFF
--- a/app/serializers/user_serializer.rb
+++ b/app/serializers/user_serializer.rb
@@ -31,7 +31,9 @@ class UserSerializer
   end
 
   attributes :core_induction_programme do |user|
-    case user.core_induction_programme&.name
+    core_induction_programme = user.core_induction_programme || find_school_cohort(user)&.core_induction_programme
+
+    case core_induction_programme&.name
     when "Ambition Institute"
       CIP_TYPES[:ambition]
     when "Education Development Trust"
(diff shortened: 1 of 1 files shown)


ethanmills:
This is an interesting one, and I don't fully understand the requirements around moving schools etc. My gut feeling is that copying the CIP to the ECT profile would be a safe thing to do, similar to what T&P are working on for uplift. I don't think mentor's should have a CIP on their profile though, I think they should have a list of CIPs from any cohorts they're part of and any ECTs they have

kicferk1:
> This is an interesting one, and I don't fully understand the requirements around moving schools etc. My gut feeling is that copying the CIP to the ECT profile would be a safe thing to do, similar to what T&P are working on for uplift. I don't think mentor's should have a CIP on their profile though, I think they should have a list of CIPs from any cohorts they're part of and any ECTs they have

Ehh, I think mentor having more than one CIP was discussed quite a lot around E&L needs, and was moved into the category of 'we will do it eventually'. I'm happy to make another PR that saves the CIP choice on the ECT / mentor profile, but it will probably be quite a bit larger - backfilling the data, worrying about how schools updating their CIP works, worrying how users changing school works, etc.

Are you happy for this to go in in the meantime?

app/serializers/user_serializer.rb
```diff
@@ -46,10 +48,13 @@ class UserSerializer
   end
 
   attributes :induction_programme_choice do |user|
+    find_school_cohort(user)&.induction_programme_choice
+  end
+
+  def self.find_school_cohort(user)
     @school_cohorts ||= SchoolCohort.all.to_a
```
ethanmills, line 55: Leave for now, but I'd be interested to know if we actually get a performance boost from reading potentially ~20k records (and growing every year) into memory to save 100 queries

ethanmills (approved):

app/serializers/user_serializer.rb
kicferk1: Hmm, not sure about performance, but I am pretty sure it saves us loads of very similar logs

PR #562, opened 2021-06-22
CPDTP-195 Make cohort mandatory for Early Career Teachers

blackrat:
### Context

Cohort was not a mandatory attribute for Early Career Teachers (ECT), but really should be.

- Make cohort required in the database and set a default value for existing records that are currently null.
- Change factories to add cohorts to ECTs.
- Make a :current trait that uses the existing seeded value when this year is used in a factory.
  - See https://dev.to/jooeycheng/factorybot-findorcreateby-3h8k for details

### Guidance to review

Is defaulting the cohort value to 'current', the best and only option? 
Are the assumptions sound?

### Testing

All existing tests and features should still pass.

### Review Checks

It's fundamentally a refactor to make something concrete that has been previously assumed.

### DIFF
--- a/db/migrate/20210621160955_change_early_career_teacher_profile_to_make_cohorts_required.rb
+++ b/db/migrate/20210621160955_change_early_career_teacher_profile_to_make_cohorts_required.rb
@@ -2,6 +2,6 @@
 
 class ChangeEarlyCareerTeacherProfileToMakeCohortsRequired < ActiveRecord::Migration[6.1]
   def change
-    change_column_null :early_career_teacher_profiles, :cohort_id, false, Cohort.current.id
+    add_check_constraint :early_career_teacher_profiles, "cohort_id IS NOT NULL", name: "early_career_teacher_profiles_cohort_id_null", validate: false
   end
 end
--- a/db/migrate/20210621160955_change_early_career_teacher_profile_to_make_cohorts_required.rb
+++ b/db/migrate/20210621160955_change_early_career_teacher_profile_to_make_cohorts_required.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class ChangeEarlyCareerTeacherProfileToMakeCohortsRequired < ActiveRecord::Migration[6.1]
+  def change
+    change_column_null :early_career_teacher_profiles, :cohort_id, false, Cohort.current.id
+  end
+end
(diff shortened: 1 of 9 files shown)


kicferk1 (approved):
I might have another look at the test setup later this week, but for now I think the benefit of non-nullable cohorts on ECT profiles outweighs a bit of awkwardness in factory bot setup.

spec/factories/cohorts.rb
```diff
@@ -5,7 +5,12 @@
     start_year { Faker::Number.unique.between(from: 2021, to: 2100) }
 
     trait :current do
-      start_year { 2021 }
+      # In order for factory_bot to correctly use seeds in the database you have to jump through a couple of hoops,
+      # Detailed in: https://dev.to/jooeycheng/factorybot-findorcreateby-3h8k
```
timabell, line 9: Super :+1:

timabell (approved):

db/migrate/20210621160955_change_early_career_teacher_profile_to_make_cohorts_required.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class ChangeEarlyCareerTeacherProfileToMakeCohortsRequired < ActiveRecord::Migration[6.1]
+  def change
+    change_column_null :early_career_teacher_profiles, :cohort_id, false, Cohort.current.id
```
jamesrose, line 5: Setting `NOT NULL` on an existing column isn't safe. Try something like:

```suggestion
    add_check_constraint :early_career_teacher_profiles, "cohort_id IS NOT NULL", name: "early_career_teacher_profiles_cohort_id_null", validate: false
```

jamesrose (changes requested):

db/migrate/20210621160955_change_early_career_teacher_profile_to_make_cohorts_required.rb
ethanmills: We would need to define an up and down to make this reversible, right?

ethanmills: Ah, it's reversible. Nice!

ethanmills (approved):

db/migrate/20210621160955_change_early_career_teacher_profile_to_make_cohorts_required.rb
blackrat: Yes. Was just about to comment on that with:
https://blog.saeloun.com/2021/01/08/rails-6-check-constraints-database-migrations.html

jamesrose (approved):

blackrat:
Closing to let Terraform do its thing with the review app

PR #563, opened 2021-06-22
CPDRP-317 fix missing list classes

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/browse/CPDRP-317)
Fix styles on list on sign-in page

Add correct govuk styles to list on sign-in page 

### Guidance to review
Go to sign-in page, the list should be correctly formatted with the right styles

### Review Checks

image

### DIFF
--- a/app/views/users/sessions/new.html.erb
+++ b/app/views/users/sessions/new.html.erb
@@ -17,7 +17,7 @@
       <%= f.govuk_submit "Sign in" %>
     <% end %>
     <h2 class="govuk-heading-m">How you can access this service</h2>
-    <ul>
+    <ul class="govuk-list govuk-list--bullet">
       <li>you must be nominated or sent sign in details to access this service. <%= govuk_link_to "Find out how to get an account", check_account_path %></li>
       <li>we don’t use DfE sign in or passwords, you just need your account’s email address</li>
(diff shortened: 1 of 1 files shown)


app/views/start/index.html.erb
```diff
@@ -7,7 +7,7 @@
     <p>Your school must complete these steps before your early career teachers (ECTs) start their statutory induction programme.
       For example, if your ECTs are starting in September, you need to complete these steps before then.</p>
 
-    <%= govuk_link_to "Start", new_user_session_path, button: true %>
+    <%= govuk_start_now_button text: "Start now", href: new_user_session_path %>
```
kicferk1, line 10: Is that in the right PR?

kicferk1 (approved):
Don't get me wrong, I am happy to have two small visual fixes in one PR, but it looks like one of them has a PR for itself https://github.com/DFE-Digital/early-careers-framework/pull/564

app/views/start/index.html.erb
tonyheadford: Good spot! How did I manage that :(

tonyheadford: Removed other changes

PR #565, opened 2021-06-22
CPDRP-319: change width of privacy policy area

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/browse/CPDRP-319)
Change width of column for privacy policy text

Changes the style on the privacy policy page to use `govuk-grid-column-three-quarters` class instead of `govuk-grid-column-two-thirds`.

### Review Checks

image

### DIFF
--- a/app/views/privacy_policies/show.html.erb
+++ b/app/views/privacy_policies/show.html.erb
@@ -1,7 +1,7 @@
 <% content_for :title, "Privacy policy" %>
 
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-three-quarters">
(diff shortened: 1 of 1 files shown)


kicferk1 (commented):
So the ticket says that it's useful to have this width for sentences. Surely this is not the only page that has long sentences, and I think our default width is `two-thirds`.

So should we double check whether we want this change applied across a wider range of pages?

ethanmills:
> So the ticket says that it's useful to have this width for sentences. Surely this is not the only page that has long sentences, and I think our default width is `two-thirds`.
> 
> So should we double check whether we want this change applied across a wider range of pages?

Just to add some context to this, we use two thirds because that's what's recommended https://design-system.service.gov.uk/styles/layout/

ethanmills (approved):

PR #566, opened 2021-06-22
CPDRP-199: Add a task to clear S3 buckets

ethanmills:
Clear uploaded CSVs from S3 on each deploy & db wipe to dev and review.
Also clear them before tearing down a review app. We cannot delete the
bucket if it is non-empty.

This might not catch 100% of cases, but that's still a lot better than at the moment.
I looked for something we could do in terraform or cloudfoundry to delete the bucket without it being empty, but instead found lots of threads of people choosing not to implement that...

### DIFF
--- a/terraform/workspace-variables/dev.tfvars
+++ b/terraform/workspace-variables/dev.tfvars
@@ -11,4 +11,4 @@ paas_app_stopped = false
 paas_web_app_deployment_strategy = "blue-green-v2"
 paas_web_app_instances = 1
 paas_web_app_memory = 1024
-paas_web_app_start_command = "bundle exec rake cf:on_first_instance db:migrate && /app/bin/delayed_job --pool=mailers --pool=*:2 start && rails db:safe_reset && rails cron:schedule && rails s"
+paas_web_app_start_command = "bundle exec rake cf:on_first_instance db:migrate s3:delete_csvs && /app/bin/delayed_job --pool=mailers --pool=*:2 start && rails db:safe_reset && rails cron:schedule && rails s"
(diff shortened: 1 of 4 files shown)


terraform/workspace-variables/dev.tfvars
```diff
@@ -11,4 +11,4 @@ paas_app_stopped = false
 paas_web_app_deployment_strategy = "blue-green-v2"
 paas_web_app_instances = 1
 paas_web_app_memory = 1024
-paas_web_app_start_command = "bundle exec rake cf:on_first_instance db:migrate && /app/bin/delayed_job --pool=mailers --pool=*:2 start && rails db:safe_reset && rails cron:schedule && rails s"
+paas_web_app_start_command = "bundle exec rake cf:on_first_instance db:migrate s3:delete_csv && /app/bin/delayed_job --pool=mailers --pool=*:2 start && rails db:safe_reset && rails cron:schedule && rails s"
```
kicferk1, line 14: You need an extra `s` at the end of the command name - `delete_csv` vs `delete_csvs` here and in review tfvars file.

kicferk1 (commented):
Code looks pretty good to me, but the proof of the eating is in the pudding or something. We should upload some csv on this PR and see if the destroy workflow triggered on close works.

terraform/workspace-variables/dev.tfvars
ethanmills: Yep, spotted that, thanks!

ethanmills:
Will come back to this later, closing for cleanliness

PR #570, opened 2021-06-22
CPDTP-226 add in strong migrations gem

blackrat:
### Context

As part of a previous PR review it was suggested that we add in the `strong_migrations' gem to ensure that any future migrations are checked for any dangerous activity and any preferred alternative solutions are used instead. This gem enables warnings and proposes changes to ensure that the database integrity is preserved.

Add in and initialise the `strong_migrations` gem.

### Review Checks

### DIFF
--- a/config/initializers/strong_migrations.rb
+++ b/config/initializers/strong_migrations.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+# Mark existing migrations as safe
+StrongMigrations.start_after = 20_210_622_170_835
+
+# Set timeouts for migrations
+# If you use PgBouncer in transaction mode, delete these lines and set timeouts on the database user
+StrongMigrations.lock_timeout = 10.seconds
+StrongMigrations.statement_timeout = 1.hour
+
+# Analyze tables after indexes are added
+# Outdated statistics can sometimes hurt performance
+StrongMigrations.auto_analyze = true
+
+# Set the version of the production database
+# so the right checks are run in development
+# StrongMigrations.target_version = 10
+
+# Add custom checks
+# StrongMigrations.add_check do |method, args|
+#   if method == :add_index && args[0].to_s == "users"
+#     stop! "No more indexes on the users table"
+#   end
+# end
+
+# Make some operations safe by default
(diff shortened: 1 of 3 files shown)


config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+# Mark existing migrations as safe
+StrongMigrations.start_after = 20_210_622_170_835
```
kicferk1, line 4: `20_210_622_170_835`? 😂 
`2021_06_22_17_08_35` is what this really comes from, right? I think. Does it really force you to specify the datetime as an integer?

config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1,28 @@
...
+# Add custom checks
+# StrongMigrations.add_check do |method, args|
+#   if method == :add_index && args[0].to_s == "users"
+#     stop! "No more indexes on the users table"
+#   end
+# end
+
+# Make some operations safe by default
+# See https://github.com/ankane/strong_migrations#safe-by-default
+# StrongMigrations.safe_by_default = true
```
kicferk1, line 28: Do we want to turn this on?

blackrat: Probably not since it includes setting not null on an existing column.
From the readme.md: 

> **Safe by Default**
> Make operations safe by default.
> 
> - adding and removing an index
> - adding a foreign key
> - adding a check constraint
> - setting NOT NULL on an existing column
>

config/initializers/strong_migrations.rb
blackrat: It does, and rubocop changes it to that format.

blackrat: It uses the database version as a cause for comparison:
`
  def self.check_enabled?(check, version: nil)
    if enabled_checks[check]
      start_after = enabled_checks[check][:start_after] || StrongMigrations.start_after
      !version || version > start_after
    else
      false
    end
  end
`

jamesrose (approved):

PR #572, opened 2021-06-23
Change page width and heading style

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/browse/CPDRP-324)
Fix page width and heading style on choose CIP training materials page

Add elements to make page width `govuk-grid-column-three-quarters`
Change style of `h1` heading to `govuk-heading-xl`

### Review Checks

image

### DIFF
--- a/app/views/schools/core_programme/materials/edit.html.erb
+++ b/app/views/schools/core_programme/materials/edit.html.erb
@@ -2,11 +2,19 @@
 
 <% content_for :before_content, govuk_back_link( text: "Back", href: { action: :info }) %>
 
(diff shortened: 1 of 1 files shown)


ethanmills (approved):

kicferk1:
Why are we making some pages `two-thirds` width and others `three-quarters`? Are there any rules guiding it?

tonyheadford:
> Why are we making some pages `two-thirds` width and others `three-quarters`? Are there any rules guiding it?

I think generally we'd use two-thirds as per GDS guidance, but the designer wants three-quarters in some instances to provide a little more width. I don't think there are any hard and fast rules, just what the designer feels is most appropriate I guess.

PR #574, opened 2021-06-23
Send partnership notification emails to support where the is no school email

ethanmills:
Support need to follow up with these schools

Resolves https://sentry.io/organizations/dfe-bat/issues/2447441719/?project=5748989&query=is%3Aunresolved


kicferk1 (commented):
Just for this one email? What about nomination email?

Also, those emails will look weird in the support mailbox. But I guess if they know that they will get them it is fine.

tonyheadford (approved):
LGTM

PR #577, opened 2021-06-23
CDPTP-195 Set payment uplift flags on new EarlyCareerTeacherProfile records

blackrat:
### Context

Track and Pay require flags against the Early Career Teacher Profiles to ensure that we are picking up the participants
to whom uplift is required. The decision about the flags boils down to the fact that we don't care if the underlying information changes, we just care if it was applicable at the point of the registration.

This is to ensure that when a new registration happens, we set those flags against the early career teacher.

Add code to pick up the current uplift flags from sparsity and pupil premium and set them in the early career teacher profile.

### Guidance to review

Would rather have this in a service or other separated section, but there's been some work on building a multistep form that I don't want to refactor at the moment to make this more separated, so it's a callback for now.

### Testing

New early_career_teacher_profile records will have the appropriate flags set.

### Review Checks

No visible changes.

### DIFF
--- a/app/models/participant_declaration.rb
+++ b/app/models/participant_declaration.rb
@@ -4,18 +4,19 @@ class ParticipantDeclaration < ApplicationRecord
   belongs_to :lead_provider
   belongs_to :early_career_teacher_profile
 
+  # Helper scopes
   scope :for_lead_provider, ->(lead_provider) { where(lead_provider: lead_provider) }
-  scope :active_for_lead_provider, ->(lead_provider) { active.for_lead_provider(lead_provider) }
-  scope :count_active_for_lead_provider, ->(lead_provider) { active_for_lead_provider(lead_provider).count }
-  scope :count_active_for_lead_provider_between, ->(lead_provider, start_date, end_date) { declared_as_between(start_date, end_date).count_active_for_lead_provider(lead_provider) }
-  scope :count_active_uplift_for_lead_provider, ->(lead_provider) { active_uplift_for_lead_provider(lead_provider).count }
-
+  scope :for_declaration, ->(declaration_type) { where(declaration_type: declaration_type) }
+  scope :started, -> { for_declaration("started").order(declaration_date: "desc").unique_early_career_teacher_profile_id }
+  scope :uplift, -> { joins(:early_career_teacher_profile).merge(EarlyCareerTeacherProfile.uplift) }
   scope :unique_early_career_teacher_profile_id, -> { select(:early_career_teacher_profile_id).distinct }
-  scope :active, -> { where(declaration_type: "started").order(declaration_date: "desc").unique_early_career_teacher_profile_id }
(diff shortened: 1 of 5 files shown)


app/models/participant_declaration.rb
```diff
@@ -4,18 +4,19 @@ class ParticipantDeclaration < ApplicationRecord
...
   scope :declared_as_between, ->(start_date, end_date) { where(declaration_date: start_date..end_date) }
   scope :submitted_between, ->(start_date, end_date) { where(created_at: start_date..end_date) }
-  scope :active_uplift_for_lead_provider, lambda { |lead_provider|
-    active_for_lead_provider(lead_provider).joins(:early_career_teacher_profile)
-      .where("early_career_teacher_profiles.sparsity_uplift OR early_career_teacher_profiles.pupil_premium_uplift")
-  }
+
+  # Declaration aggregation scopes
+  scope :count_active_for_lead_provider, ->(lead_provider) { active_for_lead_provider(lead_provider).count }
+  scope :count_active_uplift_for_lead_provider, ->(lead_provider) { active_for_lead_provider(lead_provider).uplift.count }
```
plisovin, line 21: Looks much cleaner now

spec/services/early_career_teachers/create_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+RSpec.describe EarlyCareerTeachers::Create do
+  let(:user) { create :user }
+  let(:school) { create :school }
+  let(:pupil_premium_school) { create :school, :pupil_premium_uplift }
+  let(:sparsity_school) { create :school, :sparsity_uplift }
+  let(:uplift_school) { create :school, :sparsity_uplift, :pupil_premium_uplift }
+  let(:cohort) { create :cohort, :current }
+  let(:mentor_profile) { create :mentor_profile }
```
kicferk1, line 10: Should we test that mentor profile is set on ect profile somewhere?

spec/services/early_career_teachers/create_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  it "creates an Early Career Teacher Profile record" do
+    before_count = EarlyCareerTeacherProfile.count
+    early_career_teacher_profile = described_class.call(email: user.email, full_name: user.full_name, school_id: school.id, cohort_id: cohort.id, mentor_profile_id: mentor_profile.id)
+    after_count = EarlyCareerTeacherProfile.count
+    expect(after_count).to eq(before_count + 1)
```
kicferk1, line 16: I would make this its own test, without checking anything else. And use the form of 
```
    expect {
      described_class.call(email: user.email, full_name: user.full_name, school_id: school.id, cohort_id: cohort.id, mentor_profile_id: mentor_profile.id)
    }.to change { EarlyCareerTeacherProfile.count }.by(1)
```

Also, should we add a test for the edge case when user does not exist? And the test could say that 'when the user does not exist, it creates a new user'.

ethanmills (approved):
Happy with this once the test comments are addressed

kicferk1 (approved):

PR #578, opened 2021-06-23
Make school tables responsive using NHS responsive table component

callumacrae:
### Context

https://dfedigital.atlassian.net/browse/CPDRP-452

the nhs responsive table is real nice so i added it to our codebase (after switching all the helpers to `govuk-` ones)

It didn't seem to actually add any styling which is ideal - the NHS design system seems to just style all `table` elements. That's why the elements have both `govuk-` and `nhsuk-` classes :)

### Review Checks

### DIFF
--- a/app/components/lead_providers/your_schools/table.html.erb
+++ b/app/components/lead_providers/your_schools/table.html.erb
@@ -1,16 +1,18 @@
-<table data-test="schools-table" class="govuk-table lead-provider-your-schools-table">
-  <thead>
-    <tr class="govuk-table__row">
-      <th scope="col" class="govuk-table__header">School</th>
-      <th scope="col" class="govuk-table__header">URN</th>
-      <th scope="col" class="govuk-table__header">Delivery partner</th>
-      <th scope="col" class="govuk-table__header govuk-table__header--numeric">ECTs</th>
-      <th scope="col" class="govuk-table__header govuk-table__header--numeric">Mentors</th>
-    </tr>
-  </thead>
-  <tbody class="govuk-table__body">
-    <%= render LeadProviders::YourSchools::TableRow.with_collection(partnerships) %>
-  </tbody>
-</table>
+<div class="nhsuk-table-container">
+  <table data-test="schools-table" class="govuk-table nhsuk-table-responsive lead-provider-your-schools-table">
+    <thead  class="nhsuk-table__head">
+      <tr class="govuk-table__row">
+        <th scope="col" class="govuk-table__header">School</th>
+        <th scope="col" class="govuk-table__header">URN</th>
+        <th scope="col" class="govuk-table__header">Delivery partner</th>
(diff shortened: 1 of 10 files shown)


app/components/lead_providers/your_schools/table.html.erb
```diff
@@ -1,16 +1,18 @@
...
-    </tr>
-  </thead>
-  <tbody class="govuk-table__body">
-    <%= render LeadProviders::YourSchools::TableRow.with_collection(partnerships) %>
-  </tbody>
-</table>
+<div class="nhsuk-table-container">
+  <table data-test="schools-table" class="govuk-table nhsuk-table-responsive lead-provider-your-schools-table">
+    <thead  class="nhsuk-table__head">
+      <tr class="govuk-table__row">
```
kicferk1, line 4: Does this want `nhsuk-table__row`?

app/views/admin/schools/index.html.erb
```diff
@@ -15,26 +15,43 @@
...
-      <th scope="col" class="govuk-table__header">Induction tutor</th>
-    </tr>
-  </thead>
-  <tbody class="govuk-table__body">
-    <% @schools.each do |school| %>
+<div class="nhsuk-table-container">
+  <table class="govuk-table nhsuk-table-responsive">
+    <thead class="nhsuk-table__head">
       <tr class="govuk-table__row">
-        <td class="govuk-table__cell"><%= govuk_link_to school.name, admin_school_path(school) %></td>
```
kicferk1, line 31: Ditto here, does this want `nhsuk-table__row`?

app/webpacker/styles/vendor/_nhs_tables.scss
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/* Numeric tables
+========================================================================== */
+
+/**
+ * Right aligns table cells for numeric tables.
+ */
+
+.nhsuk-table__header--numeric,
```
kicferk1, line 167: Is this used? Should it be?

app/components/lead_providers/your_schools/table.html.erb
callumacrae: Not required in header, from the looks of things: https://service-manual.nhs.uk/design-system/components/table

app/webpacker/styles/vendor/_nhs_tables.scss
callumacrae: nope, but we're using the govuk class to do the same thing. I wanted to change this file as little as possible which is why there's a few things in here that aren't used.

app/views/admin/schools/index.html.erb
callumacrae: same as above (even though github has put your comment in a weird place - i think you're talking about the line on the other side of the diff?)

kicferk1 (approved):

PR #580, opened 2021-06-23
Reduce alerting noise

ethanmills:
- Add separate alert configs per environment
- Relax dev alert criteria
- Reduce severity of alerts in other environments
- Change monitoring stack to fork to allow setting slack template
- Only notify @channel for high severity alerts


jamesrose (approved):
Awesome, thanks!

PR #582, opened 2021-06-24
Send emails to MAT schools

ethanmills:
### Context
I've been given a list of URNs and told to send the schools an email with a nomination link

### DIFF
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -97,6 +97,32 @@ class InviteSchools
     end
   end
 
+  def invite_mats(school_urns)
+    school_urns.each do |urn|
+      school = School.eligible.find_by(urn: urn)
+
+      if school.nil?
+        logger.info "School not found, urn: #{urn} ... skipping"
+        next
+      end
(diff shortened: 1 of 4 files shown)


app/services/invite_schools.rb
```diff
@@ -97,6 +97,32 @@ def invite_to_beta(school_urns)
...
+
+      if school.nil?
+        logger.info "School not found, urn: #{urn} ... skipping"
+        next
+      end
+
+      if school.contact_email.blank?
+        logger.info "No contact details for school urn: #{urn} ... skipping"
+        next
+      end
```
kicferk1, line 112: Feels like this logic gets repeated a lot in this file. Should it be extracted to something like `find_school_to_email` method that can return `nil`, and log some relevant info?

kicferk1 (approved):

app/services/invite_schools.rb
ethanmills: Agreed. I'm going to put this in for now, then take a look

PR #584, opened 2021-06-24
/api/v1/users endpoint now supports email filter

asmega:
### Context

- ECF does not allow users to have duplicate emails (for obvious reasons)
- NPQ registration can create user accounts, however it is possible for a user to already exist in ECF with this same email address
- NPQ registration should check if such an account exist and use that if it exists otherwise create the account

- Allow the endpoint `/api/v1/users` to support filtering by email

### Guidance to review

- Should not negatively impact the existing endpoint for other integrations

### Testing

- Test in review app

### Review Checks

- note down the token
- create a new user via command line or find an existing user and note down the email

```
curl -H "Content-Type: application/vnd.api+json" -H "Authorization: Bearer TOKEN" "https://ecf-review-pr-584.london.cloudapps.digital/api/v1/users?filter[email]=EMAIL"
```

- should return the found user

### DIFF
--- a/spec/requests/api/v1/users_spec.rb
+++ b/spec/requests/api/v1/users_spec.rb
@@ -87,6 +87,21 @@ RSpec.describe "API Users", type: :request do
         get "/api/v1/users", params: { filter: { updated_since: 1.day.ago.iso8601 } }
         expect(parsed_response["data"].size).to eql(2)
       end
+
+      context "when filtering by email" do
+        it "returns users that match" do
+          email = User.all.sample.email
+          get "/api/v1/users", params: { filter: { email: email } }
+          expect(parsed_response["data"].size).to eql(1)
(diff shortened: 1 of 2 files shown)


spec/requests/api/v1/users_spec.rb
```diff
@@ -87,6 +87,21 @@
         get "/api/v1/users", params: { filter: { updated_since: 1.day.ago.iso8601 } }
         expect(parsed_response["data"].size).to eql(2)
       end
+
+      context "when filtering by email" do
+        it "returns users that match" do
+          email = User.all.sample.email
```
kicferk1, line 93: Out of curiosity, does `User.sample` do the same trick as `User.all.sample`?

kicferk1 (approved):

spec/requests/api/v1/users_spec.rb
asmega: nah doesn't work

PR #586, opened 2021-06-24
Increased rate limit of nomination emails

BroiSatse:
From 1 per 24 hours to 1 per 5 minutes and 5 per 24 hours.

https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-523


app/controllers/nominations/request_nomination_invite_controller.rb
```diff
@@ -36,7 +36,7 @@ def receive_school
       redirect_to not_eligible_request_nomination_invite_path
     elsif @nomination_request_form.school.registered?
       redirect_to already_nominated_request_nomination_invite_path
-    elsif @nomination_request_form.email_limit_reached?
+    elsif @nomination_request_form.reached_email_limit.present?
```
kicferk1, line 39: Why do we need to check for `present?`? I would expect the form to give us nice, clean boolean

BroiSatse: It doesn't - this is why I renamed the method from `email_limit_reached` to `reached_email_limit`. It can be `nil` in most of the cases, but if one of the limits is reached it will return that limit details.

app/views/nominations/request_nomination_invite/limit_reached.html.erb
```diff
@@ -2,9 +2,8 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-header-xl">You can only send one email per day</h1>
-    <p>You have already sent an email today to nominate your induction lead or tutor.</p>
-    <p>You can try again tomorrow.</p>
+    <h1 class="govuk-header-xl">You can only send <%= @limit[:max] %> email per <%= @limit[:within].inspect %></h1>
```
ethanmills, line 5: `email.pluralize(@limit[:max])`?

ethanmills (approved):

PR #588, opened 2021-06-24
Escape special characters to fix flakey test

ethanmills:
Because #587 really shouldn't have broken any tests


kicferk1 (approved):

tonyheadford (commented):
Would prefer if we didn't do this, we can check `assigns` for the right objects or we write feature tests to check content. Requests specs aren't really meant for this which is why this breaks

jamesrose (approved):

ethanmills:
@tonyheadford happy to chat about test strategy more generally at a lunchtime catchup. I didn't write this test, just noticed that it was failing

PR #590, opened 2021-06-25
Content security policy

mattonfoot:
### Context
The previous version of the Content Security Policy blocked the API docs and didn't report correctly. 

This change adds a very permissive CSP policy backed by a restrictive "report only" policy. It also addresses the lack of screen shots on the API Docs that would have identified the blocked content.

### Testing
- Cypress tests for sandbox modified
- csp_report_controller_spec checks logging for expected logs

### Review Checks
- [ ] API docs page have visual tests via cypress + percy
- [ ] testing is of sufficient level - Q: do we need tests that will break if the CSP headers change?
- [ ] Check the CSP to see if anything specific stands out as missing

The developer console will need to be up. As you browse the website you should see no CSP report errors displayed or the only ones you find should be labeled "report only".

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -11,10 +11,7 @@ SecureHeaders::Configuration.default do |config|
 
   google_analytics = %w[www.google-analytics.com ssl.google-analytics.com *.googletagmanager.com tagmanager.google.com *.googleusercontent.com *.gstatic.com]
 
-  config.csp = {
-    default_src: %w[* data: 'unsafe-inline'],
-    script_src: %w[* 'unsafe-inline'],
-  }
+  config.csp = SecureHeaders::OPT_OUT
 
   config.csp_report_only = {
     default_src: %w['none'],
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -11,21 +11,26 @@ SecureHeaders::Configuration.default do |config|
 
   google_analytics = %w[www.google-analytics.com ssl.google-analytics.com *.googletagmanager.com tagmanager.google.com *.googleusercontent.com *.gstatic.com]
 
+  config.csp = {
+    default_src: %w['self' http: https: data: 'unsafe-inline'],
+    script_src: %w['self' http: https: 'unsafe-inline'],
+  }
+
   config.csp_report_only = {
     default_src: %w['none'],
     base_uri: %w['self'],
     block_all_mixed_content: true, # see http://www.w3.org/TR/mixed-content/
     child_src: %w['self'],
     connect_src: %W['self'] + google_analytics,
-    font_src: %w['self' *.gov.uk fonts.gstatic.com fonts.googleapis.com],
+    font_src: %w['self' *.gov.uk fonts.gstatic.com],
     form_action: %w['self'],
     frame_ancestors: %w['self'],
     frame_src: %w['self'],
-    img_src: %W['self' *.gov.uk] + google_analytics,
+    img_src: %W['self' data: *.gov.uk] + google_analytics,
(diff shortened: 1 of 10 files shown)


spec/cypress/integration/Sandbox.feature
```diff
@@ -13,5 +13,4 @@ Feature: Sandbox landing page
     Given I am on "the sandbox landing page" page
     When I click on "link" containing "Review our API documentation"
     Then I should be on "API Documentation" page
-    And the page should be accessible
```
timabell: I think the front page should still be checked for a11y

config/initializers/secure_headers.rb
```diff
@@ -0,0 +1,39 @@
...
+  config.x_download_options = "noopen"
+  config.x_permitted_cross_domain_policies = "none"
+  config.referrer_policy = %w[origin-when-cross-origin strict-origin-when-cross-origin]
+
+  google_analytics = %w[www.google-analytics.com ssl.google-analytics.com *.googletagmanager.com tagmanager.google.com *.googleusercontent.com *.gstatic.com]
+
+  config.csp = {
+    default_src: %w[* data: 'unsafe-inline'],
+    script_src: %w[* 'unsafe-inline'],
+  }
```
ethanmills, line 17:
```suggestion
  config.csp = SecureHeaders::OPT_OUT
```

ethanmills (changes requested):
Let's be extra safe about this one

config/initializers/secure_headers.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# rubocop:disable Lint/PercentStringArray
+SecureHeaders::Configuration.default do |config|
+  config.x_frame_options = "DENY"
```
ethanmills, line 5: Is this going to affect GA? I think that falls back to a frame for no-js

mattonfoot: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options it stops your website rendering in an iframe

ethanmills: We have

```
    <!-- Google Tag Manager (noscript) -->
    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-N58Z5PG"
    height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
    <!-- End Google Tag Manager (noscript) -->
```

ethanmills (approved):
I've check the review app, I'm seeing a report-only (only) policy, and violations are successfully making their way through to Logit. Slight concern about GA with frames, but it should be a minor of users

config/initializers/secure_headers.rb
mattonfoot: thats us including an iframe not our website in an iframe on another website

mattonfoot: So I think it is these lines that will cause issues for the google iframe 

frame_ancestors: %w['self'],
frame_src: %w['self'],

So will do some investigation

ethanmills: Ah gotcha. Looks good then

ethanmills: Those are both report-only, so we can mop that up with the rest

config/initializers/secure_headers.rb
```diff
@@ -0,0 +1,36 @@
...
+  config.x_content_type_options = "nosniff"
+  config.x_xss_protection = "1; mode=block"
```
ethanmills, line 7: Conflicts with #604

mattonfoot: Good spot, adjusted to suit PEN test

PR #591, opened 2021-06-25
Send beta chaser emails

ethanmills:
Email beta schools without participants again

### DIFF
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -10,12 +10,8 @@ class InviteSchools
     logger.info "Emailing schools"
 
     school_urns.each do |urn|
-      school = School.eligible.find_by(urn: urn)
-
-      if school.nil?
-        logger.info "School not found, urn: #{urn} ... skipping"
-        next
-      end
+      school = find_eligible_school(urn)
+      next if school.nil?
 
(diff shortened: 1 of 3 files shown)


app/services/invite_schools.rb
```diff
@@ -90,21 +78,40 @@ def invite_to_beta(school_urns)
...
+    beta_schools = School.where(id: beta_school_ids)
+    beta_schools_without_participants = beta_schools.left_outer_joins(:early_career_teacher_profiles)
+                                                    .left_outer_joins(:mentor_profiles)
+                                                    .where(
+                                                      early_career_teacher_profiles: { id: nil },
+                                                      mentor_profiles: { id: nil },
+                                                    )
+
+    beta_schools_without_participants.each do |school|
+      induction_coordinator = school.induction_coordinators.first
```
kicferk1, line 99: will that always be non-nil?

kicferk1 (approved):

app/services/invite_schools.rb
kicferk1: (by `this` I mean induction coordinator whose email we use later)

ethanmills: There's an edge case where we invite a school to beta then remove their induction coordinator. I'm happy for that to fail though

PR #592, opened 2021-06-25
Ensure email does not end with a dot

ethanmills:
Sentry has caught errors caused by people entering emails ending with a
dot

### DIFF
--- a/app/validators/notify_email_validator.rb
+++ b/app/validators/notify_email_validator.rb
@@ -19,6 +19,7 @@ class NotifyEmailValidator < ActiveModel::EachValidator
     return false if email.include?("..")
 
     hostname = email_match[1]
+    return false unless hostname.last.match(/[a-z0-9]/i)
 
     parts = hostname.split(".")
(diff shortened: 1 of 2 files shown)


kicferk1 (approved):

kicferk1:
We got the initial validation from notify, right? Should they also do that check? Do they already do it?

ethanmills:
They already do the check, it's a difference in `split` between python and ruby. If you have "a.b." python gives `['a', 'b', '']` and ruby gives `["a", "b"]`

timabell:
Incidentally, a dot at the end of a domain is valid and means it's fully qualified (FQDN) http://www.dns-sd.org/trailingdotsindomainnames.html

ethanmills:
> Incidentally, a dot at the end of a domain is valid and means it's fully qualified (FQDN) http://www.dns-sd.org/trailingdotsindomainnames.html

Yep! However, Notify accept a subset of RFC compliant emails, and we're mostly interested in pleasing them

PR #596, opened 2021-06-28
CPDTP-204 Fix issue with undefined local variable RecordParticipantEvent

plisovin:
### Context

The issue appeared when making call to the API for declarations

```
undefined local variable or method `lead_provider'
```

Fixed rspec test to catch this error

### DIFF
--- a/app/services/record_participant_event.rb
+++ b/app/services/record_participant_event.rb
@@ -59,7 +59,7 @@ private
   end
 
   def validate_provider!
-    raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || lead_provider == actual_lead_provider
+    raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || @params[:lead_provider] == actual_lead_provider
   end
 
   def required_params
--- a/app/services/record_participant_event.rb
+++ b/app/services/record_participant_event.rb
@@ -54,12 +54,16 @@ private
     ParticipantDeclaration.create(@params.slice(*required_params))
   end
 
+  def lead_provider
+    params[:lead_provider]
+  end
+
   def actual_lead_provider
     SchoolCohort.find_by(school: early_career_teacher_profile.school, cohort: early_career_teacher_profile.cohort)&.lead_provider
   end
 
   def validate_provider!
-    raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || @params[:lead_provider] == actual_lead_provider
+    raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || lead_provider == actual_lead_provider
   end
 
   def required_params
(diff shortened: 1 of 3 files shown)


app/services/record_participant_event.rb
```diff
@@ -59,7 +59,7 @@ def actual_lead_provider
   end
 
   def validate_provider!
-    raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || lead_provider == actual_lead_provider
+    raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || @params[:lead_provider] == actual_lead_provider
```
blackrat, line 62: params is defined using an attr_accessor, so the "@" isn't required.
Originally `lead_provider` was a method generated by InitializeWithConfig. May be better to keep it as a method in case we undo the previous refactor that took that out.

spec/requests/api/v1/provider_events_spec.rb
```diff
@@ -8,6 +8,15 @@
...
     let(:payload) { create(:early_career_teacher_profile) }
+    let(:delivery_partner) { create(:delivery_partner) }
+    let!(:school_cohort) { create(:school_cohort, school: payload.school, cohort: payload.cohort) }
+    let!(:partnership) do
+      create(:partnership,
+             school: payload.school,
+             lead_provider: lead_provider,
+             cohort: payload.cohort,
+             delivery_partner: delivery_partner)
+    end
```
kicferk1, line 19: So we added a bunch of setup, but we haven't added any tests. I would expect a unit test like "when lead providers don't match, recording participant event raises an error".

Probably not in this file, I imagine `record_participant_event` could use its own test suite to demonstrate and test those edge cases.

spec/services/record_participant_service_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  let!(:partnership) do
+    create(:partnership,
+           school: ect_profile.school,
+           lead_provider: lead_provider,
+           cohort: ect_profile.cohort,
+           delivery_partner: delivery_partner)
+  end
+
+  context "when lead providers don't match" do
+    it "should make recording participant event raise an error" do
```
kicferk1, line 29: Ah, I see you copied my words exactly.

I think in the context of the test, this description should be "raises a ParameterMissing error" :)

kicferk1 (approved):

spec/services/record_participant_service_spec.rb
plisovin: ah yeah, thanks

Cypress test seems to fail though for some reason

kicferk1: I think this cypress failure was fixed on `develop` on Friday, try rebasing on top of that?

kicferk1: Or maybe that particular error is fixed on https://github.com/DFE-Digital/early-careers-framework/pull/598 - if a quick rebase doesn't work, wait for Ethan's second fix to go in :)

PR #597, opened 2021-06-28
CPDTP-204 Don't output mentor ids in sandbox seed output

timabell:
### Context

You can't currently submit a declaration against mentor user ids so this was a pit of failure... into which I fell.

* Don't output mentor ids in sandbox seed output.

### Guidance to review

:ship:

### Testing

* rails db:reset
* See new output with less ids
* Optionally submit one of the ids to the declaration endpoint, but note there's another bug blocking this being fixed in #596


kicferk1 (approved):
Will we need to bring it back in one way or another soon-ish (given that mentors are participants)?

timabell:
> Will we need to bring it back in one way or another soon-ish (given that mentors are participants)?

Yes indeed, though probably not in exactly this form.

ethanmills:
Sorry!

PR #599, opened 2021-06-28
Added Javascript sentry

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-487

The above bug was (most likely) caused by the double form submission, which somehow got around our JS shields. I **guess** this could be caused by some JS error we don't know about, so this PR introduces client-side Sentry error collecting. Let's hope we won't be flooded with errors...

### DIFF
--- a/app/views/layouts/_application.html.erb
+++ b/app/views/layouts/_application.html.erb
@@ -1,6 +1,13 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template ">
   <head>
+    <% if Rails.env.to_s.in? Sentry.configuration.enabled_environments %>
+      <%= tag :meta, name: :release, content: Sentry.configuration.release %>
+      <%= tag :meta, name: :"sentry-dsn", content: Sentry.configuration.dsn.to_s %>
+      <%= tag :meta, name: :"sentry-trace", content: Sentry.get_current_scope&.span&.to_sentry_trace %>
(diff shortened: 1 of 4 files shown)


app/views/layouts/_application.html.erb
```diff
@@ -1,6 +1,12 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template ">
   <head>
+    <% if Rails.env.to_s.in? Sentry.configuration.enabled_environments %>
+      <%= tag :meta, name: :release, content: Sentry.configuration.release %>
+      <%= tag :meta, name: :"sentry-dsn", content: Sentry.configuration.dsn.to_s %>
+      <%= javascript_pack_tag('sentry') unless params[:nojs] === "nojs" %>
```
kicferk1, line 7: two equal signs please, we are still writing Ruby code here :)

ethanmills:
Do we need to do this? https://docs.sentry.io/platforms/javascript/performance/connect-services/#pageload

app/views/layouts/_application.html.erb
BroiSatse: Good catch, I'll correct this in the place I've copied it from as well :P

ethanmills (approved):

kicferk1 (approved):

PR #601, opened 2021-06-28
CPDTP-204 Return `id` of newly created declaration in API

timabell:
### Context

This patch allows providers to (at a later time) reference declarations they have created by returning to them the id of the newly created declaration record.

This may be needed to allow them to "void" declarations when they realise there has been a mistake (to be confirmed).

Currently based on branch for #596

* Refactor params validation
* Return id of newly created declaration in API

### Guidance to review

Have a look at the updated swagger, and try it out.

Here is a successful request with `id` returned:

successful request with `id` returned

### Testing

* run db:reset to get seed ids (watch out for just-merged #597 that isn't in this branch yet)
* use swagger to post real declarations
* see declaration id now in returned json
 
https://ecf-review-pr-601.london.cloudapps.digital/api-docs/index.html

### Pre-merge todo list

- [x] #596 merged
- [x] Change base branch

### DIFF
--- a/spec/docs/participant_declarations_spec.rb
+++ b/spec/docs/participant_declarations_spec.rb
@@ -63,7 +63,7 @@ RSpec.describe "Participant Declarations", type: :request, swagger_doc: "v1/api_
         },
       }, description: "The event declaration date"
 
-      response 204, "Successful" do
+      response 200, "Successful" do
         let(:fresh_user) { create(:user, :early_career_teacher) }
         let(:params) do
           {
@@ -75,7 +75,7 @@ RSpec.describe "Participant Declarations", type: :request, swagger_doc: "v1/api_
         run_test!
       end
 
-      response 204, "Successful" do
+      response 200, "Successful" do
         let(:params) do
           {
             "participant_id" => user.id,
@@ -87,6 +87,8 @@ RSpec.describe "Participant Declarations", type: :request, swagger_doc: "v1/api_
         before do
           RecordParticipantEvent.call(HashWithIndifferentAccess.new({ lead_provider: lead_provider, raw_event: params.to_json }).merge(params))
         end
+
+        schema "$ref": "#/components/schemas/id"
 
         run_test!
       end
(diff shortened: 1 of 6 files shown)


spec/requests/api/v1/provider_events_spec.rb
```diff
@@ -37,14 +37,19 @@
...
 
-      it "returns 204 status when successful" do
-        post "/api/v1/participant-declarations", params: params.to_json
-        expect(response.status).to eq 204
+      it "create declaration record and return id when successful" do
+        expect {
+          post "/api/v1/participant-declarations", params: params.to_json
+        }.to change(ParticipantDeclaration, :count).by(1)
+        expect(response.status).to eq 200
+        expect(parsed_response["id"]).to eq(ParticipantDeclaration.last.id)
```
kicferk1, line 52: `ParticipantDeclaration .order(:created_at).last` please, `last` the default db order is... alphabetical order of uids (I think).

timabell: Done, thanks for that one. Much more robust.

spec/docs/participant_declarations_spec.rb
```diff
@@ -88,6 +88,8 @@
           RecordParticipantEvent.call(HashWithIndifferentAccess.new({ lead_provider: lead_provider, raw_event: params.to_json }).merge(params))
         end
 
+        schema "$ref": "#/components/schemas/id"
```
kicferk1, line 91: Does this work in swagger?

I thought we didn't have anything like that set up (and it is a little bit of a faff to do)

kicferk1: Well, apparently it does

kicferk1 (approved):

spec/docs/participant_declarations_spec.rb
timabell: It seems it does, it'd based on the return type we already had in place for failures. I've updated the PR description with a screenshot of swagger so you can see it documenting the returned json (as well as a real response)

jamesrose (approved):

PR #606, opened 2021-06-29
Participant profiles

BroiSatse:
### Context

This PR combines mentor_profiles and early_career_teacher_profile in a single participant_profile table.

Note: This is in no way completed refactor, just the first step. There's a lot of coupling between profiles and user models and those will be relaxed in following PRs. 

As this PR is application, it will conflict all the time, so please do not block this PR on non-critical issues as those will be fixed in following PRs.

### Notes

I've tried to keep the scope of this PR to an absolute minimum, which is why it introduces a few oddities, like frequent repetition of `profile.user.full_name`. This is because many methods and objects are still expecting an instance of `User` and not `ParticipationProfile` - this will be fixed in the next PRs.

I have not yet remove the `EarlyCareerTeacherProfile` nor `MentorProfile` models from the code. This is to allow some additional data integrity checks after this is deployed. I've checked however that both classes are not being used.


app/forms/schools/add_participant_form.rb
```diff
@@ -65,11 +65,13 @@ def type_options
     end
 
     def can_add_self?
-      User.is_participant.in_school(school_cohort.school).exclude? current_user
+      ParticipantProfile
+        .where(school: school_cohort.school)
+        .none?
```
ethanmills, line 70: Is this right?

app/models/user.rb
```diff
@@ -103,9 +106,9 @@ def school
     includes(early_career_teacher_profile: %i[cohort school], mentor_profile: %i[cohort school])
   }
 
-  scope :is_participant, lambda {
-    includes_school.where.not(early_career_teacher_profile: { id: nil }).or(User.where.not(mentor_profile: { id: nil }))
-  }
+  # scope :is_participant, lambda {
+  #   includes_school.where.not(early_career_teacher_profile: { id: nil }).or(User.where.not(mentor_profile: { id: nil }))
+  # }
```
ethanmills, line 111: Remove?

app/forms/schools/add_participant_form.rb
kicferk1, line 70: `school_cohort.school.participant_profiles.none?`?

db/migrate/20210629104700_populate_participant_profiles.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class PopulateParticipantProfiles < ActiveRecord::Migration[6.1]
```
ethanmills, line 3: How long will this take to run on 2000 participants? I believe all migrations need to run in <60s for deploy to succeed

app/models/participant_profile.rb
```diff
@@ -0,0 +1,39 @@
...
+  class ECT < self
+    belongs_to :mentor_profile, class_name: "Mentor", optional: true
+
+    def ect?
+      true
+    end
+  end
+
+  class Mentor < self
+    self.ignored_columns = %i[mentor_profile_id]
```
kicferk1, line 33: What does this do?

app/models/school.rb
```diff
@@ -74,7 +77,7 @@ def delivery_partner_for(year)
   end
 
   def early_career_teacher_profiles_for(year)
-    early_career_teacher_profiles.joins(:cohort).where(cohort: { start_year: year })
+    participant_profiles.ects.where(cohort_id: Cohort.where(start_year: year).select(:id))
```
kicferk1, line 80: Do we need to change to use `cohort_id` from `cohort` in that query?

db/migrate/20210629104700_populate_participant_profiles.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  class MentorProfile < ApplicationRecord
+  end
+
+  class ParticipantProfile < ApplicationRecord
+    self.inheritance_column = nil
+  end
+
+  def up
+    MentorProfile.find_each do |ect_profile|
```
kicferk1, line 15: `mentor_profile`

db/migrate/20210629104700_populate_participant_profiles.rb
```diff
@@ -0,0 +1,40 @@
...
+      ParticipantProfile.create!(
+        id: ect_profile.id,
+        type: "ParticipantProfile::Mentor",
+        user_id: ect_profile.user_id,
+        school_id: ect_profile.school_id,
+        cohort_id: ect_profile.cohort_id,
```
kicferk1, line 21: Mentor profiles can theoretically have `nil` cohorts. I don't think there are any in production that do, but to keep our sanity, can we fall back to `Cohort.current.id`?

db/migrate/20210629170845_remove_legacy_participants_profiles_f_ks.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class RemoveLegacyParticipantsProfilesFKs < ActiveRecord::Migration[6.1]
+  def change
+    remove_foreign_key :participant_declarations, :early_career_teacher_profiles
+    add_foreign_key :participant_declarations, :participant_profiles, column: :early_career_teacher_profile_id, validate: false
```
kicferk1, line 6: What does `validate: false` do?

app/forms/schools/add_participant_form.rb
ethanmills: I'm going with "no" by the failing test...

timabell:
@BroiSatse Currently the PR title "Participant profiles" doesn't tell me what it will change. I'd suggest updating it to say what change this PR will make. Perhaps "Switch to single-table-inheritance for participant profiles" if that is accurate.

I am struggling to follow the description of the PR. Could you make it clearer? I don't see an explanation of the motivation for making the change in there. I know there has been discussion in slack on the subject but I think it is good to repeat the rationale here alongside the change itself.

I think of PR titles and descriptions in a similar way to commit messages and summaries but at a higher level, so a lot of the guidance at https://timwise.co.uk/2016/03/18/yet-another-good-commit-messages-post/ applies.

db/migrate/20210629104700_populate_participant_profiles.rb
```diff
@@ -0,0 +1,40 @@
...
+        core_induction_programme_id: ect_profile.core_induction_programme_id,
+      )
+    end
+
+    EarlyCareerTeacherProfile.find_each do |ect_profile|
```
timabell, line 26: Are we likely to hit a race condition where production is still adding records to this table after you've copied them all across? It seems like this change might require downtime to ensure we don't lose data.

timabell:
I'm a bit concerned that this doesn't seem to remove the old data and tables having copied everything out of them. This seems to me like it increases the risk of mistakes. If you don't want to delete the old tables in this migration then I'd suggest at least renaming them so that they can't be easily mistaken for current data and written to be any code that is unaware of the change. Something like a rename to `_old_early_career_teacher_profiles` etc, and a follow-up task somewhere to ensure they are deleted in a future migration.

timabell:
The commits on this PR need much better descriptions and titles to be useful to future readers.

I'd suggest reading https://timwise.co.uk/2016/03/18/yet-another-good-commit-messages-post/ and then doing an interactive rebase to improve them

app/models/participant_profile.rb
BroiSatse: Not as much as I would like it to do - it removes the column from model inspection and from `attributes`. Annoyingly it does not remove getter and setter though...

app/models/school.rb
BroiSatse: The new query does not perform join which in ActiveRecord is alwyas a massive win - `joins` does not perform preloading and will make things really complicated when this is combined with `includes` and conditions on joined table. It is best to avoid `joins` whenever possible.

BroiSatse:
I'll be closing this PR for now to reopen it in two separate PRs to tackle migration race condition.

PR #607, opened 2021-06-29
CPDRP-525 managing school changes

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/browse/CPDRP-525)

> This is labelled `blocked-merge` because it's a WIP for using a GetHelpWithTech style approach for handling GIAS changes to schools and there is still thinking needed. However it would be great to get feedback on it!

What this does:
 * Takes the GIAS school and school link extracts and imports them into a "staging" area
 * Adds or updates schools in the staging area from the GIAS extract
 * Records additions and attribute changes
 * Presents (placeholder) views of schools that need to be added, closed or updated.

Addition of `DataStage::School` model as a staging point for GIAS data
Addition of `DataStage::SchoolLink` model as representation of GIAS school links
Addition of `DataStage::SchoolChange` model that records attributes changes to `DataStage::School`s (although now I'm thinking we may not need this)
Addition of `DataStage::FetchGiasDataFiles` and `DataStage::UpdateStagedSchools` for fetching and applying school data from GIAS.
Addition of some controllers for simple viewing of changes under `/admin/gias`
A bit of refactoring of `School` to extract some common stuff
Addition of `GiasPolicy` for `Pundit`

### Some thoughts after building this
We may not need the `DataStage::SchoolChange` model as it might be more appropriate just to have an indication that a school has changed and then compare that against the existing "live" school in the front-end for admin users to see?

The front-end views are only really placeholders done in a similar style to GHwT. We _could_ show a lot more info on those detail pages, especially for existing "live" schools when comparing such as cohort and programme choices etc.

### Guidance to review
Run the import process with:

```ruby
  DataStage::FetchGiasDataFiles.call { |files| DataStage::UpdateStagedSchools.call(files) }
```

### Review Checks

image
image
image

image

image
image
image

### DIFF
--- a/app/controllers/admin/gias/school_changes_controller.rb
+++ b/app/controllers/admin/gias/school_changes_controller.rb
@@ -6,10 +6,8 @@ module Admin
       before_action { authorize :gias }
 
       def index
-        @schools_to_open = schools_scope.schools_to_add.order(urn: :asc)
-        @schools_to_close = schools_scope.schools_to_close.order(urn: :asc)
-        @changed_schools = schools_scope.includes(:school_changes).where(school_changes: { status: :changed, handled: false })
-
+        @schools = schools_scope.includes(:school_changes)
+          .where(school_changes: { status: :changed, handled: false })
       end
 
       def show
@@ -17,6 +15,7 @@ module Admin
       end
 
     private
+
       def schools_scope
         policy_scope(DataStage::School, policy_scope_class: GiasPolicy::Scope)
       end
--- a/app/controllers/admin/gias/school_changes_controller.rb
+++ b/app/controllers/admin/gias/school_changes_controller.rb
@@ -6,8 +6,7 @@ module Admin
       before_action { authorize :gias }
 
       def index
-        @schools = schools_scope.includes(:school_changes)
-          .where(school_changes: { status: :changed, handled: false })
+        @schools = schools_scope.schools_with_changes.order(urn: :asc)
       end
 
       def show
--- a/app/controllers/admin/gias/school_changes_controller.rb
+++ b/app/controllers/admin/gias/school_changes_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Admin
+  module Gias
+    class SchoolChangesController < Admin::BaseController
+      before_action { authorize :gias }
+
+      def index
+        @schools_to_open = schools_scope.schools_to_add.order(urn: :asc)
+        @schools_to_close = schools_scope.schools_to_close.order(urn: :asc)
+        @changed_schools = schools_scope.includes(:school_changes).where(school_changes: { status: :changed, handled: false })
+
+      end
+
+      def show
+        @school = schools_scope.find_by(urn: params[:id])
+      end
+
+    private
+      def schools_scope
+        policy_scope(DataStage::School, policy_scope_class: GiasPolicy::Scope)
+      end
+    end
+  end
+end
--- a/app/models/concerns/gias_helpers.rb
+++ b/app/models/concerns/gias_helpers.rb
@@ -18,6 +18,10 @@ module GiasHelpers
     }, _suffix: "status"
   end
 
+  def name_and_urn
+    "#{name} (#{urn})"
+  end
+
 private
 
   def open?
--- a/app/models/concerns/gias_helpers.rb
+++ b/app/models/concerns/gias_helpers.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module GiasHelpers
+  extend ActiveSupport::Concern
+  include GiasTypes
+
+  included do
+    scope :currently_open, -> { where(school_status_code: GiasTypes::ELIGIBLE_STATUS_CODES) }
+    scope :eligible_establishment_type, -> { where(school_type_code: GiasTypes::ELIGIBLE_TYPE_CODES) }
+    scope :in_england, -> { where("administrative_district_code ILIKE 'E%'") }
+    scope :eligible, -> { currently_open.eligible_establishment_type.in_england }
+
+    enum school_status_name: {
+      open: "Open",
+      closed: "Closed",
+      proposed_to_close: "Open, but proposed to close",
+      proposed_to_open: "Proposed to open",
+    }, _suffix: "status"
+  end
+
+private
+
+  def open?
+    open_status_code?(school_status_code)
+  end
+
+  def eligible_establishment_type?
+    eligible_establishment_code?(school_type_code)
+  end
+
+  def in_england?
+    english_district_code?(administrative_district_code)
+  end
+end
--- a/app/models/data_stage/school.rb
+++ b/app/models/data_stage/school.rb
@@ -21,9 +21,11 @@ module DataStage
     after_create :log_create
     before_update :log_changes
 
-    scope :schools_to_add, -> { currently_open.joins("LEFT JOIN schools s ON (data_stage_schools.urn = s.urn)").where("s.urn IS NULL") }
+    scope :schools_to_add, -> { currently_open.left_joins(:counterpart).where(counterpart: { urn: nil }) }
 
-    scope :schools_to_close, -> { closed_status.joins("LEFT JOIN schools s ON (data_stage_schools.urn = s.urn)").where("s.school_status_code IN (?)", ELIGIBLE_STATUS_CODES) }
+    scope :schools_to_close, -> { closed_status.left_joins(:counterpart).where(counterpart: { school_status_code: GiasTypes::ELIGIBLE_STATUS_CODES }) }
+
+    scope :schools_with_changes, -> { includes(:school_changes).where(school_changes: { status: :changed, handled: false }) }
 
   private
 
--- a/app/models/data_stage/school.rb
+++ b/app/models/data_stage/school.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module DataStage
+  class School < ApplicationRecord
+    include GiasHelpers
+
+    self.table_name = "data_stage_schools"
+
+    has_many :school_changes, class_name: "DataStage::SchoolChange",
+                              foreign_key: :data_stage_school_id,
+                              dependent: :destroy
+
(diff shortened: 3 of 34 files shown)


app/controllers/admin/gias/school_changes_controller.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Admin
+  module Gias
+    class SchoolChangesController < Admin::BaseController
+      before_action { authorize :gias }
+
+      def index
+        @schools = schools_scope.includes(:school_changes)
+          .where(school_changes: { status: :changed, handled: false })
```
kicferk1, line 10: same about the scope

app/services/data_stage/fetch_gias_data_files.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require "gias_api_client"
+require "csv"
+
+module DataStage
+  class FetchGiasDataFiles < ::BaseService
```
kicferk1, line 7: Is this used anywhere?

app/services/data_stage/update_staged_schools.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    def call
+      update_schools
+      update_school_links
+    end
+
+  private
+
+    def update_schools
+      if @school_data_file
```
kicferk1, line 20: When does it not exist? Should we log that it happened (and maybe use early returns to reduce nesting)?

app/services/data_stage/update_staged_schools.rb
```diff
@@ -0,0 +1,96 @@
...
+        end
+      end
+    end
+
+    def eligible_row?(row)
+      english_district_code?(row.fetch("DistrictAdministrative (code)")) &&
+        eligible_establishment_code?(row.fetch("TypeOfEstablishment (code)").to_i)
+    end
+
+    def filtered_attributes(data)
```
kicferk1, line 65: Feels like this looks very similar to the school importer, and for good reason too, they do the same job. Is this going to essentially replace it? If not, should we commonise some things between them?

app/views/admin/gias/home/index.html.erb
```diff
@@ -0,0 +1,30 @@
+<% content_for :title, "GIAS changes to schools" %>
+
+<h1 class="govuk-heading-xl">GIAS changes to schools</h1>
+
+<h2 class="govuk-heading-l govuk-!-font-size-27">
```
kicferk1, line 5: I don't think we want to set the font size explicitly :)

app/views/admin/gias/school_changes/show.html.erb
```diff
@@ -0,0 +1,28 @@
...
+<table class="govuk-table">
+  <thead class="govuk-table__head">
+    <tr class="govuk-table__row">
+      <th scope="col" class="govuk-table__header">Attribute</th>
+      <th scope="col" class="govuk-table__header">Current value</th>
+      <th scope="col" class="govuk-table__header">New value</th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+    <% @school.attributes.except("id", "created_at", "updated_at", "urn").each do |k,v| %>
```
kicferk1, line 15: Can we use `key` and `value` instead of `k` and `v`?

kicferk1 (commented):
Just double checking, this doesn't add the ability for the user to actually accept the changes, right? Just to show the differences?

Looks pretty good overall :)

app/services/data_stage/update_staged_schools.rb
tonyheadford: So, currently this is a refactored version of the school importer. I split fetching the files out from the actual updating to simplify things, plus I don't want to change the existing one currently so that we can still update schools directly, until this makes it out properly.  Eventually this should replace the importer as it will no longer be needed.

app/views/admin/gias/home/index.html.erb
tonyheadford: The views are only placeholders, this was lifted from GHwT where the designer wanted it like that!

app/services/data_stage/update_staged_schools.rb
tonyheadford: Having this in means it would be possible to only update schools or school links if you wanted to (eg dev or testing or manual support).

app/services/data_stage/fetch_gias_data_files.rb
tonyheadford: Yes, it fetches the GIAS files. At the moment there is no job set up to do this automatically but in the Guidance to review bit above I have an example of how it is used

app/models/data_stage/school.rb
```diff
@@ -0,0 +1,40 @@
...
+                            dependent: :destroy
+
+    has_one :counterpart, class_name: "::School",
+                          foreign_key: :urn,
+                          primary_key: :urn
+
+    after_create :log_create
+    before_update :log_changes
+
+    scope :schools_to_add, -> { currently_open.joins("LEFT JOIN schools s ON (data_stage_schools.urn = s.urn)").where("s.urn IS NULL") }
```
ethanmills, line 24: This is the same, right?

```
2.7.2 :007 > DataStage::School.currently_open.left_outer_joins(:counterpart).where(counterpart: {urn: nil})
  DataStage::School Load (0.8ms)  SELECT "data_stage_schools".* FROM "data_stage_schools" LEFT OUTER JOIN "schools" "counterpart" ON "counterpart"."urn" = "data_stage_schools"."urn" WHERE "data_stage_schools"."school_status_code" IN ($1, $2) AND "counterpart"."urn" IS NULL /* loading for inspect */ LIMIT $3  [[nil, 1], [nil, 3], ["LIMIT", 11]]
```

app/models/data_stage/school.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    scope :schools_to_close, -> { closed_status.joins("LEFT JOIN schools s ON (data_stage_schools.urn = s.urn)").where("s.school_status_code IN (?)", ELIGIBLE_STATUS_CODES) }
```
ethanmills, line 26: Ditto

```
2.7.2 :009 > DataStage::School.closed_status.left_outer_joins(:counterpart).where(counterpart: {school_status_code: GiasTypes::ELIGIBLE_STATUS_CODES})
  DataStage::School Load (1.2ms)  SELECT "data_stage_schools".* FROM "data_stage_schools" LEFT OUTER JOIN "schools" "counterpart" ON "counterpart"."urn" = "data_stage_schools"."urn" WHERE "data_stage_schools"."school_status_name" = $1 AND "counterpart"."school_status_code" IN ($2, $3) /* loading for inspect */ LIMIT $4  [["school_status_name", "Closed"], [nil, 1], [nil, 3], ["LIMIT", 11]]
```

app/models/concerns/gias_helpers.rb
```diff
@@ -0,0 +1,38 @@
...
+    scope :eligible_establishment_type, -> { where(school_type_code: GiasTypes::ELIGIBLE_TYPE_CODES) }
+    scope :in_england, -> { where("administrative_district_code ILIKE 'E%'") }
+    scope :eligible, -> { currently_open.eligible_establishment_type.in_england }
+
+    enum school_status_name: {
+      open: "Open",
+      closed: "Closed",
+      proposed_to_close: "Open, but proposed to close",
+      proposed_to_open: "Proposed to open",
+    }, _suffix: "status"
```
ethanmills, line 18: I find it surprising that we do open on code and closed on name

app/views/admin/gias/school_changes/show.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      <% current_value = @school.counterpart.send(k) %>
+      <% if current_value != v %>
+        <tr class="govuk-table__row">
+          <th scope="row" class="govuk-table__header"><%= k.humanize %></th>
+          <td class="govuk-table__cell"><%= current_value %></td>
+          <td class="govuk-table__cell">
+              <strong><i><%= v %></i></strong>
```
callumacrae, line 22: `i` doesn't really mean italic anymore, this would probably be best off replaced with `<em>` or a classname (preferably a classname)

app/models/concerns/gias_helpers.rb
tonyheadford: I did think we should look at doing away with the code attribute and simplify this to just an enum with open or closed as that's all we're really interested in (i.e. `open` and `proposed_to_close` become simply `open` and the other two become `closed`) and we could handle that during import to the data stage?

app/views/admin/gias/school_changes/show.html.erb
tonyheadford: Yes, good catch, you're right , I couldn't remember `em` last night when I added that!

app/models/data_stage/school.rb
tonyheadford: Nice!

app/models/concerns/gias_helpers.rb
ethanmills: I used code because I didn't trust the format of the name field. If you didn't have any issues with it on ghwt then I don't see why we shouldn't move to just name 👍

app/views/admin/gias/home/index.html.erb
tonyheadford: Removed.

kicferk1:
Soo, why does it have a blocked-merge label? What is it missing / what is uncertain about it?

kicferk1:
> Soo, why does it have a blocked-merge label? What is it missing / what is uncertain about it?

Oh, I just can't read the description 😂 

> This is labelled blocked-merge because it's a WIP for using a GetHelpWithTech style approach for handling GIAS changes to schools and there is still thinking needed. However it would be great to get feedback on it!

Soo, can handling changes be a separate PR? As far as I can tell, this adds value - cause it gives the admin users the ability to see the scale of the problem :)

ethanmills (approved):
LGTM. Is there any risk in putting this in production and running a one-off? Be good to get some user feedback

PR #608, opened 2021-06-30
CPDTP-224 Refactor MentorProfile creation form and add uplift

blackrat:
### Context

As part of the uplift piece for payment, the sparsity uplift and pupil premium uplift flags were added to the `EarlyCareerTeacherProfile`. This adds in the corresponding flags to the `MentorProfile` and carries on the form
refactoring that was done in the earlier piece.

- Add in the new flags and flag creation to the Form
- Extract MentorProfile Creation into a separate Mentors::Create service and remove duplication from this and EarlyCareerTeachers::Create by pulling the common School/Cohort code into a shared module.
- Add in tests for MentorProfile creation in line with those added for EarlyCareerTeacherProfile creation

### Guidance to review

Functional changes are only that the uplift flags will be populated and are backfilled.
Everything else is a refactor.

### Testing

New tests added to test the population of the MentorProfile table with the appropriate flags set.

### Review Checks

Should have no noticeable change, but the MentorProfile records in the database should be updated with the appropriate sparsity and pupil premium uplift flags.

### DIFF
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -108,26 +108,19 @@ module Schools
     end
 
     def save!
-      if type == :ect
-        EarlyCareerTeachers::Create.call(
-          full_name: full_name,
-          email: email,
-          cohort_id: school_cohort.cohort_id,
-          school_id: school_cohort.school_id,
-          mentor_profile_id: mentor&.mentor_profile&.id,
-        )
-      else
-        ActiveRecord::Base.transaction do
-          # TODO: What if email matches but with different name?
-          user = User.find_or_create_by!(full_name: full_name, email: email)
-
-          MentorProfile.create!(
-            user: user,
-            school_id: school_cohort.school_id,
-            cohort_id: school_cohort.cohort_id,
-          )
-        end
-      end
+      creator = case type
+                when :ect
+                  EarlyCareerTeachers::Create
+                when :mentor
+                  Mentors::Create
+                end
+      creator.constantize.call(
+        full_name: full_name,
+        email: email,
+        cohort_id: school_cohort.cohort_id,
+        school_id: school_cohort.school_id,
+        mentor_profile_id: mentor&.mentor_profile&.id,
+      )
     end
   end
 end
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -115,7 +115,7 @@ module Schools
     end
 
     def save!
-      creators[participant_type]&.call(
+      creators[participant_type].call(
         full_name: full_name,
         email: email,
         cohort_id: school_cohort.cohort_id,
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -107,14 +107,15 @@ module Schools
       @current_user ||= User.find_by(id: current_user_id)
     end
 
+    def creators
+      {
+        ect: EarlyCareerTeachers::Create,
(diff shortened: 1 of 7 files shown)


ethanmills:
This is going to conflict with #606

app/forms/schools/add_participant_form.rb
```diff
@@ -108,26 +108,19 @@ def current_user
...
-          )
-        end
-      end
+      creator = case type
+                when :ect
+                  EarlyCareerTeachers::Create
+                when :mentor
+                  Mentors::Create
+                end
+      creator.constantize.call(
```
kicferk1, line 117: Why do we need `constantize`? Internet tells me that it is a method on a string, that lets you find the class from that string, and use it as a class. 

But I don't think `creator` is a string, I think it is a class. So we could (I hope) just use `creator.call`. Am I thinking right?

blackrat: Changed.

blackrat:
> This is going to conflict with #606

Hopefully not a huge conflict. I think only the names within the two services should have to change to reflect the new namespace for the base models.

kicferk1 (approved):
Please talk to @BroiSatse before merging it, and maybe pair on resolving the conflicts :) 

In an ideal world the more complex thing (https://github.com/DFE-Digital/early-careers-framework/pull/606) would go in first, but I think it is also a bit less ready :/ So the second best thing is to make the merge conflicts as painless as possible.

blackrat:
Talked to Stan about possible conflicts with #606 and we're going to do a combined refactor after the ProfileDeclarations go in.

app/forms/schools/add_participant_form.rb
```diff
@@ -107,27 +107,21 @@ def current_user
...
-          user = User.find_or_create_by!(full_name: full_name, email: email)
-
-          MentorProfile.create!(
-            user: user,
-            school_id: school_cohort.school_id,
-            cohort_id: school_cohort.cohort_id,
-          )
-        end
-      end
+      creators[type]&.call(
```
ethanmills, line 118: I think this should be `creators[participant_type]`.  I suspect the error you were having was caused by `type = :self`

ethanmills (approved):

app/forms/schools/add_participant_form.rb
blackrat: Yes. I suspect that you are correct. Highlights that there mustn't have been a unit test for that scenario.

app/services/mentors/create.rb
```diff
@@ -0,0 +1,34 @@
...
+    def mentor_attributes
+      {
+        school_id: school_id,
+        cohort_id: cohort_id,
+        sparsity_uplift: sparsity_uplift?(start_year),
+        pupil_premium_uplift: pupil_premium_uplift?(start_year),
+      }
+    end
+  end
+end
```
jamesrose, line 34: Do these need to be separate services? Seems to me you could achieve the same thing with something like this:

```ruby
include SchoolCohortDelegator

def call(type, attrs)
    user = User.find_or_create_by!(full_name: full_name, email: email)
    attrs.merge!({
      user: user,
      sparsity_uplift: sparsity_uplift?(start_year),
      pupil_premium_uplift: pupil_premium_uplift?(start_year),
    })
    type.create!(attrs.compact)
  end
end
```

blackrat: Don't want to do that yet, since the `mentor_id` for one, but not the other, was part of the wider discussion about the Participant refactoring that is currently ongoing.

PR #610, opened 2021-06-30
Stable API tokens in seeds for development

timabell:

* Stable API tokens in seeds for development
* Remove output from sandbox seeds file, the providers already have their sandbox tokens so this is no longer needed.

### Context

Currently working on https://dfedigital.atlassian.net/browse/CPDTP-232 "Seed API sandbox environment with mentor data"

I wanted to remove the token generation from the sandbox seeds as that has already served its one-time purpose of making tokens for providers to test with and it was getting in the way of the current task.

We have been using that output for local development testing, so in order to continue to facilitate that use of the tokens I've moved it to a more permanent home in the initial_seeds file and made them stable for convenience.

### Guidance to review

Are we happy with well known tokens in the code? I *think* it's a good trade-off of convenience here as it will save a lot of passing around tokens.

### Testing

`bundle exec db:reset`

Try one of the new tokens locally.

### DIFF
--- a/db/seeds/initial_seed.rb
+++ b/db/seeds/initial_seed.rb
@@ -7,29 +7,19 @@ edt_cip = CoreInductionProgramme.find_or_create_by!(name: "Education Development
 teach_first_cip = CoreInductionProgramme.find_or_create_by!(name: "Teach First")
 ucl_cip = CoreInductionProgramme.find_or_create_by!(name: "UCL Institute of Education")
 
-ambition = LeadProvider.find_or_create_by!(name: "Ambition Institute")
-ambition.update!(cohorts: [cohort_2021]) unless ambition.cohorts.any?
-LeadProviderCip.find_or_create_by!(lead_provider: ambition, cohort: cohort_2021, core_induction_programme: ambition_cip)
-
-bpn = LeadProvider.find_or_create_by!(name: "Best Practice Network")
-bpn.update!(cohorts: [cohort_2021]) unless bpn.cohorts.any?
-LeadProviderCip.find_or_create_by!(lead_provider: bpn, cohort: cohort_2021, core_induction_programme: ucl_cip)
-
-capita = LeadProvider.find_or_create_by!(name: "Capita")
-capita.update!(cohorts: [cohort_2021]) unless capita.cohorts.any?
-LeadProviderCip.find_or_create_by!(lead_provider: capita, cohort: cohort_2021, core_induction_programme: ambition_cip)
-
-edt = LeadProvider.find_or_create_by!(name: "Education Development Trust")
-edt.update!(cohorts: [cohort_2021]) unless edt.cohorts.any?
-LeadProviderCip.find_or_create_by!(lead_provider: edt, cohort: cohort_2021, core_induction_programme: edt_cip)
-
-teach_first = LeadProvider.find_or_create_by!(name: "Teach First")
-teach_first.update!(cohorts: [cohort_2021]) unless teach_first.cohorts.any?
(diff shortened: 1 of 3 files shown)


db/seeds/initial_seed.rb
```diff
@@ -59,3 +59,12 @@
...
   NpqCourse.find_or_create_by!(name: hash[:name], id: hash[:id])
 end
+
+if Rails.env.development?
+  LeadProviderApiToken.create_with_known_token!("_488ase4rzqrLo6p3Kzy", lead_provider: ambition)
+  LeadProviderApiToken.create_with_known_token!("KCC_Frtiy79M3zLNMhic", lead_provider: bpn)
+  LeadProviderApiToken.create_with_known_token!("bBvWZ5dr6wmbUNqidnY-", lead_provider: capita)
+  LeadProviderApiToken.create_with_known_token!("kDyukqzjhuPw7j3kgLT2", lead_provider: edt)
+  LeadProviderApiToken.create_with_known_token!("usQncFvBGym5UoGcbsMs", lead_provider: teach_first)
+  LeadProviderApiToken.create_with_known_token!("RCxkriTrzw4yhnn8zjyy", lead_provider: ucl)
```
ethanmills, line 69: Do they even need to be in this format? For dev could they not just be e.g. `ambition-token`?

ethanmills (approved):

db/seeds/initial_seed.rb
timabell: I guess so, I just kept them as per the generated ones for consistency

db/seeds/sandbox_data.rb
```diff
@@ -14,7 +14,6 @@
 providers_names = ["Capita", "Teach First", "UCL", "Best Practice Network", "Ambition", "Education Development Trust"]
 
 def generate_provider_token(lead_provider, school, cohort, logger)
```
timabell, line 16: Oops, method name doesn't make sense any more...

db/seeds/initial_seed.rb
timabell: Done, thanks. Also did some refactoring to make it more DRY and make the values that vary stand out.

db/seeds/initial_seed.rb
```diff
@@ -7,29 +7,19 @@
...
+  { provider_name: "Capita", cip: ambition_cip, token: "capita-token" },
+  { provider_name: "Education Development Trust", cip: edt_cip, token: "edt-token" },
+  { provider_name: "Teach First", cip: teach_first_cip, token: "teach-first-token" },
+  { provider_name: "UCL Institute of Education", cip: ucl_cip, token: "ucl-token" },
+].each do |seed|
+  provider = LeadProvider.find_or_create_by!(name: seed[:provider_name])
+  provider.update!(cohorts: [cohort_2021]) unless provider.cohorts.any?
+  LeadProviderCip.find_or_create_by!(lead_provider: provider, cohort: cohort_2021, core_induction_programme: seed[:cip])
+  LeadProviderApiToken.create_with_known_token!(seed[:token], lead_provider: provider)
+end
```
kicferk1, line 22: Soo, making this stuff public is not too bad. But I don't want this in production. Can we hide it behind environment check? Or perhaps move token generation to a seed file that doesn't get run in production?

I know we have already ran the seed there, but, well, better safe than sorry.

timabell: Damn I forgot to move the environment guard across when I moved it in the refactor commit, good spot @kicferk1 , thank you!

kicferk1 (approved):

db/seeds/initial_seed.rb
timabell: Done. :sweat_smile:

PR #611, opened 2021-06-30
CPDTP-224 Prepare for Profile Declarations switchover

blackrat:
### Context

This is the new structure to hold information about the
profiles that the ParticipantDeclarations will refer to.
The EarlyCareerTeacherProfile association will move
to being a `declarable` using delegated types.
This form of multi-table inheritance was introduced in
Rails 6.1: https://api.rubyonrails.org/classes/ActiveRecord/DelegatedType.html

- Create migrations for the 3 new tables, profile_declarations, early_career_teacher_profile_declarations, and mentor_profile_declarations.
- These tables will be used to join the participant declarations to the appropriate profiles.
- For now, only the structure will be in place. Later commits will add in the functionality to migrate to these and use them.

### Guidance to review

Structural changes to capture profile declaration differences.

### Review Checks


app/models/profile_declaration.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class ProfileDeclaration < ApplicationRecord
+  delegated_type :declarable, types: %w[EarlyCareerTeacherProfileDeclaration MentorProfileDeclaration]
+  belongs_to :participant_declaration
+  belongs_to :lead_provider
+
+  scope :ect_profiles, lambda {
```
ethanmills, line 8: Are there any specs for this?

plisovin (approved):

app/models/profile_declaration.rb
blackrat: Not in this PR. Pavel was writing them and I was going to leave them out and add them in, but they're coming.

app/models/profile_declaration.rb
```diff
@@ -0,0 +1,28 @@
...
+                       }
+  scope :mentor_profiles, lambda {
+                            joins(
+                              "left join mentor_profile_declarations
+  on declarable_type='MentorProfileDeclaration'
+  and declarable_id=mentor_profile_declarations.id
+  left join mentor_profiles on
+  mentor_profile_declarations.mentor_profile_id = mentor_profiles.id",
+                            )
+                          }
```
ethanmills, line 25: Are there specs for any of this?

blackrat: There are. In the following PR, but I could pull them from here if you'd prefer them to go in together.

ethanmills:
@blackrat sorry, deleted and readded my comment to add context

> Not in this PR. Pavel was writing them for the scopes, but they're coming in the next PR or the one after when we start to migrate the data in. I could remove the 3 scopes for now?

I don't feel I can review these scopes without them, they're sufficiently complex that I'm not 100% that I understand the intention.

ethanmills:
I think I've missed some conversations here. Why is this three tables, not just one with a polymorphic association?

jamesrose:
Echoing @ethanmills. It's not clear to me that why we need three tables here.

app/models/profile_declaration.rb
jamesrose, line 8: This is a bit messy, but that might be partly because of the formatting. Can you do it in Arel?

app/models/profile_declaration.rb
blackrat: Removed from this PR

blackrat:
> I think I've missed some conversations here. Why is this three tables, not just one with a polymorphic association? Is a declaration for an ECT fundamentally different for one for a mentor?

That was the original suggestion on this, and it's probably going to collapse down to one again when @BroiSatse has a complete refactor and we link off ParticipantProfile directly. Short answer was that @timabell there were some concerns voiced about polymorphic associations linking directly to profiles and this gives them a proxy.

timabell:
> I think I've missed some conversations here. Why is this three tables, not just one with a polymorphic association? Is a declaration for an ECT fundamentally different for one for a mentor?

If I understand what you mean by "polymorphic association" correctly (a single field in the declaration table that references primary keys in multiple other tables), then I have an issue with the inability to apply foreign key constraints at the database layer. In my experience this inevitably leads to bad data over time.

This seems to be quite a good article on the tradeoffs: https://hashrocket.com/blog/posts/modeling-polymorphic-associations-in-a-relational-database

A more specific piece on the problems with polymorphic foreign keys: http://duhallowgreygeek.com/polymorphic-association-bad-sql-smell/

---

Re using a single tables:

Having two nullable fields in one declaration table with proper foreign keys to ect/mentor would be less smelly at the database layer than a single polymorphic foreign key with no constraints.

I'm not sure what the tradeoff is here between a single table with multiple nullable foreign key fields versus how it's currently in the patch with multiple tables (akin to multiple table inheritance). Both are good database modelling depending on whether the two types of declaration are really the same of different things in the real world.

db/migrate/20210630000200_create_profile_declarations.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateProfileDeclarations < ActiveRecord::Migration[6.1]
+  def change
+    create_table :profile_declarations, id: :uuid do |t|
+      t.references :participant_declaration, null: false, index: true
+      t.references :lead_provider, null: false, index: true
+      t.references :declarable, polymorphic: true, index: true
```
timabell, line 8: No likey because we can't have database foreign key constraints

https://stackoverflow.com/questions/922184/why-can-you-not-have-a-foreign-key-in-a-polymorphic-association/922341#922341

I wasn't aware that the use of `delegated_type` would result in this :disappointed:

timabell:
#606 collapses the profiles into single-table-inheritance (STI) so the extra complexity in this PR to link to multiple profile tables would then become redundant

kicferk1:
> #606 collapses the profiles into single-table-inheritance (STI) so the extra complexity in this PR to link to multiple profile tables would then become redundant

In this case, it's a shame that the work has been done, but I would much prefer to wait for #606 to be merged first.

blackrat:
> > #606 collapses the profiles into single-table-inheritance (STI) so the extra complexity in this PR to link to multiple profile tables would then become redundant
> 
> In this case, it's a shame that the work has been done, but I would much prefer to wait for #606 to be merged first.

To be fair, quite a bit of this would still need to have been done. Separating out the declarations as they are sent and storing them without any changes from the participant uuid to the profile was something that is a better fit in any case. Having the "Profile declarations" as a point of contact with the profiles, will still be a thing, it's just that it'll collapse into a single table rather than a master table with extensions to allow for two other table types to be referenced. Dropping two migrations and backfilling by picking up the new "participant_profile_id' is potentially as simple as transition either way.

app/models/early_career_teacher_profile_declaration.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class EarlyCareerTeacherProfileDeclaration < ApplicationRecord
+  belongs_to :early_career_teacher_profile
+  include Declarable
+end
```
kicferk1, line 6: What is the job of this class? As I see it, if we inlined `Declarable`, we would get

```
class EarlyCareerTeacherProfileDeclaration < ApplicationRecord
  belongs_to :early_career_teacher_profile
  has_one :profile_declaration, as: :declarable, touch: true
end
```

It looks a lot like a middle man without any other function than being a middle man. I guess it keeps the time of when it was created, but so does `ProfileDeclaration`. 

Why would we want to do it this way, rather than by linking directly to the relevant profiles?

blackrat: It is exactly that. A proxy for the profile. Part of the reason was to not have polymorphic foreign keys to different profiles types directly. This mirrors the current profile layout and will continue to do so later. If, and when, the profiles collapse into the `participant_profiles`, the `participant_profile_declarations` will follow suit. (Especially if npq_profiles also becomes linked in that way.)

app/models/profile_declaration.rb
blackrat: Scope has been pulled out of this PR.

blackrat:
Closed since the migrations have been modified and we need to regenerate the review app.

db/migrate/20210630000200_create_profile_declarations.rb
blackrat: Yes. It seems to just put some formalisation around referencing polymorphic connections.

PR #614, opened 2021-07-01
Fix turnip step name

timabell:
### Context

We were getting a warning that this test is broken:

> No such step: 'there are 200 pupil premium participants who have started that are eligible for the uplift payment'
> ./spec/lib/payment_calculator/turnip_features/ecf/uplift_payment.feature:7

We previously tried making missing steps into failures but the error message is unhelpful when configured that way. See https://github.com/jnicklas/turnip/pull/240

* Fix turnip step name

### DIFF
--- a/spec/lib/payment_calculator/turnip_steps/ecf/uplift_payment_steps.rb
+++ b/spec/lib/payment_calculator/turnip_steps/ecf/uplift_payment_steps.rb
@@ -13,7 +13,7 @@ module UpliftPaymentsSteps
     @uplift_eligible_participants = value.to_i
   end
 
-  step "And there are :value pupil premium participants who have started that are eligible for the uplift payment" do |value|
+  step "there are :value pupil premium participants who have started that are eligible for the uplift payment" do |value|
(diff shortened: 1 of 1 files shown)


kicferk1 (commented):
Just so I understand, was the test passing? That sounds a bit surprising :/ 

Unhelpful message when test fails still sounds better than a test passing and not testing the stuff you want, right? 

Also, is this step used anywhere? Do any outputs need changing because of changed name?

timabell:
> Just so I understand, was the test passing? That sounds a bit surprising :/
> 

It was spitting out a warning and skipping the test :cry: 

> Unhelpful message when test fails still sounds better than a test passing and not testing the stuff you want, right?
> 

Possibly, though the alternative would make development pretty painful without my proposed patch to turnip so we balked at turning that setting on.

> Also, is this step used anywhere? Do any outputs need changing because of changed name?

It is used in one test, and that test now passes instead of being skipped.

P.S. there is talk of dropping the turnip tests entirely in favour of unit tests so this may all become moot soon.

kicferk1 (approved):

PR #615, opened 2021-07-01
CPDRP-548: Induction tutors change participant details

ethanmills:
### Context
CPDRP-548
Let induction tutors edit name and email address of participants.
Blocked on content design

### Guidance to review
Ideas for how to make the whole email validation thing less messy appreciated.
Will conflict with Stan's work

### Review Checks

Sign in as school-leader@example.com and edit some details

### DIFF
--- a/app/controllers/schools/participants_controller.rb
+++ b/app/controllers/schools/participants_controller.rb
@@ -2,7 +2,7 @@
 
 class Schools::ParticipantsController < Schools::BaseController
   before_action :set_school_cohort
-  before_action :set_participant, except: :index
+  before_action :set_participant, except: %i[index email_used]
   before_action :check_feature_flag
 
   def index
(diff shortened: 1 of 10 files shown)


ethanmills:
This is blocked on content design, specifically for the error pages. I've asked if we really need two different error pages here, waiting on content design for that too

app/forms/schools/update_participant_name_form.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Schools
+  class UpdateParticipantNameForm
+    include ActiveModel::Model
+
+    attr_accessor :full_name
+    validates :full_name, presence: true
+  end
```
kicferk1, line 9: This feels like it's barely doing anything.

Do we want to move towards a pattern of having a form for every data input in our app?

kicferk1 (commented):
Soo, when we edit users as admins we don't bother having separate edit pages for their name and for their email, right?

I imagine it came from our designers, but I would be interested to know the reasoning behind having separate pages for them.

app/controllers/schools/participants_controller.rb
```diff
@@ -19,6 +19,31 @@ def show
...
 
+  def edit_name; end
+
+  def update_name
+    @participant.assign_attributes(params.require(:user).permit(:full_name))
+    render "schools/participants/edit_name" and return unless @participant.valid?
+
+    @participant.save!
+    set_success_message(heading: "The participant's name has been updated")
+    redirect_to schools_participant_path(id: @participant)
```
kicferk1, line 30:
```
if participant.update(params.require(:user).permit(:full_name))
  set_success_message(heading: "The participant's name has been updated")
  redirect_to schools_participant_path(id: @participant)
else
  render "schools/participants/edit_name"
end
```
seems to be a more 'railsy' way of doing this, I will let you choose whether you like it more or not :)

kicferk1 (approved):
It looks good to me, maybe the flow could use some more classic Rails flow.

I imagine the error pages will be pretty standard, but I will try to have another look when you push them :)

PR #616, opened 2021-07-01
CDPTP-224 Prepare database to link participant_declarations to profiles using polymorphic associations

blackrat:
### Context

This PR changes adds in the table and model to allow us to have a polymorphic link between the ParticipantDeclarations
(which will be set up to use the user_id as opposed to the profiles in a later PR), and the required Profiles (either ECT or Mentor), by creating a join table which points to the participant_declaration on one side and the profile on the other.

This is based on the multi-table inheritance introduced in
Rails 6.1: https://api.rubyonrails.org/classes/ActiveRecord/DelegatedType.html

It is also based on PR #611, but simplifies the structure by eliminating the intermediate tables.

ProfileDeclaration table and model and the Declarable module (as per the DelegatedType guidelines)
Note: There is a migration required, which will occur after this base structure is in place in the next PR, then
a final PR will switch to the new structure.

### Guidance to review

Bear in mind that this is an interim step, but it still should not break anything current. It's basically putting scaffolding in place to allow for the next set of migrations and the population of new structure to take place for new declarations coming in.

### Review Checks


spec/factories/declarable.rb
```diff
@@ -0,0 +1,12 @@
...
+FactoryBot.define do
+  factory :declarable do
+    trait :early_career_teacher_profile_declaration do
+      early_career_teacher_profile_declaration
+    end
+    trait :mentor_profile_declaration do
+      mentor_profile_declaration
+    end
+  end
+end
```
kicferk1, line 12: Is this useful anymore?

blackrat: No. Probably not. Should be linking directly to the profiles.

app/models/profile_declaration.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class ProfileDeclaration < ApplicationRecord
+  delegated_type :declarable, types: %w[EarlyCareerTeacherProfile MentorProfile]
+  belongs_to :participant_declaration
+  belongs_to :lead_provider
```
ethanmills, line 6: why does this need a lead provider if participant declaration has one too?

kicferk1 (approved):
This looks good to me. `declarable` factory seems unused, but I think it will be used soon enough :)

app/models/profile_declaration.rb
blackrat: I had a conversation with the PM about NPQs where the association with lead provider is different and declarations can come in from multiple lead providers for the same participant. I saw this as becoming a composite primary key, but I'm probably thinking too far ahead.

PR #619, opened 2021-07-01
Created shadow ParticipantProfile model

BroiSatse:
This PR is the first of series of PRs for refactoring current `EarlyCareerTeacherProfile` and `MentorProfile` into a single STI `ParticipantProfile` model.

### Context

At the moment, everywhere in the code `@participant` is always an instance of `User` and `Participant` simply does not exist. We have ParticipantPolicy, which takes an instance of user, we have ParticipantsController (in school and admin namespaces) which both loads Users using participant-related scopes on User class. We have a lot of code like `@participant.is_mentor? ? @participant.mentor_profile : @participant.early_career_profile` and we're passing an instance of user from method to method and searching for the relevant profiles. This feels like a missed Polymorphism opportunity which we'll bring back here.

### Further steps

This PR simply creates a `ParticipantProfile` model which is to shadow current `EarlyCareerTeacherProfile` and `MentorProfile`. This new model is not being used or referenced anywhere else in the code... for now. Further steps.

2. Once this is deployed, we populate missing records. At this point, shadow ParticipantProfile and legacy profile tables will always be in sync.
3. Once table is populated, we can switch all the profile references to use new `ParticipantProfile` instead of legacy model. At this stage we will preserve as much of original interface as possible. We would also switch the legacy profiles to shadow the new participant profiles.
4. Clean-up: removing unnecessary scopes and simplifying the code wherever we can - the legacy profiles can be now safely removed after validation the data is consistend.

### DIFF
--- a/db/migrate/20210629103240_create_participant_profiles.rb
+++ b/db/migrate/20210629103240_create_participant_profiles.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class CreateParticipantProfiles < ActiveRecord::Migration[6.1]
+  def change
+    create_table :participant_profiles do |t|
+      t.string :type, null: false
+
+      t.references :user, foreign_key: true, index: true, null: false
+      t.references :school, foreign_key: true, index: true, null: false
+      t.references :core_induction_programme, foreign_key: true, index: true, null: true
+      t.references :cohort, foreign_key: true, index: true, null: false
+
+      # No foregin key yet - not all records will be present
+      t.references :mentor_profile, foreign_key: false, index: true, null: true
+
+      t.boolean :sparsity_uplift, default: false, null: false
+      t.boolean :pupil_premium_uplift, default: false, null: false
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


db/migrate/20210629103240_create_participant_profiles.rb
```diff
@@ -0,0 +1,22 @@
...
+    create_table :participant_profiles do |t|
+      t.string :type, null: false
+
+      t.references :user, foreign_key: true, index: true, null: false
+      t.references :school, foreign_key: true, index: true, null: false
+      t.references :core_induction_programme, foreign_key: true, index: true, null: true
+      t.references :cohort, foreign_key: true, index: true, null: false
+
+      # No foregin key yet - not all records will be present
+      t.references :mentor_profile, foreign_key: false, index: true, null: true
```
kicferk1, line 14: Why won't all the records be present? Cause our migration will take a while?
If we ensured we migrate the mentor profiles first, they should all be present, right?

kicferk1 (approved):

db/migrate/20210629103240_create_participant_profiles.rb
BroiSatse: There's no migration to populate the records in this PR - we'll be just creating new records after this is merged. Merging this first means that we don't need any downtime when migrating the existing records (as the new records created when migration is running won't be lost).

PR #620, opened 2021-07-01
Update participant validation service

ethanmills:
### Context
To get ready for automated validation against DQT, we need to return more information when checking participants.

Change the participant validation service to return correct TRN, QTS status, and active flag status of a participant

### DIFF
--- a/app/services/participant_validation_service.rb
+++ b/app/services/participant_validation_service.rb
@@ -1,21 +1,44 @@
 # frozen_string_literal: true
 
 class ParticipantValidationService
-  def validate(trn:, nino:, full_name:, dob:)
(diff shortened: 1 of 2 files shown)


app/services/participant_validation_service.rb
```diff
@@ -1,21 +1,44 @@
...
+    @nino = nino
+  end
+
+  def validate
+    validated_record = matching_record(trn: trn, nino: nino, full_name: full_name, dob: date_of_birth)
+    return if validated_record.nil?
+
+    {
+      trn: validated_record[:teacher_reference_number],
+      qts: validated_record[:qts_date].present? && validated_record[:qts_date] != "null",
```
kicferk1, line 23: I just love the null check 😢

kicferk1 (approved):

PR #621, opened 2021-07-01
Add docs for cypress + cucumber

callumacrae:
Figured it would be a good idea to get this info out of my head and into a markdown document before i leave :)

### DIFF
--- a/documentation/cypress_cucumber.md
+++ b/documentation/cypress_cucumber.md
@@ -1,0 +1,133 @@
+## Cypress + cucumber setup
+
+The integration tests in this repository are written using Gherkin syntax and
+powered by the following technologies:
+
+- [Cypress]
+- [CypressOnRails]
+- [cypress-cucumber-preprocessor]
+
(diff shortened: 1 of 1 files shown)


documentation/cypress_cucumber.md
```diff
@@ -0,0 +1,133 @@
...
+following in your spec:
+
+- `Given scenario "test_scenario" has been run`
+
+#### Visual and accessibility testing
+
+We use [Percy] for visual testing and [cypress-axe] for basic automated
+accessibility testing. Every time you add a test for a new page (or significant
+state change to an existing page), add the following lines to your spec:
+
```
kicferk1, line 113: Worth saying how to find out what the accessibility issue is when it fails? As in, Open the console and click on the step that was run before (on?) the check?

kicferk1 (approved):
Worth mentioning tricks around dangling state?

kicferk1 (approved):

ethanmills (approved):

PR #622, opened 2021-07-02
CDPTP-224 Remove lead_provider from profile_declarations

blackrat:
### Context

There was a thought to use the lead_provider/declarable_id as an
index, but it's a premature thought for now. With the principle
of YAGNI, we will add it back if and when we need it.

Remove lead_provider from profile_declarations

### Review Checks


ethanmills (approved):
Just to check, there wasn't a `has_many` on LeadProvider, right?

PR #625, opened 2021-07-02
CPDTP-224 add mentors to participant declarations

blackrat:
### Context

This change will allow participant declarations to be submitted for mentors as well as those already for early career teachers.

- Drop old ParticipantDeclarations, since they're only sandbox tests and not actually used for anything else. Dropping
them will have no effect on the API submissions, but will prevent locking on the table whilst it is being modified.
- Update ParticipantDeclarations to link to the user_id and the ProfileDeclaration rather than the EarlyCareerTeacherProfile.
- Link Profiles to the ProfileDeclarations
- Change RecordParticipantEvent to RecordParticipantDeclaration to make it more consistent with other names; and
  - Change the validation to make checks and creation profile agnostic
  - Create the ParticipantDeclaration and associated ProfileDeclaration as part of the RecordParticipantDeclaration 
  - Change existing tests to allow for the new shape and verification

### Guidance to review

This PR goes along with the previous scaffolding to use the profile_declarations as the link between the incoming declarations and the stored values.
Although it adds in mentors, it's the first time that the new shape is actually used.

### Review Checks


db/migrate/20210701000410_truncate_old_participant_declaration_data.rb
```diff
@@ -0,0 +1,12 @@
...
+
+class TruncateOldParticipantDeclarationData < ActiveRecord::Migration[6.1]
+  def up
+    ParticipantDeclaration.delete_all
+  end
+
+  def down
+    ParticipantDeclaration.delete_all
+    ProfileDeclaration.delete_all
+  end
```
ethanmills, line 11: Not sure about this one... Why does it need to be a migration, and why does it need to be reversible?

blackrat: It didn't make much sense to keep records that aren't used for anything apart from the ones on the sandbox for testing the API endpoint, so rather than a migration that converted them to user_ids from early_career_teacher_profile_ids, just dropping them felt more like a pragmatic solution. 
The down can probably be dropped, but down then up could potentially have extra (useless) records in the ProfileDeclaration so either the ProfileDeclarations should be dropped on down, or added to the delete_all on up (or even both).

app/models/profile_declaration.rb
```diff
@@ -3,4 +3,12 @@
...
   delegated_type :declarable, types: %w[EarlyCareerTeacherProfile MentorProfile]
   belongs_to :participant_declaration
+
+  scope :uplift, lambda {
+                   joins(
+                     "left join early_career_teacher_profiles
+                      on declarable_type='EarlyCareerTeacherProfile'
+                      and declarable_id=early_career_teacher_profiles.id",
+                   ).merge(EarlyCareerTeacherProfile.uplift)
+                 }
```
kicferk1, line 13: why `lambda` here and ` -> {` syntax in the other scopes?

app/models/participant_declaration.rb
```diff
@@ -1,16 +1,17 @@
 # frozen_string_literal: true
 
 class ParticipantDeclaration < ApplicationRecord
+  has_one :profile_declaration
   belongs_to :lead_provider
-  belongs_to :early_career_teacher_profile
+  belongs_to :user
```
kicferk1, line 6: Do we care about any live data for this change?

db/migrate/20210701000410_truncate_old_participant_declaration_data.rb
ethanmills: I would expect this to happen on the sandbox console. I think migrations are for schema changes, and when it is correct to change data in them, that should be focussed on production

blackrat: In principle, I agree, but this is tied up to a migration, so we'd have to perform it or the migrations will fail.

https://codingitwrong.com/2016/07/29/how-to-seed-and-migrate-rails-data.html

Particularly,

2. Transform Data Outside of Migrations When Possible
<removed for brevity>
**This isn’t always possible, however. The main time it’s not possible is when data needs to be transformed in order to do a schema migration. For example, if you’re changing what table a column is stored in, you will want to add the new column, transfer the data over, and then drop the old column. This can’t be done easily outside of schema migrations. And in this case, the data transformation is closely tied to the schema change, so keeping it in a migration makes sense.**

app/models/profile_declaration.rb
blackrat: rubocop

app/models/participant_declaration.rb
blackrat: No. We drop all of the existing ones in the migration before changing the shape. It's all scratch data at this point anyway.

kicferk1 (approved):

app/services/record_participant_declaration.rb
```diff
@@ -0,0 +1,107 @@
...
+  attr_accessor :params
+
+  class << self
+    def call(params)
+      new(params).call
+    end
+
+    def required_params
+      %i[participant_id lead_provider declaration_type declaration_date raw_event]
+    end
```
ethanmills, line 15: Do we ever use this, other than in the instance method?

app/services/record_participant_declaration.rb
```diff
@@ -0,0 +1,107 @@
...
+    errors = JsonSchema::ValidateBodyAgainstSchema.call(schema: schema, body: params[:raw_event])
+    raise ActionController::ParameterMissing, (errors.map { |error| error.sub(/\sin schema.*$/, "") }) unless errors.empty?
+  end
+
+  def schema
+    JSON.parse(File.read(JsonSchema::VersionEventFileName.call(version: "0.2")))
+  end
+
+  def add_participant_profile_params!
+    raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless user && profile_type
```
ethanmills, line 50: Is this not just `user&.participant?`

app/services/record_participant_declaration.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+  end
+
+  def lead_provider
+    params[:lead_provider]
+  end
+
+  def actual_lead_provider
+    SchoolCohort.find_by(school: user_profile.school, cohort: user_profile.cohort)&.lead_provider
+  end
```
ethanmills, line 98: `actual_lead_provider`? Then what is `lead_provider`? `lead_provider_id`? `reported_lead_provider_id`?

db/migrate/20210701000400_add_user_id_to_participant_declarations.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddUserIdToParticipantDeclarations < ActiveRecord::Migration[6.1]
+  disable_ddl_transaction!
```
ethanmills, line 4: Why do we need this?

db/migrate/20210701000400_add_user_id_to_participant_declarations.rb
```diff
@@ -0,0 +1,9 @@
...
+
+  def change
+    add_reference :participant_declarations, :user, null: true, type: :uuid, index: { algorithm: :concurrently }
```
ethanmills, line 7: Why `:concurrently`?

db/migrate/20210701000410_truncate_old_participant_declaration_data.rb
blackrat: Which is what it was originally going to do. I'm not wedded to doing this in the migration though. If we're ok with a manual console change and control when the deploy happens to ensure that it's timely enough, I'm ok with dropping it from this PR.

app/services/record_participant_declaration.rb
blackrat: Indeed it is.

db/migrate/20210701000400_add_user_id_to_participant_declarations.rb
blackrat: It's from strong_migrations and required when doing the `index: { algorithm: concurrently }`

db/migrate/20210701000400_add_user_id_to_participant_declarations.rb
blackrat: Again. A stong_migrations thing.

Specifically, when you run it without both those lines it give:

```
=== Dangerous operation detected #strong_migrations ===

Adding an index non-concurrently locks the table. Instead, use:

class AddUserIdToParticipantDeclarations < ActiveRecord::Migration[6.1]
  disable_ddl_transaction!

  def change
    add_reference :participant_declarations, :user, null: true, type: :uuid, index: {algorithm: :concurrently}
  end
end

```

db/migrate/20210701000410_truncate_old_participant_declaration_data.rb
blackrat: Dropped the change. We just need to make sure that we do this manually.

app/services/record_participant_declaration.rb
blackrat: It can be called externally too, which is why it's a class level public. It can be a useful pattern for an intermediate service to follow when it is called from a controller and wants to persist the passed in data.

db/migrate/20210701000400_add_user_id_to_participant_declarations.rb
ethanmills: Fair enough

ethanmills (approved):

app/services/record_participant_declaration.rb
blackrat: lead_provider is the one that comes in from the API Token. Actual lead_provider is the one that the user_profile has association with. One of the tests is that the correct lead_provider is submitting the declaration.

PR #626, opened 2021-07-02
Making ParticipantProfile to shadow legacy profile models

BroiSatse:
Another PR for ParticipantProfile migration / refactor.

This will ensure that any new changes to MentorProfile and EarlyCareerTeacherProfile will be automatically propagated to ParticipantProfile. 

Once merged, we could quickly populate ParticipantProfiles with:

```ruby
MentorProfile.all.find_each do |profile|
  profile.send :save_participant_profile
end
EarlyCareerTeacherProfile.all.find_each do |profile|
  profile.send :save_participant_profile
end
```

### DIFF
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -32,6 +32,8 @@ class ParticipantProfile < ApplicationRecord
   class Mentor < self
     self.ignored_columns = %i[mentor_profile_id]
 
+    has_many :mentee_profiles, class_name: "ParticipantProfile::ECT", foreign_key: :mentor_profile_id, dependent: :nullify
+
     def mentor?
       true
     end
--- a/spec/cypress/support/on-rails.js
+++ b/spec/cypress/support/on-rails.js
@@ -14,8 +14,16 @@ Cypress.Commands.add("appCommands", (body) => {
 
 let createdRecords = {};
 export default {
-  getCreatedRecord(factory, index = 0) {
-    return createdRecords[factory] && createdRecords[factory][index];
+  getCreatedRecord(factory, givenIndex = -1) {
+    if (!createdRecords[factory]) {
+      return null;
+    }
+
+    let index = givenIndex;
+    if (index < 0) {
+      index = createdRecords[factory].length + index;
+    }
+    return createdRecords[factory][index];
   },
 };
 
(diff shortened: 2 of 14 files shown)


app/models/participant_profile.rb
```diff
@@ -3,7 +3,7 @@
 class ParticipantProfile < ApplicationRecord
   belongs_to :user
   belongs_to :school
-  belongs_to :cohort
+  belongs_to :cohort, optional: true
```
kicferk1, line 6: Do we really need to relax this? I know mentor profiles technically have it as optional, but I don't think any live ones should have that problem :P

BroiSatse: Yeah, but ones created in tests do...

kicferk1: I think we did something to around making it non-optional for ects recently, and I think refactoring tests was a part of that (https://github.com/DFE-Digital/early-careers-framework/pull/562).

Do you want to see if we can use the same idea for a quick fix for mentor profiles in tests?

BroiSatse: Well, most likely this is just a matter of adding `cohort` to the factory. If we are certain that this should not be optional I'm happy to change this :)

kicferk1: I am certain this should not be optional. Mentors without cohorts should not exist with our current setup, and I don't think we know of any cases when they should - it was just that mentor profile was copy-pasted from ECT profile at a time when we haven't fixed the same issue on ECT profiles.

spec/cypress/support/on-rails.js
```diff
@@ -14,8 +14,16 @@ Cypress.Commands.add("appCommands", (body) => {
...
+  getCreatedRecord(factory, givenIndex = -1) {
+    if (!createdRecords[factory]) {
+      return null;
+    }
+
+    let index = givenIndex;
+    if (index < 0) {
+      index = createdRecords[factory].length + index;
+    }
+    return createdRecords[factory][index];
```
kicferk1, line 26: Huh, what is this for?

kicferk1 (approved):
Looks good, just a question on cypress setup

spec/cypress/support/on-rails.js
BroiSatse: That is a bit that handles steps like "school was created with created cohort". By default it should take the most recently created record, but it was always taking the first one - which was causing some odd failures.

There's a much bigger bug there to fix though - for some reason, the `createdRecords` is not being reset between the tests (or rather between some tests). This need more debugging and this seems to be fixing the issue for that PR.

BroiSatse:
@kicferk1 Just added few more changes, would be great for you to accept those as well :)

kicferk1 (approved):

PR #628, opened 2021-07-02
CPDTP-304 Allow reset of sandbox data

timabell:
### Context

There are multiple lead providers with the same name in the sandbox. After discussion we think it'll be expedient to reset the whole database and recreate the tokens rather than manually fixup the data.

* Allow reset of sandbox data

### Review Checks

### DIFF
--- a/lib/tasks/safe_reset.rake
+++ b/lib/tasks/safe_reset.rake
@@ -3,7 +3,7 @@
 namespace :db do
   desc "Remove all database content, then seed it. Only for dev environments."
   task safe_reset: :environment do
-    if Rails.env.development? || Rails.env.deployed_development?
+    if Rails.env.development? || Rails.env.deployed_development? || Rails.env.sandbox?
       connection = ActiveRecord::Base.connection
       tables = connection
                    .execute("SELECT * FROM pg_catalog.pg_tables;")


kicferk1 (approved):
I said it like three times already, buut...

Make sure you can recreate their tokens before you run it on sandbox 😂

lib/tasks/safe_reset.rake
```diff
@@ -3,7 +3,7 @@
 namespace :db do
   desc "Remove all database content, then seed it. Only for dev environments."
   task safe_reset: :environment do
-    if Rails.env.development? || Rails.env.deployed_development?
+    if Rails.env.development? || Rails.env.deployed_development? || Rails.env.sandbox?
```
ethanmills, line 6: Have you thought twice? 😛

ethanmills (approved):

lib/tasks/safe_reset.rake
timabell: Could you elaborate on your concerns? I'm not sure what you're getting at

kicferk1: I think it's a joke about the thing that the script prints out :D 

> puts "You should think twice before recreating staging or production database!"
puts "Fire off Rails console and copy the code from this task if you really have to."
puts "Make sure you know what you are doing and maybe do a back-up first."

timabell: I haven't had a direct response so I'm going to assume Michal is right and  it's fine and go ahead and merge. We can always take it out again.

PR #631, opened 2021-07-05
CPDRP-260 API for data studio

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/browse/CPDRP-260)
Create endpoint to expose school rollout info for Data Studio analytics.

Adds a new private endpoint `/api/v1/data-studio/school-rollout`
Adds `Api::V1::DataStudio::SchoolRolloutController`
Adds `DataStudioApiToken`

### Guidance to review
Query is deliberately kept as a SQL query rather than ActiveRecord/Arel to enable it to be easily tweaked by colleagues who don't know Ruby.

### Review Checks

Create a `DataStudioApiToken` from the console and use it to access the endpoint using Postman or similar.


db/seeds/dummy_structures.rb
```diff
@@ -106,6 +106,7 @@
 if Rails.env.deployed_development?
   EngageAndLearnApiToken.find_or_create_by!(hashed_token: "dfce9a34c6f982e8adb4b903f8b6064682e6ad1f7858c41ed8a0a7468abc8896")
   NpqRegistrationApiToken.find_or_create_by!(hashed_token: "1dae3836ed90df4b796eff1f4a4713247ac5bc8a00352ea46eee621d74cd4fcf")
+  DataStudioApiToken.find_or_create_by!(hashed_token: "c7123fb0e2aecb17e1089e01849d71665983e200e891fe726341a08f176c1d64")
```
kicferk1, line 109: Just to make sure - how did you get this hashed token exactly?

I would expect the answer to be 'I ssh-ed into review app, made a new token, and got the hashed version of that from there'. 

If you got it from a local version of your app, it won't work - cause the base that devise uses to digest tokens is different between dev and deployed dev (I think).

app/controllers/api/v1/data_studio/school_rollout_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+              LEFT OUTER JOIN partnerships p on s.id = p.school_id
+              LEFT OUTER JOIN core_induction_programmes cip on sc.core_induction_programme_id = cip.id
+              LEFT OUTER JOIN lead_providers lp on p.lead_provider_id = lp.id
+              LEFT OUTER JOIN delivery_partners dp on p.delivery_partner_id = dp.id 
+            WHERE
+              s.school_status_code IN (1, 3)
+              AND s.school_type_code IN (1, 2, 3, 5, 6, 7, 8, 12, 14, 15, 18, 28, 31, 32, 33, 34, 35, 36, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48)
+              AND s.administrative_district_code ILIKE 'E%'
+            ORDER BY s.urn
+          SQL
```
kicferk1, line 67: Soo, couple of things here.

Number one, let's use full names for things like `school`, `nomination_email`, etc. There are like 5 or 6 of them, and it's hard to keep track of all of them when reading it.

Number two, perhaps making number one irrelevant is: should we try to make this an active record query, perhaps with its own serializer? That would hopefully make us able to use school scopes to select the right schools rather than copy-pasting the right status codes into our where clauses. Using a serializer for it would also make it significantly easier to see what fields we are returning, how, and when.

Thoughts?

db/seeds/dummy_structures.rb
tonyheadford: I ssh'd into dev and did it there, it appears to work.

app/controllers/api/v1/data_studio/school_rollout_controller.rb
tonyheadford: As I mentioned in the PR text, this was intentional so that our data analysts who know SQL but maybe not ActiveRecord/Arel can understand it and modify it if need be

tonyheadford: Agree about the abbreviated table names 👍

kicferk1: > As I mentioned in the PR text, this was intentional so that our data analysts who know SQL but maybe not ActiveRecord/Arel can understand it and modify it if need be. In the previous incarnation this was all ActiveRecord.

Hehe, I do like the idea of saying "no, you fix it" to the data analysts 😂 

Could we add a code comment saying something similar? So that an overly eager dev does not try to "fix" it for the reasons I mentioned above?

kicferk1 (approved):

PR #635, opened 2021-07-05
CDPTP-223 Include mentors in aggregation

plisovin:
### Context

This change will add mentor ‘started’ declarations in aggregation 

- add profile scopes to ```ProfileDeclaration``` for ecf and mentor profiles
- refactored ```record_participant_declaration.rb```

### Review Checks


app/models/profile_declaration.rb
```diff
@@ -4,11 +4,21 @@ class ProfileDeclaration < ApplicationRecord
...
+    )
+  }
+
+  scope :mentor_profiles, lambda {
+    joins(
+      "left join mentor_profiles
+       on declarable_type='MentorProfile'
+       and declarable_id=mentor_profiles.id",
+    )
+  }
```
kicferk1, line 21: Can we point them at new classes before Stan ends up deleting the old classes? Maybe in a separate PR, I just don't want it to get lost.

kicferk1 (approved):

app/models/profile_declaration.rb
blackrat: Is that his final naming? Having them inherit from the eigenclass, I thought, was just an interim position. I'll check with @BroiSatse.

PR #636, opened 2021-07-05
add /api/v1/participant-validation endpoint

asmega:
### Context

- NPQ registration needs to verify TRNs of applicants
- Currently this is done with the `/api/v1/dqt-records` endpoint
- A constraint of the current implementation of this endpoint means the TRNs must match in order for participant to be valid
- With the recent addition of `ParticipantValidationService` we can now wrap this service in an api endpoint
- These means it is now possible to verify participants even if the TRN they entered is incorrect but all other pieces of information are correct

- added `/api/v1/participant-validation` api endpoint which wraps `ParticipantValidationService`

### Guidance to review

- If everything goes smoothly and nothing else is using `/api/v1/dqt-records`, It can be teared down in a later change

### Testing

- Test in review app

### Review Checks

- note down the token
- call new endpoint
```
curl -H "Authorization: Bearer UNHASHED_TOKEN" https://ecf-review-pr-636.london.cloudapps.digital/api/v1/participant-validation/VALID_TRN?full_name=CORRECT_NAME&date_of_birth=YYYY-MM-DD&nino=CORRECT_NINO
```
- see csv extract for data points to test against
- should return correct TRN
- change TRN so it is incorrect
- should still return correct TRN assuming all other fields are correct

### DIFF
--- a/app/controllers/api/v1/participant_validation_controller.rb
+++ b/app/controllers/api/v1/participant_validation_controller.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class ParticipantValidationController < Api::ApiController
+      include ApiTokenAuthenticatable
+
+      def show
+        record = ParticipantValidationService.validate(
+          trn: params[:id],
+          full_name: params[:full_name],
+          date_of_birth: Date.iso8601(params[:date_of_birth]),
+          nino: params[:nino],
+        )
+
+        if record.present?
+          render json: ParticipantValidationSerializer.new(OpenStruct.new(record)).serializable_hash.to_json
+        else
+          head :not_found
+        end
+      end
+
+    private
+
+      def access_scope
+        ApiToken.where(private_api_access: true)
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/participant_validation_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+module Api
+  module V1
+    class ParticipantValidationController < Api::ApiController
+      include ApiTokenAuthenticatable
+
+      def show
+        record = ParticipantValidationService.validate(
+          trn: params[:id],
+          full_name: params[:full_name],
+          date_of_birth: Date.iso8601(params[:date_of_birth]),
```
jamesrose, line 12: This will blow up if `params[:date_of_birth` is nil.

asmega: its mandatory on the service, see https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/services/participant_validation_service.rb#L10 so blowing up is expected. the alternative is i can guard and raise relevant 4xx errors back to the client?

ethanmills (approved):
LGTM

PR #637, opened 2021-07-05
CPDTP-202 Switch from classic autoloader to zeitwerk

blackrat:
### Context

For historical reasons, we have been using the "classic" autoloader. This PR switches it for :zeitwerk

- change application autoloader to :zeitwerk
- Removed "Concerns" from namespace
- Added in inflector for UTM
- Removed spurious Api:: from the Dummy Api controller

### Guidance to review

Things should behave as before. Only naming changes have been applied.

### Testing

All existing tests pass. Only changes are to namespaces and automatically found names in code where they were previously incorrect, but still being included.

### Review Checks

### DIFF
--- a/spec/controllers/api/api_controller_spec.rb
+++ b/spec/controllers/api/api_controller_spec.rb
@@ -3,7 +3,7 @@
 require "rails_helper"
 
 class DummyController < Api::ApiController
-  include Api::ApiTokenAuthenticatable
+  include ApiTokenAuthenticatable
 end
 
 describe DummyController, type: :controller do
--- a/app/controllers/api/school_search_controller.rb
+++ b/app/controllers/api/school_search_controller.rb
@@ -7,7 +7,7 @@ class Api::SchoolSearchController < Api::ApiController
         .eligible
         .ransack(name_or_urn_cont: params[:search_key]).result
         .limit(10)
-      decorated_schools = schools.map { |school| ::Decorators::SchoolDecorator.new(school) }
+      decorated_schools = schools.map { |school| SchoolDecorator.new(school) }
       render json: SchoolSerializer.render(decorated_schools)
     else
       render json: []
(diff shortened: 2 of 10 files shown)


spec/controllers/api/api_controller_spec.rb
```diff
@@ -3,7 +3,7 @@
 require "rails_helper"
 
 class DummyController < Api::ApiController
-  include Api::ApiTokenAuthenticatable
+  include ApiTokenAuthenticatable
 end
```
kicferk1, line 7: Why this change?

kicferk1 (approved):

ethanmills:
Uh oh...

```
/usr/local/bundle/gems/zeitwerk-2.4.2/lib/zeitwerk/loader/callbacks.rb:18:in `on_file_autoloaded': expected file /app/app/decorators/school_decorator.rb to define constant SchoolDecorator, but didn't (Zeitwerk::NameError)
```

spec/controllers/api/api_controller_spec.rb
blackrat: It didn't match the actual namespace since it appears in the `concerns` directory which is treated as a root namespace by zeitwerk

blackrat:
> Uh oh...
> 
> ```
> /usr/local/bundle/gems/zeitwerk-2.4.2/lib/zeitwerk/loader/callbacks.rb:18:in `on_file_autoloaded': expected file /app/app/decorators/school_decorator.rb to define constant SchoolDecorator, but didn't (Zeitwerk::NameError)
> ```

Looks like Decorators is also a root namespace. I'll change to drop the module from it.

ethanmills:
"GET /nominations/choose-school" - (127.0.0.1)): #<ActionView::Template::Error: uninitialized constant NominationRequestForm::Decorators>

PR #638, opened 2021-07-05
Participant profiles the switch

BroiSatse:
This PR switches all the association and code to use ParticipantProfile in place of EarlyCareerTeacherProfile and MentorProfile.

Note: The data has been already synced and migrated. New ECT and mentor profile automatically create ParticipantProfile records, so no data will be lost.


app/models/legacy_ect_profile.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-class EarlyCareerTeacherProfile < ApplicationRecord
+class LegacyECTeacherProfile < ApplicationRecord
+  self.table_name = "early_career_teacher_profiles"
```
kicferk1, line 4: When do we axe them?

app/models/profile_declaration.rb
```diff
@@ -1,14 +1,8 @@
...
   belongs_to :participant_declaration
 
-  scope :uplift, lambda {
-                   joins(
-                     "left join early_career_teacher_profiles
-                      on declarable_type='EarlyCareerTeacherProfile'
-                      and declarable_id=early_career_teacher_profiles.id",
-                   ).merge(EarlyCareerTeacherProfile.uplift)
-                 }
+  scope :uplift, -> { where(participant_profile_id: ParticipantProfile::ECT.uplift.select(:id)) }
```
kicferk1, line 7: https://github.com/DFE-Digital/early-careers-framework/pull/635/files#r663994053 probably needs some conflict resolving with this, right?

app/models/legacy_ect_profile.rb
BroiSatse: Soon - I just keep them to ensure there are no data missing after the switch.

app/models/profile_declaration.rb
BroiSatse: Already resolved :)

kicferk1 (approved):

PR #640, opened 2021-07-06
Skip magic link auth for certain domains on non-production environments

jamesrose:
Draft pending confirmation on applicable domains.

- Expand mock login to cover all non-production environments, including sandbox.
- Set whitelist of email domains to apply mock login to

- Can you log in with an email address with an applicable domain without using the magic-link process?

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -3,10 +3,29 @@
 class Users::SessionsController < Devise::SessionsController
   include ApplicationHelper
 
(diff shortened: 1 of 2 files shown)


kicferk1 (approved):
Looks reasonable to me. We will still need to add a bunch of users with relevant emails to our db, either via console or seeding I guess.

We trust that lead providers won't try to mess with each other on sandbox, right? 😂

jamesrose:
@kicferk1 Yes, we are trusting they won't log in as each other. They _would_ need to know the full email address of an existing user.

PR #643, opened 2021-07-06
DQT NI number data may be cased incorrectly

asmega:
### Context

- I have seen NI numbers in prod API return NI numbers that are mixed cased

- To better out matching this makes NI matching case insensitive

### Guidance to review

- nil guard is present incase api data is not provided

### Testing

- Test on review app
- Test data seems to be upcased correctly so enter downcased data to verify case insensitivity

### Review Checks

- ssh into review app
- fire up rails console
- see csv test data to create a test case
- `ParticipantValidationService.validate(trn: "CORRECT_TRN", full_name: "CORRECT_NAME", date_of_birth: WRONG_DATE, nino: "INCORRECTLY_CASED_NINO")`
- should return correct record

### DIFF
--- a/spec/services/participant_validation_service_spec.rb
+++ b/spec/services/participant_validation_service_spec.rb
@@ -53,15 +53,21 @@ RSpec.describe ParticipantValidationService do
         ).to eql({ trn: trn, qts: true, active_alert: false })
       end
 
-      it "returns the validated details when nino is wrong" do
+      it "returns the validated details when name is wrong" do
         expect(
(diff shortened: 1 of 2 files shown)


spec/services/participant_validation_service_spec.rb
```diff
@@ -53,17 +53,23 @@
         ).to eql({ trn: trn, qts: true, active_alert: false })
       end
 
-      it "returns the validated details when nino is wrong" do
+      it "returns the validated details when name is wrong" do
```
ethanmills, line 56: Good catch, thanks

kicferk1 (approved):

ethanmills (approved):
Looks good!

PR #647, opened 2021-07-07
Cpdrp 532 it materials link

xswoodsx:
### Context

Ticket: [532](https://dfedigital.atlassian.net/browse/CPDRP-532) 

Adding a link to the induction tutor materials on the list cohorts page.

### Testing
Cypress tests were already in there. 

### Review Checks

- Login in as an induction tutor 
- Go to your cohorts. It should show you the text set out in the ticket with a link to the chosen cips induction materials page.


app/controllers/pages_controller.rb
```diff
@@ -6,6 +6,22 @@ def show
...
+    case provider_name
+    when "ambition-institute"
+      "ambition"
+    when "educational-development-trust"
+      "educational_development_trust"
+    when "teach-first"
+      "teach_first"
+    when "ucl-institute-of-education"
+      "ucl"
+    end
```
kicferk1, line 25: Should we rename the folders instead, so we can use `provider_name.underscore` for all of them?

app/views/schools/cohorts/show.html.erb
```diff
@@ -63,5 +63,15 @@
         </ul>
       </li>
     </ol>
+
+    <% if @school_cohort.core_induction_programme.present? %>
+      <div class="govuk-inset-text">Use these handbooks and training materials to help you plan your school’s
+        <%= govuk_link_to @school_cohort.core_induction_programme.name, induction_tutor_materials_path(
+          provider: @school_cohort.core_induction_programme.name.downcase.tr(" ", "-"), year: :year_one
```
kicferk1, line 70: `tr(" ", "-")` feels like `dasherize` - can we use that instead?

app/controllers/pages_controller.rb
xswoodsx: i was thinking to do that but thought it better to change code rather than folder names if providers ever change their names. Don't mind either way.

app/views/schools/cohorts/show.html.erb
xswoodsx: dasherize replaces underscores with hyphens by the looks of it - https://apidock.com/rails/String/dasherize

kicferk1 (approved):

PR #656, opened 2021-07-07
introduce CpdLeadProvider

asmega:
### Context

- Currently there is `LeadProvider` (which are ECF specific providers) and `NpqLeadProvider` and there is some overlap between the 2 entities
- We want a way of unifying this so a provider is capable of accessing multiple APIs with a single token
- Whilst at the same time keeping the data modelling sane

- Introduce `CpdLeadProvider` which encompasses all lead providers
- These can then be associated to `LeadProvider` and `NpqLeadProvider`
- Populate `CpdLeadProvider` and the data between the different tables

### Guidance to review

- A future change will follow to make the association between `LeadProvider` + `CpdLeadProvider` or `NpqLeadProvider` + `CpdLeadProvider` mandatory. It is only optional here due to missing data
- Another future change is that api token are currently associated with `LeadProvider` this will need to be associated with `CpdLeadProvider` so one api token is capable of accessing multiple APIs

### Testing

- There are no real user facing changes. and the app should behave as before
- Data associations should be correctly setup between `CpdLeadProvider`, `LeadProvider` and `NpqLeadProvider`

### Review Checks

- Can't test in review app as data is removed then reseeded causing the outcome of the data migrations to be no longer observable

### DIFF
--- a/db/migrate/20210707151757_populate_lp_and_cpd_lp_assocs.rb
+++ b/db/migrate/20210707151757_populate_lp_and_cpd_lp_assocs.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class PopulateLpAndCpdLpAssocs < ActiveRecord::Migration[6.1]
+  def up
+    LeadProvider.all.each do |lp|
+      lp.update(cpd_lead_provider: CpdLeadProvider.find_by(name: lp.name))
+    end
+
+    NpqLeadProvider.all.each do |lp|
+      lp.update(cpd_lead_provider: CpdLeadProvider.find_by(name: lp.name))
+    end
+  end
+
+  def down
+    LeadProvider.update_all(cpd_lead_provider: nil)
+    NpqLeadProvider.update_all(cpd_lead_provider: nil)
+  end
+end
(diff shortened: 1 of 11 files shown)


db/migrate/20210707151757_populate_lp_and_cpd_lp_assocs.rb
```diff
@@ -0,0 +1,18 @@
...
+
+class PopulateLpAndCpdLpAssocs < ActiveRecord::Migration[6.1]
+  def up
+    LeadProvider.all.each do |lp|
+      lp.update(cpd_lead_provider: CpdLeadProvider.find_by(name: lp.name))
+    end
+
+    NpqLeadProvider.all.each do |lp|
+      lp.update(cpd_lead_provider: CpdLeadProvider.find_by(name: lp.name))
+    end
```
kicferk1, line 11: Can we do `update!` rather than `update`? I know, I see no way in hell this can fail, but I think it's a good idea to enforce this as a convention.

kicferk1 (commented):
So will the api tokens need to belong to `CpdLeadProvider` after this change? That's the original problem we are trying to solve, right?

ethanmills (approved):
Just to check, the next steps are to make CpdLeadProvider required, and to change the API keys over?

asmega:
> Just to check, the next steps are to make CpdLeadProvider required, and to change the API keys over?

Yes. I'm doing the keys now. I will probably do the required-ness part after that as the key change requires another column

asmega:
> So will the api tokens need to belong to `CpdLeadProvider` after this change? That's the original problem we are trying to solve, right?

Yes, my next change will make them belong to both. There will then be another change to remove the original direct association with `LeadProvider` and one will have to proxy thru `CpdLeadProvider` instead

timabell:
For context `NpqLeadProvider` was introduced in #536

PR #657, opened 2021-07-07
CPDRP-578: Fix change induction coordinator logic

ethanmills:
CPDRP-578

Previously, changing induction coordinator would not work if that user
was also a mentor. If they managed more than one school, changing any of
 those schools would delete the user altogether. This change:
 - Makes it possible to change induction coordinator but leave the
 mentor unchanged
 - Makes it possible to change induction coordinator for a school
 without removing the old coordinator from any other schools

# Testing
## Change induction tutor but leave mentor
- Sign in as school-leader@example.com
- Add yourself as a mentor
- Sign in as admin@example.com
- Change induction tutor for school with urn 999999
- ensure school-leader@example.com still shows in list of participants for school
## Change induction tutor with multiple schools
- sign in as admin@example.com
- add school-leader@example.com as induction tutor to another school
- Change induction tutor for school urn 999999
- Sign in as school-leader@example.com
- Ensure you see the school added in step 2

### DIFF
--- a/app/services/create_induction_tutor.rb
+++ b/app/services/create_induction_tutor.rb
@@ -11,7 +11,7 @@ class CreateInductionTutor < BaseService
 
   def call
     ActiveRecord::Base.transaction do
-      school.induction_coordinators.first.destroy! if school.induction_coordinators.first
+      remove_existing_induction_coordinator
 
       if (user = User.find_by(email: email))&.induction_coordinator?
         raise if user.full_name != full_name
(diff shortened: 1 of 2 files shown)


app/services/create_induction_tutor.rb
```diff
@@ -34,4 +34,19 @@ def start_url
...
   end
+
+private
+
+  def remove_existing_induction_coordinator
+    existing_induction_coordinator = school.induction_coordinators.first
+    return if existing_induction_coordinator.nil?
+
+    if existing_induction_coordinator.induction_coordinator_profile.schools.count > 1
+      existing_induction_coordinator.induction_coordinator_profile.schools.delete(school)
```
kicferk1, line 45: Just double checking, this deletes the relation between IC and school, right? And cannot possibly delete the school itself?

kicferk1 (approved):
Looks pretty good, the only question is me being paranoid

app/services/create_induction_tutor.rb
ethanmills: I'll add it to the test 🙂

kicferk1 (approved):

PR #662, opened 2021-07-08
Shadowing NPQ profile by ParticipantProfile

BroiSatse:
It was decided that NPQs are still partiicpants and it is useful for them to be represented in the ParticipantProfile table as well.

In this PR we oncly create shadowing ParticipantProfile::NPQ model to ensure data consistency when we run the data migration. We'll be switching to MTI once the data is migrated.

### DIFF
--- a/app/models/participant_profile/mentor.rb
+++ b/app/models/participant_profile/mentor.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class ParticipantProfile::Mentor < ParticipantProfile
+  self.ignored_columns = %i[mentor_profile_id]
+
+  has_many :mentee_profiles,
+           class_name: "ParticipantProfile::ECT",
+           foreign_key: :mentor_profile_id,
+           dependent: :nullify
+  has_many :mentees, through: :mentee_profiles, source: :user
+
+  belongs_to :cohort
+  belongs_to :school
+  belongs_to :core_induction_programme, optional: true
+
+  def mentor?
+    true
+  end
+end
--- a/app/controllers/api/v1/npq_profiles_controller.rb
+++ b/app/controllers/api/v1/npq_profiles_controller.rb
@@ -2,7 +2,7 @@
 
 module Api
   module V1
-    class NpqProfilesController < Api::ApiController
+    class NPQProfilesController < Api::ApiController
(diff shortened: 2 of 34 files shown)


app/models/participant_profile/mentor.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  has_many :mentee_profiles,
+           class_name: "ParticipantProfile::ECT",
+           foreign_key: :mentor_profile_id,
+           dependent: :nullify
+  has_many :mentees, through: :mentee_profiles, source: :user
+
+  belongs_to :cohort
+  belongs_to :core_induction_programme, optional: true
+
```
kicferk1, line 14: we should inherit that from `ParticipantProfile`, right?

kicferk1: Also, I wonder if we should move the line 
```
belongs_to :core_induction_programme, optional: true
```
to mentor and ect profiles now, since it really doesn't make sense for NPQ profiles

BroiSatse: Yup, just did that! :D I think there might be an extra abstract model between ParticipantProfile and ECT/Mentor (with school / cohort association) but it is potentially too much for now yet?

kicferk1: Hmmm. Wasn't the original purpose of `ParticipantProfile` to commonize the similarities between mentors and ects?

I am a bit worried that we will want to have queries like `fetch me all participants with their schools and cohorts`, and by adding NPQ profiles into participants we make those queries more complicated. Thoughts?

kicferk1 (approved):

app/models/participant_profile/mentor.rb
BroiSatse: So it depends on what we want to query - to fetch all the pariticpants with schools/cohorts when they exist we'd need to do manually preload them. I might add some moethods on base to do that easily later on. I **think** it willl still be simpler than getting a list of users  based on profiles and then try fetching all three types of profiles to determine what sort of profile they have.

It was noted by few people that NPQs are pretty different to other participants. Which would even stronger justify the extra common base for ECT and Mentors - if you want to query all ECT-related profile, just query this extra base model, which will automatically leave out all the NPQs.

PR #664, opened 2021-07-08
LeadProviderApiToken now hangs off CpdLeadProvider

asmega:
### Context

- ECF and NPQ have different and overlapping providers
- We want providers to be able to access our apis with a single api token

- Make existing `LeadProviderApiToken` be associated with `CpdLeadProvider` which contains all providers we deal with
- Existing api endpoints continue to work by associating with `LeadProvider` thru `CpdLeadProvider`
- Updated seeds to ensure seed setup continues to work

### Guidance to review

- Once this is deployed we can remove `LeadProviderApiToken#lead_provider`
- We will also be able to mark `belongs_to` associations are required since all data migrations should be complete
- This should also allows to introduce NPQ related api endpoints that can now be accessed by providers via the updated tokening system

### Testing

- Should be able to access existing api endpoints with existing tokens

### Review Checks

- I don't believe this can be tested in the review app due to the way the database is reset after migrations therefore the outcomes of data change migrations are not observable
- Having said that one can check the update of the process did work with a primitive of...
```
(LeadProvider.all.pluck(:name) + NPQLeadProvider.all.pluck(:name)).uniq.count == CpdLeadProvider.count
```

### DIFF
--- a/db/migrate/20210708091338_move_lp_tokens.rb
+++ b/db/migrate/20210708091338_move_lp_tokens.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class MoveLpTokens < ActiveRecord::Migration[6.1]
+  def change
+    LeadProviderApiToken.all.each do |token|
+      token.update!(cpd_lead_provider: token.lead_provider.cpd_lead_provider)
+    end
+  end
+end
--- a/db/seeds/initial_seed.rb
+++ b/db/seeds/initial_seed.rb
@@ -8,19 +8,16 @@ teach_first_cip = CoreInductionProgramme.find_or_create_by!(name: "Teach First")
 ucl_cip = CoreInductionProgramme.find_or_create_by!(name: "UCL Institute of Education")
 
 [
-  { provider_name: "Ambition Institute", cip: ambition_cip, token: "ambition-token" },
-  { provider_name: "Best Practice Network", cip: ucl_cip, token: "best-practice-token" },
-  { provider_name: "Capita", cip: ambition_cip, token: "capita-token" },
-  { provider_name: "Education Development Trust", cip: edt_cip, token: "edt-token" },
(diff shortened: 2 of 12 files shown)


db/seeds/initial_seed.rb
```diff
@@ -51,3 +48,31 @@
 ].each do |hash|
   NPQCourse.find_or_create_by!(name: hash[:name], id: hash[:id])
 end
+
+all_provider_names = (LeadProvider.pluck(:name) + NPQLeadProvider.pluck(:name)).uniq
+
+all_provider_names.each do |name|
+  CpdLeadProvider.create!(name: name)
+end
```
kicferk1, line 56: lets' use `find_or_create_by!` please - idempotent seeds are a big convenience.

db/migrate/20210708091338_move_lp_tokens.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class MoveLpTokens < ActiveRecord::Migration[6.1]
+  def change
+    LeadProviderApiToken.all.each do |token|
+      token.update(cpd_lead_provider: token.lead_provider.cpd_lead_provider)
```
kicferk1, line 6: `update!` please :)

db/seeds/initial_seed.rb
asmega: sorry, totally forgot to do this. should be fixed now

kicferk1 (approved):

timabell:
@asmega would you mind explaining the reason behind this schema change, I'm struggling to understand it from the description.

---

> We want providers to be able to access our apis with a single api token

As far as I can tell this was already the case.

---

> ECF and NPQ have different and overlapping providers

I don't entirely understand this statement. I assume this is the key reason so would you mind expanding on this a bit for my benefit?

Thanks. Sorry for the post-merge comment, I came across the change in schema and didn't understand it.

PR #666, opened 2021-07-08
CPDRP-585: Add a status enum to participant profiles

ethanmills:
### Context
CPDRP-585
This is the first in a series of PRs which will allow admins to delete participants, and make that information available to LPs in the API

Add status enum to participant profiles, defaulting to active

### Testing
Added unit test for enum

### DIFF
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -3,6 +3,11 @@
 class ParticipantProfile < ApplicationRecord
   belongs_to :user
 
+  enum status: {
+    active: "active",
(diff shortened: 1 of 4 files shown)


BroiSatse (approved):

kicferk1:
Uhh, from Track and Pay code that I've read, there is a `ParticipantRecordStateMachine`, which is included in `ParticipationRecord`, which tracks whether a user is active. 

Why do we want an enum on participant profile for this? Will we remove the other solution?

PR #669, opened 2021-07-09
CPDRP-575: Add chaser emails

ethanmills:
Add and update a series of targeted mailshots.

I haven't added rake tasks for these because they're going to be scheduled (console, `InviteSchools.new.delay(run_at: sometime, queue: "mailers", priority: 1).send_induction_coordinator_choose_route_chasers`)

### DIFF
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -169,7 +169,7 @@ class InviteSchools
     induction_coordinators_not_in_partnership.find_each do |induction_coordinator|
       school_without_provider = induction_coordinator.schools.first do |school|
         school.school_cohorts.first.induction_programme_choice == "full_induction_programme" &&
-          school.partnerships.none? { |partnership| partnership.challenge_reason.nil? }
+          !school.partnered?(Cohort.current)
       end
       SchoolMailer.induction_coordinator_reminder_to_choose_provider_email(
         recipient: induction_coordinator.email,
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -134,6 +134,75 @@ class InviteSchools
     end
   end
 
+  def send_induction_coordinator_choose_route_chasers
+    induction_coordinators_not_chosen_route = User.includes(:induction_coordinator_profile, schools: :school_cohorts)
+                                                  .joins(:induction_coordinator_profile)
+                                                  .left_outer_joins(schools: :school_cohorts)
+                                                  .where(school_cohorts: { id: nil })
+                                                  .where.not(last_sign_in_at: nil)
+    induction_coordinators_not_chosen_route.find_each do |induction_coordinator|
+      school_without_cohort = induction_coordinator.schools
+                                                   .first { |school| school.school_cohorts.none? }
+      SchoolMailer.induction_coordinator_reminder_to_choose_route_email(
+        recipient: induction_coordinator.email,
+        name: induction_coordinator.full_name,
+        school_name: school_without_cohort.name,
+        sign_in_url: choose_route_chaser_sign_in_url,
+      ).deliver_later
+    rescue StandardError
+      logger.info "Error emailing induction coordinator, email: #{induction_coordinator.email} ... skipping"
+    end
+  end
+
+  def send_induction_coordinator_choose_provider_chasers
+    induction_coordinators_not_in_partnership = User.includes(
+      :induction_coordinator_profile,
+      schools: %i[school_cohorts partnerships],
+    ).where(
+      id:
+        School.unpartnered(Cohort.current.start_year)
+              .joins(:school_cohorts, :induction_coordinator_profiles)
+              .where(school_cohorts: { induction_programme_choice: "full_induction_programme" })
+              .pluck(:user_id),
+    )
+
+    induction_coordinators_not_in_partnership.find_each do |induction_coordinator|
+      school_without_provider = induction_coordinator.schools.first do |school|
+        school.school_cohorts.first.induction_programme_choice == "full_induction_programme" &&
+          school.partnerships.none? { |partnership| partnership.challenge_reason.nil? }
+      end
+      SchoolMailer.induction_coordinator_reminder_to_choose_provider_email(
+        recipient: induction_coordinator.email,
+        name: induction_coordinator.full_name,
+        school_name: school_without_provider.name,
+        sign_in_url: choose_provider_chaser_sign_in_url,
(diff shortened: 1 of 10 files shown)


app/services/invite_schools.rb
```diff
@@ -134,6 +134,75 @@ def send_induction_coordinator_sign_in_chasers
...
+              .joins(:school_cohorts, :induction_coordinator_profiles)
+              .where(school_cohorts: { induction_programme_choice: "full_induction_programme" })
+              .pluck(:user_id),
+    )
+
+    induction_coordinators_not_in_partnership.find_each do |induction_coordinator|
+      school_without_provider = induction_coordinator.schools.first do |school|
+        school.school_cohorts.first.induction_programme_choice == "full_induction_programme" &&
+          school.partnerships.none? { |partnership| partnership.challenge_reason.nil? }
+      end
```
kicferk1, line 173: should we make it a `filter` rather than `first`? If a coordinator has many schools and hasn't chosen a partner for any of them, maybe they should get an email for each?

app/services/invite_schools.rb
```diff
@@ -134,6 +134,75 @@ def send_induction_coordinator_sign_in_chasers
...
 
+  def send_induction_coordinator_choose_route_chasers
+    induction_coordinators_not_chosen_route = User.includes(:induction_coordinator_profile, schools: :school_cohorts)
+                                                  .joins(:induction_coordinator_profile)
+                                                  .left_outer_joins(schools: :school_cohorts)
+                                                  .where(school_cohorts: { id: nil })
+                                                  .where.not(last_sign_in_at: nil)
+    induction_coordinators_not_chosen_route.find_each do |induction_coordinator|
+      school_without_cohort = induction_coordinator.schools
+                                                   .first { |school| school.school_cohorts.none? }
```
kicferk1, line 145: Ditto here, single school vs multiple schools

app/services/invite_schools.rb
```diff
@@ -134,6 +134,75 @@ def send_induction_coordinator_sign_in_chasers
     end
   end
 
+  def send_induction_coordinator_choose_route_chasers
+    induction_coordinators_not_chosen_route = User.includes(:induction_coordinator_profile, schools: :school_cohorts)
+                                                  .joins(:induction_coordinator_profile)
+                                                  .left_outer_joins(schools: :school_cohorts)
+                                                  .where(school_cohorts: { id: nil })
+                                                  .where.not(last_sign_in_at: nil)
+    induction_coordinators_not_chosen_route.find_each do |induction_coordinator|
```
kicferk1, line 143: Why `find_each` instead of `each`?

asmega (approved):
Looks fine

app/services/invite_schools.rb
```diff
@@ -134,6 +134,75 @@ def send_induction_coordinator_sign_in_chasers
...
+        sign_in_url: choose_provider_chaser_sign_in_url,
+      ).deliver_later
+    rescue StandardError
+      logger.info "Error emailing induction coordinator, email: #{induction_coordinator.email} ... skipping"
+    end
+  end
+
+  def send_induction_coordinator_choose_materials_chasers
+    induction_coordinators_not_chosen_materials = User.includes(:induction_coordinator_profile, schools: :school_cohorts)
+                                                      .joins(:induction_coordinator_profile, schools: :school_cohorts)
```
kicferk1, line 187: Should we make `.includes(:induction_coordinator_profile, schools: :school_cohorts).joins(:induction_coordinator_profile, schools: :school_cohorts)` a scope?

app/services/invite_schools.rb
```diff
@@ -134,6 +134,75 @@ def send_induction_coordinator_sign_in_chasers
...
+        School.unpartnered(Cohort.current.start_year)
+              .joins(:school_cohorts, :induction_coordinator_profiles)
+              .where(school_cohorts: { induction_programme_choice: "full_induction_programme" })
+              .pluck(:user_id),
+    )
+
+    induction_coordinators_not_in_partnership.find_each do |induction_coordinator|
+      school_without_provider = induction_coordinator.schools.first do |school|
+        school.school_cohorts.first.induction_programme_choice == "full_induction_programme" &&
+          school.partnerships.none? { |partnership| partnership.challenge_reason.nil? }
```
kicferk1, line 172: Can this be a boolean method on a school?

spec/services/invite_schools_spec.rb
```diff
@@ -337,4 +338,171 @@
...
+    it "does not email coordinators who have chosen providers for all their schools" do
+      induction_coordinator = create(:user, :induction_coordinator, last_sign_in_at: Time.zone.now, current_sign_in_at: Time.zone.now)
+      create(:school_cohort, school: induction_coordinator.schools.first, induction_programme_choice: "full_induction_programme", cohort: cohort)
+      create(:partnership, school: induction_coordinator.schools.first, cohort: cohort)
+      InviteSchools.new.send_induction_coordinator_choose_provider_chasers
+      expect(SchoolMailer).not_to delay_email_delivery_of(:induction_coordinator_reminder_to_choose_provider_email)
+    end
+
+    it "does not email coordinators who only have CIP schools" do
+      induction_coordinator = create(:user, :induction_coordinator, last_sign_in_at: Time.zone.now, current_sign_in_at: Time.zone.now)
```
kicferk1, line 425: Can we commonise this line in our test?

app/services/invite_schools.rb
ethanmills: First was a decision, some of these have 40 schools, we only want to email them once

app/services/invite_schools.rb
ethanmills: Batch the queries. It's quite a beefy query, with thousands of results. Probably not necessary, but seemed prudent

app/services/invite_schools.rb
ethanmills: I would normally write this in ActiveRecord terms, but here I'm trying to avoid N+1 queries and do everything with the loaded model

ethanmills: Heh, looks like rails is cleverer than I thought

app/services/invite_schools.rb
ethanmills: I don't think so. I'm not sure what we would call it, and it doesn't belong to a user. If anything it's a property of the induction_coordinator_profile, but that would just make this more difficult

kicferk1 (approved):

PR #671, opened 2021-07-09
Renamed NPQProfile to NPQValidationData

BroiSatse:
Unlike previous MentorProfile and EarlyCareerTeacherProfile, NPQProfile contains a lot of data that is specific only to NPQ participants. Additionally, it is already exposed via API which is being used by `npq_register`.

Because of this, I have renamed the `NPQProfile` model to `NPQValidationData` and associated it with the `ParticipantProfile::NPQ` model. To avoid any API changes, I have kept the logic which automatically creates ParticpantProfile::NPQ when NPQValidationData is created via API (which we can tackle later on).

### DIFF
--- a/app/models/npq_validation_data.rb
+++ b/app/models/npq_validation_data.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
-class NPQProfile < ApplicationRecord
+class NPQValidationData < ApplicationRecord
+  # TODO: Rename table
+  self.table_name = "npq_profiles"
+
+  has_one :profile, class_name: "ParticipantProfile::NPQ", foreign_key: :id
   belongs_to :user
   belongs_to :npq_lead_provider
   belongs_to :npq_course
--- a/app/models/participant_profile/npq.rb
+++ b/app/models/participant_profile/npq.rb
@@ -2,4 +2,6 @@
 
 class ParticipantProfile::NPQ < ParticipantProfile
   belongs_to :school, optional: true
+
+  has_one :validation_data, class_name: "NPQValidationData", foreign_key: :id, dependent: :destroy
 end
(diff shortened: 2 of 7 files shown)


app/models/npq_validation_data.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
-class NPQProfile < ApplicationRecord
+class NPQValidationData < ApplicationRecord
+  # TODO: Rename table
+  self.table_name = "npq_profiles"
+
+  has_one :profile, class_name: "ParticipantProfile::NPQ", foreign_key: :id
```
ethanmills, line 7: We're using the primary key as a foreign key? Are the guaranteed to stay in sync for newly created records?

ethanmills: Do we have any DB enforced consistency?

BroiSatse: Yes, they have been migrated this way and since association are defined this way it is pretty much guaranteed. Having association via id-id is quite interesting IMO as it effectively prevents reassigning the profiles (which is perfect for situations when one table is an extension of other).

However it is not really a proper RailsWay (tm) - so theoretically it is possible to reassign associated object which would in fact change the associated object id (which might be quite problematic). I'd tackle this in a separate PR.

tonyheadford: Will this id <-> id association lead us to not being able to delete either one?

BroiSatse: @tonyheadford - Etiher one of them can be deleted and we can controll the behaviour on deletion. There's really nothing special in id<->id association.

BroiSatse: @ethanmills There's no need for the db consistency here. Both of this record represent the same domain model (NPQ Participant) in two different bounded context (which is very common in DDD - https://martinfowler.com/bliki/BoundedContext.html) and none of them really requires the presence of the other one.

ethanmills (approved):

PR #674, opened 2021-07-12
Don't alert on 4xx in production

ethanmills:
Retain alerts in dev/stag at lower level to catch issues before live


jamesrose (approved):
🙏

PR #676, opened 2021-07-12
Payments-spec cleanup patches

timabell:
### Context

These commits fell out of the work I'm doing to convert the turnip specs to plain rspec.

https://dfedigital.atlassian.net/browse/CPDTP-302

* Remove rounding from `output_payment_calculations_spec`
* Use `build_stubbed` in payment specs (much faster)
* Aggregate failures in `service_fee_calculations_spec`
* Refactor tests - rename `DummyClass` to `ModuleTestHarness`

### Guidance to review

Have a look at the individual commits.

### DIFF
--- a/spec/lib/payment_calculator/ecf/contract/output_payment_calculations_spec.rb
+++ b/spec/lib/payment_calculator/ecf/contract/output_payment_calculations_spec.rb
@@ -1,13 +1,13 @@
 # frozen_string_literal: true
 
 require "payment_calculator/ecf/contract/output_payment_calculations"
-class DummyClass
+class ModuleTestHarness
   include PaymentCalculator::Ecf::Contract::OutputPaymentCalculations
 end
 
 describe ::PaymentCalculator::Ecf::Contract::OutputPaymentCalculations do
   let(:contract) { create(:call_off_contract) }
-  let(:call_off_contract) { DummyClass.new({ contract: contract }) }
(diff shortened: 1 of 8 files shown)


kicferk1 (approved):
I would be careful using `build_stubbed`. If we save miliseconds on the run, then it feels insignificant compared to the overhead of 'firing off rspec', for example.

I would be a bit worried about false positives - in a hypothetical case when without saving a db record we would never get into failed state. I _think_ it's not a problem here, but that's the thing - is a couple milliseconds per run worth the uncertainty of 'I think' rather than 'I am sure'? Thoughts?

timabell:
> I would be careful using `build_stubbed`. If we save miliseconds on the run, then it feels insignificant compared to the overhead of 'firing off rspec', for example.
> 
> I would be a bit worried about false positives - in a hypothetical case when without saving a db record we would never get into failed state. I _think_ it's not a problem here, but that's the thing - is a couple milliseconds per run worth the uncertainty of 'I think' rather than 'I am sure'? Thoughts?

You raise a good question.

We do have coverage elsewhere that will also exercise this code in more of an integration-test style, so I think we can safely unit-test the individual payment classes and then have bigger integration tests on the whole set together such as https://github.com/DFE-Digital/early-careers-framework/blob/636aa14a4b259cc8bd03bf008b49bbe2494b5f76/spec/services/calculation_orchestrator_spec.rb#L5-L6

PR #678, opened 2021-07-12
Time to add participants chaser emails

ethanmills:


spec/services/invite_schools_spec.rb
```diff
@@ -504,6 +519,120 @@
...
+      create(:school_cohort, school: school_without_participants, induction_programme_choice: "core_induction_programme")
+      induction_coordinator = create(:user, :induction_coordinator, school_ids: [school_without_participants.id, *schools_with_participants.map(&:id)])
+
+      InviteSchools.new.send_induction_coordinator_add_participants_email
+      expect(SchoolMailer).to delay_email_delivery_of(:induction_coordinator_add_participants_email)
+                                .with(hash_including(
+                                        recipient: induction_coordinator.email,
+                                        name: induction_coordinator.full_name,
+                                        sign_in_url: String,
+                                      ))
```
kicferk1, line 627: Can we commonise these expects somehow? They are 6 lines long each, and the difference seems to be tiny (if at all)

kicferk1 (approved):

PR #680, opened 2021-07-13
NPQ applications api

asmega:
### Context

- Providers need to be able to fetch NPQ applications from us

- Add `/api/v1/npq-applications` endpoint
- Can fetch json and csv responses
- Added swagger docs for this new endpoints so providers can see how to use it

### Guidance to review

- JSON endpoint accepts pagination and the CSV endpoint does not. This is by design to mirror the participants endpoint for consistency
- I've implemented the filter `updated_since` to also mirror the partcipants endpoint for consistency
- I have opted to not use `JSONAPI::Resources` although powerful it seems to lack flexibility and extensibility
- This feature is behind a feature flag so it can be made available in sandbox whilst disabled in production

### Testing

- Test in review app

### Review Checks

- Create or use an existing `LeadProviderApiToken`
- Ensure there exists some data to fetch from the api
```rb
NPQValidationData.create!(npq_course: NPQCourse.all.sample, npq_lead_provider: NPQLeadProvider.all.sample, date_of_birth: 50.year
s.ago, teacher_reference_number: rand(1_000_000..9_999_999).to_s, school_urn: rand(100_000..999_999), headteacher_status: "no", funding_choice: "s
chool", user: User.all.sample)
```
- Make json api call
```sh
curl -H "Authorization: Bearer TOKEN" https://ecf-review-pr-680.london.cloudapps.digital/api/v1/npq-applications
```
- Make CSV api call
```sh
curl -H "Authorization: Bearer TOKEN" https://ecf-review-pr-680.london.cloudapps.digital/api/v1/npq-applications.csv
```
- Test data contains only applications for the current lead provider based off the api token 
- Test pagination
- Test filtering
- Check the docs https://ecf-review-pr-680.london.cloudapps.digital/api-docs

### DIFF
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module Api
+  module V1
+    class NPQApplicationsController < Api::ApiController
+      include ApiTokenAuthenticatable
+      include ApiPagination
+
+      def index
+        respond_to do |format|
+          format.json do
+            render json: NpqApplicationSerializer.new(paginate(scope)).serializable_hash
+          end
+
+          format.csv do
+            render body: NpqApplicationCsvSerializer.new(scope).call
+          end
+        end
+      end
+
+    private
+
+      def npq_lead_provider
+        current_api_token.cpd_lead_provider.npq_lead_provider
+      end
+
+      def scope
+        npq_lead_provider.npq_profiles.includes(:user, :npq_course)
+      end
+
+      def access_scope
+        LeadProviderApiToken.all
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -11,11 +11,11 @@ module Api
       def index
         respond_to do |format|
           format.json do
-            render json: NPQApplicationSerializer.new(paginate(scope)).serializable_hash
+            render json: NPQApplicationSerializer.new(paginate(query_scope)).serializable_hash
           end
 
           format.csv do
-            render body: NPQApplicationCsvSerializer.new(scope).call
+            render body: NPQApplicationCsvSerializer.new(query_scope).call
           end
         end
       end
@@ -26,10 +26,10 @@ module Api
         current_api_token.cpd_lead_provider.npq_lead_provider
       end
 
-      def scope
-        s = npq_lead_provider.npq_profiles.includes(:user, :npq_course)
-        s = s.where("updated_at > ?", Time.iso8601(updated_since_filter)) if updated_since_filter.present?
-        s
+      def query_scope
+        scope = npq_lead_provider.npq_profiles.includes(:user, :npq_course)
+        scope = scope.where("updated_at > ?", Time.iso8601(updated_since_filter)) if updated_since_filter.present?
+        scope
       end
 
       def filter
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -11,11 +11,11 @@ module Api
       def index
         respond_to do |format|
           format.json do
-            render json: NpqApplicationSerializer.new(paginate(scope)).serializable_hash
+            render json: NPQApplicationSerializer.new(paginate(scope)).serializable_hash
           end
 
           format.csv do
-            render body: NpqApplicationCsvSerializer.new(scope).call
+            render body: NPQApplicationCsvSerializer.new(scope).call
           end
         end
       end
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -41,7 +41,7 @@ module Api
       end
 
       def access_scope
-        LeadProviderApiToken.all
+        LeadProviderApiToken.joins(cpd_lead_provider: [:npq_lead_provider])
       end
     end
   end
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -27,7 +27,17 @@ module Api
       end
 
       def scope
-        npq_lead_provider.npq_profiles.includes(:user, :npq_course)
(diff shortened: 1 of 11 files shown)


app/serializers/npq_application_csv_serializer.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require "csv"
+
+class NpqApplicationCsvSerializer
```
ethanmills, line 5:
```suggestion
class NPQApplicationCsvSerializer
```

app/serializers/npq_application_serializer.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class NpqApplicationSerializer
```
ethanmills, line 3:
```suggestion
class NPQApplicationSerializer
```

app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+module Api
+  module V1
+    class NPQApplicationsController < Api::ApiController
+      include ApiTokenAuthenticatable
+      include ApiPagination
+
+      def index
+        respond_to do |format|
+          format.json do
+            render json: NpqApplicationSerializer.new(paginate(scope)).serializable_hash
```
ethanmills, line 14:
```suggestion
            render json: NPQApplicationSerializer.new(paginate(scope)).serializable_hash
```

app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+          end
+
+          format.csv do
+            render body: NpqApplicationCsvSerializer.new(scope).call
```
ethanmills, line 18:
```suggestion
            render body: NPQApplicationCsvSerializer.new(scope).call
```

app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      def npq_lead_provider
+        current_api_token.cpd_lead_provider.npq_lead_provider
+      end
+
+      def scope
+        s = npq_lead_provider.npq_profiles.includes(:user, :npq_course)
+        s = s.where("updated_at > ?", Time.iso8601(updated_since_filter)) if updated_since_filter.present?
+        s
+      end
```
kicferk1, line 33: Full words please, `scope` over `s`

Maybe `applications_scope` or `query_scope`, or `applications` for the method name? We do have `access_scope` below, so just `scope` feels a bit off.

app/controllers/concerns/api_pagination.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module ApiPagination
+  extend ActiveSupport::Concern
+  include Pagy::Backend
+
+  included do
+  end
```
kicferk1, line 8: Do we need this empty block?

app/controllers/concerns/api_pagination.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module ApiPagination
+  extend ActiveSupport::Concern
+  include Pagy::Backend
```
kicferk1, line 5: Are there any places where we can reuse this? Maybe not in this PR, but if you could add a `todo` in them to use this concern it would make it easier to pick up for someone else later.

app/serializers/npq_application_csv_serializer.rb
```diff
@@ -0,0 +1,59 @@
...
+      teacher_reference_number
+      teacher_reference_number_validated
+      school_urn
+      headteacher_status
+      eligible_for_funding
+      funding_choice
+      course_id
+      course_name
+    ]
+  end
```
kicferk1, line 24: Can we call them `csv_headers`? For a second there I thought they are response headers.

app/serializers/npq_application_csv_serializer.rb
```diff
@@ -0,0 +1,59 @@
...
+
+private
+
+  def to_row(record)
+    [
+      record.id,
+      record.user_id,
+      record.user.full_name,
+      record.user.email,
+      true,
```
kicferk1, line 48: Are we sure email will always be validated? If not, can we have a todo comment here?

kicferk1 (commented):
Looks good overall, some small naming comments here and there :)

app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    private
+
+      def npq_lead_provider
+        current_api_token.cpd_lead_provider.npq_lead_provider
+      end
+
+      def scope
+        s = npq_lead_provider.npq_profiles.includes(:user, :npq_course)
```
ethanmills, line 30: npq_lead_provider can be nil here in theory, right?

app/controllers/api/v1/npq_applications_controller.rb
asmega: I will change method to `query_scope` and local variable to `scope`

app/controllers/concerns/api_pagination.rb
asmega: It can be used in other places but as you say for another PR. I don't really want to add a TODO as it doesn't seem important or pressing enough and the functionality is not degraded or missing

app/serializers/npq_application_csv_serializer.rb
asmega: I will rename to `csv_headers` and make it a private method

app/serializers/npq_application_csv_serializer.rb
asmega: Yes. I don't believe it is possible to complete the journey to submit an application until they have confirmed their email address on NPQ side

app/controllers/concerns/api_pagination.rb
ethanmills: If we get this merged I'll pick this up in my next PR, touching those controllers anyway

app/controllers/api/v1/npq_applications_controller.rb
asmega: i've made a small change so this is no longer possibly based on the `access_scope`

kicferk1 (approved):

PR #682, opened 2021-07-13
Update user serializer

xswoodsx:
### Context
[653](https://dfedigital.atlassian.net/jira/software/projects/CPDEL/boards/89?selectedIssue=CPDEL-653)

Engage and learn want to know whether a user has completed their registration before allowing them access to the service. 
This hasn't been implemented fully yet so we're automatically assigning the value of false for now. 

Adding registration completed attribute to user serializer
Fixing npq validation data factory that was breaking a test. 

### Review Checks

Run the app locally.
Send an api request using postman or insomnia and see what you get back. It should include registration completed: false for all users.


app/serializers/user_serializer.rb
```diff
@@ -51,6 +51,10 @@ class UserSerializer
     find_school_cohort(user)&.induction_programme_choice
   end
 
+  attributes :registration_completed do
+    false
```
ethanmills, line 55: Could you add a TODO against CPDRP-508 to set this to a real value please?

ethanmills (approved):
1 request for a TODO, looks good other than that

PR #688, opened 2021-07-14
add sandbox api token for NPQ

asmega:
### Context

- Ensure in sandbox env NPQ can talk to this app

- Add to seeding for sandbox hardcoded api token

### Guidance to review

- none

### Testing

- Run sandbox seeds
- Should see newly added token

### Review Checks

- not possible


kicferk1 (approved):
I don't think sandbox goes through the same cycle of destroying and re-creating its database, so I am not sure how necessary it is, but it probably won't hurt to have it.

asmega:
> I don't think sandbox goes through the same cycle of destroying and re-creating its database, so I am not sure how necessary it is, but it probably won't hurt to have it.

Sure. It scripted in such a way that it looks like it could and may be ran on sandbox. But yes agreed no harm in having.

PR #693, opened 2021-07-14
CPDRP-585: Only show active participants

ethanmills:
### Context
CPDRP-585
In a previous PR (#666 ) I added a status to participants. This change is to only display active participants

This is mostly about updating the scopes on User, School, and Lead Provider, then ensuring they're used correctly.
I read a good portion of the codebase in the course of this change, and noticed a few oddities. I'll leave comments for any unrelated changes I made along the way

### Guidance to review
Related changes in the first commit. Other things that needed fixing in the other two

### Testing
Ensuring existing tests pass, adding tests to make sure withdrawn ppts are not shown

### Review Checks

Functionality to withdraw ppts not yet implemented, so it would need to be done via the console.


app/forms/participant_mentor_form.rb
```diff
@@ -12,7 +12,9 @@ def mentor
   end
 
   def available_mentors
-    school.mentors.order(:full_name)
+    User.where(
+      id: school.mentor_profiles_for(Cohort.find(cohort_id)).select(:user_id),
+    ).order(:full_name)
```
kicferk1, line 17: Soo, this is not really something for this PR, but in general, I have a feeling most of the time when school users do an action, they really operate on a school_cohort. That's why we keep passing both schools and cohorts around. 

I wonder if we should shift our approach to suit that - have a bunch of things like profiles point at school cohorts. And if they need school or cohort, they can fetch it `through`.

If we do that, this could be a lot simpler. Something like `school_cohort.mentors.order(:full_name)`. Thoughts?

app/services/invite_schools.rb
```diff
@@ -92,12 +92,7 @@ def invite_to_beta(school_urns)
   def send_beta_chasers
     beta_school_ids = FeatureFlag.new(name: :induction_tutor_manage_participants).feature.selected_objects.pluck(:object_id)
     beta_schools = School.where(id: beta_school_ids)
-    beta_schools_without_participants = beta_schools.left_outer_joins(:early_career_teacher_profiles)
-                                                    .left_outer_joins(:mentor_profiles)
-                                                    .where(
-                                                      early_career_teacher_profiles: { id: nil },
-                                                      mentor_profiles: { id: nil },
-                                                    )
+    beta_schools_without_participants = beta_schools.where.not(id: ParticipantProfile.ecf.select(:school_id))
```
kicferk1, line 95: nice :)

spec/models/school_spec.rb
```diff
@@ -463,12 +435,104 @@
...
+      mentor_profile = create(:mentor_profile, cohort: cohort)
+
+      expect(school.mentor_profiles_for(cohort)).not_to include mentor_profile
+    end
+
+    it "does not include mentors" do
+      mentor_profile = create(:early_career_teacher_profile, school: school, cohort: cohort)
+
+      expect(school.mentor_profiles_for(cohort)).not_to include mentor_profile
+    end
```
kicferk1, line 536: Hehe, I think you mean 'it does not include ects' here ;) 

And `ect_profile` for the variable name.

spec/policies/participant_policy_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+require "rails_helper"
+
+RSpec.describe ParticipantPolicy, type: :policy do
+  subject { described_class.new(acting_user, user_under_test) }
+
+  %i[early_career_teacher_profile mentor_profile].each do |profile_type|
+    context "when the participant has #{profile_type}" do
+      # rubocop:disable Rails/SaveBang
+      let(:user_under_test) { create(profile_type).user }
+      # rubocop:enable Rails/SaveBang
```
kicferk1, line 12: Why do we need this rubocop stuff here?

app/forms/participant_mentor_form.rb
ethanmills: Yep, see https://dfedigital.atlassian.net/browse/CPDRP-545

spec/models/school_spec.rb
ethanmills: Good spot!

spec/policies/participant_policy_spec.rb
ethanmills: Eugh. So, it thinks this is an ActiveRecord#create, and wants me to use create!. It's because I'm not passing a symbol

kicferk1 (approved):

PR #700, opened 2021-07-15
CPDTP-234 Make token generation rake task apply to NPQ as well

timabell:
### Context

LeadProvider is actually EcfLeadProvider in disguise so the task as-was could only generate tokens for ECF providers. By moving up to the (new) parent entity of CpdLeadProvider we can now generate tokens for both ECF and NPQ.

Make token generation rake task apply to NPQ as well

### Guidance to review

:ship:

### Testing

Try on an NPQ provider, eg:

```
bundle exec rake lead_provider:generate_token "Church of England"
```

### DIFF
--- a/lib/tasks/lead_provider_token.rake
+++ b/lib/tasks/lead_provider_token.rake
@@ -8,20 +8,15 @@ namespace :lead_provider do
       "#{msg}\n"
     end
 
-    begin
-      lead_provider = LeadProvider.find(ARGV[1])
-      cpd_lead_provider = lead_provider.cpd_lead_provider
-    rescue StandardError
-      lead_provider = LeadProvider.find_by(name: ARGV[1])
-      cpd_lead_provider = lead_provider.cpd_lead_provider
-    end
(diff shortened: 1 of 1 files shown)


kicferk1 (approved):

lib/tasks/lead_provider_token.rake
```diff
@@ -8,20 +8,15 @@ namespace :lead_provider do
...
-    begin
-      lead_provider = LeadProvider.find(ARGV[1])
-      cpd_lead_provider = lead_provider.cpd_lead_provider
-    rescue StandardError
-      lead_provider = LeadProvider.find_by(name: ARGV[1])
-      cpd_lead_provider = lead_provider.cpd_lead_provider
-    end
+    name_or_id = ARGV[1]
+
+    lead_provider = CpdLeadProvider.find_by(id: name_or_id) || CpdLeadProvider.find_by(name: name_or_id)
```
plisovin, line 13: Smart 👍

PR #703, opened 2021-07-15
CPDRP-622 Prevent NPQ registration from blocking ECT/mentor registration

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/browse/CPDRP-622)
If user has registered for NPQ this prevents the same user from being added as a ECT/mentor.

When checking if the email already exists, only check against users that are already ECTs and Mentors so we can add NPQ registrants.

### Testing
A test user has been added with a `ParticipantProfile::NPQ` and email address `npq-registrant@example.com`
Sign-in as the `school-leader@example.com` user and attempt to add a participant with the same email address as the NPQ user and it should succeed.
Attempting to add any existing ECT or Mentor should fail.

### Review Checks

### DIFF
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -9,13 +9,6 @@ class ParticipantProfile < ApplicationRecord
     withdrawn: "withdrawn",
   }
 
-  scope :mentors, -> { where(type: Mentor.name) }
-  scope :ects, -> { where(type: ECT.name) }
-
-  scope :sparsity, -> { where(sparsity_uplift: true) }
-  scope :pupil_premium, -> { where(pupil_premium_uplift: true) }
-  scope :uplift, -> { sparsity.or(pupil_premium) }
-
   attr_reader :participant_type
 
   def ect?
--- a/app/policies/participant_policy.rb
+++ b/app/policies/participant_policy.rb
@@ -27,7 +27,8 @@ private
 
   def user_can_access?(participant)
     return false unless participant.participant?
-    return false if participant.participant_profile.withdrawn?
+    # TODO: is this correct? assumes we're only looking at mentors and ECTs?
+    return false if participant.participant_profiles.ecf.first.withdrawn?
 
     user.schools.include?(participant.school)
   end
--- a/app/policies/participant_policy.rb
+++ b/app/policies/participant_policy.rb
@@ -27,8 +27,8 @@ private
 
   def user_can_access?(participant)
     return false unless participant.participant?
-    # TODO: is this correct? assumes we're only looking at mentors and ECTs?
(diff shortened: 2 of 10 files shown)


app/policies/participant_policy.rb
```diff
@@ -27,7 +27,8 @@ def update?
 
   def user_can_access?(participant)
     return false unless participant.participant?
-    return false if participant.participant_profile.withdrawn?
+    # TODO: is this correct? assumes we're only looking at mentors and ECTs?
+    return false if participant.participant_profiles.ecf.first.withdrawn?
```
kicferk1, line 31: I think it probably works. Although I wonder what happens if the induction coordinator tries to access index page with a single withdrawn user and a few active ones. Possibly something scary. 

Does all of this logic make sense? I think not quite - we wrote the policy before adding NPQ profiles, so it is not aware of them. I would say for time being we probably should return false if the user is npq, just to be safe.

tonyheadford: Think that would then prevent us from doing what this PR is intended to fix though, because then if a user was a NPQ we wouldn't be able to administer their mentor profile if that makes sense?

tonyheadford: Agree that I think this area needs more thought at some point

kicferk1: Ah, true that. Perhaps what we want here is to check 'does this user have any active ect profiles? if not, then return false'?

kicferk1 (approved):

app/models/participant_profile.rb
```diff
@@ -9,13 +9,6 @@ class ParticipantProfile < ApplicationRecord
     withdrawn: "withdrawn",
   }
 
-  scope :mentors, -> { where(type: Mentor.name) }
-  scope :ects, -> { where(type: ECT.name) }
```
ethanmills, line 13: Why have these gone?

ethanmills: I think you need to rebase, this would fail tests on develop

ethanmills: D'oh, didn't see these were duplicated. Sorry!!

ethanmills:
have we considered how mentors who are also doing an NPQ will be displayed?

tonyheadford:
> have we considered how mentors who are also doing an NPQ will be displayed?
> 
> It looks like in some places they will appear twice: https://github.com/DFE-Digital/early-careers-framework/blob/CPDRP-622-cannot_add_participant_if_already_in_NPQ/app/controllers/admin/participants_controller.rb#L13 not sure if this is right or not

I haven't in this PR but this sounds like it could be a different story? @BroiSatse have you thought how this should be?  If we're not qualifying which type of participant we're looking for then perhaps it is valid to return all matching profiles?

PR #704, opened 2021-07-16
CPDTP-227 update provider declarations api to v0.3

blackrat:
### Context

This changes updated the documentation and code versions to support v0.3 of the API specification.
In particular, it adds in validation and documentation for other schema variants (NPQ and "retained-x" changes), as well
as adding in the new `course_type` field.

Update documentation and swagger API with the new schemas
Update validation schema to support these
Update code to continue working with the new shape of the schema and introduce a new 400 error type when the baseline structure is incorrect.

Note that this only supports the existing declaration_types. It is not mean to be a complete implementation of the API and types not currently supported will return a 422 error.

### Review Checks


etc/schema/0.3/participant_declarations/create/request_schema.json
```diff
@@ -0,0 +1,188 @@
...
+    },
+    "npq": {
+      "courses": {
+        "enum": [
+          "npq-leading-teaching",
+          "npq-leading-teaching-development",
+          "npq-leading-behaviour-culture",
+          "npq-headship",
+          "npq-senior-leadership",
+          "npq-executive-leadership"
```
mattonfoot, line 132: in the npq-applications API the LP receives a uuid and a name for the course. should there be consitency with this one, to maybe share these identifiers in the npq-applications API ?

mattonfoot (approved):

kicferk1 (approved):

PR #706, opened 2021-07-16
different feature flag for NPQ applications api

asmega:
### Context

- At the moment NPQ applications api endpoint shares the same feature flag as the participant data
- We want to split these so the features can be toggled independently

- Use different feature flag for NPQ applications api endpoint

### Guidance to review

- none

### Testing

- Toggle feature flag on `npq_applications_api`
- Should be able to hit the endpoint
- Toggle feature flag off `npq_applications_api`
- Should no longer be able to hit api endpoint

### Review Checks

- Not sure how feature flag are set in the review apps

### DIFF
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -17,6 +17,7 @@ class FeatureFlag
 
   # Short-lived feature flags
   TEMPORARY_FEATURE_FLAGS = %i[
+    npq_applications_api
     participant_data_api
     induction_tutor_manage_participants
     admin_participants
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -54,7 +54,7 @@ Rails.application.routes.draw do
       resources :users, only: %i[index create]
       resources :dqt_records, only: :show, path: "dqt-records"
       resources :participant_validation, only: :show, path: "participant-validation"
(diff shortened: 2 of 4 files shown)


kicferk1:
Do we actually want to hide this behind a feature flag? For participants we do, cause we only want them to work in sandbox. Why is this the case for npq applications?

To set the feature flag in review app, our readme says...
> To activate a new feature you can run `Feature.create!(name: 'rate_limiting', active: true)`.

which I guess you can do with Rails console. I think it is persisted despite db resets.

asmega:
> Do we actually want to hide this behind a feature flag? For participants we do, cause we only want them to work in sandbox. Why is this the case for npq applications?
> 
> To set the feature flag in review app, our readme says...
> 
> > To activate a new feature you can run `Feature.create!(name: 'rate_limiting', active: true)`.
> 
> which I guess you can do with Rails console. I think it is persisted despite db resets.

I can just delete all the feature flags instead if that is preferred

kicferk1:
> I can just delete all the feature flags instead if that is preferred?

If we don't have an environment where we want it disabled (say, if we want to enable it in production), and if we don't want the ability to turn it off with a switch of a flag... then I would say we don't need the flag.

I just don't know for sure whether the 'if's I wrote are true. Are they?

asmega:
ok, ill close this and raise another PR with the flag deleted

PR #709, opened 2021-07-16
CPDRP-545: Use school_cohort on profiles instead of school and cohort

kicferk1:
### Context

https://dfedigital.atlassian.net/browse/CPDRP-545

1. Add school_cohort association to profiles
2. That includes NPQ profiles (before merging it I should probably make sure that all NPQ profiles in prod include a school...)
3. Populate the school_cohort data on profiles with school / cohort data
4. Remove the old school / cohort fields from profiles
5. ...And fix everything in the app and in the tests that breaks because of that.

### Review Checks

It should be a refactor that doesn't change the existing behaviour.


kicferk1:
> 2. That includes NPQ profiles (before merging it I should probably make sure that all NPQ profiles in prod include a school...)

`ParticipantProfile::NPQ.where(school: nil).count` gave me 0 in prod, presumably because somewhere (in npq registration maybe?) we validate that the school with a given urn exists.

ethanmills:
What about NPQ ppts who have a school that doesn't have a school_cohort?

ethanmills (changes requested):
I don't think NPQ schools are guaranteed to have a school_cohort. I think we will lose data here

app/models/npq_validation_data.rb
```diff
@@ -29,7 +29,12 @@ class NPQValidationData < ApplicationRecord
 
   def update_participant_profile
     profile = ParticipantProfile::NPQ.find_or_initialize_by(id: id)
-    profile.school = School.find_by(urn: school_urn)
+    school = School.find_by(urn: school_urn)
+    cohort = profile.school_cohort&.cohort || Cohort.current
+    school_cohort = SchoolCohort.find_or_create_by!(school: school, cohort: cohort) do |new_school_cohort|
+      new_school_cohort.update!(induction_programme_choice: "not_yet_known")
```
ethanmills, line 35: This is a legacy option, not really used anywhere. What do these schools see?

ethanmills: This will prevent these schools from choosing provision currently

kicferk1:
I think the migration would take over the time that paas gives us, sadly. So we will need to do it in a few more steps.

PR #712, opened 2021-07-16
Bugfix npq ppts in admin views

ethanmills:
Two in one here (terrible I know)
- Only show ECF participants in admin schools view (NPQ participant don't have the correct methods/attributes and things break)
- Fix scope to work for NPQ participants - this has been breaking the /users API, and was also the direct cause of the above issue


kicferk1 (approved):
more conflicts, yay 😭

PR #714, opened 2021-07-16
CPDRP-637 Change email check to allow NPQ registrants

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/browse/CPDRP-637)
When nominating an induction tutor, allow an existing user who is an NPQ registrant
See also #703 

Change the email in use check to allow NPQ registrants.

### Review Checks

### DIFF
--- a/app/forms/nominate_induction_tutor_form.rb
+++ b/app/forms/nominate_induction_tutor_form.rb
@@ -19,9 +19,7 @@ class NominateInductionTutorForm
   end
 
   def email_already_taken?
-    User.left_outer_joins(:induction_coordinator_profile)
-        .where(induction_coordinator_profile: { id: nil })
-        .exists?(email: email)
+    ParticipantProfile.ecf.joins(:user).where(user: { email: email }).any?
   end
 
   def name_different?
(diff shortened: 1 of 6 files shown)


app/forms/nominate_induction_tutor_form.rb
```diff
@@ -19,9 +19,7 @@ def school
   end
 
   def email_already_taken?
-    User.left_outer_joins(:induction_coordinator_profile)
-        .where(induction_coordinator_profile: { id: nil })
-        .exists?(email: email)
+    ParticipantProfile.ecf.joins(:user).where(user: { email: email }).any?
```
ethanmills, line 22: This theoretically means we could end up with someone who is both and admin/lead provider and a coordinator. I don't think it will ever happen, but something to consider

ethanmills (commented):
Looks like you've got some failing tests

app/forms/nominate_induction_tutor_form.rb
tonyheadford: Yes, I know, not sure if it's something to really worry about?
I think ideally we'd have all the profiles inheriting from some common base which would simplify things.

ethanmills (approved):

PR #715, opened 2021-07-16
Add school cohorts step 1: add the field and shadow it on the models

kicferk1:
### Context

https://dfedigital.atlassian.net/browse/CPDRP-545

Add optional school_cohort field to participant profiles. Shadow it on ECT and Mentor models. 

I will need to get it deployed, and run a script to populate the school_cohort. Something like `ParticipantProfile.ecf.each { |record| record.save }` should do the trick. And I should do it in every environment.


ethanmills (approved):
This is much less scary

PR #716, opened 2021-07-16
Record whodunnit for console users in production

ethanmills:
Taken from papertrail docs. We are doing lots of console work at the moment, we should keep track of it

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -15,7 +15,7 @@ Rails.application.configure do
   config.eager_load = true
 
   # Full error reports are disabled and caching is turned on.
-  config.consider_all_requests_local       = false
+  config.consider_all_requests_local = false
   config.action_controller.perform_caching = true
 
   # Ensures that a master key has been made available in either ENV["RAILS_MASTER_KEY"]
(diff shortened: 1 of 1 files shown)


kicferk1 (approved):
Hah, too late, I have already modified all participant records!

So that saves like 20k records in papertrail I guess 😂

config/environments/production.rb
```diff
@@ -142,4 +142,20 @@
...
+                                   until email.present?
+                                     # rubocop:disable Rails/Output
+                                     puts "Enter your email address for PaperTrail"
+                                     # rubocop:enable Rails/Output
+                                     email = gets.chomp
+                                   end
+                                   email
+                                 end
+    }
+  end
```
kicferk1, line 160: Can we not do that in development?

Or perhaps in any non-prod environment?

jamesrose: Not a _huge_ fan of this. I think it might get a bit annoying. We can trace back SSH sessions to users with PaaS events.

ethanmills: @kicferk1 this is production only a the moment.

@jamesrose there have been a few times where I have been trying to figure out how a record got into the state it was, and only managed by going through Jira to find the support ticket. It's worth noting this will only pop up when you start trying to modify records

ethanmills: discussed

PR #721, opened 2021-07-19
CPDTP-324 Change rake task for payment breakdown to generate csv output

blackrat:
### Context

We were asked if our current rake task could be made to generate a csv file as a demonstrator. This updates the output and adds in the CSV headings and body to do so.

Change structure of Rake task to make it easier to maintain, by creating a small class which will perform calls into the calculator for those portions of the calculations that we need. Note that this may actually drive how the calculator works in the future, so is probably a good thing even if it's just used in the rake task.
Add in aggregation scopes for ECTs and mentors.
Change the large single table into smaller tables which are more like the proposed layout for Payment Breakdown in any case.
Add in CSV generation. This is almost certainly going to be temporary code, but does actually interrogate the database and pulls back real numbers.

### Review Checks

### DIFF
--- a/lib/tasks/payment_breakdown.rake
+++ b/lib/tasks/payment_breakdown.rake
@@ -2,6 +2,7 @@
 
 require "payment_calculator/ecf/payment_calculation"
 require "payment_calculator/ecf/uplift_calculation"
+require "tasks/payment_breakdown"
 require "terminal-table"
 
 include ActiveSupport::NumberHelper
@@ -9,11 +10,6 @@ include ActiveSupport::NumberHelper
 namespace :payment_calculation do
   desc "run payment calculator for a given lead provider"
   task breakdown: :environment do
-    logger = Logger.new($stdout)
-    logger.formatter = proc do |_severity, _datetime, _progname, msg|
-      "#{msg}\n"
-    end
-
     lead_provider = begin
                       LeadProvider.find(ARGV[1])
                     rescue StandardError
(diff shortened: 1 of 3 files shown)


lib/tasks/payment_breakdown.rake
```diff
@@ -23,63 +19,35 @@ namespace :payment_calculation do
...
-    uplift_payment = breakdown[:uplift].each_with_object({}) do |(type, value), hash|
-      hash[type] = "£#{number_to_delimited(value.to_i)}"
+  desc "generate csv payment calculations for all lead providers"
+  task csv: :environment do
+    filename = (ARGV[1] || "output.csv")
+    lead_providers = LeadProvider.all
+    CSV.open(filename, "wb") do |csv|
+      csv << Tasks::PaymentBreakdown.new(contract: lead_providers.first.call_off_contract, total_participants: 0, uplift_participants: 0).csv_headings
+      lead_providers.each do |lead_provider|
+        contract = lead_provider.call_off_contract
```
plisovin, line 39: One DRY suggestion for these 6 lines of code as it's repeated in both tasks:

```
def shared(lead_provider)
  total_participants = ParticipantEventAggregator.call({ lead_provider: lead_provider }).to_i
  uplift_participants = ParticipantUpliftAggregator.call({ lead_provider: lead_provider }).to_i
  total_ects = (ARGV[2].present? ? ARGV[2].to_i / 2 : ParticipantEventAggregator.call({ lead_provider: lead_provider, started: :count_active_ects_for_lead_provider }))
  total_mentors = (ARGV[2].present? ? ARGV[2].to_i - ARGV[2].to_i / 2 : ParticipantEventAggregator.call({ lead_provider: lead_provider, started: :count_active_mentors_for_lead_provider }))
  Tasks::PaymentBreakdown.new(contract: lead_provider.call_off_contract, total_participants: total_participants, uplift_participants: uplift_participants, total_ects: total_ects, total_mentors: total_mentors)
end
```

and then:

```
shared(lead_provider).to_table
```

```
lead_providers.each do |lead_provider|
  csv << shared(lead_provider).csv_body
end
```

blackrat: Actually, I'm going to make those blocks slightly different. The 'csv' option shouldn't take in ARGV[2], since it should always be hitting the database.

blackrat: Also. Having a method in the rake task gives it global scope, which isn't a good thing, since all rake tasks are loaded and can redefine. For reference:
https://kevinjalbert.com/defined_methods-in-rake-tasks-you-re-gonna-have-a-bad-time/

kicferk1 (approved):

PR #723, opened 2021-07-20
CPDTP-302 Convert some of the turnip feature tests to pure rspec

timabell:
### Context

We were using the BDD style feature files during isolated development of the payment calculator library to help with communication with the business analyst and to give us something clear to agree on. Now that the calculator is being connected up as part of the bigger system this doesn't make so much sense to discuss in isolation. These tests and steps are adding a development overhead that we don't currently think is worth the investment so we are removing these specs in favour of equivalent test coverage with pure rspec tests.

I was going to do this as one bigger PR to eliminate all the feature files in one go but the task keeps getting interrupted and we want smaller PRs anyway, so here's a PR that removes the first couple.

https://dfedigital.atlassian.net/browse/CPDTP-302

* Remove some of the turnip `.feature` files
* Ensure there is equivalent test coverage in the `_spec.rb` tests to replace them.

### Guidance to review

:ship:


spec/lib/payment_calculator/ecf/contract/service_fee_calculations_spec.rb
```diff
@@ -7,22 +7,62 @@ class ModuleTestHarness
...
+      expected_per_participant: 560.00,
+    },
+    {
+      # service fee variation 2
+      set_up_fee: 150_000,
+      band_a_per_participant_price: 1_400.00, # unchanged from 1
+      expected_total: 970_000.00,
+      expected_monthly: 33_448.28,
+      expected_per_participant: 485.00, # reduced from 1 by service fee component
+    },
```
kicferk1, line 47: I am not sure I like those code comments.

Do you think it would make sense to make each of those examples a `describe` or a `context`, which lets you specify some more useful descriptions?

timabell: Yeah I agree it's a little smelly. As you can tell I was struggling a bit to express intent while also keeping clean structures

kicferk1: Yeah. 
I would much prefer something like 
```
context "when conditions for variation 2 are met" # maybe say what they are explicitly, I don't actually know
  let(:set_up_fee) { 150_000 }
  let(:band_a_per_participant_price) { 1_400.00 }

  it "correctly reduces expected value per participant" do
    run_test_with_expected_values(expected_total: 970_000.00, expected_monthly: 33_448.28, expected_per_participant: 485.00)
  end
end
```

Something like that?

timabell: I'll try it as describes and see how it feels....

timabell: How about this? https://github.com/DFE-Digital/early-careers-framework/pull/723/commits/b253495aed3fc3dfd7d37344c5d3bf915ab993d9

kicferk1 (approved):

spec/lib/payment_calculator/ecf/contract/service_fee_calculations_spec.rb
kicferk1: Yeah, I think it's clear enough for my taste :)

PR #726, opened 2021-07-20
Allow admin users to impersonate induction coordinators

jamesrose:
### Context

- Allow admin users to impersonate induction coordinators
- I haven't placed any limitations on the impersonation side as to the type of user it can impersonate. That is, it will accept any user ID that is not the impersonator or another admin user. It should just work.
- The dependent gem is relatively straight forward but is well maintained and tested with devise.

Screenshot 2021-07-20 at 12 03 40
Screenshot 2021-07-20 at 12 03 44

### DIFF
--- a/app/controllers/admin/impersonates_controller.rb
+++ b/app/controllers/admin/impersonates_controller.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+module Admin
+  class ImpersonatesController < Admin::BaseController
+    skip_after_action :verify_policy_scoped
+
+    before_action :load_user, only: [:create]
+    before_action :check_self_impersonation, only: [:create]
+    before_action :check_admin_user_impersonation, only: [:create]
+    before_action { authorize @user, policy_class: ImpersonationPolicy }
+
+    def create
+      impersonate_user(@user)
+
+      redirect_to after_sign_in_path_for(@user)
+    end
+
+    def destroy
+      stop_impersonating_user
+      redirect_to after_sign_in_path_for(current_user)
+    end
+
+  private
+
+    def load_user
+      @user = User.find(params[:impersonated_user_id])
+    end
+
+    def check_self_impersonation
+      if params[:impersonated_user_id] == current_user.id.to_s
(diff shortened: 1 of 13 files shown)


app/controllers/application_controller.rb
```diff
@@ -23,7 +25,7 @@ def previous_url_for_cookies_page
   end
 
   def after_sign_in_path_for(user)
-    stored_location_for(user) || helpers.profile_dashboard_path(user)
```
jamesrose, line 26: This is a bit tenuous but `user` in this context was always the impersonator. It didn't break anything so I rolled with it.

jamesrose: Thinking about it perhaps I need to set `pundit_user` in the application controller like I did in the impersonation one.

```ruby
def pundit_user
  true_user
end
```

Will give it a go.

app/controllers/admin/impersonates_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+        redirect_to(after_sign_in_path_for(current_user))
+      end
+    end
+
+    def check_admin_user_impersonation
+      if User.find(params[:impersonated_user_id]).admin?
+        flash[:warning] = "You cannot impersonate another admin user"
+        redirect_to(after_sign_in_path_for(current_user))
+      end
+    end
```
ethanmills, line 37: Could we move this to the policy?

app/policies/privacy_policy_policy.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class PrivacyPolicyPolicy < ApplicationPolicy
```
ethanmills, line 3: Is this used anywhere? Also, the semantics seem off here, I think this should be `create?` for PrivacyPolicy::Acceptance, and the check should be `user == record.user`

asmega (approved):

kicferk1 (approved):
I like it - and it seems if need be we can extend it to impersonate other users quite easily too :)

ethanmills (approved):

PR #729, opened 2021-07-20
Spin up a very basic version of multipage flow for 2020

kicferk1:
### Context

More of an exploratory experience - figuring out how far we can push the multistep form, what we can save, what screens make sense, etc. I would not merge it, the PR is mostly for the review app. 

Sign in as `school-leader@example.com`.
Go to `/schools/999999-example-school/year-2020/add` in the url.

Known things to address:
1. 'Opt out' somewhere, perhaps the first page.
2. Multiple participants.
3. Multiple schools.
3. Saving data correctly.
4. Hide behind a feature flag?
4. Tests.

### DIFF
--- a/app/views/schools/year2020/confirm.html.erb
+++ b/app/views/schools/year2020/confirm.html.erb
@@ -1,0 +1,33 @@
+<% content_for :title, "Confirm these details" %>
+
+<% content_for :before_content, govuk_back_link(text: "Back", href: back_link_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">Confirm these details</h1>
+
+    <dl class="govuk-summary-list govuk-!-margin-bottom-9">
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          Details
+        </dt>
+        <dd class="govuk-summary-list__value">
+          <div><%= year_2020_form.full_name %></div>
+          <div><%= year_2020_form.email %></div>
+        </dd>
+        <dd class="govuk-summary-list__actions">
+          <%# govuk_link_to issue: https://github.com/DFE-Digital/govuk-components/issues/193 %>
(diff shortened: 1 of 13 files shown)


app/views/schools/year2020/confirm.html.erb
```diff
@@ -0,0 +1,33 @@
...
+        <dd class="govuk-summary-list__actions">
+          <%# govuk_link_to issue: https://github.com/DFE-Digital/govuk-components/issues/193 %>
+          <%= govuk_link_to({ action: :show, step: "details" }, button: false) do %>
+            Change <span class="govuk-visually-hidden">personal details</span>
+          <% end %>
+        </dd>
+      </div>
+    </dl>
+
+<!--    QQQQ Add another participant, display more participants -->
```
timabell, line 27: what does `QQQQ` mean?

spec/forms/schools/add_participant_form_spec.rb
```diff
@@ -57,7 +57,7 @@
     end
 
     context "when the email is not already in use" do
-      it "returns true" do
+      it "returns false" do
```
timabell, line 60: this doesn't make much more sense to me (in relation to the body of the test), and I don't understand why the logic has reversed

kicferk1:
```
describe "email_already_taken?" do
  context "when the email is not already in use" do
    it "returns false"
      ...
    end
  end
end
```
makes a lot of sense to me. The method checking if an email is already taken should return false when the email is not already in use. 

The description was simply incorrect - the logic hasn't changed (and the test contents haven't either).

app/views/schools/year2020/confirm.html.erb
kicferk1: It means "resolve it before you merge", I use it to make sure I can quickly find those in my code :P

kicferk1:
Closing in favor of https://github.com/DFE-Digital/early-careers-framework/pull/743

PR #730, opened 2021-07-20
BUGFIX: don't show delivery partners for challenged partnerships

ethanmills:
image
We shouldn't see this when the partnership is challenged


kicferk1 (commented):
Unit test?

kicferk1 (approved):

PR #731, opened 2021-07-20
Cpdtp 259 set up participant declarations to support npq

blackrat:
### Context

We're going to have a single endpoint for all participant-declarations for both NPQ and ECF data.
The recent refactor to allow the same token to be used for both means that we should have the common (i.e. CpdLeadProvider) as our stored FK to support both from the same table.

Also, the term "course_identifier" is more descriptive than "course_type" so the parameter is being switched to support that name.

The places where we will need to refer to the lead_provider and npq_lead_provider will be at the "contracts" level, so
changes apart from to the contracts parts need to reflect this.

The unused ParticipationRecord can now be deleted too, since that refers to ECTprofiles and the lead_provider, and changing it doesn't seem to make sense now that it's no longer in use.

Change controller to inject the cpd_lead_provider_id into the params which it takes from the current_user coming in on the API Token.
Update CpdLeadProvider to have an association to ParticipantDeclarations and drop and the participation_record reference from the LeadProvider.
Change the calculation aggregator to use the cpd_lead_provider to reference the records (scoping will still be done for the record types, to ensure no conflict between them).
Since the calculator uses the contract information without caring where it comes from, no changes need to be made there, only for the aggregators.
Add in and use a declaration_type for ECF declarations to create the correct records (Note. The code only supports ECF declarations at the moment. This PR is to enable them to be extended by changing the provider relationships.)
Add in the optional "evidence_held" field to the database and set the default type to enable STI.
Update existing participant_declarations records with the cpd_lead_provider reference.
Update tests and API documentation to reflect the changes to type
Regenerate swagger API docs

### Review Checks


kicferk1:
Without looking at the code, I reckon the app is broken because of some association errors. 

Take a look at the logs from it: https://kibana.logit.io/s/51151c79-e17d-4407-be3a-01960f3f65ff/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-24h,to:now))&_a=(columns:!(message),filters:!(),index:'8ac115c0-aac1-11e8-88ea-0383c11b333a',interval:auto,query:(language:kuery,query:%22ecf-review-pr-731%22),sort:!())

> /usr/local/bundle/gems/activesupport-6.1.3.2/lib/active_support/core_ext/hash/keys.rb:52:in `block in assert_valid_keys': Unknown key: :through. Valid keys are: :class_name, :anonymous_class, :primary_key, :foreign_key, :dependent, :validate, :inverse_of, :strict_loading, :autosave, :required, :touch, :polymorphic, :counter_cache, :optional, :default (ArgumentError)

I reckon https://github.com/DFE-Digital/early-careers-framework/pull/731/files#diff-04348b9a98f008637aec694e3f9765f2423c04160be56b71963c6fccce6ff34cR4 is what causes it.

app/models/participant_declaration/npq.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class ParticipantDeclaration::NPQ < ParticipantDeclaration
+  belongs_to :npq_lead_provider, through: :cpd_lead_provider
```
kicferk1, line 4: This probably should be 'has_one` - `belongs_to` in rails land means the id of the owner lives on this object.

So, `school_cohort` `belongs_to` `school` means the school cohort has school id on it. And in this sense, `through` can't work.

I reckon that's why the app is not starting.

app/services/record_participant_declaration.rb
```diff
@@ -75,9 +75,15 @@ def user_profile
     end
   end
 
+  def declaration_type
+    if early_career_teacher? || mentor?
+      ParticipantDeclaration::ECF
+    end
+  end
```
kicferk1, line 82: And otherwise it's `nil`? That doesn't look right...

db/migrate/20210720065105_change_participant_declarations_to_common_base_class.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class ChangeParticipantDeclarationsToCommonBaseClass < ActiveRecord::Migration[6.1]
+  disable_ddl_transaction!
+
+  def change
+    safety_assured do
+      change_table :participant_declarations, bulk: true do |t|
+        t.string :course_identifier, null: true
+        t.string :evidence_held, null: true
+        t.string :type, default: "ParticipantDeclaration::ECF"
```
kicferk1, line 11: Warning: this will mean anyone who tries making an abstract `ParticipantDeclaration` will end up with an ECF one, no real way of stopping it. I made that mistake with tokens.

If we don't want that, uhh, how many declarations are there in live system? We might just get away with a migration setting the type on all of them.

spec/models/participation_record_spec.rb
```diff
@@ -1,58 +0,0 @@
-# frozen_string_literal: true
-
-require "rails_helper"
-require "aasm/rspec"
-
-RSpec.describe ParticipationRecord, type: :model do
```
kicferk1, line 6: I don't see anything replacing this test.

Is that because we are getting rid of this functionality? If so... Do we have anything else using `aasm/rspec`?

app/models/participant_declaration/npq.rb
blackrat: Oops. Quite right.

blackrat: Wasn't actually adding NPQ into this PR. That shouldn't be there.

spec/models/participation_record_spec.rb
blackrat: No. It's all been pulled out.

app/services/record_participant_declaration.rb
blackrat: Current code only support ECF. We've got another ticket to add in NPQ and I want to refactor this when we add it.

kicferk1: Alright, I wrote 'that doesn't look right...' cause I saw NPQ declarations and thought they would fit here :D

kicferk1 (approved):
We are limited to something around a minute to run all the migrations, but I think we are good here - I don't suspect there ate thousands of participant declarations / events floating around in any envorionment.

db/migrate/20210720065105_change_participant_declarations_to_common_base_class.rb
blackrat: Don't really like enabling people to make Abstract instances. When the next bit goes in, it's going to have restriction and validations to prevent that. It'll either have to be ECF or NPQ. Base ones won't make sense and will fail unless the correct course types are set.

PR #732, opened 2021-07-20
CPDRP-573: admins delete participants

ethanmills:
### Context
CPDRP-573
Give admins the ability to "delete" participants. What this is actually doing is marking the selected profile as withdrawn. This only applies to ECF ppts at the moment, waiting to hear the rules for NPQ ppts.

- Cherrypick a @BroiSatse commit to change from /admin/participants/:user_id to admin/participants/:profile_id
- Add functionality to withdraw ppts

### Testing
Request specs and cypress tests

### Review Checks

Sign in as admin@example.com, look at school with urn 999999


app/controllers/admin/participants_controller.rb
```diff
@@ -2,26 +2,34 @@
...
+      render :destroy_success
     end
 
   private
 
     def load_participant
-      @participant = User.is_participant.find(params[:id])
-      authorize @participant, policy_class: ParticipantPolicy
+      @participant_profile = ParticipantProfile.find(params[:id])
+      authorize @participant_profile, policy_class: ParticipantProfilePolicy
```
kicferk1, line 32: Do we need to specify the class here?

app/views/admin/participants/show.html.erb
```diff
@@ -10,37 +10,42 @@
...
   </div>
   <div class="govuk-grid-column-one-half">
     <table class="govuk-table">
       <tbody class="govuk-table__body">
-        <% if @participant.early_career_teacher? %>
+        <% if @participant_profile.ect? %>
           <tr class="govuk-table__row">
             <th scope="row" class="govuk-table__header">Mentor</th>
-            <td class="govuk-table__cell"><%= @participant.early_career_teacher_profile.mentor&.full_name&.presence || 'Not yet assigned' %></td>
+            <td class="govuk-table__cell"><%= @participant_profile.mentor&.full_name || 'Not yet assigned' %></td>
```
kicferk1, line 32: Empty strings in ruby are truthy, hence I imagine the `presence`.

Although, I guess the `full_name` cannot be an empty string, so I think we are safe. Is that true?

spec/requests/admin/participants_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    it "renders the show template" do
+      get "/admin/participants/#{mentor_profile.id}"
+      expect(response).to render_template("admin/participants/show")
+    end
+
+    it "shows the correct participant" do
+      get "/admin/participants/#{ect_profile.id}"
+      expect(response.body).to include(CGI.escapeHTML(ect_profile.user.full_name))
+      expect(response.body).to include(CGI.escapeHTML(mentor_profile.user.full_name))
+    end
```
kicferk1, line 44: What are we testing for here?

Can we get an unrelated ect and show that they are not shown?

app/controllers/admin/participants_controller.rb
ethanmills: FYI @BroiSatse I think these are on the cherry picked commit. Will change these, but you might get a few conflicts

app/views/admin/participants/show.html.erb
ethanmills: I think if the name were blank we wouldn't want to show "not yet assigned", but no, it should never be blank. Will leave this as-is I think

app/controllers/admin/participants_controller.rb
ethanmills: Yes, this is needed
`Pundit::NotDefinedError: unable to find policy `ParticipantProfile::MentorPolicy`

app/views/admin/participants/show.html.erb
```diff
@@ -10,37 +10,42 @@
...
         </tr>
       </tbody>
     </table>
   </div>
 </div>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_link_to "Delete participant", remove_admin_participant_path %>
+  </div>
+</div>
```
kicferk1, line 51: Should we make it a button? Link at the bottom of the page looks a bit odd to my eye.

ethanmills: 🤷 just following the designs

ethanmills: Feedback to Robin if interested

kicferk1 (approved):

app/policies/participant_profile_policy.rb
```diff
@@ -1,6 +1,16 @@
 # frozen_string_literal: true
 
 class ParticipantProfilePolicy < ApplicationPolicy
+  def show?
+    admin_only
+  end
+
+  def destroy?
+    admin? && (record.ect? || record.mentor?)
```
timabell, line 9: It's not obvious to me why the restriction to ect and mentor is needed. Perhaps a code comment to explain?

ethanmills:
@timabell commit reworded

app/serializers/participant_serializer.rb
```diff
@@ -30,10 +30,8 @@ class ParticipantSerializer
   end
 
   attributes :participant_type do |user|
-    if user.early_career_teacher_profile
-      :ect
-    else
-      :mentor
+    if participant_active?(user)
+      user.early_career_teacher? ? :ect : :mentor
```
kicferk1, line 34: I wonder if for our sanity's sake we should make it something like
```
if user.early_career_teacher?
  :ect
elsif user.mentor?
  :mentor
end
```
This should still return `nil` now that you added the active restriction to mentor and ect profiles on the user, and should return `nil` for npq users too, if somehow they end up here. As opposed to returning `:mentor` for npq users.

app/policies/participant_profile_policy.rb
kicferk1: That brings up another interesting point. Will we show the link to delete the user on purely npq users that are visible from participants page? I think we will :(

kicferk1 (commented):
Still makes sense to me.

app/serializers/participant_serializer.rb
ethanmills: Yep, that sounds more robust

app/policies/participant_profile_policy.rb
ethanmills: Currently there are no links to the show page for an NPQ participant, and if you try to visit it by entering the ID then it throws and error about not having a cohort. That said, I should hide it for when it does work 🙂

ethanmills: Now done

kicferk1 (approved):

app/serializers/participant_serializer.rb
```diff
@@ -30,10 +30,8 @@ class ParticipantSerializer
   end
 
   attributes :participant_type do |user|
-    if user.early_career_teacher_profile
-      :ect
-    else
-      :mentor
+    if participant_active?(user)
```
blackrat: Now this is being repeated lots (and there seems to be use of the plural form when only one attribute is being listed, would it not be better to create a private method to wrapper it? 

e.g. 
```
class << self
 def active_participant_attribute(attr, &blk)
    attribute attr do |user|
      if participant_active?(user)
        blk.call
      end
    end
  end
end
```

then you could dry up the code with:
```
active_participant_attribute :email do |user|
  user.email
end

active_participant_attribute :full_name do |user|
    user.full_name
end
```

Comment (ethanmills) [reply app/serializers/participant_serializer.rb]: I like this, but can't get it working, some variation on 

```
NoMethodError:
  undefined method `active_participant_attribute' for ParticipantSerializer:Class
```

I'm going to go as-is for now, then come back this this if I have time

blackrat: Did you miss the "class << self"? I did a test refactor on it, and it went from 55 lines to .......

55 lines. :)

PR #733, opened 2021-07-21
Scope API tokens for access to participants API

asmega:
### Context

- This fix is in response to https://sentry.io/organizations/dfe-bat/issues/2519946117/?environment=sandbox&project=5748989&query=is%3Aunresolved
- This prevents API tokens not associated to an ECF provider such as NPQ only lead providers from accessing the participants API

- Scope API token to only those with an (ECF) `LeadProvider` to be able to access the participants API

### Guidance to review

- Existing API access should not be broken and there should be continuity of service

### Testing

- Create or use an existing `CpdLeadProvder` that has a `NPQLeadProvider` but not an (ECF) `LeadProvider`
- This will mimic a NPQ only lead provider
- Create a `LeadProviderApiToken` associated to this `CpdLeadProvider`
- Attempt to access the participants api endpoint
- Should receive a 403 and not an unexpected error

### Review Checks

- Probably easier locally, but the same instructions for local testing should apply here

### DIFF
--- a/app/controllers/api/v1/participants_controller.rb
+++ b/app/controllers/api/v1/participants_controller.rb
@@ -24,7 +24,7 @@ module Api
     private
 
       def access_scope
-        LeadProviderApiToken.all
+        LeadProviderApiToken.joins(cpd_lead_provider: [:lead_provider])
       end
 
       def to_csv(hash)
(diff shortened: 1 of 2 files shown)


kicferk1 (approved):
Is there any place that does it the other way around? As in, an endpoint where we should make sure an npq lead provider exists?

asmega:
> Is there any place that does it the other way around? As in, an endpoint where we should make sure an npq lead provider exists?

Yes, this exists. Please see https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/controllers/api/v1/npq_applications_controller.rb#L44

jamesrose (approved):

PR #734, opened 2021-07-21
Cpdrp 538 admin participant page

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-538

NOTE: The tests are still missing - I am working to add them. It might still beusefull to review it as is.


app/controllers/admin/participants/validations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    private
+
+      def load_decision
+        @participant_profile = ParticipantProfile.find(params[:id])
+        authorize @participant_profile, :validate?, policy_class: ParticipantProfilePolicy
+
+        @decision = @participant_profile.validation_decision(params[:step])
+      end
+
+      def t_scope
```
ethanmills, line 31: `t` as in translation?

app/controllers/admin/schools/participants_controller.rb
```diff
@@ -4,19 +4,18 @@ module Admin
...
 
     def index
-      @participants = @school.active_ecf_participants.order(:full_name).includes(
-        early_career_teacher_profile: [:mentor, school_cohort: %i[school cohort]],
-        mentor_profile: [school_cohort: %i[school cohort]],
-      )
+      @participant_profiles = policy_scope(ParticipantProfile::Ecf, policy_scope_class: ParticipantProfilePolicy::Scope)
+        .where(school_cohort_id: SchoolCohort.where(school: school).select(:id))
+        .includes(:user)
+        .order("users.full_name")
```
ethanmills, line 12: Would it be nicer to do something like 

```
school_participant_profiles = school.active_ecf_participant_profiles
                                    .includes(:user)
                                    .order("users.full_name")
@participant_profiles = policy_scope(school_participant_profiles, policy_scope_class: ParticipantProfilePolicy::Scope)
```
?

app/models/participant_profile/ecf.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ParticipantProfile::Ecf < ParticipantProfile
```
ethanmills, line 3: 🎉

app/models/profile_validation_decision.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class ProfileValidationDecision < ApplicationRecord
+  has_paper_trail
+
+  validates :note, presence: true
```
ethanmills, line 6: Is the note required?

app/policies/participant_profile_policy.rb
```diff
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
 class ParticipantProfilePolicy < ApplicationPolicy
+  def show?
+    return true if admin?
```
ethanmills, line 5: This is just `admin_only`

app/policies/participant_profile_policy.rb
```diff
@@ -1,6 +1,14 @@
...
+  end
+
+  def validate?
+    admin?
```
ethanmills, line 9: This is `admin_only`

spec/components/admin/participants/details_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        t(:ect, scope: "schools.participants.type"),
+        mentor_profile.user.full_name,
+        participant_profile.school.name,
+        participant_profile.cohort.display_name,
+      )
+    end
+  end
+
+  # context "for Mentor profile" do
+  #   let!(:participant_profile) { create :participant_profile, :ect }
```
ethanmills, line 23: Still working on this?

spec/requests/admin/schools/participants_spec.rb
```diff
@@ -5,10 +5,9 @@
 RSpec.describe "Admin::Schools::Participants", type: :request do
   let(:admin_user) { create(:user, :admin) }
   let(:school) { create(:school) }
-  let(:school_cohort) { create(:school_cohort, school: school) }
-  let!(:ect_user) { create(:user, :early_career_teacher, school_cohort: school_cohort) }
-  let!(:mentor_user) { create(:user, :mentor, school_cohort: school_cohort) }
-  let!(:unrelated_user) { create(:user, :mentor) }
+  let!(:ect_profile) { create :participant_profile, :ect, school: school }
+  let!(:mentor_profile) { create :participant_profile, :mentor, school: school }
```
ethanmills, line 9: This doesn't look right?

app/components/admin/participants/details/details.html+ect_pending.erb
```diff
@@ -0,0 +1,30 @@
+<h2 class="govuk-heading-m">Details</h2>
```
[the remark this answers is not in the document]
kicferk1: Also, we have `ect_pending`, `mentor_pending`, `npq`, `npq_pending`. Are we missing something? :P

app/components/admin/participants/details.rb
```diff
@@ -0,0 +1,16 @@
...
+        @profile = profile
+        @variant = [profile.participant_type, (:pending if profile.pending?)].compact.join("_")
+      end
+
+    private
+
+      attr_reader :profile
+    end
+  end
+end
```
kicferk1, line 16: Where can I see html for that? Is there some magic that means `@variant` just renders the right template?

app/components/admin/participants/table_row.rb
```diff
@@ -13,6 +13,12 @@ def initialize(profile:)
 
       attr_reader :profile
       delegate :school, to: :profile
+
+      def validation_status
+        return { text: "Not ready", colour: "grey" } unless profile.npq?
+
+        { text: "Pending", colour: "yellow" }
+      end
```
kicferk1, line 21: Is this used?

app/components/admin/participants/validation_tasks.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <li>
+      <table data-test="admin-participants-validation-tasks" class="govuk-table">
+        <thead  class="nhsuk-table__head">
+          <tr class="govuk-table__row">
+            <th class="govuk-table__header" colspan="2"><%= t :name, scope: "schools.participants.validations.#{profile.participant_type}.#{step}" %></td>
+          </tr>
+        </thead>
+        <tbody>
+          <tr class="govuk-table__row">
+            <td class="govuk-!-width-full">
```
kicferk1, line 14: huh, what does this do in a responsive table?

app/components/admin/participants/validation_tasks.html.erb
```diff
@@ -0,0 +1,29 @@
...
+              <%= govuk_link_to(
+                    t(:link_text, scope: "schools.participants.validations.#{profile.participant_type}.#{step}"),
+                    validation_step_admin_participant_path(profile, step)
+                  )
+              %>
+            </td>
+            <td>
+              <%= govuk_tag tag_attributes(step) %>
+            </td>
```
kicferk1, line 23: Does putting it in a table (rather than in a flexbox of our own making) cause the tag to break over multiple lines for unlucky viewport sizes?

app/controllers/admin/schools/participants_controller.rb
```diff
@@ -4,19 +4,18 @@ module Admin
...
     skip_after_action :verify_policy_scoped
-    before_action :set_school
 
     def index
-      @participants = @school.active_ecf_participants.order(:full_name).includes(
-        early_career_teacher_profile: [:mentor, school_cohort: %i[school cohort]],
-        mentor_profile: [school_cohort: %i[school cohort]],
-      )
+      @participant_profiles = policy_scope(ParticipantProfile::Ecf, policy_scope_class: ParticipantProfilePolicy::Scope)
+        .where(school_cohort_id: SchoolCohort.where(school: school).select(:id))
```
kicferk1, line 10: `ParticipantProfile::Ecf` has a school on it, so maybe we could write `where(school: school)`? Or `where(schools: school` depending on how mean rails feels today? :P

app/models/participant_profile.rb
```diff
@@ -2,14 +2,29 @@
 
 class ParticipantProfile < ApplicationRecord
   belongs_to :user
+  # belongs_to :school, optional: true
```
kicferk1, line 5: burn it with fire :D 

or you know, just remove commented out code ;)

app/models/participant_profile.rb
```diff
@@ -2,14 +2,29 @@
...
+  scope :npqs, -> { where(type: NPQ.name) }
+
+  scope :sparsity, -> { where(sparsity_uplift: true) }
+  scope :pupil_premium, -> { where(pupil_premium_uplift: true) }
+  scope :uplift, -> { sparsity.or(pupil_premium) }
+
   attr_reader :participant_type
 
+  class_attribute :validation_steps
+  self.validation_steps = []
```
kicferk1, line 26: Should this be a class method rather than class attribute? Are we planning to dynamically change it?

app/models/profile_validation_decision.rb
```diff
@@ -0,0 +1,8 @@
...
+  validates :approved, inclusion: { in: [true, false] }
```
kicferk1, line 7: What other values can there be for what I assume is a boolean?

app/views/admin/participants/validations/show.html.erb
```diff
@@ -0,0 +1,17 @@
+<% content_for :title, "Participant #{t :name, scope: t_scope}" %>
+<% content_for :before_content, govuk_back_link(text: "Back", href: admin_participant_path(@participant_profile)) %>
+
+<span class="govuk-caption-l"><%= @participant_profile.user.full_name %>
+<h1 class="govuk-heading-l"><%= t :header, scope: t_scope %></h1>
+
+<p><%= t :description, scope: t_scope %></p>
+
+<%= form_for @decision, url: {}, method: :post do |f| %>
```
kicferk1, line 9: What does `url: {}` do?

app/views/admin/participants/validations/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <%= f.govuk_radio_buttons_fieldset :approved, inline: true, legend: { text: "Change this decision", tag: 'h2', size: 'm' } do %>
+    <%= f.govuk_radio_button :approved, 1, label: { text: 'Approved' }, checked: @decision.approved %>
+    <%= f.govuk_radio_button :approved, 0, label: { text: 'Rejected' }, checked: @decision.approved == false %>
```
kicferk1, line 12: Let's use double quotes everywhere please :)

config/routes.rb
```diff
@@ -146,7 +146,14 @@
...
     end
 
-    resources :participants, only: %i[show index]
+    resources :participants, only: %i[show index] do
+      member do
+        scope path: "validations", controller: "participants/validations" do
+          get ":step", action: :show, as: :validation_step
+          post ":step", action: :update
+        end
+      end
```
kicferk1, line 155: would this be easy to make resourceful?

spec/components/admin/participants/details_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        participant_profile.user.email,
+        t(:ect, scope: "schools.participants.type"),
+        mentor_profile.user.full_name,
+        participant_profile.school.name,
+        participant_profile.cohort.display_name,
+      )
+    end
+  end
+
+  # context "for Mentor profile" do
```
kicferk1, line 22: I know the PR says tests will come later, but just to ping, this needs to be uncommented probably :)

spec/components/admin/participants/validation_status_tag_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  component { described_class.new profile: participant_profile }
+
+  context "for ECT profile" do
+    let!(:participant_profile) { create :participant_profile, :ect }
+
+    it { is_expected.to have_selector(".govuk-tag.govuk-tag--grey", text: "Not ready") }
+  end
+
+  context "for Mentor profile" do
+    let!(:participant_profile) { create :participant_profile, :ect }
```
kicferk1, line 13: I reckon you need a different profile here (mentor one)

kicferk1 (commented):
so many conflicts 😢

app/models/participant_profile/ecf.rb
kicferk1: 🎉

app/models/participant_profile/npq.rb
```diff
@@ -6,10 +6,21 @@ class ParticipantProfile::NPQ < ParticipantProfile
 
   has_one :validation_data, class_name: "NPQValidationData", foreign_key: :id, dependent: :destroy
 
+  self.validation_steps = %i[identity decision]
```
ethanmills, line 9: `.freeze` ?

app/policies/participant_profile_policy.rb
BroiSatse: For now yes, but we also display participant profiles to SIC (which currently either does not use policy or use the old `ParticipantPolicy` which authorises User models. I have my doubts about `admin_only` method, I think `admin?` would be a much better name for it. `admin_only || record.some_condition` is a massive conceptual riddle

app/components/admin/participants/details/details.html+ect_pending.erb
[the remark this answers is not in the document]
BroiSatse: Unfortunately, this convention is pretty out of our hands - it is how `versions` of the ViewComponent works - during compilation phase they search for files matching the name of the component (either in the same folder or in the folder with same name) and then they split it on dots, take the second part (normally `html` or `text`), split that on `+` sign and whatever follows is added to `context_lookup.versions` array. What's even more annoying is that they use that name to define public method `call_#{version}` so you're not allowed to use just any characters in the version...

I've spent quite a while searching for a more explicit way to pass/override template to be used for the component, the close second place was having a separate component per each type and one component that selects which one to render...

As per missing templates - yes, we're missing `ect` and `mentor`. As for now, all the ect and mentor profiles are pending validation - the exact requirement for ECT validations are quite different from NPQ, and since NPQ is a prioirty for now I have not added the other validations yet. So, for `ect` and `mentor` profiles, we'll be always displaying `ect_pending` and `mentor_pending` for now.

app/components/admin/participants/details.rb
BroiSatse: Yup, that's the magic with that weird file name. Honestly, I do not like it at all. I might issue a PR to ViewComponents to allow `component_name/#{variant}.html.erb` which feels so much better..

app/components/admin/participants/table_row.rb
BroiSatse: Ah, no. It was replaced by the extracted component! :)

app/components/admin/participants/validation_tasks.html.erb
BroiSatse: It forces this column to take as much space as possible, effectively pushing the tag as far to the right as possible. Tables are ultra-weird when it comes to sizing...

app/views/admin/participants/validations/show.html.erb
BroiSatse: It causes the form to submit to the current path. The given hash is passed to `url_for`, which reverse_merge it with `url_options` extracted from the current route.

There's one interesting probelm I hit with it before: using `url: ''` theoretically should have the same effect, as submitting to the current path is a default for html's form tag. However - it breaks CRSF token generation as the form is given a token for path '', which is then verified agains the actual path...

app/views/admin/participants/validations/show.html.erb
BroiSatse: Uh, shall we add erb-lint?

config/routes.rb
BroiSatse: Not 100% resourceful. The relation between create and update are a bit bleak here - you cannot really create or change validation_step - those are defined by hardcoded array depending on the type of profile.

Potentially, each step could be seen as a separate resource without a create action. However, I had a bit of a hard time imagining how that would translate to resource. Theoretically, each form submission is not an update of the record but rather a completely new decision (and we will later on display the history of decisions). For now, I've decided to use papertrail for that, but I'm really not happy about this. Actually, event_log would be much better match here....

app/models/participant_profile/npq.rb
BroiSatse: Ah, good idea! One note though - in not so distant future those steps will need to take a few additional options like `on` to specify that given step decisions are to be stored against another object. On one hand it could live on that other record, but I think I'd prefer the whole validation logic to be encapsulated in a single place.

app/components/admin/participants/validation_tasks.html.erb
BroiSatse: IIRC html tables never wraps - but it might just happily go out of the screen.

spec/requests/admin/schools/participants_spec.rb
BroiSatse: Anything apart from `ect_profile` not being used, I don't see it! 00

app/components/admin/participants/details/details.html+npq_pending.erb
```diff
@@ -0,0 +1,50 @@
...
+      <th scope="row" class="govuk-table__header">DOB entered</th>
+      <td class="govuk-table__cell"><%= profile.validation_data&.date_of_birth %></td>
+    </tr>
+    <tr class="govuk-table__row">
+      <th scope="row" class="govuk-table__header">TRN entered</th>
+      <td class="govuk-table__cell"><%= profile.validation_data&.teacher_reference_number %></td>
+    </tr>
+    <tr class="govuk-table__row">
+      <th scope="row" class="govuk-table__header">NINo entered</th>
+      <td class="govuk-table__cell"><%#= profile.validation_data&.nino %></td>
```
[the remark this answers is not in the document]
BroiSatse: Yeah, we're expected to display NI number here - but we don't have it! The data is coming from register-npq and NI has not been added to NPQProfile. I thought to resolve it a bit later (as this field would not live on participant profile any way - it is another field for TeacherProfile).

app/controllers/admin/participants/validations_controller.rb
BroiSatse: Yes, might rename to i18n_scope? Or maybe `step_i18n_scope` to be more clear of its purpose?

app/controllers/admin/schools/participants_controller.rb
BroiSatse: TBH, I think I'll be fighting with `active_ecf_participant_profiles` a bit :) Since this is a subset of `ecf_participants` association, we now need to know that there are 2 associations going through the same foreign key that might both return the same records but which preloading logic is different. Another reason to limit number of associations...

app/components/admin/participants/details/details.html+ect_pending.erb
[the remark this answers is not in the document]
kicferk1: Uhh, the magic and limitations surrounding `versions` sounds like a very good idea to have one component per user type and an extra component which uses a switch to display the right one. 

Saves anyone else the need to dive deep into ViewComponent versions and being horrified at what they discover :P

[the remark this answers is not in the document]
kicferk1: And with your component test setup, we could easily unit test that the right component gets rendered for the right user. Which I don't think is particularly easy to verify with `versions` (or am I wrong and it is very easy?)

app/views/admin/participants/validations/show.html.erb
kicferk1: Maybe. Perhaps not in this PR ;)

app/models/profile_validation_decision.rb
ethanmills: Does this mean that it's not possible to leave a note without making a decision?

app/models/participant_profile/npq.rb
```diff
@@ -6,10 +6,21 @@ class ParticipantProfile::NPQ < ParticipantProfile
...
     true
   end
 
+  def approved?
+    validation_decision(:decision).approved?
+  end
+
+  def rejected?
+    decision = validation_decision(:decision)
+    decision.persisted? && !decision.approved?
```
ethanmills, line 21: This is nullable, right? This treats "no decision yet" as rejection, surely?

app/controllers/admin/participants/validations_controller.rb
ethanmills: I think that would be clearer, yes 🙂

spec/requests/admin/schools/participants_spec.rb
ethanmills: Shouldn't these take a `school_cohort`, not a school?

app/models/participant_profile/npq.rb
BroiSatse: "no decision yet" mean that the decision is not yet persisted :)

app/models/profile_validation_decision.rb
BroiSatse: Good question - I think best way to find out is to make it required and wait for feedback :P

spec/requests/admin/schools/participants_spec.rb
BroiSatse: Ah, after the refactor. Hmm, good question - both profiles do have `has_one :school, through: :school_cohort` so my guess is that by passing `school` instead of cohort FactoryBot will create a cohort (with a school) and then assigns the school through that cohort, effectively leaving us with some orphaned school records. It's not really an issue for the test, but would be nice to avoid this. :)

app/components/admin/participants/details/details.html+ect_pending.erb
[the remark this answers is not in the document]
BroiSatse: It is easy enough to test that the correct content is displayed with standard content matchers. The one annoying thing is the sheer number of files in the director - 12 files just to handle that switch without any difference in behaviour...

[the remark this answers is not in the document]
BroiSatse: That being said, I think I'd like to bite the bullet here as I'd like this merged sooner than later. I am ahppy to refactor in next PR, but amount of conflicts arising here every day is really getting in the way...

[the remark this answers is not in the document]
tonyheadford: Agree with @kicferk1 I would be much happier seeing some extra components rather than being forced into this

app/models/participant_profile/npq.rb
ethanmills: If I wanted to leave a comment without selecting a decision - then persist it - this would show "rejected"?

BroiSatse: For now, you cannot leave a comment without making decision

app/components/admin/participants/details/details.html+ect_pending.erb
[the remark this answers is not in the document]
BroiSatse: Ok, this is now changed to use mutliple sub classes instead of variants.

app/controllers/admin/schools/participants_controller.rb
BroiSatse: Interestingly enough - tried this but it didn't work as rails tried to filter on `school` column which does not exist...

app/models/profile_validation_decision.rb
BroiSatse: It rejects `nil`. As for now it is not possible to leave a comment without making a decision.

app/models/participant_profile.rb
BroiSatse: I think I went for a class_attribute to be able to define id declarative in a single line rather than as an override to a class methods (in 3 lines). I don't think there's much reason to choose one over the other, but If you feel strongly about this I am happy to change it to a method...

kicferk1: Main reason I find putting it in an attribute weird is that it makes it easily editable (I think), which, well, we probably do not want, especially on a class level.

I don't feel strongly about it though.

app/components/admin/participants/validation_tasks.rb
```diff
@@ -0,0 +1,34 @@
...
+module Admin
+  module Participants
+    class ValidationTasks < BaseComponent
+      def initialize(profile:)
+        @profile = profile
+      end
+
+      def render?
+        profile.npq?
+      end
```
kicferk1, line 12: Can we move this check to the parent component?

spec/cypress/app_commands/scenarios/admin/school_participants.rb
```diff
@@ -44,3 +44,10 @@
                   user: FactoryBot.create(:user, full_name: "Unrelated ect"),
                   school_cohort: another_school_cohort,
                   created_at: Date.parse("29/12/2020")
+
+npq_profile = FactoryBot.create :participant_profile, :npq, validation_data: nil
+
+FactoryBot.create :npq_validation_data,
+                  id: npq_profile.id,
+                  user: FactoryBot.create(:user, full_name: "Natalie Portman Quebec"),
+                  created_at: Date.parse("19/09/2019")
```
kicferk1, line 53: Do we need to pin date of birth to avoid randomness in percy?

app/components/admin/participants/validation_tasks.rb
BroiSatse: Ah, you mean the view? I think this is a perfect example when `render?` makes sense - this component is to be rendered for all the types n the future, but currently it only makes sense for `npq` profiles. When the feature is fully complete that method will be gone and `render?` method will be completely removed - which is nice that we have a single place to be removed...

spec/cypress/app_commands/scenarios/admin/school_participants.rb
BroiSatse: Yup!

app/components/admin/participants/validation_tasks.rb
kicferk1: Alright, sure.

kicferk1 (approved):

ethanmills (approved):

PR #735, opened 2021-07-22
CPDTP-329 Add in model validations for participant_declarations

blackrat:
### Context

Now that this is using STI we can put in proper course type validations for NPQ and ECF courses.

Inclusion validations to make sure that the correct course identifier is used for each declaration type. For NPQ it reads from the NPQCourse identifiers held in the database, so that if more courses are added, we can validate against those without having to make code changes too.
Presence validation to ensure that the required fields are passed to the model.

Factories and tests updated to ensure that correct course_identifiers are linked to the correct profile types.

### Review Checks

### DIFF
--- a/app/models/participant_declaration.rb
+++ b/app/models/participant_declaration.rb
@@ -5,6 +5,9 @@ class ParticipantDeclaration < ApplicationRecord
   belongs_to :cpd_lead_provider
   belongs_to :user
 
+  validates :course_identifier, inclusion: { in: :valid_courses }
+  validates :course_identifier, :user, :cpd_lead_provider, :declaration_date, :declaration_type, presence: true
+
   # Helper scopes
   scope :for_lead_provider, ->(cpd_lead_provider) { where(cpd_lead_provider: cpd_lead_provider) }
   scope :for_declaration, ->(declaration_type) { where(declaration_type: declaration_type) }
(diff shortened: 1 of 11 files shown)


spec/models/participant_declaration_spec.rb
```diff
@@ -16,11 +16,14 @@
...
                cpd_lead_provider: call_off_contract.lead_provider.cpd_lead_provider,
-               profile_type: profile_type)
+               profile_type: profile_type,
+               course_identifier: course_identifier)
       end
 
       context "for mentor was created" do
         let(:profile_type) { :mentor_profile }
+        let(:course_identifier) { "ecf-mentor" }
+        let(:course) { :ecf_mentor }
```
kicferk1, line 26: Is `course` used? Doesn't look like it.

spec/models/participant_declaration_spec.rb
```diff
@@ -29,6 +32,7 @@
 
       context "for early career teacher was created" do
         let(:profile_type) { :early_career_teacher_profile }
+        let(:course_identifier) { "ecf-induction" }
```
kicferk1, line 35: Same with `course_identifier` actually

blackrat: > Same with `course_identifier` actually

`course_identifier` is used in the create factory (line 20)

spec/models/participant_declaration_spec.rb
blackrat: Don't think it is in this one. I'll pull it.

kicferk1 (approved):

app/models/participant_declaration.rb
```diff
@@ -5,6 +5,9 @@ class ParticipantDeclaration < ApplicationRecord
   belongs_to :cpd_lead_provider
   belongs_to :user
 
+  validates :course_identifier, inclusion: { in: :valid_courses }
```
ethanmills, line 8: Do we need an `allow_blank: true` here to prevent duplicates?

Issue #336, opened 2021-05-10
Cohorts Validation has an illegal LocalJumpError when running with coverage


blackrat:
### What did you do? (steps to reproduce)

Ran the rspec test suite with coverage

### What happened? (actual results)
```
The test failed with:
Admin::Suppliers::DeliveryPartners POST /admin/suppliers/new/delivery-partner/choose-cohorts displays an error when no cohorts are selected for a lead provider
     Failure/Error: break

     LocalJumpError:
       break from proc-closure
     # ./app/forms/delivery_partner_form.rb:86:in 'block in cohorts_validation'
```

ethanmills:
I suspect this is an issue with your setup. Nobody else has encountered this as far as I'm aware, and it hasn't come up in CI

blackrat:
It's a definite issue since you shouldn't break inside a block. changing it to a return will fix it, since it's just wants to record an error and return.

kicferk1:
```
  def cohorts_validation
    # Ensure all selected lead providers have at least one selected cohort
    # This is indicated by the presence of a provider relationship for that lead provider
    lead_provider_ids&.filter(&:present?)&.each do |lead_provider_id|
      unless chosen_provider_relationships.pluck(:lead_provider_id).include?(lead_provider_id)
        errors.add(:provider_relationship_hashes, :blank, message: "Choose at least one cohort for every lead provider")
        break
      end
    end
  end
```
is the code it probably talks about. Feel free to refactor it not to use `break`, or to use it better. Although truth be told, the code looks reasonable to me - if we ever find a cohort that's missing, we stop looking. I think `break` refers to the `each do`, is there some subtle ruby magic that I am missing? Should we be using `return` instead? :P

tonyheadford:
Not sure that `return` will help, I think it's still possible to get that error if you `return` from a proc

blackrat:
Both break and return have different issues when used inside procs or lambdas. Best thing is to avoid them. Simplest fix is to drop the break line and switch the .each to .any?

I've done a quick test and refactor. All tests are passing still. Also double verified that what I thought about errors.add (and any?) were true.

PR #736, opened 2021-07-22
Issue#336 Stop the cohorts_validation using break from within a validation proc

blackrat:
### Context

Fixes issue #336
Breaks the coverage tests since a break inside a proc is an illegal instruction.

- Switching each to any? and dropping the break will stop after the first error is added, which was the intent.

errors.add returns the errors object, which is truthy, so will stop the loop and only add it once.

### Review Checks

### DIFF
--- a/app/forms/delivery_partner_form.rb
+++ b/app/forms/delivery_partner_form.rb
@@ -75,10 +75,18 @@ class DeliveryPartnerForm
     end
   end
 
+  def current_lead_provider_ids
+    lead_provider_ids&.filter(&:present?)
+  end
+
+  def chosen_provider_relationships_lead_provider_ids
+    chosen_provider_relationships.pluck(:lead_provider_id)
+  end
+
 private
 
   def lead_providers_validation
-    unless lead_provider_ids&.filter(&:present?)&.any?
+    unless current_lead_provider_ids&.any?
       errors.add(:lead_provider_ids, :blank, message: "Choose at least one")
     end
   end
@@ -86,10 +94,8 @@ private
   def cohorts_validation
     # Ensure all selected lead providers have at least one selected cohort
     # This is indicated by the presence of a provider relationship for that lead provider
-    lead_provider_ids&.filter(&:present?)&.any? do |lead_provider_id|
-      unless chosen_provider_relationships.pluck(:lead_provider_id).include?(lead_provider_id)
-        errors.add(:provider_relationship_hashes, :blank, message: "Choose at least one cohort for every lead provider")
-      end
+    if (current_lead_provider_ids - chosen_provider_relationships_lead_provider_ids).any?
+      errors.add(:provider_relationship_hashes, :blank, message: "Choose at least one cohort for every lead provider")
     end
   end
 end
--- a/app/forms/delivery_partner_form.rb
+++ b/app/forms/delivery_partner_form.rb
@@ -86,10 +86,9 @@ private
   def cohorts_validation
     # Ensure all selected lead providers have at least one selected cohort
     # This is indicated by the presence of a provider relationship for that lead provider
-    lead_provider_ids&.filter(&:present?)&.each do |lead_provider_id|
+    lead_provider_ids&.filter(&:present?)&.any? do |lead_provider_id|
       unless chosen_provider_relationships.pluck(:lead_provider_id).include?(lead_provider_id)
         errors.add(:provider_relationship_hashes, :blank, message: "Choose at least one cohort for every lead provider")
-        break
       end
     end
   end


app/forms/delivery_partner_form.rb
```diff
@@ -86,10 +86,9 @@ def lead_providers_validation
   def cohorts_validation
     # Ensure all selected lead providers have at least one selected cohort
     # This is indicated by the presence of a provider relationship for that lead provider
-    lead_provider_ids&.filter(&:present?)&.each do |lead_provider_id|
+    lead_provider_ids&.filter(&:present?)&.any? do |lead_provider_id|
       unless chosen_provider_relationships.pluck(:lead_provider_id).include?(lead_provider_id)
         errors.add(:provider_relationship_hashes, :blank, message: "Choose at least one cohort for every lead provider")
-        break
```
kicferk1, line 92: What happens if we have have two `lead_provider_ids `?

Do we get duplicate error?

kicferk1: Also, in principle I dislike side effects in functions like `map`, `filter`, and yes, `any` - `each` is the 'principled' functional array method that is designed to allow for those.

Maybe a slightly more semantically correct way of going about it would be
```
cohorts_selected =  lead_provider_ids&.filter(&:present?)&.all? do | lead_provider_id | 
  chosen_provider_relationships.pluck(:lead_provider_id).include?(lead_provider_id)}
end

errors.add(:provider_relationship_hashes, :blank, message: "Choose at least one cohort for every lead provider") unless cohorts_selected
```

Thoughts?

ethanmills: > Do we get duplicate error?
I would assume that `any?` short-circuits and so prevents that. I prefer your suggestion, however

blackrat: The former stops as soon as it hits the condition, the latter runs them all. I prefer the former as closer to the original. 
If you wanted to have a "true" at the end of the block to make it less of a side-effect.

timabell: I can see the argument for `any` being more of a functional list/comprehension type thing so having procedural stuff in it being a bit odd, but given it currently breaks the coverage I'd probably take it as is and then polish later if we feel like it.

The use of `all` will likely be less performant than `any` as it has to run for the whole set rather than exit early.

timabell: how about using `first` and then checking for whether that returned anything....

timabell: I think the use of `errors.add` to generate a `true` for the `unless` is a bit hard to follow, I'll try and make something I like better....

kicferk1: Performance would be my last concern here, we are talking about order of magnitude of '6 lead providers' 😂

timabell: Excuse the deleted comments, I had to play with the code to really grok this.

I agree with @kicferk1 on the suggested change, sorry @blackrat !

blackrat: I dislike the variable usage. I'm more inclined to pull that into a different method and refer to it that way. Thought?

kicferk1: I don't mind putting the code checking if there is a cohort selected for each LP into its own method.

blackrat: Actually, I had some time to put more thought into this. It was a quick first thing that worked with minimal code, but in thinking about it, it just comes down to an array comparison. If there are any elements in the first array that aren't in the second element, then we have an error. i.e. if (A-B) is not empty add the error in, otherwise it's fine.

app/forms/delivery_partner_form.rb
```diff
@@ -86,10 +86,10 @@ def lead_providers_validation
...
     # Ensure all selected lead providers have at least one selected cohort
     # This is indicated by the presence of a provider relationship for that lead provider
-    lead_provider_ids&.filter(&:present?)&.any? do |lead_provider_id|
-      unless chosen_provider_relationships.pluck(:lead_provider_id).include?(lead_provider_id)
-        errors.add(:provider_relationship_hashes, :blank, message: "Choose at least one cohort for every lead provider")
-      end
+    # Boils down to a set comparison of Arrays of lead_provider_ids
+    # i.e. (A-B) returns all elements in A not included in B, so if there are any, that a validation failure.
+    if (lead_provider_ids&.filter(&:present?)-chosen_provider_relationships.pluck(:lead_provider_id)).any?
+      errors.add(:provider_relationship_hashes, :blank, message: "Choose at least one cohort for every lead provider")
```
kicferk1, line 92: There are now 4 lines of code comments here, which I doubt anyone will read in full.

I reckon introducing some local variables with expressive names will do more for code maintainability than code comments.
```
selected_lead_provider_ids = lead_provider_ids&.filter(&:present?)
lead_providers_ids_with_cohort = leachosen_provider_relationships.pluck(:lead_provider_id)
if (selected_lead_provider_ids - lead_providers_ids_with_cohort).any?
  ...
```
That said, I think we are spending more time on it than it's worth :P

kicferk1 (approved):

app/forms/delivery_partner_form.rb
blackrat: I've extracted the `lead_provider_ids&filter(&:present?)`into a method since it's used in more than one place. The new comment was more to show the rationale behind it, and I'm not sure how valuable the original comment is. My initial thought was to extract into a method with a name close to that comment description. You're right that we're spending a long time over a tiny issue though. Working code replacing broken code was the desire here.

kicferk1 (approved):

timabell (approved):

app/forms/delivery_partner_form.rb
ethanmills: Set comparison, perhaps, but it's possible to have multiple copies of the lead provider in chosen_provider_relationships

PR #737, opened 2021-07-22
Correction to login spec name

timabell:
### Context

Noticed this while working on adding a finance user type.

Correction to login spec name

This now matches some of the other tests and is less misleading

### Guidance to review

:ship:


ethanmills (approved):
we've still got some inconsistency with one user type using the generic dashboard. For another time 🙂

PR #738, opened 2021-07-22
CPDTP-339 Add a new finance user type and give them a default URL

timabell:
### Context

These users are DfE internal users intended to be able to view payment information.

https://dfedigital.atlassian.net/browse/CPDTP-339

* Add a new finance user type and give them default URL that they will be redirected to on login
* Add a seed finance user
* Add a rake task for creating finance users (via a service class)

This patch is largely copied from the existing admin user code.

### Guidance to review

* Rather than testing the rake task with all the stdout redirection that involves, the logic for the rake task has been pushed into a service class and that is tested instead.

### Testing

```bash
bundle exec rails users:finance:create["Marge Simpson",marge@example.com]
bundle exec rails console
> User.find_by(email: "marge@example.com").finance?
...
true
```

Sign in to https://ecf-review-pr-738.london.cloudapps.digital/ as `finance@example.com` and see the redirect to https://ecf-review-pr-738.london.cloudapps.digital/finance/lead_providers which 404's as expected until the next PR lands.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -213,6 +213,10 @@ Rails.application.routes.draw do
     resources :induction_coordinators, only: %i[index edit update], path: "induction-coordinators"
   end
 
+  namespace :finance do
+    resources :lead_providers, only: %i[index]
+  end
+
   namespace :schools do
     resources :dashboard, controller: :dashboard, only: %i[index show], path: "/", param: :school_id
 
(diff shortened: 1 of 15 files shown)


config/routes.rb
```diff
@@ -213,6 +213,10 @@
     resources :induction_coordinators, only: %i[index edit update], path: "induction-coordinators"
   end
 
+  namespace :finance do
+    resources :lead_providers, only: %i[index]
+  end
```
kicferk1, line 218: Without a controller and a view this will either break or do something very default for Rails. Are we happy with that?

timabell: That's correct, @plisovin is following up with another PR to fill that in

kicferk1 (approved):

PR #739, opened 2021-07-22
CPDTP-348 Create lead provider index page for finance users

plisovin:
### Context

Finance users should be able to land on a page showing list of providers, the follow up work will allow them to click on any of them and get price breakdown

https://dfedigital.atlassian.net/browse/CPDTP-348

Add index page for lead providers and ability for finance users to display them

Selection_20210723-01-6a67e6a67e6a67e

### Testing

Create finance user
Login to the web app

### DIFF
--- a/app/controllers/finance/lead_providers_controller.rb
+++ b/app/controllers/finance/lead_providers_controller.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class Finance::LeadProvidersController < Finance::BaseController
+  def index
+    @lead_providers = LeadProvider.all
+  end
+end
--- a/spec/policies/finance_profile_policy_spec.rb
+++ b/spec/policies/finance_profile_policy_spec.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe FinanceProfilePolicy, type: :policy do
+  subject { described_class.new(user, finance_profile) }
+  let(:finance_profile) { create(:finance_profile) }
+
+  context "being a finance" do
+    let(:user) { create(:user, :finance) }
+
+    it { is_expected.to permit_action(:index) }
+    it { is_expected.to forbid_new_and_create_actions }
+    it { is_expected.to forbid_edit_and_update_actions }
+  end
+
+  context "not being a finance" do
+    let(:user) { create(:user) }
+
+    it { is_expected.to forbid_new_and_create_actions }
+    it { is_expected.to forbid_edit_and_update_actions }
+    it { is_expected.to forbid_action(:destroy) }
+    it { is_expected.to forbid_action(:index) }
+  end
+end
--- a/spec/requests/finance/lead_providers_controller_spec.rb
+++ b/spec/requests/finance/lead_providers_controller_spec.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Lead Providers for Finance users", type: :request do
+  let(:finance_user) { create(:user, :finance) }
+  let!(:lead_providers) { create_list(:lead_provider, 5) }
+
+  before do
+    sign_in finance_user
+  end
+
+  describe "GET /finance/lead-providers" do
+    it "renders the index template" do
+      get "/finance/lead-providers"
+
+      expect(response).to render_template("finance/lead_providers/index")
+      assigns(:lead_providers).should eq(lead_providers)
+    end
+  end
+end
--- a/app/controllers/finance/base_controller.rb
+++ b/app/controllers/finance/base_controller.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class Finance::BaseController < ApplicationController
+  include Pundit
+
+  before_action :authenticate_user!
+  before_action :ensure_finance
+
+private
+
+  def ensure_finance
+    raise Pundit::NotAuthorizedError, "Forbidden" unless current_user.finance?
+  end
+end
--- a/app/policies/finance_profile_policy.rb
+++ b/app/policies/finance_profile_policy.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class FinanceProfilePolicy < ApplicationPolicy
+  def index?
+    !!user&.finance?
+  end
+
+  class Scope < Scope
+    def resolve
+      return scope.all if user.finance?
+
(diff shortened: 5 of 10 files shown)


app/controllers/finance/lead_providers_controller.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class Finance::LeadProvidersController < Finance::BaseController
+  skip_after_action :verify_authorized
+  skip_after_action :verify_policy_scoped
```
kicferk1, line 5: We probably should either not skip them, or not include them in the base controller. As I see it, `ensure_finance` is enough for our current needs.

app/controllers/finance/lead_providers_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  def index
+    @lead_providers = LeadProvider.all
+  end
+
+  def show; end
```
kicferk1, line 11: If we really want to keep the pundit methods (`verify_authorized` and `verify_policy_scoped`), then we should use them here. In particular, scope in the index, and authorize in show. Also, in show, should we load a lead provider by id?

plisovin:
Cypress feature is coming soon

spec/policies/finance_profile_policy_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  context "being a finance" do
+    it { is_expected.to permit_action(:index) }
+  end
+
+  context "not being a finance" do
+    it { is_expected.to forbid_new_and_create_actions }
+    it { is_expected.to forbid_edit_and_update_actions }
+    it { is_expected.to forbid_action(:destroy) }
+  end
```
kicferk1, line 18: What about the new, create, etc. actions for a finance user? 

What about index action for non-finance user? Also, I reckon when you add those tests they will break. I think you want your `let(:user)` inside each context :)

spec/requests/finance/lead_providers_controller_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  describe "GET /finance/lead_providers" do
+    it "renders the index template" do
+      get "/finance/lead_providers"
+
+      expect(response).to render_template("finance/lead_providers/index")
+      expect(
+        Nokogiri.parse(response.body).css("table.govuk-table").text.gsub(/\n([ \t]*\n)+/, "\n").gsub(/^\s+|\s+$/, "").split("\n"),
+      ).to include(*lead_providers.map(&:name))
```
kicferk1, line 22: Uhh, can we use `assigns` to check the lead providers, and maybe have a cypress tests to have visuals and accessibility check?

timabell:
We need to add the cypress/percy coverage before we merge this....

app/controllers/finance/lead_providers_controller.rb
kicferk1: Also also, this does not have an existing route, so we will need to either add it, or remove this action.

plisovin: You are right, scope is not required as every user will be able to see all lead providers, at least for now

spec/factories/lead_provider.rb
```diff
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   factory :lead_provider do
-    name  { "Lead Provider" }
+    name  { Faker::Name.name }
```
kicferk1, line 5: I reckon this will cause a lot of randomness to show up in percy snapshots :/

plisovin: Reverted

spec/policies/finance_profile_policy_spec.rb
plisovin: Done

spec/policies/finance_profile_policy_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe FinanceProfilePolicy, type: :policy do
+  subject { described_class.new(user, finance_profile) }
+  let(:user) { create(:user, :finance) }
+  let(:finance_profile) { create(:finance_profile) }
```
kicferk1, line 8: I reckon those lets are unused :)

spec/requests/finance/lead_providers_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Lead Providers for Finance users", type: :request do
+  let(:name) { Faker::Name.name }
+  let(:email) { Faker::Internet.email }
```
kicferk1, line 7: As are there :)

kicferk1 (approved):
Some more cleanup in rspec, but looks good overall

timabell:
I think the commits could do with a bit of tidying up before merging.

spec/policies/finance_profile_policy_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe FinanceProfilePolicy, type: :policy do
+  subject { described_class.new(user, finance_profile) }
+  let(:finance_profile) { create(:finance_profile) }
+
+  context "being a finance" do
```
timabell, line 9:
```suggestion
  context "being a finance user" do
```

spec/policies/finance_profile_policy_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    let(:user) { create(:user, :finance) }
+
+    it { is_expected.to permit_action(:index) }
+    it { is_expected.to forbid_new_and_create_actions }
+    it { is_expected.to forbid_edit_and_update_actions }
+  end
+
+  context "not being a finance" do
```
timabell, line 17:
```suggestion
  context "not being a finance user" do
```

spec/policies/finance_profile_policy_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    let(:user) { create(:user) }
+
+    it { is_expected.to forbid_new_and_create_actions }
+    it { is_expected.to forbid_edit_and_update_actions }
+    it { is_expected.to forbid_action(:destroy) }
```
timabell, line 22: ditch the new/edit/create/update/destroy seeing as we only implement index?

PR #742, opened 2021-07-23
service to import NPQ manual validation CSV data

asmega:
### Context

- The automatic TRN validation at the moment is not great due to missing data
- Therefore there is currently a manual validation process in place involving CSVs
- After the manual validation has been performed the data needs to be re-imported back into the service 

- Add CSV importer for NPQ applications

### Guidance to review

- Added new folder and therefore namespace for importers
- I suggest we start grouping the importers together at least otherwise the services folder will just become unmanageable 
- `ll app/services | grep -i "importer.rb$" | wc -l` => `4`
- I'm sure there will be other groupings in the future

### Testing

- Start rails console
- Create test NPQManualValidation records with blank or incorrect TRNs
- Create a new CSV with headers `application_ecf_id,validated_trn` 
- Populate CSV with test records and correct TRNs
- `svc = Importers::NPQManualValidation.new(path_to_csv: Rails.root.join("manual-validation.csv"))` 
- `svc.call`
- Records should now have correct TRN and marked as verified

### Review Checks

- Probably easier locally but same instructions from testing apply here

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,3 +32,7 @@ Rails/UnknownEnv:
 
 Naming/VariableNumber:
   EnforcedStyle: snake_case
+
+Rails/Output:
+  Exclude:
+    - "app/services/importers/*"
--- a/app/services/importers/npq_manual_validation.rb
+++ b/app/services/importers/npq_manual_validation.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+class Importers::NPQManualValidation
+  attr_reader :path_to_csv
+
+  def initialize(path_to_csv:)
+    @path_to_csv = path_to_csv
+  end
+
+  def call
+    check_headers
+
+    rows.each do |row|
+      data = NPQValidationData.find_by(id: row["application_ecf_id"])
+
+      puts "no NPQValidationData found for #{row['application_ecf_id']}" if data.nil?
+      next if data.nil?
+
+      puts "updating trn for NPQValidationData: #{row['application_ecf_id']} with trn: #{row['validated_trn']}"
+
+      data.update!(teacher_reference_number: row["validated_trn"], teacher_reference_number_verified: true)
+    end
+  end
+
+private
+
+  def check_headers
+    unless rows.headers == %w[application_ecf_id validated_trn]
+      raise NameError, "Invalid headers"
+    end
+  end
+
+  def rows
+    @rows ||= CSV.read(path_to_csv, headers: true)
+  end
+end
--- a/spec/lib/importers/npq_manual_validation.rb
+++ b/spec/lib/importers/npq_manual_validation.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Importers::NPQManualValidation do
+  let(:profile) { create(:npq_validation_data) }
+
+  let(:file) { Tempfile.new("test.csv") }
+
+  around do |example|
+    original_stdout = $stdout
+    $stdout = File.open(File::NULL, "w")
+
+    example.run
+
+    $stdout = original_stdout
+  end
+
+  describe "#call" do
+    subject do
+      described_class.new(path_to_csv: file.path)
+    end
+
+    context "with well formed csv" do
+      before do
+        file.write("application_ecf_id,validated_trn")
+        file.write("\n")
+        file.write("123,7654321")
+        file.write("\n")
+        file.write("#{profile.id},7654321")
+        file.rewind
+      end
+
+      it "updates trn" do
+        expect {
+          subject.call
+        }.to change { profile.reload.teacher_reference_number }.to("7654321")
+      end
+
+      it "updates teacher_reference_number_verified to true" do
+        expect {
+          subject.call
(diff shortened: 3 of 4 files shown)


.rubocop.yml
```diff
@@ -32,3 +32,7 @@ Rails/UnknownEnv:
 
 Naming/VariableNumber:
   EnforcedStyle: snake_case
+
+Rails/Output:
+  Exclude:
+    - "app/services/importers/*"
```
kicferk1, line 38: Why do we need to add it? So that `puts` is not flagged by rubocop?

In `SchoolDataImporter` we use `Rails.logger` instead - probably worth keeping logging mechanism consistent.

app/services/importers/npq_manual_validation.rb
```diff
@@ -0,0 +1,36 @@
...
+    @path_to_csv = path_to_csv
+  end
+
+  def call
+    check_headers
+
+    rows.each do |row|
+      data = NPQValidationData.find_by(id: row["application_ecf_id"])
+
+      puts "no NPQValidationData found for #{row['application_ecf_id']}" if data.nil?
```
kicferk1, line 16: Do we want to log this as a sentry exception?

spec/lib/importers/npq_manual_validation.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    context "with well formed csv" do
+      before do
+        file.write("application_ecf_id,validated_trn")
+        file.write("\n")
+        file.write("123,7654321")
+        file.write("\n")
+        file.write("#{profile.id},7654321")
+        file.rewind
+      end
```
kicferk1, line 32: Do we want a fixture for that? (like the files in `spec/fixtures/files`?)

spec/lib/importers/npq_manual_validation.rb
```diff
@@ -0,0 +1,61 @@
...
+  let(:file) { Tempfile.new("test.csv") }
+
+  around do |example|
+    original_stdout = $stdout
+    $stdout = File.open(File::NULL, "w")
+
+    example.run
+
+    $stdout = original_stdout
+  end
```
kicferk1, line 17: What does this do? Hides logs in tests?

kicferk1:
> I suggest we start grouping the importers together at least otherwise the services folder will just become unmanageable

I agree that our services folder is getting bloated. That said, I am not sure I like grouping things together by the type of task they are meant to do. I reckon we should try to group them based on the business logic they operate on. Track and Pay services want to be together in one directory, whatever we might call it.

ethanmills:
What the plan for using this? Is it `scp`? Either way, would be good to get something in the readme

ethanmills:
@asmega is this still needed?

spec/lib/importers/npq_manual_validation.rb
asmega: Yes

asmega:
I can live without it. I run this from the console, so would save me from having to load the class from scratch. Plus I thought others might want to see what is actually being ran.

.rubocop.yml
asmega: I run this on the console so kind of want to see it in stdout and not lost somewhere in the logs

spec/lib/importers/npq_manual_validation.rb
asmega: I can move it to fixtures but really see the benefit of doing so, that's why I made it inline

kicferk1: I think csv files are ugly. That said, in this particular scenario I think keeping them inline makes sense - we want the `profile.id` to be something that is in our database, so it means either constructing the file at runtime (as you do here), or hard-coding the uid, which is not ideal.

.rubocop.yml
kicferk1:
```
/app # /usr/local/bin/bundle exec rails c
Loading production environment (Rails 6.1.3.2)
irb(main):001:0> Rails.logger.info "Test"
I, [2021-08-02T09:14:41.794245 #727]  INFO -- : Test
=> 53
```
Not sure what `53` is in there, but... This comes from our production app. `Rails.logger` still outputs to stdout, I think.

ethanmills:
@asmega I'd be happy to see this go in. Would you be able to add a line to the readme on how you get the CSV on to the server to  run this task?

asmega:
> @asmega I'd be happy to see this go in. Would you be able to add a line to the readme on how you get the CSV on to the server to run this task?

I've now added a section to the README detailing this

ethanmills (approved):

.rubocop.yml
asmega: If I did change to `Rails.logger` I don't see how logging the output would be very beneficial though?

kicferk1: It would remove the rubocop unhappiness 😂 

And I guess it would push us more in the direction of consistency of logging in our services / tasks.

I don't really mind the functionality.

asmega:
I'm closing this for now as https://github.com/DFE-Digital/early-careers-framework/pull/878 will impact this. I will re-open once i get a chance to patch it

PR #743, opened 2021-07-26
2020 cohort: Add the first actual version of the flow.

kicferk1:
### Context

https://dfedigital.atlassian.net/browse/CPDEL-726

Add a form, a controller and a bunch of views for 2020 data input. Hide them behind a new feature flag.

### Guidance to review

Multiple participants will come in a later PR. Entry points into the flow (almost certainly emails) will come in a later PR.

This replaces https://github.com/DFE-Digital/early-careers-framework/pull/729. You might notice it is not using the multipage form. That's because there are a bunch of edge cases that I know of already (such as opting out) and things we will do for the first time (managing multiple participants), which I think the multipage form is not really geared towards handling.

And I think for speed of iterative improvement it will be better to use a more flexible approach (arbitrary controller actions) - with an ambition to refactor it to use our multipage form when we know for sure what exactly we need to add to it to fit all our requirements. 

### Testing
rspec for the form, request spec for the controller, percy for accessibility and visual tests

### Review Checks

1. Go to `/schools/999999-example-school/year-2020/start` in the review app.

### DIFF
--- a/app/controllers/schools/year2020_controller.rb
+++ b/app/controllers/schools/year2020_controller.rb
@@ -17,7 +17,7 @@ module Schools
     def choose_induction_programme
       render :select_induction_programme and return unless @year_2020_form.valid? :choose_induction_programme
 
-      session[SESSION_KEY] = @year_2020_form.serializable_hash
+      store_year_2020_session
       if @year_2020_form.opt_out?
         @year_2020_form.opt_out!
         redirect_to action: :no_accredited_materials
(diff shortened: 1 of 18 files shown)


ethanmills:
Does this only allow you to add one participant?

kicferk1:
> Does this only allow you to add one participant?

Yes - multiple participants will be its own PR, cause I reckon this one is big enough as is.

ethanmills (approved):

PR #744, opened 2021-07-26
Dupe ECF users api

asmega:
### Context

- `/api/v1/users` endpoint at the moment contains ECF specific implementation
- NPQ at the moment requires a generic endpoint regarding users without any scoping to any subset of users

- Duplicate the `/api/v1/users` endpoint at `/api/v1/ecf-users`
- Once deployed ECF can be pointed to `/api/v1/ecf-users`
- This will free then free up `/api/v1/users` where all ECF specific implementation can be removed to make it generic

### Guidance to review

- I feel long-term we want to deprecate the `/api/v1/ecf-users` endpoints. And opt for filtering to return users only ECF is concerned about. And use https://jsonapi.org/format/#fetching-relationships for fetching desired associations 

### Testing

```sh
curl -H "Content-Type: application/vnd.api+json" -H "Authorization: Bearer PRIVATE_TOKEN" "http://localhost:3001/api/v1/ecf-users"
```
- should return some users just like before

### Review Checks

- Same concept as testing locally

### DIFF
--- a/app/controllers/api/v1/ecf_users_controller.rb
+++ b/app/controllers/api/v1/ecf_users_controller.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class ECFUsersController < Api::ApiController
+      include ApiTokenAuthenticatable
+      include ApiPagination
+
+      def index
+        render json: UserSerializer.new(paginate(users)).serializable_hash.to_json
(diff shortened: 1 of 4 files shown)


kicferk1 (approved):
Thanks, I will look at pointing E&L at the new endpoint once this gets released :)

PR #745, opened 2021-07-26
CPDTP-259 submit started declaration for npq event

blackrat:
### Context

NPQ "started" declarations need to be added to the participant-declarations endpoint in order to start collecting them.

- Change the shape of the controller to call a Factory which will fan-out to the different declaration recorders.
- Add in optional params to the controller, to allow an additional "evidence_held' field to be added later.
- Have the NPQCourse model pass back the list of npq-course-identifiers
- Have User perform a check for npq_profiles
- Split out the Declaration Recorder into separate ones for npq, early_career_teachers and mentors
- Create an abstract Base for these three for common functionality, and an ECF base for early_career_teachers and mentors
- Manage which one is called with a factory keyed of the course-identifier value

### Review Checks


app/services/record_declarations/npq.rb
```diff
@@ -0,0 +1,33 @@
...
+        NPQCourse.identifiers
+      end
+    end
+
+    def participant?
+      npq?
+    end
+
+    def user_profile
+      npq_profiles.includes({ validation_data: [:npq_course] }).where('npq_courses.identifier': course).first
```
kicferk1, line 18: Should this also depend on the lead provider?

etc/schema/0.3/npq/participant_declarations/create/request_schema.json
```diff
@@ -152,7 +86,7 @@
           "declaration_date",
           "declaration_type",
           "participant_id",
-          "course_identifier"
+          "course_type"
```
kicferk1, line 89: Are we sure we wanna change it?

spec/services/record_declarations/ecf/early_career_teacher_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+  let(:delivery_partner) { create(:delivery_partner) }
+  let!(:school_cohort) { create(:school_cohort, school: ect_profile.school, cohort: ect_profile.cohort) }
+  let!(:partnership) do
+    create(:partnership,
+           school: ect_profile.school,
+           lead_provider: cpd_lead_provider.lead_provider,
+           cohort: ect_profile.cohort,
+           delivery_partner: delivery_partner)
+  end
```
kicferk1, line 38: This setup seems very similar between tests. Can we see if there is some way of commonising it?

app/services/record_declarations/base.rb
```diff
@@ -0,0 +1,99 @@
...
+      add_participant_profile_params!
+
+      declaration = create_record!
+      validate_provider!
+      { id: declaration.id }
+    end
+
+  private
+
+    delegate :school, :cohort, to: :user_profile
```
kicferk1, line 35: This is not really gonna work for npq profiles - they don't have a cohort. I reckon it should be moved to ecf recorders.

app/services/record_declarations/ecf.rb
```diff
@@ -0,0 +1,19 @@
...
+      super.merge({ schema_path: "ecf/participant_declarations" })
+    end
+
+    def declaration_type
+      ParticipantDeclaration::ECF
+    end
+
+    def actual_lead_provider
+      SchoolCohort.find_by(school: school, cohort: cohort)&.lead_provider&.cpd_lead_provider
+    end
```
kicferk1, line 17: `user_profile.school_cohort.lead_provider&.cpd_lead_provider` should work

app/services/record_declarations/base.rb
```diff
@@ -0,0 +1,99 @@
...
+      params[:cpd_lead_provider]
+    end
+
+    def validate_provider!
+      raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || lead_provider_from_token == actual_lead_provider
+    end
+
+    def required_params
+      (self.class.required_params - [:participant_id] + [:user_id])
+    end
```
kicferk1, line 97: Uhh, what is this for exactly?

app/services/record_declarations/base.rb
```diff
@@ -0,0 +1,99 @@
...
+      end
+    end
+
+    def lead_provider_from_token
+      params[:cpd_lead_provider]
+    end
+
+    def validate_provider!
+      raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || lead_provider_from_token == actual_lead_provider
+    end
```
kicferk1, line 93: When does `actual_lead_provider.nil?` case occur? When ect/mentor's school is not in a partnership?

If soo, I would expect us to tell the LP that they gave us an invalid participant. Right?

app/services/record_participant_declaration.rb
```diff
@@ -1,110 +1,28 @@
 # frozen_string_literal: true
 
-require "json_schema/validate_body_against_schema"
-
 class RecordParticipantDeclaration
```
kicferk1, line 3: What job does this class do now? As I see it, not much of one. Maybe we should cut out the middle man and go call our factory in `ParticipantDeclarationsController`?

app/services/record_declarations/base.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    delegate :school, :cohort, to: :user_profile
+
+    def initialize(params)
+      @params = params
+      @params[:user_id] = params[:participant_id]
+    end
+
+    def validate_schema!
+      errors = JsonSchema::ValidateBodyAgainstSchema.call(schema: schema, body: params[:raw_event])
```
kicferk1, line 43: Are we sure this is the right place to validate `raw_event`? `ParticipantDeclarationsController` does `validate_params!`, would we not want to validate the raw event there?

etc/schema/0.3/npq/participant_declarations/create/request_schema.json
blackrat: Ah. No. We don't.

app/services/record_declarations/ecf.rb
blackrat: Cool. In that case the need for delegation mentioned before disappears.

app/services/record_declarations/base.rb
blackrat: Makes sure that only the required parameters are passed on to the declaration type in create_record!

app/services/record_declarations/base.rb
blackrat: Ahh. Looks like that returned after a merge. I'll remove it again.

app/services/record_declarations/base.rb
blackrat: There are several schemas that we want to validate against. Putting that into the controller would require much more logic than is required with the factory approach which only validates against the schema deemed as valid for that course and type.

app/models/user.rb
```diff
@@ -47,6 +47,10 @@ def mentor?
     mentor_profile&.active? || false
   end
 
+  def npq?
+    npq_profiles.any?
```
ethanmills, line 51: This should have an active check I think?

app/services/record_declarations/base.rb
```diff
@@ -0,0 +1,97 @@
...
+
+    def schema_validation_params
+      { version: "0.3" }
+    end
+
+    def schema
+      JSON.parse(File.read(JsonSchema::VersionEventFileName.call(schema_validation_params)))
+    end
+
+    def add_participant_profile_params!
```
ethanmills, line 53: Is this named correctly?

app/services/record_declarations/base.rb
```diff
@@ -0,0 +1,97 @@
...
+    def not_implemented_error
+      raise NotImplementedError, "Method must be implemented"
+    end
+
+    def call
+      validate_schema!
+      add_participant_profile_params!
+
+      declaration = create_record!
+      validate_provider!
```
ethanmills, line 29: Why does this happen after `create_record!`?

spec/services/record_declarations/ecf/early_career_teacher_spec.rb
kicferk1: Probably not in this PR.

app/services/record_declarations/base.rb
kicferk1: Cause we want to store it if the providers are shadily trying to declare participants which don't belong to them. I think.

app/services/record_declarations/base.rb
kicferk1: Good point, `validate_participant_params!` sounds like a much better name.

app/services/record_participant_declaration.rb
kicferk1: I think this class now does even less of a job than it used to when I wrote this comment.

I'm like 90% sure we should inline it - but we can probably do it in a later PR.

kicferk1 (approved):
Please address Ethan's comments before merging. 

Code looks reasonable to me - there are a few things we will need to fix (such as nil lead provider) - could we add JIRA tickets for them so we don't forget?

ethanmills (approved):

PR #746, opened 2021-07-26
TRA API will now send us padded TRNs

asmega:
### Context

- We have people that failed auto validation. turns out these people were simply missing from the temporary database
- but my understanding was they were in the database but did not have the padded trn. ie the look up failed
- eg user entered `0123456` we look for that but don’t find it because its stored under 123456
- so the temporary api now pads the request to TRA so it always finds the record
- on top of that responses to us will now contain a padded TRN
- therefore any matching we do our end must also be on a padded TRN
- without this change users who have been padding their own TRN will get auto validated. but those entering 6 digit trns will start failing

- API response will now be padded with leading zeros till 7 character length
- therefore any matching our side will also need to be padded
- service API has therefore changed as we now return padded TRNS
- which mirrors what TRA will now send to us

### Guidance to review

- No integrating parts will be broken as a result of this change

### Testing

- Connect to the test instance api
```ruby
ParticipantValidationService.validate(
  trn: "0000001",
  nino: "WRONG",
  full_name: "FULL NAME FROM EXTRACT",
  date_of_birth: DOB_FROM_EXTRACT,
)
```
- should return test record

### Review Checks

- Testing instructions apply here

### DIFF
--- a/app/services/participant_validation_service.rb
+++ b/app/services/participant_validation_service.rb
@@ -39,8 +39,10 @@ private
     dqt_record = dqt_record(trn, nino)
     return if dqt_record.nil?
 
+    padded_trn = trn.rjust(7, "0")
+
     matches = 0
-    trn_matches = trn == dqt_record[:teacher_reference_number]
+    trn_matches = padded_trn == dqt_record[:teacher_reference_number]
     matches += 1 if trn_matches
     name_matches = full_name == dqt_record[:full_name]
     matches += 1 if name_matches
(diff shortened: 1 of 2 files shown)


kicferk1 (approved):
Do we have any applicationsin db that need to be padded?

Would we even want to do that?

asmega:
> Do we have any applicationsin db that need to be padded?
> 
> Would we even want to do that?

Yeah i can do that by hand after this is in

ethanmills (approved):

PR #748, opened 2021-07-26
CPDRP-616: Allow admins to change induction choice

ethanmills:
### Context
CPDRP-616
Allow admins to change which programme a school is doing for a cohort. There will be follow-up PRs to allow challenging partnerships and changing CIP materials

- Extract partnership challenge logic. This has the added benefit of making this easier on the console
- Add service to change induction programme choice and cip materials
- Display full information about induction choice to admins
- Change programme functionality
- Add rspec feature test 

### Guidance to review
I've written all the feature steps inline at the moment. I intend to extract them as they're reused

### Review Checks

admin@example.com


app/controllers/admin/schools/cohorts_controller.rb
```diff
@@ -10,8 +10,8 @@ def index
         @school_cohorts = policy_scope(SchoolCohort)
           .joins(:school)
           .where(school: @school)
-          .joins(:cohort)
-          .order("cohorts.start_year ASC")
+
+        @cohorts = Cohort.all.order(start_year: :asc)
```
kicferk1, line 14: Will that include the 2020 cohort? Is that a good idea?

app/services/admin/change_induction_service.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module Admin
+  class ChangeInductionService
```
kicferk1, line 4: We might want to make it not work for 2020 cohort - or restrict it to just allow changing CIP perhaps?

kicferk1 (commented):
Looks good overall - I'm just worried about 2020 cohort :P And maybe a little worried about repercussions for Track and Pay, but they are probably something to address later.

spec/requests/admin/schools/cohorts_spec.rb
```diff
@@ -17,6 +26,15 @@
       get "/admin/schools/#{school.id}/cohorts"
       expect(response).to render_template("admin/schools/cohorts/index")
       expect(assigns(:school_cohorts)).to match_array school.school_cohorts
+      Cohort.all.each do |cohort|
+        expect(response.body).to include cohort.start_year
+      end
```
kicferk1, line 31:
```
  1) Admin::Schools::Cohorts GET /admin/schools/:school_id/cohorts renders the index template
     Failure/Error: expect(response.body).to include cohort.start_year

     TypeError:
       no implicit conversion of Integer into String
     # ./spec/requests/admin/schools/cohorts_spec.rb:30:in `block (4 levels) in <top (required)>'
     # ./spec/requests/admin/schools/cohorts_spec.rb:29:in `block (3 levels) in <top (required)>'
```
argh!

app/controllers/admin/schools/cohorts/change_programme_controller.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Admin
+  module Schools
+    module Cohorts
+      class ChangeProgrammeController < ::Admin::BaseController
+        skip_after_action :verify_authorized
+        skip_after_action :verify_policy_scoped
```
kicferk1, line 8: Just a general moan about this.

spec/features_helper.rb
```diff
@@ -8,7 +8,7 @@
 
 Capybara.register_driver :chrome_headless do |app|
   options = ::Selenium::WebDriver::Chrome::Options.new
-  options.add_argument("--headless")
+  options.add_argument("--headless") unless ENV["HEADLESS"] == "false"
```
tonyheadford, line 11:
```suggestion
  options.add_argument("--headless") unless ENV["NOT_HEADLESS"]
```
Flows a bit nicer

app/controllers/admin/schools/cohorts/change_programme_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+          set_success_message heading: "Induction programme has been changed"
+          redirect_to admin_school_cohorts_path
+        end
+
+      private
+
+        def set_induction_choice_form
+          @induction_choice_form = InductionChoiceForm.new(params.require(:induction_choice_form).permit(:programme_choice))
+          @induction_choice_form.cohort = @cohort
```
kicferk1, line 32: Do we want to set school here too?

kicferk1: We could probably set the form before each action, if we handled empty params. Up to you whether you prefer it :) 

```
    def change_programme_params
      return {} unless params.key?(: induction_choice_form)

      params.require(: induction_choice_form).permit(: programme_choice)
    end
```

spec/services/admin/change_induction_choice_service_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+      end
+    end
+  end
+
+  describe "#challenge_partnership" do
+    let(:school) { create(:school_cohort, induction_programme_choice: "full_induction_programme", cohort: cohort).school }
+    let!(:partnership) { create(:partnership, cohort: cohort, school: school) }
+
+    it "should record the reason for challenging the partnership"
+    it "should email the lead provider"
```
ethanmills, line 221: Missed these, picking up in the next one where I use them

tonyheadford (approved):
Nice

spec/features/admin/schools/changing_provision_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    then_i_should_be_on_the_school_cohorts_page
+    and_there_should_be_a_success_banner
+  end
+
+private
+
+  def given_there_is_a_fip_school_in_2021
+    cohort = create(:cohort, start_year: 2021)
+    @school_cohort = create(:school_cohort, cohort: cohort, induction_programme_choice: "full_induction_programme")
+  end
```
kicferk1, line 31: Not sure if this is the right PR for it, but I would love for us to have a similar shared steps setup to what we have in cypress tests.

app/controllers/admin/schools/cohorts/change_programme_controller.rb
ethanmills: Yeah, understood. I wanted to bake a load of the logic about what was allowed and not into the policy, not the service, but couldn't find a nice way to do it, since it depends on previous and current state

app/controllers/admin/schools/cohorts/change_programme_controller.rb
ethanmills: It doesn't have or need a school. I think I prefer this way, since we keep the require on the param

spec/features/admin/schools/changing_provision_spec.rb
ethanmills: I've written all the feature steps inline at the moment. I intend to extract them as they're reused

kicferk1 (approved):

PR #749, opened 2021-07-26
Enable capybara so we can write rspec feature tests

tonyheadford:
### Context
Enable writing feature specs in RSpec (task from dev retro).  I've added an example feature spec with the work-around to prevent rutabaga from trying to load `.feature` file
Once the rutabaga feature tests for the payments calculator are refactored and removed, we can remove the `rutabaga: false` param.

Add `require` for capybara dependencies.
Add example feature test.
Fix spec that was using `.and` which isn't supported with Capybara::RSpecMatchers.

### Review Checks


ethanmills (approved):
Are we planning on hooking Percy up to this?

tonyheadford:
> Are we planning on hooking Percy up to this?

Yes, ideally we can hook this up with the accessibility tests and percy. Looking to build on this during the participant validation

PR #750, opened 2021-07-27
Touch User on update to participant profile

ethanmills:
The participants API uses updated_at on the User model

### DIFF
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -2,7 +2,7 @@
 
 class ParticipantProfile < ApplicationRecord
   has_paper_trail
-  belongs_to :user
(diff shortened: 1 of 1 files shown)


kicferk1 (approved):
Should we also touch `school_cohort` on the ecf participant profiles? So when induction programme choice changes (for example), we mark the participant as updated?

kicferk1:
Argh, percy randomness strikes again 😢

ethanmills:
> Argh, percy randomness strikes again 😢

I'll get it fixed separately 😢

ethanmills:
We also need to do this for Partnerships and SchoolCohorts, but that will have to be callbacks I think. To be done separately

PR #751, opened 2021-07-27
2020 Invite email: Add invite email for 2020 Cohort

kicferk1:
### Context

We will want to send schools an email about 2020 cohort. This should include a link - when they click it, they should go to the start of our flow.

Add a new email to school mailer, and another method to the service inviting schools.
Add some tests.

We want to send the email to:
> State-funded English schools. In essence, the exact same schools who are eligible for FIP.

(from slack message)

You can't.

### DIFF
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -224,7 +224,7 @@ class InviteSchools
   def send_year2020_invite_email
     return unless FeatureFlag.active?(:year_2020_data_entry)
 
-    School.currently_open.each do |school|
+    School.eligible.each do |school|
       recipient = school.contact_email
       SchoolMailer.year2020_invite_email(recipient: recipient, start_url: year2020_start_url(school)).deliver_later if recipient.present?
     end
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -221,12 +221,29 @@ class InviteSchools
     end
   end
 
+  def send_year2020_invite_email
+    return unless FeatureFlag.active?(:year_2020_data_entry)
+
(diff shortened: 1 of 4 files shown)


app/services/invite_schools.rb
```diff
@@ -221,6 +221,15 @@ def send_induction_coordinator_add_participants_email
     end
   end
 
+  def send_year2020_invite_email
+    return unless FeatureFlag.active?(:year_2020_data_entry)
+
+    School.currently_open.each do |school|
```
ethanmills, line 227: I read your comment, but looking at slack I think this should be `eligible`

kicferk1: Yep, you got to it literal seconds before I fixed it :D

app/services/invite_schools.rb
```diff
@@ -221,6 +221,15 @@ def send_induction_coordinator_add_participants_email
     end
   end
 
+  def send_year2020_invite_email
+    return unless FeatureFlag.active?(:year_2020_data_entry)
```
ethanmills, line 225: Are they planning a phased rollout of this? I really hope so....

ethanmills (approved):

app/services/invite_schools.rb
kicferk1: Maybe? I imagine it will end up being phased, and this is just a first iteration that we can build on when we know more requirements.

kicferk1: The exact comms plan and email content will be done next sprint, apparently. So this will be improved upon - for now it's more of a foundation to make the future work easier.

PR #752, opened 2021-07-27
Shadow TeacherProfile

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-759

This PR creates TeacherProfile and automatically creates them on each ParticipantProfile save. This way we can safely switch to the new TeacherProfile without any downtime.

Once this is deplyed we will run `data:populate_teacher_profiles` after which we can switch to use the teacher profile and break participant_profile->user association in the next PR.

### DIFF
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -3,6 +3,7 @@
 class ParticipantProfile < ApplicationRecord
   has_paper_trail
   belongs_to :user
+  belongs_to :teacher_profile, optional: true
   has_many :validation_decisions, class_name: "ProfileValidationDecision"
 
   enum status: {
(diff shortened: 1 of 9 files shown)


app/models/participant_profile.rb
```diff
@@ -57,4 +61,10 @@ def validation_decision(name)
     decision = validation_decisions.find { |record| record.validation_step.to_s == name.to_s }
     decision || validation_decisions.build(validation_step: name)
   end
+
+  def sync_teacher_profile
+    self.teacher_profile = user.teacher_profile || user.build_teacher_profile
```
kicferk1, line 66: What is `build_teacher_profile`?

BroiSatse: It is a `belongs_to` equivalent of `records.build`: https://guides.rubyonrails.org/association_basics.html#belongs-to-association-reference

kicferk1 (approved):

PR #755, opened 2021-07-27
Touch User whenever a change is made that is relevant to the participants API

ethanmills:
Currently for example, when a school forms a partnership and their participants become visible to an LP, that will not be reflected in the updated_since parameter for the participants API endpoint

### DIFF
--- a/spec/components/admin/participants/table_spec.rb
+++ b/spec/components/admin/participants/table_spec.rb
@@ -4,13 +4,13 @@ RSpec.describe Admin::Participants::Table, type: :view_component do
   let!(:participant_profiles) { create_list :participant_profile, rand(11..15) }
   let(:page) { rand 1..2 }
 
-  component { described_class.new profiles: ParticipantProfile.all, page: page }
+  component { described_class.new profiles: ParticipantProfile.all.order(:id), page: page }
   request_path "/admin/participants"
 
   stub_component Admin::Participants::TableRow
 
   it "renders table row for each participant profile on this page" do
-    expected_profiles = participant_profiles.each_slice(10).to_a[page - 1]
+    expected_profiles = participant_profiles.sort_by(&:id).each_slice(10).to_a[page - 1]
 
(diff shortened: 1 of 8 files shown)


kicferk1 (approved):
Not sure why the test is failing, very unlikely that your changes broke it.

spec/components/admin/participants/table_spec.rb
```diff
@@ -4,13 +4,13 @@
...
 
-  component { described_class.new profiles: ParticipantProfile.all, page: page }
+  component { described_class.new profiles: ParticipantProfile.all.order(:id), page: page }
   request_path "/admin/participants"
 
   stub_component Admin::Participants::TableRow
 
   it "renders table row for each participant profile on this page" do
-    expected_profiles = participant_profiles.each_slice(10).to_a[page - 1]
+    expected_profiles = participant_profiles.sort_by(&:id).each_slice(10).to_a[page - 1]
```
kicferk1, line 13: what were they ordered by before? :O

BroiSatse: They weren't, which was causing issues on pagination. This component is expecting olready ordered relation.

ethanmills: Depends which way the wind was blowing 🤷

BroiSatse:
That feels strange - why would we expose updated_at of the user via API?

kicferk1:
> That feels strange - why would we expose updated_at of the user via API?

We might not - but a filter on the api saying "only return users which have changed since xxx" is a very useful tool - so you don't have to fetch all 20k users each time you want to get the new or updated ones.

ethanmills:
I spoke to Ben S over on Apply since they've solved the same problem, this is pretty much what they ended up with too. We might need to add a specific field to track relevant updates if the LPs get annoyed that we're showing them records that haven't visibly changed, but this will do for now

PR #756, opened 2021-07-27
Remove random data from visual tests

ethanmills:

### DIFF
--- a/spec/cypress/app_commands/scenarios/admin/school_participants.rb
+++ b/spec/cypress/app_commands/scenarios/admin/school_participants.rb
@@ -45,11 +45,19 @@ FactoryBot.create :participant_profile,
                   school_cohort: another_school_cohort,
                   created_at: Date.parse("29/12/2020")
 
-npq_profile = FactoryBot.create :participant_profile, :npq, validation_data: nil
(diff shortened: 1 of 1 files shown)


kicferk1 (commented):
Is that enough? I reckon we need email for the user too.

spec/cypress/app_commands/scenarios/admin/school_participants.rb
```diff
@@ -45,11 +45,12 @@
...
+npq_profile = FactoryBot.create :participant_profile, :npq, validation_data: nil, updated_at: Date.parse("19/09/2019")
 
 FactoryBot.create :npq_validation_data,
                   id: npq_profile.id,
                   user: FactoryBot.create(:user, full_name: "Natalie Portman Quebec"),
                   date_of_birth: Date.parse("10/12/1982"),
                   nino: "NI123456",
-                  created_at: Date.parse("19/09/2019")
+                  created_at: Date.parse("19/09/2019"),
+                  teacher_reference_number: "9780824"
```
kicferk1, line 56: Syntax is wrong here ;)

ethanmills:
> Is that enough? I reckon we need email for the user too.

Good spot. The diff I was looking at was being unhelpful

kicferk1 (approved):

PR #758, opened 2021-07-27
prevent screenshots and video on cypress tests

ethanmills:
This should speed up the test runs. We don't use them anyway

I tried to do this before, but only changed the local setting


ethanmills:
Seems to have gone from ~13mins to ~9

kicferk1 (approved):

PR #759, opened 2021-07-27
Fixed race condition in teacher profile shadowing

BroiSatse:
### Context

Oops! There's a bug in TeacheProfile shadowing code causing new teacher profiles not being associated with newly created participant profiles: `self.techer_profile = <new_instance_of_record>` does not set the foregin_key value and, for some reason `self.teacher_profile.save` did not update the foreign key...

### DIFF
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -63,8 +63,10 @@ class ParticipantProfile < ApplicationRecord
   end
 
   def sync_teacher_profile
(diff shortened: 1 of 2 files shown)


kicferk1 (approved):
Is this PR coming out of the well-principled approach, or did we break something with it?

BroiSatse:
@kicferk1 - As the teacher profile is not being used anywhere it doesn't break anything - we will need to re-run data migration as the current shadowing is not fully working (it is creating TeacherProfile but fails to associate them with the PariticpantProfile).

PR #760, opened 2021-07-27
CPDTP-259 Eliminate the nil check from validate_provider!

blackrat:
### Context

The old code allowed a lead_provider to be nil if a declaration were submitted before the lead_provider was associated with the user profile. This was erroneous and this patch fixes it and updates the tests to require a partnership. It also eliminates an extra 200 test from the API specification, since those tests are included in the request and as the swagger spec is mainly to generate a schema and documentation, they aren't as useful to have them in there.

- Remove the nil check for the lead_provider associated with the ECF from the declaration
- Remove the extra 200 check from the swagger documentation tests
- Fix the tests to make them work and generate the correct actual_lead_provider for the other parameters specified

### Review Checks

### DIFF
--- a/spec/docs/participant_declarations_spec.rb
+++ b/spec/docs/participant_declarations_spec.rb
@@ -5,6 +5,13 @@ require "swagger_helper"
 RSpec.describe "Participant Declarations", type: :request, swagger_doc: "v1/api_spec.json" do
   let(:early_career_teacher_profile) { create(:early_career_teacher_profile) }
   let(:cohort) { early_career_teacher_profile.cohort }
+  let!(:school_cohort) { create(:school_cohort, school: early_career_teacher_profile.school, cohort: early_career_teacher_profile.cohort) }
+  let!(:ect_partnership) do
+    create(:partnership,
+           school: early_career_teacher_profile.school,
+           lead_provider: cpd_lead_provider.lead_provider,
+           cohort: early_career_teacher_profile.cohort)
+  end
   let(:user) { early_career_teacher_profile.user }
   let(:lead_provider) { create(:lead_provider) }
   let(:cpd_lead_provider) { create(:cpd_lead_provider, lead_provider: lead_provider) }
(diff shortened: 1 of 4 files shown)


spec/docs/participant_declarations_spec.rb
```diff
@@ -5,6 +5,13 @@
 RSpec.describe "Participant Declarations", type: :request, swagger_doc: "v1/api_spec.json" do
   let(:early_career_teacher_profile) { create(:early_career_teacher_profile) }
   let(:cohort) { early_career_teacher_profile.cohort }
+  let!(:school_cohort) { create(:school_cohort, school: early_career_teacher_profile.school, cohort: early_career_teacher_profile.cohort) }
```
ethanmills, line 8: This shouldn't work I don't think - `create(:early_career_teacher_profile)` already creates a SchoolCohort, and I thought SchoolCohorts were unique on cohort&school - maybe not.
Either way this should be 
```suggestion
  let!(:school_cohort) { early_career_teacher_profile.school_cohort }
```

timabell (approved):

spec/services/record_declarations/ecf/mentor_spec.rb
```diff
@@ -28,13 +28,19 @@
     ect_params.merge({ user_id: induction_coordinator_profile.user_id })
   end
   let(:delivery_partner) { create(:delivery_partner) }
-  let!(:school_cohort) { create(:school_cohort, school: ect_profile.school, cohort: ect_profile.cohort) }
-  let!(:partnership) do
+  let!(:ect_school_cohort) { create(:school_cohort, school: ect_profile.school, cohort: ect_profile.cohort) }
```
ethanmills, line 31: ditto

spec/docs/participant_declarations_spec.rb
blackrat: They weren't linking up with the partnership, which is where we were seeing issues and why it had to become explicit either one way or the other.

blackrat: I'll try with this, since the issue that I was having was a disjoin between the mentor and ect profile school_cohorts and fixing one may fix the other.

blackrat:
PR invalidated by other changes to the codebase switching validation to rails validation and away from schema validation.

PR #761, opened 2021-07-27
/api/v1/users now returns all users

asmega:
### Context

- `/api/v1/users` used to return ECF specific users and this has been duplicated to `/api/v1/ecf-users`
- This endpoint is therefore free to deal with the concept of all users in CPD
- The use case for this change is so NPQ is able to query for existing users in the system based on email address

- `/api/v1/users` now returns all users
- this is no longer scoped to ECF specific users
- serialization is now generalised removing all ECF specific attributes

### Guidance to review

- Besides NPQ no other client that I am aware of is using this endpoint and therefore safe to change. Is this the case?
- There should be no changes to the endpoint `/api/v1/ecf-users`

### Testing

- See testing on review app

### Review Checks

- Create or use an existing token with private api access
- Create or use an existing non ECF user
- Search for them 
```sh
curl -H "Content-Type: application/vnd.api+json" -H "Authorization: Bearer TOKEN" "https://ecf-review-pr-761.london.cloudapps.digital/api/v1/users?filter[email]=user@example.com"
```
- Should return the user based off their email
- JSON should no longer return an ECF specific properties

### DIFF
--- a/app/controllers/api/v1/ecf_users_controller.rb
+++ b/app/controllers/api/v1/ecf_users_controller.rb
@@ -7,7 +7,7 @@ module Api
       include ApiPagination
 
       def index
-        render json: UserSerializer.new(paginate(users)).serializable_hash.to_json
+        render json: ECFUserSerializer.new(paginate(users)).serializable_hash.to_json
       end
 
       def create
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -51,7 +51,7 @@ module Api
       end
 
       def users
-        users = User.is_ecf_participant
+        users = User.all
 
         if updated_since.present?
           users = users.changed_since(updated_since)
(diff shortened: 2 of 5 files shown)


kicferk1 (approved):
LGTM

Do you reckon we should cut out `create` endpoint from the ecf controller? I reckon the way it is used makes it more reasonable to keep it in generic users one. Thoughts?

asmega:
> LGTM
> 
> Do you reckon we should cut out `create` endpoint from the ecf controller? I reckon the way it is used makes it more reasonable to keep it in generic users one. Thoughts?

If ECF doesn't need it then yes you can delete. I don't know enough to say whether it is "safe" to delete though. I think use the generic where possible. I only duplicated the endpoint for continuity. Long term i think it would be great just to have the users endpoints. If you'd need specific data we can implement filters. If you need extra data I think we should be fetching the relations/association as opposed to having one endpoint return everything.

PR #764, opened 2021-07-28
2020 Multiple participants: Add adding, removing and updating participants

kicferk1:
### Context

https://dfedigital.atlassian.net/browse/CPDEL-727

Add `participants` field to the 2020 form. That's where the data for all participants is stored.
Add pages and routes for edit/update, and remove/delete teachers.

### Review Checks

Go to `/schools/999999-example-school/year-2020/start` and play around :)

### DIFF
--- a/app/forms/schools/year2020_form.rb
+++ b/app/forms/schools/year2020_form.rb
@@ -5,19 +5,20 @@ module Schools
     include ActiveModel::Model
     include ActiveModel::Serialization
 
-    attr_accessor :school_id, :induction_programme_choice, :core_induction_programme_id, :full_name, :email
+    attr_accessor :school_id, :induction_programme_choice, :core_induction_programme_id, :full_name, :email, :participants
 
     validates :induction_programme_choice, presence: true, on: :choose_induction_programme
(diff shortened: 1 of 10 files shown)


app/forms/schools/year2020_form.rb
```diff
@@ -57,12 +95,14 @@ def save!
...
         school_cohort.save!
 
-        EarlyCareerTeachers::Create.call(
-          full_name: full_name,
-          email: email,
-          school_cohort: school_cohort,
-          mentor_profile_id: nil,
-        )
+        get_participants.each do |participant|
+          EarlyCareerTeachers::Create.call(
```
ethanmills, line 99: Do we test that this is called the right number of times with the correct args?

kicferk1: Hmm, `correct number of times` - probably yes, we have a test saying 'if form has 3 participants, and you save, then the school cohort should have 3 participants'. Close enough.

But they could in principle have the wrong details. I will add a test.

ethanmills (approved):
Be good to get the test for correct user details, otherwise happy

PR #765, opened 2021-07-28
Instrument the JSONAPI::Serializer s

ethanmills:
There's something slow on the API, and I think it might be this


blackrat:
From the documentation, it reads like that should speed things up.

blackrat (approved):

PR #767, opened 2021-07-28
CPDTP-246 Add a page to show the ECF payment breakdown display

timabell:
### Context

https://dfedigital.atlassian.net/browse/CPDTP-246

* Add a page to show the ECF payment breakdown display

Provider names now have links:

Selection_20210728-01-00a9100a9100a91

Click the link and see the payment breakdown:

Selection_20210728-01-c1748c1748c1748

Go to https://ecf-review-pr-767.london.cloudapps.digital/ and sign in as `finance@example.com`
### Guidance to review

:ship:

### Testing

New cypress test.

### Review Checks


app/controllers/finance/lead_providers_controller.rb
```diff
@@ -1,7 +1,34 @@
...
+      contract: @ecf_lead_provider.call_off_contract,
+      aggregator: ::ParticipantEventAggregator,
+      uplift_aggregator: ::ParticipantUpliftAggregator,
+      calculator: ::PaymentCalculator::Ecf::PaymentCalculation,
+      event_type: :started,
+    )
+
+    @total_ect = ParticipantEventAggregator.call({ cpd_lead_provider: cpd_lead_provider, started: :count_active_ects_for_lead_provider })
+    @total_mentors = ParticipantEventAggregator.call({ cpd_lead_provider: cpd_lead_provider, started: :count_active_mentors_for_lead_provider })
+    @total_participants = ParticipantEventAggregator.call({ cpd_lead_provider: cpd_lead_provider, started: :count_active_for_lead_provider })
```
kicferk1, line 25: I'm looking at the code for `ParticipantEventAggregator` and I have a hard time understanding what's going on in there.

```
ParticipantEventAggregator.call({ cpd_lead_provider: cpd_lead_provider, started: :count_active_ects_for_lead_provider })

...

class ParticipantEventAggregator
  include HasDIParameters

  def call(event_type: :started)
    recorder.send(params[event_type], cpd_lead_provider)
  end

private

  def default_params
    {
      recorder: ParticipantDeclaration,
      started: :count_active_for_lead_provider,
    }
  end
end
```
If I see this correctly... `ParticipantEventAggregator.call` makes a new instance of `ParticipantEventAggregator`, and `call`s it with the same parameters. Which the instance ignores, because the only thing its call cares about is `event_type`. It gets all other things from fields that were created in its `new` call. 
In the meantime, the params with which we called the class need to include the hash for `event_type` value, and the value of that hash needs to be a key matching a scope on a model somewhere. 

Other complexities aside, I think it is not a great idea to have our `new` and `call` methods take the same set of arguments - because at best those sets are independent, which means both methods receive and ignore a lot of needless arguments, and at worst they aren't independent, and you can't `call` with a different value of a param than you `new` - unless you break the general pattern...

Not sure if this is the right PR to fix it in, probably not :(

app/controllers/finance/lead_providers_controller.rb
```diff
@@ -1,7 +1,34 @@
...
+    @uplift_participants = ParticipantUpliftAggregator.call({ cpd_lead_provider: cpd_lead_provider }).to_i
```
kicferk1, line 26: Why do we need to call `to_i` here?

app/views/finance/lead_providers/show.html.erb
```diff
@@ -0,0 +1,98 @@
+<% content_for :title, "Lead Provider Payment Breakdown" %>
+<%
+  def band_index_to_heading(number)
+    ("A".."Z").to_a[number]
+  end
+%>
```
kicferk1, line 6: This should be a helper method in the controller.

app/views/finance/lead_providers/show.html.erb
```diff
@@ -0,0 +1,98 @@
...
+      <th class="govuk-table__header">Band</th>
+      <th class="govuk-table__header">Number of Participants</th>
+      <th class="govuk-table__header">Payment amount per person</th>
+      <th class="govuk-table__header">Payment amount monthly</th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+    <% @breakdown[:service_fees].each_with_index do |service_fee, i| %>
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell"><%= band_index_to_heading(i) %></td>
```
kicferk1, line 49: Or maybe the value of `band_index_to_heading` should be in the `service_fee` - I imagine wherever we want to display them it would be useful to know whether it's band A, B or C.

app/views/finance/lead_providers/show.html.erb
```diff
@@ -0,0 +1,98 @@
...
+      <td class="govuk-table__cell"><%= @ecf_lead_provider.call_off_contract.bands[i].number_of_participants_in_this_band(@ecf_lead_provider.call_off_contract[:recruitment_target]) %></td>
```
kicferk1, line 50: Should `service_fee` include this in its output?

app/views/finance/lead_providers/show.html.erb
```diff
@@ -0,0 +1,98 @@
...
+      <td class="govuk-table__cell"><%= number_to_currency service_fee[:service_fee_per_participant] %></td>
```
kicferk1, line 51: Where does `number_to_currency` come from?

app/views/finance/lead_providers/show.html.erb
```diff
@@ -0,0 +1,98 @@
...
+      <th class="govuk-table__header">Number of Participants</th>
+      <th class="govuk-table__header">Payment amount per person</th>
+      <th class="govuk-table__header">Payment amount monthly</th>
+    </tr>
+  </thead>
+  <tbody class="govuk-table__body">
+    <% @breakdown[:output_payments].each_with_index do |service_fee, i| %>
+    <tr class="govuk-table__row">
+      <td class="govuk-table__cell"><%= band_index_to_heading(i) %></td>
+      <td class="govuk-table__cell"><%= @ecf_lead_provider.call_off_contract.bands[i].number_of_participants_in_this_band(@ecf_lead_provider.call_off_contract[:recruitment_target]) %></td>
```
kicferk1, line 72: Ditto about whether we should include those things in `service_fee`

config/application.rb
```diff
@@ -30,6 +30,7 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.0
     config.autoloader = :zeitwerk
+    config.i18n.default_locale = :'en-GB'
```
kicferk1, line 33: Is this so we can print numbers in the right currency? Isn't it a bit of an overkill?

app/views/finance/lead_providers/show.html.erb
timabell: Yeah I think this view knows far too much about the internals of things further down.

app/views/finance/lead_providers/show.html.erb
timabell: Rails

https://apidock.com/rails/ActionView/Helpers/NumberHelper/number_to_currency

config/application.rb
timabell: Yes it is. Seems reasonable to me, this is a GB app. Without this we'd have to pass the locale to every call to the currency formatter.... (and I couldn't actually make that work for some reason I didn't get to the bottom of)

app/views/finance/lead_providers/show.html.erb
```diff
@@ -0,0 +1,98 @@
...
+        <th class="govuk-table__header govuk-!-width-one-half">Milestone</th>
+        <td class="govuk-table__cell" data-test="lead-provider-name">Started</td>
+      </tr>
+      <tr class="govuk-table__row">
+        <th class="govuk-table__header">Recruitment target</th>
+        <td class="govuk-table__cell" data-test="lead-provider-name"><%= @ecf_lead_provider.call_off_contract[:recruitment_target] %></td>
+      </tr>
+      <tr class="govuk-table__row">
+        <th class="govuk-table__header">Current ECTs</th>
+        <td class="govuk-table__cell" data-test="lead-provider-name"><%= @total_ect %></td>
```
timabell, line 23: Generally, I'm not sure we have enough test coverage on this patch, I'm not currently confident that tests would catch breakages if numbers went missing. (I got the PR out anyway so we could get moving and make a more shared decision). Making a comment on a line so we can use the "resolve" feature of github.

kicferk1: Percy snapshots would catch those numbers missing (I hope). If you want to make doubly sure, you can check the `assigns` of `total_ect` variable in rspec - similar to
```
  describe "GET /finance/lead-providers" do
    it "renders the index template" do
      get "/finance/lead-providers"

      expect(response).to render_template("finance/lead_providers/index")
      expect(assigns(:lead_providers)).to eq(lead_providers)
    end
  end
```

spec/factories/call_off_contract.rb
```diff
@@ -6,6 +6,7 @@
     uplift_amount { 100 }
     recruitment_target { 2000 }
     set_up_fee { 149_651 }
+    lead_provider { build(:lead_provider, cpd_lead_provider: build(:cpd_lead_provider)) }
```
timabell, line 9: I'm wondering if this new one from the build will get persisted or just end up in-memory

spec/requests/finance/lead_providers_controller_spec.rb
```diff
@@ -15,7 +16,16 @@
       get "/finance/lead-providers"
 
       expect(response).to render_template("finance/lead_providers/index")
-      expect(assigns(:lead_providers)).to eq(lead_providers)
+      expect(assigns(:ecf_lead_providers)).to eq(ecf_lead_providers)
+    end
+  end
+
+  describe "GET /finance/lead-providers/{:id}" do
```
timabell, line 23: I think we need more coverage in the request spec

spec/services/record_declarations/ecf/early_career_teacher_spec.rb
```diff
@@ -3,7 +3,8 @@
 require "rails_helper"
 
 RSpec.describe RecordDeclarations::ECF::EarlyCareerTeacher do
-  let(:cpd_lead_provider) { create(:cpd_lead_provider) }
+  let(:ecf_lead_provider) { create(:lead_provider) }
+  let(:cpd_lead_provider) { create(:cpd_lead_provider, lead_provider: ecf_lead_provider) }
```
timabell, line 7: Incidentally, I want to make the `[ecf_]lead_provider` factory automatically generate its parent `cpd_lead_provider` record, but I'd already spent far too long on the factories and this patch was the first variant where I got everything to work together. Further iteration welcome!

app/controllers/finance/lead_providers_controller.rb
timabell: https://github.com/DFE-Digital/early-careers-framework/blob/9a5480476680773d0964d8f7ddd2e540c8a20937/app/services/participant_event_aggregator.rb :eyes:
Perhaps @blackrat would be best placed to respond to this one

app/views/finance/lead_providers/show.html.erb
blackrat: This would be best placed as a decorator.

app/views/finance/lead_providers/show.html.erb
blackrat: Another candidate for decorators

app/controllers/finance/lead_providers_controller.rb
timabell: Gone, cheers

app/controllers/finance/lead_providers_controller.rb
```diff
@@ -1,7 +1,34 @@
...
+    @ecf_lead_provider = lead_provider_scope.find(params[:id])
+    cpd_lead_provider = @ecf_lead_provider.cpd_lead_provider
+
+    @breakdown = CalculationOrchestrator.call(
+      cpd_lead_provider: cpd_lead_provider,
+      contract: @ecf_lead_provider.call_off_contract,
+      aggregator: ::ParticipantEventAggregator,
+      uplift_aggregator: ::ParticipantUpliftAggregator,
+      calculator: ::PaymentCalculator::Ecf::PaymentCalculation,
+      event_type: :started,
```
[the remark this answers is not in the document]
timabell: Done, good spot

app/controllers/finance/lead_providers_controller.rb
```diff
@@ -1,7 +1,30 @@
...
+
+    @total_ect = ParticipantEventAggregator.call({ cpd_lead_provider: cpd_lead_provider, started: :count_active_ects_for_lead_provider })
+    @total_mentors = ParticipantEventAggregator.call({ cpd_lead_provider: cpd_lead_provider, started: :count_active_mentors_for_lead_provider })
+    @total_participants = ParticipantEventAggregator.call({ cpd_lead_provider: cpd_lead_provider })
+    @uplift_participants = ParticipantUpliftAggregator.call({ cpd_lead_provider: cpd_lead_provider })
+  end
+
+private
+
+  def lead_provider_scope
```
blackrat, line 27: Why not `ecf_lead_provider_scope` if we're trying to make that distinction between cpd, ecf and npq?

app/policies/finance_profile_policy.rb
```diff
@@ -5,6 +5,10 @@ def index?
     !!user&.finance?
   end
 
+  def show?
+    !!user&.finance?
```
blackrat, line 9: Why `!!`? That turns nil (which is falsey) into false and anything not falsey into `true`. I've never seen this as useful in a flag.

app/views/finance/lead_providers/show.html.erb
blackrat: Numbers of participants in a band is not a service fee. It feels more like a rename/refactor to get the correct structure of things when we decide what that is.

config/locales/devise.en-GB.yml
```diff
@@ -1,6 +1,6 @@
 # Additional translations at https://github.com/heartcombo/devise/wiki/I18n
 
-en:
+en-GB:
```
blackrat, line 3: I'm finding it hard to believe that this is right. Is there a reference somewhere? Surely en-GB should only apply to things with locale specialisation (like the spelling of specialisation) rather than to all English.

blackrat (changes requested):

app/policies/finance_profile_policy.rb
timabell: @plisovin :eyes: 
If I recall it was to cast `nil` to `false`

config/locales/devise.en-GB.yml
timabell: I agree that it would be good if it would fall-back, however it didn't in my experimenting (I tried it without the rename first and all the tests broke). Suggestions / references welcome.

app/views/finance/lead_providers/show.html.erb
timabell: @blackrat is going to rustle up a patch :+1:

app/views/finance/lead_providers/show.html.erb
kicferk1: Sure, numbers of participants is not a service fee, in the same way as email is not a user. 

But just like user fields are strongly related to user, I reckon number of participants is strongly related to the service fee. I mean, we surface `service_fee_total`, and `service_fee_per_participant` - if we really wanted to we could divide one by the other and get total participants this way. Likewise for band name - every method in the service fee calculator depends on the band, so I don't see much harm in surfacing it.

I think we can stick a comment in the service fee calculator saying something like "TODO: Figure out if this needs to be renamed to something more accurate", but I would not want that to be the justification we use for putting
```
@ecf_lead_provider.call_off_contract.bands[i].number_of_participants_in_this_band(@ecf_lead_provider.call_off_contract[:recruitment_target])
```
in our view - that seems much worse than making `service_fee` an imprecise name.

PR #768, opened 2021-07-28
CPDTP-361 CPDTP-360 CPDTP-357 CPDTP-356 remaining ecf and npq payment declarations

blackrat:
### Context

New payment focussed participant-declarations are required to support "retained-x" and "completed" declarations.

* https://dfedigital.atlassian.net/browse/CPDTP-361 - "Submit a Completed declaration for an ECF student"
* https://dfedigital.atlassian.net/browse/CPDTP-360 - "Submit a Retained-x declaration for an ECF student"
* https://dfedigital.atlassian.net/browse/CPDTP-357 - "Submit a Retained-x declaration for an NPQ student"
* https://dfedigital.atlassian.net/browse/CPDTP-356 - "Submit a completed declaration for an NPQ student"

- Add in the declarations types
- Change the structure to make it a "pick-list" of modules required for each
- Allow for each combination of ECF and NPQ + payment event_types to have their own schemas
- Update the factories to build the required DeclarationRecorder

### Review Checks


app/services/record_declarations/npq_course_proxy.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module RecordDeclarations
+  class NPQCourseProxy
+    include NPQ
+  end
+end
```
kicferk1, line 7: Why do we need this?

app/services/record_declarations/event_factory.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    private
+
+      def event_namespace_for_event(course)
+        event_identifiers[course.underscore.intern].to_s
+      end
+
+      def event_identifiers
+        started_identifiers.merge(retained_identifiers)
+      end
```
kicferk1, line 18: So is this equivalent to a hash like so:
```
{
  started: "Started",
  completed: "Started",
  retained_1: "Retained",
  retained_2: "Retained",
  retained_3: "Retained",
  retained_4: "Retained",
}
```

If so, can we please hard-code and inline it?

And perhaps a more important question: why is `completed` identified with `Started`?

app/services/record_declarations/mentor.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# frozen_string_literal: true
```
kicferk1, line 3: I don't think we need 2 of those

app/services/record_declarations/retained.rb
```diff
@@ -0,0 +1,17 @@
...
+    extend ActiveSupport::Concern
+
+    included do
+      extend RetainedClassMethods
+    end
+
+    module RetainedClassMethods
+      def required_params
+        %i[user_id cpd_lead_provider declaration_type declaration_date course_identifier evidence_held raw_event]
+      end
```
kicferk1, line 14: Those are very similar to `StartedCompletedClassMethods` ones. For my understanding, do we not expect `evidence_held` to be present on non-retained declarations?

app/services/record_declarations/event_factory.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module RecordDeclarations
+  class EventFactory
```
kicferk1, line 4: Is this file tested? Is `RecorderFactory` tested?

app/services/record_participant_declaration.rb
```diff
@@ -10,7 +13,8 @@ def call(params)
   end
 
   def call
-    RecordDeclarations::RecorderFactory.call(course_identifier).call(params)
+    recorder = "::RecordDeclarations::#{::RecordDeclarations::EventFactory.call(event)}::#{::RecordDeclarations::RecorderFactory.call(course_identifier)}".constantize
```
kicferk1, line 16: I think this should be thoroughly unit tested.

I would also be very tempted to replace `::RecordDeclarations::EventFactory.call(event)` and `::RecordDeclarations::RecorderFactory.call(course_identifier)` with helper methods in this file - I am not sure what value havign them separated in their own files brings. 

It certainly makes it harder to double check that the combinations of values they spit out match the names of the recorders we define.

spec/services/record_declarations/started/early_career_teacher_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require "rails_helper"
 
-RSpec.describe RecordDeclarations::ECF::EarlyCareerTeacher do
+RSpec.describe RecordDeclarations::Started::EarlyCareerTeacher do
```
kicferk1, line 5: Do we need a whole bunch more tests (for other events?)

app/services/record_declarations/event_factory.rb
blackrat: No. This is generated because when we change induction courses these will have different values.

app/services/record_declarations/retained.rb
blackrat: Yes. This is currently the biggest difference between them.

app/services/record_declarations/event_factory.rb
ethanmills: What do you mean by change induction course? What will the values be?

timabell (commented):
This PR/patch is composed of lots of small pieces, it is quite hard to understand how the pieces fit together and operate as a whole machine from a cold reading of just a diff or even the resultant files.

I think this could be mitigated by adding code-doc descriptions to every class/module in the machine explaining the intention behind each and how they fit together into the bigger piece.

It might also be worth adding a readme in one of the directories (or maybe an ADR) that explains in plain english the thinking that went into this and the way the architecture is supposed to work and evolve.

As it stands I wouldn't fancy coming to maintain this, and I could not really approve this PR without spending an very long time loading all the context from first principles and inspecting the running code locally.

app/services/record_declarations/event_factory.rb
blackrat: Completed and Started have equivalences at the moment. We're probably going to add more specifics since the declaration types are going to change based on other factors, but at the moment, the "started" and "completed" both are the same (with the exception of the stored declaration_type). These may need to be more explicit later, but for the purposes of standing up the API to support them, this is the correct overall shape for now.

etc/schema/0.3/npq/participant_declarations/started/create/request.json
```diff
@@ -5,7 +5,7 @@
       "declaration_type": "started",
       "declaration_date": "2015-02-14T02:21:34",
       "participant_id": "0cdc3c2f-20e3-4721-97f1-485b397ae7e9",
-      "course_type": "leading-teaching"
+      "course_type": "npq-leading-teaching"
```
timabell, line 8: I'm not sure what I'm missing, but I don't understand why this only mentions one of the 6 NPQ course types

app/services/record_declarations/event_factory.rb
timabell: @ethanmills .... @blackrat deleted the comment you responded to, much confusion ensues lol

app/services/record_declarations/event_factory.rb
```diff
@@ -0,0 +1,29 @@
...
+      def event_namespace_for_event(event)
+        event_identifiers[event.underscore.intern].to_s
+      end
+
+      def event_identifiers
+        started_identifiers.merge(retained_identifiers)
+      end
+
+      def started_identifiers
+        %i[started completed].index_with { |_event| "Started" }
```
timabell, line 21: https://www.bigbinary.com/blog/rails-6-adds-enumerable-index_with :eyes:

app/services/record_declarations/event_factory.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      def event_namespace_for_event(event)
+        event_identifiers[event.underscore.intern].to_s
+      end
+
+      def event_identifiers
+        started_identifiers.merge(retained_identifiers)
+      end
+
+      def started_identifiers
```
timabell, line 20: This highlights to me that we should *not* be conflating "started" and "completed" as they are only coincidentally the same. This conflation is visible externally in https://github.com/DFE-Digital/early-careers-framework/blob/13bebc716998e5ab8942ac721d92ef5e7b5ccdb5/swagger/v1/component_schemas/ECFParticipantStartedDeclaration.yml
We should follow up with further work to disentangle started and completed across the whole stack.

app/services/record_declarations/npq_course_proxy.rb
blackrat: The other option was to make the NPQ module callable directly.

blackrat: Or to embed the NPQCourseProxy inside the class that uses it. I'll do that.

app/services/record_declarations/event_factory.rb
kicferk1: I will need to spike into a nicer way of validating api requests early next week :P 

For time being I think this is ok. `started` and `completed` are different in the database, but have the same shape for the purpose of schema validation. And this code's purpose is to find the right schema validator. We might end up renaming them to something nicer later down the line.

kicferk1 (approved):
I think we will need to dig into the schema validation and potentially restructure it - I will have a go at it next week. For time being this looks reasonable enough - we can get the event types we were after, and they are validated.

etc/schema/0.3/npq/participant_declarations/started/create/request.json
blackrat: It's an example request to match the request_schema.json in the same directory

timabell: Ah I see. Thanks

PR #770, opened 2021-07-29
CPDRP-616: admin challenge partnership

ethanmills:
### Context
CPDRP-616 PR 2/3 
Give admins the ability to challenge a partnership on behalf of a school. There is no time limit on this at the moment, so we will need to revisit this once the declarations start coming in

### Review Checks

### DIFF
--- a/app/controllers/admin/schools/cohorts/challenge_partnership_controller.rb
+++ b/app/controllers/admin/schools/cohorts/challenge_partnership_controller.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module Admin
+  module Schools
+    module Cohorts
+      class ChallengePartnershipController < ::Admin::BaseController
+        skip_after_action :verify_policy_scoped
+        before_action :set_challenge_partnership_form, only: %i[confirm create]
+
+        def new
+          school = School.friendly.find params[:school_id]
+          cohort = ::Cohort.find_by(start_year: params[:id])
+          partnership = Partnership.find_by(school: school, cohort: cohort)
+          authorize partnership, :update?
+          @challenge_partnership_form = ChallengePartnershipForm.new(
+            school_name: school.name,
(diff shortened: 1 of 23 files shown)


app/controllers/admin/schools/cohorts/challenge_partnership_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+          @challenge_partnership_form.challenge!
+          set_success_message heading: "Induction programme has been changed"
+          redirect_to admin_school_cohorts_path
+        end
+
+      private
+
+        def set_challenge_partnership_form
+          @challenge_partnership_form = ChallengePartnershipForm.new(
+            params.require(:challenge_partnership_form).permit(:partnership_id, :challenge_reason, :lead_provider_name),
```
kicferk1, line 39: `lead_provider_name`? Don't we set that?

kicferk1: As in, why do we need this in params?

ethanmills: Just makes the confirm step a little nicer: this way the name is passed through from the create form, instead of just passing through the partnership ID, then finding the lead provider from the partnership and setting it on the form again

kicferk1: I wonder if the form should have helper methods `lead_provider_name`, `delivery_partner_name` - and all we would do is set the `partnership` (or `partnership_id`) on it. That should remove our need to set `lead_provider_name` explicitly (and implicitly via params). Thoughts?

ethanmills: Few extra queries this way, but worth it

kicferk1 (approved):

PR #771, opened 2021-07-29
Switching to use TeacherProfile model

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-759

This PR switches the `User.has_many :participant_profiles` association into a User.has_many :participant_profiles, through: :teacher_profile`. 

Next step: clean-up - remove participant_profile.user_id column and reduce number of participant-related logic from user.


app/controllers/api/v1/ecf_users_controller.rb
```diff
@@ -51,7 +51,7 @@ def email
       end
 
       def users
-        users = User.is_ecf_participant
+        users = User.where(id: ParticipantProfile::ECF.joins(:teacher_profile).select("teacher_profiles.user_id"))
```
ethanmills, line 54: Is suspect typing this everywhere is going to get boring. Not one for this PR, but would we be able to refactor this to some kind of helper or service object? C# would have ParticipantService that goes and gets the participants

app/models/teacher_profile.rb
```diff
@@ -4,5 +4,12 @@ class TeacherProfile < ApplicationRecord
   belongs_to :user
```
ethanmills, line 4: `touch: true`

app/models/user.rb
```diff
@@ -56,11 +56,11 @@ def mentor?
   end
 
   def npq?
-    npq_profiles.any?(&:active?)
+    npq_profiles.any?
```
ethanmills, line 59: This is a change in behaviour (and not a good one...)

app/models/user.rb
```diff
@@ -56,11 +56,11 @@ def mentor?
...
   end
 
   def participant?
-    early_career_teacher? || mentor?
+    participant_profiles.any?
```
ethanmills, line 63: This is a change in behaviour

app/models/user.rb
```diff
@@ -92,9 +92,7 @@ def cohort
   end
 
   def school
-    return early_career_teacher_profile.school if early_career_teacher?
-    return mentor_profile.school if mentor?
-    return induction_coordinator_profile.schools.first if induction_coordinator?
+    teacher_profile&.school # POTENTIAL CHANGE OF BEHAVIOUR!
```
ethanmills, line 95: Yeah this worries me a bit. I'd have to do some digging to be confident in this

db/seeds/test_data.rb
```diff
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-
 DOMAIN = "@digital.education.gov.uk" # Prevent low effort email scraping
```
ethanmills, line 2: Linter is unlikely to be happy with this?

spec/cypress/app_commands/scenarios/school_participants.rb
```diff
@@ -18,9 +18,31 @@
 coordinator = FactoryBot.create(:user, :induction_coordinator, email: "school-leader@example.com", full_name: "Ms School Leader")
 coordinator.induction_coordinator_profile.schools = [school]
 
-mentor = FactoryBot.create(:user, :mentor, full_name: "Abdul Mentor", id: "51223b41-a562-4d94-b50c-0ce59a8bb34d", school_cohort: school_cohort)
+mentor_profile = FactoryBot.create(
+  :participant_profile, :mentor,
```
ethanmills, line 22: Can you split these down on to new lines please? I find this really hard to read

spec/factories/teacher_profiles.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :teacher_profile do
+    user
+
+    trn { Random.alphanumeric(10) }
```
ethanmills, line 7: 7 digits

spec/factories/users.rb
```diff
@@ -33,19 +33,22 @@
     end
 
     trait :early_career_teacher do
-      early_career_teacher_profile
+      teacher_profile
+      # early_career_teacher_profile { association :early_career_teacher_profile, teacher_profile: teacher_profile }
```
ethanmills, line 37: Why?

spec/models/participant_profile_spec.rb
```diff
@@ -13,9 +14,10 @@
...
   it "updates the updated_at on the users" do
     freeze_time
-    user = create(:user, :mentor, updated_at: 2.weeks.ago)
+    user = create(:user, updated_at: 2.weeks.ago)
+    profile = create(:participant_profile, user: user)
 
-    user.mentor_profile.touch
+    profile.touch
     expect(user.reload.updated_at).to be_within(1.second).of Time.zone.now
   end
```
ethanmills: surprised this one works without the touch on user in teacher profile?

spec/models/user_spec.rb
```diff
@@ -144,14 +145,14 @@
     end
 
     it "is false when the ect profile is withdrawn" do
-      user = create(:early_career_teacher_profile, status: "withdrawn").user
+      user = create(:participant_profile, :ect, status: "withdrawn").user
```
ethanmills, line 148: Consistency please, either symbols everywhere or not at all

spec/models/user_spec.rb
```diff
@@ -351,26 +350,26 @@
...
-      admin = create(:user, :admin)
-      lp = create(:user, :lead_provider)
-      npq = create(:participant_profile, :npq).user
-
-      expect(User.is_ecf_participant).not_to include(admin)
-      expect(User.is_ecf_participant).not_to include(lp)
-      expect(User.is_ecf_participant).not_to include(npq)
-    end
-  end
+  # describe "scope :is_ecf_participant" do
```
ethanmills, line 353: remove?

app/controllers/api/v1/ecf_users_controller.rb
BroiSatse: Actually, the logic here is a bit odd. We are inside of `users_controller` - why do we filter users to ecf only? It would make sense for participants_controller, but inside user_controller it is a sign of violated boundaries...

app/models/user.rb
BroiSatse: Yup, corrected already! :)

app/models/user.rb
BroiSatse: I've restored original implementation. One thing that is odd here is the scenario when SIC is also a mentor - would those two profile always belong to the same school?

spec/models/participant_profile_spec.rb
BroiSatse: Indeed, will investiagate :)

spec/models/user_spec.rb
BroiSatse: Would prefer a trait, tbh :)

app/models/user.rb
BroiSatse: Original implementation restored. However, this method returns `false` for NPQ participant - which is not correct with the language

app/controllers/api/v1/ecf_users_controller.rb
kicferk1: Eeeh, we are in `ecf_users_controller`, so filtering them to be just ecf users if alright imo. That's a recent change, prior to that it was simply named poorly :)

BroiSatse: Ah, must have changed on merge with development :D So another note - there are no ECF users, there are ECF participants and that's what we should be serializing here. But that's for later

app/models/user.rb
ethanmills: Agreed, but we should change the language, not the behaviour

ethanmills (approved):

PR #772, opened 2021-07-30
Do not explicitly specify percy nonce to allow test re-runs

ethanmills:
Per documentation https://docs.percy.io/docs/parallel-test-suites#advanced-details the nonce should be
automatically provided by github actions. Fixing it to the run number
prevents re-running a failed test run

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,10 +6,19 @@ on:
       - '**'
 
 jobs:
+  create-nonce:
+    name: Create nonce
+    runs-on: ubuntu-20.04
+    outputs:
+      nonce: ${{ steps.generate-nonce.outputs.nonce }}
+    steps:
+      - id: generate-nonce
+        run: echo "::set-output nonce=$(tr -dc 0-9 </dev/urandom | head -c 18 ; echo '')"
+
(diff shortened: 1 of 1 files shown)


kicferk1 (approved):

ethanmills:
I feel lied to, Percy isn't working and Github clearly isn't setting the nonce

tonyheadford:
> I feel lied to, Percy isn't working and Github clearly isn't setting the nonce

Yes, I couldn't find any mention of Github actions supporting parallelising Percy, I set it to the `github.run_number` as that is supposed to be incremented and work across multiple invocations but it didn't by the sound of it?

ethanmills:
> > I feel lied to, Percy isn't working and Github clearly isn't setting the nonce
> 
> Yes, I couldn't find any mention of Github actions supporting parallelising Percy, I set it to the `github.run_number` as that is supposed to be incremented and work across multiple invocations but it didn't by the sound of it?

Sadly not https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#:~:text=run.-,this%20number%20does%20not%20change%20if%20you%20re-run%20the%20workflow%20run.,-github

PR #775, opened 2021-07-30
Delete flapping randomized feature flag tests

timabell:

Delete flapping randomized feature flag tests

### Context

Tests should be repeatable. For reasons that are not explained in the history these tests use a randomized set of factories, as a result this test file is randomly failing causing a drag on other development. This should never have been allowed into the main branch.

In order to restore sanity to development this patch deletes the offending randomization and the tests that depend on it. I shall leave it to the author of these tests to re-do the coverage in a more stable way.

### DIFF
--- a/spec/services/feature_flag_spec.rb
+++ b/spec/services/feature_flag_spec.rb
@@ -31,25 +31,6 @@ RSpec.describe FeatureFlag do
         expect(described_class.active?(feature_name, for: double)).to be true
       end
     end
-
-    context "with object given via for: keyword" do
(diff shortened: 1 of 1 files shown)


ethanmills (changes requested):
We can't remove these without replacing them with something else

blackrat:
The FactoryBot linter also seems to break on the factories that these use. Running against all factories highlights these issues. Random isn't normally a good choice, but I agree with @ethanmills that we should really replace them rather than just pulling them.

PR #778, opened 2021-08-02
CPDTP-329 T&P: Rails validations for declarations

kicferk1:
### Context
Track and Pay uses several layers of validation - params checking in controller, schema checking in 'recorders' and model validation on the model. 

It looks like the model validation is doing largely the same job as schema validation, and it can be brought up to speed to do all of it. In the long term, I reckon we should rely on Rails validations more - in principle if a dev in rails console needs to deal with support ticket, it will make their life much easier if the validations are on the model level. 

https://dfedigital.atlassian.net/browse/CPDTP-329

Give a tiny upgrade to Rails validations to handle valid course types for ects.
Swap schema validation for Rails validation in the recorders.
Remove things relating to schema validation in recorders. 

### Guidance to review
I made sure the existing tests pass. I don't know if all the behaviour of schema validations was captured in automated tests. If there is anything that we are losing with it, I would love to know, so that I can bring it back (or at least know that it is an issue, so we can judge how serious it is). 

### Testing
Made sure existing tests pass.

### Review Checks

It should not change behaviour.


app/services/record_declarations/base.rb
```diff
@@ -18,23 +16,17 @@ def call(params)
       def not_implemented_error
         raise NotImplementedError, "Method must be implemented"
       end
-
-      def schema_validation_params
-        { version: "0.3" }
```
blackrat, line 23: How are you going to handle versioning?

blackrat (changes requested):

app/services/record_declarations/base.rb
kicferk1: What do we need versioning for exactly? At the moment?

kicferk1: In principle, if (when?) we add `v2` version of the api, we will either need to add a v2 declaration model, or extract v1 validation into their own layer, and add v2 validation in that layer.

We could think of doing something like that now, I just don't know what the value of it is.

blackrat:
I really don't see what this gives us. Rails validations were added to ensure that any console level changes don't break the underlying data structures, but as an enhancement rather than a replacement for schema level validations.

It's removing working schema level validation which uses versioned schemas which will later enable us to verify the stored declarations against whichever version they were recorded against. The next step in this was to start recording the version against the schema, which admittedly would mainly have been a 1.0 feature and easier to keep in step with the API docs schemas.

Not being able to compare those with the swagger documentation easily is going to cause us more pain rather than less.

kicferk1:
> I really don't see what this gives us. Rails validations were added to ensure that any console level changes don't break the underlying data structures, but as an enhancement rather than a replacement for schema level validations.

Well, having an 'enhancement validation' in model layer is a problem, because treating it as enhancement causes validations to diverge. `valid_courses` on models allowed an ECT to do the mentor course, and vice-versa. Keeping the enhancement validations up to date with real validations is a cost. 

What this gives us is:
1. Single source of truth of what is a valid declaration.
2. Following the same pattern of validation that the rest of the repository does. Which is `ActiveModel::Model` validations - be it on models or on forms, the syntax and capabilities of this style of validation are much more standard (for this team and codebase, and probably for any Rails codebase) than json schema validations.
3. Easily extensible model-level validations. Schema can check that a field is not in an enum, but it cannot check that, say, a participant id is present in our database. Or that the event type makes sense in context of what has been declared for the participant. That's why we have to check it manually: `raise ActionController::ParameterMissing, [I18n.t(:invalid_participant)] unless participant?`. But Rails validations can refer back to our database, with minimal effort. 

> It's removing working schema level validation which uses versioned schemas which will later enable us to verify the stored declarations against whichever version they were recorded against. The next step in this was to start recording the version against the schema, which admittedly would mainly have been a 1.0 feature and easier to keep in step with the API docs schemas.

I think we should worry about versioning declarations and schemas when we need to support multiple versions. There are a lot of ways this can be done, and at the moment we have no way of knowing which is the right one.

> Not being able to compare those with the swagger documentation easily is going to cause us more pain rather than less.

As I understand it, rswag does that for us - it fires of a request to our application, and checks that the outputs of our application match the schema (and response code, and so on) that we told it about. When I go to `spec/docs/participant_declarations_spec.rb` and change something about the expected response (say, schema), it fails. So are I reckon comparing our validations with the swagger documentation is a matter of writing enough rswag specs - which I think you have been doing so far anyway :)

kicferk1:
I decided to put more effort into validating in the services. Conveniently, the current setup is pretty good at making a service for each type of declaration - so I decided to add validation to those services.

Also... our retained tests tested 'Started' events 😭

app/services/record_declarations/base.rb
kicferk1: Just to follow up on it, if we had to add versioning, we would need to have `v1` and `v2` versions of the services that record the declarations. Those services could then have different validation.

app/models/participant_declaration.rb
```diff
@@ -5,7 +5,7 @@ class ParticipantDeclaration < ApplicationRecord
   belongs_to :cpd_lead_provider
   belongs_to :user
 
-  validates :course_identifier, inclusion: { in: :valid_courses }
+  validates :course_identifier, inclusion: { in: :valid_courses, message: "The property '#/course_identifier' must be an available course to '#/participant_id'" }
```
blackrat, line 8: Should we allow this to be missing here, since the presence check following this should take precedence.

app/services/record_declarations/base.rb
```diff
@@ -1,12 +1,37 @@
...
-require "json_schema/validate_body_against_schema"
-
 module RecordDeclarations
   class Base
-    attr_accessor :params
+    include ActiveModel::Model
+
+    attr_accessor :course_identifier, :user_id, :raw_event, :lead_provider_from_token, :declaration_date, :declaration_type, :evidence_held
+
+    validates :course_identifier, inclusion: { in: :valid_courses_for_user, message: "The property '#/course_identifier' must be an available course to '#/participant_id'" }
```
blackrat, line 9: I18n?

app/services/record_declarations/base.rb
```diff
@@ -1,12 +1,37 @@
...
+    validates :declaration_type, inclusion: { in: :valid_declaration_types, message: "The property '#/declaration_type' must be an available for course_identifier '#/course_identifier'" }
+    validates :course_identifier, presence: { message: "The property '#/course_identifier' must be present" }
+    validates :declaration_date, presence: { message: "The property '#/declaration_date' must be present" }
+    validates :declaration_type, presence: { message: "The property '#/declaration_type' must be present" }
+    validates :user, presence: { message: "The participant must be exist" }
```
blackrat, line 14: Not good english

app/services/record_declarations/base.rb
```diff
@@ -1,12 +1,37 @@
...
-
 module RecordDeclarations
   class Base
-    attr_accessor :params
+    include ActiveModel::Model
+
+    attr_accessor :course_identifier, :user_id, :raw_event, :lead_provider_from_token, :declaration_date, :declaration_type, :evidence_held
+
+    validates :course_identifier, inclusion: { in: :valid_courses_for_user, message: "The property '#/course_identifier' must be an available course to '#/participant_id'" }
+    validates :declaration_type, inclusion: { in: :valid_declaration_types, message: "The property '#/declaration_type' must be an available for course_identifier '#/course_identifier'" }
```
blackrat, line 10: not good english

app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -6,7 +6,7 @@ class ParticipantDeclarationsController < Api::ApiController
       include ApiTokenAuthenticatable
 
       def create
-        params = HashWithIndifferentAccess.new({ raw_event: request.raw_post, cpd_lead_provider: cpd_lead_provider }).merge(permitted_params["attributes"] || {})
+        params = HashWithIndifferentAccess.new({ raw_event: request.raw_post, lead_provider_from_token: cpd_lead_provider }).merge(permitted_params["attributes"] || {})
```
blackrat, line 9: Not sure I understand the name change.  I'd probably have changed the lead_provider_from_database rather than this one.

app/services/record_declarations/base.rb
```diff
@@ -1,12 +1,37 @@
...
+    validates :course_identifier, presence: { message: "The property '#/course_identifier' must be present" }
+    validates :declaration_date, presence: { message: "The property '#/declaration_date' must be present" }
+    validates :declaration_type, presence: { message: "The property '#/declaration_type' must be present" }
+    validates :user, presence: { message: "The participant must be exist" }
+    validates :lead_provider_from_token, presence: { message: "The lead provider must be present" }
```
blackrat, line 15: Is this the same? It wasn't supposed to be just a presence check. I'm not seeing where it's checking that the participant is attached to that lead_provider or are we allowing the course check to do that?

spec/factories/participant_declaration.rb
```diff
@@ -8,20 +8,23 @@
     declaration_type { "started" }
 
     factory :ect_participant_declaration do
+      user { create(:user, :early_career_teacher) }
```
blackrat, line 11: Should we be creating users like this as part of the declaration?

blackrat (changes requested):

app/services/record_declarations/base.rb
```diff
@@ -1,11 +1,36 @@
 # frozen_string_literal: true
 
-require "json_schema/validate_body_against_schema"
-
 module RecordDeclarations
   class Base
+    include ActiveModel::Model
+
+    attr_accessor :course_identifier, :user, :cpd_lead_provider, :declaration_date, :declaration_type, :evidence_held
```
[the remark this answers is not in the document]
kicferk1: Outdated, I reckon it does just that now :)

app/services/record_declarations/base.rb
```diff
@@ -45,37 +63,24 @@ def call
...
     def create_record!
       ActiveRecord::Base.transaction do
-        declaration_type.create!(params.slice(*self.class.required_params)).tap do |participant_declaration|
+        declaration_type.create!(
+          course_identifier: course_identifier,
+          declaration_date: declaration_date,
+          declaration_type: declaration_type,
+          cpd_lead_provider: cpd_lead_provider,
+          user: user,
+          evidence_held: evidence_held,
```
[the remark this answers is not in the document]
kicferk1: I think the latest version does that?

app/models/participant_declaration.rb
kicferk1: I would not want it to be missed. I am happy to look at re-ordering it (and adding more meaningful tests) in a future PR.

app/services/record_declarations/base.rb
kicferk1: I hate i18n with a passion, but fine, if you want it a lot I can put it in there.

spec/factories/participant_declaration.rb
kicferk1: I don't understand your comment. What do you mean 'as part of the declaration'?

app/services/record_declarations/base.rb
kicferk1: This does not check that this was the right provider for the user.

This checks that the provider from token exists - cause if it doesn't the creation of our record will fail.

app/controllers/api/v1/participant_declarations_controller.rb
kicferk1: I took it from
```
    def lead_provider_from_token
      params[:cpd_lead_provider]
    end
```

I think it explains what it is very well, and I am not sure this PR is the best place to change it.

app/services/record_declarations/base.rb
kicferk1: And the old check that the lead providers match is still in the same place it used to be in.

spec/factories/participant_declaration.rb
kicferk1: In any case, I wonder if https://dfedigital.atlassian.net/browse/CPDTP-391 is not the better ticket to do it under.

kicferk1: Oh, I think I get it. Do you mean "why does participant declaration create a user?"? 

The answer to that is "because without a user participant declaration makes no sense". And the extension is "without ect user, ect declaration makes no sense".

app/models/participant_declaration.rb
kicferk1: Did you mention that as part of https://dfedigital.atlassian.net/browse/CPDTP-362 you would like to have a look at how valuable our errors are?

blackrat (approved):

PR #779, opened 2021-08-02
Ignoring user_id on participant profiles

BroiSatse:
The column is unused now

### DIFF
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -27,7 +27,7 @@ class ParticipantProfile < ApplicationRecord
   class_attribute :validation_steps
   self.validation_steps = []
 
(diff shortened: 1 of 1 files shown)


ethanmills (approved):

ethanmills (changes requested):
Is it safe to do this on NPQ profiles?

BroiSatse:
It should be, all NPQ profiles belong to teacher_profile and (and to user through it) just as any other participation profile. The logic syncing profile with NPQ validation data record is coming in next PR. Ignored columns are pretty silly to be honest - they only remove column from attributes, but all setters and getters continue to working...

ethanmills (approved):

PR #784, opened 2021-08-02
Fix API URLs on API usage guide

jamesrose:
### Context

These links are currently 404ing. They should be the API paths for each environment.


ethanmills:
Are they definitely meant to link there, not to https://ecf-sandbox.london.cloudapps.digital/api-docs/index.html ?

ethanmills (approved):
Better than what it is currently definitely

PR #785, opened 2021-08-02
Fix production compilation and performance webpacker settings

jamesrose:
### Context

A change was introduced in bea7a4ad827179dfcd63d8811fb3a10fe2c60956 to add the new sandbox environment to webpacker. It pushed two config settings into the sandbox environment instead of the production environment.

This PR pushes the values back into the scope of production which is then inherited by the other non-test environments.

### DIFF
--- a/config/webpacker.yml
+++ b/config/webpacker.yml
@@ -81,14 +81,14 @@ test:
 production: &production
   <<: *default
 
-sandbox: &production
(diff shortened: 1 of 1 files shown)


ethanmills (approved):
There's an assets-precompile step in the dockerfile that I'm not sure if correct too, this has reminded me

PR #786, opened 2021-08-02
Switched school participant list to load participant profiles instead of users

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-598

Admin part has been tackled with manual admin validation journey

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -22,6 +22,8 @@ require "paper_trail/frameworks/rspec"
 # require features_helper after support files have been loaded
 require "features_helper"
 
+Capybara.server = :puma, { Silent: true }
+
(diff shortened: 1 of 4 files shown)


spec/rails_helper.rb
```diff
@@ -22,6 +22,8 @@
 # require features_helper after support files have been loaded
 require "features_helper"
 
+Capybara.server = :puma, { Silent: true }
+
```
kicferk1, line 26: What's this?

ethanmills (approved):

spec/rails_helper.rb
BroiSatse: This removes the extra output on start of Capybara server in test, which is just ugly and unnecessary

PR #788, opened 2021-08-03
Removed participant_profile user_id column

BroiSatse:
`user_id` column on participant_profiles are now obsolete, as ParticipantProfile to User association now goes through TeacherProfile.

NOTE: This is possible best to be merged in a week time when we are 100% sure there are no issues with the new association and we don't need a fail-safe mechanism to migrate missing data.


ethanmills:
Yes, I'd be happier waiting a little while for this

ethanmills:
Happy with this in principle, just not yet. I don't think there's much chance of rot if we leave this unmerged for a week or so

BroiSatse:
@ethanmills One potential issue would be someone writing a new code relying on user_id column. But that's what reviews are for :)

ethanmills (approved):

BroiSatse:
Closing to re-open

PR #790, opened 2021-08-03
NPQ API to accept/reject an application

asmega:
### Context

- As part of the end of the NPQ lifecycle before we hit track and pay we need for providers to acknowledge NPQ applications

- NPQ applications start out in a starting state of `pending`
- Providers can then hit one of two endpoints to `accept` or `reject`
- This will then change the `status` to `accepted` or `rejected` respectively
- Once an application has been accepted or rejected it cannot change state
- The first provider to accept an application will cause all other NPQ applications on the user to become rejected
- Swagger docs have been added for these endpoints
- Also included some other swagger doc changes related to NPQ application that were not correctly documented

### Guidance to review

- Do we only allow this once? Can a provider accept then reject an application?

### Review Checks

### DIFF
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -32,11 +32,14 @@ module Api
 
       def accept
         profile = npq_lead_provider.npq_profiles.includes(:user, :npq_course).find(params[:id])
+        other_profiles = NPQValidationData.where(profile: (profile.user.npq_profiles - [profile.profile]))
 
-        if profile.update(status: "accepted")
-          render json: NPQApplicationSerializer.new(profile).serializable_hash
-        else
-          render json: { errors: Api::ErrorFactory.new(error: profile.errors).call }, status: :bad_request
+        ActiveRecord::Base.transaction do
+          if profile.update(status: "accepted") && other_profiles.update(status: "rejected")
+            render json: NPQApplicationSerializer.new(profile).serializable_hash
+          else
+            render json: { errors: Api::ErrorFactory.new(error: profile.errors).call }, status: :bad_request
+          end
         end
       end
 
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -20,6 +20,22 @@ module Api
         end
       end
 
+      def reject
+        profile = npq_lead_provider.npq_profiles.includes(:user, :npq_course).find(params[:id])
+
+        profile.update!(status: "rejected")
+
+        render json: NPQApplicationSerializer.new(profile).serializable_hash
+      end
+
+      def accept
+        profile = npq_lead_provider.npq_profiles.includes(:user, :npq_course).find(params[:id])
+
+        profile.update!(status: "accepted")
+
+        render json: NPQApplicationSerializer.new(profile).serializable_hash
+      end
+
     private
 
       def npq_lead_provider
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -23,17 +23,21 @@ module Api
       def reject
         profile = npq_lead_provider.npq_profiles.includes(:user, :npq_course).find(params[:id])
 
-        profile.update!(status: "rejected")
-
-        render json: NPQApplicationSerializer.new(profile).serializable_hash
+        if profile.update(status: "rejected")
+          render json: NPQApplicationSerializer.new(profile).serializable_hash
+        else
+          render json: { errors: Api::ErrorFactory.new(error: profile.errors).call }, status: :bad_request
+        end
       end
 
       def accept
         profile = npq_lead_provider.npq_profiles.includes(:user, :npq_course).find(params[:id])
 
-        profile.update!(status: "accepted")
-
-        render json: NPQApplicationSerializer.new(profile).serializable_hash
+        if profile.update(status: "accepted")
+          render json: NPQApplicationSerializer.new(profile).serializable_hash
+        else
+          render json: { errors: Api::ErrorFactory.new(error: profile.errors).call }, status: :bad_request
+        end
       end
 
     private
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -23,10 +23,10 @@ module Api
       def reject
         profile = npq_lead_provider.npq_profiles.includes(:user, :npq_course).find(params[:id])
 
-        if profile.update(status: "rejected")
+        if profile.update(lead_provider_approval_status: "rejected")
           render json: NPQApplicationSerializer.new(profile).serializable_hash
         else
-          render json: { errors: Api::ErrorFactory.new(error: profile.errors).call }, status: :bad_request
+          render json: { errors: Api::ErrorFactory.new(model: profile).call }, status: :bad_request
         end
       end
 
@@ -35,10 +35,10 @@ module Api
         other_profiles = NPQValidationData.where(profile: (profile.user.npq_profiles - [profile.profile]))
 
         ActiveRecord::Base.transaction do
-          if profile.update(status: "accepted") && other_profiles.update(status: "rejected")
+          if profile.update(lead_provider_approval_status: "accepted") && other_profiles.update(lead_provider_approval_status: "rejected")
             render json: NPQApplicationSerializer.new(profile).serializable_hash
           else
-            render json: { errors: Api::ErrorFactory.new(error: profile.errors).call }, status: :bad_request
+            render json: { errors: Api::ErrorFactory.new(model: profile).call }, status: :bad_request
           end
         end
       end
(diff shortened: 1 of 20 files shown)


db/schema.rb
```diff
@@ -366,6 +366,7 @@
     t.boolean "eligible_for_funding", default: false, null: false
     t.text "funding_choice"
     t.text "nino"
+    t.text "status", default: "pending", null: false
```
kicferk1, line 369: There are too many statuses to my liking :( Participant profiles have statuses, too, active or withdrawn.

Can we name it `lead_provider_approval_status`, or something like that?

swagger/v1/component_schemas/NpqApplicationAttributes.yml
```diff
@@ -9,11 +9,9 @@ required:
   - teacher_reference_number
   - teacher_reference_number_validated
   - school_urn
-  - headteacher_status
   - eligible_for_funding
-  - funding_choice
-  - course_id
-  - course_name
```
kicferk1, line 16: Why are we cutting them out? were they never there in the first place, or did something change?

kicferk1 (commented):
Looks pretty good to me, I am just a bit worried about potential confusion caused by naming things `status` on various different models.

swagger/v1/component_schemas/NpqApplicationAttributes.yml
asmega: so there are 2 things happening here:
- the documentation is out of data.`course_id` and `course_name` no longer exists and now comes under `course_indentifier`
- `headteacher_status` and `headteacher_status` are valid attributes but if you scroll up this is under `required`. and the documentation is incorrect that these fields are not required and may be omitted from the response. This is because `headteacher_status` is only available if they do a headship NPQ and `funding_choice` is only available if they are not `eligible_for_funding`

kicferk1: Ah, ok. That makes sense, thanks :)

db/schema.rb
asmega: this has now been renamed to `lead_provider_approval_status` for lead provider facing side eg in the api it is still displayed as `status`

app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -20,6 +20,29 @@ def index
         end
       end
 
+      def reject
+        profile = npq_lead_provider.npq_profiles.includes(:user, :npq_course).find(params[:id])
```
ethanmills, line 24: This would make a good `policy_scope` I expect

app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -20,6 +20,29 @@ def index
...
+          render json: { errors: Api::ErrorFactory.new(model: profile).call }, status: :bad_request
+        end
+      end
+
+      def accept
+        profile = npq_lead_provider.npq_profiles.includes(:user, :npq_course).find(params[:id])
+        other_profiles = NPQValidationData.where(profile: (profile.user.npq_profiles - [profile.profile]))
+
+        ActiveRecord::Base.transaction do
+          if profile.update(lead_provider_approval_status: "accepted") && other_profiles.update(lead_provider_approval_status: "rejected")
```
ethanmills, line 38: Do we set papertrail whodunnit anywhere? Would be good to be able to work out how something got rejected

spec/requests/api/v1/npq_applications_spec.rb
```diff
@@ -167,4 +165,97 @@
...
+      let!(:other_npq_profile) { create(:npq_validation_data, npq_lead_provider: npq_lead_provider, user: user) }
+
+      it "rejects all other NPQs" do
+        post "/api/v1/npq-applications/#{npq_profile.id}/accept"
+
+        expect(other_npq_profile.reload.lead_provider_approval_status).to eql("rejected")
+      end
+    end
+
+    context "application has been rejected" do
```
ethanmills, line 241: Is it possible to apply to another once you have an accepted application? If so, it would be good to see a test where we 400 when one of the other applications is accepted

ethanmills (approved):

app/controllers/api/v1/npq_applications_controller.rb
kicferk1: Does papertrail work with `current_user` by default? If so, `ApiTokenAuthenticatable` should (I think) take care of it.

kicferk1 (approved):

kicferk1 (approved):

app/controllers/api/v1/npq_applications_controller.rb
asmega: `whodunnit` was not set for the api. I have added to `ApiTokenAuthenticatable`. Also the format of the `whodunnit` for the api is now a `GlobalID` thing

app/services/api/error_factory.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+
+    def call
+      error.errors.map do |e|
+        {
+          title: "#{e.attribute.to_s.humanize} is #{e.type.to_s.humanize.downcase}",
+          detail: e.message,
+        }
+      end
+    end
```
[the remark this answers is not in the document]
asmega: This is a first attempt which kind of works. You've commented on an older version. It now takes the model instead of errors as is needed if we want to do any `i18n` calls. Improvements are welcomed

app/models/npq_validation_data.rb
```diff
@@ -4,6 +4,8 @@ class NPQValidationData < ApplicationRecord
   # TODO: Rename table
   self.table_name = "npq_profiles"
 
+  has_paper_trail
```
ethanmills, line 7: We don't want this papertrailed. This contains all the PII that we're going to delete

ethanmills (changes requested):

spec/requests/api/v1/npq_applications_spec.rb
asmega: Yes it is possible but It would actually start causing problems. We think the chances of this happening are low. If the problem does arise we'll have investigate by hand and resolve with an offline process then reconcile after that by hand. Once we start figuring out cohorts it should reduce the problem over time but it'll still be a problem

asmega: having now thought about this there once something is set it should not be able to change. i will make sure that is the case

app/controllers/concerns/api_token_authenticatable.rb
```diff
@@ -26,6 +27,14 @@ def current_user
 
 private
 
+  def set_custom_paper_trail_whodunnit
+    PaperTrail.request.whodunnit = if current_user.is_a?(ApplicationRecord)
+                                     current_user.to_global_id.to_s
```
ethanmills, line 32: Do we set it to a global id anywhere else, or just in the API?

app/models/npq_validation_data.rb
asmega: can we set those as a dependency so when the validation data is deleted so it the paper trail versions? my understanding at the moment is that `NPQValidationData` is the "profile" as we not sorted this part out yet

app/controllers/concerns/api_token_authenticatable.rb
asmega: just api at the moment. I can make this change in the "normal" app in another PR once i've finished this

app/models/npq_validation_data.rb
asmega: I've pulled the commit out. I'll move part of it into another PR once I'm done here

app/controllers/concerns/api_token_authenticatable.rb
asmega: This will now go into another PR

app/controllers/api/v1/npq_applications_controller.rb
asmega: reverted the paper trail commit and will go into another PR

spec/requests/api/v1/npq_applications_spec.rb
asmega: i've added an additional test (turns out it was already implemented) where once an application is accepted it cannot then be rejected by accepting a future application.

> Is it possible to apply to another once you have an accepted application? If so, it would be good to see a test where we 400 when one of the other applications is accepted

I think i misinterpreted this and was thinking of another issue. The answer is yes but I don't see why we would 400 is one of the other applications is accepted?

app/controllers/api/v1/npq_applications_controller.rb
asmega: do we need pundit here? we're just affecting objects the token owns. theres no different access levels of users here or the need to deal with permissions from what I can tell

spec/requests/api/v1/npq_applications_spec.rb
ethanmills: Because presumably you can't have two accepted applications?

ethanmills (approved):

spec/requests/api/v1/npq_applications_spec.rb
asmega: i think we'd have have take this an edge case and deal with it offline at a later date

PR #791, opened 2021-08-03
Check first name only for api/v1/participant-validation

asmega:
### Context

- Policy has given the go ahead to auto validate TRNs based on first name and not the whole name

- `ParticipantValidationService` can now be passed a config hash to check first name only
- This feature is disabled by default and will check the whole name by default
- The api endpoint `api/v1/participant-validation` checks first name now

### Guidance to review

- none

### Testing

- See how to test in a review app below

### Review Checks

- Use existing `NPQRegistrationApiToken` or create a new one
- See data extract spreadsheet for test data
```
curl -H "Authorization: Bearer TOKEN" https://ecf-review-pr-791.london.cloudapps.digital/api/v1/participant-validation/TRN?full_name=FULL_NAME_WITH_WONRG_LAST_NAME&date_of_birth=DOB&nino=
```
- Should return record regardless of last names

### DIFF
--- a/app/services/participant_validation_service.rb
+++ b/app/services/participant_validation_service.rb
@@ -51,7 +51,7 @@ private
     matches += 1 if trn_matches
 
     name_matches = if check_first_name_only?
-                     full_name.split(" ").first.to_s == dqt_record[:full_name].split(" ").first.to_s
+                     full_name.split(" ").first == dqt_record[:full_name].split(" ").first
                    else
                      full_name == dqt_record[:full_name]
                    end
--- a/app/services/participant_validation_service.rb
+++ b/app/services/participant_validation_service.rb
@@ -1,17 +1,18 @@
 # frozen_string_literal: true
 
 class ParticipantValidationService
-  attr_reader :trn, :nino, :full_name, :date_of_birth
+  attr_reader :trn, :nino, :full_name, :date_of_birth, :config
 
-  def self.validate(trn:, full_name:, date_of_birth:, nino: nil)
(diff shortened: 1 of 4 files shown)


app/services/participant_validation_service.rb
```diff
@@ -44,8 +49,15 @@ def matching_record(trn:, nino:, full_name:, dob:)
     matches = 0
     trn_matches = padded_trn == dqt_record[:teacher_reference_number]
     matches += 1 if trn_matches
-    name_matches = full_name == dqt_record[:full_name]
+
+    name_matches = if check_first_name_only?
+                     full_name.split(" ").first.to_s == dqt_record[:full_name].split(" ").first.to_s
```
kicferk1, line 54: I wonder, do we need `to_s` here? I would hope `split` would already give us a string.

I guess that makes sure neither value is `nil`, just not sure if it's worth it.

kicferk1 (approved):

app/services/participant_validation_service.rb
asmega: hmm i think i glossed over this so it looks more like a mistake. i will remove the `to_s`. If one or both values are `nil` it shouldn't affect `==`.

asmega: FYI I've now removed the `to_s` in a new commit

kicferk1 (approved):

ethanmills (approved):
Thanks for sorting this. Just had confirmation that this applies to R&P too

asmega:
Sorry for the delay. still waiting for sign off. If you need this in let me know then I can remove the controller change and add that line later and you will have the updated service

ethanmills:
Blocked per Phil's comment ☝️

PR #794, opened 2021-08-04
CPDTP-246 Add Payment Breakdown pages to Finance

blackrat:
### Context

This ticket enables the display of payment breakdowns for an ECF Lead Provider.

Change the aggregation passed through from single items to a hash of different types (:ects, :mentors, :all) which are used for different parts of the output.
Create header, service_fee, output_payment and other_fee objects and collections associated with them to allow for object and collection rendering in a clean manner.
Defer rounding to display and implement a standard method to set currency and precision without needing a locale change.

### Guidance to review

- Submit some valid ecf declarations for a particular lead provider
- Spin up the app and login as a Finance user.
- Visit the page for the lead provider
- If required, the figures can be checked against the rake payment breakdown

### Review Checks


spec/services/calculation_orchestrator_spec.rb
```diff
@@ -4,58 +4,112 @@
...
+      when Array
+        # I know this is a hash, if it wasn't this wouldn't work and you'd have to handle the values more elegantly
+        hash[k].each do |service_fee_hash|
+          with_rounding(service_fee_hash, &blk)
+        end
+      else
+        next
+      end
+    end
+  end
```
kicferk1, line 111: What is this for? It is used to set the precision of our results to be 2 every time, apparently. Should we perhaps change the `normal_outcome` to have the correct precision? Or the orchestrator to spit out numbers with precision 2?

app/services/calculation_orchestrator.rb
```diff
@@ -8,12 +8,57 @@ class << self
...
+                 calculator: ::PaymentCalculator::Ecf::PaymentCalculation)
+    @cpd_lead_provider = cpd_lead_provider
+    @contract = contract
+    @aggregator = aggregator
+    @calculator = calculator
+  end
+
+  def aggregators(event_type:)
+    @aggregators ||= Hash.new { |hash, key| hash[key] = aggregate(aggregation_type: key, event_type: event_type) }
+  end
```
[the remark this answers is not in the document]
blackrat: Yes. It's creating the hash with an initialiser that knows how to get the values of any keys that it doesn't have. Since aggregations could be an expensive operation, they should only be called when required.

spec/services/calculation_orchestrator_spec.rb
blackrat: It's a recursive descent method to walk through the payment breakdown hash and convert any BigDecimal values to a specific rounding. Only used for this breakdown hash (which is why the comment) and these tests (which is why the location). Since this is financial data, it shouldn't be rounded until we need to display it, but the tests don't need to work to 15 digits of precision.

This is set to round to 2 for now. It's only for testing purposes. The display output will be 2 or 0 digits, depending on requirements. (Options for these are currently being discussed.)

.rspec
```diff
@@ -1,4 +1,3 @@
 --require rails_helper
 --require spec_helper
---require turnip/rspec
```
kicferk1, line 3: Can we also uninstall the gem? I think @tonyheadford will be very happy to hear about that

blackrat: Yes. I was looking at producing some features, but it looks like RSpec are implementing their own.

kicferk1 (approved):

PR #797, opened 2021-08-04
Ensure only one TeacherProfile per user

ethanmills:
### Context
Each user should have at most one teacher profile. Add a database constraint to ensure that.


db/schema.rb
```diff
@@ -585,8 +585,11 @@
     t.string "administrative_district_code"
     t.string "administrative_district_name"
     t.string "slug"
+    t.boolean "section_41_approved", default: false, null: false
```
asmega, line 588: are these suppose to be here? as one would expect the migration for these to be present in this pr

ethanmills: uh oh

spec/cypress/integration/nominations/NominateInductionCoordinator.feature
```diff
@@ -56,7 +56,7 @@ Feature: Nominate induction tutor
     And the page should be accessible
     And percy should be sent snapshot called "Start nominations name different"
 
-    Given user was created as "teacher" with email "different-user-type@example.com"
+    Given user was created with email "different-user-type@example.com"
```
kicferk1, line 59: Out of curiosity, did this start to fail here, or is it a cheeky fix of an incorrect test?

kicferk1 (approved):

spec/cypress/integration/nominations/NominateInductionCoordinator.feature
ethanmills: This started to fail here. This line created a teacher profile for the user, then the next line creates another

PR #798, opened 2021-08-05
CPDTP-251: Add schedules

kicferk1:
### Context

https://dfedigital.atlassian.net/browse/CPDTP-251

We want to make sure declarations are made at the correct time. And we want to have a way of representing the time the payments get made.

Add schedule and milestone model.
Add basic unit tests.
Add schedule seeds for known schedules (https://docs.google.com/presentation/d/1HBVEZabnGuakaAKJ6d5AGJthC7Bl21cex4MN5K8rus4/edit#slide=id.ge582d819d0_0_304)

### Guidance to review
I think all participants profiles will need to be pointing at a schedule. By default probably September standard one. We might need to worry about how exactly we know which schedule they are doing - is it IC's job, or participant's job, or LP's job to tell us. Not sure, not in scope for this ticket, just something to be aware. 

I think all participant declarations will be implicitly made against milestones - they have `declaration_date` on them, and from that we can work out which milestone they belong to (a Milestone with the earliest milestone date that is after the declaration date). I don't think we should store it in the database for time being. perhaps at a future time we will for easier querying, but for now it is just something to be aware of.

### Testing
Unit tests for models

### Review Checks

You can't. I guess if you are a dev you can ssh into review app and check out the schedules and milestones in Rails console.

### DIFF
--- a/app/models/milestone.rb
+++ b/app/models/milestone.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class Milestone < ApplicationRecord
+  belongs_to :schedule
+end
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -3,6 +3,10 @@
 class ParticipantProfile < ApplicationRecord
   has_paper_trail
   belongs_to :teacher_profile, touch: true
+
+  # TODO: Back-fill, so that every profile has a schedule
+  belongs_to :schedule, optional: true, touch: true
+
   has_one :user, through: :teacher_profile
 
   has_many :validation_decisions, class_name: "ProfileValidationDecision"
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class Schedule < ApplicationRecord
+  has_many :milestones
+  has_many :participant_profiles
+end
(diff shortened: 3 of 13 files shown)


app/models/participant_profile.rb
```diff
@@ -3,6 +3,10 @@
 class ParticipantProfile < ApplicationRecord
   has_paper_trail
   belongs_to :teacher_profile, touch: true
+
+  # TODO: Back-fill, so that every profile has a schedule
+  belongs_to :schedule, optional: true, touch: true
```
ethanmills, line 8: why the touch?

app/models/milestone.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class Milestone < ApplicationRecord
```
ethanmills, line 3: Do we want to start namespacing things? `Payment::Milestone` for example?

ethanmills (approved):

app/models/participant_profile.rb
kicferk1: I reckon if the schedule the participant is on changes, we would like to tell the lead providers about it in participant api. Especially if we end up letting IC's or participants change the schedule. 

Although... I guess maybe we want to only add that one after we populate the profiles with reasonable defaults to avoid 40k updated records 😂

app/models/milestone.rb
kicferk1: Hmm, maybe. Let's go with `Finance` namespace, I think it is used in a few places now.

jamesrose (approved):

PR #799, opened 2021-08-05
add OWasp Action to test staging on manual trigger

mattonfoot:
### Context
An OWasp security scan task that can be triggered manually against the staging environment

Add the zapscan container to a separate github action which can be triggered manually and add a basic zap config that can be used to ignore specific security tests if needed.

### Guidance to review
Review output of OWasp security testing action 

https://github.com/DFE-Digital/early-careers-framework/actions/workflows/owasp.yml


.github/workflows/owasp.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.4
+
+      - name: ZAP Scan
+        uses: zaproxy/action-full-scan@v0.2.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          docker_name: 'owasp/zap2docker-stable'
+          target: 'https://s-manage-training-for-early-career-teachers.education.gov.uk/sandbox'
```
kicferk1, line 21: Soo, what will it test exactly? Just the things discoverable from that page?

mattonfoot: It is a spider, and that page has links to most places it needs to get to, it may need parallel runs but it is a start. unfortunately I can't run it manually until it is on main branch! so will have to make it on pull for this branch first

mattonfoot: there is also an API scanner I will look to run in parallel that runs off the openapi spec

https://www.zaproxy.org/docs/docker/api-scan/

jamesrose (approved):

PR #801, opened 2021-08-05
CPDRP-806: BUGFIX: do not withdraw participants when changing between CIP and FIP

ethanmills:
A series of compounded errors lead to this one getting though 😞

### DIFF
--- a/app/services/admin/change_induction_service.rb
+++ b/app/services/admin/change_induction_service.rb
@@ -48,7 +48,7 @@ module Admin
     end
 
     def withdraw_participants_if_required(new_provision)
-      return if %w[core_induction_programme full_induction_programme].include? new_provision
+      return if %i[core_induction_programme full_induction_programme].include? new_provision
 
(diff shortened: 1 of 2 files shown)


app/services/admin/change_induction_service.rb
```diff
@@ -48,7 +48,7 @@ def perform_induction_change(new_provision)
     end
 
     def withdraw_participants_if_required(new_provision)
-      return if %w[core_induction_programme full_induction_programme].include? new_provision
+      return if %i[core_induction_programme full_induction_programme].include? new_provision
```
kicferk1, line 51: I sometimes hate Ruby's distinction between strings and symbols, and this is certainly one of those times 😭

kicferk1 (approved):

app/services/admin/change_induction_service.rb
ethanmills: If we had some kind of type system...

PR #805, opened 2021-08-06
Fix validation for event type

kicferk1:
### Context
We missed that when we switched from json schema validator into rails validations.

Add validations for evidence type. Add tests to show how they work.

### Guidance to review
I also changed the schemas. One was the npq retained event, because we need it to match our actual validation.

The other was for ecf started event. I think we don't want or need them on that event.

### Testing
Unit tests for the validations.

### Review Checks

Try submitting a retained event with invalid evidence held, it should fail. Before it succeeded.

### DIFF
--- a/app/services/record_declarations/retained/mentor.rb
+++ b/app/services/record_declarations/retained/mentor.rb
@@ -5,6 +5,10 @@ module RecordDeclarations
     class Mentor < ::RecordDeclarations::Base
       include RecordDeclarations::Mentor
       include Retained
+
+      def valid_evidence_types
+        %w[training-event-attended]
+      end
     end
   end
 end
--- a/app/services/record_declarations/retained/mentor.rb
+++ b/app/services/record_declarations/retained/mentor.rb
@@ -7,7 +7,7 @@ module RecordDeclarations
       include Retained
 
       def valid_evidence_types
-        %w[training-event-attended]
+        %w[training-event-attended self-study-material-completed other]
       end
     end
   end
--- a/app/services/record_declarations/retained/npq.rb
+++ b/app/services/record_declarations/retained/npq.rb
@@ -5,6 +5,10 @@ module RecordDeclarations
     class NPQ < ::RecordDeclarations::Base
       include RecordDeclarations::NPQ
       include Retained
+
+      def valid_evidence_types
+        %w[training-event-attended]
+      end
     end
   end
 end
--- a/app/services/record_declarations/retained/npq.rb
+++ b/app/services/record_declarations/retained/npq.rb
@@ -7,7 +7,7 @@ module RecordDeclarations
       include Retained
 
       def valid_evidence_types
-        %w[training-event-attended]
+        %w[training-event-attended other]
       end
     end
   end
--- a/app/services/record_declarations/retained/npq.rb
+++ b/app/services/record_declarations/retained/npq.rb
@@ -7,7 +7,7 @@ module RecordDeclarations
       include Retained
 
       def valid_evidence_types
-        %w[training-event-attended other]
+        %w[yes]
       end
     end
   end
--- a/etc/schema/0.3/npq/participant_declarations/retained/create/request_schema.json
+++ b/etc/schema/0.3/npq/participant_declarations/retained/create/request_schema.json
@@ -36,8 +36,7 @@
     },
     "evidence_type": {
       "enum": [
-        "training-event-attended",
-        "self-study-material-completed"
+        "training-event-attended"
       ]
     },
     "npq": {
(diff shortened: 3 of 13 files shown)


app/services/record_declarations/retained/mentor.rb
```diff
@@ -5,6 +5,10 @@ module Retained
     class Mentor < ::RecordDeclarations::Base
       include RecordDeclarations::Mentor
       include Retained
+
+      def valid_evidence_types
+        %w[training-event-attended]
```
blackrat, line 10: This isn't right. Should be the same as ECT.

blackrat (changes requested):

app/services/record_declarations/retained/npq.rb
```diff
@@ -5,6 +5,10 @@ module Retained
     class NPQ < ::RecordDeclarations::Base
       include RecordDeclarations::NPQ
       include Retained
+
+      def valid_evidence_types
+        %w[training-event-attended]
```
blackrat, line 10: We don't know what these are for NPQ, but they've not been pinned down yet.

etc/schema/0.3/npq/participant_declarations/retained/create/request_schema.json
```diff
@@ -36,8 +36,7 @@
     },
     "evidence_type": {
       "enum": [
-        "training-event-attended",
-        "self-study-material-completed"
+        "training-event-attended"
```
blackrat, line 39: Don't know why this has changed.

app/services/record_declarations/retained/mentor.rb
kicferk1: Why on earth would mentors attend self-study sessions?

blackrat (changes requested):
ECF started we don't need.

From previous communications:
- NPQ is still being decided, so it was a call from the team to use the same as ECF as placeholders. 
- ECTs and Mentors are the same. If you've got more information on that changing, can you link it?

app/services/record_declarations/retained/mentor.rb
kicferk1: I will double check with people who ought to know, but from my understanding, the course that mentors and ects are doing is induction into teaching. It makes sense for ECTs to do self-study sessions - after all, there is a lot they have to learn about teaching, from a lot of course materials.

But mentors are experienced teachers. I don't think they get time to do self study on the induction materials, and from the little UR I remember them doing on E&L, I don't think they really want to either. So it seemed weird to me that they would have self study sessions under FIP.

kicferk1: And apparently I was wrong, mentors can do self study. The more you know.

etc/schema/0.3/npq/participant_declarations/retained/create/request_schema.json
kicferk1: Because I want to keep it consistent with the actual validation. This is for NPQ, which we don't really know much about, but I thought you told me they would not need the self study materials?

kicferk1:
Alright, based on https://ukgovernmentdfe.slack.com/archives/C01BE7G56F7/p1628240977059800 we do indeed want to allow self study for mentors. I guess my understanding of FIP is not perfect.

From the same conversation it also seems we want to allow "other" type there.

etc/schema/0.3/npq/participant_declarations/retained/create/request_schema.json
blackrat: No. Only thing that I knew was that there is probably going to be a third one for ECF ("other"), but that's not cast in stone yet as far as I know.

app/services/record_declarations/retained/npq.rb
```diff
@@ -5,6 +5,10 @@ module Retained
     class NPQ < ::RecordDeclarations::Base
       include RecordDeclarations::NPQ
       include Retained
+
+      def valid_evidence_types
+        %w[training-event-attended other]
```
blackrat, line 10: "yes" as a placeholder for NPQs. (same slack thread)

etc/schema/0.3/npq/participant_declarations/retained/create/request_schema.json
blackrat: likewise here

spec/services/record_declarations/retained/early_career_teacher_spec.rb
```diff
@@ -16,7 +16,7 @@
       declaration_type: "retained-1",
       course_identifier: "ecf-induction",
       lead_provider_from_token: another_lead_provider,
-      evidence_held: "Test evidence",
+      evidence_held: %w[training-event-attended self-study-material-completed other].sample,
```
blackrat, line 19: Do we really want to do more "sample" based testing? Wouldn't it be better to test them all rather than risk a random failure that would be harder to diagnose?

blackrat (approved):

PR #806, opened 2021-08-06
CPDTP-371 Create header to more tightly control date validation for the non-production systems

plisovin:
### Context

There is a requirement for lead providers to be able to test declaration date validation in sandbox environment. They should be able to pass a date in the ```header X-With-Server-Date``` and the server will use it as current date. Future ticket will cover validations.

Middleware to use timecop to set date for the scope of a request

### Testing

Set an additional header and check the server side date will be set to that date.

Screenshot 2021-08-06 at 10 27 06

### Review Checks

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -101,6 +101,8 @@ gem "rouge"
 
 gem "ransack"
 
+gem "timecop"
+
 # Payment breackdown
 
 gem "terminal-table"
--- a/Gemfile
+++ b/Gemfile
@@ -101,8 +101,6 @@ gem "rouge"
 
 gem "ransack"
 
-gem "timecop"
-
 # Payment breackdown
 
 gem "terminal-table"
@@ -148,6 +146,7 @@ end
 
 group :development, :deployed_development, :test, :sandbox do
   gem "faker"
+  gem "timecop"
 end
 
 group :development do
--- a/config/application.rb
+++ b/config/application.rb
@@ -40,6 +40,10 @@ module EarlyCareerFramework
       require_dependency(c)
     end
 
+    Dir.glob(Rails.root + "app/middlewares/*.rb").sort.each do |file|
+      require file
+    end
+
     Dir.glob(Rails.root + "app/serializers/**/*_serializer*.rb").each do |c|
       require_dependency(c)
     end
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -5,6 +5,9 @@ require "active_support/core_ext/integer/time"
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  # Used to handle HTTP_X_WITH_SERVER_DATE header for server side datetime overwrite
+  config.middleware.use TimeTraveler
+
   # In the development environment your application's code is reloaded any time
   # it changes. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
--- a/config/environments/sandbox.rb
+++ b/config/environments/sandbox.rb
@@ -5,6 +5,8 @@ require "active_support/core_ext/integer/time"
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  config.middleware.use TimeTraveler
+
   # Code is not reloaded between requests.
   config.cache_classes = true
 
--- a/config/environments/sandbox.rb
+++ b/config/environments/sandbox.rb
@@ -5,6 +5,7 @@ require "active_support/core_ext/integer/time"
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  # Used to handle HTTP_X_WITH_SERVER_DATE header for server side datetime overwrite
   config.middleware.use TimeTraveler
 
   # Code is not reloaded between requests.
--- a/spec/middlewares/time_traveler_spec.rb
+++ b/spec/middlewares/time_traveler_spec.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+describe TimeTraveler do
+  context "when called with a POST request" do
+    let(:app) { proc { [200, {}, %w[Live]] } }
+    let(:middleware) { described_class.new(app) }
+    let(:request) { Rack::MockRequest.new(middleware) }
+    let(:header_date) { "2021-01-02T03:04:05.000Z" }
+    let(:expected_date) { Time.zone.parse(header_date) }
+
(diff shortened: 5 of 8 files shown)


config/environments/sandbox.rb
```diff
@@ -5,6 +5,8 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  config.middleware.use TimeTraveler
+
```
kicferk1, line 9: Let's add it to deployed development too. That's what is on the dev environment and in review apps.

Also, let's duplicate the comment from `development` config, it makes sense here too :)

spec/middlewares/time_traveler_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    let(:request) { Rack::MockRequest.new(middleware) }
+    let(:header_date) { "2021-01-02T03:04:05.000Z" }
+    let(:expected_date) { Time.zone.parse(header_date) }
+
+    context "when server date header passed" do
+      it "passes the request through unchanged" do
+        request.get("/", "X_WITH_SERVER_DATE" => header_date) do
+          expect(Time.zone.now).to eq(expected_date)
+        end
+      end
```
kicferk1, line 16: Can we have another test here to show that date for other requests is unaffected?
```
      it "changes the current time on the server for a single request" do
        request.get("/", "X_WITH_SERVER_DATE" => header_date) do
          expect(Time.zone.now).to eq(expected_date)
        end

        request.get("/") do
          expect(Time.zone.now).to eq(some_other_date)
        end
      end
```
`some_other_date` being some set time that we set with `travel_to`?

config/application.rb
```diff
@@ -40,6 +40,10 @@ class Application < Rails::Application
       require_dependency(c)
     end
 
+    Dir.glob(Rails.root + "app/middlewares/*.rb").sort.each do |file|
+      require file
+    end
```
kicferk1, line 45: This doesn't actually use the middleware, just loads it so that other files don't need to `require` it explicitly, right?

spec/middlewares/time_traveler_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+describe TimeTraveler do
+  context "when called with a POST request" do
```
kicferk1, line 4: We are not making post requests in those tests. I would axe the most outside context, it's not like there is any other context in this test file.

Gemfile
```diff
@@ -101,6 +101,8 @@ gem "rouge"
 
 gem "ransack"
 
+gem "timecop"
```
ethanmills, line 104: Is there a way to prevent this from being loaded in production?

ethanmills (changes requested):
Is there anything to stop someone from doing this on production?

kicferk1 (commented):
Looks good overall, the test could use some better wording.

Some concerns, which I hope you checked are:

Can one request change the time for other requests? I hope not.

If middleware is not used in config files, the header does nothing, right? We really don't want to let this affect production.

Gemfile
ethanmills: Aha, I've just spotted the initializers

config/application.rb
ethanmills: Is this needed with Zeitwerk? I was under the impression everything under /app was autoloaded?

ethanmills (approved):
Undoing request for changes

Gemfile
plisovin: Yeah, I will move it to sandbox, develop etc group

config/application.rb
plisovin: Wasn't autoloaded unless I was missing something locally

Gemfile
blackrat: Should be all environments except production.

config/environments/development.rb
```diff
@@ -5,6 +5,9 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  # Used to handle HTTP_X_WITH_SERVER_DATE header for server side datetime overwrite
+  config.middleware.use TimeTraveler
```
blackrat, line 9: Should also be in `deployed_development` and non-production environments.

spec/middlewares/time_traveler_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    let(:middleware) { described_class.new(app) }
+    let(:request) { Rack::MockRequest.new(middleware) }
+    let(:header_date) { "2021-01-02T03:04:05.000Z" }
+    let(:expected_date) { Time.zone.parse(header_date) }
+
+    context "when server date header passed" do
+      it "passes the request through unchanged" do
+        request.get("/", "X_WITH_SERVER_DATE" => header_date) do
+          expect(Time.zone.now).to eq(expected_date)
+        end
```
blackrat, line 15: Does this work if it's not yet enabled in the test environment?

Gemfile
ethanmills: Even staging?

spec/middlewares/time_traveler_spec.rb
plisovin: Apparently request block didn't work at all. Changed it to return time in the response and refactored contexts [here](66d81bb)

spec/middlewares/time_traveler_spec.rb
plisovin: Fixed in [66d81bb](66d81bb)

spec/middlewares/time_traveler_spec.rb
plisovin: Added test environment in 66d81bb

config/environments/sandbox.rb
plisovin: Done

plisovin: Added additional check for non-production environment https://github.com/DFE-Digital/early-careers-framework/blob/CPDTP-371-create-header-to-control-dates/app/middlewares/time_traveler.rb#L9

config/environments/development.rb
plisovin: Added in the latest commit

Gemfile
blackrat: Are there any other exceptions on staging where it's not an 100% match to production? If not, then no.

blackrat (approved):

kicferk1 (approved):

PR #807, opened 2021-08-06
Additionally set test API keys on deployed_development

ethanmills:
Is there any reason we shouldn't do this? Would make testing things much easier


kicferk1 (approved):
I don't think there is any harm with adding it to dev and review apps.

kicferk1:
I guess staging is a question, but it is not like there is any real participant data there.

jamesrose (approved):

ethanmills:
> I guess staging is a question, but it is not like there is any real participant data there.

This shouldn't do staging

PR #809, opened 2021-08-09
CPDTP-252 Add declaration format and future date validations

plisovin:

Added 2 param validators for record declaration's date: 
 - RCF3339 date format 
 - future declaration date.

### Guidance to review

Screenshot 2021-08-09 at 15 23 48

Screenshot 2021-08-09 at 15 24 19

### Review Checks


spec/services/record_declarations/param_validation_spec.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+describe "Record Declaration Validation" do
```
kicferk1, line 3: Should these tests live in `spec/services/record_declarations/retained/early_career_teacher_spec.rb`?

kicferk1 (commented):
Looks pretty good, just a question about where we want the tests to live :)

spec/services/record_declarations/param_validation_spec.rb
plisovin: well, yeah, I've grouped them here as well as all schema validation tests. Will move them into event specific folders

plisovin: Moved them in 81469b5

app/services/record_declarations/base.rb
```diff
@@ -10,9 +10,12 @@ class Base
     validates :declaration_type, inclusion: { in: :valid_declaration_types, message: I18n.t(:invalid_declaration_type) }
     validates :course_identifier, presence: { message: I18n.t(:missing_course_identifier) }
     validates :declaration_date, presence: { message: I18n.t(:missing_declaration_date) }
+    validates :declaration_date, declaration_date: true
+    validates :declaration_date, future_date: true
```
kicferk1, line 14: Do we want to `allow_blank: true` for both of them? So that the we only catch the blank case once?

spec/services/record_declarations/retained/early_career_teacher_spec.rb
```diff
@@ -71,4 +71,13 @@ def generate_raw_event(params)
       expect { described_class.call(params.merge(declaration_type: "invalid")) }.to raise_error(ActionController::ParameterMissing)
     end
   end
+
+  context "when declaration date is invalid" do
+    it "raises a ParameterMissing error" do
+      params = ect_params.merge({ declaration_date: "2021-06-21 08:46:29" })
+      params[:raw_event] = generate_raw_event(params)
+      expected_msg = "param is missing or the value is empty: [\"The property '#/declaration_date' must be a valid RCF3339 date\"]"
```
kicferk1, line 79: I think `ApiController` is changing the actual response to be more like 
```
"{\"bad_or_missing_parameters\":\"The property '#/declaration_date' must be a valid RCF3339 date\"}"
```

Writing down the message might confuse someone into thinking that this is what our api outputs look like.

plisovin: good point, we can change to match a substring

plisovin: Like [here](https://github.com/DFE-Digital/early-careers-framework/blob/2caff755dec2ae3b8ca851e063373659e53d9f88/spec/services/record_declarations/started/early_career_teacher_spec.rb#L84)

kicferk1 (approved):

kicferk1:
There is a small conflict, looks like we both added a test for schema validation to the end of the file. I think we want to keep both tests :) 

Other than that, looks good to me!

app/validators/declaration_date_validator.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class DeclarationDateValidator < ActiveModel::EachValidator
+  RCF3339_DATE_REGEX = /\A\d{4}-\d{2}-\d{2}T(\d{2}):(\d{2}):(\d{2})([\.,]\d+)?(Z|[+-](\d{2})(:?\d{2})?)?\z/i.freeze
```
blackrat, line 4: can't we just use:
`validates_format_of: :declaration_date, with: /\A\d{4}-\d{2}-\d{2}T(\d{2}):(\d{2}):(\d{2})([\.,]\d+)?(Z|[+-](\d{2})(:?\d{2})?)?\z/i`

blackrat (changes requested):

blackrat (changes requested):
Do we need a separate validator.

app/validators/declaration_date_validator.rb
plisovin: it was there originally, I just moved it to a separate one as we may reuse it for other classes and it looks quite long in the ```Base```

blackrat (approved):

PR #810, opened 2021-08-09
CPDRP-812: validation test scenarios

ethanmills:
### Context
CPDRP-812
Test data for manual testing of participant validation. Documented in https://docs.google.com/document/d/1PlBZQHqSmD3QxYzTpAvd98-rbLzkww-jdync3fYUVv4/edit#

Also changed:
- Use preprod DQT api in pre-live environments (not the broken "dev" environment)
- Make the seeds idempotent


db/seeds/test_data.rb
```diff
@@ -208,3 +208,102 @@
...
+end
+ParticipantProfile::NPQ.find_or_create_by!(teacher_profile: teacher_profile)
+
+user = User.find_or_create_by!(email: "fip-mentor2@example.com") do |u|
+  u.full_name = "FIP Mentor"
+end
+teacher_profile = TeacherProfile.find_or_create_by!(user: user)
+ParticipantProfile::Mentor.find_or_create_by!(teacher_profile: teacher_profile) do |mentor_profile|
+  mentor_profile.school_cohort = School.find_by(urn: "000105").school_cohorts.find_by(cohort: Cohort.current)
+end
```
kicferk1, line 257: Soo, are they linked in any way, other than that we know they are the same?

Or is the point of linking when the mentor user enters their trn which is the same into our dqt validation?

kicferk1 (approved):
Looks pretty good to me, did we write down the emails for testing somewhere prominent so that we don't need to look through the code to know how to test an edge case?

db/seeds/test_data.rb
ethanmills: They're not linked in any particular way, that's just the email I told people to use in the instructions

ethanmills:
> Looks pretty good to me, did we write down the emails for testing somewhere prominent so that we don't need to look through the code to know how to test an edge case?

Google doc in description 🙂

PR #811, opened 2021-08-09
Add accept/reject guidance to NPQ usage guide

jamesrose:
### Context

- Add accept/reject guidance to NPQ usage guide

### DIFF
--- a/spec/requests/lead_providers/npq_usage_spec.rb
+++ b/spec/requests/lead_providers/npq_usage_spec.rb
@@ -4,22 +4,28 @@ require "rails_helper"
 
 RSpec.describe "NPQ usage guide", type: :request do
   describe "GET /lead-providers/guidance/npq-usage" do
-    it "should show the NPQ usage guide" do
+    before do
       get lead_providers_guidance_npq_usage_path
+    end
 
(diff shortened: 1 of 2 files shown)


spec/requests/lead_providers/npq_usage_spec.rb
```diff
@@ -21,5 +21,13 @@
 
       expect(response.body).to include("Declaring that an NPQ participant has started their induction")
     end
+
+    it "should explain how to accept an NPQ application" do
+      expect(response.body).to include("Accept an NPQ application")
```
ethanmills, line 26: Missing a get here, maybe move them all to a before block?

asmega:
It's fine but I think we want to add to it. I feel we should have more guidance around the lifecycle. e.g. a user has applied but now wants to study a different course. we should outline the steps as i don’t think a provider would know what to do

spec/requests/lead_providers/npq_usage_spec.rb
jamesrose: Cheers. I blindly copied it!

asmega (approved):

PR #812, opened 2021-08-09
Fixing data synchronisation on NPQValidationData updates

BroiSatse:
It seems that NPQValidationData did not synchronize very well with participant and teacher profiles. This PR should fix TRN synchronisation issue (as highlighted by tests).

Long term goal should be to remove this syncing...


asmega (approved):

kicferk1:
The review app fails to deploy, saying that the name is already taken. My guess is that it is some weird interaction with force pushing in an unfortunate moment. You can either open another PR with this change, or close and re-open this one to get the app working.

jamesrose:
@BroiSatse Does this need any remedial fixes on production?

BroiSatse:
Review app deployment fails due to the race condition between two parallel deployments caused by badly timed push force. Closing to reopen

PR #813, opened 2021-08-09
CPDRP-659: Extend participant API with additional information

ethanmills:
### Context
CPDRP-659
Add the following details to participants API
- TRN
- eligible for funding
- uplift

I will follow this up with a PR to add test data to the sandbox

Still to do:
- [x] Document eligibility and uplift

### Review Checks

### DIFF
--- a/app/serializers/participant_serializer.rb
+++ b/app/serializers/participant_serializer.rb
@@ -14,7 +14,7 @@ class ParticipantSerializer
     end
 
     def participant_active?(user)
-      user.early_career_teacher_profile&.active? || user.mentor_profile&.active?
+      user.teacher_profile.ecf_profile&.active?
     end
   end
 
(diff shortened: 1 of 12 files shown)


app/serializers/participant_serializer.rb
```diff
@@ -24,28 +24,35 @@ def participant_active?(user)
...
-    user.early_career_teacher_profile&.status || user.mentor_profile&.status || "withdrawn"
+    user.teacher_profile.ecf_profile&.status || "withdrawn"
+  end
+
+  active_participant_attribute :teacher_reference_number do |user|
+    user.teacher_profile.trn
+  end
+
+  active_participant_attribute :teacher_reference_number_validated do |user|
+    user.teacher_profile.trn.present?
```
kicferk1, line 56: That field looks... excessive. Are we planning to surface some more complex things here? Cause if not, we can just rename `teacher_reference_number` to `validated_ teacher_reference_number` and convey the same meaning.

jamesrose: We might want to surface TRNs that are not yet validated @kicferk1. This mirrors NPQ applications, so we'd want to change it in both places.

spec/serializers/participant_serializer_spec.rb
```diff
@@ -24,14 +25,67 @@
...
       end
 
       it "outputs correctly formatted serialized ECTs" do
-        expected_json_string = "{\"data\":{\"id\":\"#{ect.id}\",\"type\":\"participant\",\"attributes\":{\"email\":null,\"full_name\":null,\"mentor_id\":null,\"school_urn\":null,\"participant_type\":null,\"cohort\":null,\"status\":\"withdrawn\"}}}"
+        expected_json_string = "{\"data\":{\"id\":\"#{ect.id}\",\"type\":\"participant\",\"attributes\":{\"email\":null,\"full_name\":null,\"mentor_id\":null,\"school_urn\":null,\"participant_type\":null,\"cohort\":null,\"status\":\"withdrawn\",\"teacher_reference_number\":null,\"teacher_reference_number_validated\":null,\"eligible_for_funding\":null,\"pupil_premium_uplift\":null,\"sparsity_uplift\":null}}}"
         expect(ParticipantSerializer.new(ect).serializable_hash.to_json).to eq expected_json_string
       end
     end
+
+    describe "funding_eligibility" do
```
kicferk1, line 38: `eligible_for_funding`?

kicferk1: So that it matches field name?

lib/tasks/sandbox_data.rake
```diff
@@ -70,3 +70,14 @@ def create_school_and_associations(lead_provider, cohort, index)
...
   school
 end
+
+def random_trn
+  return if [true, false].sample
+
+  sprintf("%07i", Random.random_number(9_999_999))
+end
+
+def create_teacher
```
blackrat, line 80: Think it's about time we moved these into an included module. These are global scope for rake, so if any other rake task creates the same method it could cause spurious errors since they could overwrite each other.

lib/tasks/sandbox_data.rake
```diff
@@ -70,3 +70,14 @@ def create_school_and_associations(lead_provider, cohort, index)
...
+  mentor = User.create!(full_name: Faker::Name.name, email: Faker::Internet.email)
+  TeacherProfile.create!(user: mentor, trn: random_trn)
```
blackrat, line 82: Does this create a teacher profile with a `trn: <random number>` or a `trn: true`?

blackrat (changes requested):

lib/tasks/sandbox_data.rake
ethanmills: Yes

blackrat (approved):

spec/serializers/participant_serializer_spec.rb
blackrat: I'll change that in the next PR

PR #816, opened 2021-08-09
CPDRP-799 feature flags and emails for beta

tonyheadford:
### Context
[Jira](https://dfedigital.atlassian.net/browse/CPDRP-799)
Enable feature flags and email participants for private beta of participant validation

Mailers for ECTs and Mentors (FIP)
Add method to `InviteSchools` to activate `:participant_validation` feature flag for each school in a list of URNs provided by the caller and send the appropriate email to the school's recipients.
Add UTM campaigns for `participant-validation-beta` and `participant-validation-research`

### Review Checks

### DIFF
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -230,6 +230,34 @@ class InviteSchools
     end
   end
 
+  def feature_flag_and_send_participant_validation_beta_emails(array_of_urns:)
+    start_url = participant_validation_start_url
+    user_research_url = participant_validation_research_url
+
+    School.where(urn: array_of_urns).find_each do |school|
+      FeatureFlag.activate(:participant_validation, for: school)
+
+      school.active_ecf_participant_profiles.each do |profile|
+        if profile.ect?
+          SchoolMailer.participant_validation_ect_email(
+            recipient: profile.user.email,
+            school_name: school.name,
+            start_url: start_url,
+            user_research_url: user_research_url
+          ).deliver_later
+        elsif profile.mentor?
+          SchoolMailer.participant_validation_fip_mentor_email(
+            recipient: profile.user.email,
+            school_name: school.name,
+            start_url: start_url
+          ).deliver_later
+        end
+      rescue StandardError
+        logger.info "Error sending participant validation email to: #{profile&.user&.email} ... skipping"
+      end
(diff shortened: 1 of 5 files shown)


app/services/invite_schools.rb
```diff
@@ -230,6 +230,34 @@ def send_year2020_invite_email
...
+            start_url: start_url,
+            user_research_url: user_research_url,
+          ).deliver_later
+        elsif profile.mentor?
+          SchoolMailer.participant_validation_fip_mentor_email(
+            recipient: profile.user.email,
+            school_name: school.name,
+            start_url: start_url,
+          ).deliver_later
+        end
```
kicferk1, line 254: Should we check that the school is actually doing FIP before we send this emails? CIP schools have mentors too, couple hundreds of them.

kicferk1: Or should we perhaps make the email work with both cip and fip mentors? Probably a product question which one we want (or maybe we want something else).

kicferk1 (commented):
I'm slightly worry about naming - if the mentor emails are just for FIP mentors, then we should make sure we don't send them to CIP mentors (or resolve this in some other way).

ethanmills:
Looks like the template isn't set up quite correctly, but the mentor email will also have a UR link

app/services/invite_schools.rb
tonyheadford: Not sure how much of an issue it is at this point as this is to email a pre-selected set of FIP schools for the beta, but generally we should do this.

tonyheadford:
> Looks like the template isn't set up quite correctly, but the mentor email will also have a UR link. I've messaged @Lloyd-K

Thanks, yes I did wonder, I was going to ask her this morning

ethanmills:
> > Looks like the template isn't set up quite correctly, but the mentor email will also have a UR link. I've messaged @Lloyd-K
> 
> Thanks, yes I did wonder, I was going to ask her this morning

@tonyheadford Looks like it's been updated

app/services/invite_schools.rb
tonyheadford: Added a check for this just to be on the safe side.

kicferk1 (approved):

PR #825, opened 2021-08-11
T&P: Split audit from logic in participant declarations

kicferk1:
### Context

We want to know what LPs are doing with our api, especially if they don't really use it right. 

https://dfedigital.atlassian.net/browse/CPDTP-383 says that "all declarations that pass schema validations should be stored".

We were storing raw event details on the declarations. I don't think that's the right place to store them. Truth be told, I am not sure we want to store them in our app at all (maybe logit is more appropriate), but to preserve a way of storing the raw requests we are sent I added a middleware that audits our api.

1. Move `raw_event` into its own model, store it via the api before any validation or authentication.
2. Add `ParticipantDeclarationAttempt` model - which should store things that pass schema validation but not business logic validation.

### Testing
Added unit tests for the middleware.
Added request spec to show the case when attempt is recorded but declaration is not. 

Try making a declaration against a participant that is not in your lead provider scope.
Then ssh into the app - the attempt should be saved, but the declaration should not.

### DIFF
--- a/app/controllers/api/v1/participant_declarations_controller.rb
+++ b/app/controllers/api/v1/participant_declarations_controller.rb
@@ -6,7 +6,7 @@ module Api
       include ApiTokenAuthenticatable
 
       def create
-        params = HashWithIndifferentAccess.new({ raw_event: request.raw_post, lead_provider_from_token: cpd_lead_provider }).merge(permitted_params["attributes"] || {})
+        params = HashWithIndifferentAccess.new({ lead_provider_from_token: cpd_lead_provider }).merge(permitted_params["attributes"] || {})
         validate_params!(params)
         render json: RecordParticipantDeclaration.call(convert_params_for_declaration(params))
       end
--- a/app/middlewares/lead_provider_request_auditor.rb
+++ b/app/middlewares/lead_provider_request_auditor.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class LeadProviderRequestAuditor
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    store_request_details(env)
+    @app.call(env)
+  end
+
+private
+
+  def store_request_details(env)
+    path = env["PATH_INFO"]
+    if audited_paths.any? { |pattern| pattern.match?(path) }
+      request = Rack::Request.new(env)
+      body = request.body.read.squish
+      ApiRequestAudit.create!(path: path, body: body)
+    end
+  rescue StandardError => e
+    Sentry.capture_exception(e)
+  end
+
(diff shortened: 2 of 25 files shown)


app/middlewares/lead_provider_request_auditor.rb
```diff
@@ -0,0 +1,29 @@
...
+    @app.call(env)
+  end
+
+private
+
+  def store_request_details(env)
+    path = env["PATH_INFO"]
+    if audited_paths.any? { |pattern| pattern.match?(path) }
+      request = Rack::Request.new(env)
+      body = request.body.read.squish
```
ethanmills, line 19: Do we need to store a hashed token or anything here? How would we use this?

jamesrose:
Echo'ing @ethanmills here. LogIt shouldn't be used as a data store.

app/services/record_declarations/base.rb
```diff
@@ -54,8 +54,10 @@ def call
         raise ActionController::ParameterMissing, errors.map(&:message)
       end
 
-      declaration = create_record!
+      create_declaration_attempt!
       validate_provider!
+      declaration = create_record!
+
```
jamesrose, line 60:
```suggestion
      declaration_attempt = create_declaration_attempt!
      validate_provider!
      declaration = create_record!
      declaration_attempt.update!(declaration: declaration)
```

Have you thought about whether we'd like to reference the declaration attempt back to the declaration itself? Would that be useful?

jamesrose (approved):

app/middlewares/lead_provider_request_auditor.rb
kicferk1: I think we would dump the contents of the audited endpoints into some lovely google studio to see something. 

I think until I get some concrete use cases (which so far I don't think we have), I prefer to keep it as simple as possible.

app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -6,7 +6,7 @@ class ParticipantDeclarationsController < Api::ApiController
       include ApiTokenAuthenticatable
 
       def create
-        params = HashWithIndifferentAccess.new({ raw_event: request.raw_post, lead_provider_from_token: cpd_lead_provider }).merge(permitted_params["attributes"] || {})
+        params = HashWithIndifferentAccess.new({ lead_provider_from_token: cpd_lead_provider }).merge(permitted_params["attributes"] || {})
```
blackrat, line 9: We need this to protect against schema API changes for declarations. It keeps us immutable and protects against API changes.

blackrat (changes requested):

PR #827, opened 2021-08-11
Fix typo in participant validation journey

tonyheadford:
### Context
Typo was causing error to be raised and not storing validation data under certain circumstances

### Review Checks


kicferk1 (approved):

ethanmills (approved):
I've approved this because we need to get on with testing, but I think we should write a test that would have caught this

tonyheadford:
> I've approved this because we need to get on with testing, but I think we should write a test that would have caught this

Yes absolutely, I'm looking at that now but thought we should get this out asap

PR #828, opened 2021-08-11
Set fields on previous induction to nil until we have the data to calculate the correct value

ethanmills:
This should make it easier to determine which records have or haven't been processed when we get the data


jamesrose (approved):

ethanmills:
Trying to get review app....

ethanmills:
Change already made

PR #832, opened 2021-08-11
CPDTP-391 Extract common setup for record declarations test

plisovin:
### Context

Record declaration tests have common setup blocks in them. This is the first step to extract into shared context and use it in other areas for example in request type rspec test.

### DIFF
--- a/spec/services/record_declarations/retained/early_career_teacher_spec.rb
+++ b/spec/services/record_declarations/retained/early_career_teacher_spec.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "rails_helper"
+
 require_relative "../../../shared/context/service_record_declaration_params.rb"
 require_relative "../../../shared/context/lead_provider_profiles_and_courses.rb"
 
--- a/spec/services/record_declarations/retained/early_career_teacher_spec.rb
+++ b/spec/services/record_declarations/retained/early_career_teacher_spec.rb
@@ -1,41 +1,12 @@
 # frozen_string_literal: true
 
 require "rails_helper"
+require_relative "../../../shared/context/service_record_declaration_params.rb"
+require_relative "../../../shared/context/lead_provider_profiles_and_courses.rb"
 
 RSpec.describe RecordDeclarations::Retained::EarlyCareerTeacher do
-  let(:cpd_lead_provider) { create(:cpd_lead_provider, :with_lead_provider) }
-  let(:another_lead_provider) { create(:cpd_lead_provider, name: "Unknown") }
-  let(:ect_profile) { create(:early_career_teacher_profile) }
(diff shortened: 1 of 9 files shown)


spec/services/record_declarations/retained/early_career_teacher_spec.rb
```diff
@@ -1,46 +1,12 @@
 # frozen_string_literal: true
 
 require "rails_helper"
+require "./rspec/shared/context/service_record_declaration_params.rb"
+require "./rspec/shared/context/leadprovider_profiles_and_courses.rb"
```
kicferk1: Let's name it `lead_provider` rather than `leadprovider` just to keep consistent across codebase.

spec/shared/context/leadprovider_profiles_and_courses.rb
```diff
@@ -0,0 +1,30 @@
...
+    create(:partnership,
+           school: ect_profile.school,
+           lead_provider: cpd_lead_provider.lead_provider,
+           cohort: ect_profile.cohort,
+           delivery_partner: delivery_partner)
+  end
+
+  def generate_raw_event(params)
+    params.except(:raw_event, :cpd_lead_provider).to_json
+  end
```
kicferk1, line 29: That probably does not belong in this file - cause this file does not know about params

kicferk1: The other shared context likely wants it more :)

spec/shared/context/leadprovider_profiles_and_courses.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  let(:delivery_partner) { create(:delivery_partner) }
+  let!(:school_cohort) { create(:school_cohort, school: ect_profile.school, cohort: ect_profile.cohort) }
+  let!(:partnership) do
+    create(:partnership,
+           school: ect_profile.school,
+           lead_provider: cpd_lead_provider.lead_provider,
+           cohort: ect_profile.cohort,
+           delivery_partner: delivery_partner)
+  end
```
kicferk1, line 25: Can we group certain parts together? Say, one part for ECF setup, another part for NPQ setup?

kicferk1 (commented):
Looks pretty neat, should make it much easier to write new tests that want the same basic T&P setup.

spec/shared/context/leadprovider_profiles_and_courses.rb
plisovin: You mean logically or in separate files? ECF stuff comes first here, then NPQ, and then school cohort. I can add comment to make it clear

kicferk1 (approved):

blackrat (approved):
Looks like a good first step to me. Will definitely help with writing tests.

PR #839, opened 2021-08-11
CPDRP-508-DQT API returns a boolean not a string value

tonyheadford:
### Context
From participant validation testing - [Jira](https://dfedigital.atlassian.net/browse/CPDRP-508?focusedCommentId=32255)
Fix issue where test API actually returns a boolean not a string "No" value

Expect a boolean instead of a string value for the `:active_alert` flag value returned from DQT API.

### Review Checks


kicferk1 (approved):
Is that a recent change in the API behaviour?

Did we get told about it?

If it is, and if we didn't, should we talk to whoever is managing that api and ask that they let us know about changes like that? Alternatively, should we think about smoke testing the dqt api?

PR #840, opened 2021-08-11
Do not count two blank NINos as being a match

ethanmills:
At present the validation service counts a match a blank NINo from the participant and  a blank from the DQT. This is wrong.


kicferk1 (approved):
Can any ther record be blank? trn? name? date of birth?

Also, maybe it is time to extract `nino_matches` into a private method?

asmega (approved):

jamesrose (approved):

ethanmills:
Also, maybe it is time to extract nino_matches into a private method?

Will come back to this maybe

PR #841, opened 2021-08-11
Squish whitespace in user entered data

tonyheadford:
### Context
From participant validation testing - [Jira](https://dfedigital.atlassian.net/browse/CPDRP-508?focusedCommentId=32248)
Fix issue where user could enter data with spurious trailing/leading whitespace and fail validation.

Squish the participant details so that leading trailing and multiple embedded whitespace are removed.

### Review Checks

### DIFF
--- a/app/controllers/participants/validations_controller.rb
+++ b/app/controllers/participants/validations_controller.rb
@@ -122,10 +122,8 @@ module Participants
     end
 
     def validate_participant_details_and_redirect
-      result = ParticipantValidationService.validate(trn: @participant_validation_form.trn,
-                                                     full_name: @participant_validation_form.name,
-                                                     date_of_birth: @participant_validation_form.date_of_birth,
-                                                     nino: @participant_validation_form.national_insurance_number)
+      result = ParticipantValidationService.validate(participant_details)
+
       if result.nil?
         @participant_validation_form.increment_validation_attempts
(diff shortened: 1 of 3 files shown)


app/controllers/participants/validations_controller.rb
```diff
@@ -147,13 +145,23 @@ def validate_participant_details_and_redirect
...
+          case key
+          when :name
+            :full_name
+          when :national_insurance_number
+            :nino
+          else
+            key
+          end
+        end
+    end
```
jamesrose, line 161:
```suggestion
    def participant_details
      @participant_validation_form.attributes.slice(:trn, :name, :date_of_birth, :national_insurance_number)
    end
```

I think I'd do the middle ground, and then closer to what you had in the caller:

```ruby
result = ParticipantValidationService.validate(trn: participant_details[:trn],
                                                     full_name: participant_details[:name],
                                                     date_of_birth: participant_details[:date_of_birth],
                                                     nino: participant_details[:national_insurance_number])
```

What do you think?

jamesrose (approved):
Nice one. Thanks for picking that up.

app/controllers/participants/validations_controller.rb
jamesrose: Fair, your call, I'll leave it up to you.

PR #847, opened 2021-08-12
NPQ profiles API endpoint refactor

asmega:
### Context

- There are changes that need to starting happening to the `api/v1/npq-profiles` endpoint
- The endpoint implemented `JSONAPI::ActsAsResourceController` from a third party gem
- This gem although brought a lot to the table out of the box is extraordinarily constraining and difficult to work with when try to extend it
- This will now make the endpoint more easily changeable for the next set of changes that need to happen

-  `api/v1/npq-profiles` now uses (mostly) vanilla rails
- Removed the gem `jsonapi-resources` as it is no longer used

### Guidance to review

- There should be little to no behavioural change

### Testing

- Can test in review app
- Or you can fire up `npq-registration` locally and should be able to register and see new data created in ECF

### Review Checks

- Create or find an existing `NPQRegistrationApiToken`
- Create or find an existing user
- Call the endpoint
```
curl -X POST -H "Authorization: Bearer TOKEN" -H "Content-Type: application/vnd.api+json" -d '{"data":{"type":"npq_profiles","attributes":{"teacher_reference_number":"1234567","teacher_reference_number_verified":true,"date_of_birth":"1990-12-13","school_urn":"123456","headteacher_status":"no"},"relationships":{"user":{"data":{"type":"users","id":"USER_ID_HERE"}},"npq_lead_provider":{"data":{"type":"npq_lead_providers","id":"NPQ_LEAD_PROVIDER_ID"}},"npq_course":{"data":{"type":"npq_courses","id":"NPQ_COURSE_ID"}}}}}' https://ecf-review-pr-847.london.cloudapps.digital/api/v1/npq-profiles
```
- Should see new `NPQValidationData` record persisted


app/controllers/api/v1/npq_profiles_controller.rb
```diff
@@ -4,13 +4,54 @@ module Api
   module V1
     class NPQProfilesController < Api::ApiController
       include ApiTokenAuthenticatable
-      include JSONAPI::ActsAsResourceController
```
kicferk1, line 7: I am glad we are cutting it out

app/controllers/api/v1/npq_profiles_controller.rb
```diff
@@ -4,13 +4,54 @@ module Api
...
+        validation_data = NPQValidationData.new(validation_data_params)
+        validation_data.npq_course = NPQCourse.find(npq_course_param)
+        validation_data.npq_lead_provider = NPQLeadProvider.find(npq_lead_provider_param)
+        validation_data.user = User.find(user_param)
+
+        if validation_data.save
+          render status: :created,
+                 content_type: "application/vnd.api+json",
+                 json: NPQValidationDataSerializer.new(validation_data).serializable_hash
+        end
```
kicferk1, line 18: If it is not saved, do we want to surface validation errors?

config/routes.rb
```diff
@@ -62,7 +62,7 @@
         end
       end
 
-      jsonapi_resources :npq_profiles, only: [:create]
+      resources :npq_profiles, only: [:create], path: "npq-profiles"
```
kicferk1, line 65: what was the old path? Will it need the change on the client?

app/controllers/api/v1/npq_profiles_controller.rb
asmega: It's in the next commit. Sorry, I saw your message after I made that commit.

config/routes.rb
asmega: endpoint has not changed. `jsonapi_resources` under the hood converts to a hyphen for the path

kicferk1 (approved):

ethanmills (approved):

jamesrose (approved):
Nice one.

PR #849, opened 2021-08-12
Temporarily disable auditing middleware

kicferk1:
### Context
For some reason the middleware causes 400s. I need to dig deeper into it, for now we should disable it.


jamesrose (approved):

jamesrose:
@ethanmills Yes, I think so.

PR #853, opened 2021-08-12
Fix and re-enable middleware

kicferk1:
### Context
Middleware was found out to be broken, and disabled earlier today.

1. Enable middlewares in test environment.
2. Switch to use a controller concern
3. Add tests checking for the content of audit object
4. Enable middlewares in other environments.

### Guidance to review

Just to be triple sure, we should check that it works with review app. Good thing `xxx-token` tokens work in review apps.


config/environments/deployed_development.rb
```diff
@@ -6,11 +6,9 @@
   # Settings specified here will take precedence over those in config/application.rb.
 
   # Used to handle HTTP_X_WITH_SERVER_DATE header for server side datetime overwrite
-  # TODO: Enable after more thorough testing
-  # config.middleware.use TimeTraveler
+  config.middleware.use TimeTraveler
```
ethanmills, line 9: We sure this is working too?

kicferk1: I will test that on review app too, although now that it it enabled in tests I am a lot more confident it will not break anything.

ethanmills (approved):

spec/middlewares/lead_provider_request_auditor_spec.rb
```diff
@@ -1,19 +1,15 @@
 # frozen_string_literal: true
 
-describe LeadProviderRequestAuditor do
-  let(:app) { proc { [200, {}, Time.zone.now.to_s] } }
-  let(:middleware) { described_class.new(app) }
-  let(:request) { Rack::MockRequest.new(middleware) }
-
+describe LeadProviderRequestAuditor, type: :request do
```
blackrat, line 3: Was the problem the middleware mock not being sufficient to test with? Shouldn't we have tests that replicate what raised and what happens in the event of a raise? This doesn't tell me what caused the raise; where the raise was handled; so hard to judge whether it fixes the issue.

blackrat (changes requested):

kicferk1:
> Can reproduce the API call that made this fail in a test or a feature test?

If you check out develop and enable the middleware in tests, the request tests for participant declarations fail. Sooo... We could, but it is already there.

config/environments/deployed_development.rb
jamesrose: I was definitely seeing a nil error or something somewhere re: this. Whether it came from this middleware I don't know.

spec/middlewares/lead_provider_request_auditor_spec.rb
kicferk1: In fact, if I change the auditor implementation to the old version, I see
```
ActionDispatch::Http::Parameters::ParseError: no implicit conversion of nil into String
```
in all `requests/api/v1/participant_declarations_spec.rb` tests. Which is, I think, what we were seeing before.

kicferk1:
I played with the review app. Sending `X_WITH_SERVER_DATE` with `2022-01-02T03:04:05.000Z` did indeed make a declaration with `created_at` in next year. It seems to work just fine.

I also had it with rack middlewares. I decided to move the auditing logic into a controller concern, so we get reasonable Rails behaviour, instead of worrying about reading request properties mutating that request.

config/environments/deployed_development.rb
ethanmills: I wonder if this was just showing up in the stacktrace as a caller?

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -55,9 +55,11 @@ def build_params(attributes)
       end
 
       it "create declaration record and declaration attempt and return id when successful" do
-        expect { post "/api/v1/participant-declarations", params: build_params(valid_params) }
+        params = build_params(valid_params)
+        expect { post "/api/v1/participant-declarations", params: params }
             .to change(ParticipantDeclaration, :count).by(1)
             .and change(ParticipantDeclarationAttempt, :count).by(1)
+        expect(ApiRequestAudit.last.body).to eq(params.to_s)
```
ethanmills, line 62: Implicit ordering with `last`, is it correct?

plisovin (approved):

spec/requests/api/v1/participant_declarations_spec.rb
kicferk1: Well, there should only be one, but I fixed it.

spec/middlewares/lead_provider_request_auditor_spec.rb
blackrat: I don't think that's enough. Would expect a test (possibly with a stub) to make sure that what is passed in isn't changed when it is forwarded

blackrat (approved):

PR #857, opened 2021-08-13
[CPDEL-788] Fix error for users with no cohort

FriedSock:
### Context

https://dfedigital.atlassian.net/browse/CPDEL-649

The previous method shouldn't be used any more as it uses attributes that aren't
the source of truth. We should look at the school cohort to find the user' cohort.

Also use the find_school_cohort method and amend the includes in the controller to
avoid N+1 lookups to find the cohort related to the school cohort.

The previous implementation was throwing an exception on the users index endpoint, 
which is bad because it prevents some users from being synced with engage and learn.

Unfortunately it is hard to create a test for this as a user can't be created without a cohort
without bypassing validations

### Review Checks

### DIFF
--- a/app/serializers/ecf_user_serializer.rb
+++ b/app/serializers/ecf_user_serializer.rb
@@ -63,7 +63,7 @@ class ECFUserSerializer
   end
 
   attributes :cohort do |user|
-    user.cohort.start_year
+    find_school_cohort(user)&.cohort&.start_year
   end
 
   def self.find_school_cohort(user)
--- a/app/controllers/api/v1/ecf_users_controller.rb
+++ b/app/controllers/api/v1/ecf_users_controller.rb
@@ -53,7 +53,10 @@ module Api
       def users
         users = User.where(id: ParticipantProfile::ECF.joins(:teacher_profile).select("teacher_profiles.user_id"))
                     .includes(teacher_profile: {
-                      ecf_profile: %i[school_cohort core_induction_programme],
+                      ecf_profile: [
+                        :core_induction_programme,
+                        school_cohort: :cohort,
+                      ],
                     })
 
         if updated_since.present?


ethanmills:
They should have a cohort through school_cohort

spec/requests/api/v1/ecf_users_spec.rb
```diff
@@ -73,6 +73,19 @@
         expect(parsed_response["data"][0]["attributes"]["cohort"]).to eql(2021)
       end
 
+      context "there is a user without a cohort" do
+        before do
+          school = create(:school)
+          teacher = create(:user, :teacher)
+          create(:participant_profile, participant_type: :ect, school: school, teacher_profile: teacher.teacher_profile)
```
ethanmills, line 80: We no longer set `school` or `cohort` on ParticipantProfile::ECT, but rather `school_cohort` which captures both

ethanmills:
We should probably remove the cohort method on User - it doesn't belong there. Accessing the cohort this way in the EcfUserSerializer will also likely cause N+1 queries - I missed the review that added it. 

I'll suggest a change shortly

app/serializers/ecf_user_serializer.rb
```diff
@@ -63,7 +63,7 @@ class ECFUserSerializer
   end
 
   attributes :cohort do |user|
-    user.cohort.start_year
+    user.cohort&.start_year
```
ethanmills, line 66:
```suggestion
    find_school_cohort(user)&.cohort&.start_year
```

FriedSock: 🤦 Can't believe I missed the method right there

spec/requests/api/v1/ecf_users_spec.rb
FriedSock: Hmm, struggling to think of a test to add instead of this now, this is how you would set up a user to break the previous implementation but now it is impossible to create a profile without a school cohort, or a school cohort without a cohort 🤔

ethanmills (approved):
Great, thanks for sorting!

PR #858, opened 2021-08-13
CPDTP-410 Adding integration tests for declarations API

plisovin:
### Context

Additional feature tests to cover the service-wide API solution, including:

- Provider collecting data via ECF or NPQ participant API endpoint
- Provider sending data via declaration API endpoint
- Validation occurring in Track and Pay
- Response sent to provider to confirm success or error


spec/features/participant-declarations/integration_test_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+           cohort: ect_profile.cohort,
+           delivery_partner: delivery_partner)
+
+    @participant_id = ect_profile.user.id
+  end
+
+  def when_the_participant_details_are_passed_to_the_lead_provider
+    @session.get("/api/v1/participants",
+                 headers: { "Authorization": "Bearer #{@token}" })
+    participants = JSON.parse(@session.response.body).dig("data").map { |a| a["id"] }
```
kicferk1, line 35: Let's use a better name than `a`. `participant`?

spec/features/participant-declarations/integration_test_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+    @params = common_params(@participant_id, "ecf-induction")
+  end
+
+  def then_the_declaration_made_is_valid
+    expect {
+      @response = @session.post("/api/v1/participant-declarations",
+                                params: @params,
+                                headers: { "Authorization": "Bearer #{@token}" })
+    }.to change(ParticipantDeclaration, :count).by(1)
+  end
```
kicferk1, line 49: I would set `@response` in `and_the_lead_provider_submits_a_declaration_for_the_participant_using_the_same_unique_id` - cause that's what it says it does. 

Then in `then_the_declaration_made_is_valid` we could use the id from the response and assert that a participant declaration with that id exists. I think that would still show validity.

spec/features/participant-declarations/integration_test_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+  #------------
+
+  scenario "ECT details sent to provider, declaration sent using same unique ID, errors exist in declaration" do
+    given_there_is_a_token
+    and_there_is_a_participant
+    and_there_is_wrong_declaration_params(@participant_id)
+
+    response = @session.post("/api/v1/participant-declarations", params: @params, headers: { "Authorization": "Bearer #{@token}" })
+    expect(response).to eq(422)
+  end
```
kicferk1, line 75: I like the scenario above much more - I think we want to make our methods as close to the descriptions in the acceptance criteria as possible :)

spec/features/participant-declarations/integration_test_spec.rb
```diff
@@ -0,0 +1,144 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.feature "Integration Test", type: :feature do
```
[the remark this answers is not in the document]
plisovin: Yeah, agree - renamed in 892c2e1

spec/features/participant-declarations/integration_test_spec.rb
plisovin: Done in 892c2e1

spec/features/participant-declarations/integration_test_spec.rb
plisovin: yeah, good idea, refactored in 892c2e1

spec/features/participant-declarations/integration_test_spec.rb
plisovin: Refactored all the other scenarios to the one above

spec/features/participant-declarations/submit_participant_declarations_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+    @session.get("/api/v1/participants",
+                 headers: { "Authorization": "Bearer #{@token}" })
+    participants = JSON.parse(@session.response.body).dig("data").map { |participant| participant["id"] }
+    expect(participants.first).to eq(@participant_id)
+  end
+
+  def and_the_lead_provider_submits_a_declaration_for_the_participant_using_the_same_unique_id(participant_id, course_identifier)
+    params = common_params(participant_id, course_identifier)
+    submit_request(params)
+  end
```
kicferk1, line 118: Should we split it into three methods to avoid the parameters?
Say
```
and_the_lead_provider_submits_a_declaration_for_the_ect_using_their_id
and_the_lead_provider_submits_a_declaration_for_the_mentor_using_their_id
and_the_lead_provider_submits_a_declaration_for_the_npq_using_their_id
```

spec/features/participant-declarations/submit_participant_declarations_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+
+  def and_the_lead_provider_submits_a_declaration_for_the_participant_using_the_same_unique_id_with_error(course_identifier)
+    params = common_params("", course_identifier)
+    params["data"]["attributes"].reject! { |a| a["participant_id"] }
+    submit_request(params)
+  end
```
kicferk1, line 124: `and_the_lead_provider_submits_a_declaration_without_participant_id` - also, maybe we don't care about course identifier here :)

spec/features/participant-declarations/submit_participant_declarations_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+  def setup
+    lead_provider = create(:lead_provider)
+    @cpd_lead_provider = create(:cpd_lead_provider, lead_provider: lead_provider)
+    @token = LeadProviderApiToken.create_with_random_token!(cpd_lead_provider: @cpd_lead_provider)
+    @session = ActionDispatch::Integration::Session.new(Rails.application)
+
+    @params = common_params(@participant_id, "ecf-induction")
+  end
+
+private
```
kicferk1, line 153: That's second `private` in the file, let's remove it.

spec/features/participant-declarations/submit_participant_declarations_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+
+  def submit_request(params)
+    @response_http_code = @session.post("/api/v1/participant-declarations",
+                                        params: params,
+                                        headers: { "Authorization": "Bearer #{@token}" })
```
kicferk1, line 158: I think we should use `body` rather than `params` - I think `params` will make url-encoded version fo what you give it, which is not the end of the world, but it is also not what we expect from LPs.

spec/features/participant-declarations/submit_participant_declarations_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+        "type":"participant-declaration",
+        "attributes": {
+           "participant_id": "#{participant_id}",
+           "declaration_type": "started",
+           "declaration_date": "2021-01-01T01:01:01.000Z",
+      		 "course_identifier": "#{course_identifier}"
+         }
+       }
+      }
+    DATA
```
kicferk1, line 176: Can we do something like
```
      {
      "data":{
        "type":"participant-declaration",
        "attributes": {
           "participant_id": "#{participant_id}",
           "declaration_date": "2021-01-01T01:01:01.000Z",
      		 "course_identifier": "#{course_identifier}"
         }
       }
      }.to_json
```
?
Like on https://github.com/DFE-Digital/early-careers-framework/pull/853/files#diff-c04e64bd0173a9b1c625d63330800bfaf04749509a1d200fb79ccadfcc9c8878R1?

kicferk1 (commented):
Looks pretty good, I would prefer no parameters in methods we use for testing, unless it is absolutely necessary - they break the english rules a bit.

If we need them, we should do stuff like `someone_does_something_with_id(come_id)` to make it more readable.

spec/features/participant-declarations/submit_participant_declarations_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+  def given_an_npq_participant_has_been_entered_onto_the_dfe_service
+    npq_lead_provider = create(:npq_lead_provider, cpd_lead_provider: @cpd_lead_provider)
+    npq_course = create(:npq_course, identifier: "npq-leading-teaching")
+    @npq_profile_id = create(:npq_validation_data,
+                             npq_lead_provider: npq_lead_provider,
+                             npq_course: npq_course).user.id
+  end
+
+  def when_the_participant_details_are_passed_to_the_lead_provider
+    @session.get("/api/v1/participants",
```
plisovin, line 109: btw - I don't think npq user id will be returned in the ```/participants``` API call

spec/features/participant-declarations/submit_participant_declarations_spec.rb
plisovin: I think it's slightly different functionality, there is no body https://www.rubydoc.info/docs/rails/4.1.7/ActionDispatch/Integration/RequestHelpers#post-instance_method

kicferk1 (approved):
I think the tests look good, probably worth chatting to Rhys about whether we have a ticket for giving LPs npq participants.

PR #859, opened 2021-08-13
Add milestone date validation

kicferk1:
### Context
We want to check that the declaration has been sent in the correct milestone period.

This does not check that the declaration type (started, retained) matches the milestone. 

1. Add start dates to milestones
2. Remove non-standard schedules from the seed
3. Make all participant declaration tests aware that they need to send the correct dates on them. 

### Guidance to review
As you can see, if a participant has no schedule, the API call will fail - cause it cannot check the dates. I can temporarily change it to be an early return (so the validation passes), but it really tells me we should populate schedules on participant profiles. 

I need to set up schedules for all test participants so this can be properly tested. 

So I think I will extract the migration part from this PR, add a rake task, deploy everywhere and run a script so all participants are on September schedule. 

### Review Checks

Try sending a declaration for a participant before 1st of September. It should fail.

### DIFF
--- a/app/services/record_declarations/base.rb
+++ b/app/services/record_declarations/base.rb
@@ -3,6 +3,7 @@
 module RecordDeclarations
   class Base
     include ActiveModel::Model
+    RCF3339_DATE_REGEX = /\A\d{4}-\d{2}-\d{2}T(\d{2}):(\d{2}):(\d{2})([\.,]\d+)?(Z|[+-](\d{2})(:?\d{2})?)?\z/i.freeze
 
     attr_accessor :course_identifier, :user_id, :lead_provider_from_token, :declaration_date, :declaration_type, :evidence_held
 
@@ -10,13 +11,12 @@ module RecordDeclarations
     validates :declaration_type, inclusion: { in: :valid_declaration_types, message: I18n.t(:invalid_declaration_type) }
     validates :course_identifier, presence: { message: I18n.t(:missing_course_identifier) }
     validates :declaration_date, presence: { message: I18n.t(:missing_declaration_date) }
-    validates :declaration_date, declaration_date: true, allow_blank: true
-    validates :declaration_date, future_date: true, allow_blank: true
     validates :declaration_type, presence: { message: I18n.t(:missing_declaration_type) }
     validates :user, presence: { message: I18n.t(:invalid_participant) }
     validates :lead_provider_from_token, presence: { message: I18n.t(:missing_lead_provider) }
 
     validate :profile_exists
+    validate :date_has_the_right_format
 
     def profile_exists
       return if errors.any?
@@ -24,6 +24,15 @@ module RecordDeclarations
       unless user_profile
         errors.add(:user_profile, I18n.t(:invalid_participant))
       end
+    end
+
+    def date_has_the_right_format
+      return if declaration_date.blank?
+
+      errors.add(:declaration_date, I18n.t(:invalid_declaration_date)) unless declaration_date.match(RCF3339_DATE_REGEX)
+      errors.add(:declaration_date, I18n.t(:future_declaration_date)) if parsed_date > Time.zone.now
+    rescue StandardError
+      errors.add(:declaration_date, I18n.t(:invalid_declaration_date))
     end
 
     def valid_courses_for_user
@@ -113,9 +122,7 @@ module RecordDeclarations
     end
 
     def parsed_date
-      ActiveRecord::Type::Date.new.cast(declaration_date)
-    rescue StandardError
-      raise ActionController::ParameterMissing, I18n.t(:invalid_declaration_date)
+      Time.zone.parse(declaration_date)
     end
 
     def validate_schedule!
--- a/app/services/record_declarations/base.rb
+++ b/app/services/record_declarations/base.rb
@@ -128,11 +128,11 @@ module RecordDeclarations
     end
 
     def validate_milestone!
-      unless milestone.start_date < parsed_date
+      unless milestone.start_date.beginning_of_day < parsed_date
         raise ActionController::ParameterMissing, I18n.t(:declaration_before_milestone_start)
       end
 
-      unless milestone.milestone_date > parsed_date
+      unless parsed_date <= milestone.milestone_date.end_of_day
         raise ActionController::ParameterMissing, I18n.t(:declaration_after_milestone_cutoff)
       end
     end
@@ -142,17 +142,17 @@ module RecordDeclarations
         raise ActionController::ParameterMissing, I18n.t(:schedule_missing)
       end
 
-      declaration_to_milestone_map[declaration_type.to_sym]
+      declaration_to_milestone_map[declaration_type]
     end
 
     def declaration_to_milestone_map
       {
-        "started": schedule.milestones[0],
-        "retained-1": schedule.milestones[1],
-        "retained-2": schedule.milestones[2],
-        "retained-3": schedule.milestones[3],
-        "retained-4": schedule.milestones[4],
-        "completed": schedule.milestones.last,
+        "started" => schedule.milestones[0],
+        "retained-1" => schedule.milestones[1],
+        "retained-2" => schedule.milestones[2],
+        "retained-3" => schedule.milestones[3],
+        "retained-4" => schedule.milestones[4],
+        "completed" => schedule.milestones.last,
       }
     end
   end
--- a/app/services/record_declarations/base.rb
+++ b/app/services/record_declarations/base.rb
@@ -14,6 +14,7 @@ module RecordDeclarations
     validates :declaration_type, presence: { message: I18n.t(:missing_declaration_type) }
     validates :user, presence: { message: I18n.t(:invalid_participant) }
     validates :lead_provider_from_token, presence: { message: I18n.t(:missing_lead_provider) }
+    validates :parsed_date, future_date: true, allow_blank: true
 
     validate :profile_exists
     validate :date_has_the_right_format
@@ -116,7 +117,7 @@ module RecordDeclarations
       return if declaration_date.blank?
 
       errors.add(:declaration_date, I18n.t(:invalid_declaration_date)) unless declaration_date.match(RFC3339_DATE_REGEX)
-      errors.add(:declaration_date, I18n.t(:future_declaration_date)) if parsed_date > Time.zone.now
+      parsed_date
     rescue StandardError
       errors.add(:declaration_date, I18n.t(:invalid_declaration_date))
     end
--- a/app/services/record_declarations/base.rb
+++ b/app/services/record_declarations/base.rb
@@ -139,7 +139,6 @@ module RecordDeclarations
       end
 
       next_milestone = schedule.milestones[existing_declarations.count]
-      parsed_date = self.parsed_date
       unless next_milestone.start_date < parsed_date
         raise ActionController::ParameterMissing, I18n.t(:declaration_before_milestone_start)
       end
--- a/app/services/record_declarations/base.rb
+++ b/app/services/record_declarations/base.rb
@@ -59,6 +59,7 @@ module RecordDeclarations
 
       declaration_attempt = create_declaration_attempt!
       validate_provider!
+      validate_schedule!
       declaration = create_record!
       declaration_attempt.update!(participant_declaration: declaration)
 
@@ -109,6 +110,36 @@ module RecordDeclarations
     def validate_provider!
       # TODO: Remove the nil? check and fix the test setup so that they build the school cohort, partnership and give us back the actual lead_provider.
       raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || lead_provider_from_token == actual_lead_provider
+    end
+
+    def parsed_date
+      ActiveRecord::Type::Date.new.cast(declaration_date)
+    rescue StandardError
+      raise ActionController::ParameterMissing, I18n.t(:invalid_declaration_date)
+    end
+
+    def validate_schedule!
+      schedule = user_profile.schedule
+
+      unless schedule
+        raise ActionController::ParameterMissing, I18n.t(:schedule_missing)
+      end
+
+      existing_declarations = user_profile.participant_declarations
+
+      if existing_declarations.count >= schedule.milestones.count
+        raise ActionController::ParameterMissing, I18n.t(:too_many_declarations)
+      end
+
+      next_milestone = schedule.milestones[existing_declarations.count]
+      parsed_date = self.parsed_date
+      unless next_milestone.start_date < parsed_date
+        raise ActionController::ParameterMissing, I18n.t(:declaration_before_milestone_start)
+      end
+
+      unless next_milestone.milestone_date > parsed_date
+        raise ActionController::ParameterMissing, I18n.t(:declaration_after_milestone_cutoff)
+      end
     end
   end
 end
--- a/app/services/record_declarations/base.rb
+++ b/app/services/record_declarations/base.rb
@@ -18,31 +18,6 @@ module RecordDeclarations
     validate :profile_exists
     validate :date_has_the_right_format
 
-    def profile_exists
-      return if errors.any?
-
-      unless user_profile
-        errors.add(:user_profile, I18n.t(:invalid_participant))
-      end
-    end
-
-    def date_has_the_right_format
-      return if declaration_date.blank?
-
-      errors.add(:declaration_date, I18n.t(:invalid_declaration_date)) unless declaration_date.match(RCF3339_DATE_REGEX)
-      errors.add(:declaration_date, I18n.t(:future_declaration_date)) if parsed_date > Time.zone.now
-    rescue StandardError
-      errors.add(:declaration_date, I18n.t(:invalid_declaration_date))
-    end
-
-    def valid_courses_for_user
-      valid_courses = []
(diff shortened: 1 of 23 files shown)


app/services/record_declarations/base.rb
```diff
@@ -110,5 +111,35 @@ def validate_provider!
       # TODO: Remove the nil? check and fix the test setup so that they build the school cohort, partnership and give us back the actual lead_provider.
       raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || lead_provider_from_token == actual_lead_provider
     end
+
+    def parsed_date
+      ActiveRecord::Type::Date.new.cast(declaration_date)
```
jamesrose, line 116: Does this raise the error below?

kicferk1: Good question. I think we should never actually get in here, I will refactor the setup to make sure of that.

spec/services/record_participant_declaration_service_spec.rb
```diff
@@ -2,62 +2,27 @@
 
 require "rails_helper"
 
+require_relative "../shared/context/service_record_declaration_params.rb"
```
asmega, line 5: i don't think i like this for several reasons:
- we're just hiding a load of test setup elsewhere in the code. if someone needs to look into these tests they'd have to look at the shared contexts.
- because the variable names are elsewhere you have to figure out how they affect the current file despite being elsewhere 
- if they want to change any of test setup the will affect every other test setup where it is being called.
- there are 'let!' defined so as soon as the context is called it has affected state
- it loads NPQ and ECF when someone may just want to test one thing

kicferk1: Prior to having this setup we were literally copy-pasting the same setup across many different files. It looked much worse than it does now.

That said, I am open to suggestions on how to improve it. Or which of a myriad other solutions for commonising test setup would be better. But the short gist of the problem is that T&P testing scenarios tend to require a bunch of involved setup (schools which are in partnership with LPs and have some teachers for ECF part is probably the worst), and I don't want to have to copy-paste it whenever I need to make a new test suite.

asmega: I guess, we'll try it out and see what happens. hopefully it'll just work

app/services/record_declarations/base.rb
```diff
@@ -26,6 +26,15 @@ def profile_exists
       end
     end
 
+    def date_has_the_right_format
```
asmega, line 29: are you able to make this private if only used by validation?

asmega: personally i like to prefix validation methods with `validate_` that way i know it'll add errors potentially. otherwise someone might think `date_has_the_right_format` just returns a boolean with no side effects. i guess this is optional

app/services/record_declarations/base.rb
```diff
@@ -26,6 +26,15 @@ def profile_exists
...
+      return if declaration_date.blank?
+
+      errors.add(:declaration_date, I18n.t(:invalid_declaration_date)) unless declaration_date.match(RCF3339_DATE_REGEX)
```
asmega, line 32: if `invalid_declaration_date` is a new error message i expected a new respective entry in the locales file

app/services/record_declarations/base.rb
kicferk1: Hmm, so then the top of the file would read
```
validate :validate_xxx
```

Feels a bit awkward. I made it and a few other methods private.

app/services/record_declarations/base.rb
asmega: so we seem to have a load translations including `invalid_declaration_date` at the root level of the locales file. these should all be namespaced accordingly. otherwise we're gonna get a very flat locale file which i think is undesired. so the line should read
```
      errors.add(:declaration_date, :invalid_declaration_date) unless declaration_date.match(RCF3339_DATE_REGEX)
```
the locale should be on something like
```
en.activemodel.errors.models.OBJECT_NAME.attributes.declaration_date.invalid_declaration_date
```
however this is probably outside the scope of this PR but something that may want addressing soon

kicferk1: I will look into it in a spare moment. Ehh, I hate i18n.

app/services/record_declarations/base.rb
```diff
@@ -110,5 +120,33 @@ def validate_provider!
...
+      end
+
+      existing_declarations = user_profile.participant_declarations
+
+      if existing_declarations.count >= schedule.milestones.count
+        raise ActionController::ParameterMissing, I18n.t(:too_many_declarations)
+      end
+
+      next_milestone = schedule.milestones[existing_declarations.count]
+      parsed_date = self.parsed_date
```
asmega, line 142: i think you can delete this line. i don't see the point of assigning the method return to a local variable of the same name?

kicferk1: Ah, I think this is an artifact of an old refactor - when I was parsing inline

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -14,29 +14,17 @@
       {
         participant_id: ect_profile.user.id,
         declaration_type: "started",
-        declaration_date: (Time.zone.now - 1.week).iso8601,
+        declaration_date: (ect_profile.schedule.milestones.first.start_date + 1.day).rfc3339,
```
asmega, line 17: why are we changing from `iso8601` to `rfc3339`? there are other places using `iso8601` in the API. we should probably change these globally for our api. otherwise some endpoints need `iso8601` and others `rfc3339` which as an api consumer is a bad experience

app/services/record_declarations/base.rb
```diff
@@ -106,9 +95,57 @@ def create_record!
...
+      errors.add(:declaration_date, I18n.t(:invalid_declaration_date))
+    end
+
     def validate_provider!
       # TODO: Remove the nil? check and fix the test setup so that they build the school cohort, partnership and give us back the actual lead_provider.
       raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || lead_provider_from_token == actual_lead_provider
     end
+
+    def validate_schedule!
+      schedule = user_profile.schedule
```
asmega, line 129: this is ripe for extracting to a method along with `existing_declarations` and `next_milestone` unless it is problematic?

spec/requests/api/v1/participant_declarations_spec.rb
kicferk1: The reason I changed it here is because `(ect_profile.schedule.milestones.first.start_date + 1.day).iso8601` does not work - presumably it does not include the time part, which `rfc3339` does. If I use `(ect_profile.schedule.milestones.first.start_date + 24.hours). iso8601` it works.

The larger point of 'we should use the same format across the API'. Yeah. We probably should. Something to improve in another PR.

app/services/record_declarations/base.rb
```diff
@@ -3,38 +3,23 @@
 module RecordDeclarations
   class Base
     include ActiveModel::Model
+    RCF3339_DATE_REGEX = /\A\d{4}-\d{2}-\d{2}T(\d{2}):(\d{2}):(\d{2})([\.,]\d+)?(Z|[+-](\d{2})(:?\d{2})?)?\z/i.freeze
```
ethanmills, line 6: RFC

ethanmills: Also, can we use https://apidock.com/rails/ActiveSupport/TimeZone/rfc3339 ?

ethanmills: Ah sorry, just noticed this was a copy-paste. Let's at least fix the typo

spec/requests/api/v1/participant_declarations_spec.rb
asmega: if you did not explicitly want to change the API, i recommend `(ect_profile.schedule.milestones.first.start_date + 24.hours). iso8601` over the current change. as with this change we are not explicitly saying the client should now send us `rfc3339` and we are testing for that and not `iso8601`.

If you did want to change to `rfc3339` the API the docs at https://ecf-sandbox.london.cloudapps.digital/lead-providers/guidance/reference#post-api-v1-participant-declarations will need updating. But i'd like to also know why we decided to change.

spec/factories/participant_profiles.rb
```diff
@@ -17,6 +17,7 @@
       end
     end
 
+    schedule { create(:schedule) }
```
ethanmills, line 20: Why do we need the create here?

spec/requests/api/v1/participant_declarations_spec.rb
ethanmills: rfc3339 is a profile of iso8601. All rfc3339 is iso8601, but not the other way around

ethanmills (approved):

spec/requests/api/v1/participant_declarations_spec.rb
kicferk1: The docs say 
```
string, ISO 8601 date with time and timezone
```
ISO without time and timezone is invalid according to them, although I guess we could say `rfc3339` in there.

As for why, well, we care about the exact time on declarations. I think in general this is not true in all of our endpoints - things like date of birth won't require this degree of precision.

spec/factories/participant_profiles.rb
kicferk1: We probably don't :)

spec/requests/api/v1/participant_declarations_spec.rb
blackrat: We specify ISO 8601 date with time and timezone, which is what RFC3339 defines as a stricter parsing format. From the Ruby Docs:

"If the time or zone components are missing then an ArgumentError will be raised. This is much stricter than either parse or iso8601 which allow for missing components."

As we don't want to allow missing components, this is what is defined by RFC3339.

app/services/record_declarations/base.rb
```diff
@@ -106,9 +96,56 @@ def create_record!
       end
     end
 
+    def valid_courses_for_user
+      valid_courses = []
+      valid_courses << "ecf-mentor" if user.mentor?
+      valid_courses << "ecf-induction" if user.early_career_teacher?
+      valid_courses += NPQCourse.all.map(&:identifier) if user.npq?
```
blackrat, line 103: It's a design smell for the Base classes to have knowledge of subclasses and this is just proxying that.

app/services/record_declarations/base.rb
```diff
@@ -106,9 +96,56 @@ def create_record!
...
+      unless user_profile
+        errors.add(:user_profile, I18n.t(:invalid_participant))
+      end
+    end
+
+    def date_has_the_right_format
+      return if declaration_date.blank?
+
+      errors.add(:declaration_date, I18n.t(:invalid_declaration_date)) unless declaration_date.match(RFC3339_DATE_REGEX)
+      errors.add(:declaration_date, I18n.t(:future_declaration_date)) if parsed_date > Time.zone.now
```
blackrat, line 119: I'd have expected to see the format and business future date logic as separate validations.

blackrat (changes requested):

app/services/record_declarations/base.rb
```diff
@@ -106,9 +96,56 @@ def create_record!
...
+    end
+
+    def next_milestone
+      unless schedule
+        raise ActionController::ParameterMissing, I18n.t(:schedule_missing)
+      end
+
+      if participant_declarations.count >= schedule.milestones.count
+        raise ActionController::ParameterMissing, I18n.t(:too_many_declarations)
+      end
```
blackrat, line 146: What's this? There's no business decision on restricting the number of participant_declarations yet.

blackrat (changes requested):

app/services/record_declarations/base.rb
kicferk1: Base class has no knowledge what subclasses are doing.

What this computes is the valid courses for the user. Those do not change in child classes. We then check that the course we are sent is in the list of courses valid for the user. This again, does not change in child classes.

In fact, which chilld class is used to validate is picked on the basis of course identifier. So if we wanted to move this check to child classes, we would have to validate that the user has access to the course which the child class corresponds to. Which I guess could make sense...

Something like
```
# In mentor child class

def validate_user_can_access_course
  errors.add(...) unless user.mentor?
end
```

But definitely not in this PR. The only reason it touches this stuff is to move it from non-private to private.

app/services/record_declarations/base.rb
kicferk1: This is what protects us from out of bound exception.

blackrat: Out of bounds how? That's really nothing to do with milestone dates and doesn't seem to be a good fit for this PR.

spec/requests/api/v1/participant_declarations_spec.rb
asmega: I'm happy to moving to `rfc3339`. I don't believe they are quite super/sub sets of one another. and there are subtle differences. here is a naive example for the ruby implementation
```
irb(main):030:0> DateTime.iso8601("20-12-09T16:09:53+00:00")
=> Wed, 09 Dec 2020 16:09:53 +0000
irb(main):031:0> DateTime.rfc3339("20-12-09T16:09:53+00:00")
Traceback (most recent call last):
        2: from (irb):31
        1: from (irb):31:in `rfc3339'
Date::Error (invalid date)
```
So I think we just want to update the docs to reflect that and remove any ambiguity. Happy for this to be in another PR

asmega (approved):

app/services/record_declarations/base.rb
kicferk1: There are like 5 or 6 milestones per schedule. To get the milestone we check against this implementation says 'count the existing declarations, assume they are 1-1 to the milestones, and so pick the milestone in the ordered list of milestones that is next'. And what happens when there is no next and we receive a declaration?

I don't expect this solution will stay in forever - one other thing we want to validate, which will probably superseed this, is that the declaration type is right for the milestone in question. But until we do that, this is a simple way of fetching a milestone and showing that date validations work for this milestone.

spec/requests/api/v1/participant_declarations_spec.rb
blackrat: The former is an invalid date since it doesn't say what the offset it being applied to. For the full representation it needs to specify offset from a specific timezone. It's entirely applicable for that offset to be from localtime, but the Z forces it to be from UTC. Hence it's unambiguous adoption.

app/services/record_declarations/base.rb
```diff
@@ -106,9 +97,56 @@ def create_record!
...
+
+      schedule.milestones[participant_declarations.count]
```
blackrat, line 149: This would be better as a lookup table. It's not just an index into the milestones.

app/services/record_declarations/base.rb
```diff
@@ -106,9 +97,63 @@ def create_record!
...
+    rescue StandardError
+      errors.add(:declaration_date, I18n.t(:invalid_declaration_date))
+    end
+
     def validate_provider!
       # TODO: Remove the nil? check and fix the test setup so that they build the school cohort, partnership and give us back the actual lead_provider.
       raise ActionController::ParameterMissing, I18n.t(:invalid_participant) unless actual_lead_provider.nil? || lead_provider_from_token == actual_lead_provider
     end
+
+    def validate_milestone!
```
blackrat, line 130: Probably one of these should actually match the date too (i.e. <= or >=)

app/services/record_declarations/base.rb
blackrat: Sure. The "can access course" shouldn't be necessary if the factory is doing the correct job. This check was more relevant for schema validation, so the mentor child class should only need to validate that it's the correct course without the unless user.mentor? That would be implied by the fact that you're using that class in the first place.

app/services/record_declarations/base.rb
```diff
@@ -106,9 +97,63 @@ def create_record!
...
+        raise ActionController::ParameterMissing, I18n.t(:declaration_after_milestone_cutoff)
+      end
+    end
+
+    def milestone
+      unless schedule
+        raise ActionController::ParameterMissing, I18n.t(:schedule_missing)
+      end
+
+      declaration_to_milestone_map[declaration_type.to_sym]
```
blackrat, line 145: Shouldn't switch from string to symbol if the keys are still strings.

app/services/record_declarations/base.rb
```diff
@@ -106,9 +97,63 @@ def create_record!
...
+
+    def declaration_to_milestone_map
+      {
+        "started": schedule.milestones[0],
+        "retained-1": schedule.milestones[1],
+        "retained-2": schedule.milestones[2],
+        "retained-3": schedule.milestones[3],
+        "retained-4": schedule.milestones[4],
+        "completed": schedule.milestones.last,
+      }
```
blackrat, line 156: As in: these keys.

app/services/record_declarations/base.rb
```diff
@@ -106,9 +97,63 @@ def create_record!
...
+    end
```
blackrat, line 157: Probably easier to follow is you do

`declaration_to_milestone_map[declaration_type.underscore.to_sym]`

and have

`def declaration_to_milestone_map
{
started: schedule.milestone[0],
retained_1: schedule.milestons[1],
.
.
.
etc.
}`

blackrat (approved):

PR #860, opened 2021-08-13
Populate schedules on participant profiles

kicferk1:
### Context
I think every participant profile should have a schedule. That's because we will want to validate whether the declaration date on the declarations LPs send us matches that schedule.

Find all places where we create participant profiles, and add a default schedule to them. Yes, NPQ participants will need their own schedule, I think, but I don't know what it is - when we do we will add it to prod and run an update.

When this is merged and deployed, I will create the default schedule on each environment, and run something like 
```
ParticipantProfile.each { |profile| profile.update!(schedule: Finance::Schedule.default) }
```

### Guidance to review
I added `start_date` migration here because I want the milestones I create in production to have the right shape. Or as right as they can have as far as I know.

### DIFF
--- a/db/seeds/schedules.rb
+++ b/db/seeds/schedules.rb
@@ -2,65 +2,17 @@
 
 ecf_september_standard_2021 = Finance::Schedule.find_or_create_by!(name: "ECF September standard 2021")
 [
-  { name: "Output 1 - Participant Start", milestone_date: Date.new(2021, 10, 31), payment_date: Date.new(2021, 11, 30) },
-  { name: "Output 2 – Retention Point 1", milestone_date: Date.new(2022, 1, 31), payment_date: Date.new(2022, 2, 28) },
-  { name: "Output 3 – Retention Point 2", milestone_date: Date.new(2022, 4, 30), payment_date: Date.new(2022, 5, 31) },
-  { name: "Output 4 – Retention Point 3", milestone_date: Date.new(2022, 9, 30), payment_date: Date.new(2022, 10, 31) },
-  { name: "Output 5 – Retention Point 4", milestone_date: Date.new(2023, 1, 31), payment_date: Date.new(2023, 2, 28) },
-  { name: "Output 6 – Participant Completion", milestone_date: Date.new(2023, 4, 30), payment_date: Date.new(2023, 5, 31) },
+  { name: "Output 1 - Participant Start", start_date: Date.new(2021, 9, 1), milestone_date: Date.new(2021, 10, 31), payment_date: Date.new(2021, 11, 30) },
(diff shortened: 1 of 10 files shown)


ethanmills:
Could this not all just be 

```ruby
before_create :set_default_payment_schedule

def set_default_payment_schedule
  self.payment_schedule ||= Finance::Schedule.default
end
```
?

db/seeds/schedules.rb
```diff
@@ -2,65 +2,17 @@
...
-  { name: "Output 6 – Participant Completion", milestone_date: Date.new(2023, 4, 30), payment_date: Date.new(2023, 5, 31) },
+  { name: "Output 1 - Participant Start", start_date: Date.new(2021, 9, 1), milestone_date: Date.new(2021, 10, 31), payment_date: Date.new(2021, 11, 30) },
+  { name: "Output 2 – Retention Point 1", start_date: Date.new(2021, 11, 1), milestone_date: Date.new(2022, 1, 31), payment_date: Date.new(2022, 2, 28) },
+  { name: "Output 3 – Retention Point 2", start_date: Date.new(2022, 2, 1), milestone_date: Date.new(2022, 4, 30), payment_date: Date.new(2022, 5, 31) },
+  { name: "Output 4 – Retention Point 3", start_date: Date.new(2022, 5, 1), milestone_date: Date.new(2022, 9, 30), payment_date: Date.new(2022, 10, 31) },
+  { name: "Output 5 – Retention Point 4", start_date: Date.new(2022, 10, 1), milestone_date: Date.new(2023, 1, 31), payment_date: Date.new(2023, 2, 28) },
+  { name: "Output 6 – Participant Completion", start_date: Date.new(2023, 2, 1), milestone_date: Date.new(2023, 4, 30), payment_date: Date.new(2023, 5, 31) },
 ].each do |hash|
   Finance::Milestone.find_or_create_by!(
```
ethanmills, line 12: has this already been run?

ethanmills:
If we intend for every ParticipantProfile to have a schedule, are we going to enforce that with a db constraint once the data is right?

db/seeds/schedules.rb
kicferk1: Pretty sure not, I will check before running it.

kicferk1:
> Could this not all just be
> 
> ```ruby
> before_create :set_default_payment_schedule
> 
> def set_default_payment_schedule
>   self.schedule ||= Finance::Schedule.default
> end
> ```
> 
> ?

It probably could. But I am not sure how much we as a team like model callbacks, I reckon not much.

kicferk1:
> If we intend for every ParticipantProfile to have a schedule, are we going to enforce that with a db constraint once the data is right?

Yep.

ethanmills (approved):

PR #861, opened 2021-08-13
Only show validated TRNs after ECF validation

ethanmills:
There is a chance that something could go wrong at ECF TRN validation
that means a TRN we get from NPQ is not correct. Be conservative in
declaring that the TRN is validated.

Also, share some TRN pre-validation

### DIFF
--- a/app/serializers/participant_serializer.rb
+++ b/app/serializers/participant_serializer.rb
@@ -16,6 +16,17 @@ class ParticipantSerializer
     def participant_active?(user)
       user.teacher_profile.ecf_profile&.active?
     end
+
+    def trn(user)
+      user.teacher_profile.trn || user.teacher_profile.ecf_profile.ecf_participant_validation_data&.trn
+    end
+
+    def validated_trn(user)
+      eligibility_status = user.teacher_profile.ecf_profile.ecf_participant_eligibility&.status
+      if %w[matched eligible].include?(eligibility_status)
+        user.teacher_profile.trn
+      end
+    end
   end
 
   set_id :id
@@ -49,11 +60,11 @@ class ParticipantSerializer
   end
 
   active_participant_attribute :teacher_reference_number do |user|
-    user.teacher_profile.trn
+    trn(user)
   end
 
   active_participant_attribute :teacher_reference_number_validated do |user|
-    user.teacher_profile.trn.present?
+    trn(user).nil? ? nil : validated_trn(user).present?
   end
 
   active_participant_attribute :eligible_for_funding do |user|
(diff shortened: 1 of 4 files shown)


jamesrose (approved):

ethanmills:
I always forget about the other set of specs 🙄

app/serializers/participant_serializer.rb
```diff
@@ -16,6 +16,17 @@ def active_participant_attribute(attr, &blk)
...
+    def trn(user)
+      user.teacher_profile.trn || user.teacher_profile.ecf_profile.ecf_participant_validation_data&.trn
+    end
+
+    def validated_trn(user)
+      eligibility_status = user.teacher_profile.ecf_profile.ecf_participant_eligibility&.status
+      if %w[matched eligible].include?(eligibility_status)
+        user.teacher_profile.trn
+      end
+    end
```
tonyheadford, line 29: So, if there is an existing TRN from NPQ and we validate a different TRN, we always use the existing TRN not the validated one?

spec/serializers/participant_serializer_spec.rb
```diff
@@ -4,20 +4,30 @@
...
     let(:mentor_cohort) { mentor.mentor_profile.cohort }
 
-    it "outputs correctly formatted serialized Mentors" do
-      expected_json_string = "{\"data\":{\"id\":\"#{mentor.id}\",\"type\":\"participant\",\"attributes\":{\"email\":\"#{mentor.email}\",\"full_name\":\"#{mentor.full_name}\",\"mentor_id\":null,\"school_urn\":\"#{mentor.mentor_profile.school.urn}\",\"participant_type\":\"mentor\",\"cohort\":#{mentor_cohort.start_year},\"status\":\"active\",\"teacher_reference_number\":\"#{mentor.teacher_profile.trn}\",\"teacher_reference_number_validated\":true,\"eligible_for_funding\":null,\"pupil_premium_uplift\":null,\"sparsity_uplift\":null}}}"
-      expect(ParticipantSerializer.new(mentor).serializable_hash.to_json).to eq expected_json_string
-    end
+    context "when the participant is eligible for funding" do
+      before do
+        eligibility = ECFParticipantEligibility.create!(participant_profile: ect_profile)
+        eligibility.update!(status: :eligible)
```
tonyheadford, line 17: You're missing lots of sugar, as this is an enum you can do this instead:
```suggestion
        eligibility.eligible_status!
```

app/serializers/participant_serializer.rb
tonyheadford: Although in that case we would be `:manual_check` won't we?

ethanmills: Yep - so we will share the NPQ one and `teacher_reference_number_validated`  will be false

tonyheadford (approved):

PR #863, opened 2021-08-13
Refactor tests around NPQ

asmega:
### Context

- There is a callback `NPQValidationData#update_participant_profile`
- It creates and persists multiple database records
- It needs to be removed so we can have finer grain control of when it happens and the behaviour of it
- However many tests rely on this mechanism for the setup of its test data
- Therefore removing this method cannot be done first time round
- By reducing the dependency of this mechanism in the tests and instead setting up the data correctly this callback can be more easily removed in the next change

- Removes most of test reliance on `NPQValidationData#update_participant_profile`

### Guidance to review

- There should be no behavioural change
- The tests should continue to work if `NPQValidationData#update_participant_profile` had been deleted

### Testing

- n/a

### Review Checks

- No behavioural changes have been made so app should behave the same

### DIFF
--- a/spec/services/record_participant_declaration_service_spec.rb
+++ b/spec/services/record_participant_declaration_service_spec.rb
@@ -5,85 +5,119 @@ require "rails_helper"
 RSpec.describe RecordParticipantDeclaration do
   let(:ecf_lead_provider) { create(:lead_provider) }
   let(:cpd_lead_provider) { create(:cpd_lead_provider, lead_provider: ecf_lead_provider) }
-  let(:another_lead_provider) { create(:cpd_lead_provider, name: "Unknown") }
+  let(:another_cpd_lead_provider) { create(:cpd_lead_provider, name: "Unknown") }
   let(:npq_lead_provider) { create(:npq_lead_provider, cpd_lead_provider: cpd_lead_provider) }
-  let!(:npq_course) { create(:npq_course, identifier: "npq-leading-teaching") }
-  let(:npq_profile) do
-    create(:npq_validation_data,
-           npq_lead_provider: npq_lead_provider,
-           npq_course: npq_course)
-  end
+
+  let(:user) { create(:user) }
+  let(:teacher_profile) { create(:teacher_profile, user: user) }
   let(:induction_coordinator_profile) { create(:induction_coordinator_profile) }
-  let(:params) do
-    {
-      user_id: npq_profile.user.id,
-      declaration_date: "2021-06-21T08:46:29Z",
-      declaration_type: "started",
-      course_identifier: "npq-leading-teaching",
-      lead_provider_from_token: another_lead_provider,
-    }
-  end
 
   context "when sending event for an npq course" do
-    let(:npq_params) do
-      params.merge({ lead_provider_from_token: cpd_lead_provider })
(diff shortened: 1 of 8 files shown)


spec/services/record_participant_declaration_service_spec.rb
```diff
@@ -5,85 +5,119 @@
...
       it "raises a ParameterMissing error" do
         expect { described_class.call(params) }.to raise_error(ActionController::ParameterMissing)
       end
     end
 
     context "when valid user is an early_career_teacher" do
+      before do
+        school_cohort
+        partnership
+      end
```
kicferk1, line 79: Can we use `let!` where we define them to ensure they are run before tests?

spec/services/record_participant_declaration_service_spec.rb
```diff
@@ -5,85 +5,119 @@
 RSpec.describe RecordParticipantDeclaration do
   let(:ecf_lead_provider) { create(:lead_provider) }
   let(:cpd_lead_provider) { create(:cpd_lead_provider, lead_provider: ecf_lead_provider) }
-  let(:another_lead_provider) { create(:cpd_lead_provider, name: "Unknown") }
+  let(:another_cpd_lead_provider) { create(:cpd_lead_provider, name: "Unknown") }
```
kicferk1, line 8: I smell... conflicts.

In here and in shared setup, with the work I've done on https://github.com/DFE-Digital/early-careers-framework/pull/859/files#diff-b04de0b996bbd7c765305b53215f91f9edad6ef4df41f5fe83615d74a3ffb898L5

I'm happy to deal with the shared setup conflicts myself, but could you use the diff from the link above to make this test use common track and pay test setup? When you do, I hope things will 'just work' - and if not I am happy to pair on it.

asmega: I think it's just easier to fix the conflicts after the merge. Feel free to merge first and I'll patch up afterwards

spec/services/record_participant_declaration_service_spec.rb
asmega: I can't replace the `let` with `let!`. The test will fail as the behaviour and outcome would be different. The tests are written in a quirky way. Keeping them lazy then evaluating when needed seems to me the best solution. Please let me know if you have a better suggestion

spec/services/record_participant_declaration_service_spec.rb
asmega: oh wait. i misunderstood and was reading another part of the PR. hmm do we really want these as shared contexts? i foresee these being painful sharing all this setup stuff. if someone wants to test something slightly different it just won't work as it's fully coupled. I'd rather opt for duplication here personally.

i see shared contexts more of a way to test for a very specific behaviour common to many things. are we even in that area?

kicferk1: I think we want a base shared setup for T&P tests - and to diverge from that setup if some individual tests require it. I think almost all T&P tests require some ECF participant in a school which is in a partnership with a lead provider, and writing it down everywhere is a pain. This is simply a place where we haven't used the shared setup yet (it is a pretty new thing).

ethanmills (approved):

asmega:
Other work has affected this PR so I'll open a new one up in a similar vein when I'm done

PR #865, opened 2021-08-16
Prevent induction coordinator from adding themselves as a mentor at multiple schools

ethanmills:
Change the check to ensure the induction coordinator is not a mentor at
all before allowing them to add themselves at a school

### DIFF
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -65,7 +65,7 @@ module Schools
     end
 
     def can_add_self?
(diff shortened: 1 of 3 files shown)


kicferk1 (commented):
Does this want to be unit tested?

Also, are we sure this is the behaviour we want? Do we ever communicate to ICs that they can nly be a mentor in one school?

ethanmills:
> Also, are we sure this is the behaviour we want? Do we ever communicate to ICs that they can nly be a mentor in one school?

Everyone can only be a mentor in one school

kicferk1 (approved):

PR #866, opened 2021-08-16
Simply feature flag test to prevent random failures

ethanmills:
I don't think we really lose anything by simplifying these, do we?


kicferk1 (approved):
I think we do indeed only activate them for schools and users. I guess the original purpose was to show that they can be activated for literally any object in a database, and the complexity is that some factories don't make sense in isolation. I am happy with this chanage.

jamesrose (approved):

PR #867, opened 2021-08-16
CPDTP-423 Generate unique TRNS for testing and sandbox purposes.

blackrat:
### Context

Existing TRN generation code just did a sample without a comparison which meant that scaling it up was more likely to result in collisions.
This class generates unique and unassigned TRNs without having too much impact on the database.

Create a generator which caches all possibly TRNs and removes the ones already assigned.
Add an enumerator which will return the next unassigned TRN or raise an error if all of the available ones are exhausted.

### Guidance to review

All initial population of the huge array happens during start-up.
A shuffled version of this and the read of the database to determine which ones are actually is use happens the first time that #next is called. 
Once this happens, the available is treated as a stack and the next available value is popped off it.
This minimises the chances of a collision when generating 100_000 values and keeps the time spent handling exceptions low.

### Testing

Specific tests have been written to generate a large number of unique TRNs and benchmark the time taken to create them.

### Review Checks

### DIFF
--- a/lib/tasks/trn_generator.rb
+++ b/lib/tasks/trn_generator.rb
@@ -3,12 +3,12 @@
 class TRNGenerator
   class << self
     def next
-      taken.push(available.pop || (raise "TRN available list exhausted"))[-1]
+      available.pop || (raise "TRN available list exhausted")
     end
 
   private
 
-    ALL_TRNS = ("001111".."9999999").to_a.freeze unless defined?(ALL_TRNS)
+    ALL_TRNS = ("0011111".."9999999").to_a.freeze unless defined?(ALL_TRNS)
 
     def available
       @available ||= (ALL_TRNS - taken).shuffle
--- a/lib/tasks/trn_generator.rb
+++ b/lib/tasks/trn_generator.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class TRNGenerator
+  class << self
+    def next
+      taken.push(available.pop || (raise "TRN available list exhausted"))[-1]
+    end
+
+  private
+
+    ALL_TRNS = ("001111".."9999999").to_a.freeze unless defined?(ALL_TRNS)
+
+    def available
+      @available ||= (ALL_TRNS - taken).shuffle
+    end
+
+    def taken
+      @taken ||= TeacherProfile.where.not(trn: nil).pluck(:trn)
+    end
+  end
+end
--- a/lib/tasks/trn_generator.rb
+++ b/lib/tasks/trn_generator.rb
@@ -1,21 +1,25 @@
 # frozen_string_literal: true
 
+ALL_TRNS = (1111..9_999_999).freeze unless defined?(ALL_TRNS)
+
 class TRNGenerator
   class << self
     def next
-      available.pop || (raise "TRN available list exhausted")
+      sprintf("%07d", taken.push(available.pop || (raise "TRN available list exhausted"))[-1])
     end
 
   private
 
-    ALL_TRNS = ("0011111".."9999999").to_a.freeze unless defined?(ALL_TRNS)
-
     def available
-      @available ||= (ALL_TRNS - taken).shuffle
+      @available.presence || reseed
+    end
+
+    def reseed
+      @available = (10_000.times.map { rand(ALL_TRNS) }.uniq - taken)
     end
 
     def taken
-      @taken ||= TeacherProfile.where.not(trn: nil).pluck(:trn)
+      @taken ||= TeacherProfile.where.not(trn: nil).pluck(:trn).map(&:to_i)
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/trn_generator.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class TRNGenerator
+  class << self
+    def next
+      available.pop || (raise "TRN available list exhausted")
+    end
+
+  private
+
+    ALL_TRNS = ("0011111".."9999999").to_a.freeze unless defined?(ALL_TRNS)
```
kicferk1, line 11: Soooo... Will this just load 10 milion strings into memory when this class is loaded?

Can we test how it works on a review app?

lib/tasks/trn_generator.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def available
+      @available ||= (ALL_TRNS - taken).shuffle
+    end
```
kicferk1, line 15: I wonder how much we need the shuffle. What's the risk of generating TRNs sequentially? (which would presumably allow us not to keep track of an array of strings couple of millions long)

lib/tasks/trn_generator.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def taken
+      @taken ||= TeacherProfile.where.not(trn: nil).pluck(:trn)
```
asmega, line 18: if you memoize this it means the query result is cached. this means when you call `next`, `available` is not updated therefore you can generate duplicate trns. is that desired?

spec/lib/tasks/trn_generator_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    expect(::TeacherProfile.pluck(:trn)).not_to include(TRNGenerator.next)
+  end
+
+  it "generates 100_000 unallocated TRNs in under a second" do
+    benchmark = Benchmark.measure do
+      100_000.times do
+        TRNGenerator.next
+      end
+    end
+    expect(benchmark.total).to be < 1
```
asmega, line 17: because of the memoization this is probably why it is so performant. I expect if you wanted actually different TRNs the operation is more expensive for 100k outputs

lib/tasks/trn_generator.rb
blackrat: Yes. That's the point of it having so many. I wanted to make sure that the review app could cope with them all if we for some reason wanted to generate all possibly valid 10M records. I think that it'll be unlikely that we need more than 200k at any time, so the next PR will restrict this for the rake task.

lib/tasks/trn_generator.rb
blackrat: Check again. You're missing the `available.pop` which is used by next

spec/lib/tasks/trn_generator_spec.rb
blackrat: Again. You're missing the `available.pop`. This is performant because it's dealing with an internal array and just treating it like a stack.

lib/tasks/trn_generator.rb
blackrat: We only ever look at the head of the array, so internally it's a single pointer update. Not hard to keep track of what essentially is a stack.

asmega:
Wouldn't a far better strategy be something like:
```
available_trns = ALL_TRNS - existing trns
available_trns.shuffle!

def next
  available_trns.pop
end
```

asmega:
> Wouldn't a far better strategy be something like:
> 
> ```
> available_trns = ALL_TRNS - existing_trns
> available_trns.shuffle!
> 
> def next
>   available_trns.pop
> end
> ```

Ok, my mistake looks like we are doing this!

kicferk1 (approved):
So long as the review app can load this into memory, we are probably fine on any other environment.

ethanmills (approved):

ethanmills (changes requested):
This appears to OOM 
image

lib/tasks/trn_generator.rb
ethanmills, line 11: This is at least 700MB, right?

kicferk1: 10 milion TRNs, times 7, times however many bytes per character, times however inefficient Ruby is with it. I would hope closer to 70MB than 700, but it clearly makes our app unhappy 😢

ethanmills:
Feel free to dismiss my request for changes when you've got something else

lib/tasks/trn_generator.rb
ethanmills: Indeed. Looks like we were running at somewhere between 60% and 80% utilization

blackrat: I've changed the strategy to keep these as integers and generate a unique map. That should minimise the hit on it.

kicferk1 (approved):
This seems to behave well in a review app. As in, it doesn't crash it and gives me some random TRNs

ethanmills (approved):

PR #870, opened 2021-08-17
Cpdtp 423 create tool to generate valid data for lead providers to test against

blackrat:
### Context

Creating test data can be a minefield if you don't have the correct tools in place. This rake task allows us to create batches of data for lead providers with representative samples of how the expected schools and participants would look.

- 11% sparsity_uplift
- 22% pupil_premium_uplift
- 14% withdrawn record sets for participants
- Randomised mentors and early_career_teachers with and without their equivalents (i.e. some mentors without ECTs and some ECTs without mentors)
- Even split of valid and missing TRN data for participants.

- Move methods for generation of data into callable modules and classes and out of the rake task
- Cascade creation of records where required to fulfil the numbers of schools and participants per school for each lead provider.

### Guidance to review

Run the task and watch it generate up to 60k records for the 6 named lead_providers.

### Review Checks

### DIFF
--- a/lib/tasks/sandbox_data.rake
+++ b/lib/tasks/sandbox_data.rake
@@ -1,83 +1,17 @@
 # frozen_string_literal: true
 
-namespace :lead_provider do
-  desc "create seed mentors for API testing"
-  task seed_mentors: :environment do
+require "tasks/valid_test_data_generator"
+
+namespace :lead_providers do
+  desc "create seed schools and participants for API testing"
+  task seed_schools_and_participants: :environment do
     logger = Logger.new($stdout)
     logger.formatter = proc do |_severity, _datetime, _progname, msg|
       "#{msg}\n"
     end
 
-    providers_names = ["Capita", "Teach First", "UCL Institute of Education", "Best Practice Network", "Ambition Institute", "Education Development Trust"]
-    cohort_2021 = Cohort.find_by!(start_year: "2021")
-
-    providers_names.each_with_index do |provider_name, index|
-      logger.info("Generating seed mentors for '#{provider_name}'...")
-      lead_provider = LeadProvider.find_by!(name: provider_name)
-      school = create_school_and_associations(lead_provider, cohort_2021, index)
-      generate_mentors(lead_provider, school, cohort_2021, logger)
+    ["Capita", "Teach First", "UCL Institute of Education", "Best Practice Network", "Ambition Institute", "Education Development Trust"].each do |provider|
+      ValidTestDataGenerator::LeadProviderPopulator.call(name: provider, total_schools: 100, participants_per_school: 100)
     end
   end
-end
-
-def generate_mentors(lead_provider, school, cohort, logger)
-  existing_mentor_count = lead_provider.ecf_participant_profiles.mentors.count
-  existing_ect_count = lead_provider.ecf_participant_profiles.ects.count
-  school_cohort = SchoolCohort.find_or_create_by!(school: school, cohort: cohort)
-
-  10.times do
-    mentor = create_teacher
-    mentor_profile = ParticipantProfile::Mentor.create!(teacher_profile: mentor, school_cohort: school_cohort, schedule: Finance::Schedule.default)
-
-    ect_count = rand(0..3)
-    ect_count.times do
-      ect = create_teacher
-      ParticipantProfile::ECT.create!(teacher_profile: ect, school_cohort: school_cohort, mentor_profile: mentor_profile, schedule: Finance::Schedule.default)
-    end
-    logger.info(" Mentor with user_id #{mentor.id} generated with #{ect_count} ECTs")
-  end
-  2.times do
-    mentor = create_teacher
-    mentor_profile = ParticipantProfile::Mentor.create!(teacher_profile: mentor, school_cohort: school_cohort, status: "withdrawn", schedule: Finance::Schedule.default)
-
-    ect = create_teacher
-    ParticipantProfile::ECT.create!(teacher_profile: ect, school_cohort: school_cohort, mentor_profile: mentor_profile, status: "withdrawn", schedule: Finance::Schedule.default)
-  end
-  new_mentor_count = lead_provider.ecf_participant_profiles.mentors.count
-  logger.info(" Before: #{existing_mentor_count} mentors, after: #{new_mentor_count}")
-  new_ect_count = lead_provider.ecf_participant_profiles.ects.count
-  logger.info(" Before: #{existing_ect_count} ECTs, after: #{new_ect_count}")
-end
-
-def create_school_and_associations(lead_provider, cohort, index)
-  school = School.find_or_create_by!(
-    urn: sprintf("%06d", (10_000 + index)),
-  ) do |s|
-    s.name = Faker::Company.name
-    s.address_line1 = Faker::Address.street_address
-    s.postcode = Faker::Address.postcode
-  end
-
-  Partnership.find_or_create_by!(
-    school: school,
-    lead_provider: lead_provider,
-    cohort: cohort,
-  ) do |partnership|
-    partnership.delivery_partner = DeliveryPartner.create!(name: Faker::Company.name)
-  end
-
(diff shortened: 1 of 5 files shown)


lib/tasks/school_urn_generator.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+ALL_URNS = (1..999_999).freeze unless defined?(ALL_URNS)
+
+class SchoolURNGenerator
+  class << self
+    def next
+      sprintf("%06d", taken.push(available.pop || (raise "URN available list exhausted"))[-1])
```
kicferk1, line 8: There is like 4 or 5 things in that line. Please do it across multiple lines. In particular I don't know what `[-1]` is doing here. I think it is only facilitating sticking all of it in one line and avoiding a local variable, which I reckon is not a good enough reason to justify itself.

lib/tasks/school_urn_generator.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  private
+
+    def available
+      @available.presence || reseed
+    end
+
+    def reseed
+      @available = (1_000.times.map { rand(ALL_URNS) }.uniq - taken)
+    end
```
kicferk1, line 19: What is this doing? Why do we need 1000 copies of `rand(ALL_URNS)`?

ethanmills (approved):
This is a big improvement, looks good. As mentioned, it would be nice to have something other than a strict one to one between ECTs and Mentors

lib/tasks/school_urn_generator.rb
kicferk1: Is this saying that available urns are 1000 random urns, but we remove duplicates and taken? Why do we need this? So the memory doesn't go haywire?

If so, why would we not just use `rand(ALL_URNS)` and re-roll when we run into a taken object? What's the improvement here, other than we do it 1000 times?

lib/tasks/school_urn_generator.rb
kicferk1: Also, all comments in this file relate to TRN validation too. I think I approved an old, and pretty different version of it, sorry I didn't take another look.

blackrat:
> This is a big improvement, looks good. As mentioned, it would be nice to have something other than a strict one to one between ECTs and Mentors

Yes. I'll make that change before merging.

lib/tasks/valid_test_data_generator.rb
```diff
@@ -0,0 +1,92 @@
...
+      @lead_provider = ::LeadProvider.find_or_create_by!(name: name)
+    end
+
+    def generate_new_schools(count:)
+      count.times { create_fip_school_with_cohort(urn: SchoolURNGenerator.next) }
+    end
+
+    def find_or_create_participants(school:, number_of_participants:)
+      generate_new_participants(school: school, count: number_of_participants - school.ecf_participants.count) if school.ecf_participants.count < number_of_participants
+      school.ecf_participants.order("created_at desc").limit(number_of_participants).to_a
```
kicferk1, line 35: What is this last line doing? Nothing that we use from what I can see.

lib/tasks/sandbox_data.rake
```diff
@@ -1,83 +1,17 @@
...
-    providers_names = ["Capita", "Teach First", "UCL Institute of Education", "Best Practice Network", "Ambition Institute", "Education Development Trust"]
-    cohort_2021 = Cohort.find_by!(start_year: "2021")
-
-    providers_names.each_with_index do |provider_name, index|
-      logger.info("Generating seed mentors for '#{provider_name}'...")
-      lead_provider = LeadProvider.find_by!(name: provider_name)
-      school = create_school_and_associations(lead_provider, cohort_2021, index)
-      generate_mentors(lead_provider, school, cohort_2021, logger)
+    ["Capita", "Teach First", "UCL Institute of Education", "Best Practice Network", "Ambition Institute", "Education Development Trust"].each do |provider|
+      ValidTestDataGenerator::LeadProviderPopulator.call(name: provider, total_schools: 100, participants_per_school: 100)
```
kicferk1, line 14: So it's like 60k participants in total, right?

lib/tasks/sandbox_data.rake
```diff
@@ -1,83 +1,17 @@
 # frozen_string_literal: true
 
-namespace :lead_provider do
-  desc "create seed mentors for API testing"
-  task seed_mentors: :environment do
+require "tasks/valid_test_data_generator"
```
kicferk1, line 3: I wonder if we should put it in `services` in some namespace. The fact that tasks are the only ones using it is true at the moment, but in principle I could see other things using it - say our seeds for review apps.

lib/tasks/school_urn_generator.rb
kicferk1: Ah, I guess if we fail to find a new element in 1000 attempts the code above will fail raise the exception.

lib/tasks/sandbox_data.rake
blackrat: Yes. 10k for each provider.

lib/tasks/valid_test_data_generator.rb
blackrat: Returning an array of the chosen schools. I was mapping them before, but you're correct in that it's now a redundant line.

lib/tasks/school_urn_generator.rb
blackrat: Yes. That's correct. It's a much lower memory version that create 1000 random URNs, removes duplicate and taken and that becomes the available set for the next set of runs. If it's empty, it raises when next is called.

blackrat: It's a memory optimisation and doing like this means that we don't get any collisions and don't need to cater for them in this.

lib/tasks/sandbox_data.rake
blackrat: Maybe later. I think that the TRN and URN generators could be used, for example, by factories.

lib/tasks/school_urn_generator.rb
kicferk1: Alright, fair enough

kicferk1 (commented):
Linter is sad.

Have you tested it in a review app? Just to be on the safe side?

Might be worth checking with Rhys about whether we need any NPQ participants - I think the setup for them would need to be a bit different, and could be its own PR. And I don't know if we need it for this week.

I just want to avoid any surprises due to communication errors.

blackrat:
> Linter is sad.
> 
Yeah. Had to disable auto rubocop and it's greedy about it's styles.
> 
> Have you tested it in a review app? Just to be on the safe side?
Not yet, but will.
> 
> Might be worth checking with Rhys about whether we need any NPQ participants - I think the setup for them would need to be a bit different, and could be its own PR. And I don't know if we need it for this week.

We don't need that one just yet. This one has a big "participants for schools focus", so NPQ is out of scope at present for this and the setup for them will indeed be very different.

lib/tasks/valid_test_data_generator.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    def create_participant(school_cohort:, profile_type: :ect, mentor_profile: nil, status: "active", sparsity_uplift: false, pupil_premium_uplift: false)
+      name = Faker::Name.name
+      user = User.create!(full_name: name, email: Faker::Internet.email(name: name))
+      teacher_profile = TeacherProfile.create!(user: user, trn: random_or_nil_trn)
+      if profile_type == :ect
+        ParticipantProfile::ECT.create!(teacher_profile: teacher_profile, school_cohort: school_cohort, mentor_profile: mentor_profile, status: status, sparsity_uplift: sparsity_uplift, pupil_premium_uplift: pupil_premium_uplift)
+      else
+        ParticipantProfile::Mentor.create!(teacher_profile: teacher_profile, school_cohort: school_cohort, status: status, sparsity_uplift: sparsity_uplift, pupil_premium_uplift: pupil_premium_uplift)
+      end
```
kicferk1, line 64: We need schedule here, or the providers will have a bad time.

kicferk1 (approved):

PR #873, opened 2021-08-17
CPDTP-294: Enforce schedules on participant profiles

kicferk1:
### Context
Participant profiles need schedules to validate that their declarations are sent at the right time.

Make all participant profiles require schedules. I will double check that this is the case in prod and sandbox before merging, but it should be.


kicferk1:
https://github.com/DFE-Digital/early-careers-framework/pull/877 does it better

PR #877, opened 2021-08-18
CPDTP-294: Enforce schedules on participant profiles

kicferk1:
### Context
Participant profiles need schedules to validate that their declarations are sent at the right time.

Make all participant profiles require schedules. I will double check that this is the case in prod and sandbox before merging, but it should be.

I needed to wrap setting null in `safety_assured` because our postgres is 11, rather than 12+, and the recommended strong migrations way only works for 12+.

### DIFF
--- a/db/migrate/20210817084817_validate_remove_null_from_participant_profile_schedules.rb
+++ b/db/migrate/20210817084817_validate_remove_null_from_participant_profile_schedules.rb
@@ -3,7 +3,7 @@
 class ValidateRemoveNullFromParticipantProfileSchedules < ActiveRecord::Migration[6.1]
   def change
     validate_check_constraint :participant_profiles, name: "participant_profiles_schedule_id_null"
-    change_column_null :participant_profiles, :schedule_id, false
+    safety_assured { change_column_null :participant_profiles, :schedule_id, false }
     remove_check_constraint :participant_profiles, name: "participant_profiles_schedule_id_null"
   end
 end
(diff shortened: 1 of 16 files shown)


db/migrate/20210817084817_validate_remove_null_from_participant_profile_schedules.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ValidateRemoveNullFromParticipantProfileSchedules < ActiveRecord::Migration[6.1]
+  def change
+    validate_check_constraint :participant_profiles, name: "participant_profiles_schedule_id_null"
```
blackrat, line 5: I thought the whole point of strong_migrations was to have you do the former check_constraint, then run a required data change (in a console) and then a subsequent migration to run the validation and removal. I'd expect this to fail, force you to run a session to fix the data and then allow a redeploy and re-run.

jamesrose: https://github.com/ankane/strong_migrations#setting-not-null-on-an-existing-column

kicferk1: I did just what the docs said. It works locally. Presumably the comment in the docs, 
```
    # in Postgres 12+, you can then safely set NOT NULL on the column
```
is what is going wrong - our postgres is 11...

kicferk1:
Aaaand the reason was the difference in postgres versions.

I will update description of this PR and close the old one.

ethanmills (approved):

kicferk1:
I double checked there are 0 schedule-less participant profiles in prod and sandbox. Let's roll.

PR #881, opened 2021-08-18
Update pupil premium dataset

jamesrose:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDRP-534

We imported pupil premium data in Feb 2021 that was not the 21/22 dataset. This PR updates the pupil premium dataset to the [21/22 set](https://www.gov.uk/government/publications/pupil-premium-allocations-and-conditions-of-grant-2021-to-2022).

Once this PR is merged and deployed I will:

1. Re-run the `PupilPremiumImporter` so that the new dataset is imported.
2. Run the temporary `pupil_premium:update_cached_pupil_premium_uplift` task so that the cached pupil premium uplift values on participant profiles are updated.
3. Open a PR to remove the Rake task.


kicferk1 (approved):
Do we need to communicate this to LPs? They were (presumably) contacting schools and making partnerships based on the old pupil premium data.

Also, do we need to do the same for sparcity?

ethanmills (approved):
Glad it turned out we could use the up to date data!

jamesrose:
I was wrong about this. We can't use the 21/22 dataset this cohort. The next one will be 1:1 with the dataset:cohort year.

PR #882, opened 2021-08-18
CPDTP-421: Add a smoke test to participant declaration endpoints

kicferk1:
### Context

We want some smoke tests. As in, tests running against something that is not a test environment. For starters I will put it against a review app.

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -10,26 +10,6 @@ jobs:
   deploy:
     runs-on: ubuntu-20.04
 
-    env:
-      RAILS_ENV: test
-      DB_USERNAME: postgres
-      DB_PASSWORD: ''
-      DB_HOSTNAME: 127.0.0.1
-      DB_PORT: 5432
-      PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
-      PERCY_PARALLEL_NONCE: ${{ needs.create-nonce.outputs.nonce }}
-      PERCY_PARALLEL_TOTAL: 2
-
-    services:
-      postgres:
-        image: postgres:11.6-alpine
-        env:
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: ''
-        ports:
-          - 5432:5432
-        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
-
     steps:
       - uses: softprops/turnstyle@v1
         name: Check workflow concurrency
@@ -134,27 +114,32 @@ jobs:
           msg: "Created review app at https://ecf-review-pr-${{ github.event.number }}.london.cloudapps.digital"
           check_for_duplicate_msg: true
 
-      # avoid complexities of dispatching a separate flow
-      - name: Set up Ruby
-        uses: ruby/setup-ruby@v1
+      - name: Trigger Smoke Test
+        uses: benc-uk/workflow-dispatch@v1.1
         with:
-          bundler-cache: true
+          workflow: Smoke Test
+          token: ${{ secrets.PERSONAL_TOKEN }}
+          inputs: '{"paas_environment": "${{ env.ENVIRONMENT }}", "sha": "${{ github.event.pull_request.head.sha }}"}'
+          ref: ${{ github.event.pull_request.head.ref }}
 
-      - name: Set up Node
-        uses: actions/setup-node@v2.4.0
+      - name: Wait for smoke test
+        id: wait_for_smoke_test
+        uses: fountainhead/action-wait-for-check@v1.0.0
         with:
-          node-version: '12.x'
+          token: ${{ secrets.PERSONAL_TOKEN }}
+          checkName: Smoke Test ${{ env.ENVIRONMENT }} ${{ github.event.pull_request.head.sha }}
+          ref: ${{ github.event.pull_request.head.ref }}
+          timeoutSeconds: 300
+          intervalSeconds: 15
 
-      - name: set environment (workflow dispatch)
-        if: ${{ github.event_name == 'workflow_dispatch' }}
-        run: |
-          echo "ENVIRONMENT=${{ github.event.inputs.paas_environment }}" >> $GITHUB_ENV
+      - name: Exit whole workflow if ${{ env.ENVIRONMENT }} smoke test was not successful
+        if: steps.wait_for_smoke_test.outputs.conclusion != 'success'
+        run: exit 1
 
-      - name: Set up test database
-        run: bin/rails db:create db:schema:load
-
-      - name: Run deployment smoke test
-        run: bundle exec rspec spec/smoke_tests/participant_declarations_spec.rb --tag smoke_test
-
-      - name: print environment
-        run: echo ${{ env.ENVIRONMENT}}
+      - name: comment on PR
+        uses: unsplash/comment-on-pr@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          msg: "Ran smoke test against the review app"
+          check_for_duplicate_msg: true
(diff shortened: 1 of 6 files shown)


ethanmills:
I know we're moving away, but would Cypress be a better fit here? No need for all the ruby machinery, designed to run against a different server

kicferk1:
> I know we're moving away, but would Cypress be a better fit here? No need for all the ruby machinery, designed to run against a different server

If I get annoyed enough with Ruby setup I will look into setting up cypress for it

.github/workflows/review.yml
```diff
@@ -112,3 +113,33 @@ jobs:
         with:
           msg: "Created review app at https://ecf-review-pr-${{ github.event.number }}.london.cloudapps.digital"
           check_for_duplicate_msg: true
+
+      - name: Trigger Smoke Test
+        uses: benc-uk/workflow-dispatch@v1.1
+        with:
+          workflow: Smoke Test
+          token: ${{ secrets.PERSONAL_TOKEN }}
```
asmega, line 121: sorry, what is this token used for? and can it be named more aptly?

.github/workflows/review.yml
```diff
@@ -112,3 +113,33 @@ jobs:
...
+      - name: Exit whole workflow if ${{ env.ENVIRONMENT }} smoke test was not successful
+        if: steps.wait_for_smoke_test.outputs.conclusion != 'success'
+        run: exit 1
+
+      - name: comment on PR
+        uses: unsplash/comment-on-pr@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          msg: "Ran smoke test against the review app"
```
jamesrose, line 144:
```suggestion
          msg: "Smoke tests passed against the review app."
```

.github/workflows/smoke_test.yml
```diff
@@ -37,10 +58,13 @@ jobs:
...
+      - name: Set up test database
+        run: bin/rails db:create db:schema:load
 
       - name: Run deployment smoke test
-        run:  bundle exec rspec spec/smoke_tests/participant_declarations_spec.rb --tag smoke_test
+        run: bundle exec rspec spec/smoke_tests/participant_declarations_spec.rb --tag smoke_test
 
       - name: print environment
-        run: echo ${{ env.PAAS_ENVIRONMENT}}
+        run: echo ${{ env.ENVIRONMENT}}
```
ethanmills, line 70: Why?

.github/workflows/smoke_test.yml
```diff
@@ -1,6 +1,8 @@
 name: Smoke Test
 
 on:
+  schedule:
+    - cron: '*/5 * * * *'
```
jamesrose, line 5: What's this scheduling?

.github/workflows/review.yml
```diff
@@ -112,3 +113,33 @@ jobs:
...
+        with:
+          token: ${{ secrets.PERSONAL_TOKEN }}
+          checkName: Smoke Test ${{ env.ENVIRONMENT }} ${{ github.event.pull_request.head.sha }}
+          ref: ${{ github.event.pull_request.head.ref }}
+          timeoutSeconds: 300
+          intervalSeconds: 15
+
+      - name: Exit whole workflow if ${{ env.ENVIRONMENT }} smoke test was not successful
+        if: steps.wait_for_smoke_test.outputs.conclusion != 'success'
+        run: exit 1
```
jamesrose, line 137: Do we get notified if it was not successful? How do we tell?

.github/workflows/smoke_test.yml
```diff
@@ -37,10 +58,13 @@ jobs:
...
         run: |
-          echo "PAAS_ENVIRONMENT=${{ github.event.inputs.paas_environment }}" >> $GITHUB_ENV
+          echo "ENVIRONMENT=${{ github.event.inputs.paas_environment }}" >> $GITHUB_ENV
+
+      - name: Set up test database
+        run: bin/rails db:create db:schema:load
 
       - name: Run deployment smoke test
-        run:  bundle exec rspec spec/smoke_tests/participant_declarations_spec.rb --tag smoke_test
+        run: bundle exec rspec spec/smoke_tests/participant_declarations_spec.rb --tag smoke_test
```
ethanmills, line 67:
```suggestion
        run: bundle exec rspec spec/smoke_tests/**/*_spec.rb --tag smoke_test
```

ethanmills (approved):
Looks good to me in general, let's see if it works

.github/workflows/smoke_test.yml
```diff
@@ -18,11 +20,30 @@ jobs:
 
     runs-on: ubuntu-20.04
 
+    # Our rspec is deeply coupled with a database. I had a go at decoupling it for smoke tests, and gave up.
```
asmega, line 23: My guess is because `.rspec` loads everything
```
--require rails_helper
--require spec_helper
--require payment_calculator
```
it should really be only loading `spec_helper` then you'd probably be free to do what you want

.github/workflows/smoke_test.yml
ethanmills, line 5: Do we set any default parameters for this?

asmega, line 5: is every 5 minutes overkill?

kicferk1: Hmm, maybe I should remove this all-together for time being.

.github/workflows/review.yml
kicferk1: Well, we get told the review app workflow failed (I get emails about it for my PRs), and it should have a little red cross on github. Do you think we should get anything more from review smoke tests?

.github/workflows/smoke_test.yml
kicferk1: Cause `PAAS_ENVIRONMENT` is what Teaching Vacancies use, and `ENVIRONMENT` is what we use in our actions.

.github/workflows/review.yml
kicferk1: To dispatch an action you cannot use a default github token provided in the environment. You need a 'personal' token. 

I can change the secret name to something more apt if you want? 

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token

.github/workflows/smoke_test.yml
kicferk1: Copy paste artifact, I think for time being we don't need it.

.github/workflows/smoke_test.yml
kicferk1: This might be one reason, but there is more. We clean database after each run for one
```
RSpec.configure do |rspec|
  rspec.before(:suite) { DatabaseCleaner.clean_with :truncation }
end
```
We check migration status
```
# Checks for pending migrations and applies them before tests are run.
# If you are not using ActiveRecord, you can remove these lines.
begin
  ActiveRecord::Migration.maintain_test_schema!
rescue ActiveRecord::PendingMigrationError => e
  puts e.to_s.strip
  exit 1
end
```
Aaand possibly do more. 

I think trying to cut all of them out of our rspec is, well, much more effort than giving the flow a dummy database, as much as it hurts my heart :(

.github/workflows/review.yml
jamesrose: No, that makes sense. I wasn't sure if it would fail the pipeline.

jamesrose (approved):
Nice work, thanks for this! It's great to have something that we can build on over time.

.github/workflows/smoke_test.yml
ethanmills: I meant why do we echo it out here? Especially after the test step - that means we only print it if the test passes

kicferk1: Moved it earlier

PR #891, opened 2021-08-19
Only set TRN on teacher profile if NPQ validated

asmega:
### Context

- My understanding is the `TeacherProfile` should be a source of truth for a TRN
- At the moment this is set when an NPQ comes in which may contain an unvalidated TRN

- Ensure only set `TeacherProfile#trn` if the trn on the NPQ has been validated

### Guidance to review

- My assumptions for TRN on `TeacherProfile` are correct?

### Testing

- See testing on review app

### Review Checks

- Through api calls
- Create an NPQ registration with an unvalidated TRN
- TRN should not be set on `TeacherProfile`
- Create an NPQ registration with a validated TRN
- TRN should be set on the `TeacherProfile`

### DIFF
--- a/app/services/npq/create_or_update_profile.rb
+++ b/app/services/npq/create_or_update_profile.rb
@@ -9,7 +9,10 @@ module NPQ
     end
 
     def call
-      teacher_profile.trn = npq_validation_data.teacher_reference_number
+      if npq_validation_data.teacher_reference_number_verified?
+        teacher_profile.trn = npq_validation_data.teacher_reference_number
+      end
+
       teacher_profile.school = school
       teacher_profile.save!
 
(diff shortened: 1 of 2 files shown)


kicferk1:
Sooo, how does `teacher_reference_number_verified` get set to true? 

And if it is false, is there a process for validation?

asmega:
> Sooo, how does `teacher_reference_number_verified` get set to true?

2 ways. It can come in as `true` when the NPQ registration comes in. It comes in as true when NPQ auto validated the TRN with the api endpoint `Api::V1::ParticipantValidationController`

Secondly through manual validation. This is me doing processing CSVs by hand. It previously used this process mentioned here https://github.com/DFE-Digital/early-careers-framework/pull/742 however with the introduction of `NPQ::CreateOrUpdateProfile` it will need to be slightly tweaked so it includes this new class/service

> And if it is false, is there a process for validation?
 
Yes, NPQ has manual process involving CSVs and support with direct access to TRA/DQT data and sometimes emailing users

kicferk1 (approved):

PR #893, opened 2021-08-19
Add NPQ course to ParticipantProfile

asmega:
### Context

- A `TeacherProfile` can have many `ParticipantProfile::NPQ`
- At the moment these are mapped one-to-one for each NPQ application that comes in
- However we only want an `ParticipantProfile::NPQ` per NPQ course

- This adds NPQ course to `participant_profiles` table
- This will enable the ability for `ParticipantProfile::NPQ` to be scoped on courses in a later change
- For new applications populate `ParticipantProfile::NPQ` with the NPQ course
- Once deployed I will retro back fill `ParticipantProfile::NPQ#npq_course` based on the attached application

### Guidance to review

- `safety_assured` was used in the migration as since it is new column with no data i don't see the duration of the operation being particularly very long

### Testing

- See how to test in review app

### Review Checks

- Via api calls
- Create a new NPQ application
- Should now see the course being populated for the new participant profile

### DIFF
--- a/app/models/participant_profile/npq.rb
+++ b/app/models/participant_profile/npq.rb
@@ -3,6 +3,7 @@
 class ParticipantProfile::NPQ < ParticipantProfile
   self.ignored_columns = %i[mentor_profile_id school_cohort_id]
   belongs_to :school, optional: true
+  belongs_to :npq_course, optional: true
 
   has_one :validation_data, class_name: "NPQValidationData", foreign_key: :id, dependent: :destroy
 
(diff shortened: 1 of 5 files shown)


app/models/participant_profile/npq.rb
```diff
@@ -3,6 +3,7 @@
 class ParticipantProfile::NPQ < ParticipantProfile
   self.ignored_columns = %i[mentor_profile_id school_cohort_id]
   belongs_to :school, optional: true
+  belongs_to :npq_course, optional: true
```
kicferk1, line 6: Will we eventually make it non-optional here?

kicferk1 (approved):

app/models/participant_profile/npq.rb
asmega: yea i guess we should change that when we get there

PR #894, opened 2021-08-19
CPDRP-842: validation status tags

ethanmills:
### Context
CPDRP-842
Now that we've started validations, we need to show something other than pending against all these participants

- Change the school participants routes to use profile ID
- Move some of the associations from ParticipantProfile to ParticipantProfile::ECF
- Add the tags

### Review Checks


app/components/participant_status_tag_component.rb
```diff
@@ -0,0 +1,33 @@
...
+    @profile = profile
+    @admin = admin
+  end
+
+  def call
+    if profile.npq?
+      render Admin::Participants::NPQValidationStatusTag.new(profile: profile)
+    else
+      govuk_tag tag_attributes
+    end
```
kicferk1, line 14: I sense a disturbing lack of symetry. Do we want a `Admin::Participants::ECFValidationStatusTag`?

app/components/participant_status_tag_component.rb
```diff
@@ -0,0 +1,33 @@
...
+    return { text: "Manual checks needed", colour: "turquoise" } if admin && manual_check_needed
+    return { text: "DfE checking eligibility", colour: "blue" } if profile.ecf_participant_validation_data.present?
+
+    { text: "DfE requested details from participant", colour: "yellow" }
+  end
+
+  def manual_check_needed
+    profile.ecf_participant_eligibility&.manual_check_status? ||
+      (profile.ecf_participant_validation_data.present? && profile.ecf_participant_eligibility.nil?)
+  end
```
kicferk1, line 32: Does this want to live on the profile?

app/controllers/schools/participants_controller.rb
```diff
@@ -96,15 +97,15 @@ def check_feature_flag
   end
 
   def set_participant
-    @participant = User.find(params[:participant_id] || params[:id])
-    authorize @participant, policy_class: ParticipantPolicy
+    @profile = ParticipantProfile.find(params[:participant_id] || params[:id])
```
kicferk1, line 100: Will that work? what is `participant_id`? I thought it would be the id of the user?

spec/requests/schools/participants_request_spec.rb
```diff
@@ -67,13 +69,13 @@
 
   describe "GET /schools/:school_id/cohorts/:start_year/participants/:id" do
     it "renders participant template" do
-      get "/schools/#{school.slug}/cohorts/#{cohort.start_year}/participants/#{ect_user.id}"
+      get "/schools/#{school.slug}/cohorts/#{cohort.start_year}/participants/#{ect_profile.id}"
```
kicferk1, line 72: I guess we can hope that people don't bookmark those urls, and it's probably a fair bet.

app/components/participant_status_tag_component.rb
ethanmills: No, because this isn't just for admins

ethanmills: The asymmetry here is that there is no SIT equivalent for NPQ

kicferk1: I meant more that NPQ get a component to cover their tag attributes, but ECF's tag attributes are handled in the general component

ethanmills: I can extract if you'd prefer. I was mostly leaving the NPQ stuff alone because I expect it will need to change soon anyway

app/components/participant_status_tag_component.rb
ethanmills: Hmm, maybe. Most of this sort of things lives on the eligibility record, but this obviously can't. Feels a little odd having it on the profile, but probably less odd than exposing it here

app/controllers/schools/participants_controller.rb
ethanmills: I changed it to be the ID of a profile

app/components/participant_status_tag_component.rb
kicferk1: Hmm, alright. I would hope there will be more in common than different between them eventually.

ethanmills: They should end up pretty similar, but not identical

ethanmills:
This adds some N+1 queries, but it's fairly low traffic and low numbers. I don't foresee much impact. Anyone have a different view?

spec/requests/schools/participants_request_spec.rb
ethanmills: Yeah I'm not too worried

kicferk1 (approved):

PR #895, opened 2021-08-19
Provide registration detail in ECF user api endpoint

ethanmills:
### Review Checks


kicferk1 (approved):
LGTM

FriedSock (approved):
LGTM, but can we hold off merging this until we have made some invite logic changes in E&L? So we can avoid a manual task of sending invites to anyone who registers before we deploy that.

PR #897, opened 2021-08-20
CPDTP-340 Give providers feedback about declaration pay-ability

plisovin:
### Context

Additional action (index) on participant-declaration endpoint to return list of declarations together with the eligibility for the payment status

GET request with pagination i.e. GET ```/api/v1/participant-declarations?page[per_page]=200&page[page]=1```

### Review Checks

### DIFF
--- a/app/serializers/participant_declaration_serializer.rb
+++ b/app/serializers/participant_declaration_serializer.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require "jsonapi/serializer/instrumentation"
+
+class ParticipantDeclarationSerializer
+  include JSONAPI::Serializer
+  include JSONAPI::Serializer::Instrumentation
+
+  set_id :id
+  attributes :participant_id, :declaration_type, :declaration_date, :course_identifier
+
+  attributes :eligible_for_payment do |participant_declaration|
+    participant_declaration.profile_declaration.participant_profile.ecf_participant_eligibility&.status == "eligible"
+  end
+
+  attribute(:participant_id, &:user_id)
+end
--- a/app/serializers/participant_declaration_serializer.rb
+++ b/app/serializers/participant_declaration_serializer.rb
@@ -10,7 +10,7 @@ class ParticipantDeclarationSerializer
   attributes :participant_id, :declaration_type, :declaration_date, :course_identifier
 
   attributes :eligible_for_payment do |participant_declaration|
-    participant_declaration.profile_declaration.participant_profile.ecf_participant_eligibility&.status == "eligible"
+    participant_declaration.profile_declaration.participant_profile.ecf_participant_eligibility&.eligible_status? || false
   end
 
   attribute(:participant_id, &:user_id)
--- a/spec/requests/api/v1/participant_declarations_spec.rb
+++ b/spec/requests/api/v1/participant_declarations_spec.rb
@@ -146,22 +146,22 @@ RSpec.describe "participant-declarations endpoint spec", type: :request do
     context "when authorized" do
       context "when there is a non eligible declaration" do
         let(:expected_response) do
-          JSON.parse(<<-DATA)
-            {"data":
-              [
-                {
-                  "id":"#{participant_declaration.id}",
-                  "type":"participant_declaration",
-                  "attributes": {
-                    "participant_id": "#{ect_profile.user.id}",
-                    "declaration_type": "started",
-                    "declaration_date": "#{participant_declaration.declaration_date.rfc3339(3)}",
-                    "course_identifier": "ecf-induction",
-                    "eligible_for_payment": false
-                  }
-              }]
-            }
-          DATA
+          {
+            "data" =>
+            [
+              {
+                "id" => participant_declaration.id,
+                "type" => "participant_declaration",
+                "attributes" => {
+                  "participant_id" => ect_profile.user.id,
+                  "declaration_type" => "started",
+                  "declaration_date" => participant_declaration.declaration_date.rfc3339(3),
+                  "course_identifier" => "ecf-induction",
+                  "eligible_for_payment" => false,
+                },
+              },
+            ],
+          }
         end
 
         before do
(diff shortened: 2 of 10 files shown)


app/serializers/participant_declaration_serializer.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class ParticipantDeclarationSerializer
+  include JSONAPI::Serializer
+  include JSONAPI::Serializer::Instrumentation
+
+  set_id :id
+  attributes :participant_id, :declaration_type, :declaration_date, :course_identifier
+
+  attributes :eligible_for_payment do |participant_declaration|
+    participant_declaration.profile_declaration.participant_profile.ecf_participant_eligibility&.status == "eligible"
```
kicferk1, line 13: I think via power of enum this can be
`participant_declaration.profile_declaration.participant_profile.ecf_participant_eligibility&.eligible_status?`

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -125,4 +125,95 @@ def build_params(attributes)
...
+                  "attributes": {
+                    "participant_id": "#{ect_profile.user.id}",
+                    "declaration_type": "started",
+                    "declaration_date": "#{participant_declaration.declaration_date.rfc3339(3)}",
+                    "course_identifier": "ecf-induction",
+                    "eligible_for_payment": false
+                  }
+              }]
+            }
+          DATA
```
kicferk1, line 164: JSON.parse takes string and returns a ruby hash.

Why don't we skip that, and `<<-Data` steps and define our hash in pure ruby?

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -125,4 +125,95 @@ def build_params(attributes)
...
+          expect(response.status).to eq 200
+
+          expect(JSON.parse(response.body)).to eq(expected_response)
+        end
+      end
+
+      context "when there is an eligible declaration" do
+        before do
+          eligibility = ECFParticipantEligibility.create!(participant_profile_id: ect_profile.id)
+          eligibility.update!({ status: "eligible" })
```
kicferk1, line 183: `eligibility.eligible_status!` via magic of enum

kicferk1 (commented):
Can we have a swagger doc for this endpoint too?

plisovin:
> Can we have a swagger doc for this endpoint too?

yeah, good point, I will add

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -125,4 +125,95 @@ def build_params(attributes)
...
+
+      context "when there is an eligible declaration" do
+        before do
+          eligibility = ECFParticipantEligibility.create!(participant_profile_id: ect_profile.id)
+          eligibility.eligible_status!
+        end
+
+        let(:expected_response) do
+          JSON.parse(<<-DATA)
+            {"data":
```
kicferk1, line 188: Let's use a hash here too please :)

swagger/v1/component_schemas/MultipleParticipantDeclarationsResponse.yml
```diff
@@ -0,0 +1,26 @@
...
+properties:
+  data:
+    type: array
+    items:
+      $ref: "#/components/schemas/ParticipantDeclarations"
+    example:
+      - id: db3a7848-7308-4879-942a-c4a70ced400a
+        type: "participant-declaration"
+        attributes:
+          participant_id: db3a7848-7308-4879-942a-c4a70ced400a
```
kicferk1, line 14: I think we should make those id-s different - so someone reading the docs does not think they are the same thing.

swagger/v1/component_schemas/ParticipantDeclaration.yml
```diff
@@ -20,3 +20,4 @@ example:
       declaration_type: started
       declaration_date: 2021-05-31T02:21:32Z
       course_identifier: ecf-mentor
+      eligible_for_payment: true
```
kicferk1, line 23: I think this file specifies the thing that we expect to receive from the provider. 

But `eligible_for_payment` should be in a schema that we give back to the provider. 

If we don't have a file for that already, `ParticipantDeclarationRecordedResponse`? 

Maybe we want to rename that file to `ParticipantDeclarationResponse` so we can use it in both endpoints.

swagger/v1/api_spec.json
```diff
@@ -1391,7 +1483,8 @@
               "participant_id": "db3a7848-7308-4879-942a-c4a70ced400a",
               "declaration_type": "started",
               "declaration_date": "2021-05-31T02:21:32.000Z",
-              "course_identifier": "ecf-mentor"
+              "course_identifier": "ecf-mentor",
+              "eligible_for_payment": true
```
kicferk1, line 1487: That is the place that now gets `eligible_for_payment` where it does not belong, I think.

swagger/v1/component_schemas/ParticipantDeclaration.yml
plisovin: yeah, my bad, shouldn't be here

plisovin: ```ParticipantDeclarationRecordedResponse```  only contains an ID as this is what's been returned from the post request

kicferk1: Well, yeah, we can kill two birds with one stone and extend it. I think in the grand scheme of things we want a return after create to be in the same format as the return in get list.

But if you want you can make a new yml file for the endpoint you added in this PR, and we will merge them in a later PR.

plisovin: I've added ```ParticipantDeclarationResponse.yml``` which is referenced from ```MultipleParticipantDeclarationsResponse.yml``` as we can't reuse ```ParticipantDeclaration.yaml``` here due to new read-only flag returned by the GET participant-declaration API

spec/requests/api/v1/participant_declarations_spec.rb
plisovin: Done in 2589d47

kicferk1 (approved):

PR #898, opened 2021-08-20
Minor admin participant tweaks

asmega:
### Context

- Minor tweaks around admin participants area

- Eager load validation decisions to prevent n+1
- Stop treating data in table as numerical, this removes it from being right aligned

### Guidance to review

- None

### Testing

- Login as an admin
- View admin participants area
- Assuming there is data, there should be no n+1 queries in the logs
- All data in the table should be left aligned

### Review Checks

- Same as local testing


kicferk1 (approved):
Looks good :)

PR #899, opened 2021-08-20
Service to import NPQ manual validation CSV data

asmega:
### Context

- Based off https://github.com/DFE-Digital/early-careers-framework/pull/742
- The automatic TRN validation at the moment is not great due to missing data
- Therefore there is currently a manual validation process in place involving CSVs
- After the manual validation has been performed the data needs to be re-imported back into the service 

- Add CSV importer for NPQ applications

### Guidance to review

- Added new folder and therefore namespace for importers
- Differences from previous PR:
  - Renamed spec to correct location
  - Added `NPQ::CreateOrUpdateProfile` service

### Testing

- Start rails console
- Create test NPQValidationData records with blank or incorrect TRNs
- Create a new CSV with headers `application_ecf_id,validated_trn` 
- Populate CSV with test records and correct TRNs
- `svc = Importers::NPQManualValidation.new(path_to_csv: Rails.root.join("manual-validation.csv"))` 
- `svc.call`
- Records should now have correct TRN and marked as verified

### Review Checks

- Probably easier locally but same instructions from testing apply here

### DIFF
--- a/app/services/importers/npq_manual_validation.rb
+++ b/app/services/importers/npq_manual_validation.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+class Importers::NPQManualValidation
+  attr_reader :path_to_csv
+
+  def initialize(path_to_csv:)
+    @path_to_csv = path_to_csv
+  end
+
+  def call
+    check_headers
+
+    rows.each do |row|
+      data = NPQValidationData.find_by(id: row["application_ecf_id"])
+
+      puts "no NPQValidationData found for #{row['application_ecf_id']}" if data.nil?
+      next if data.nil?
+
+      puts "updating trn for NPQValidationData: #{row['application_ecf_id']} with trn: #{row['validated_trn']}"
+
+      ApplicationRecord.transaction do
+        data.update!(teacher_reference_number: row["validated_trn"], teacher_reference_number_verified: true)
+        NPQ::CreateOrUpdateProfile.new(npq_validation_data: data).call
+      end
+    end
+  end
+
+private
+
+  def check_headers
+    unless rows.headers == %w[application_ecf_id validated_trn]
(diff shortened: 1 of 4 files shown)


app/services/importers/npq_manual_validation.rb
```diff
@@ -0,0 +1,38 @@
...
+      data = NPQValidationData.find_by(id: row["application_ecf_id"])
+
+      puts "no NPQValidationData found for #{row['application_ecf_id']}" if data.nil?
+      next if data.nil?
+
+      puts "updating trn for NPQValidationData: #{row['application_ecf_id']} with trn: #{row['validated_trn']}"
+
+      data.update!(teacher_reference_number: row["validated_trn"], teacher_reference_number_verified: true)
+
+      NPQ::CreateOrUpdateProfile.new(npq_validation_data: data).call
```
ethanmills, line 23: I can see you getting errors here if there's already a teacher profile with that TRN

ethanmills: This should probably be wrapped in a transaction

asmega: Sure can wrap in transaction however don't follow where the error would be. Looking at the code multiple `TeacherProfile`s with the same TRN are permitted. `NPQ::CreateOrUpdateProfile` only creates or updates the TRN on the singleton `TeacherProfile`

ethanmills (approved):
Think a transaction would be good, looks good to me

app/services/importers/npq_manual_validation.rb
asmega: it's in a transaction now

kicferk1:
Looks like smoke tests failed, I think if you rebase on develop they will pass.

kicferk1 (approved):

PR #903, opened 2021-08-20
CDPTP-206 Change current status to suffix existing actions with _record for the current enumerated values

blackrat:
### Context

The current "active" and "withdrawn" status flags refer to the participant_profile record status as opposed to the participant and show if the record itself is available.

There is a new state coming in a follow-up PR that also sets a "withdrawn" state which is more about modelling the participant and their relationships with the course that they are or may no longer be following. i.e. withdrawn from a course as opposed to a withdrawn record.

In order to avoid some confusion and complexity around the differences and similarities of these logically distinct states, this first changes the actions and queries on this "active" and "withdrawn" status.

- modify the existing dynamically generated actions and queries to have an "_record" suffix, so that "withdrawn" becomes "withdrawn_record" and "withdrawn?" becomes "withdrawn_record?" and perform the same changes for "active" to "active_record" and "active_record?" (Confusion with ActiveRecord should be minor and easily overcome in comparison). 
- Update tests with these. Since this is an internal code method change only, API and database should not be effected.
- Change the trait "withdrawn" to "withdrawn_record" to reflect the intent.
- Clean up the tests so it's clear what they are testing vis a vis record withdrawal prior to putting in the lead_provider withdrawal.

### Guidance to review

This should only change the method calls

### Review Checks

### DIFF
--- a/app/controllers/admin/participants_controller.rb
+++ b/app/controllers/admin/participants_controller.rb
@@ -13,7 +13,7 @@ module Admin
       school_cohort_ids = SchoolCohort.ransack(school_name_or_school_urn_cont: params[:query]).result.pluck(:id)
       query = "%#{(params[:query] || '').downcase}%"
       @participant_profiles = policy_scope(ParticipantProfile).joins(:user)
-                                                              .active
+                                                              .active_record
                                                               .includes(:validation_decisions)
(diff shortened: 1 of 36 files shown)


kicferk1 (approved):
Right, so this enum is now referred to `active_record` and `withdrawn_record`. In the future PR that you mentioned, what do you plan to call the new enum to make it easily distinguishable?

Also... You got a ticket number for this PR? :D

PR #906, opened 2021-08-23
CPDTP-207: Only aggregate payable declarations

kicferk1:
### Context
We want to introduce the concept of "payability" to declaations, in parallel to "fundability" of participants.
A declaration is "payable" if the participant is "fundable", plus potentially some extra conditions.
https://dfedigital.atlassian.net/browse/CPDTP-207

And we do not want to count the non-payable declarations in our calculations.

1. Add 'payable' boolean and scope to participant declarations - at some point we might want to move to some other mechanism of storing it, or maybe computing it on the fly. But for time being this is the simplest thing I could think of.
2. Use that new scope in the aggregator.
3. Refactor the aggregator so that I can convince myself that I know what it is doing. In particular the line `aggregator.call({ cpd_lead_provider: cpd_lead_provider, event_type => aggregation_types[event_type][aggregation_type] }, event_type: event_type)` had me stumped. 
I think the old aggregator was supporting an initialisation with a bunch of instructions for what type of scope to use for which event. But I also don't think this was used anywhere - we seem to just define new aggregators when we want to pass another scope. Which I think is cheap enough and simpler.
4. Remove `early_career_teacher_profile` and `mentor_profile` factories - `participant_profile` with relevant traits should be enough, and right about now mainainance of those old factories got me annoyed. Specifically in tests for aggregator, where I wanted to show that we can substitute an NPQ setup for ECF setup (rather than defining a dummy setup). 

### Guidance to review
Going commit-by-commit should make things easier to follow. 

I think at some point soon it will be convenient to make a service named `UpdateDeclarationPayability` so we can easily controll the logic of that. And so we can back-fill that data for existing participants. Well, existing Sandbox participants anyway. 

This PR doesn't actually provide a way of getting a declaration to a payable state. Which I guess brings up a question to product (that I still need to ask): how, if at all, should we surface non-payable declarations to finance users?

### Testing
Add a unit test for aggregator showing it does not count non-payable declarations. 

You can't.


spec/services/participant_event_aggregator_service_spec.rb
```diff
@@ -2,43 +2,37 @@
 
 require "rails_helper"
 
-class DummyRecorder
```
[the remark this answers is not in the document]
kicferk1: The immediate impulse to remove it came from 'oh, I need to define a dummy that has `payable` method, which returns something that has `count` method'. At which point I was like 'well, might as well show an actual use case for it, which is NPQ'. 

And the result is... No need for dummy recorder.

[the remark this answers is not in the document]
blackrat: I'd have moved the payable to the scopes, as mentioned in the previous comment and left this one as is.
Changing the use case makes it more of an integration than a unit test.

kicferk1:
I think overriding a migration is causing some annoyances. Gonna close and re-open the PR to clean it up.

app/models/participant_profile.rb
```diff
@@ -43,6 +43,10 @@ def mentor?
     false
   end
 
+  def ecf?
+    ecf? || mentor?
```
blackrat, line 47: Circular reference, surely.

app/models/participant_profile/npq.rb
```diff
@@ -25,4 +25,8 @@ def rejected?
   def participant_type
     :npq
   end
+
+  def fundable?
+    false
```
blackrat, line 30: Do we need this if it's already in the base class?

app/services/calculation_orchestrator.rb
```diff
@@ -1,19 +1,18 @@
...
 
-require "participant_event_aggregator"
 require "payment_calculator/ecf/payment_calculation"
 
 class CalculationOrchestrator
   class << self
     def call(cpd_lead_provider:,
              contract:,
-             aggregator: ::ParticipantEventAggregator,
+             recorder: ParticipantDeclaration::ECF,
```
blackrat, line 9: Even though we've remove the proxy, I'd still have called it the aggregator, since it's acting like an aggregator for the orchestrator.

kicferk1: Hmm, alright, I went for `recorder` cause that is what it was called in the old aggregator :)

app/services/record_declarations/base.rb
```diff
@@ -89,10 +90,11 @@ def create_record!
           user: user,
           evidence_held: evidence_held,
         ).tap do |participant_declaration|
-          ProfileDeclaration.create!(
+          profile_declaration = ProfileDeclaration.create!(
             participant_declaration: participant_declaration,
             participant_profile: user_profile,
           )
+          profile_declaration.update!(payable: participant_declaration.currently_payable)
```
blackrat, line 97: Why as an external to the create! ?

app/models/participant_declaration.rb
```diff
@@ -1,30 +1,52 @@
 # frozen_string_literal: true
 
 class ParticipantDeclaration < ApplicationRecord
-  has_one :profile_declaration
+  has_many :profile_declarations
```
blackrat, line 4: Do we want to track fundable vs non-fundable as multiple profile_declarations? Is that easier than having a trackable fundability as an association within the profile declaration? I saw the profile declaration as being an extension to the participant_declaration so that we could add meta-data to it. Both mechanisms should work, but I think it merits further discussion.

kicferk1: Do you mean `has_one :payable_profile_declaration`? `has_many :payable_profile_declarations`? 

I think that we just don't know how this stuff will be used exactly. I am happy to say that we will care about when declaration became payable, and so to put payable flag on profile declaration - but anything beyond this seems much too speculative for my taste.

PR #907, opened 2021-08-23
CDPTP-206_Fix up the broken validation catch-all at the controller as a pre-step for implementing the withdrawn action

blackrat:
### Context

The participant-declaration controller had a global "catch-all" exception handler which was incorrectly always returning a message stating that the participant_id didn't match the course_identifier.
Working on the common interface for the withdrawn message highlighted this problem, when, for example, the underlying user didn't exist and a user.mentor? call (for example) would raise an exception that was then swallowed.

Refactor "participant-declarations" controller and validations as a step towards merging participant validation code into a block which will support both.
Move constant values into the class to avoid failures when the user creation fails for other validation types.

### Review Checks

### DIFF
--- a/app/services/record_declarations/npq.rb
+++ b/app/services/record_declarations/npq.rb
@@ -6,6 +6,7 @@ module RecordDeclarations
 
     included do
       delegate :npq?, :npq_profiles, to: :user
+      extend NPQClassMethods
     end
 
     def participant?
@@ -26,6 +27,12 @@ module RecordDeclarations
 
     def valid_declaration_types
       %w[started completed retained-1 retained-2]
+    end
+
+    module NPQClassMethods
+      def valid_courses_for_user
+        NPQCourse.identifiers
+      end
     end
   end
 end
--- a/app/controllers/api/v1/participant_declarations_controller.rb
+++ b/app/controllers/api/v1/participant_declarations_controller.rb
@@ -7,9 +7,8 @@ module Api
       include ApiTokenAuthenticatable
 
       def create
-        params = HashWithIndifferentAccess.new({ lead_provider_from_token: cpd_lead_provider }).merge(permitted_params["attributes"] || {})
-        validate_params!(params)
-        render json: RecordParticipantDeclaration.call(convert_params_for_declaration(params))
+        params = HashWithIndifferentAccess.new({ cpd_lead_provider: cpd_lead_provider }).merge(permitted_params["attributes"] || {})
+        render json: RecordParticipantDeclaration.call(params)
       end
 
     private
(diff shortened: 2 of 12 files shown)


app/services/record_declarations/npq.rb
```diff
@@ -27,5 +28,11 @@ def actual_lead_provider
     def valid_declaration_types
       %w[started completed retained-1 retained-2]
     end
+
+    module NPQClassMethods
+      def valid_courses_for_user
+        NPQCourse.identifiers
+      end
+    end
```
kicferk1, line 36: This need to make a new module to include sounds like a good reason to keep it as an instance method.

More importantly, I am wondering what the course identifier inclusion is meant to validate in the grand scheme of things. As I see it, we have a setup in controller which decides based on `course-identifier` which selects the service type. 

And then we validate that the service that we selected has access to the course identifier that we used to select it? Is this a valuable validation?

If we get a mentor participant and an `ecf-induction` course, the setup in controller will pick the ECT recorder, and validate that the course identifier is in `ecf-induction`. In particular, it won't fail. The old setup would construct the list of courses that the user can access, which the controller does not check when selecting the recorder, and then compare it with the course identifier - and so it could fail.

Although even the old setup does not seem that useful now that I look at it - because in the case I mentioned, we would not have the profile. And fetching the profile is actually more thorough too - cause it checks for the right NPQ identifier.

So the question really becomes: should we validate `course_identifier`? I don't see much value in it - it seems checking that the user has the right profile for the given course identifier is covering all edge cases. Thoughts?

kicferk1: I guess we can treat it as a safety check for developers messing around in rails console - but then I think we should rename the method to something like `valid_courses_for_recorder`. Thoughts?

blackrat: The validations at this level were put in place as much to have support and the console handle the issue and the move from schema-based validations was also to achieve this. Hence the safety check that you later mention.

In the example you give, that lack of failure isn't true. In the case of the `ecf-induction` course, the `user_profile` will return nil, since it wants an `early_career_teacher` induction which means that the `participant_id` will receive the error.

The move to the class based rather than the user.ect? and user.mentor? was because the error was being hidden due to a catch-all. When the participant_id passed the presence and format tests (the latter was missing with the move away from schema validations, so didn't fail), the find by for user would return `nil` which was then had its exception rewritten by the catch-all that this PR is meant to fix.

blackrat: Just `valid_courses` should convey the correct meaning. I want to use it in other places to, so it's not just going to be for the recorder.

kicferk1: Alright, I think I'm happy to keep it as a sanity check for support users, but if that's what we are keeping it as let's make it more obvious that that is what the list of course identifiers is for. `valid_courses_for_recorder` is my first go, if you can come up with a better name go for it. 

`valid_courses_for_user` counds a lot more like we are validating data, rather than sanity checking that the service is used right.

blackrat: I'll change this in the next PR

kicferk1 (approved):

PR #913, opened 2021-08-25
CPDTP-437 update ambition institute on the sandbox

blackrat:
### Context

One of our providers has asked for different data as a replacement in the sandbox. This adds in a script to safely remove the old data.

Remove the participants and schools added as part of the old data population script before adding in the new data.

### Review Checks

### DIFF
--- a/lib/tasks/valid_test_data_generator.rb
+++ b/lib/tasks/valid_test_data_generator.rb
@@ -107,9 +107,11 @@ module ValidTestDataGenerator
 
   class AmbitionSpecificPopulater < LeadProviderPopulater
     class << self
+      FIRST_AMBITION_SEED_DATA_TIME = ("2021-08-18 13:43".."2021-08-18 13:49").freeze
+
       def call(name:, total_schools: 3, participants_per_school: 3000)
         generator = new(name: name)
-        generator.remove_old_data(created_at: "2021-08-18 13:43".."2021-08-18 13:49")
+        generator.remove_old_data(created_at: FIRST_AMBITION_SEED_DATA_TIME)
         generator.call(total_schools: total_schools, participants_per_school: participants_per_school)
       end
     end
--- a/lib/tasks/valid_test_data_generator.rb
+++ b/lib/tasks/valid_test_data_generator.rb
@@ -6,7 +6,7 @@ require "tasks/trn_generator"
 module ValidTestDataGenerator
   class LeadProviderPopulater
     class << self
-      def call(name:, total_schools: 10, participants_per_school: 100, created_at: )
+      def call(name:, total_schools: 10, participants_per_school: 100)
         new(name: name).call(total_schools: total_schools, participants_per_school: participants_per_school)
       end
     end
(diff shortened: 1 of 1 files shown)


lib/tasks/valid_test_data_generator.rb
```diff
@@ -108,7 +108,37 @@ def weighted_choice(selection:, odds:)
   class AmbitionSpecificPopulater < LeadProviderPopulater
     class << self
       def call(name:, total_schools: 3, participants_per_school: 3000)
-        new(name: name).call(total_schools: total_schools, participants_per_school: participants_per_school)
+        generator = new(name: name)
+        generator.remove_old_data(created_at: "2021-08-18 13:43".."2021-08-18 13:49")
```
kicferk1, line 112: This `created_at` range seems very specific for something going into the general rake task.

Can we maybe put it in a constant, like "FIRST_DATA_SEED_TIME" or something?

lib/tasks/valid_test_data_generator.rb
```diff
@@ -108,7 +108,37 @@ def weighted_choice(selection:, odds:)
...
+
+    def remove_old_data(created_at:)
+      participants = lead_provider.ecf_participants.where(created_at: created_at)
+      participants.all.each do |participant|
+        tp = participant.teacher_profile
+        # Just in case the withdrawn merge happens before this is run.
+        # tp.participant_profiles.each do |pp|
+        #   pp.participant_profile_states.destroy_all!
+        # end
+        tp.destroy!
```
kicferk1, line 125: I hope we don't need that, we have 
```
  has_one :teacher_profile, dependent: :destroy
```
on `user` model.

lib/tasks/valid_test_data_generator.rb
```diff
@@ -108,7 +108,37 @@ def weighted_choice(selection:, odds:)
...
+        partnership = Partnership.find_by(lead_provider: lead_provider, school: school, cohort: Cohort.current)
+        delivery_partner = partnership.delivery_partner
+        provider_relationship = ProviderRelationship.find_by(lead_provider: lead_provider,
+                                                             cohort: Cohort.current,
+                                                             delivery_partner: delivery_partner)
+        provider_relationship.destroy!
+        partnership.destroy!
+        delivery_partner.destroy!
+        school_cohort = SchoolCohort.find_by(school: school, cohort: Cohort.current, induction_programme_choice: "full_induction_programme")
+        school_cohort.ecf_participant_profiles.destroy_all
```
kicferk1, line 139: Likewise here, I think `has_many :participant_profiles, dependent: :destroy` in teacher profile will take care of them before you reach this line.

lib/tasks/valid_test_data_generator.rb
blackrat: Yes. Whole rake task is very specific too though. It can be run again, but will always only add a delta to the previous run. I'll call it FIRST_AMBITION_SEED_DATA_TIME, since that was when their pass happened rather than all of them.

lib/tasks/valid_test_data_generator.rb
blackrat: I had thought about doing an .each(&:destroy!) on participants, but I thought this would make it more explicit and let me double check the ordering of creation/destruction if I had to do it again later.

lib/tasks/valid_test_data_generator.rb
blackrat: Same comment. I don't mind extra code when it's something that I may forget later.

kicferk1 (approved):

lib/tasks/valid_test_data_generator.rb
blackrat: Since it's now in a commit, I'll make that change too.

PR #916, opened 2021-08-25
Support UKPRNs for NPQs

asmega:
### Context

- UKPRN - UK Provider Reference Number - Another identifier for learning providers in the UK - register available at https://www.ukrlp.co.uk/ - It is an 8 digit number
- NPQ will officially support Further Education (FE) bodies in the near future although it is currently possible to sign up with an FE body
- These FE bodies include Local Authorities (LAs)
- These LAs only have UKPRNs associated with them and are not assigned a URN

- Incoming new NPQs will have a URN and UKPRN populated where possible
- Outgoing NPQ applications will have URN and UKPRN where the data is available
- After deployment we can retro fill any missing UKPRNs that we have so these are available to providers and for consistency
- Updated NPQ docs + guidance

### Guidance to review

- none

### Testing

- Hook up to NPQ registrations app
- Register with a school with a `urn` and `ukprn`
- Should see `NPQValidationData` with `urn` and `ukprn` set
- Register a new application with an LA
- Should see new `NPQValidationData` with `urn` missing (as it will not have one assigned) and `ukprn` populated
- Fetch these 2 record via api calls with json and csv
- Should see correct `urn` and `ukprn`s

### Review Checks

- Might be easier to test locally, otherwise you'll have to hand craft all the api calls

### DIFF
--- a/app/serializers/npq_application_csv_serializer.rb
+++ b/app/serializers/npq_application_csv_serializer.rb
@@ -31,6 +31,7 @@ private
       teacher_reference_number
       teacher_reference_number_validated
       school_urn
+      school_ukprn
       headteacher_status
       eligible_for_funding
       funding_choice
@@ -49,6 +50,7 @@ private
       record.teacher_reference_number,
       record.teacher_reference_number_verified,
       record.school_urn,
+      record.school_ukprn,
       record.headteacher_status,
       record.eligible_for_funding,
       record.funding_choice,
--- a/app/services/npq/create_or_update_profile.rb
+++ b/app/services/npq/create_or_update_profile.rb
@@ -13,14 +13,14 @@ module NPQ
         teacher_profile.trn = npq_validation_data.teacher_reference_number
       end
 
-      teacher_profile.school = school
       teacher_profile.save!
 
       participant_profile.schedule ||= Finance::Schedule.default
       participant_profile.npq_course ||= npq_validation_data.npq_course
-      participant_profile.school = school
       participant_profile.teacher_profile = teacher_profile
       participant_profile.user = user
+      participant_profile.school_urn = npq_validation_data.school_urn
+      participant_profile.school_ukprn = npq_validation_data.school_ukprn
       participant_profile.save!
     end
 
@@ -36,10 +36,6 @@ module NPQ
 
     def user
       @user ||= npq_validation_data.user
-    end
-
-    def school
-      @school ||= School.find_by(urn: npq_validation_data.school_urn)
     end
   end
 end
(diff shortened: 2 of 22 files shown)


app/serializers/npq_application_csv_serializer.rb
```diff
@@ -31,6 +31,7 @@ def csv_headers
       teacher_reference_number
       teacher_reference_number_validated
       school_urn
+      school_ukprn
```
kicferk1, line 34: Does the order matter here?

If so, do we want to add the new fields to the end of the list, rather than in the middle? I think I heard of some providers relying on indeces of rows rather than names in the header to figure out what's what 😭

swagger/v1/component_schemas/NpqApplicationAttributes.yml
```diff
@@ -8,7 +8,6 @@ required:
   - email_validated
   - teacher_reference_number
   - teacher_reference_number_validated
-  - school_urn
```
kicferk1, line 11: Is this not a required attribute? Without it, how could we (or the provider) know which school the participant is in?

app/serializers/npq_application_csv_serializer.rb
asmega: My understanding is that version 1 is not finalised and subject to changes on our end. I think we should to work to design at least a half decent api. I would deem grouping CSV columns logically to be sensible. If a provider chose to integrate with CSVs programatically and chose to work off indices and not field names then they should understand that implementing in such a fashion would be brittle therefore understood the risks of doing so.

I don't think we should work to design an API due to naive client implementations whilst it is still a work in progress.

kicferk1: Fair enough.

app/services/npq/create_or_update_profile.rb
```diff
@@ -13,14 +13,14 @@ def call
...
-      teacher_profile.school = school
       teacher_profile.save!
 
       participant_profile.schedule ||= Finance::Schedule.default
       participant_profile.npq_course ||= npq_validation_data.npq_course
-      participant_profile.school = school
       participant_profile.teacher_profile = teacher_profile
       participant_profile.user = user
+      participant_profile.school_urn = npq_validation_data.school_urn
+      participant_profile.school_ukprn = npq_validation_data.school_ukprn
```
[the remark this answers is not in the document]
asmega: I've made the URN visible for both the index view and details page. I've left the name blank though as I don't want to do any look up at this point

kicferk1 (approved):

PR #918, opened 2021-08-26
CPDRP-848 CIP only onboarding journey

FriedSock:
### Context

https://dfedigital.atlassian.net/browse/CPDRP-848

- Remove the advisory page before induction programme choice. It was decided that the information on this page is a bit superfluous as it is repeated on the next page where you actually get to make the choice. It also seemed simpler than defining a different set of copy for CIP-only schools.

- Copy updates to the induction programme choice form

- Hide the `full_induction_programme` choice for CIP-only schools.

### Testing

Added a new factory trait for cip only schools, tests for the form itself regarding the list of options.

### Review Checks

- Log in as an induction coordinator for a school that hasn't made any choices yet. Check that the first page you are directed to is the induction programme choice page.
- Log in as an induction coordinator for the CIP only school: cip-only-induction-tutor@example.com and compare the copy and options on the page with how it looked for a FIP eligible school.


app/models/concerns/gias_helpers.rb
```diff
@@ -10,6 +10,8 @@ module GiasHelpers
     scope :in_england, -> { where("administrative_district_code ILIKE 'E%'") }
     scope :section_41, -> { where(section_41_approved: true) }
     scope :eligible, -> { currently_open.eligible_establishment_type.in_england.or(currently_open.in_england.section_41) }
+    scope :cip_only, -> { currently_open.where(school_type_code: GiasTypes::CIP_ONLY_TYPE_CODES) }
```
ethanmills, line 13: This isn't quite right. If you're a CIP_ONLY type code, but have section 41 approval and are in England then you're fully eligible, not cip only

db/seeds/test_data.rb
```diff
@@ -98,6 +98,20 @@
   )
 end
 
+School.find_or_create_by!(urn: "1818181818") do |school|
```
ethanmills, line 101: This URN is the wrong length

app/models/concerns/gias_helpers.rb
ethanmills, line 13: Is there a spec for this?

FriedSock: Yeah this probably doesn't make sense as a scope in isolation as the query is going to add superfluous checks. Ill just add a test for the eligible_or_cip_only scope

ethanmills (approved):

PR #929, opened 2021-08-31
Add job to retry failed validations

ethanmills:
### Context
The DQT API went down last week, and there are records marked with API failure. We could fix it as a one-off, but I think the API is flaky enough that it's worth making this a regular thing.

I looked for ways to commonise this with the logic in the controller, but couldn't quite figure it out

### DIFF
--- a/app/jobs/validation_retry_job.rb
+++ b/app/jobs/validation_retry_job.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+class ValidationRetryJob < CronJob
+  self.cron_expression = "15 3 * * *"
+
+  queue_as :validation_retry
+
+  def perform
+    ECFParticipantValidationData.where(api_failure: true).each do |validation_data|
+      validation_result = ParticipantValidationService.validate(trn: validation_data.trn,
+                                                                full_name: validation_data.full_name,
+                                                                date_of_birth: validation_data.date_of_birth,
+                                                                nino: validation_data.nino)
+      ActiveRecord::Base.transaction do
+        validation_data.update!(api_failure: false)
+        if validation_result.present?
+          eligibility_data = store_eligibility_data!(validation_data, validation_result)
+          eligibility_data.manual_check_status! unless store_trn!(validation_result[:trn], validation_data.participant_profile)
+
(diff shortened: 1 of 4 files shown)


app/jobs/validation_retry_job.rb
```diff
@@ -0,0 +1,51 @@
...
+        if validation_result.present?
+          eligibility_data = store_eligibility_data!(validation_data, validation_result)
+          eligibility_data.manual_check_status! unless store_trn!(validation_result[:trn], validation_data.participant_profile)
+
+          validation_data.destroy! if eligibility_data.eligible_status?
+        end
+      end
+    rescue StandardError => e
+      Rails.logger.error("Problem with DQT API on retry: " + e.message)
+      Sentry.capture_message("Problem with DQT API on retry: " + e.message)
```
kicferk1, line 25: Do we need both?

kicferk1 (commented):
I guess if we wanted to commonise the parts between the job and the controller, we could put them in a service - `ECFValidationService` or something? 

Or perhaps the logic belongs in `ParticipantValidationService`.

Thoughts?

ethanmills:
> I guess if we wanted to commonise the parts between the job and the controller, we could put them in a service - `ECFValidationService` or something?

So much of the controller redirect logic is closely tied to the outcomes here that it's hard to extract it out. We can't even commonise storing the results, because the ECFParticipantEligibilityData already exists here.

> Or perhaps the logic belongs in ParticipantValidationService.

this is not ECF specific, the NPQ API uses it too. I wouldn't want to put extra stuff in there, I think it should just be about matching DQT records

app/jobs/validation_retry_job.rb
ethanmills: I don't think it does any harm

kicferk1 (approved):

app/jobs/validation_retry_job.rb
kicferk1: I am just worried that in other places when we capture sentry we don't explicitly log an error in rails. It would be nice to be consistent.

Is there some way to tell Sentry to rails log the errors it catches?

PR #932, opened 2021-08-31
CPDTP-362 Change to return full participant json object instead of just an ID

plisovin:
### Context

Submission of a participant declaration should return a full participant declaration json object rather than just an ID

Modified response
Fixed rspec tests

### Review Checks

### DIFF
--- a/app/services/record_declarations/base.rb
+++ b/app/services/record_declarations/base.rb
@@ -43,9 +43,7 @@ module RecordDeclarations
       declaration.refresh_payability!
       declaration_attempt.update!(participant_declaration: declaration)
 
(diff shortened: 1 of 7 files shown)


app/services/record_declarations/base.rb
```diff
@@ -43,7 +43,9 @@ def call
       declaration.refresh_payability!
       declaration_attempt.update!(participant_declaration: declaration)
 
-      { id: declaration.id }
+      participant_declarations_hash = ParticipantDeclarationSerializer.new(declaration).serializable_hash
```
blackrat, line 46: Why introduce a new variable and then just to_json it?

plisovin: Done in 78054cc

kicferk1 (approved):

PR #933, opened 2021-08-31
Bump all production-like environments to four instances

ethanmills:
We're expecting an increase in load at the beginning of September. Throw more compute at the problem - bound to work


kicferk1 (approved):
Do we expect that much more load on non-prod environments? Also, @FriedSock @xswoodsx , FYI - if our monitoring shows that E&L is struggling, this is how we would increase its power.

ethanmills:
I expect the lead providers to pound the sandbox. Staging... Staging is useless

PR #935, opened 2021-08-31
CPDTP-369: Change participant's assigned induction schedule

kicferk1:
### Context
We want LPs to be able to change participant schedules. https://dfedigital.atlassian.net/browse/CPDTP-369 

Add an endpoint mimicking `withdraw` one.
Add services mimicking `withdraw` ones.
Add tests and docs mimicking... you guessed it, `withdraw` ones.

Add identifiers to schedules.
Add a new schedule - will need to make sure I re-run seeds when this goes in and we deploy.

### Guidance to review
At the moment we store which schedule participant profile pointed at in papertrail, but not anywhere else. I need to double check that this is ok to do. In principle if a provider wants to back-date something, it could be useful. And I guess tracking what LPs are doing in our business logic could be useful. 

But 'I guess' is not quite enough, and there are other ways of achieving those goals too. 

If we end up wanting it, I would be tempted to stick it on `ParticipantProfileState`.

### Review Checks

Participants should have schedule identifier on them (on participant endpoint).
Using the endpoint it should be possible to change it from `ecf-september-standard-2021` to `ecf-september-extended-2021`.


app/services/factories/course_identifier.rb
```diff
@@ -10,7 +10,7 @@ def call(course)
     private
 
       def recorder_klass_name_for_course_identifier(course)
-        declaration_identifiers[course.underscore.intern].to_s
+        declaration_identifiers[course.underscore.intern].to_s if course
```
blackrat, line 13: Is this enough. Should we also have the caller with a catch block as per the event factory?

app/services/factories/event.rb
```diff
@@ -12,7 +12,7 @@ def call(event)
     private
 
       def event_namespace_for_event(event)
-        event_identifiers[event.underscore.intern].to_s
+        event_identifiers[event.underscore.intern].to_s if event
```
blackrat, line 15: Do we need this and the catch block in the caller?

blackrat:
Now that we have an identifier for the schedules, is that a better way to handle the milestone lookups in the previous "schedules and milestones" creation PR?

app/services/factories/course_identifier.rb
kicferk1:
```
      def call(course)
        recorder_klass_name_for_course_identifier(course).presence || (raise ActionController::ParameterMissing, I18n.t(:invalid_course))
      end
```
should prevent the caller from doing anything.

app/services/factories/event.rb
kicferk1: No, we don't, good spot :)

app/services/factories/course_identifier.rb
kicferk1: Alright, I made it similar to event factory :)

kicferk1:
> Now that we have an identifier for the schedules, is that a better way to handle the milestone lookups in the previous "schedules and milestones" creation PR?

This PR adds a new way of fetching schedules (by their identifier). But in the milestone lookups, we already know which schedule we are using - we get it from participant profile.

So I think this PR does not provide any additional value there. We would need to add milestone lookups for that.

swagger/v1/api_spec.json
```diff
@@ -587,21 +636,60 @@
...
+        }
+      },
+      "ECFParticipantChangeScheduleRequest": {
+        "description": "The change schedule request for a participant",
+        "type": "object",
+        "required": [
+          "data"
+        ],
+        "properties": {
+          "data": {
```
blackrat, line 678: Does `data` not need to have a `type` and `description` too to make this a valid schema?

kicferk1: Interesting, it seems to work fine, but I added them.

swagger/v1/api_spec.json
```diff
@@ -676,13 +676,15 @@
...
+            "type": "object",
+            "description": "The request data",
             "required": [
               "type",
               "attributes"
             ],
             "properties": {
               "type": {
-                "description": "The data type",
+                "description": "The data attributes",
```
blackrat: It's not the attributes, it is the type of request that's being made.

swagger/v1/api_spec.json
```diff
@@ -811,13 +813,15 @@
...
```
blackrat: Same comment

swagger/v1/component_schemas/ECFParticipantChangeScheduleRequest.yml
```diff
@@ -4,12 +4,14 @@ required:
...
   data:
+    type: object
+    description: "The request data"
     required:
       - type
       - attributes
     properties:
       type:
-        description: "The data type"
+        description: "The data attributes"
```
blackrat, line 14: and again

swagger/v1/component_schemas/ECFParticipantWithdrawRequest.yml
```diff
@@ -4,12 +4,14 @@ required:
...
   data:
+    type: object
+    description: "The request data"
     required:
       - type
       - attributes
     properties:
       type:
-        description: "The data type"
+        description: "The data attributes"
```
blackrat, line 14: and again.

blackrat (approved):

PR #937, opened 2021-09-01
CPDTP-362 fix ECF and NPQ naming

blackrat:
### Context

Making the naming of the components and namespaces consistent with the inflections used.

- Change some file and namespace naming, so that we have internal and external consistency when ECF and NPQ are used

### Guidance to review

- As well as the file name changes, this will also change the display of the API documentation to make the names consistent.

### Review Checks

View the API documentation


swagger/v1/component_schemas/ECFParticipantAttributes.yml
```diff
@@ -0,0 +1,78 @@
+description: "The data attributes associated with an ECF participant"
```
kicferk1, line 1: Should we also delete the old version of this file and all the other incorrectly named yml files?

blackrat: Yes. Weirdly enough they weren't on my local machine. Could be a Mac case-indifferent filing system thing.

kicferk1 (approved):

jamesrose (approved):

PR #938, opened 2021-09-01
CPDRP-908 Add validation notification emails

FriedSock:
### Context
This email will be a notification to SITs to tell them about the upcoming validation release. We want to exclude
schools that have added participants as part of the validation beta, but practically everyone else who it will get
released to

Add a new method that we will call from the console to trigger emails to these users. 

exact specification of audience:
```
1. SITs who HAVE added their ECTs and mentors BUT have not been contacted for validation Beta
2. SITs who have signed in AND chosen their programme, but have not added their ECTs and mentors
```

### Review Checks


app/mailers/participant_validation_mailer.rb
```diff
@@ -9,6 +9,8 @@ class ParticipantValidationMailer < ApplicationMailer
   ENGAGE_BETA_MENTOR_TEMPLATE = "99dacb2d-2255-4aa8-9076-46fa6093f1e5"
   INDUCTION_COORDINATOR_NOTIFICATION_TEMPLATE = "d560fb2e-243d-48b1-bf61-c7e111f56858"
   INDUCTION_COORDINATOR_NOTIFICATION_UR_TEMPLATE = "afb54050-7ebd-43af-ad83-7fa6795d1523"
+  # TODO: add real uuid, this is a dummy one
+  INDUCTION_COORDINATOR_VALIDATION_NOTIFICATION_TEMPLATE = "4eb4455d-6b33-4af3-aa40-43b47bfc5501"
```
ethanmills, line 13: Not sure this is the notification email, is it? This might need renaming

app/mailers/participant_validation_mailer.rb
```diff
@@ -105,6 +107,19 @@ def induction_coordinator_email(recipient:, school_name:, start_url:)
...
 
+  def induction_coordinator_validation_notification_email(recipient:, school_name:, start_url:)
+    template_mail(
+      INDUCTION_COORDINATOR_VALIDATION_NOTIFICATION_TEMPLATE,
+      to: recipient,
+      rails_mailer: mailer_name,
+      rails_mail_template: action_name,
+      personalisation: {
+        school_name: school_name,
+        start_url: start_url,
```
ethanmills, line 118: Probably going to want a step-by-step url in here?

app/services/validation_beta_service.rb
```diff
@@ -22,6 +22,14 @@ def invite_schools(array_of_urns:, participant_research:, coordinator_research:)
     end
   end
 
+  def notify_induction_coordinators_about_validation
+    School.eligible.find_each do |school|
```
ethanmills, line 26: And cip schools?

app/services/validation_beta_service.rb
```diff
@@ -22,6 +22,14 @@ def invite_schools(array_of_urns:, participant_research:, coordinator_research:)
...
+      if added_participants_and_not_in_beta(school) || chosen_programme_but_not_added_participants(school)
```
ethanmills, line 27: This condition simplifies to "chosen programme and not in beta"

app/services/validation_beta_service.rb
```diff
@@ -22,6 +22,14 @@ def invite_schools(array_of_urns:, participant_research:, coordinator_research:)
...
+        send_validation_notification(school)
```
ethanmills, line 28: Are we ok with SITs at multiple schools getting lots of notifications?

app/services/validation_beta_service.rb
FriedSock: Do you mean cip-only schools? They won't have added any participants as they aren't invited yet right? And we are emailing them separately later in the week

app/services/validation_beta_service.rb
FriedSock: > 2. SITs who have signed in AND chosen their programme, but have not added their ECTs and mentors

I understood this as the second set of schools could be in beta, will clarify

FriedSock: Got you, if they haven't added participants they wouldn't be invited to beta

app/mailers/participant_validation_mailer.rb
FriedSock: I don't know, we are notifying them about the validation changes aren't we? Do you have a suggestion?

ethanmills: In beta, we also sent a notification that we've asked their participants to validate. If we're still going to do that, this needs to be called something like "prompt to update". If not, the other templates should probably get a BETA in the name (then be deleted once we know we don't need them). Asked here: https://ukgovernmentdfe.slack.com/archives/C01J1J4FM50/p1630492003125600

ethanmills:
Why the label @jamesrose ?

kicferk1:
> Why the label @jamesrose ?

The fake uid for notify template perhaps?

app/services/validation_beta_service.rb
tonyheadford: I think what @ethanmills means is that this will send an email to CIP schools?

app/services/validation_beta_service.rb
FriedSock: Fixed this now

app/mailers/participant_validation_mailer.rb
FriedSock: I've asked about this

app/services/validation_beta_service.rb
FriedSock: The query has changed quite a bit now to loop over induction coordinators to make sure we just send one email each. 
https://github.com/DFE-Digital/early-careers-framework/pull/938/files#diff-4c570b8127fbf99f726f6b1a409d563cc29b16e9c464c92b90173c71b58ef416R25-R31
I didn't know we wanted to make a distinction between CIP or FIP schools, just that they had chosen a programme, do we need to?

ethanmills: Nope, I meant cip-only schools. If we're targeting them separately then that's fine

app/mailers/participant_validation_mailer.rb
```diff
@@ -105,6 +108,21 @@ def induction_coordinator_email(recipient:, school_name:, start_url:)
...
+    template_mail(
+      INDUCTION_COORDINATOR_CHECK_ECT_AND_MENTOR_TEMPLATE,
+      to: recipient,
+      rails_mailer: mailer_name,
+      rails_mail_template: action_name,
+      personalisation: {
+        statutory_guidance: STATUTORY_GUIDANCE_LINK,
+        sign_in: new_user_session_url,
+        step_by_step: step_by_step_url,
+        resend_email: resend_email_request_nomination_invite_url,
```
ethanmills, line 121: Missing UTM params here. We'd normally pass these in as params so we can get the right campaign

FriedSock: Added now

app/services/validation_beta_service.rb
```diff
@@ -185,5 +193,35 @@ def send_induction_coordinator_notification(participant_research, school)
...
+      host: Rails.application.config.domain,
+      **UTMService.email(campaign, campaign),
+    )
+
+    step_by_step_url = Rails.application.routes.url_helpers.step_by_step_url(
+      host: Rails.application.config.domain,
+      **UTMService.email(campaign, campaign),
+    )
+
+    resend_email_url = Rails.application.routes.url_helpers.step_by_step_url(
```
ethanmills, line 209: Wrong url?

ethanmills (approved):

PR #940, opened 2021-09-01
CPDTP-362 Consistent errors in the declaration API

plisovin:
### Context

Errors are wrapped in the expected format i.e.

```
{
  "errors": [
    {
      "title": "Bad or missing parameters",
      "detail": "is invalid"
    }
  ]
}
```

New factory to generate the json error object from the provided parameters.
Also validation for evidence_held has been fixed

### Review Checks

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -19,11 +19,11 @@ module Api
     end
 
     def missing_parameter_response(exception)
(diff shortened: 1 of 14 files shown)


kicferk1 (approved):
This looks neat. Should we also think about using `Api::ErrorFactory` for the errors coming from our service-level validations? Perhaps in another PR?

kicferk1 (approved):

PR #941, opened 2021-09-01
CPDRP-800: Add analytics db

ethanmills:
### Context
We want to stream some anonymised updates to a more accessible database for analytics. I've set up a couple of postgres DBs in GCP for this.

When this is merged, you may need to set the analytics db credentials in your `.env` file. You can use the same database server as for the primary application database. You may then need to run `rails db:setup`. 

Once this is on production, the following should be run to backfill the analytics data:

```ruby
ParticipantProfile.ecf.each { |profile| Analytics::ECFValidationService.upsert_record(profile) }
```


asmega:
Do we need consider latency? PaaS and PaaS postgres are probably right next to each other. I have no idea where GCP postgres instances are in relation to PaaS. Are we happy for these calls to be synchronous?

spec/services/mentors/create_spec.rb
```diff
@@ -16,7 +16,7 @@
         school_cohort: school_cohort,
       )
     }.to change { ParticipantProfile::Mentor.count }.by(1)
-     .and not_change { User.count }
+                                                    .and not_change { User.count }
```
kicferk1, line 19: Sad whitespace

kicferk1 (approved):
It looks reasonable, let's see what it does.

ethanmills:
> Do we need consider latency? PaaS and PaaS postgres are probably right next to each other. I have no idea where GCP postgres instances are in relation to PaaS. Are we happy for these calls to be synchronous?

Nope, they're asynchronous: https://github.com/DFE-Digital/early-careers-framework/pull/941/files#diff-7acba4097cf257b1582b1e26d564e98c97623ee7ffe375a108815ff9f0fedad7R26

PR #942, opened 2021-09-01
Preventing accidental inclusion of url_helpers into ALL ruby objects

BroiSatse:
https://dfedigital.atlassian.net/browse/CPDRP-893

There seems to be just a single test causing issues in other tests. This PR fixes the issues.

One of the tests were invoking one of the rake tasks, which in turn was calling `include <URL_HELPERS>` without any receiver. In result, url_helpers has been included into Object, making it available for all the application objects (including `nil`). In the meantime, the depth of rails routing is invoking `url_for` on the `_routes` - `_routes` is a memoized method and its memoization is dependant on `@router.respond_to?(:url_for)` (`_routes` should be an instance of ActionDispatch::Routing::RouteSet`). After the inclusion of url_helpers into `nil`, this caused the instance variable not to be set at all leading to `wrong number of arguments 4, expecting 0..1` as it was invoking `<helpers>#url_for` on `nil` instead of `RouteSet#url_for`.

### DIFF
--- a/spec/smoke_tests/participant_declarations_spec.rb
+++ b/spec/smoke_tests/participant_declarations_spec.rb
@@ -3,24 +3,19 @@
 require "capybara/rspec"
 require "yaml"
 
-CLOUDAPPS_DOMAIN = "london.cloudapps.digital"
-
 RSpec.describe "Participant API availability", smoke_test: true do
   context "ECF participant declarations" do
     let(:smoke_test_domain) { fetch_smoke_test_domain }
 
     before do
+      skip "No domain for smoke tests found" unless smoke_test_domain
       WebMock.allow_net_connect!
     end
 
     it "ensures participant declaration api allows declarations" do
-      if smoke_test_domain
-        participant = fetch_first_active_participant
-        res = make_started_declaration(participant)
-        expect(res).to be_a Net::HTTPSuccess
-      else
-        puts "No domain for smoke tests found, skipping..."
-      end
+      participant = fetch_first_active_participant
+      res = make_started_declaration(participant)
+      expect(res).to be_a Net::HTTPSuccess
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -96,14 +96,16 @@
...
   #
   #   # Seed global randomization in this process using the `--seed` CLI option.
   #   # Setting this allows you to use `--seed` to deterministically reproduce
   #   # test failures related to randomization by passing the same `--seed` value
   #   # as the one that triggered the failure.
-  # Kernel.srand config.seed
+  Kernel.srand config.seed
   config.extend WithModel
+
+  config.bisect_runner = :shell
```
kicferk1, line 108: What is this for?

spec/smoke_tests/participant_declarations_spec.rb
```diff
@@ -30,7 +25,7 @@ def fetch_smoke_test_domain
     paas_environment = ENV.fetch("ENVIRONMENT")
     YAML.load_file("#{__dir__}/../../terraform/workspace-variables/#{paas_environment}_app_env.yml")["DOMAIN"]
   rescue Errno::ENOENT
-    "ecf-#{paas_environment}.#{CLOUDAPPS_DOMAIN}"
```
jamesrose, line 33: Isn't this the same thing?

spec/spec_helper.rb
BroiSatse: Default bisect runner is `fork` which tries to run all the tests within a single system process. `shell` is executing each run by invoking `rspec` command in a shell subprocess. The later is quite  a bit slower, but `fork` has a tendency to end up in a Mutex deadlock - and in our case it did pretty much all the time making bisect completely unusable.

spec/smoke_tests/participant_declarations_spec.rb
BroiSatse: Yes, this constant has been previously defined on line 6 in the same file. I've decided to remove the constant as it was only used in this one place and I have quite strong feeling about local-but-global constants hanging around.

kicferk1 (approved):

PR #946, opened 2021-09-01
CPDRP-908 Don't email schools that have opted out of updates

FriedSock:
### Context

A piece of logic was missed in this previous PR: https://github.com/DFE-Digital/early-careers-framework/pull/938

We need to adjust this to make sure we don't email any schools that have opted out of updates

As above

### Review Checks

### DIFF
--- a/app/services/validation_beta_service.rb
+++ b/app/services/validation_beta_service.rb
@@ -24,7 +24,7 @@ class ValidationBetaService
 
   def tell_induction_coordinators_to_check_ect_and_mentor_information
     InductionCoordinatorProfile.find_each do |ic|
-      if ic.schools.any? { |school| chosen_programme_and_not_in_beta(school) }
+      if ic.schools.not_opted_out.any? { |school| chosen_programme_and_not_in_beta(school) }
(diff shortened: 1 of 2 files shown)


app/services/validation_beta_service.rb
```diff
@@ -24,7 +24,7 @@ def invite_schools(array_of_urns:, participant_research:, coordinator_research:)
 
   def tell_induction_coordinators_to_check_ect_and_mentor_information
     InductionCoordinatorProfile.find_each do |ic|
-      if ic.schools.any? { |school| chosen_programme_and_not_in_beta(school) }
+      if ic.schools.any? { |school| chosen_programme_and_not_in_beta(school) && !opted_out_of_updates(school) }
```
kicferk1, line 27: We have a scope on schools, `not_opted_out`.

That means we can do 
```
if ic.schools.not_opted_out.any? { |school| chosen_programme_and_not_in_beta(school) }
```
without the extra cumbersome check.
```

kicferk1 (approved):

PR #948, opened 2021-09-01
Cpdtp 367 deferral

blackrat:
### Context

Adding in the ability for a lead_provider to defer a participant
Cleaning up a lot of the commonality between similar actions

- Simplify the controller and eliminate the intermediate factory handler
- Create a simple state machine to handle transitions as validations and extend for deferred
- Extract validation and action execution functionality into modules for composition
- Add in a "strawman" abstract interface to simplify some of the delegation protection
- Fix a bug in the ECF participant delegation
- Replace custom validators with inclusion validation and move more messages to I18n activemodel section
- Extend tests to cover deferral and new state transition checks.

### Review Checks


kicferk1 (commented):
I know it's a draft PR, so some of my comments will probably be invalid soon-ish.

The general feeling I have about it is that the refactor to remove base service classes for each action and to have a single base class does not look too convincing - the cracks in the walls are that we need to do a few hacks to handle changing schedules correctly. 

But as our system complexity grows and actions become more sophisticated, we probably will find more need for divergence on an action level. So even if those services are small, I would keep them.

app/services/participants/defer/ecf.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Participants
+  module Defer
+    class ECF < ::Participants::Base
+      include Participants::ECF
+      include StateValidation
+      attr_accessor :reason
+      validates :reason, presence: true
+      validates :reason, "deferred/ecf": true
```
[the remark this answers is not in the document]
blackrat: Messages will be in the en.yml file, but attached to activemodel using the rails validation standard, so we don't need to be explicit here.
Good spot on allow_blank. We don't want to repeat error messages.

app/services/participants/defer/state_validation.rb
```diff
@@ -0,0 +1,24 @@
...
+          :deferred
+        end
+
+        def states_to_transition_from
+          %i[active]
+        end
+      end
+    end
+  end
+end
```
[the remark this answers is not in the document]
blackrat: ignore this, I'm sketching up the transition validations with allowed states.

app/services/participants/base.rb
```diff
@@ -0,0 +1,63 @@
...
+    class << self
+      def call(params:)
+        new(params: params).call
+      end
+
+      def not_implemented_error
+        raise NotImplementedError, "Method must be implemented"
+      end
+
+      delegate :valid_courses, :state_to_transition_to, to: :not_implemented_error
```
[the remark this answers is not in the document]
blackrat: Ignore. It's a placeholder for testing state and won't be there too long.

app/services/participants/base.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    def business_validation!
+      validate_provider!
+    end
+
+    def perform_action!
+      ParticipantProfileState.create!(participant_profile: user_profile, state: self.class.state_to_transition_to, reason: reason)
+      user_profile
+    end
```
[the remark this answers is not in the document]
blackrat: Ignore. It's not going to state there.

app/services/participants/base.rb
```diff
@@ -0,0 +1,63 @@
...
+        raise ActionController::ParameterMissing, errors.map(&:message)
+      end
+
+      business_validation!
+      perform_action!
+    end
+
+    def business_validation!
+      validate_provider!
+    end
```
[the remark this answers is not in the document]
blackrat: We do. This allows the `business_validation!` to be overridden separately from the `action!`, which will be in a later commit.

app/services/participants/base.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      delegate :valid_courses, :state_to_transition_to, to: :not_implemented_error
+    end
+
+    def call
+      unless valid?
+        raise ActionController::ParameterMissing, errors.map(&:message)
+      end
+
+      business_validation!
```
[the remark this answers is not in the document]
blackrat: Yes. I want to move that one, since the errors that it will give should be different when it's an invalid state transition as opposed to a guard against changing other things.

app/services/participants/change_schedule/ecf.rb
```diff
@@ -4,6 +4,12 @@ module Participants
   module ChangeSchedule
     class ECF < Base
       include Participants::ECF
+      include ScheduleValidation
+
+      def perform_action!
+        user_profile.update_schedule!(schedule)
+        user_profile
+      end
     end
```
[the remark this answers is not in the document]
blackrat: I see a lot of this being composable.
For example, "reasons" are required for some state transitions but not others, so they could be shared between `deferred` and `withdrawn`.
I see `resume` and `resume with schedule_change` as just `resume` and passed either a schedule or defaulting to what it had before. There may be validation required on "not" changing the schedule in the latter instance, so it should be checked for validity before a resumption is allowed.

app/services/participants/defer/ecf.rb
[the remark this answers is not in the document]
blackrat: Actually, that's now gone too.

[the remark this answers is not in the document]
blackrat: But the inclusion that's replaced it needs it.

app/services/participants/base.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    def initialize(params:)
+      params.each do |param, value|
+        send("#{param}=", value)
+      end
+    end
+
+    def not_implemented_error
+      self.class.not_implemented_error
+    end
```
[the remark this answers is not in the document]
kicferk1: Ehh, I don't know.

I think a module for it is a better solution than duplication. That said,

> It's a better pattern for pure abstract base classes

Maybe it is. Makes me wonder why it's not a part of Ruby 😂

app/services/participants/actions/base.rb
```diff
@@ -0,0 +1,38 @@
...
+          raise NotImplementedError, "Method must be implemented"
+        end
+
+        delegate :recorder_namespace, to: :not_implemented_error
+      end
+
+      def call
+        recorder = "#{self.class.recorder_namespace}::#{::Factories::CourseIdentifier.call(course_identifier)}".constantize
+        recorder.call(params: params)
+      end
```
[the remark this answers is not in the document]
blackrat: Makes the factory a bit harder to test, but could trade that off with a private method test.

app/services/participants/base.rb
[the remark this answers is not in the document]
blackrat: > Makes me wonder why it's not a part of Ruby 😂

Don't get me started on things that should be part of Ruby. 🔢

swagger/v1/component_schemas/ECFParticipantAttributes.yml
```diff
@@ -70,12 +70,13 @@ properties:
     type: boolean
     nullable: true
     example: true
-  state:
-    description: "The state of the ECF participant"
+  training_status:
+    description: "The training status of the ECF participant"
```
kicferk1, line 74: Can we have it in its own little PR please? I would like this to go in as quick as possible, and this PR is large and complex :/

blackrat: Done. This one adds in the "defer" status as well though, which isn't in the previous PR, so it'll still be a (smaller) change in this PR.

app/models/participant_profile_state.rb
```diff
@@ -1,9 +1,26 @@
...
+    current_state = participant_profile.participant_profile_state
+    return true if current_state.nil?
+
+    case state
+    when "active"
+      errors.add(:base, "already active") if current_state.active?
+    when "withdrawn"
+      errors.add(:base, I18n.t(:invalid_withdrawal)) if current_state.withdrawn?
+    when "deferred"
+      errors.add(:base, I18n.t(:invalid_withdrawal)) if current_state.withdrawn?
```
kicferk1, line 14: I wonder if those two should be the same error messages. 
> The participant is already withdrawn

sounds good when you try to withdraw, perhaps slightly less good when you try to defer.

app/services/participants/base.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+    end
+
+    def call
+      unless valid?
+        raise ActionController::ParameterMissing, errors.map(&:message)
+      end
+
+      business_validation!
+      action!
```
kicferk1, line 23: I liked `perform_action!` more as the name.

app/services/participants/base.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    def business_validation!
+      validate_provider!
+    end
```
kicferk1, line 28: `business_validation` is never overwritten, and all it does is one line. Can we inline it?

app/services/participants/base.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def initialize(params:)
+      params.each do |param, value|
+        send("#{param}=", value)
+      end
+    end
+
+    def validate_provider!
+      return if errors.any?
+      raise ActionController::ParameterMissing, [I18n.t(:invalid_participant)] unless matches_lead_provider?
```
kicferk1, line 42: Does `matches_lead_provider` need to be marked as abstract?

app/services/participants/base.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require "abstract_interface"
+
+module Participants
+  class Base
+    include ProfileAttributes
```
kicferk1, line 7: I just noticed, without this inclusion, we would not have `ActiveModel::Validations` in this base class. But it refers to errors. Should we include it by default?

app/services/participants/change_schedule/validate_and_change.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Participants
+  module ChangeSchedule
+    module ValidateAndChange
```
kicferk1, line 5: Can we call it `ValidateAndChangeSchedule`? So that we don't get confused with other `ValidateAndChange` modules?

app/services/participants/defer/ecf.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Participants
+  module Defer
+    class ECF < ::Participants::Base
+      class << self
+        def reasons
+          %w[
+            other
+          ].freeze
```
kicferk1, line 10: I guess this is changing soon

app/services/participants/defer/validate_and_change.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Participants
+  module Defer
+    module ValidateAndChange
```
kicferk1, line 5: Ditto, I really don't like the idea of modules being named the same thing

app/services/participants/withdraw/ecf.rb
```diff
@@ -2,10 +2,23 @@
...
+            mentor-no-longer-being-mentor
+            school-left-fip
+            career-break
+            passed-induction
+            other
+          ].freeze
+        end
+      end
 
-      validates :reason, "withdrawn/ecf": true
```
kicferk1, line 8: I like getting rid of the validators in favour of inclusion validations :)

app/services/participants/withdraw/validate_and_change.rb
```diff
@@ -0,0 +1,21 @@
...
+module Participants
+  module Withdraw
+    module ValidateAndChange
+      extend ActiveSupport::Concern
+      include ActiveModel::Validations
+
+      included do
+        attr_accessor :reason
+        validates :reason, presence: true
+        validates :reason, inclusion: { in: reasons, case_sensitive: false }, allow_blank: true
```
kicferk1, line 12: Why do we not want it to be case sensitive?

config/locales/en.yml
```diff
@@ -39,12 +39,12 @@ en:
...
-    mentor_no_longer_mentor: "Mentor no longer being mentor"
-    school_left_fip: "School left FIP"
-    career_break: "Career break"
-    passed_induction: "Passed Induction"
+    left_teaching_profession: Left teaching profession
+    moved_school: Moved school
+    mentor_no_longer_mentor: Mentor no longer being mentor
+    school_left_fip: School left FIP
+    career_break: Career break
+    passed_induction: Passed Induction
```
kicferk1, line 47: What's the difference? If it is significant, then we have more things to clean up in this file...

spec/docs/participants_spec.rb
```diff
@@ -91,6 +91,61 @@
...
+      parameter name: :params,
+                in: :body,
+                type: :object,
+                style: :deepObject,
+                required: true,
+                schema: {
+                  "$ref": "#/components/schemas/ECFParticipantDeferRequest",
+                }
+
+      response "200", "The ECF participant being withdrawn" do
```
kicferk1, line 125: deferred rather than withdrawn

spec/services/participants/defer/early_career_teacher_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  context "when lead providers don't match" do
+    it "raises a ParameterMissing error" do
+      expect { described_class.call(params: participant_params.merge({ cpd_lead_provider: another_lead_provider })) }.to raise_error(ActionController::ParameterMissing)
+    end
+  end
+
+  context "when valid user is an early_career_teacher" do
+    it "creates a withdrawn state for that user's profile" do
```
kicferk1, line 25: deferred

Also, in the test we should check the actual state

spec/services/participants/defer/mentor_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    it "raises a ParameterMissing error" do
+      expect { described_class.call(params: participant_params.merge({ cpd_lead_provider: another_lead_provider })) }.to raise_error(ActionController::ParameterMissing)
+    end
+  end
+
+  context "when valid user is an mentor" do
+    it "creates a withdrawn state for that user's profile" do
+      expect { described_class.call(params: participant_params) }
+        .to change { ParticipantProfileState.count }.by(1)
+    end
```
kicferk1, line 28: Ditto

spec/services/participants/defer/npq_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    it "raises a ParameterMissing error" do
+      expect { described_class.call(params: participant_params.merge({ cpd_lead_provider: another_lead_provider })) }.to raise_error(ActionController::ParameterMissing)
+    end
+  end
+
+  context "when valid user is an NPQ" do
+    it "creates a withdrawn state for that user's profile" do
+      expect { described_class.call(params: participant_params) }
+        .to change { ParticipantProfileState.count }.by(1)
+    end
```
kicferk1, line 29: Ditto

swagger/v1/api_spec.json
```diff
@@ -493,6 +493,55 @@
...
+              }
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "required": true,
+            "example": "",
```
kicferk1, line 524: Should we give id examples here?

app/services/participants/base.rb
blackrat: True. I changed it and then folded it into a standard validation for the inherited class. I'll inline.

app/services/participants/change_schedule/validate_and_change.rb
blackrat: Sure. I'll also rename the other to ValidateAndChangeState

kicferk1 (approved):

kicferk1 (approved):
🚢

kicferk1 (approved):

PR #949, opened 2021-09-02
Increase API pagination max per page from 250 to 3000

jamesrose:
### Context

A provider would like to be able to retrieve more participants using the API so that they can continue their integration. I have tested up to 3000 participants which was returned in c. 1.5s, which seems reasonable to me.

### DIFF
--- a/swagger/v1/api_spec.json
+++ b/swagger/v1/api_spec.json
@@ -1718,7 +1718,7 @@
           },
           "per_page": {
             "type": "integer",
-            "description": "The number items to display on a page. Defaults to 100. Maximum is 250, if the value is greater that the maximum allowed it will fallback to 250.",
+            "description": "The number items to display on a page. Defaults to 100. Maximum is 3000, if the value is greater that the maximum allowed it will fallback to 3000.",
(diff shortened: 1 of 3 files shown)


kicferk1 (approved):

swagger/v1/api_spec.json
```diff
@@ -1718,7 +1718,7 @@
           },
           "per_page": {
             "type": "integer",
-            "description": "The number items to display on a page. Defaults to 100. Maximum is 250, if the value is greater that the maximum allowed it will fallback to 250.",
+            "description": "The number items to display on a page. Defaults to 250. Maximum is 3000, if the value is greater that the maximum allowed it will fallback to 250.",
```
FriedSock, line 1721: the `default_per_page` is still set to 100, this bit also needs changing:
```
if the value is greater that the maximum allowed it will fallback to ~250~ 3000
```

jamesrose: Good spot. I knew there wasn't something right about that.

FriedSock (approved):

PR #952, opened 2021-09-02
CPDTP-429 Give providers feedback about declaration pay-ability

plisovin:
### Context

Lead providers need information about their declarations for a specific participant

A new filter ```participant_id``` has been added to the participant-declarations API 

### Review Checks

### DIFF
--- a/app/models/participant_declaration.rb
+++ b/app/models/participant_declaration.rb
@@ -21,7 +21,6 @@ class ParticipantDeclaration < ApplicationRecord
   scope :npq, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.npq_profiles) }
   scope :payable, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.where(payable: true)) }
   scope :unique_id, -> { select(:user_id).distinct }
-  scope :for_participant, ->(*args) { where(user_id: args[0]) if args.any? }
 
   # Time dependent Range scopes
   scope :declared_as_between, ->(start_date, end_date) { where(declaration_date: start_date..end_date) }
--- a/app/models/participant_declaration.rb
+++ b/app/models/participant_declaration.rb
@@ -21,6 +21,7 @@ class ParticipantDeclaration < ApplicationRecord
   scope :npq, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.npq_profiles) }
   scope :payable, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.where(payable: true)) }
   scope :unique_id, -> { select(:user_id).distinct }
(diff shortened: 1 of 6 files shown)


app/models/participant_declaration.rb
```diff
@@ -21,6 +21,7 @@ class ParticipantDeclaration < ApplicationRecord
   scope :npq, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.npq_profiles) }
   scope :payable, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.where(payable: true)) }
   scope :unique_id, -> { select(:user_id).distinct }
+  scope :for_participant, ->(*args) { where(user_id: args[0]) if args.any? }
```
kicferk1, line 24: I don't think this is the place to check that args exist. I think that should be done in the controller, and this scope should require a user id.

kicferk1 (commented):
In participants and npq applications we have something similar - namely, an `updated_since` filter parameter. Can we follow the convention they are using, and stick the parameters we filter the response by into `filter` parameter? That way, if (when) we add more filters, we can stick them there.

Also, this needs swagger documentation.

app/models/participant_declaration.rb
plisovin: ok, made it similar to npq application scope

spec/docs/participant_declarations_spec.rb
```diff
@@ -101,6 +101,18 @@
       tags "participants declarations"
       security [bearerAuth: []]
 
+      parameter name: :filter,
+                in: :query,
+                schema: {
+                  "$ref": "#/components/schemas/ListFilter",
```
kicferk1, line 107: That component is 
```
description: "Filter a list of records to return more specific results"
type: object
properties:
  updated_since:
    description: "Return only records that have been updated since this date and time (ISO 8601 date format)"
    type: :string
    example: "2021-05-13T11:21:55Z"
```
Which doesn't really match the filter in our case. Can we add participant_id field to it?

swagger/v1/component_schemas/ListFilter.yml
```diff
@@ -5,3 +5,7 @@ properties:
     description: "Return only records that have been updated since this date and time (ISO 8601 date format)"
     type: :string
     example: "2021-05-13T11:21:55Z"
+  participant_id:
+    description: "The unique id of the participant"
+    type: :string
+    example: "ab3a7848-1208-7679-942a-b4a70eed400a"
```
kicferk1, line 11: Hmm, perhaps at some point we should split it, cause maybe not all of our filters will make sense for all the endpoints, but it's probably ok for now.

kicferk1 (approved):

PR #953, opened 2021-09-03
CPDTP-428: Csv participant declarations

kicferk1:
### Context
We want to support CSV formats to make ourselves as flexible to the providers importing data as possible.
https://dfedigital.atlassian.net/browse/CPDTP-428

Add an csv format support to existing participant declarations index endpoint.
Pull 'to_csv' method into its own module to share it between controllers.
Specify the format of declaration date in serialiser explicitly to avoid `to_s` in csv computation returning a different format.
Add documentation and tests.
Fix the githook saying `/components/schemas/XXXXXX` is a bloody AWS key. 

### Guidance to review
I imagine it will have some conflicts with https://github.com/DFE-Digital/early-careers-framework/pull/952, but they shouldn't be too terrible. 

### Testing
Added request specs for the csv index endpoint

Hit the `/api/v1/participant-declarations.csv` endpoint with the right token.

### DIFF
--- a/app/controllers/concerns/api_csv.rb
+++ b/app/controllers/concerns/api_csv.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 module ApiCsv
-  extend ActiveSupport::Concern
-
 private
 
   def to_csv(hash)
--- a/app/controllers/concerns/api_csv.rb
+++ b/app/controllers/concerns/api_csv.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module ApiCsv
+  extend ActiveSupport::Concern
+
+private
+
+  def to_csv(hash)
+    return "" if hash[:data].empty?
+
+    headers = %w[id]
+    attributes = hash[:data].first[:attributes].keys
+    headers.concat(attributes.map(&:to_s))
+    CSV.generate(headers: headers, write_headers: true) do |csv|
+      hash[:data].each do |item|
+        row = [item[:id]]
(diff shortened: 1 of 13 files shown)


app/controllers/concerns/api_csv.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module ApiCsv
+  extend ActiveSupport::Concern
```
blackrat, line 4: Don't need to extend Concern unless you're using `included` `prepended` or one of the other methods that it adds. This is just a plain ol' grouping module.

swagger/v1/component_schemas/MultipleParticipantDeclarationsCsvResponse.yml
```diff
@@ -0,0 +1,14 @@
+description: "A list of participant declarations in the Comma Separated Value (CSV) format"
+type: string
+format: base64
```
blackrat, line 3: Is it?

app/controllers/concerns/api_csv.rb
kicferk1: Good spot :)

swagger/v1/component_schemas/MultipleParticipantDeclarationsCsvResponse.yml
kicferk1: No it ain't.
I need to purge that from a few other places too...

blackrat (approved):

PR #956, opened 2021-09-03
CPDRP-908 Add mailer for ECTs and mentors who have not yet added validation info

jamesrose:
### Context

This PR sends an email to ECTs and mentors who have not yet added validation information, irrespective of whether they are in the beta or not.

Template: https://www.notifications.service.gov.uk/services/d1207ebf-ac0c-47b8-b1bf-16dc899e0923/templates/50ee41e5-06b9-41cf-9afd-d5bc4db356c4

Things to check:

1. Is the query correct?
2. Are the email template params correct?

### DIFF
--- a/app/services/validation_beta_service.rb
+++ b/app/services/validation_beta_service.rb
@@ -30,6 +30,15 @@ class ValidationBetaService
     end
   end
 
+  def tell_ects_and_mentors_to_add_validation_information
+    ParticipantProfile::ECF
+      .includes(:ecf_participant_eligibility, :ecf_participant_validation_data, :school_cohort, :school)
+      .where(school_cohort: { cohort: Cohort.current }) # Chosen program
+      .where(ecf_participant_eligibility: { participant_profile_id: nil })
+      .or(ParticipantProfile::ECF.where(ecf_participant_validation_data: { participant_profile_id: nil }))
+      .find_each { |profile| send_we_need_information_for_your_programme(profile, profile.school) }
+  end
+
 private
 
   def participant_validation_start_url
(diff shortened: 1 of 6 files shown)


spec/mailers/participant_validation_mailer_spec.rb
```diff
@@ -131,6 +131,21 @@
     end
   end
 
+  describe "#we_need_information_for_your_programme_email" do
+    let(:induction_coordinator_email) do
+      described_class.we_need_information_for_your_programme_email(
+        recipient: recipient,
+        school_name: school_name,
+        participant_start: "example.com/start-validation",
+      )
```
kicferk1, line 140: This needs to be `start_url` rather than `participant_start`.

app/services/validation_beta_service.rb
```diff
@@ -30,6 +30,15 @@ def tell_induction_coordinators_to_check_ect_and_mentor_information
     end
   end
 
+  def tell_ects_and_mentors_to_add_validation_information
+    ParticipantProfile::ECF
+      .includes(:ecf_participant_eligibility, :ecf_participant_validation_data, :school_cohort, :school)
+      .where(school_cohort: { cohort: Cohort.current }) # Chosen program
+      .where(ecf_participant_eligibility: { participant_profile_id: nil })
+      .or(ParticipantProfile::ECF.where(ecf_participant_validation_data: { participant_profile_id: nil }))
```
kicferk1, line 38: That seems a bit weird - why do we say 'where the participant eligibility has no profile', rather than 'where there is no participant eligibility? 

```
      .where(ecf_participant_eligibility: nil)
      .or(ParticipantProfile::ECF.where(ecf_participant_validation_data: nil))
```

spec/mailers/participant_validation_mailer_spec.rb
jamesrose: Cheers

app/services/validation_beta_service.rb
jamesrose: These are both associations and we want to check that there are either zero `ecf_participant_eligibility` or zero `ecf_participant_validation_data` associated.

kicferk1: Yeah, that part makes sense. What feels weird is 
```
.where(ecf_participant_eligibility: { participant_profile_id: nil })
```
vs
```
.where(ecf_participant_eligibility: nil)
```
Are we not be able to do the latter because `ecf_participant_eligibility` is a `has_one` association?

kicferk1 (approved):

app/services/validation_beta_service.rb
jamesrose: Ah, yes, will return to it. I thought it was a `has_many`.

jamesrose: @kicferk1 I tried this in the console and I don't think that works.

kicferk1: Ah well, too bad. We have tests that look convincing, so it should behave fine :)

PR #958, opened 2021-09-03
CPDRP-908 Add email to tell SITs we have asked ECTs and mentors for info

jamesrose:
### Context

Send another mailer to tell SITs that we have asked their ECTs and mentors for validation information. This largely uses the same criteria as https://github.com/DFE-Digital/early-careers-framework/pull/938 but because we needed a school name I had to make sure that we weren't sending to the same SIT for multiple schools. Other than that, the criteria and tests are the same.

Notify template: https://www.notifications.service.gov.uk/services/d1207ebf-ac0c-47b8-b1bf-16dc899e0923/templates/d560fb2e-243d-48b1-bf61-c7e111f56858

### DIFF
--- a/app/services/validation_beta_service.rb
+++ b/app/services/validation_beta_service.rb
@@ -30,6 +30,17 @@ class ValidationBetaService
     end
   end
 
+  def tell_induction_coordinators_we_asked_ects_and_mentors_for_information
+    InductionCoordinatorProfile.find_each do |ic|
+      ic.schools.not_opted_out.each do |school|
+        if chosen_programme_and_not_in_beta(school)
+          send_asked_ects_and_mentors_for_information(ic, school)
+          break
+        end
+      end
+    end
+  end
+
   def tell_ects_and_mentors_to_add_validation_information
     ParticipantProfile::ECF
       .includes(:ecf_participant_eligibility, :ecf_participant_validation_data, :school_cohort, :school)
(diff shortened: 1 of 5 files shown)


app/services/validation_beta_service.rb
```diff
@@ -30,6 +30,17 @@ def tell_induction_coordinators_to_check_ect_and_mentor_information
...
+  def tell_induction_coordinators_we_asked_ects_and_mentors_for_information
+    InductionCoordinatorProfile.find_each do |ic|
+      ic.schools.not_opted_out.each do |school|
+        if chosen_programme_and_not_in_beta(school)
+          send_asked_ects_and_mentors_for_information(ic, school)
+          break
+        end
+      end
+    end
+  end
```
kicferk1, line 42: Will this send at most one email per IC? (due to `break`)

If so, is it a good idea to include school name? 

```
  def tell_induction_coordinators_to_check_ect_and_mentor_information
    InductionCoordinatorProfile.find_each do |ic|
      if ic.schools.not_opted_out.any? { |school| chosen_programme_and_not_in_beta(school) }
        send_check_ect_and_mentor(ic)
      end
    end
  end
```
seems to be doing something very similar, and because it does not need school name, it is more functional.

Also, just in general, if we include the school name in the email, should we not send one email per school? If an IC has two schools, and they get one email with one school name, will they get confused and assume the other school is irrelevant?

jamesrose: I agree — I'll return to Shani about this.

jamesrose: My code intention was to send to one per IC because multiple didn't make sense to me. We are essentially just asking them to sign in. I'll confirm and update if neccessary.

kicferk1 (approved):

PR #959, opened 2021-09-03
Push NPQ accept/reject stats to BigQuery

asmega:
### Context

- We need to be able to reporting for the accept/reject on NPQ validations

- Add the ability to push data into BigQuery in a streaming append only fashion
- On `NPQValidationData` creation and updating of status stream this data to BigQuery asynchronously

### Guidance to review

- I'm aware of the existence of https://github.com/DFE-Digital/early-careers-framework/pull/941, however it has not been merged yet

### Testing

- Test in review app

### Review Checks

- `cf ssh ecf-review-pr-959`
- start a rails console
- create an application with:
```
record = NPQValidationData.create!(user: User.all.sample, npq_lead_provider: NPQLeadProvider.all.sample, npq_course: NPQCourse.all.sample)
```
- should see `pending` row added to `enrollments_deployed_development` in BigQuery
- update status with:
```
record.update(lead_provider_approval_status: "rejected")
```
- should now see additional row with status `rejected`


.github/workflows/deploy_to_dev.yml
```diff
@@ -96,4 +96,11 @@ jobs:
           export TF_VAR_paas_app_docker_image=dfedigital/early-careers-framework-prod:${{ github.sha }}
           cd terraform/app
           terraform init -reconfigure -input=false -backend-config="bucket=paas-s3-broker-prod-lon-7f2ca242-9929-4662-a79c-c454ea56ea7b"
-          terraform apply -input=false -auto-approve -var-file ../workspace-variables/dev.tfvars -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}"}' -var 'logstash_url=${{secrets.SYSLOG_DRAIN_URL}}' -var 'docker_username=${{ secrets.DOCKER_USERNAME }}' -var 'docker_password=${{ secrets.DOCKER_DEV_PASSWORD }}'
+          terraform apply \
+            -input=false \
+            -auto-approve \
+            -var-file ../workspace-variables/dev.tfvars \
+            -var='secret_paas_app_env_values={"RAILS_MASTER_KEY":"${{secrets.RAILS_MASTER_KEY_DEV}}","GOOGLE_BQ_SERVICE_ACCOUNT_BASE64":"${{secrets.GOOGLE_BQ_SERVICE_ACCOUNT_BASE64}}"}' \
```
kicferk1, line 103: So, `GOOGLE_BQ_SERVICE_ACCOUNT_BASE64` is a secret that is only used inside the app?

So in principle, it could be put inside Rails credentials, and be accessible using `RAILS_MASTER_KEY` for a particular environment, right?

spec/models/npq_validation_data_spec.rb
```diff
@@ -11,4 +11,26 @@
...
+    context "when lead_provider_approval_status is modified" do
+      it "fires NPQ::StreamBigQueryEnrollmentJob" do
+        subject
+
+        expect {
+          subject.update(lead_provider_approval_status: "accepted")
+        }.to change(Delayed::Job, :count).by(1)
+      end
+    end
+  end
```
kicferk1, line 35: Can we have a test showing when the BigQuery callback is not fired?

.github/workflows/deploy_to_dev.yml
asmega: Sure I can change. Although I'm not a fan of this. I believe the application code should be "configuationless" or "stateless" with these things being injected in. Otherwise code changes are needed for something which should be a configuration change. Now this has been blurred and less clean cut.

kicferk1: Well, I think the other side of this spectrum is "if our secrets are held in github, then we need like 4 copies of them per secret, and our github gets littered very quickly". Which is not quite true here, cause we use the same secret for all envs.

I don't really mind which way it goes.

spec/models/npq_validation_data_spec.rb
asmega: I've now added a test for this

kicferk1 (approved):

.github/workflows/deploy_to_dev.yml
asmega: Well we're using it anyway, in for a penny in for a pound.

asmega:
sorry i have to close and re-open as the review changes affected the review app and no longer works as expected

PR #962, opened 2021-09-06
CPDRP-868 CIP only invites

FriedSock:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDRP-868

- Remove redirect to cip-only information page in the nomination flow.
- Add email for sending invitations to cip-only schools (excluding welsh schools)

### Review Checks


app/controllers/nominations/request_nomination_invite_controller.rb
```diff
@@ -30,9 +30,7 @@ def receive_school
 
     session[:nomination_request_form] = @nomination_request_form.serializable_hash
 
-    if @nomination_request_form.school.cip_only?
-      redirect_to cip_only_request_nomination_invite_path
-    elsif !@nomination_request_form.school.eligible?
+    if !@nomination_request_form.school.eligible? && !@nomination_request_form.school.cip_only?
```
kicferk1, line 33: Sorry, what's the intention for `cip_only` schools here?

That it behaves as if the school is eligible?

Maybe we should add a method like `can_access_service` on a school, that combines `eligible` and `cip_only`?

app/services/invite_schools.rb
```diff
@@ -230,6 +230,25 @@ def send_year2020_invite_email
     end
   end
 
+  def invite_cip_only_schools
+    # We want to exclude sending invites to welsh schools (code 30), which is the only other cip only
+    # type code in GiasTypes::CIP_ONLY_TYPE_CODES
+    School.currently_open.where(school_type_code: [10, 11, 37]).where(section_41_approved: false).find_each do |school|
```
kicferk1, line 236: Can we put that in `GiasHelpers`? 
At least the codes, but perhaps also the general scope, which means `cip_only_except_welsh`

app/controllers/nominations/request_nomination_invite_controller.rb
FriedSock: Yes that is the intention

app/services/invite_schools.rb
FriedSock: Ok, the reason I didn't do that is because its going into an email that we send once (we should be able to delete this code after we send the invites. But I can do it if you prefer

kicferk1: Yeah, that makes sense, but I really don't want to see school type codes anywhere outside of closed and isolated modules, cause they are 'magic numbers'.

Feel free to add `TODO` comments around the code that tells us to delete it when we no longer need it :)

kicferk1 (approved):

PR #964, opened 2021-09-06
Reduce needless touching of participants

kicferk1:
### Context
This morning a provider asked us why we updated a lot of participant records on sandbox around 6am this morning.

Well... because our gias import job ran at 6am, it saved all schools (most with no changes), which triggered `after_commit` hook for ALL schools. Which touched all participants for all GIAS schools.

1. Remove touching participants when school is updated. I am not sure we should update participants of school when school name changes.
2. Reduce touching participants on school cohort. Here I think we need to update them, because school cohort has a lot of things that matter for participants. Say, what induction programme they are following.

### Guidance to review
I can alter point 1 to be more like point 2. If we really want to. I just am not sure we really want to. 

### Testing
Unit tests documenting new behaviour.

### DIFF
--- a/spec/models/partnership_spec.rb
+++ b/spec/models/partnership_spec.rb
@@ -25,7 +25,8 @@ RSpec.describe Partnership, type: :model do
     user = profile.user
     user.update!(updated_at: 2.weeks.ago)
 
-    partnership.touch
+    partnership.update!(updated_at: 2.weeks.ago)
+
     expect(user.reload.updated_at).to be_within(1.second).of Time.zone.now
     expect(profile.reload.updated_at).to be_within(1.second).of Time.zone.now
   end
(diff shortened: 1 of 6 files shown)


spec/models/partnership_spec.rb
```diff
@@ -25,7 +25,8 @@
     user = profile.user
     user.update!(updated_at: 2.weeks.ago)
 
-    partnership.touch
+    partnership.update!(updated_at: 2.weeks.ago)
```
asmega, line 28: random fyi, touch can take arguments eg
```
partnership.touch(time: 2.weeks.ago)
```
don't think you need to change though

asmega (approved):

spec/models/partnership_spec.rb
kicferk1: Good to know!

jamesrose (approved):

PR #972, opened 2021-09-07
only install production yarn dependencies

asmega:
## Ticket and context

Ticket: n/a

- We are installing yarn dev dependencies into our docker images
- These aren't needed as far as I can tell as the test suite does not go through our built images
- So it's just cluttering up the produced image
- From sample size of one test run, it has shaved off about 2 minutes of the deploy to review app time 

- see inline comments

### Code quality checks

### HTML Checks

### Gotchas

- visit https://ecf-review-pr-972.london.cloudapps.digital
- app should function as normal

 ### External API changes

- none

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN apk -U upgrade && \
     bundle config set no-cache 'true' && \
     bundle config set no-binstubs 'true' && \
     bundle --retry=5 --jobs=4 --without=development test && \
-    yarn install --check-files && \
+    yarn install --check-files --production && \
     apk del .gem-installdeps && \
     rm -rf /usr/local/bundle/cache && \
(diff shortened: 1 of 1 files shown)


kicferk1 (approved):

Dockerfile
```diff
@@ -55,8 +55,7 @@ ENV GOVUK_APP_DOMAIN="http://localhost:3000" \
 WORKDIR /app
 COPY . .
 
-RUN yarn jest --passWithNoTests && \
```
asmega, line 58: I'm not sure why this `jest` line is here?

kicferk1: I think in ancient days we were not sure how we would test front-end, and jest might have come from whichever place we copy-pasted the dockerfile from :P

PR #973, opened 2021-09-07
Removed participant_validation feature flag

BroiSatse:
Removing :participant_validation feature flag!


app/services/validation_beta_service.rb
```diff
@@ -1,46 +1,6 @@
...
-
-  def tell_induction_coordinators_to_check_ect_and_mentor_information
-    InductionCoordinatorProfile.find_each do |ic|
-      if ic.schools.not_opted_out.any? { |school| chosen_programme_and_not_in_beta(school) }
-        send_check_ect_and_mentor(ic)
-      end
-    end
-  end
-
-  def tell_induction_coordinators_we_asked_ects_and_mentors_for_information
```
jamesrose, line 33: We need this one - it's #10 on the schedule, we haven't sent it yet.

config/routes.rb
```diff
@@ -254,7 +254,7 @@
     resource :no_access, only: :show, controller: "no_access"
     resource :start_registrations, path: "/start-registration", only: :show
 
-    authenticated :user, ->(user) { FeatureFlag.active?(:participant_validation, for: user.teacher_profile&.participant_profiles&.ecf&.active_record&.first&.school) } do
+    authenticated :user do
```
tonyheadford, line 257: Think this can go as it was only needed to get the context for the feature flag check

jamesrose (approved):
Nice one 🚢

config/routes.rb
BroiSatse: Ah nice. Also, we have a multistep form for this kind of thing, I'll try to change it later on. :)

PR #974, opened 2021-09-07
Standardise updated_since filter processing

kicferk1:
## Ticket and context

A provider asked us why a format of a filter works on NPQ applications, but not on participants.
IMO they should work everywhere. And it does not hurt us to make the date format range a little wider.

I added the test to participants api - I guess technically I could add tests to all places using the filter, but since it is a shared functionality, should I really?

### Code quality checks

### HTML Checks

### Gotchas

2. Curl `/api/v1/participants.json?filter[updated_since]=1970-01-01T00%3A00%3A00%2B01%3A00&page[page]=1&page[per_page]=100` with proper authorisation (say, ambition-token).
3. You should get 200 response and a list of values.

## External API changes

Does this change anything in our external APIs? If so, did you remember to update the CHANGELOG for Lead Providers? (lead-providers/guidance/release-notes) - Yes

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -8,6 +8,7 @@ module Api
     rescue_from ActionController::ParameterMissing, with: :missing_parameter_response
     rescue_from ActionController::BadRequest, with: :bad_request_response
     rescue_from ActiveRecord::StatementInvalid, with: :bad_request_response
+    rescue_from ArgumentError, with: :bad_request_response
 
   private
 
--- a/app/controllers/concerns/api_filter.rb
+++ b/app/controllers/concerns/api_filter.rb
@@ -8,9 +8,9 @@ private
   end
 
   def updated_since
-    return unless filter[:updated_since]
+    return if filter[:updated_since].blank?
 
-    Time.iso8601(filter[:updated_since]) if filter[:updated_since].present?
+    Time.iso8601(filter[:updated_since])
   rescue ArgumentError
     Time.iso8601(URI.decode_www_form_component(filter[:updated_since]))
   end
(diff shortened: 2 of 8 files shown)


app/controllers/concerns/api_filter.rb
```diff
@@ -0,0 +1,17 @@
...
+private
+
+  def filter
+    params[:filter] ||= {}
+  end
+
+  def updated_since
+    return unless filter[:updated_since]
+
+    Time.iso8601(filter[:updated_since]) if filter[:updated_since].present?
```
asmega, line 13:
```suggestion
    return if filter[:updated_since].blank?

    Time.iso8601(filter[:updated_since])
```

app/controllers/api/api_controller.rb
```diff
@@ -8,6 +8,7 @@ class ApiController < ActionController::API
     rescue_from ActionController::ParameterMissing, with: :missing_parameter_response
     rescue_from ActionController::BadRequest, with: :bad_request_response
     rescue_from ActiveRecord::StatementInvalid, with: :bad_request_response
+    rescue_from ArgumentError, with: :bad_request_response
```
asmega, line 11: is this line covered? I ask because I was recently here and have doubts about `Api::ParamErrorFactory` and would rather have coverage

kicferk1: When I remove it 
```
          context "when updated_since in an invalid format" do
            it "returns a 400 status" do
              get "/api/v1/participants", params: { filter: { updated_since: "23rm21" } }
              expect(response.status).to eq 400
            end
          end
```
fails

tonyheadford: I added that spec in the other day (#957) - the issue appeared to be that the `updated_since` was URI encoded and when passed to the query raised the `ActiveRecord::StatementInvalid` error. So I added the line to unencode `updated_since` as well as that `rescue_from`. If things have shifted you might now be raising a different error.

asmega (approved):

PR #975, opened 2021-09-07
add filter parameter check on npq applications api

asmega:
## Ticket and context

Ticket: https://trello.com/c/pTTNAH4t/466-ecf-api-handle-incorrect-input-for-filters

- An error is raised when api users pass incorrect filter to npq applications api endpoint
- This should now no longer error
- It should now return meaningful error message to the api user

### Code quality checks

### HTML Checks

### Gotchas

- create a private api token
```
LeadProviderApiToken.create_with_random_token!(cpd_lead_provider: CpdLeadProvider.all.sample)
```
- make malformed request
```
curl -H "Authorization: Bearer TOKEN" "https://ecf-review-pr-975.london.cloudapps.digital/api/v1/npq-applications?filter=asd"
```
- expect:
```
{"errors":[{"title":"Bad parameter","detail":"Filter must be a hash"}]}
```

## External API changes

- Although this does affect the external api I don't feel it needs to be documented on the specific endpoint
- Do we have general documentation around errors? if not I assume we should add some outside the scope of this change

### DIFF
--- a/app/controllers/concerns/api_filter.rb
+++ b/app/controllers/concerns/api_filter.rb
@@ -1,7 +1,20 @@
 # frozen_string_literal: true
 
 module ApiFilter
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :validate_filter_param
+  end
+
 private
+
+  def validate_filter_param
+    unless filter.as_json.is_a?(Hash)
+      error_factory = Api::ParamErrorFactory.new(params: ["Filter must be a hash"], error: "Bad parameter")
+      render json: { errors: error_factory.call }, status: :bad_request
+    end
+  end
 
   def filter
     params[:filter] ||= {}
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -59,6 +61,13 @@ def filter
         params[:filter] ||= {}
       end
 
+      def validate_filter_param
+        unless filter.as_json.is_a?(Hash)
+          error_factory = Api::ParamErrorFactory.new(params: ["Filter must be a hash"], error: "Bad parameter")
+          render json: { errors: error_factory.call }, status: :bad_request
+        end
+      end
```
kicferk1, line 69: Should we have this check in `ApiFilter`? Perhaps when we try to access the filter we could check its format, and raise an error that will get caught by api controller error handling?

That way we can add this check to every place that is using `filter` url parameter in one go.

asmega: Yeah, already on it. That's why I waited for your stuff to go in first. I've tweaked and rebased

kicferk1 (approved):

PR #976, opened 2021-09-07
CPDRP-849 Update schools from GIAS data

tonyheadford:
## Ticket and context

Ticket: [Jira](https://dfedigital.atlassian.net/browse/CPDRP-849)
Automatically refresh GIAS data and update live schools

## Tech review
There are 2 parts to this:

1. An update to the GIAS update process to identify and handle simple changes to school data, and create `DataStage::SchoolChange` records for complex changes. It also replaces the `SchoolDataImporter` at this point, updating the live schools.

2. ~~A handler `DataStage::ProcessSchoolChanges` for `DataStage::SchoolChange` in the case where a school has closed and reopened e.g. an academy conversion. This results in a new school. In cases where the is an existing predecessor, migrate the assets from the predecessor to the new school.  This is a bit scary so at the moment the implementation is there but it is not automatically wired up. Once we've confirmed that it is doing the right thing then this could be added as the final step in the `ImportGiasDataJob`~~

***Part 2 will now be in a separate PR***

### Code quality checks

### HTML Checks

### Gotchas

## External API changes


app/controllers/admin/gias/schools_to_add_controller.rb
```diff
@@ -18,6 +18,11 @@ def show
       def schools_scope
         policy_scope(DataStage::School, policy_scope_class: GiasPolicy::Scope)
       end
+
+      def fetch_schools_to_open
+        (schools_scope.schools_to_add + schools_scope.schools_to_open)
+          .sort { |a, b| a.urn.to_i <=> b.urn.to_i }
```
kicferk1, line 24: Will `sort_by { |school school.urn.to_i }` do the trick too?

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,90 @@
...
+      ukprn
+    ].freeze
+
+    def call
+      DataStage::SchoolChange.unhandled.status_changed.find_each do |change|
+        next if has_unhandled_attributes?(change.attribute_changes)
+
+        if change.attribute_changes.key? "school_status_code"
+          handle_status_code_change(change)
+        end
```
kicferk1, line 22: Should we handle status code change before we skip?

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,90 @@
...
+    def handle_status_code_change(change)
+      from_code = change.attribute_changes["school_status_code"]&.first
+      to_code = change.attribute_changes["school_status_code"]&.last
+
+      if open_status_code?(from_code) && !open_status_code?(to_code)
+        close_school(change)
+      elsif !open_status_code?(from_code) && open_status_code?(to_code)
+        open_school(change)
+      else
+        # just sync then
```
kicferk1, line 41: That comment does not seem necessary

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,90 @@
...
+        change.school.create_or_sync_counterpart!
+      end
+    end
+
+    def open_school(change)
+      school = change.school
+
+      # is there a predecessor
+      last_link = school.school_links.order(created_at: :asc).last
```
kicferk1, line 50: If we call the variable `predecessor` then the comment is unnecessary, right?

kicferk1: Unless the comment is lying, and the actual check for whether there is a predecessor is
```
last_link.link_type == "Predecessor"
```
line.

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,90 @@
...
+          # transfer assets if predecessor
+          if last_link.present?
+            predecessor = ::School.find_by(urn: last_link.link_urn)
+            transfer_assets_from_predecessor(school: school.counterpart, predecessor: predecessor) if predecessor.present?
+          end
+
+          change.update!(handled: true)
+        end
+      end
+    end
```
kicferk1, line 66: Can we handle `last_link` being `nil` in a separate branch from it being `predecessor`, rather than intertwining them?

Also, what's the difference between `DataStage::School.transaction` and `ActiveRecord::Base.transaction`? Should we be using the latter, given that asset changes deal with non-school records?

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    def close_school(change)
+      # let any related school open transfer any assets
+      # just sync attribute changes
+      DataStage::School.transaction do
+        # only sync changes don't create
+        change.school.create_or_sync_counterpart! if change.school.counterpart.present?
+        change.update!(handled: true)
+      end
+    end
```
kicferk1, line 76: What does `let any related school open transfer any assets` mean?

Also, why do we check if the counterpart is present here? `create_or_sync_counterpart` does it for us (I think)

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    def transfer_assets_from_predecessor(school:, predecessor:)
+      # move everything to the new school
```
kicferk1, line 79: I don't think this code comment is particularly useful here.

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,90 @@
...
+      Partnership.active.where(school: predecessor).each { |partnership| partnership.update!(school: school) }
+      # TODO: What should we do here if the school already has a school_cohort (if anything)?
```
kicferk1, line 81: That sounds like a pain, because what if the school cohorts have different induction choices on them? 😢 

I think if the school cohorts are for a different cohort we should be fine to just move it.

If the cohort is the same, and if existing school cohort is 'almost identical' to predecessor school cohort, we can probably 'merge' them - just move everything pointing at old school cohort into new school cohort and delete the old one.

But if they have the same cohort, and they are different (say, different induction choice) then we are in trouble.

This actually is also a problem with partnerhsips - if a school has an active partnership, and its predecessor has an actie partnership with a different provider or delivery partner, I don't know if we should just transfer the partnership to new school :/

app/services/data_stage/process_school_changes.rb
tonyheadford: Not at the moment as the status code change is the only change we want to handle, we don't want to do that if there are other major changes too (don't understand what's needed for those yet) because we'd then mark the change as being handled. I've got some discussions with the business about what we should do planned.

app/services/data_stage/process_school_changes.rb
tonyheadford: Yes missing a word/grammar in the comment, it should be "let any related 'school open' change handle the transfer of assets"
In this case we don't want to create a counterpart if it doesn't exist as we're closing the school - the `create_or_sync_counterpart!` will created it if it doesn't exist.

app/services/data_stage/process_school_changes.rb
tonyheadford: Yes, this is the stuff that we need to understand. The `ProcessSchoolChanges` service isn't actually called from anywhere yet, it's a placeholder for discussion.

app/services/data_stage/process_school_changes.rb
tonyheadford: I don't think there is any difference but I like to use the model that things are relating to, but you're right it should probably be `::School.transaction` as the assets are related to the live school not the staged one

tonyheadford: I want them to be wrapped in the transaction so that the `change.update!` is only done if the create/sync/transfer are done - not sure if we'd just be less DRY if I split this up?

app/services/data_stage/process_school_changes.rb
tonyheadford: `predecessor` is used later but fair point about the comment

app/services/data_stage/process_school_changes.rb
tonyheadford: At the moment I'm only trying to handle the case where a school closes and reopens, (where there is a `predecessor` but not one from a merger or split).  In this case there shouldn't be any choices made on the school as we've just created it - but there may be if it's been added/updated already manually.

app/services/data_stage/process_school_changes.rb
tonyheadford: Moving this class to a separate PR

app/services/data_stage/process_school_changes.rb
tonyheadford: Moving this class to a separate PR

app/services/data_stage/process_school_changes.rb
tonyheadford: Moving this class to a separate PR

app/services/data_stage/process_school_changes.rb
tonyheadford: Moving this class to a separate PR

app/services/data_stage/process_school_changes.rb
tonyheadford: Moving this class to a separate PR

app/services/data_stage/process_school_changes.rb
tonyheadford: Moving this class to a separate PR

app/models/concerns/gias_helpers.rb
```diff
@@ -24,12 +24,14 @@ def name_and_urn
...
 
-private
-
   def open?
     open_status_code?(school_status_code)
   end
 
+  def closed?
+    !open_status_code?(school_status_code)
+  end
```
kicferk1, line 33:
```
  def closed?
    !open
  end
```
?

kicferk1 (approved):

app/models/concerns/gias_helpers.rb
tonyheadford: Yes, then I thought that's just an extra call .... :)

PR #977, opened 2021-09-08
Set a minimum number of requests captured on staging before we raise an alert

kicferk1:
## Ticket and context
I got annoyed by "high rate of failing requests" notification on staging. They seem to be 404s by bots, they happen every now and then in low volumes. But staging is just not ussed for much, so they compose a high percentage of requests, which triggers our alarms.

https://www.robustperception.io/combining-alert-conditions is where I got the `and` from.

### DIFF
--- a/terraform/monitoring/config/staging/alert.rules.yml
+++ b/terraform/monitoring/config/staging/alert.rules.yml
@@ -3,7 +3,7 @@ groups:
     rules:
       - alert: RequestsFailuresElevated
         # Condition for alerting
-        expr: ((sum(rate(requests{app="ecf-staging", status_range=~"0xx|4xx|5xx"}[5m]))) / (sum(rate(requests{app="ecf-staging"}[5m])) ) > 0.1) and ((sum(rate(requests{app="ecf-staging"}[5m])) > 20)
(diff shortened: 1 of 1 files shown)


terraform/monitoring/config/staging/alert.rules.yml
```diff
@@ -3,7 +3,7 @@ groups:
     rules:
       - alert: RequestsFailuresElevated
         # Condition for alerting
-        expr: (sum(rate(requests{app="ecf-staging", status_range=~"0xx|4xx|5xx"}[5m]))) / (sum(rate(requests{app="ecf-staging"}[5m])) ) > 0.1
+        expr: ((sum(rate(requests{app="ecf-staging", status_range=~"0xx|4xx|5xx"}[5m]))) / (sum(rate(requests{app="ecf-staging"}[5m])) ) > 0.1) and ((sum(rate(requests{app="ecf-staging"}[5m])) > 20)
```
jamesrose, line 6: Could just exclude 4xx here from staging?

kicferk1: That would work too

FriedSock (approved):

jamesrose (approved):

PR #981, opened 2021-09-08
CPDTP-434 Reject different date duplicates

plisovin:
## Ticket and context

When a participant declaration has been created, any subsequent duplicate declaration for this participant with any different declaration date will be rejected with the 409 status code.

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -9,6 +9,7 @@ module Api
     rescue_from ActionController::BadRequest, with: :bad_request_response
     rescue_from ActiveRecord::StatementInvalid, with: :bad_request_response
     rescue_from ArgumentError, with: :bad_request_response
+    rescue_from ActiveRecord::RecordNotUnique, with: :conflict_response
 
   private
 
@@ -26,6 +27,10 @@ module Api
 
     def bad_request_response(exception)
       render json: { errors: Api::ParamErrorFactory.new(error: "Bad request", params: exception.message).call }, status: :bad_request
+    end
+
+    def conflict_response(exception)
+      render json: { errors: Api::ParamErrorFactory.new(error: "Conflict", params: exception.message).call }, status: :conflict
     end
   end
 end
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -9,7 +9,7 @@ module Api
     rescue_from ActionController::BadRequest, with: :bad_request_response
     rescue_from ActiveRecord::StatementInvalid, with: :bad_request_response
     rescue_from ArgumentError, with: :bad_request_response
-    rescue_from ActiveRecord::RecordNotUnique, with: :conflict_response
+    rescue_from ActiveRecord::RecordNotUnique, with: :bad_request_response
 
   private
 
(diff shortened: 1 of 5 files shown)


app/services/record_declarations/base.rb
```diff
@@ -86,6 +89,17 @@ def find_or_create_record!
       end
     end
 
+    def record_exist?
+      declaration = self.class.declaration_model.find_by(
+        user: user,
+        course_identifier: course_identifier,
+        declaration_type: declaration_type,
+        cpd_lead_provider: cpd_lead_provider,
+      )
```
kicferk1, line 98: I would say the cpd_lead_provider does not need to be there either - if user switches LPs, and has a started declaration, we should prevent the new LP from submitting another started declaration.

app/services/record_declarations/base.rb
```diff
@@ -39,7 +39,10 @@ def call
       validate_provider!
       validate_milestone!
 
+      raise ActiveRecord::RecordNotUnique, "Declaration with given participant ID already exists" if record_exist?
+
       declaration = find_or_create_record!
+
```
kicferk1, line 45: Should we do something like:
```
declaration = find_or_create_record! # make it more lenient - find by things we use in record_exists?
validate_declaration_uniqueness!(declaration)
...

def validate_declaration_uniqueness!(declaration)
  # check date
  # check provider
  # check anything else that we want to check
  # raise if the checks fail
end
```

app/services/record_declarations/base.rb
```diff
@@ -86,6 +89,17 @@ def find_or_create_record!
       end
     end
 
+    def record_exist?
```
kicferk1, line 92: What this really means is `record_exists_and_does_not_match_incoming_request?`, or something to that effect.
I would prefer to relax `find_or_create_record` query, and have a validation done after we get the record.

spec/shared/context/service_record_declaration_params.rb
```diff
@@ -14,6 +14,9 @@
   let(:ect_params) do
     params.merge({ cpd_lead_provider: cpd_lead_provider })
   end
+  let(:ect_params_with_different_date) do
+    ect_params.merge({ declaration_date: (ect_declaration_date + 1.second).rfc3339 })
+  end
```
kicferk1, line 19: Does this really belong in shared context? We seem to be using it just once.

app/services/record_declarations/base.rb
plisovin: The reason I think it should be done before the declaration creation as otherwise we end up with duplicate declarations after ```find_or_create_record``` but in the ticket requirement we don't want to do that. What do you think?

spec/shared/context/service_record_declaration_params.rb
plisovin: Moved to the test in 9a99d84

app/services/record_declarations/base.rb
plisovin: renamed in e3ed1fe

kicferk1 (approved):

kicferk1 (approved):

app/controllers/api/api_controller.rb
```diff
@@ -9,6 +9,7 @@ class ApiController < ActionController::API
     rescue_from ActionController::BadRequest, with: :bad_request_response
     rescue_from ActiveRecord::StatementInvalid, with: :bad_request_response
     rescue_from ArgumentError, with: :bad_request_response
+    rescue_from ActiveRecord::RecordNotUnique, with: :conflict_response
```
mattonfoot, line 12: do we need a new response type here will `:bad_request_response` not be suitable?

plisovin: I don't mind, any other gov projects use ```conflict_response```?

plisovin: @mattonfoot changed to ```bad_request``` in 9de72bb4

PR #984, opened 2021-09-08
Add updated_since filter to /api/v1/participant-declarations

jamesrose:
## Ticket and context

Add updated_since filter to `GET /api/v1/participant-declarations`

Allow participant declarations to be filtered by when they were created (or last updated). A provider tried to do this previously and I don't see any reason why we wouldn't let them.

### Code quality checks
- [X] All commit messages are meaningful and true
- [X] Added enough automated tests

### DIFF
--- a/app/controllers/api/v1/participant_declarations_controller.rb
+++ b/app/controllers/api/v1/participant_declarations_controller.rb
@@ -31,6 +31,7 @@ module Api
       def query_scope
         scope = ParticipantDeclaration.for_lead_provider(cpd_lead_provider)
         scope = scope.where("user_id = ?", participant_id_filter) if participant_id_filter.present?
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -42,6 +43,10 @@ def participant_id_filter
         filter[:participant_id]
       end
 
+      def updated_since_filter
+        filter[:updated_since]
+      end
```
kicferk1, line 48: Let's use the new and improved `ApiFilter` concern for it :)

jamesrose: @kicferk1 🙏 Thank you!

jamesrose: @kicferk1 Much better, thanks. Updated.

kicferk1 (approved):

PR #985, opened 2021-09-08
Add parallel tests for rspec for dev use

asmega:
## Ticket and context

Ticket: none

- Adds ability to run rspec test suite in parallel
- This at the moment is for local developer use only as a first step
- Maybe after we can patch it into the CI process
- Has dropped down from 5/6 minutes runtime to about 90 seconds for my local machine

### Code quality checks

### HTML Checks

### Gotchas

- n/a

## External API changes

- none

### DIFF
--- a/app/controllers/lead_providers/openapi_controller.rb
+++ b/app/controllers/lead_providers/openapi_controller.rb
@@ -2,6 +2,8 @@
 
 # source: https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/controllers/api_docs/vendor_api_docs/openapi_controller.rb
 
+require "lead_provider_api_specification"
+
 module LeadProviders
   class OpenapiController < ApplicationController
     def api_docs
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -75,4 +75,8 @@ Rails.application.configure do
   # config.action_view.annotate_rendered_view_with_filenames = true
 
   config.cip_resource_bucket = "https://paas-s3-broker-prod-lon-ac28a7a5-2bc2-4d3b-8d16-a88eaef65526.s3.amazonaws.com"
+
+  if config.respond_to?(:web_console)
+    config.web_console.development_only = false
+  end
 end
--- a/.rspec_parallel
+++ b/.rspec_parallel
@@ -1,0 +1,5 @@
+--require rails_helper
+--require spec_helper
+--require payment_calculator
+--format progress
+--format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log
--- a/Gemfile
+++ b/Gemfile
@@ -142,6 +142,8 @@ group :development, :test do
   # Swagger generator
   gem "multi_json"
   gem "open_api-rswag-specs", ">= 0.1.0"
+
+  gem "parallel_tests"
 end
 
 group :development, :deployed_development, :test, :sandbox do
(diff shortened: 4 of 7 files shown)


config/environments/test.rb
```diff
@@ -75,4 +75,8 @@
   # config.action_view.annotate_rendered_view_with_filenames = true
 
   config.cip_resource_bucket = "https://paas-s3-broker-prod-lon-ac28a7a5-2bc2-4d3b-8d16-a88eaef65526.s3.amazonaws.com"
+
+  if config.respond_to?(:web_console)
+    config.web_console.development_only = false
+  end
```
kicferk1, line 81: what's this for?

app/controllers/lead_providers/openapi_controller.rb
```diff
@@ -2,6 +2,8 @@
 
 # source: https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/controllers/api_docs/vendor_api_docs/openapi_controller.rb
 
+require "lead_provider_api_specification"
+
```
kicferk1, line 6: Why do we need it here?

app/controllers/lead_providers/openapi_controller.rb
```diff
@@ -2,6 +2,8 @@
 
 # source: https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/controllers/api_docs/vendor_api_docs/openapi_controller.rb
 
+require "lead_provider_api_specification"
```
asmega, line 5: rails does not load `/lib` so should be required by hand is my understanding

kicferk1 (commented):
Looks pretty good. I imagine it won't work in CI in the current form, right?

config/environments/test.rb
```diff
@@ -75,4 +75,8 @@
   # config.action_view.annotate_rendered_view_with_filenames = true
 
   config.cip_resource_bucket = "https://paas-s3-broker-prod-lon-ac28a7a5-2bc2-4d3b-8d16-a88eaef65526.s3.amazonaws.com"
+
+  if config.respond_to?(:web_console)
+    config.web_console.development_only = false
```
asmega, line 80: seems to be some kind of quirk from `parallel_tests`. I don't actually know why but assume it for some reason it loads the dev gems whilst in test mode

asmega: actually maybe to do with spring and the way it loads

app/controllers/lead_providers/openapi_controller.rb
asmega: prior to this `lead_provider_api_specification` is loaded by `app/controllers/lead_providers/guidance_controller.rb`. If that file is never loaded then `lead_provider_api_specification` will also not be available. `/lib` is not autoloaded by rails therefore must be required by hand before it is available

asmega: from `config/environments/test.rb` `config.eager_load = false` therefore not everything is loaded prior to test suite running.

kicferk1: Ah, alright, that makes sense

kicferk1 (approved):

asmega:
> Looks pretty good. I imagine it won't work in CI in the current form, right?

It's not hooked up so won't work. I think we want to run it locally for now to iron out any weirdness before updating CI with potential flakiness.

PR #986, opened 2021-09-08
CPDTP-424 Allow a provider to query the status of a specific declaration

plisovin:
## Ticket and context

Providers need ability to query the pay-ability status of a specific declaration. 

## Tech review

A new show routes has been added to query by declaration id i.e. 
```
/api/v1/participant-declarations/d4a4e4f0-24f5-4934-8736-19e58a0cc01c
```

### Code quality checks

### HTML Checks

### Gotchas

## External API changes


kicferk1 (commented):
I think to get a specific declaration details we should just add a `show` endpoint, or equivalent of that.

So you go to `GET .../participant-declarations/uid`

And this endpoint will return a single declaration response.

After all, the index endpoint returns a list, but if you have exact declaration id, surely what you are after is a single object.

plisovin:
> I think to get a specific declaration details we should just add a `show` endpoint, or equivalent of that.
> 
> So you go to `GET .../participant-declarations/uid`
> 
> And this endpoint will return a single declaration response.
> 
> After all, the index endpoint returns a list, but if you have exact declaration id, surely what you are after is a single object.

yeah, makes sense

app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -26,12 +26,21 @@ def index
         end
       end
 
+      def show
+        record = ParticipantDeclaration.find_by(id: params[:id])
```
kicferk1, line 30: `ParticipantDeclaration.for_lead_provider(cpd_lead_provider).find_by(id: params[:id])`?

We don't want to show the declaration of Ambition to UCL (or anyone else), even if the chances they get the right UID is slim.

plisovin: Yeah, good point, updated in 01e5060 and also added to swagger docs

swagger/v1/api_spec.json
```diff
@@ -422,6 +422,59 @@
...
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "A single participant declaration",
+            "content": {
+              "application/vnd.api+json": {
+                "schema": {
+                  "$ref": "#/components/schemas/ParticipantDeclarationsRecordedResponse"
```
kicferk1, line 457: `ParticipantDeclarationsRecordedResponse` does not sound that great in here - we don't record anything in a get endpoint. 

Should we rename it `ParticipantDeclarationsResponse`, now that we reuse it between get and post endpoints?

kicferk1 (commented):
Looks pretty good to me.

Do we need to update changelog? And / or usage guide?

plisovin:
> Looks pretty good to me.
> 
> Do we need to update changelog? And / or usage guide?

yep, I think so, I will look into this

plisovin:
> @plisovin I made some changes in this area just now so you might need to rebase. Also, agree with @kicferk1 .

@jamesrose do you know why participant_id has been removed from the ListFilter i.e. https://github.com/DFE-Digital/early-careers-framework/blob/develop/swagger/v1/component_schemas/ListFilter.yml

plisovin:
> > @plisovin I made some changes in this area just now so you might need to rebase. Also, agree with @kicferk1 .
> 
> @jamesrose do you know why participant_id has been removed from the ListFilter i.e. https://github.com/DFE-Digital/early-careers-framework/blob/develop/swagger/v1/component_schemas/ListFilter.yml

ah it's been moved, ignore this

kicferk1 (approved):
Double check whether we want guidance for it with James C, but other than that it looks good to me.

kicferk1 (approved):

PR #990, opened 2021-09-09
CPDTP-368: Resume participant

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-368

When a participant is deferred, they can be resumed. For time being changing their schedule is a separate operation, to simplify this endpoint. In general we might add an optional schedule identifier param to this endpoint at some point.

If a participant is 'active by default' - does not have a state object in the db - then we have a hard time preventing them from being resumed (could be done but I think it would be pretty ugly). I think we should run come command to make an active profile state for each participant profile without one - at least on ECF side of things.

### Code quality checks

### HTML Checks

### Gotchas

2. Get list of participants
3. Hit the defer endpoint for one of them
4. Hit the resume endpoint for the same one - they should switch to being active
5. Hit it for an active or withdrawn participant - it should fail

## External API changes

Does this change anything in our external APIs? If so, did you remember to update the CHANGELOG for Lead Providers? (lead-providers/guidance/release-notes) - it does.
I updated guidance and release notes, will send a review app to James C and Conor before merging for review.


spec/lib/lead_provider_api_specification_spec.rb
```diff
@@ -18,6 +18,7 @@
         /api/v1/participants/{id}/defer
         /api/v1/participants/{id}/withdraw
         /api/v1/participants/{id}/change-schedule
+        /api/v1/participants/{id}/resume
```
blackrat, line 21: Not a big deal, but it might be worth having defer and resume together, since they're more closely related. I'd also have "withdraw" last.

app/views/lead_providers/guidance/ecf_usage.html.erb
```diff
@@ -4,6 +4,7 @@
   <li><a href="#continuing-the-ecf-registration-process" class="govuk-link">Continuing the ECF registration process</a></li>
   <li><a href="#notifying-of-withdrawal" class="govuk-link">Notify that an ECF participant has withdrawn from their course</a></li>
   <li><a href="#notifying-of-deferral" class="govuk-link">Notify that an ECF participant is taking a break from their course</a></li>
+  <li><a href="#notifying-of-resume" class="govuk-link">Notify that an ECF participant is taking a break from their course</a></li>
```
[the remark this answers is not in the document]
kicferk1: Done

spec/requests/api/v1/participants_spec.rb
```diff
@@ -277,6 +277,60 @@
           expect(parsed_response.dig("data", "attributes", "schedule_identifier")).to eql("ecf-september-extended-2021")
         end
       end
+
+      describe "JSON Participant Deferral" do
```
jamesrose, line 281: > JSON

😬

jamesrose:
Looks good to me! I'll defer to your team for the final approval.

spec/requests/api/v1/participants_spec.rb
kicferk1: 🙈

app/services/participants/resume/validate_and_change_state.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Participants
+  module Resume
+    module ValidateAndChangeState
+      extend ActiveSupport::Concern
```
blackrat, line 6: Doesn't need to extend ActiveSupport::Concern.

spec/docs/participants_spec.rb
```diff
@@ -146,8 +146,8 @@
     end
   end
 
-  path "/api/v1/participants/{id}/withdraw" do
-    put "Withdraw a participant from a course" do
+  path "/api/v1/participants/{id}/resume" do
+    put "Resume a participant into a course" do
```
blackrat, line 150: I'd word this slightly differently to indicate they are "returning to their course after a break". Doesn't read well at the moment.

spec/services/participants/withdraw/mentor_spec.rb
```diff
@@ -25,7 +25,7 @@
     it "creates a withdrawn state for that user's profile" do
       expect { described_class.call(params: participant_params) }
         .to change { ParticipantProfileState.count }.by(1)
-      expect { User.find(participant_params[:participant_id]).mentor_profile.withdrawn? }
+      expect { ParticipantProfileState.order(created_at: :desc).first.withdrawn? }
```
blackrat, line 28: Why? The former mechanism ensure that the user profile has the correct state. That's testing the business logic. The changed one only that the last entry is the expected value, which is testing the implementation.

kicferk1:
```
it "creates a withdrawn state for that user's profile"
```
Didn't sound quite like what the test was doing.

Fine, I will being back the old logic and rename the test.

spec/docs/participants_spec.rb
kicferk1: I updated it to have a similar copy to our guidance

blackrat (approved):

PR #992, opened 2021-09-09
admins can filter participants by type

asmega:
## Ticket and context

Ticket: A tiny part of https://trello.com/c/3FttpebV/406-handle-duplicate-signups

### Screenshots

Desktop
Screenshot 2021-09-09 at 12 14 00

Mobile
Screenshot 2021-09-09 at 12 14 46

### Code quality checks

### HTML Checks

### Gotchas

- login as an admin
- filter participants by different types
- should only return participants of selected type

## External API changes

- none

### DIFF
--- a/app/components/search_box.html.erb
+++ b/app/components/search_box.html.erb
@@ -9,6 +9,18 @@
           )
       %>
     </div>
+
+    <% filters.each do |filter| %>
+      <div class="search-box__input">
+        <%= f.govuk_collection_select filter[:field],
+          filter[:options],
+          :id,
+          :name,
+          label: { text: "Filter by #{filter[:field]}" },
+          options: { selected: filter[:value] } %>
+      </div>
+    <% end %>
+
     <%= f.govuk_submit "Search", classes: "search-box__button", "data-test": "search-button" %>
   <% end %>
 </div>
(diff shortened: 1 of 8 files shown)


app/components/search_box.html.erb
```diff
@@ -9,6 +9,20 @@
...
+        [
+          OpenStruct.new(id: "", name: ""),
+          OpenStruct.new(id: "ParticipantProfile::ECT", name: "ECT"),
+          OpenStruct.new(id: "ParticipantProfile::Mentor", name: "Mentor"),
+          OpenStruct.new(id: "ParticipantProfile::NPQ", name: "NPQ")
+        ],
+        :id,
+        :name,
+        label: { text: "Filter by type" } %>
+    </div>
```
kicferk1, line 24: Can we make it stay after search, like query?

asmega: Yes. Actually this has pointed me to another bug whereby this component is shared on other views such as schools. so let me fix that too

asmega: Ok, filters are now passed in which solves the "used in different area problems". One can now specify the value so it'll get passed down to the filter to select it.

kicferk1 (approved):

jamesrose (approved):
Yes!

PR #993, opened 2021-09-09
Service to send reminders to STIs

BroiSatse:
https://dfedigital.atlassian.net/browse/CPDRP-931

### DIFF
--- a/app/mailers/school_mailer.rb
+++ b/app/mailers/school_mailer.rb
@@ -18,7 +18,7 @@ class SchoolMailer < ApplicationMailer
   ADD_PARTICIPANTS_EMAIL_TEMPLATE = "721787d0-74bc-42a0-a064-ee0c1cb58edb"
   YEAR2020_INVITE_EMAIL_TEMPLATE = "d4b53e26-4630-43a5-b89e-3c668061a41c"
 
-  def remind_sits_to_setup_cohort(recipient:, school_name:, campaign: nil)
+  def remind_induction_coordinator_to_setup_cohort_email(recipient:, school_name:, campaign: nil)
     campaign_tracking = campaign ? UTMService.email(campaign, campaign) : {}
 
     template_mail(
--- a/app/mailers/school_mailer.rb
+++ b/app/mailers/school_mailer.rb
@@ -30,8 +30,8 @@ class SchoolMailer < ApplicationMailer
         subject: "You need to set up your ECT training programme",
         school_name: school_name,
         sign_in: new_user_session_url(**campaign_tracking),
-        step_by_step: step_by_step_url(**campaign_tracking)
-      }
+        step_by_step: step_by_step_url(**campaign_tracking),
+      },
     )
   end
 
--- a/app/mailers/school_mailer.rb
+++ b/app/mailers/school_mailer.rb
@@ -18,6 +18,23 @@ class SchoolMailer < ApplicationMailer
   ADD_PARTICIPANTS_EMAIL_TEMPLATE = "721787d0-74bc-42a0-a064-ee0c1cb58edb"
   YEAR2020_INVITE_EMAIL_TEMPLATE = "d4b53e26-4630-43a5-b89e-3c668061a41c"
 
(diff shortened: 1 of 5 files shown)


app/services/validation_beta_service.rb
```diff
@@ -1,6 +1,19 @@
...
+    School.where(id: empty_school_cohorts.select(:school_id)).includes(:induction_coordinators).find_each do |school|
+      school.induction_coordinators.each do |sit|
+        SchoolMailer.remind_sits_to_setup_cohort(
+          recipient: sit.email,
+          school_name: school.name,
+          campaign: :sit_to_complete_steps,
+        ).deliver_later
+      end
+    end
+  end
```
kicferk1, line 15: Soo, question one, if an IC has multiple schools, should they get more than one email? In other cases we tend to just send them one, but this will not be so kind.

Question two, schools can opt out - `not_opted_out` scope on schools. Should we respect that?

Question three, school cohort has states like `no_early_career_teachers`, `school_funded_fip`. I am pretty sure we don't collect teachers for those two. Should we exclude those school cohorts?

kicferk1 (commented):
Unit tests for who gets what emails?

app/services/validation_beta_service.rb
BroiSatse: All good questions, will check with Nicola. :)

BroiSatse:
@kicferk1 Per unit tests, had a chat with @jamesrose about those. So, IMO, unit tests are there to ensure maintainability of the code, so they can pick up quickly if something is broken in the future. This service will run once and will be deleted, so I don't think writing a full test suite for it would be justified here

kicferk1 (approved):

app/mailers/school_mailer.rb
```diff
@@ -18,6 +18,23 @@ class SchoolMailer < ApplicationMailer
   ADD_PARTICIPANTS_EMAIL_TEMPLATE = "721787d0-74bc-42a0-a064-ee0c1cb58edb"
   YEAR2020_INVITE_EMAIL_TEMPLATE = "d4b53e26-4630-43a5-b89e-3c668061a41c"
 
+  def remind_sits_to_setup_cohort(recipient:, school_name:, campaign: nil)
```
jamesrose, line 21:
```suggestion
  def remind_induction_coordinator_to_setup_cohort_email(recipient:, school_name:, campaign: nil)
```

nit: _email consistency

app/services/validation_beta_service.rb
```diff
@@ -1,6 +1,31 @@
...
+      .where(opt_out_of_updates: false)
+      .where.not(id: ParticipantProfile::ECF.select(:school_cohort_id))
+
+    emails_sent = []
+
+    School.where(id: empty_school_cohorts.select(:school_id)).includes(:induction_coordinators).find_each do |school|
+      school.induction_coordinators.each do |sit|
+        next if emails_sent.include?(sit.email)
+
+        SchoolMailer.remind_sits_to_setup_cohort(
```
jamesrose, line 16:
```suggestion
        SchoolMailer.remind_induction_coordinator_to_setup_cohort_email(
```

app/services/validation_beta_service.rb
```diff
@@ -1,6 +1,31 @@
...
+    empty_school_cohorts = SchoolCohort
+      .where(induction_programme_choice: %i[full_induction_programme core_induction_programme not_yet_known])
+      .where(opt_out_of_updates: false)
+      .where.not(id: ParticipantProfile::ECF.select(:school_cohort_id))
+
+    emails_sent = []
+
+    School.where(id: empty_school_cohorts.select(:school_id)).includes(:induction_coordinators).find_each do |school|
+      school.induction_coordinators.each do |sit|
+        next if emails_sent.include?(sit.email)
```
jamesrose, line 14:
```suggestion
        next if emails_sent.include?(sit.email)
```

This isn't idempotent because we aren't persisting anything against the SIT. If the instance dies while you're running this, can you run it again?

jamesrose (approved):

jamesrose (approved):

PR #995, opened 2021-09-10
CPDTP-208: Void declarations

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-208

We want the providers to be able to tell us that declaration was submitted in error - and to void them. 

I think we care about when the declaration was voided, so I stuck it in `ProfileDeclaration`s. 

### Code quality checks

2. Make a declaration
3. Try voiding it

## External API changes

Does this change anything in our external APIs? If so, did you remember to update the CHANGELOG for Lead Providers? (lead-providers/guidance/release-notes) - still to do


app/services/void_participant_declaration.rb
```diff
@@ -0,0 +1,24 @@
...
+class VoidParticipantDeclaration
+  attr_accessor :cpd_lead_provider, :id
+
+  def call
+    declaration = ParticipantDeclaration.find(id)
+
+    raise ActionController::BadRequest, "Declaration is already voided" if declaration.voided
+
+    latest_declaration = declaration.participant_profile.participant_declarations.order(declaration_date: :desc).first
+    raise ActionController::BadRequest, "Can only void last declaration" if latest_declaration != declaration
```
jamesrose, line 12: Why?

kicferk1: That's what the ticket says 😂 

I think it's unclear how we would handle voiding if there is a follow-up declaration. Presumably if the started declaration is voided, then retained declaration is at least suspect too. It's a business problem to be tackled lated, so we just don't lt the providers do it for now.

app/models/participant_declaration.rb
```diff
@@ -50,8 +50,22 @@ def refresh_payability!
...
       )
       reload
     end
   end
+
+  def void!
+    return if current_profile_declaration&.voided
+
+    # TODO: Prevent voiding a processed declaration - that requires clawbacks
+    ProfileDeclaration.create!(
```
jamesrose, line 63: I didn't expect us to create a new record, but to update an existing one with `voided_at: Time.now.current` or similar. What's the motivation for this approach?

app/services/void_participant_declaration.rb
jamesrose: @kicferk1 What if they got the date wrong on the `started` declaration, but only realised later, after the first retained one?

kicferk1: I think if that's the case the voiding will have to trigger a clawback, because we will have most likely processed it. When we build that functionality we will probably add in the ability to void more declarations.

Or keep the scenario that you described as a support request until it becomes too commonplace and annoying.

app/models/participant_declaration.rb
kicferk1: Hmmm, I think I like keeping a single `voided_at` date. I didn't think too hard on it, the general feeling I had was "well, we need to keep track of when it happened, and those things should live in profile declarations".

db/migrate/20210909154907_add_voiding_to_declarations.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddVoidingToDeclarations < ActiveRecord::Migration[6.1]
+  def change
+    add_column :participant_declarations, :voided_at, :datetime
```
blackrat, line 5: Should this be in the metadata/profile rather than here?

kicferk1: You mean in `profile_declarations`? I don't think so.

Truth be told I am not 100% sure anything should be in `profile_declarations` (why would `profile` need to be there instead of on the declaration itself?) but voiding in particular is a one-way street. We do not plan to ever un-void a declaration. So storing it on the declaration itself seems perfectly fine for me, and it is easier.

Am I missing something?

app/serializers/participant_declaration_serializer.rb
```diff
@@ -18,5 +18,9 @@ class ParticipantDeclarationSerializer
     declaration.declaration_date.rfc3339
   end
 
+  attribute :voided do |declaration|
+    declaration.voided || false
+  end
+
```
blackrat, line 24: Are we doing this as opposed to having it a declaration state? Do you see this changing as a change to the API later?

kicferk1: I think it can be greatly simplified now - this code was written on top of a previous version which did indeed store voided data on profile declarations.

I think now it should be just `attribute :voided`. Good spot!

app/models/participant_declaration.rb
```diff
@@ -54,4 +54,13 @@ def refresh_payability!
       reload
     end
   end
+
+  def voided
```
blackrat, line 58: Don't like this in here if we're allowing multiple voids on the same record. Are we just going to keep the first one of those even if we don't reject them?

db/migrate/20210909154907_add_voiding_to_declarations.rb
blackrat: What if support does the "unvoid". I can see the "I voided the wrong declaration by mistake" being a thing. I don't like the idea of anything touching the actual declaration where are they could just delete an incorrect voiding profile.

spec/services/void_participant_declaration_service_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      described_class.new(cpd_lead_provider: cpd_lead_provider, id: declaration.id).call
+      expect(declaration.reload.voided).to be_truthy
+    end
+
+    it "does not void a voided declaration" do
+      declaration = ParticipantDeclaration.order(:created_at).last
+      described_class.new(cpd_lead_provider: cpd_lead_provider, id: declaration.id).call
+      expect {
+        described_class.new(cpd_lead_provider: cpd_lead_provider, id: declaration.id).call
+      }.to raise_error ActionController::BadRequest
```
blackrat, line 31: Why BadRequest? Surely it's an Unprocessable entity. The form of the request is ok, it's just the content.

spec/services/void_participant_declaration_service_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    it "only voids the last declaration" do
+      old_declaration = create(:participant_declaration, declaration_date: start_date + 1.day, course_identifier: "ecf-induction", declaration_type: "started")
+      create(:profile_declaration, participant_declaration: old_declaration, participant_profile: ect_profile)
+      expect {
+        described_class.new(cpd_lead_provider: cpd_lead_provider, id: old_declaration.id).call
+      }.to raise_error ActionController::BadRequest
```
blackrat, line 39: Same comment

app/models/participant_declaration.rb
kicferk1: I don't think this PR wants to allow multiple voids per single record. I am pretty sure more than one void is not a valid business state in general.

app/services/void_participant_declaration.rb
```diff
@@ -6,10 +6,10 @@ class VoidParticipantDeclaration
   def call
     declaration = ParticipantDeclaration.find(id)
```
blackrat, line 7: Shouldn't this be fetching declarations scoped for that lead_provider? Wouldn't this let one lead provider void another lead provider's declarations.

blackrat (approved):

app/views/lead_providers/guidance/release_notes.html.erb
```diff
@@ -2,6 +2,9 @@
   If you have any questions or comments about these notes, please contact DfE via Slack or email.
 </p>
 
+<h2 class="govuk-heading-l">15th September 2021</h2>
+<p class="govuk-body-m">Ability to resume void participant declarations.</p>
```
blackrat, line 6: Should just be void

PR #996, opened 2021-09-10
Add training status states to all places creating new profiles

kicferk1:
## Ticket and context

https://github.com/DFE-Digital/early-careers-framework/pull/990 adds resuming participants. We should not resume a participant when they are already active.

But we haven't made a training status for each participant profile yet - some are active by the fact of not having a training status.

Soo, we should just add a training status to all existing profiles. When this is merged and deployed I will run something like 
```
ParticipantProfile.each { |profile| ParticipantProfileState.find_or_create_by!({ participant_profile: profile }) }
```

Nothing I can think of.

### Code quality checks

### DIFF
--- a/app/services/early_career_teachers/create.rb
+++ b/app/services/early_career_teachers/create.rb
@@ -17,7 +17,7 @@ module EarlyCareerTeachers
         end
 
         ParticipantProfile::ECT.create!({ teacher_profile: teacher_profile, schedule: Finance::Schedule.default }.merge(ect_attributes)) do |profile|
-          ParticipantProfileState.create!({ participant_profile: profile })
+          ParticipantProfileState.find_or_create_by!(participant_profile: profile)
         end
       end
     end
--- a/app/services/early_career_teachers/create.rb
+++ b/app/services/early_career_teachers/create.rb
@@ -17,7 +17,7 @@ module EarlyCareerTeachers
         end
 
         ParticipantProfile::ECT.create!({ teacher_profile: teacher_profile, schedule: Finance::Schedule.default }.merge(ect_attributes)) do |profile|
-          ParticipantProfileState.find_or_create_by!(participant_profile: profile)
+          ParticipantProfileState.create!(participant_profile: profile)
         end
       end
     end
--- a/app/services/npq/create_or_update_profile.rb
+++ b/app/services/npq/create_or_update_profile.rb
@@ -23,7 +23,7 @@ module NPQ
       participant_profile.school_ukprn = npq_validation_data.school_ukprn
       participant_profile.save!
 
-      ParticipantProfileState.find_or_create_by!({ participant_profile: participant_profile })
+      ParticipantProfileState.find_or_create_by!(participant_profile: participant_profile)
     end
 
   private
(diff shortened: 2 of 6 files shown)


app/services/early_career_teachers/create.rb
```diff
@@ -17,7 +17,7 @@ def call
         end
 
         ParticipantProfile::ECT.create!({ teacher_profile: teacher_profile, schedule: Finance::Schedule.default }.merge(ect_attributes)) do |profile|
-          ParticipantProfileState.create!({ participant_profile: profile })
+          ParticipantProfileState.find_or_create_by!(participant_profile: profile)
```
blackrat, line 20: Why? It should only be possibly to create, since it's in the create block for the profile.

app/services/npq/create_or_update_profile.rb
```diff
@@ -22,6 +22,8 @@ def call
       participant_profile.school_urn = npq_validation_data.school_urn
       participant_profile.school_ukprn = npq_validation_data.school_ukprn
       participant_profile.save!
+
+      ParticipantProfileState.find_or_create_by!({ participant_profile: participant_profile })
```
blackrat, line 26: I'm assuming that the thinking is that we just want to create if it's a new profile?

lib/tasks/valid_test_data_generator.rb
```diff
@@ -60,11 +60,11 @@ def create_participant(school_cohort:, profile_type: :ect, mentor_profile: nil,
       schedule = Finance::Schedule.default
       if profile_type == :ect
         ParticipantProfile::ECT.create!(teacher_profile: teacher_profile, school_cohort: school_cohort, mentor_profile: mentor_profile, status: status, sparsity_uplift: sparsity_uplift, pupil_premium_uplift: pupil_premium_uplift, schedule: schedule) do |profile|
-          ParticipantProfileState.create!({ participant_profile: profile })
+          ParticipantProfileState.find_or_create_by!(participant_profile: profile)
```
blackrat, line 63: Don't need to do find when it's in the create block of the profile.

lib/tasks/valid_test_data_generator.rb
```diff
@@ -60,11 +60,11 @@ def create_participant(school_cohort:, profile_type: :ect, mentor_profile: nil,
...
         end
       else
         ParticipantProfile::Mentor.create!(teacher_profile: teacher_profile, school_cohort: school_cohort, status: status, sparsity_uplift: sparsity_uplift, pupil_premium_uplift: pupil_premium_uplift, schedule: schedule) do |profile|
-          ParticipantProfileState.create!({ participant_profile: profile })
+          ParticipantProfileState.find_or_create_by!(participant_profile: profile)
```
blackrat, line 67: Same comment.

app/services/early_career_teachers/create.rb
kicferk1: Ah, a global find and replace is why :P

app/services/npq/create_or_update_profile.rb
kicferk1: That's right.

jamesrose:
```suggestion
      ParticipantProfileState.find_or_create_by!(participant_profile: participant_profile)
```

nit: consistency

jamesrose (approved):

PR #997, opened 2021-09-10
Setting up action mailer default url host based on domain

BroiSatse:
Currently, `action_mailer.default_url_options.host` is blank in almost all environments. This leads to us adding `host: Rails.application.config.domain` to all url_helpers that are used for email purposes. If this is missed, the emails will work in development, but not in production, which is problematic.

Now, ideally this would be set in the application.rb to be shared by all environment. However, `config.domain` is set in each environment file which loads after the `application.rb` so this would not work - that's why we need to work within the initializer. Additionally, the initializer files run after railtie initializers, including ActionMailer::Railtie which is copying populating `ActionMailer::Base` configuration with `Rails.application.config.action_mailer`, whcih is why it is setting the host directly on `ActionMailer::Base` - changes to the config in intiializer won't take effect.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -43,7 +43,6 @@ Rails.application.configure do
   config.action_mailer.delivery_method = :notify
   config.action_mailer.perform_deliveries = true
   config.action_mailer.perform_caching = false
-  config.action_mailer.default_url_options = { host: config.domain }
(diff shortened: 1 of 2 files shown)


jamesrose (approved):

tonyheadford (approved):

kicferk1 (approved):
Did you test the emails get sent and their urls go to the correct place in the review app?

BroiSatse:
@kicferk1 They did :)

PR #998, opened 2021-09-10
CDPTP-473 Add basic ECF contract display for validation of contract information

blackrat:
## Ticket and context

This adds a view of contract information which allows the underlying providers contract details which forms the basis of the calculations.
There will be a wider "contract summary" and proper design, but for now, this gives a general shape that can be viewed.

Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDTP/boards/87?selectedIssue=CPDTP-473

## Tech review

This PR is mainly for the view components and controller actions, since it's more likely that we'll have a landing page and a more details Contract Summary once it's gone through the content design phase.

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/app/components/finance/ecf/contract.rb
+++ b/app/components/finance/ecf/contract.rb
@@ -15,7 +15,7 @@ module Finance
         contract.bands.each_with_index.map do |band, index|
           {
             index: index,
-            band: band
+            band: band,
           }
         end
       end
--- a/app/components/finance/ecf/contract.rb
+++ b/app/components/finance/ecf/contract.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module Finance
+  module ECF
+    class Contract < BaseComponent
+      include FinanceHelper
+      attr_accessor :contract
+      delegate :uplift_target, :uplift_amount, :recruitment_target, :set_up_fee, to: :contract
+
+      def name
+        contract.lead_provider.name
+      end
+
+      def bands
+        contract.bands.each_with_index.map do |band, index|
+          {
+            index: index,
+            band: band
+          }
+        end
+      end
+
+    private
+
+      def initialize(contract:)
+        @contract = contract
+      end
+    end
+  end
+end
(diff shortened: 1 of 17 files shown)


app/components/finance/ecf/contract.rb
```diff
@@ -0,0 +1,30 @@
...
+      delegate :uplift_target, :uplift_amount, :recruitment_target, :set_up_fee, to: :contract
+
+      def name
+        contract.lead_provider.name
+      end
+
+      def bands
+        contract.bands.each_with_index.map do |band, index|
+          {
+            index: index,
```
kicferk1, line 17: Should we compute the band identifier here?

app/helpers/finance_helper.rb
```diff
@@ -4,4 +4,12 @@ module FinanceHelper
...
     number_to_currency number, precision: 2, unit: "£"
   end
+
+  def float_to_percentage(number)
+    number_to_percentage(number * 100, precision: 0)
+  end
+
+  def band_to_identifier(index)
+    ("A".ord + index).chr
+  end
```
kicferk1, line 14: We are doing the same in the payment calculator and the output payment aggregator.

I wonder if an easier thing for us would be have something like `bands_with_identifiers` method on the contract model. Or even adding band name to the database. Thoughts?

kicferk1 (commented):
Can we have some accessibility and percy tests for the new page? Preferably in rspec rather than in cypress?

I know that this look is likely going to change, but it is easier to remember to add those tests now than remember to do it when we get the final version.

app/helpers/finance_helper.rb
blackrat: It's really only a decorator. Bands are ordered correctly defined by the limits, and I don't think they're needed apart for for display purposes. I did think about having them as a locale based rather than a fixed string. That may be better and allow for more flexibility when we start with D and potentially other bands based on the contractual changes that are happening. I see it more like the percentage and pound display, so having the other "output" displays using this as opposed to duplicating is probably a refactor that I'd prefer to do.

blackrat:
This is transient, and there's another branch which will change this already. Design for how it's going to look is underway and it's more in keeping with the design. That definitely will have percy/rspec-capybara too.

kicferk1 (approved):

kicferk1 (approved):

PR #1003, opened 2021-09-13
Minor correctio to validation reminder emails

BroiSatse:

We do not want to ask withdrawn participants to add their details. Since SIT mentors will reaceive the separate email, sending generic mentor emails to them would be just a duplication.

### DIFF
--- a/app/services/validation_beta_service.rb
+++ b/app/services/validation_beta_service.rb
@@ -28,7 +28,8 @@ class ValidationBetaService
   # ECTs who have not added their details for validation
   def tell_ects_to_add_validation_information
     ParticipantProfile::ECT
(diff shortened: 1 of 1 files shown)


kicferk1 (approved):
The google doc is not very good at telling me why we want this change. A bit more context in PR description would be nice.

As in: "We don't want to send anything to withdrawn participants. We don't want to send mentor validation info to ICs". We haven't sent these emails in live yet, right?

BroiSatse:
@kicferk1 Some description added. We have already sent those emails last week, we'll be sending them again today (`request_for_details` column ensures partiicpants who has received it last time won't be nudged again) and likely again on wednesday.

PR #1006, opened 2021-09-13
Preparing Sit's participants notified emails

BroiSatse:

1. We ensure we only notifies SIT about us notifying their participants that they've added after 7th of September
2. We ensure we do not notify SIT of opt-out cohorts

### DIFF
--- a/app/services/validation_beta_service.rb
+++ b/app/services/validation_beta_service.rb
@@ -110,14 +110,25 @@ class ValidationBetaService
   end
 
   def tell_induction_coordinators_we_asked_ects_and_mentors_for_information
-    InductionCoordinatorProfile.find_each do |ic|
-      ic.schools.not_opted_out.each do |school|
-        if chosen_programme_and_not_in_beta(school)
-          send_induction_coordinators_we_asked_ects_and_mentors_for_information(ic, school)
-          break
+    school_cohorts = SchoolCohort
+      .where(
+        id: ParticipantProfile::ECF.where("created_at > ?", Date.new(2021, 9, 7)).select(:school_cohort_id),
+        induction_programme_choice: %w[core_induction_programme full_induction_programme],
+      )
+
+    school_ids = school_cohorts.select(:school_id)
+
+    InductionCoordinatorProfile
+      .includes(:schools)
+      .where(schools: { id: school_ids })
+      .find_each do |ic|
+        ic.schools.not_opted_out.where(id: school_ids) do |school|
+          if chosen_programme_and_not_in_beta(school)
+            send_induction_coordinators_we_asked_ects_and_mentors_for_information(ic, school)
+            break
+          end
         end
       end
-    end
   end
 
 private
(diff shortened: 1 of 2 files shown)


app/services/validation_beta_service.rb
```diff
@@ -110,14 +110,24 @@ def tell_induction_coordinators_who_are_mentors_to_add_validation_information
...
-        if chosen_programme_and_not_in_beta(school)
-          send_induction_coordinators_we_asked_ects_and_mentors_for_information(ic, school)
-          break
+    school_cohorts = SchoolCohort
+      .where(
+        id: ParticipantProfile::ECF.where("created_at > ?", Date.new(2021, 9, 7)).select(:school_cohort_id),
+        induction_programme_choice: %i[core_induction_programme full_induction_programme]
+      )
+
+
```
kicferk1: We can cut out one line of whitespace here I think

app/services/validation_beta_service.rb
```diff
@@ -110,14 +110,23 @@ def tell_induction_coordinators_who_are_mentors_to_add_validation_information
...
+    InductionCoordinatorProfile
+      .includes(:schools)
+      .where(schools: { id: school_cohorts.select(:school_id) })
+      .find_each do |ic|
+        ic.schools.not_opted_out.each do |school|
+          if chosen_programme_and_not_in_beta(school)
+            send_induction_coordinators_we_asked_ects_and_mentors_for_information(ic, school)
+            break
+          end
```
kicferk1, line 127: If an IC has two schools, and added participants to one of them, we would surely want to send an email with that school name.

But the code here says 'take that IC, for each of their schools check if they are not opted out, check they have chosen programme, and not in beta and send them email with that school'. 

Which I guess will give us the wrong school name a bunch of times.

Not sure if it matters much, but if it does then we should probably have the `each` iteration on schools rather than on ICs - and maybe store a list of ICs that already saw an email.

kicferk1 (commented):
I am not sure I like the idea of adding a bunch of complexity wihtout adding unit tests for it :P

app/services/validation_beta_service.rb
BroiSatse: Good catch! Fixed.

kicferk1 (approved):

spec/services/validation_beta_service_spec.rb
```diff
@@ -25,6 +25,7 @@
   let(:urns) { schools.map(&:urn) }
 
   describe "#tell_induction_coordinators_we_asked_ects_and_mentors_for_information" do
+    before { skip "Not worth fixing this - should be deleted shortly!" }
```
kicferk1, line 28: 😂

BroiSatse: Yeah, I'd need to add a participant and write two separate scenarios to test if they've been added before, on and after 7th of september... Not worth it really...

kicferk1: Do we have a JIRA ticket for deleting this stuff? If so, maybe add it to the skip message :P

PR #1010, opened 2021-09-13
Fixed a silly bug in 'we have asked your participants' email

BroiSatse:
Please don't judge me - you missed that as well... :D:D:D


kicferk1 (approved):
This is why we write unit tests 😂

PR #1011, opened 2021-09-14
Create dependabot.yml

ethanmills:
## Ticket and context
Let's see if Dependabot can help keep our dependencies up to date. We currently only get security updates - let's see if keeping everything more up to date is practical. 

Recommended per https://docs.publishing.service.gov.uk/manual/manage-ruby-dependencies.html

Ticket: N/A

Very open to challenges on this. I'm suggesting we try it out, and turn it off if it doesn't work for us

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,7 +13,7 @@ updates:
       - dependency-name: "*"
         update-types: ["version-update:semver-major"]
     open-pull-requests-limit: 5
-  - package-ecosysyem: "npm"
+  - package-ecosystem: "npm"
     directory: "/"
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,32 @@
...
+    open-pull-requests-limit: 5
+  - package-ecosysyem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    ignore:
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]
+    open-pull-requests-limit: 5
+  - package-ecosystem: "githib-actions"
```
FriedSock, line 24: typo?

ethanmills: Yes typo. The actions we depend on are versioned: https://github.com/DFE-Digital/early-careers-framework/blob/develop/.github/workflows/run_task_dev.yml#L20

kicferk1 (approved):
So do we think this is going to remove the need for snyk? In the long run anyway?

ethanmills:
> So do we think this is going to remove the need for snyk? In the long run anyway?

Possibly, we'll see how it behaves

PR #1013, opened 2021-09-14
CPDTP-296 Reject a declaration if the participant is deemed 'withdrawn' or 'deferred'

plisovin:
## Ticket and context

Ticket: Declarations submitted by providers for participants no longer on the ECF course should be rejected.
Also if a declaration is backdated and the participant profile was active in that milestone for which the declaration is being submitted then it should be accepted.

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -89,15 +89,4 @@ class ParticipantProfile < ApplicationRecord
     # TODO: Do we need to store when this happens outside of papertrail?
     update!(schedule: schedule)
   end
-
-  def validate_participant_state(declaration_date, milestone)
-    if (%w[withdrawn deferred] & participant_profile_states.map(&:state)).any?
-      raise ActionController::ParameterMissing, I18n.t(:declaration_on_incorrect_state) if declaration_date >= Time.zone.now
-
-      active_state_date = participant_profile_states.where(state: "active").order(:created_at).last.created_at
-      unless active_state_date && active_state_date <= milestone.milestone_date.end_of_day
-        raise ActionController::ParameterMissing, I18n.t(:declaration_on_incorrect_state)
-      end
-    end
-  end
 end
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -89,4 +89,15 @@ class ParticipantProfile < ApplicationRecord
     # TODO: Do we need to store when this happens outside of papertrail?
     update!(schedule: schedule)
   end
+
+  def validate_participant_state(declaration_date, milestone)
+    if (%w[withdrawn deferred] & participant_profile_states.map(&:state)).any?
+      raise ActionController::ParameterMissing, I18n.t(:declaration_on_incorrect_state) if declaration_date >= Time.zone.now
+
+      active_state_date = participant_profile_states.where(state: "active").order(:created_at).last.created_at
+      unless active_state_date && active_state_date <= milestone.milestone_date.end_of_day
+        raise ActionController::ParameterMissing, I18n.t(:declaration_on_incorrect_state)
+      end
+    end
+  end
 end
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -81,6 +81,10 @@ class ParticipantProfile < ApplicationRecord
     decision || validation_decisions.build(validation_step: name)
   end
 
+  def state_at(declaration_date)
+    participant_profile_states.where("created_at < ?", declaration_date).order(:created_at).last
+  end
+
   def fundable?
     false
   end
--- a/app/services/participants/base.rb
+++ b/app/services/participants/base.rb
@@ -26,7 +26,7 @@ module Participants
 
   private
 
-    implement_instance_method :participant_profile_state, :participant_profile_states, :perform_action!, :matches_lead_provider?
+    implement_instance_method :participant_profile_state, :perform_action!, :matches_lead_provider?
 
     def initialize(params:)
       params.each do |param, value|
--- a/app/services/participants/base.rb
+++ b/app/services/participants/base.rb
@@ -26,7 +26,7 @@ module Participants
 
   private
 
-    implement_instance_method :participant_profile_state, :perform_action!, :matches_lead_provider?
+    implement_instance_method :participant_profile_state, :participant_profile_states, :perform_action!, :matches_lead_provider?
 
     def initialize(params:)
       params.each do |param, value|
--- a/app/services/participants/ecf.rb
+++ b/app/services/participants/ecf.rb
@@ -6,7 +6,7 @@ module Participants
 
     included do
       delegate :early_career_teacher?, :mentor_profile, :early_career_teacher_profile, :participant?, to: :user, allow_nil: true
-      delegate :school_cohort, :participant_profile_state, :participant_profile_states, to: :user_profile, allow_nil: true
+      delegate :school_cohort, :participant_profile_state, to: :user_profile, allow_nil: true
     end
 
     def matches_lead_provider?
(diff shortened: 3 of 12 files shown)


app/models/participant_profile.rb
```diff
@@ -89,4 +89,15 @@ def update_schedule!(schedule)
...
+  def validate_participant_state(declaration_date, milestone)
+    if (%w[withdrawn deferred] & participant_profile_states.map(&:state)).any?
+      raise ActionController::ParameterMissing, I18n.t(:declaration_on_incorrect_state) if declaration_date >= Time.zone.now
+
+      active_state_date = participant_profile_states.where(state: "active").order(:created_at).last.created_at
+      unless active_state_date && active_state_date <= milestone.milestone_date.end_of_day
+        raise ActionController::ParameterMissing, I18n.t(:declaration_on_incorrect_state)
+      end
+    end
+  end
```
kicferk1, line 102: Why is this in participant profile model? I would expect it to live in the relevant service (`app/services/record_declarations/base.rb`)

app/services/participants/base.rb
```diff
@@ -26,7 +26,7 @@ def call
 
   private
 
-    implement_instance_method :participant_profile_state, :perform_action!, :matches_lead_provider?
+    implement_instance_method :participant_profile_state, :participant_profile_states, :perform_action!, :matches_lead_provider?
```
kicferk1, line 29: This is a service for switching participant state, right? Why do we need this change here?

app/services/participants/ecf.rb
```diff
@@ -6,7 +6,7 @@ module ECF
 
     included do
       delegate :early_career_teacher?, :mentor_profile, :early_career_teacher_profile, :participant?, to: :user, allow_nil: true
-      delegate :school_cohort, :participant_profile_state, to: :user_profile, allow_nil: true
+      delegate :school_cohort, :participant_profile_state, :validate_participant_state, to: :user_profile, allow_nil: true
```
kicferk1, line 9: Don't think we need this in participant services.

app/services/participants/npq.rb
```diff
@@ -5,7 +5,7 @@ module NPQ
     extend ActiveSupport::Concern
 
     included do
-      delegate :validation_data, :participant_profile_state, to: :user_profile, allow_nil: true
+      delegate :validation_data, :participant_profile_state, :validate_participant_state, to: :user_profile, allow_nil: true
```
kicferk1, line 8: Ditto.

kicferk1 (commented):
Can we have an integration test for it?

Say, a provider fetches list of participants, they withdraw / defer one, and try to make a declaration against them (and they fail)?

app/services/participants/ecf.rb
plisovin: Left there after refactor, will remove

app/services/participants/base.rb
plisovin: It was actually in the base class then I moved it to the user_profile as I think it's more appropriate to have it there.

app/models/participant_profile.rb
```diff
@@ -89,4 +89,15 @@ def update_schedule!(schedule)
...
+
+  def validate_participant_state(declaration_date, milestone)
+    if (%w[withdrawn deferred] & participant_profile_states.map(&:state)).any?
+      raise ActionController::ParameterMissing, I18n.t(:declaration_on_incorrect_state) if declaration_date >= Time.zone.now
+
+      active_state_date = participant_profile_states.where(state: "active").order(:created_at).last.created_at
+      unless active_state_date && active_state_date <= milestone.milestone_date.end_of_day
+        raise ActionController::ParameterMissing, I18n.t(:declaration_on_incorrect_state)
+      end
+    end
```
kicferk1, line 101: I think what this should do is: fetch the last participant state before the declaration date (using `created_at`).

Then it says 'Ok, if that state is active, I let you make the declaration. If it is not, you cannot make the declaration'.

I do not think this should validate that declaration date is not in future, or anything relating to milestones - because this should have been validated already.

app/services/participants/base.rb
kicferk1: I mean `app/services/participants/base.rb` and `app/services/record_declarations/base.rb` are not the same, and I would only expect this PR to be touching the latter.

app/models/participant_profile.rb
plisovin: I think it's good idea. I've pushed the change

app/services/record_declarations/base.rb
```diff
@@ -119,6 +120,11 @@ def validate_milestone!
       end
     end
 
+    def validate_participant_state!
+      last_state = participant_profile_states.where("created_at < ?", declaration_date).order(:created_at).last
+      raise ActionController::ParameterMissing, I18n.t(:declaration_on_incorrect_state) unless last_state&.state.nil? || last_state&.state == "active"
```
jamesrose, line 125:
```suggestion
      raise ActionController::ParameterMissing, I18n.t(:declaration_on_incorrect_state) unless last_state&.state.nil? || last_state.active?
```

We can use enum helpers here, and the first condition allow us to be unsafe in the latter.

plisovin: yeah, I also don't think we can have nil state here

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -108,6 +108,43 @@ def build_params(attributes)
...
 
+      context "when declaration is rejected" do
+        let(:valid_params) do
+          {
+            participant_id: ect_profile.user.id,
+            declaration_type: "started",
+            declaration_date: (ect_profile.schedule.milestones.first.start_date + 2.days).rfc3339,
+            course_identifier: "ecf-induction",
+          }
+        end
```
kicferk1, line 119: I don't think `when declaration is rejected` is a meaningful context here. That statement is the outsome, so it better belongs in the `it` statements.

Also, why do we need different `valid_params` here? Do we actually need this context?

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -108,6 +108,43 @@ def build_params(attributes)
...
+            ect_profile.participant_profile_states.create({ state: "withdrawn", created_at: Time.zone.now - 1.hour })
+          end
+
+          it "returns 422 when participant is withdrawn" do
+            params = build_params(valid_params)
+            post "/api/v1/participant-declarations", params: params
+            expect(response.status).to eq 422
+            expect(response.body).to eq({ errors: [{ title: "Bad or missing parameters", detail: I18n.t(:declaration_on_incorrect_state) }] }.to_json)
+          end
+        end
```
kicferk1, line 132: With the outer context, it reads as "when participant is withdrawn it returns 422 when participant is withdrawn".

Either remove the outside context (and put the before block inside the test), or remove `when participant is withdrawn` from `it` statement please.

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -108,6 +108,43 @@ def build_params(attributes)
...
+            ect_profile.participant_profile_states.create({ state: "deferred", created_at: Time.zone.now - 1.hour })
+          end
+
+          it "returns 422 when participant is withdrawn" do
+            params = build_params(valid_params)
+            post "/api/v1/participant-declarations", params: params
+            expect(response.status).to eq 422
+            expect(response.body).to eq({ errors: [{ title: "Bad or missing parameters", detail: I18n.t(:declaration_on_incorrect_state) }] }.to_json)
+          end
+        end
```
kicferk1, line 145: Similar context / test case naming issue here. Also, `it "returns 422 when participant is withdrawn" do` probably wants to be `it "returns 422 when participant is deferred" do` - `withdrawn` vs `deferred` copy-paste error :)

kicferk1 (approved):

app/services/record_declarations/base.rb
blackrat: Shouldn't the state checking be a method on the profile as opposed to a reach through?

PR #1014, opened 2021-09-14
Cpdrp 937 participant added email

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-937


app/forms/schools/add_participant_form.rb
```diff
@@ -120,12 +121,19 @@ def creators
...
+      ActiveRecord::Base.transaction do
+        profile = creators[participant_type].call(
+          full_name: full_name,
+          email: email,
+          school_cohort: school_cohort,
+          mentor_profile_id: mentor&.mentor_profile&.id,
+        )
+
+        ParticipantMailer.participant_added(participant_profile: profile).deliver_later
+        profile.update_column(:request_for_details_sent_at, Time.zone.now)
```
BroiSatse, line 133: Question - should we put this in the mailer? On one hand, mailer traditionally does not change the state, on the other this column exists in precisely this purpose so it would avoid a bit of duplication around the code?

app/forms/schools/add_participant_form.rb
```diff
@@ -120,12 +121,19 @@ def creators
...
+        ParticipantDetailsReminderJob.schedule(profile)
+        profile
```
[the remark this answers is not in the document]
BroiSatse: I personally find those creator classes a bit confusing - there is a massive amount of repetition between those services and they effectively only differ in one field. Ideally, I'd be looking at this form as a domain service object - the only thing that I'd be thinking of abstracting from it is its multi-steppiness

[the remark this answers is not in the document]
ethanmills: I'm not sure that's a strong enough reason not to use it here. We're ending up with nested transactions at the moment, and I think that's worse.

app/services/validation_beta_service.rb
```diff
@@ -9,7 +9,7 @@ def remind_sits_to_add_participants
 
     School.where(id: empty_school_cohorts.select(:school_id)).includes(:induction_coordinators).find_each do |school|
       school.induction_coordinator_profiles.each do |sit|
-        next if sit.reminder_email_sent_at.present?
+        next if sit.reminder_email_sent_at.blank? || sit.reminder_email_sent_at > time
```
ethanmills, line 12: What's this doing?

ethanmills: How would it be in the future?

BroiSatse: Ah, that should not get in...

ethanmills (approved):

PR #1015, opened 2021-09-15
Remove the analytics database from staging

ethanmills:
Now that we know the approach works, remove the staging analytics database for cost savings


kicferk1 (approved):
While you are at it, can you make sure the schema is up to date? I am seeing 
```
t.boolean "active", default: true
```
added when re-running migrations.

Also, maybe worth adding 
```
ANALYTICS_DB_USERNAME=
ANALYTICS_DB_PASSWORD=
```
to `.env.template`?

PR #1040, opened 2021-09-15
Switch over validation to match against first name only

ethanmills:
Aim to boost automatic validation rate of mentors by only comparing their first name


kicferk1 (approved):
Looks like some tests are a touch fragile...

tonyheadford (commented):
😬  I've just changed a whole load of this to cater for that in my PR

ethanmills:
@tonyheadford how far off do you think yours is? Should I just wait for that?

ethanmills:
I'm going to put this in for now @tonyheadford , shouldn't be too hard to undo

PR #1041, opened 2021-09-16
CPDRP-902 Add pupil premium and sparsity uplifts to participants endpoint

jamesrose:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDRP-902

Following an investigation into whether we need to update the pupil premium data following the initial import earlier in the year, it was concluded that this data was not going to change this cohort.

That means we can start to expose `pupil_premium_uplift` and `sparsity_uplift` to lead providers via the participants endpoint.

- These fields can only be `true` or `false` unless the participant is withdrawn. Is that correct?

### DIFF
--- a/app/serializers/participant_serializer.rb
+++ b/app/serializers/participant_serializer.rb
@@ -71,12 +71,12 @@ class ParticipantSerializer
     user.teacher_profile.ecf_profile.ecf_participant_eligibility&.eligible_status? || nil
   end
 
-  active_participant_attribute :pupil_premium_uplift do
-    nil # TODO: CPDRP-534 - Share when we know we have the correct information
+  active_participant_attribute :pupil_premium_uplift do |user|
+    user.teacher_profile.ecf_profile&.school&.pupil_premium_uplift?(user.teacher_profile.ecf_profile.cohort.start_year)
   end
 
-  active_participant_attribute :sparsity_uplift do
-    nil # TODO: CPDRP-534 - Share when we know we have the correct information
+  active_participant_attribute :sparsity_uplift do |user|
+    user.teacher_profile.ecf_profile&.school&.sparsity_uplift?(user.teacher_profile.ecf_profile.cohort.start_year)
   end
 
   active_participant_attribute :training_status do |user|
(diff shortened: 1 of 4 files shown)


app/serializers/participant_serializer.rb
```diff
@@ -71,12 +71,12 @@ def validated_trn(user)
     user.teacher_profile.ecf_profile.ecf_participant_eligibility&.eligible_status? || nil
   end
 
-  active_participant_attribute :pupil_premium_uplift do
-    nil # TODO: CPDRP-534 - Share when we know we have the correct information
+  active_participant_attribute :pupil_premium_uplift do |user|
+    user.teacher_profile.ecf_profile&.school&.pupil_premium_uplift?(user.teacher_profile.ecf_profile.cohort.start_year)
```
kicferk1, line 75: We have that cached on `participant_profiles` too - or so I think. Should we use that instead?

Same with `sparsity_uplift`.

kicferk1 (approved):
Looks pretty good, question is where the data to surface should come from.

app/serializers/participant_serializer.rb
jamesrose: @kicferk1 @blackrat Why do T&P cache these?

blackrat: Yes. We should always be using the `participant_profiles` value. Once uplift is set on a participant, it remains set on that participant regardless if they change schools or if the school changes.

blackrat: It was a business rule given to us by product.

jamesrose: @blackrat By 'we' do you mean T&P or programme-wide?

jamesrose: Discussed offline: previous cached version is obsolete now.

blackrat: > Discussed offline: previous cached version is obsolete now.

No. That's not the case. I was investigating further. It's not effected by the school change, but it should stay with the teacher if they move schools, so caching is still relevant.

PR #1051, opened 2021-09-16
Export school analytics data to analytics database every hour

ethanmills:
## Ticket and context
I  have been dumping CSV files with this information far too regularly. It's time that stopped.

### Code quality checks

### Gotchas

## External API changes
None

### DIFF
--- a/app/services/analytics/ecf_school_service.rb
+++ b/app/services/analytics/ecf_school_service.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Analytics
+  class ECFSchoolService
+    class << self
+      def update_school_analytics
+        return unless %w[test development production].include? Rails.env
+
+        School.eligible_or_cip_only.includes(
+          :nomination_emails,
+          :induction_coordinators,
+          :school_cohorts,
+          :partnerships,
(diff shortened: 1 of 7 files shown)


app/services/analytics/ecf_school_service.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Analytics
+  class ECFSchoolService
+    class << self
+      def update_school_analytics
+        return unless %w[test development production].include? Rails.env
```
jamesrose, line 7: What's the motivation for this line?

kicferk1: Saving money? 😂 

Do we have analytics database outside of those environments?

ethanmills: yeah, those are the only environments the database exists

ethanmills: And yes, the motivation for that is saving £100 per month per environment 😛

kicferk1: 💰 💰 💰

kicferk1 (approved):

PR #1053, opened 2021-09-16
CPDTP-440 New Banding payment rules

blackrat:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-440

New banding rules are applicable to some lead providers who have been issued with revised target
This change reflects the changes to banding to support these.

1. The variable amount has changed from a fixed calculation of 60% to a variable one depending on the band, so has been moved to the DB.
2. An optional revised target which has been added to the call_off_contract as a 2nd target (with a 2% uplifted top level). NOTE: There is currently no cap on the existing 3rd band, but this may be revised too. The existing code will cope with this too if it becomes something that we set as the max value in the database. (which appears likely)
3. Output display has become conditional depending on whether the provider has contracted a revised target or not.

Viewable to the finance user in the payment breakdown and contract screens.
Rake task for payment calculation has been updated to support this.
Banding strings have been moved to indexes and we're using I18n to look them up for both the rake task and breakdown pages.

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/app/components/finance/ecf/breakdown_summary.html.erb
+++ b/app/components/finance/ecf/breakdown_summary.html.erb
@@ -25,7 +25,7 @@
       <%=recruitment_target %>
     </dd>
   </div>
-  <%- if revised_target.present? %>
+  <% if revised_target.present? %>
     <div class="govuk-summary-list__row">
       <dt class="govuk-summary-list__key">
         <%=t("finance.contract.revised_target") %>
--- a/app/components/finance/ecf/breakdown_summary.html.erb
+++ b/app/components/finance/ecf/breakdown_summary.html.erb
@@ -25,6 +25,16 @@
       <%=recruitment_target %>
     </dd>
   </div>
+  <%- if revised_target.present? %>
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">
+        <%=t("finance.contract.revised_target") %>
+      </dt>
+      <dd class="govuk-summary-list__value">
+        <%=revised_target %>
+      </dd>
+    </div>
+  <% end %>
   <div class="govuk-summary-list__row">
     <dt class="govuk-summary-list__key">
       <%=t("finance.ects") %>
--- a/app/components/finance/ecf/contract.html.erb
+++ b/app/components/finance/ecf/contract.html.erb
@@ -17,6 +17,16 @@
       <%=recruitment_target %>
     </dd>
   </div>
+  <%- if revised_target.present? %>
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">
+        <%=t("finance.contract.revised_target") %>
+      </dt>
+      <dd class="govuk-summary-list__value">
+        <%=revised_target %>
+      </dd>
+    </div>
+  <% end %>
   <div class="govuk-summary-list__row">
     <dt class="govuk-summary-list__key">
       <%=t("finance.contract.uplift_target") %>
--- a/app/components/finance/ecf/contract.html.erb
+++ b/app/components/finance/ecf/contract.html.erb
@@ -17,7 +17,7 @@
       <%=recruitment_target %>
     </dd>
   </div>
-  <%- if revised_target.present? %>
+  <% if revised_target.present? %>
     <div class="govuk-summary-list__row">
       <dt class="govuk-summary-list__key">
         <%=t("finance.contract.revised_target") %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -125,6 +125,11 @@ en:
         floor: 0
         ceiling: Unlimited
     ecf:
+      bands:
+        "0": ECF FIP Band A
+        "1": ECF FIP Band B
+        "2": ECF FIP Band C
+        "3": Additional
       payment_breakdown: ECF Payment Breakdown
       contract_information: ECF Contract Information
     service_fee:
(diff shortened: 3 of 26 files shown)


lib/tasks/payment_breakdown.rake
```diff
@@ -18,8 +18,8 @@ namespace :payment_calculation do
     raise "Unknown lead provider: #{ARGV[1]}" if cpd_lead_provider.nil?
     raise "Not an ECF lead provider #{ARGV[1]}" if cpd_lead_provider.lead_provider.nil?
 
-    total_participants = (ARGV[2] || ParticipantDeclaration::ECF.payable_for_lead_provider(cpd_lead_provider).count)
-    uplift_participants = (ARGV[3] || ParticipantDeclaration::ECF.payable_uplift_for_lead_provider(cpd_lead_provider).count)
+    total_participants = (ARGV[2] || ParticipantDeclaration::ECF.payable_for_lead_provider(cpd_lead_provider).count).to_i
+    uplift_participants = (ARGV[3] || ParticipantDeclaration::ECF.payable_uplift_for_lead_provider(cpd_lead_provider).count).to_i
```
kicferk1, line 22: Should `to_i` just apply to `ARGV[x]`? I think the count scopes are integers already

spec/models/participant_band_spec.rb
```diff
@@ -3,29 +3,99 @@
...
+        expect(bands[1].number_of_participants_in_this_band(2000)).to eq(0)
+        expect(bands[2].number_of_participants_in_this_band(2000)).to eq(0)
+        expect(bands[3].number_of_participants_in_this_band(2000)).to eq(0)
+      end
+
+      it "populates bands[1] if there are enough participants for bands[0] + bands[1]" do
+        expect(bands[0].number_of_participants_in_this_band(2001)).to eq(2000)
+        expect(bands[1].number_of_participants_in_this_band(2001)).to eq(1)
+        expect(bands[2].number_of_participants_in_this_band(2001)).to eq(0)
+        expect(bands[3].number_of_participants_in_this_band(2000)).to eq(0)
```
kicferk1, line 83: Probably want 2001 here, not that it will change the outcome

spec/models/participant_band_spec.rb
```diff
@@ -3,29 +3,99 @@
...
+        expect(bands[2].number_of_participants_in_this_band(4000)).to eq(0)
+        expect(bands[3].number_of_participants_in_this_band(4000)).to eq(0)
+      end
+
+      it "fills all the bands but no more if there are more participants than the revised recruitment target" do
+        expect(bands[0].number_of_participants_in_this_band(10_000)).to eq(2000)
+        expect(bands[1].number_of_participants_in_this_band(10_000)).to eq(2000)
+        expect(bands[2].number_of_participants_in_this_band(10_000)).to eq(500)
+        expect(bands[3].number_of_participants_in_this_band(10_000)).to eq(600)
+      end
```
kicferk1, line 98: What happens to the other 5k participants? In business rules anyway?

db/migrate/20210915155919_add_output_payment_contribution_percentage_to_band.rb
```diff
@@ -0,0 +1,5 @@
+class AddOutputPaymentContributionPercentageToBand < ActiveRecord::Migration[6.1]
+  def change
+    add_column :participant_bands, :output_payment_contribution_percantage, :integer, default: 60
```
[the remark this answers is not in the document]
blackrat: We should have the updated figures at some point later today.

spec/models/participant_band_spec.rb
blackrat: They don't get aggregated. It's a cap. I've asked if we should be capping the regular ones too, since the band[2] in that instance gets 6000. It's currently being discussed, but I think the most likely outcome will be to set the top level band to an agreed target based on the recruitment_target. That'll change the expectation for the non-revised tests.

app/components/finance/ecf/breakdown_summary.html.erb
```diff
@@ -25,6 +25,16 @@
       <%=recruitment_target %>
     </dd>
   </div>
+  <%- if revised_target.present? %>
```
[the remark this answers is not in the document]
blackrat: Shows how long I've been using rails and haml/slim. :) Shouldn't need it.

config/locales/en.yml
```diff
@@ -124,6 +125,11 @@ en:
         floor: 0
         ceiling: Unlimited
     ecf:
+      bands:
+        "0": ECF FIP Band A
+        "1": ECF FIP Band B
+        "2": ECF FIP Band C
+        "3": Additional
```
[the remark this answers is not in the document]
blackrat: Nope. It can go. Band as an index with translation feels a lot cleaner.

lib/tasks/payment_breakdown.rake
blackrat: `to_i` would turn `nil` to `0` which would pass the `truthy` test

app/components/finance/ecf/contract.html.erb
```diff
@@ -17,6 +17,16 @@
       <%=recruitment_target %>
     </dd>
   </div>
+  <%- if revised_target.present? %>
```
kicferk1, line 20: spot the extra `-` 😂

kicferk1 (approved):

PR #1059, opened 2021-09-16
Use new DQT API

asmega:
## Ticket and context

Ticket: https://trello.com/c/a7aV9n50/470-transition-to-new-dqt-api

- We have access to a test environment of the actual DQT API (so not the temporary one)
- You have to call one endpoint to fetch a JWT which is then used as a Bearer token to call the actual API
- Test API tokens have a lifespan of 1 hour. They are cached in memory and only refreshed 5 minutes prior to expiry

## Tech review

- Feature is behind a feature flag, "old" api will continue to be used until toggled on
- Encrypted credentials have been added, please check I've not made any silly typos there

### Code quality checks

### HTML Checks

### Gotchas

- review app will be deployed as `deployed_development` environment
- this means it will connect to test api instance
- remember to toggle the feature on
- there is a google document that contains test data that can be used to test against whilst in a rails console

## External API changes

- none


app/services/feature_flag.rb
```diff
@@ -28,6 +28,7 @@ def feature
     admin_change_materials
     user_research_full_booked
     sit_user_research_fully_booked
+    actual_dqt_api
```
jamesrose, line 31: 😆

asmega: Name suggestions are also welcome before we set it in stone

app/services/dqt_api_access.rb
```diff
@@ -0,0 +1,74 @@
...
+        request.set_form_data(data)
+
+        response = Net::HTTP.start(uri.hostname, uri.port, use_ssl: use_ssl?, read_timeout: 20) do |http|
+          http.request(request)
+        end
+
+        if response.code == "200"
+          @token = JSON.parse(response.body)["access_token"]
+        else
+          raise "DQT access token could not be fetched"
```
jamesrose, line 28: Would we want to raise in this situation? Or fail the validation?

lib/actual_dqt/client.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    def subscription_key
+      Rails.configuration.dqt_api_ocp_apim_subscription_key
+    end
+
+    def uri(trn:, birthdate:, nino: nil)
+      URI("#{Rails.configuration.dqt_api_url}/#{trn}?#{query_string(birthdate: birthdate, nino: nino)}")
+    end
+
+    def query_string(birthdate:, nino: nil)
```
jamesrose, line 49: You can do `Object#to_query` to generate this.

lib/actual_dqt/old_record_mapper.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module ActualDqt
+  # Given a response from new actual DQP api
+  # map to old format for backwards compatability
+  class OldRecordMapper
+    def self.translate(response)
+      return nil if response.nil?
```
jamesrose, line 8:
```suggestion
      return if response.nil?
```

nit: not needed

jamesrose (commented):
Nice. Some initial comments.

app/services/feature_flag.rb
ethanmills: `full`?

lib/actual_dqt/client.rb
asmega: oh nice. didn't actually know that existed

app/services/dqt_api_access.rb
kicferk1: I would like to see it in sentry personally

app/services/dqt_api_access.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      @token
+    end
+
+  private
+
+    def token_about_to_expire?
+      expires_at = Time.zone.at(decoded_token.dig(0, "exp"))
+      Time.zone.now + token_expiry_leeway > expires_at
+    end
```
kicferk1, line 40: Out of curiosity, how long do they last?

app/services/dqt_api_access.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    def decoded_token
+      JWT.decode @token, nil, false
```
kicferk1, line 43: Why do we need to pass `nil, false` here?

app/services/feature_flag.rb
kicferk1: How do the people writing those apis distinguish the two? Maybe we can borrow naming from them?

lib/actual_dqt/client.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    def translate_hash(hash)
+      hash.deep_transform_values do |value|
+        if value =~ /^\d{4}-\d{2}-\d{2}/
+          Date.parse(value)
+        else
+          value
+        end
+      end
+    end
```
kicferk1, line 39: Do we know the names of the fields that this should apply to? Would it be safer to parse based on that, rather than field shape?

app/services/dqt_api_access.rb
asmega: I would of thought we'd want to raise. Any async stuff can be re-queued. Any synchronous calls should know there was an error so we can look into it.

If we fail the validation silently because of technical glitch I feel that would have potentially a more negative impact.

I'm happy to raise a more specific error here and then decide what we want to do in the client

app/services/dqt_api_access.rb
asmega: Token lifespan seems to be one hour

app/services/dqt_api_access.rb
asmega: - `nil` is the key/password used for signing the token. in our case the token isn't signed hence `nil`
- `false` is whether we verify the signature. since the token isn't signed we can't verify it which is why it it set to false

app/services/feature_flag.rb
asmega: I don't see much from outside. the only maybe useful bit are from the hostname which contains `dqt`, `crm4`, `dynamics`

app/services/dqt_api_access.rb
kicferk1: Does `JWT.decode @token` break badly?
```
  def decode(jwt, key = nil, verify = true, options = {}, &keyfinder)
    Decode.new(jwt, key, verify, DEFAULT_OPTIONS.merge(options), &keyfinder).decode_segments
  end
```
is the code for the method. It seems surprising that the default values would break. But if it does break, I guess passing two extra arguments is not the end of the world.

lib/actual_dqt/client.rb
asmega: I did consider that. From the data I've seen i don't expect anything shaped like a date that is not a date so I think the above is fine. I also didn't see a particularly nice way of doing the above for each key with a given name in a nested hash which may also turn out to be optional.

kicferk1: Fair enough

app/services/dqt_api_access.rb
asmega: oh i see your point, yeah looks like i can remove them then, the README says otherwise compared to the implementation

asmega: ok it is needed as i just read the source correctly. default arguments in the code example you pasted are `nil, true` we need `nil, false` so therefore they are needed

spec/services/participant_validation_service_spec.rb
```diff
@@ -249,4 +249,296 @@ def build_validation_result(trn:, options: {})
...
+      before do
+        expect(mock_client).to receive(:get_record).and_return(dqt_record)
+      end
+
+      it "returns record with padded trn when trn is not padded" do
+        expect(validation_result).to eql({
+          trn: padded_trn,
+          qts: true,
+          active_alert: false,
+          previous_induction: true,
```
asmega, line 336: I'm not sure what these extra fields (`previous_induction`, `previous_participation`) are and appeared after i rebased. I just copied the tests across. it would be great if someone can confirm if these values are what they should be.

app/services/dqt_api_access.rb
asmega: ok, this now explicitly sends to sentry, also a custom error is raise for finer grain control later for the caller

lib/actual_dqt/client.rb
asmega: I've swapped this over now

app/services/feature_flag.rb
asmega: I've just mass renamed everything to `FullDqt` and `full_dqt_api`

asmega:
- I'd like to get this merged, the feature is toggled off by default and I can toggle on for each environment one by one
- Sandbox is to remain toggled off until data has been imported

lib/full_dqt/old_record_mapper.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module FullDqt
+  # Given a response from new full DQP api
+  # map to old format for backwards compatability
+  class OldRecordMapper
```
kicferk1, line 6: What's the plan with that? Eventually we remove the old dqt api, and we make everything work with the new format (and remove this class)?

kicferk1 (approved):
Let's see it in action.

lib/full_dqt/old_record_mapper.rb
asmega: yea something like that. this way i don't have to change any code and the feature can be on or off. once we just have just the one there won't be a need to maintain compatibility and we can just change to use the newer structure

PR #1061, opened 2021-09-17
CPDTP-499: ECF participants endpoint

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-499

I opted to route the old index participants action to the new controller in our routes. Happy with suggestions to get it done differently if you have any.

I decided to keep the swagger docs for the old endpoint, cause an undocumented endpoint doesn't feel right. Should we add a deprecation warning or something in them? 

### Code quality checks

2. Hit the new `api/v1/participants/ecf` endpoint - it should behave just like `api/v1/participants` did. and `api/v1/participants` should behave the same as it used to. 

## External API changes

Does this change anything in our external APIs? If so, did you remember to update the CHANGELOG for Lead Providers? (lead-providers/guidance/release-notes) - Added release notes and updated guidance to include the new endpoint.


app/controllers/api/v1/participants_controller.rb
```diff
@@ -6,21 +6,9 @@ module Api
...
-          format.json do
-            participant_hash = ParticipantSerializer.new(paginate(participants)).serializable_hash
-            render json: participant_hash.to_json
-          end
-          format.csv do
-            participant_hash = ParticipantSerializer.new(participants).serializable_hash
-            render body: to_csv(participant_hash)
-          end
-        end
+        redirect_to controller: "ecf_participants", action: "index", params: request.params and return
```
jamesrose, line 11: I can't see any entry point to this. Why do we need to redirect here?

kicferk1: Oh, I forgot to delete it from earlier attempts 😂 Good spot!

swagger/v1/api_spec.json
```diff
@@ -516,6 +516,129 @@
         }
       }
     },
+    "/api/v1/participants/ecf": {
```
jamesrose, line 519: I think we should remove `/api/v1/participants` from the Swagger docs. It's going to be _very_ confusing otherwise :D

jamesrose: We could also add a note here that this is the new endpoint

kicferk1: Hmmm, I could add big `DEPRECATED` letters to the old endpoint docs. 

I guess the alternative would be to say something like "also available on api/v1/participants" on the new endpoint documentation.

I'm worried that in a week a provider who hasn't upgraded yet tries to find docs for the endpoint they are using, and can't :P

mattonfoot: In the API reference we say we will include a DEPRECIATION header, I don't necessarily think we have to right now, just something to note

kicferk1: I just added some clarification to the swagger docs - including mentioning that the old endpoints are deprecated.

@jamesrose do you reckon it makes the intention clear enough? @mattonfoot any other places where we should add deprecation notice?

jamesrose: @kicferk1 IMO I'd remove the docs entirely. The endpoint will still be there and they will already know to move to the new one. Plus, it'll encourage them to move over sooner.

swagger/v1/api_spec.json
```diff
@@ -1602,8 +1602,8 @@
                   "teacher_reference_number": null,
                   "teacher_reference_number_validated": false,
                   "eligible_for_funding": null,
-                  "pupil_premium_uplift": true,
-                  "sparsity_uplift": false,
+                  "pupil_premium_uplift": null,
```
jamesrose, line 1605: These should be true and false. Active participants can't have null values in these fields anymore.

jamesrose (approved):

PR #1070, opened 2021-09-17
Allow eligible status in ECFParticipantEligiblilty

tonyheadford:
## Ticket and context

Enable the `:eligible` status to be set in `ECFParticipantEligibility` instead of forcing the status to `:matched`. Ensure that this is still handled in the validations controller as though it were `:matched` status. This is to enable us to get a picture of who will successfully validate and the different outcomes that will need to be handled.

## Tech review
When we have matched the participants TRN, the qts flag is set and there are no other flags, the eligibility status should be `:eligible`.  In this instance we should delete/not store the validation data, however as this is not complete yet, we need to hang on to the personal data needed to re-run validation against the participant once it is complete.

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/app/controllers/participants/validations_controller.rb
+++ b/app/controllers/participants/validations_controller.rb
@@ -101,7 +101,6 @@ module Participants
       if participant_profile.ecf_participant_eligibility&.eligible_status? ||
           participant_profile.ecf_participant_eligibility&.matched_status?
         # Probably successful - show success message.
-        # TODO CPDRP-884: change line above to be eligible_status?
         # Will become: TRN has been validated, qts, no flags, not done before, no previous induction
         render_completed_page
       else
--- a/app/controllers/participants/validations_controller.rb
+++ b/app/controllers/participants/validations_controller.rb
@@ -98,7 +98,8 @@ module Participants
 
     def complete
       @school = participant_profile.school
-      if participant_profile.ecf_participant_eligibility&.matched_status?
+      if participant_profile.ecf_participant_eligibility&.eligible_status? ||
+          participant_profile.ecf_participant_eligibility&.matched_status?
         # Probably successful - show success message.
         # TODO CPDRP-884: change line above to be eligible_status?
         # Will become: TRN has been validated, qts, no flags, not done before, no previous induction
--- a/spec/services/validate_participant_spec.rb
+++ b/spec/services/validate_participant_spec.rb
@@ -118,7 +118,7 @@ RSpec.describe ValidateParticipant do
             .and_return(eligible_record)
         end
 
-        it "does not save the validation data" do
+        xit "does not save the validation data" do
           service.call(participant_profile: participant_profile, validation_data: request_data)
           expect(participant_profile.reload.ecf_participant_validation_data).to be_nil
         end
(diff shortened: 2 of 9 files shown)


kicferk1 (commented):
Sweet!

Percy changes look a bit surprising - one of them seems spurious, and we need to fix it. The other seems a bit weirder, perhaps some conditional changes what text is being displayed on a dashboard. Could you double check that?

tonyheadford:
> Sweet!
> 
> Percy changes look a bit surprising - one of them seems spurious, and we need to fix it. The other seems a bit weirder, perhaps some conditional changes what text is being displayed on a dashboard. Could you double check that?

Yes, just seen that, will investigate

app/controllers/participants/validations_controller.rb
```diff
@@ -98,7 +98,8 @@ def cannot_find_details
 
     def complete
       @school = participant_profile.school
-      if participant_profile.ecf_participant_eligibility&.matched_status?
+      if participant_profile.ecf_participant_eligibility&.eligible_status? ||
+          participant_profile.ecf_participant_eligibility&.matched_status?
         # Probably successful - show success message.
         # TODO CPDRP-884: change line above to be eligible_status?
```
jamesrose, line 104:
```suggestion
```

spec/services/validate_participant_spec.rb
```diff
@@ -118,7 +118,7 @@
             .and_return(eligible_record)
         end
 
-        it "does not save the validation data" do
+        xit "does not save the validation data" do
```
kicferk1, line 121: Is that going back in when Phil's PR with new dqt api goes in? https://github.com/DFE-Digital/early-careers-framework/pull/1059

kicferk1 (approved):

spec/services/validate_participant_spec.rb
tonyheadford: Yes, once we can access the previous inductees we won't keep the validation data

ethanmills:
I believe this is surfaced in the API. Should it be?

kicferk1:
> I believe this is surfaced in the API. Should it be?

I don't see why not. The providers are interested in whether their participants are eligible for payment, and their verification status is the only thing that makes a difference in it at the moment. So it feels correct to surface that.

ethanmills:
> > I believe this is surfaced in the API. Should it be?
> 
> I don't see why not. The providers are interested in whether their participants are eligible for payment, and their verification status is the only thing that makes a difference in it at the moment. So it feels correct to surface that.

The conversations I heard were that we weren't ready to share this data yet (hence not updating the UI)

kicferk1:
> The conversations I heard were that we weren't ready to share this data yet (hence not updating the UI). Quite possible that something changed though

Uhh, if there is a very good reason not to share it, I would be interested to hear it. But in general I would be very keen to get it moving.

I guess we could be concerned about false negatives, until we get the full api merged, but they don't sound like a massive risk - at the moment the api says nobody is eligible for payment, which surely means more false negatives.

tonyheadford:
> > > I believe this is surfaced in the API. Should it be?
> > 
> > 
> > I don't see why not. The providers are interested in whether their participants are eligible for payment, and their verification status is the only thing that makes a difference in it at the moment. So it feels correct to surface that.
> 
> The conversations I heard were that we weren't ready to share this data yet (hence not updating the UI). Quite possible that something changed though

I think you might be right @ethanmills, we don't want to make it public yet, we're not letting the participants know and we haven't got the previous inductees from the API yet (assuming that will work).  It might be better to hide this in the API for now

ethanmills (approved):
LGTM!

PR #1073, opened 2021-09-17
CPDTP-280 Limit uplift payments

blackrat:
## Ticket and context

Ticket: 

Calculate the total contract value from the participants in the target * the price per band for each band.
Limit the uplift to 5% of this (currently a hardcoded value)

Split out the aggregator from the calculator again for a couple of reasons.

1. It's becoming big enough to be independent and will be more so when it will need to capture "paid" historic numbers to reconcile against future milestones and stories.
2. Doing it now, rather than when 1 becomes an issue make is easier to test the uplift cap without needing to seed the database.

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/app/models/call_off_contract.rb
+++ b/app/models/call_off_contract.rb
@@ -1,8 +1,17 @@
 # frozen_string_literal: true
 
 class CallOffContract < ApplicationRecord
-  has_many :participant_bands
   belongs_to :lead_provider
+  delegate :total_contract_value, to: :bands
+  has_many :participant_bands do
+    def total_contract_value
+      map(&:contract_value).reduce(&:+)
+    end
+  end
+
+  def uplift_cap
+    total_contract_value * 0.05
+  end
 
   def band_a
     bands.first
--- a/spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
+++ b/spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe PaymentCalculator::ECF::Contract::UpliftPaymentCalculations do
+  let(:dummy_class) { Class.new { include PaymentCalculator::ECF::Contract::UpliftPaymentCalculations } }
+  let(:call_off_contract) { create(:call_off_contract) }
+  let(:calculator) { dummy_class.new({ contract: call_off_contract }) }
+
+  context "when condition" do
+    it "returns the uplift_payment_per_participant" do
+      expect(calculator.uplift_payment_per_participant).to eq(100)
+    end
+
+    it "returns the expected uplift_payment_per_participant for a each event type" do
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :started)).to eq(100)
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_1)).to eq(0)
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_2)).to eq(0)
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_3)).to eq(0)
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_4)).to eq(0)
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :completed)).to eq(0)
+    end
+
+    it "returns a capped uplift_payment_for_events when the total exceeds 5% of the total contract value" do
+      expect(calculator.uplift_payment_for_event(uplift_participants: 10_000, event_type: :started)).to eq(99_500)
+    end
+  end
+end
--- a/spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
+++ b/spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
@@ -7,22 +7,20 @@ RSpec.describe PaymentCalculator::ECF::Contract::UpliftPaymentCalculations do
   let(:call_off_contract) { create(:call_off_contract) }
   let(:calculator) { dummy_class.new({ contract: call_off_contract }) }
 
-  context "when condition" do
-    it "returns the uplift_payment_per_participant" do
-      expect(calculator.uplift_payment_per_participant).to eq(100)
-    end
+  it "returns the uplift_payment_per_participant" do
+    expect(calculator.uplift_payment_per_participant).to eq(100)
+  end
 
-    it "returns the expected uplift_payment_per_participant for a each event type" do
-      expect(calculator.uplift_payment_per_participant_for_event(event_type: :started)).to eq(100)
-      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_1)).to eq(0)
-      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_2)).to eq(0)
-      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_3)).to eq(0)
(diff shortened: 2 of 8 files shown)


app/models/call_off_contract.rb
```diff
@@ -1,8 +1,17 @@
...
 
 class CallOffContract < ApplicationRecord
-  has_many :participant_bands
   belongs_to :lead_provider
+  delegate :total_contract_value, to: :bands
+  has_many :participant_bands do
+    def total_contract_value
+      map(&:contract_value).reduce(&:+)
+    end
+  end
```
kicferk1, line 10: What does putting `total_contract_value` in a `has_many` block do?

app/models/call_off_contract.rb
```diff
@@ -1,8 +1,17 @@
...
+
+  def uplift_cap
+    total_contract_value * 0.05
+  end
```
kicferk1, line 14: Can we have a unit test for this? Or a couple?

spec/services/calculation_orchestrator_spec.rb
```diff
@@ -74,6 +87,15 @@
...
   end
+  let(:capped_uplift) do
+    {
+      uplift: {
+        participants: 10_000,
+        per_participant: 100.0,
+        subtotal: 99_500.0,
+      },
+    }
+  end
```
kicferk1, line 98: I don't understand where these values are coming from, and how they relate to the dummy aggregator.

Does this test belong in this file, or should it live in the spec for uplift fees calculation?

kicferk1 (commented):
> Doing it now, rather than when 1 becomes an issue make is easier to test the uplift cap without needing to seed the database.

Maybe it is easier to test the uplift cap in calculation orchestrator without seeding the database, but the real question is... Should we?

Uplift should be unit tested by itself, and the orchestrator checking that uplift works is more of a sanity / integration check. In which case, we can make a contract with a tiny value, which will require minimal seeding to observe the behaviour we want. Maybe with what, 3 target participants or something?

blackrat:
> Maybe it is easier to test the uplift cap in calculation orchestrator without seeding the database, but the real question is... Should we?
As a general principle, the orchestrator shouldn't care where the numbers come from, database or otherwise. Unit testing this shouldn't depend on the db. Whether these tests should be in there or a separate uplift test section is probably more of a point though.

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe PaymentCalculator::ECF::Contract::UpliftPaymentCalculations do
+  let(:dummy_class) { Class.new { include PaymentCalculator::ECF::Contract::UpliftPaymentCalculations } }
+  let(:call_off_contract) { create(:call_off_contract) }
+  let(:calculator) { dummy_class.new({ contract: call_off_contract }) }
+
+  context "when condition" do
```
kicferk1, line 10: hehe, maybe we don't need this context if it doesn't say anything meaningful :)

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    it "returns the expected uplift_payment_per_participant for a each event type" do
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :started)).to eq(100)
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_1)).to eq(0)
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_2)).to eq(0)
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_3)).to eq(0)
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :retained_4)).to eq(0)
+      expect(calculator.uplift_payment_per_participant_for_event(event_type: :completed)).to eq(0)
+    end
```
kicferk1, line 22: Where do those values come from? Is that real business requirement, or test data?

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    it "returns a capped uplift_payment_for_events when the total exceeds 5% of the total contract value" do
+      expect(calculator.uplift_payment_for_event(uplift_participants: 10_000, event_type: :started)).to eq(99_500)
+    end
```
kicferk1, line 26: Ok, so we get 10k participants. The uplift is 99.5k. For that to be 5%, the total contract value needs to be 1990k. 

Can we make those numbers... rounder somehow? Like, set up the test contract so that total value is 1 million, and show that if that's the case, the max uplift is 50k? Or something like that?

As it stands, the test shows that the value matches the expectations of the test writer, but does little to document why it needs to be that.

I guess we could make it more explicit - calculate total contract value in this test, and show that we get 5% of it - rather than hard-coding the value. Thoughts?

spec/models/call_off_contract_spec.rb
```diff
@@ -8,8 +8,16 @@
...
 
-    it "is expected to have band_a with nil as the lowest min value" do
-      expect(call_off_contract.band_a.min).to eq(nil)
+    it "is expected to have band_a with the lowest minimum value" do
+      expect(call_off_contract.band_a.min.to_i).to eq(0)
+    end
+
+    it "is expected to have a total contract value" do
+      expect(call_off_contract.total_contract_value).to eq(2000 * 995) # recruitment_target * per_participant
+    end
```
kicferk1, line 17: Are these the default values from our test suite? Maybe we should do something like `call_off_contract.recruitment_target * call_off_contract.per_participant`? Rather than adding a comment about it?

spec/models/call_off_contract_spec.rb
```diff
@@ -8,8 +8,16 @@
...
+
+    it "is expected to have an uplift cap of 5% of the total contract value" do
+      expect(call_off_contract.uplift_cap).to eq(99_500.0)
```
kicferk1, line 20: Same as above, 99500 doesn't make it easy to mentally verify that it's 5% of total contract value :P

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
blackrat: Indeed.

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
blackrat: These are real. Uplift is only for the started event.

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
blackrat: Why? This is testing something that takes a number and multiplies it by 0.05. I think enough time has been spend on this without building even more testing around it.

spec/models/call_off_contract_spec.rb
blackrat: I did that originally, but then it's just testing both side the same.

spec/lib/payment_calculator/ecf/contract/uplift_payment_calculations_spec.rb
kicferk1: Because when I read

> it "returns a capped uplift_payment_for_events when the total exceeds 5% of the total contract value"

I want to verify that the test body matches the description. And veryfying that `99_500` is 5% of the contract value is not easy.

kicferk1 (approved):

PR #1076, opened 2021-09-20
CPDRP-966 Adjust ect validation reminder emails

FriedSock:
We want to send some more chaser emails to ects who haven't given their details for validation. These should be sent to ECTs who haven't given information for validation, and were created before automation launched. We will send these out in two batches of 5000 (there are just under 10000 meeting the criteria).

## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDRP-966

### Code quality checks

### HTML Checks

### Gotchas

## External API changes


kicferk1 (approved):

app/services/validation_beta_service.rb
```diff
@@ -25,13 +25,19 @@ def remind_sits_to_add_participants
     end
   end
 
+  # This timestamp was generated by finding the last user with request_for_details_sent_at: nil,
+  # after automation this request for details is sent automatically
+  AUTOMATION_LAUNCH_TIME = Time.zone.parse("2021-09-17 09:45:33.284979000 +0000").freeze
```
ethanmills, line 30: Could this just be `3.days.ago`? In theory, anyone covered by automation would have an email sent within the last three days, right?

FriedSock: Because we want to run this twice (two batches of approx 5000), we need this so that we don't send to the same set of users again the second time round.

ethanmills: How come? Does this not also schedule these users for automatic chasers?

FriedSock: Support want a limit on the amount of emails we send a week so they aren't overwhelmed.

Regarding the automatic chasers for these users: my understanding is that it does not, and a series of chasers are sent starting from user creation using `ParticipantDetailsReminderJob`. I can't answer "Should it?" as I wasn't involved in that work, but it seems to be by design.

app/services/utm_service.rb
```diff
@@ -20,14 +20,14 @@ class UTMService
...
     check_ect_and_mentor_info: "check-ect-and-mentor-info",
-    ects_to_add_validation_information: "ects-to-add-validation-information",
     fip_mentors_to_add_validation_information: "fip-mentors-to-add-validation-information",
     cip_mentors_to_add_validation_information: "cip-mentors-to-add-validation-information",
     induction_coordinators_who_are_mentors_to_add_validation_information: "induction-coordinators-who-are-mentors-to-add-validation-information",
     asked_ects_and_mentors_for_information: "asked-ects-and-mentors-for-information",
     sit_to_complete_steps: "sit-to-complete-steps",
     year2020_nqt_invite_school: "year2020-nqt-invite-school",
     year2020_nqt_invite_sit: "year2020-nqt-invite-sit",
+    "2109": "2109",
```
kicferk1, line 30: uhh, what? What is 2109? Can we make it more human readable?

FriedSock: Why does it need to be human readable? I can change it if you feel strongly but this was agreed with Shani.

kicferk1: So the "why" is "for the sanity of whoever uses it" 😂 

This stuff is used in google analytics to monitor how the users got to our service. And if we start making it into magic numbers, someone will have to make a spreadsheet somewhere mapping those numbers to the meaning behind them.

Also, our sanity :D

kicferk1: Also, what is 2109? 21st of September?

kicferk1: If it is, maybe `2021-09-21` or something would convey that more obviously? For a moment there I thought it was a typo of 2019 😂

FriedSock: Yes but there is also a balance to be struct with how long we end up making the url as apparently users are less likely to click on longer urls with very descriptive names. I think `ect-validation-info-2109` is sufficient enough to identify which email it is, and is also being logged in the spreadsheet

PR #1077, opened 2021-09-20
Mock participant validation

asmega:
## Ticket and context

Ticket:

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/app/controllers/api/v1/participant_validation_controller.rb
+++ b/app/controllers/api/v1/participant_validation_controller.rb
@@ -24,7 +24,11 @@ module Api
     private
 
       def validation_service_class
-        ParticipantValidationService
+        if FeatureFlag.active?(:mock_participant_validation)
+          MockParticipantValidationService
+        else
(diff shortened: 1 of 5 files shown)


kicferk1 (commented):
How do we plan to use this?

Seems pretty useful, but some documentation would be welcome. If we plan to enable that on sandbox (so that providers can validate their participants?), we should make sure this is communicated to LPs. If it's just internal then a readme note is probably fine.

asmega:
> How do we plan to use this?
> 
> Seems pretty useful, but some documentation would be welcome. If we plan to enable that on sandbox (so that providers can validate their participants?), we should make sure this is communicated to LPs. If it's just internal then a readme note is probably fine.

I can't tell if it's decided yet. The requirements are rather vague. This might change to a CSV dataset which I don't have yet. But yes this will probably be used in sandbox for validation. This way we control the data rather than depending on data on a third party API which we have no control over

PR #1078, opened 2021-09-20
Change the participant_serializer to use cached uplift

ethanmills:
Context: https://ukgovernmentdfe.slack.com/archives/D021MT22UEL/p1632147075009000

This should also fix the N+1 query on the endpoint: https://sentry.io/organizations/dfe-bat/performance/early-careers-framework:690e901e4b274632af4c4664b578b1cf/?project=5748989&query=transaction.duration%3A%3C15m+transaction.op%3Arails.request+event.type%3Atransaction+http.method%3AGET&statsPeriod=24h&transaction=Api%3A%3AV1%3A%3AParticipantsController%23index&unselectedSeries=p100%28%29

### DIFF
--- a/app/serializers/participant_serializer.rb
+++ b/app/serializers/participant_serializer.rb
@@ -72,11 +72,11 @@ class ParticipantSerializer
   end
 
   active_participant_attribute :pupil_premium_uplift do |user|
-    user.teacher_profile.ecf_profile&.school&.pupil_premium_uplift?(user.teacher_profile.ecf_profile.cohort.start_year)
+    user.teacher_profile.ecf_profile&.pupil_premium_uplift
   end
 
   active_participant_attribute :sparsity_uplift do |user|
-    user.teacher_profile.ecf_profile&.school&.sparsity_uplift?(user.teacher_profile.ecf_profile.cohort.start_year)
+    user.teacher_profile.ecf_profile&.sparsity_uplift
   end
 
   active_participant_attribute :training_status do |user|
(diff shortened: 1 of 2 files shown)


jamesrose (approved):
👍

app/serializers/participant_serializer.rb
```diff
@@ -72,11 +72,11 @@ def validated_trn(user)
   end
 
   active_participant_attribute :pupil_premium_uplift do |user|
-    user.teacher_profile.ecf_profile&.school&.pupil_premium_uplift?(user.teacher_profile.ecf_profile.cohort.start_year)
+    user.teacher_profile.ecf_profile&.pupil_premium_uplift
```
blackrat, line 75: Do we need the guard? Will `ecf_profile` not always have it set, and if not, do we want an `|| false`

app/serializers/participant_serializer.rb
```diff
@@ -72,11 +72,11 @@ def validated_trn(user)
...
   end
 
   active_participant_attribute :sparsity_uplift do |user|
-    user.teacher_profile.ecf_profile&.school&.sparsity_uplift?(user.teacher_profile.ecf_profile.cohort.start_year)
+    user.teacher_profile.ecf_profile&.sparsity_uplift
```
blackrat, line 79: same

app/serializers/participant_serializer.rb
ethanmills: They're `default false not null`

PR #1082, opened 2021-09-21
Guard against unpermitted params in participants api

kicferk1:
## Ticket and context

Someone sent us some unpermitted parameters on participant actions, and we don't want to 500 for that.

2. Try to send a 'withdraw' request for a participant, sending an extra parameter along the way.

### DIFF
--- a/app/services/participants/base.rb
+++ b/app/services/participants/base.rb
@@ -31,6 +31,8 @@ module Participants
     def initialize(params:)
       params.each do |param, value|
         send("#{param}=", value)
+      rescue NoMethodError
+        raise ActionController::UnpermittedParameters, ["Unpermitted parameter: #{param}"]
       end
     end
 
--- a/app/services/participants/base.rb
+++ b/app/services/participants/base.rb
@@ -29,11 +29,9 @@ module Participants
     implement_instance_method :participant_profile_state, :perform_action!, :matches_lead_provider?
 
     def initialize(params:)
-      params.each do |param, value|
-        send("#{param}=", value)
-      rescue NoMethodError
-        raise ActionController::UnpermittedParameters, ["Unpermitted parameter: #{param}"]
-      end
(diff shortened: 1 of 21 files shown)


app/services/participants/base.rb
```diff
@@ -31,6 +31,8 @@ def call
     def initialize(params:)
       params.each do |param, value|
         send("#{param}=", value)
+      rescue NoMethodError
```
ethanmills, line 34: Does this leave us open to some kind of injection? Are there any parameters that could be modified which shouldn't be? Picking out a specific list of parameters would be a more obviously safe way of doing it, but so long as we're confident about this...

app/services/participants/resume/base.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require "abstract_interface"
+
+module Participants
+  module Resume
+    class Base < ::Participants::Base; end
```
ethanmills, line 7: Out of curiosity, why the `::Participants` here and nowhere else?

app/services/participants/withdraw/base.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require "abstract_interface"
+
+module Participants
+  module Withdraw
+    class Base < ::Participants::Base
```
ethanmills, line 7: ditto

app/services/participants/resume/base.rb
kicferk1: I think if you are in the same module, you can short hand it. The `Base` in other places, really is `Participants::Resume::Base`, but since we are in `Participants::Resume` module, it's fine.

I guess `class Base < Participants::Base; end` is good enough for our needs. Not sure what `::` in front of it does. Something clever that also seems to work.

ethanmills (approved):
Much better, thanks!

PR #1084, opened 2021-09-21
Remove all school search code

ethanmills:
I believe this is all dead code. Shout if it isn't


kicferk1 (approved):
🪓 it

asmega:
I believe `app/webpacker/packs/school_search.js` may not be needed?

kicferk1 (approved):
I can't find anything else relating to school_search using a bunch of global search options, so I reckon that gets it all.

asmega (approved):

PR #1089, opened 2021-09-21
Add extra analytics fields

ethanmills:
Chris has asked for some more data on schools to make presenting and slicing the data easier.

I've also started capturing some metrics to aid diagnosing the memory issue


db/schema.rb
```diff
@@ -437,7 +437,6 @@
     t.datetime "updated_at", precision: 6, null: false
     t.integer "output_payment_percantage", default: 60
     t.integer "service_fee_percentage", default: 40
-    t.integer "service_fee_contribution_percentage", default: 40
```
ethanmills, line 440: This was a fresh schema dump. Does this look wrong to anyone?

kicferk1: @blackrat was that something you had on a local branch?

Anyway, I don't see it in any migrations, or used for anything, so I'm fine cutting it from schema.

kicferk1 (approved):

PR #1096, opened 2021-09-21
CPDRP-958: Update chaser for SITs to choose a provider

ethanmills:
## Ticket and context

Ticket: CPDRP-958

Another email ticket. This one had an existing method I adapted. The old template is now being deleted

I've added `delivery_params` so that you can do for example `send_induction_coordinator_choose_provider_chasers(delivery_params: {wait_until: Time.zone.parse("2021-09-22T08:00Z")})`

This probably better than delaying the entire method call, because you won't accidentally email someone lots of times if there's an error and delayed_jobs retried. One thing to bear in mind though is that the information might be out of date by the time the email is sent, so it shouldn't be scheduled too far ahead

### Code quality checks


app/mailers/school_mailer.rb
```diff
@@ -228,17 +228,13 @@ def induction_coordinator_reminder_to_choose_route_email(recipient:, name:, scho
...
       COORDINATOR_REMINDER_TO_CHOOSE_PROVIDER_EMAIL_TEMPLATE,
       to: recipient,
       rails_mailer: mailer_name,
       rails_mail_template: action_name,
-      personalisation: {
-        name: name,
-        school_name: school_name,
-        sign_in_url: sign_in_url,
-      },
+      personalisation: {},
```
kicferk1, line 237: We don't have any scheduled emails that will break if we remove the template variables, right?

kicferk1 (approved):

PR #1097, opened 2021-09-21
CPDTP-475: Schedule change validation

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-475

If there is a pending (unprocessed) declaration, and the provider tries to change schedule, we want them to void it first. The common case is someone having a declaration in September and switching to a January start schedule - we don't think the declaration from September would make much sense for them.

I added `voided` and `not_ voided` scopes to declarations because they were useful in this PR.

### Code quality checks

2. Make a declaration in Septmber for a participant
3. Try changing their schedule to a January one
4. It should fail and tell you why

## External API changes

Does this change anything in our external APIs? If so, did you remember to update the CHANGELOG for Lead Providers? (lead-providers/guidance/release-notes) - I guess I will add release notes

### DIFF
--- a/app/models/finance/schedule.rb
+++ b/app/models/finance/schedule.rb
@@ -7,15 +7,4 @@ class Finance::Schedule < ApplicationRecord
   def self.default
     find_by(name: "ECF September standard 2021")
   end
-
-  def milestone_for_declaration_type
-    {
-      "started" => milestones[0],
-      "retained-1" => milestones[1],
-      "retained-2" => milestones[2],
-      "retained-3" => milestones[3],
-      "retained-4" => milestones[4],
-      "completed" => milestones.last,
-    }
-  end
 end
--- a/app/models/finance/schedule.rb
+++ b/app/models/finance/schedule.rb
@@ -7,4 +7,15 @@ class Finance::Schedule < ApplicationRecord
   def self.default
     find_by(name: "ECF September standard 2021")
   end
+
+  def milestone_for_declaration_type
+    {
+      "started" => milestones[0],
+      "retained-1" => milestones[1],
+      "retained-2" => milestones[2],
+      "retained-3" => milestones[3],
+      "retained-4" => milestones[4],
+      "completed" => milestones.last,
+    }
+  end
 end
--- a/app/services/participants/change_schedule/validate_and_change_schedule.rb
+++ b/app/services/participants/change_schedule/validate_and_change_schedule.rb
@@ -29,6 +29,8 @@ module Participants
       end
 
       def schedule_valid_with_pending_declarations
+        return unless user_profile
+
         declarations = user_profile.participant_declarations.not_voided
         declarations.each do |declaration|
           milestone = schedule.milestone_for_declaration_type[declaration.declaration_type]
--- a/app/services/participants/change_schedule/validate_and_change_schedule.rb
+++ b/app/services/participants/change_schedule/validate_and_change_schedule.rb
@@ -33,12 +33,12 @@ module Participants
 
         declarations = user_profile.participant_declarations.not_voided
         declarations.each do |declaration|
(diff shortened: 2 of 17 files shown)


app/models/finance/schedule.rb
```diff
@@ -7,4 +7,15 @@ class Finance::Schedule < ApplicationRecord
   def self.default
     find_by(name: "ECF September standard 2021")
   end
+
+  def milestone_for_declaration_type
+    {
+      "started" => milestones[0],
```
jamesrose, line 13: Why do milestones not have a `declaration_type`?

app/services/participants/change_schedule/validate_and_change_schedule.rb
```diff
@@ -26,6 +27,22 @@ def schedule
...
         errors.add(:participant_id, I18n.t(:withdrawn_participant)) if participant_profile_state&.withdrawn?
       end
+
+      def schedule_valid_with_pending_declarations
+        return unless user_profile
+
+        declarations = user_profile.participant_declarations.not_voided
+        declarations.each do |declaration|
+          milestone = schedule.milestone_for_declaration_type[declaration.declaration_type]
+          unless milestone.start_date.beginning_of_day < declaration.declaration_date
```
jamesrose, line 37:
```suggestion
          if declaration.declaration_date < milestone.start_date.beginning_of_day
```

These read a bit better if keep the target first and don't do the negation.

app/services/participants/change_schedule/validate_and_change_schedule.rb
```diff
@@ -26,6 +27,22 @@ def schedule
...
+            errors.add(:schedule_identifier, I18n.t(:schedule_invalidates_declaration))
+          end
+
+          unless declaration.declaration_date <= milestone.milestone_date.end_of_day
```
jamesrose, line 41:
```suggestion
          if declaration.declaration_date > milestone.milestone_date.end_of_day
```

What do `start_date` and `milestone_date` represent?

kicferk1: `milestone_date` is the cut-off point for declarations. For the first milestone, it's October 31st. So all declarations for a milestone need to have a date before that. We wouldn't want the declaration date to be after the milestone date, because it would probably mean that the participant is not actually following the schedule they are given.

`start_date` is the other end of that - we don't want declarations to come in too early. For the first milestone it is some arbitraty date, usually 2 months before milestone date. For evety other milestone, it't the day after the end of previous milestone. At least in usual cases.

What it aims to model is "specific declarations should happen at specific time ranges, or the participant is not doing the course on the schedule we expect them to".

app/models/finance/schedule.rb
kicferk1: Hmm, when I was adding them I was not convinced they would have a single declaration type.

Now I know for sure that the real-life milestones won't have a single declaration type. Buuut, the milestones we use for validation should. I think. So I can probably add them.

app/services/participants/change_schedule/validate_and_change_schedule.rb
blackrat: Do we not allow declarations to be made on the milestone date itself? Do they have to be before it?

kicferk1: Well, we have tests (that I might have broken locally) that cover all edge cases.

app/models/finance/schedule.rb
kicferk1: Should be in now. I made a migration to populate it on existing milestones, because there are like 20 of them total, and splitting this into multiple PRs and deploys felt like an overkill. If you feel strongly about it I can re-visit.

jamesrose (approved):

PR #1098, opened 2021-09-21
BUGFIX: prevent NQT+1s from being sent validation invites

ethanmills:
We're emailing NQT+1 s telling them they need to validate. We should not do this. This quick and dirty PR prevents that - it is a temporary solution to buy time for a better fix

### DIFF
--- a/app/services/early_career_teachers/create.rb
+++ b/app/services/early_career_teachers/create.rb
@@ -18,9 +18,13 @@ module EarlyCareerTeachers
 
         ParticipantProfile::ECT.create!({ teacher_profile: teacher_profile, schedule: Finance::Schedule.default }.merge(ect_attributes)) do |profile|
           ParticipantProfileState.create!(participant_profile: profile)
-          ParticipantMailer.participant_added(participant_profile: profile).deliver_later
-          profile.update_column(:request_for_details_sent_at, Time.zone.now)
-          ParticipantDetailsReminderJob.schedule(profile)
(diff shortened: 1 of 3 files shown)


app/services/early_career_teachers/create.rb
```diff
@@ -30,11 +34,12 @@ def call
 
     attr_reader :full_name, :email, :school_cohort, :mentor_profile_id
 
-    def initialize(full_name:, email:, school_cohort:, mentor_profile_id: nil)
+    def initialize(full_name:, email:, school_cohort:, mentor_profile_id: nil, year_2020: false)
       @full_name = full_name
       @email = email
       @school_cohort = school_cohort
       @mentor_profile_id = mentor_profile_id
+      @year_2020 = year_2020
```
kicferk1, line 42: Use `attr_reader` like all the other variables here?

kicferk1 (approved):
Unit test?

tonyheadford (approved):

FriedSock (approved):

app/services/early_career_teachers/create.rb
ethanmills: I'll pick these up tomorrow - for now I just want to stop any more emails being sent. So long as we're happy it will work, that's enough for now

kicferk1: Sure, merge when ready :)

PR #1099, opened 2021-09-21
Redirect NQT+1 to no access page on sign in

ethanmills:
Attempt to prevent NQT+1 from validating by redirecting them to the no-access page on login. They could probably still validate if they knew the correct URL - are we concerned?

I've also cleaned up some of the changes from the related #1098


FriedSock (approved):

ethanmills:
@kicferk1 as promised, some cleanup and unit tests

kicferk1 (approved):

PR #1101, opened 2021-09-22
Experiment with reducing retained memory

ethanmills:
Based on https://github.com/rails/rails/issues/7112#issuecomment-30514153


jamesrose (approved):
🤷

PR #1102, opened 2021-09-22
Increase chaser delay to 7 days

ethanmills:
As mentioned on sprint review, increase chaser delay from 3 to 7 days


kicferk1 (approved):

jamesrose:
@ethanmills This will only take effect after a subsequent re-schedule. Is that the intention?

ethanmills:
> @ethanmills This will only take effect after a subsequent re-schedule. Is that the intention?

I think that's fine, it's not an urgent change

PR #1104, opened 2021-09-22
gem updates

ethanmills:
Pre-empt some of the dependabot updates. The major version bumps here are papertrail and rubocop. I've looked at papertrail and it seems fine.

All the changes here other than rubocop.yml and Gemfile.lock are lint fixes

The Rubocop changes seem fine, except I've disabled the requirement to call super in the initializer because our components just don't need it


kicferk1 (approved):
I like the new rubocop

ethanmills:
> I like the new rubocop

Of course you do, it has strong opinions on whitespace

PR #1105, opened 2021-09-22
BUGFIX: Find correct TRNs for people with 3/4 match

ethanmills:
The DQT has recently added a whole load of people with TRN 0. That breaks our system to find people who gave us the wrong TRN

### DIFF
--- a/app/services/participant_validation_service.rb
+++ b/app/services/participant_validation_service.rb
@@ -85,8 +85,8 @@ private
 
     # If a participant mistypes their TRN and enters someone else's, we should search by NINO instead
     # The API first matches by (mandatory) TRN, then by NINO if it finds no results. This works around that.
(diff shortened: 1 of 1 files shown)


asmega (approved):

kicferk1 (commented):
Where does "1" come from? Is that just an arbitrary value? Also, do we need to fix a bunch of data in our db?

tonyheadford (approved):
Well found

ethanmills:
> Where does "1" come from? Is that just an arbitrary value? Also, do we need to fix a bunch of data in our db?

1 is a value that we don't expect to match anything. It's not from our DB. The API takes a TRN and an optional NINO. It preferentially matches against the TRN, but if there are no results, it matches against NINO. TRN is required, so we need to use a value that won't give any matches

PR #1112, opened 2021-09-22
Add current_user class and id to our logs

kicferk1:
## Ticket and context

When we look at logs, it is useful to know who is causing them. So we can figure out what their state is in the db.

I checked that the hook in application cntroller works.
I still need to test this in a review app, cause our logs work differently in local environment. Should hopefully work just fine.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,6 +14,12 @@ class ApplicationController < ActionController::Base
     head :ok
   end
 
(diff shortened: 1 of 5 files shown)


ethanmills (approved):
This is awesome - looks super useful. Thanks!

asmega:
I think the concept of a `NullUser` may be useful in future. It would remove the need for all those guards

asmega (approved):

PR #1114, opened 2021-09-22
Post release notes in slack on deploy to staging

ethanmills:


kicferk1 (approved):
Let's see it in action

PR #1115, opened 2021-09-22
Email audit

BroiSatse:
This PR introduces automatic email recording - this way we have an easy way see all the emails that has been send in the system without need to go to notify log.


lib/mail/delivery_recorder.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    module MessageExtension
+      def associations
+        @associations ||= []
+      end
+
+      def associate_with(object, as: nil) # rubocop:disable Naming/MethodParameterName
+        associations << [object, as]
+        self
+      end
```
kicferk1, line 52: Please name this method something that does not get easily confused with the method that creates stuff in our database. Maybe `with_associated_object`?

lib/mail/delivery_recorder.rb
```diff
@@ -0,0 +1,55 @@
...
+          email.associate_with object, as: name
+        end
+      end
+    end
+
+  private
+
+    def enabled?
+      !!@enabled
+    end
```
kicferk1, line 42: What is this preventing?

app/models/email.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  UNDEFINED = Object.new
+  private_constant :UNDEFINED
+
+  scope(:associated_with, lambda do |object, as: UNDEFINED|
+    association_scope = Association.where(object: object)
+    association_scope = association_scope.where(name: as) unless as == UNDEFINED
+
+    where(id: association_scope.select(:email_id))
+  end)
```
kicferk1, line 19: 1. Why do we need an undefined object? Can't we use `as: nil` and have a guard `if as`? 
2. That scope brackets look pretty awkward, especially closing one after an `end`. Can we get rid of that?

app/models/email.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class Email < ApplicationRecord
+  class Association < ApplicationRecord
+    belongs_to :email
+    belongs_to :object, polymorphic: true
+  end
```
kicferk1, line 7: Should it be in its own file? And maybe called `EmailAssociation`?

app/models/email.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class Email < ApplicationRecord
+  class Association < ApplicationRecord
+    belongs_to :email
+    belongs_to :object, polymorphic: true
```
jamesrose, line 6: Is this ever not a user?

tonyheadford (commented):
A suggestion as I looked into doing pretty much the same for the ticket I'm working on (but have parked pending this PR) but my solution was a lot less complex...
1. Install an ActiveMailer observer that records outgoing emails. My intention was not to associate with a user but store the recipient(s) email address(es), the mailer class and method, notify id and template if we need it etc. No changes needed to any existing or future mailer.
2. Assuming it works, we already have a notify callback controller so use that to look up the notify Id and reconcile whether the mail was delivered or not.
3. If you need to find out which emails we sent to a particular user you'd look them up by the user's email from the "audit" table.
I don't know if that matches the expectation of what this was intended to do, but it sems way simpler?

lib/mail/delivery_recorder.rb
```diff
@@ -0,0 +1,55 @@
...
+          id: response.id,
+          from: response.content["from_email"],
+          to: mail.original_to,
+          template_id: response.template["id"],
+          template_version: response.template["version"],
+          uri: response.uri,
+          personalisation: mail.header["personalisation"].unparsed_value,
+        )
+
+        email.associate_with(*User.where(email: email.to).to_a)
```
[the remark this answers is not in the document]
BroiSatse: Email can have multiple receivers - we don't have a case like that yet, but there's no reason not to cover this (however i left out cc and bcc fields for now).

lib/mail/delivery_recorder.rb
BroiSatse: But this exists solely to create stuff in our database

lib/mail/delivery_recorder.rb
BroiSatse: We (probably) don't want to record emails in test environment - we use TestDeliveryMethod so we cannot record all the fields. Additionally it would slow down the suite. If at any point we want to test the recording, there are number of tweaks to do this and I'd do it per-tests or per-context only.

lib/mail/delivery_recorder.rb
kicferk1: It exists to pass the data to the `delivery_recorder`. It so happens that the recorder uses the method which is named the same to store stuff in the database at a later, asynchronous time.

jamesrose:
Here's how we did it on GHWT:

- https://github.com/DFE-Digital/get-help-with-tech/blob/main/app/controllers/notify_callbacks_controller.rb#L12
- https://github.com/DFE-Digital/get-help-with-tech/blob/main/db/schema.rb#L178

app/models/email.rb
BroiSatse: 1. `as` is optional - `associated_with(object, as: ni)l` means "give me emails that has been associated without any name` while `associated_with(object)` means "give me all emails, regardless of association name".
2. Yeah, I was very unhappy about this as well, but this is how RuboCop want to see it becasue:
  - `lambda` must be used over `->` for multiline blocks (don't really agree with it!) 
  - `scope :associated_with, lambda do |v|` passes block with `scope` method rather than `lambda` (since `lambda` is a method and not a ruby code token) - hence the need for brackets
Happy to disable rubocop lambda check here!

app/models/email.rb
BroiSatse: I see this record as a private implementation detail of `Email` logic that should not be used in the application code. As such I'd prefer to live in a single file with `Email` - our `model` folder is already unnecessary bloated.

app/models/email.rb
BroiSatse: We can associate a single email to multiple objects in the database, depending on requirements. In this PR, some emails are associated to ParticipantProfiles as well - this can be useful if we want to check how many emails has been sent in regards to a particular ParticipantRecord. I believe there is already a card in JIRA to display `request_for_details` email status in admin panel (including `bounced/delivered`)

lib/mail/delivery_recorder.rb
BroiSatse: It is not asynchronous - `deliver_later` executes a mailer method and delivery in a single go. Somehow `mail(...).associate_with(object)` reads better to me than `mail(...).with_associated_object`. I am also not certain how this could be confused with `the method that creates stuff in our database` as we don't really have any method like that anywhere. Happy to get it under voting though :)

kicferk1:
```
  def associate_with(*objects, as: nil) # rubocop:disable Naming/MethodParameterName
    objects.each do |object|
      Association.create!(
        email: self,
        object: object,
        name: as,
      )
    end
  end
```
  vs 
```
      def associate_with(object, as: nil) # rubocop:disable Naming/MethodParameterName
        associations << [object, as]
        self
      end
```
The first one is in the email class and 'creates stuff in our database'. The second one is here. It took me longer than I wanted to understand which one is called where exactly :P

BroiSatse:
@tonyheadford - This is pretty much exactly how this is written here? There are just one extras thing here - ability to associate emails with other db records. I have left the notify controller changes to do separately in the PR that actually requires it, but I'm happy to add it here as well.

lib/mail/delivery_recorder.rb
kicferk1: I mean the `!!` - is `if enabled?` ever different than `if @enabled`?

app/models/email.rb
kicferk1: > `associated_with(object)` means "give me all emails, regardless of association name"

You mean the same as `Email.all`?

BroiSatse: No, it will return all emails that are associated with an object...

tonyheadford:
> @tonyheadford - This is pretty much exactly how this is written here? There are just one extras thing here - ability to associate emails with other db records. I have left the notify controller changes to do separately in the PR that actually requires it, but I'm happy to add it here as well.

Don't think there's the need for the polymorphic associations and having to change the mailers to do the `associate_with` etc. just seems a bit over complicated?

app/models/email.rb
tonyheadford: I think that would only ever need to be for the `User` regarding whether their invite has been delivered or not? (sounds like the ticket I was working on)

kicferk1: Schools has a contact email, and for some emails we might be more interested that we delivered them to a school than to a user

BroiSatse:
@tonyheadford - There are a few good reasons to associate emails to objects. We need a link between user and email in case user change their email. Ability to display status of the `request_for_details` email for participant profile would be another one that we have right now. Finding this without associations would be a lot of complexity within a core domain space, while here we have complexity within supporting domain. We cannot escape complexity, but we can decide where it lives - I prefer more complex implementation of supporting classes in exchange of much simpler core logic. Also note that the polymorphic association lives on a "private" class and is merely an implementation detail that can change if we decide to do so.

app/models/email.rb
tonyheadford: Do we care? Is storing the email enough?

lib/mail/delivery_recorder.rb
BroiSatse: I have a strong phobia of using instance variables in the code as I prefer separation between logic and data. Instance variable `@enabled` is pure data, `enabled?` is logic (which currently encapsulates the variable). That logic can change in the future and ensuring the code is using `enabled?` instead of `@enabled` minimises the amount of refactor required.

app/models/email.rb
BroiSatse: @tonyheadford Email addresses on records can change, we need a persisted link between email and the reason why that email has been sent.

lib/mail/delivery_recorder.rb
kicferk1, line 30: Should we add `as: "recipient"` or something like that here?

app/models/email.rb
kicferk1: Right, ok. So we would need the name when the same object has been associated as two different roles for some emails. 

And one edge case is we want emails where the object is a specific role - then we pass a non-default name.

Another edge case is we want all emails for the object - and at the moment we pass in nothing, and the default value does the trick.

What I struggle to understand is: why would we want the emails where the association name is specifically `nil`. Surely each association we create has a meaningful name - or at least we can give it some name that is better than nothing. 

So I guess it boils down to "should we make `name` of the association not nullable, and when we receive a nil name in this scope just return everything for this object"?

tonyheadford:
> @tonyheadford - There are a few good reasons to associate emails to objects. We need a link between user and email in case user change their email. Ability to display status of the `request_for_details` email for participant profile would be another one that we have right now. Finding this without associations would be a lot of complexity within a core domain space, while here we have complexity within supporting domain. We cannot escape complexity, but we can decide where it lives - I prefer more complex implementation of supporting classes in exchange of much simpler core logic. Also note that the polymorphic association lives on a "private" class and is merely an implementation detail that can change if we decide to do so.

As per the lengthy Slack discussion - I don't think we need to associate with a profile. We can store an association to the `User`, the email address used for the recipient, the mailer class and mail type, Notify template etc. and the nature of the query we perform would provide the context for which profile we mean wouldn't it?  This removes the need for all the `associate_with` and it could all be handled in a simple observer.  Queries might be what is the status of the most recent participant request for details email sent to this user?

BroiSatse:
@tonyheadford - I still disagree with this. IMO, this is important not only to record the receiver of the email but also an object that given email has been sent for. ParticipantProfile is only one example of this - since a single user can have multiple profiles for various reasons (withdrawn and added; moved school; was ect and then became mentor; etc) there would be no way of answering the question "which profile was this email sent for", and even answering "what email has been sent in connection with this profile" would become much more complicated that the code in this PR. There are many instances when we should link email to other objects as well - all emails sent to school contact details have no matching user to link to and can change in time so they should be linked either to school or even school_cohort (or both - the code here is flexible enough to allow multiple links). All emails to STI should be linked to appropriate SchoolCohorts. This linking might not be required by any feature in the system, but will make debugging email issues much easier.

Link between email and object is an essential data that cannot be derived from the other data in the system. Essential data is always a source of essential (unavoidable) complexity, and I prefer that complexity to be encapsulated within an afferent, highly stable module. Additionally, since this data cannot be accurately restored later, it is better to start collect it early and drop it later if we find it to make code too complicated - however as it is all encapsulated and afferent, it should have minimal effect on the complexity of the core domain.

lib/mail/delivery_recorder.rb
BroiSatse: Hmmm, the idea behind `as` in association was to give the name to the email itself, not to the associated object role for that email... But having both would be actually nice, especially as the email can be, theoretically, associated with two object of the same class but in different role (for example a hypothetical email from one User regarding another user).

How would you see the following change in the interface here:
```
template_mail(...).tag(:request_for_participant_details).associate_with(user, as: :recipient).associate_with(profile, as: :participant_profile)
```

Justification - `tag` would be used to describe the content of the email and would be stored on the Email record. For example, we have two `request_for_participant_details` email - the initial invitation to the system. We might want to see all the request we sent, but the initial email might have additional `invitation` tag (so `tags` would be an array). Association itself would only store the role given object plays for the object (so 'recipient` `school_cohort` etc) - it is non-nullable and (probably) defaults to an underscored class name of the object. When email is displayed in admin we could then display all the links to associated objects.

lib/mail/delivery_recorder.rb
BroiSatse: AH, right! In this case I'd rename the method on the Email record to `create_association_with` to make the database operation more prominent.

app/models/email.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def create_association_with(*objects, as: nil) # rubocop:disable Naming/MethodParameterName
+    objects.each do |object|
+      Association.create!(
+        email: self,
+        object: object,
+        name: (as || object.model_name.singular),
+      )
+    end
+  end
```
kicferk1, line 25: Doesn't look like we need `UNDEFINED` anymore.

Also, do we need `name` for anything on association? I don't see it used in any of the Email internals anymore?

lib/mail/delivery_recorder.rb
```diff
@@ -0,0 +1,65 @@
...
+    module MessageExtension
+      def associations
+        @associations ||= []
+      end
+
+      def tags
+        @tags ||= []
+      end
+
+      def associate_with(object, as: nil) # rubocop:disable Naming/MethodParameterName
```
kicferk1, line 54: Why is rubocop unhappy? Doesn't like `as`? Perhaps instead of disabling it, we should give it a longer name, `associate_as`?

BroiSatse: Not sure about this. `as` is a very common keyword, even rails uses them in quite a few places and IMO it just fits here perfectly.

app/models/email.rb
BroiSatse: `name` is used within delviery_recorder: `create_association_with(*User.where(email: email.to).to_a, as: :recipient)`. It is to describer the relation between given object and the email.

kicferk1 (approved):
I am reasoably confident that it will work and let us deal with requests of "send a bunch of emails to those users, but only if we sent them some other emails in the past" in a more graceful manner than we had to so far. So overall it's an improvement over our current setup.

I am not confident we will ever have great use for emails being associated with any object in the db (like 95% of our tables come nowhere near that). And I am not really sure we need to name the association - other than "recipient", what are we really planning to put in there?

Likewise, I am not convinced recording tags is that good of an idea. I guess they could be useful if we matched them against the campaigns that we use to know how effective our emails are at getting users onto the service, but I struggle to see the use for them in their current format. Maybe we should add them to a bunch more emails to make it more obvious what their usefulness is.

So in short, this PR does the job we are sure we want, and a bunch more things. That bunch of more things is I guess why the non-technical conversations on this PR took place. I reckon for the next purely technical task adding functionality, we should either make the solution minimal, or make an explicit ticket and get it through refinement to make sure we really want all the features that will take a long time implemening. Or at least have an in-depth chat on the dev channel about it.

PR #1122, opened 2021-09-23
CPDRP-199: Add step to clear S3 bucket

ethanmills:
CPDRP-199
Terraform can't delete non-empty buckets for us

- Install CF CLI
- Sign in 
- Create S3 bucket credentials
- Set AWS keys as env variables
- Wipe bucket
- Delete credentials

### DIFF
--- a/.github/workflows/destroy.yml
+++ b/.github/workflows/destroy.yml
@@ -16,6 +16,28 @@ jobs:
         with:
           terraform_version: 0.14.0
 
+      - name: Clear S3 bucket
+        env:
+          CF_USERNAME: ${{ secrets.GOVPAAS_DEV_USERNAME }}
+          CF_PASSWORD: ${{ secrets.GOVPAAS_DEV_PASSWORD }}
+        run: |
+          wget -q -O - https://packages.cloudfoundry.org/debian/cli.cloudfoundry.org.key | sudo apt-key add -
+          echo "deb https://packages.cloudfoundry.org/debian stable main" | sudo tee /etc/apt/sources.list.d/cloudfoundry-cli.list
+          sudo apt-get update
+          sudo apt-get install cf7-cli
+          cf api https://api.london.cloud.service.gov.uk
+          cf auth
+          cf target -o "dfe" -s "earlycareers-framework-dev"
+          cf create-service-key ecf-csv-review-pr-${{ github.event.number }} ecf-csv-review-pr-${{ github.event.number }}-key -c '{"allow_external_access": true}'
+          CREDS=$(cf service-key ecf-csv-review-pr-${{ github.event.number }} ecf-csv-review-pr-${{ github.event.number }}-key)
+          echo $CREDS
+          export AWS_ACCESS_KEY_ID=$(echo $CREDS | tail -7 | jq -r '.aws_access_key_id')
+          export AWS_SECRET_ACCESS_KEY=$(echo $CREDS | tail -7 | jq -r '.aws_secret_access_key')
+          BUCKET_NAME=$(echo $CREDS | tail -7 | jq -r '.bucket_name')
+          aws s3 rm s3://$BUCKET_NAME --recursive
+          cf delete-service-key ecf-csv-review-pr-${{ github.event.number }} ecf-csv-review-pr-${{ github.event.number }}-key -f
+
+
       - name: Destroy review app
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}


.github/workflows/destroy.yml
```diff
@@ -16,6 +16,27 @@ jobs:
...
+          cf api https://api.london.cloud.service.gov.uk
+          cf auth
+          cf target -o "dfe" -s "earlycareers-framework-dev"
+          cf create-service-key ecf-csv-review-pr-${{ github.event.number }} ecf-csv-review-pr-${{ github.event.number }}-key -c '{"allow_external_access": true}'
+          CREDS=$(cf service-key ecf-csv-review-pr-${{ github.event.number }} ecf-csv-review-pr-${{ github.event.number }}-key)
+          AWS_ACCESS_KEY_ID=$(echo $CREDS | tail -7 | jq -r '.aws_access_key_id')
+          AWS_SECRET_ACCESS_KEY=$(echo $CREDS | tail -7 | jq -r '.aws_secret_access_key')
+          BUCKET_NAME=$(echo $CREDS | tail -7 | jq -r '.bucket_name')
+          aws s3 rm s3://$BUCKET_NAME --recursive
+          cf delete-service-key ecf-csv-review-pr-${{ github.event.number }} ecf-csv-review-pr-${{ github.event.number }}-key -f
```
kicferk1, line 37: Is there no way to do it via cf? https://cli.cloudfoundry.org/en-US/v7/delete-service.html

ethanmills: https://docs.cloud.service.gov.uk/deploying_services/s3/#:~:text=you%20must%20delete%20all%20objects%20from%20the%20s3%20bucket%20before%20you%20can%20delete%20the%20bucket%20itself

ethanmills: Looks like TVS do something different, but I cannot find any documentation on this https://github.com/DFE-Digital/teaching-vacancies/blob/30956313bee65abcc894216a8c915215ba86fd49/terraform/app/modules/paas/main.tf#L30

ethanmills: Aha, found it https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket#force_destroy

ethanmills: I'm not sure we can use that. They use AWS credentials and an AWS provider and provision the bucket directly. We have do everything through PaaS

kicferk1: Well, if it doesn't work out then this approach is fine

ethanmills: https://github.com/DFE-Digital/teaching-vacancies/blob/30956313bee65abcc894216a8c915215ba86fd49/terraform/app/modules/paas/versions.tf#L4

ethanmills: That won't work for us. We don't have access to AWS in the same way they do

.github/workflows/destroy.yml
```diff
@@ -16,6 +16,27 @@ jobs:
...
+          wget -q -O - https://packages.cloudfoundry.org/debian/cli.cloudfoundry.org.key | sudo apt-key add -
+          echo "deb https://packages.cloudfoundry.org/debian stable main" | sudo tee /etc/apt/sources.list.d/cloudfoundry-cli.list
+          sudo apt-get update
+          sudo apt-get install cf7-cli
+          cf api https://api.london.cloud.service.gov.uk
+          cf auth
+          cf target -o "dfe" -s "earlycareers-framework-dev"
+          cf create-service-key ecf-csv-review-pr-${{ github.event.number }} ecf-csv-review-pr-${{ github.event.number }}-key -c '{"allow_external_access": true}'
+          CREDS=$(cf service-key ecf-csv-review-pr-${{ github.event.number }} ecf-csv-review-pr-${{ github.event.number }}-key)
+          AWS_ACCESS_KEY_ID=$(echo $CREDS | tail -7 | jq -r '.aws_access_key_id')
```
kicferk1, line 33: For my understanding's sake, what's `tail -7` and `jq -r` doing here exactly?

kicferk1 (approved):

kicferk1:
I guess the proof of the eating is in the pudding, or something like that. Wanna upload a csv of partnerships to the review app and see if closing this PR will successfully destroy it? :)

.github/workflows/destroy.yml
ethanmills: service-key gives
```
{
 "aws_access_key_id": "xxx",
 "aws_region": "eu-west-2",
 "aws_secret_access_key": "xxx",
 "bucket_name": "xxx",
 "deploy_env": ""
}
```
(with the blank line at the start)
`tail` takes the last 7 lines here, without the blank line so that `jq` doesn't complain
`jq` is for json, and pulls out the values. We need `-r` for raw, so that the values aren't wrapped in quotes

PR #1134, opened 2021-09-24
Layout update to cohort dashboard

ericaporter:
## Ticket and context

CPDRP-913 has already been deployed, this is an update relating to this ticket.

Design requested that the cohort dashboard is a summary list and not a table.

The difference this will make will be the actions will be right-aligned, and will improve the spacing. 

It was possible to revert to two-thirds column width and no wrap was no longer required.
Ticket:

## Tech review
The table in schools/dashboard/show.html.erb has been updated to being 

### HTML Checks
- [x] All new pages have automated accessibility checks (checks added under a previous ticket)

### DIFF
--- a/app/views/schools/dashboard/show.html.erb
+++ b/app/views/schools/dashboard/show.html.erb
@@ -7,7 +7,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <span class="govuk-caption-xl"><%= @school.name %></span>
-    <h1 class="govuk-heading-xl govuk-!-margin-bottom-7">Manage your training</h1>
+    <h1 class="govuk-heading-xl">Manage your training</h1>
 
     <% @school_cohorts.each do |school_cohort| %>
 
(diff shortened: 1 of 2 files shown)


app/views/schools/dashboard/show.html.erb
```diff
@@ -5,66 +5,64 @@
...
+          <dt class="govuk-summary-list__key">Programme</dt>
+          <dd class="govuk-summary-list__value"><%= t(school_cohort.induction_programme_choice, scope: %i[manage_your_training induction_programmes]) %></dd>
+          <dd class="govuk-summary-list__actions"><%= govuk_link_to "View details", schools_cohort_path(cohort_id: school_cohort.cohort.start_year), class: 'app-govuk-link' %></dd>
+        </div>
+        <% if school_cohort.core_induction_programme? %>
+          <div class="govuk-summary-list__row">
+            <dt class="govuk-summary-list__key">Materials</dt>
+            <dd class="govuk-summary-list__value"><%= school_cohort.core_induction_programme&.name %></dd>
+            <% if school_cohort.core_induction_programme_id? %>
+              <dd class="govuk-summary-list__actions"><%= govuk_link_to 'Change', schools_core_programme_materials_path(cohort_id: school_cohort.cohort.start_year) %></dd>
```
ethanmills, line 42: How do these play with screenreader? Do they need some visually hidden helper text?

ethanmills (approved):

app/views/schools/dashboard/show.html.erb
ericaporter: Hi @ethanmills, Screenreader handles the summary-list fine. I've checked elsewhere in the app and can't see we are using any hidden helper text with summary-lists - but if you think this should go in, point me in the direction of an example we should follow

jamesrose (approved):

PR #1135, opened 2021-09-24
T&P: Some tech debt work on tests

kicferk1:
## Ticket and context

I got paranoid about some edge cases and while I was at it, decided to do some tech debt on our unit tests.

I might do the same for participant actions, there is a lot of duplication and possibly missing small edge cases there too. 

### Code quality checks

### DIFF
--- a/spec/support/shared_examples/participant_declaration_support.rb
+++ b/spec/support/shared_examples/participant_declaration_support.rb
@@ -1,0 +1,122 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples "a participant declaration service" do
+  context "when lead providers don't match" do
+    it "raises a ParameterMissing error" do
+      expect { described_class.call(given_params.merge(cpd_lead_provider: another_lead_provider)) }.to raise_error(ActionController::ParameterMissing)
+    end
+  end
+
+  let(:params_with_different_date) do
+    given_params.merge({ declaration_date: (cutoff_start_datetime + 1.day + 1.second).rfc3339 })
+  end
+
+  it "creates a participant and profile declaration" do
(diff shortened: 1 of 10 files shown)


app/services/record_declarations/base.rb
```diff
@@ -21,8 +21,8 @@ class Base
     delegate :schedule, :participant_declarations, to: :user_profile, allow_nil: true
 
     class << self
-      def call(params)
-        new(params).call
+      def call(params:)
+        new(params: params).call
```
plisovin, line 25: What is the benefit to have named parameter here?

kicferk1: Consistency with the participants service.

And the benefit of that is... we can have shared examples between them to show the behaviour for "when user is not a participant" or "when user does not have access to the course".

It's the difference of `described_class.call(params)` vs `described_class.call(params: params)` in tests.

And, well, I opted for making it a named parameter rather than making participant actions one an un-named parameter because I thought that would lead to fewer changes.

spec/support/shared_examples/participant_declaration_support.rb
```diff
@@ -0,0 +1,127 @@
...
+      travel_to cutoff_end_datetime + 1.day
+    end
+
+    it "raises ParameterMissing error" do
+      params = given_params.merge({ declaration_date: (cutoff_end_datetime + 1.day).rfc3339 })
+      expect { described_class.call(params: params) }.to raise_error(ActionController::ParameterMissing)
+    end
+  end
+
+  it "fails when user profile is a withdrawn record" do
```
plisovin, line 123: Another use case is ```deferred``` profile

kicferk1: Good spot :)

plisovin (approved):

PR #1139, opened 2021-09-27
CPDRP-978: Update chaser emails

ethanmills:
## Ticket and context

Ticket: CPDRP-978

Prepare to send next round of validation invites
- Remove loads of old mailer code
- Add a condition to the SIT mentor email
- Update UTM campaigns


app/services/validation_beta_service.rb
```diff
@@ -50,13 +50,14 @@ def tell_ects_to_add_validation_information(limit: VALIDATION_CHASER_BATCH_LIMIT
...
-        # a limit that will override any one we provide
-        sent += 1
-        break if sent == limit
-      end
+      # We cannot use a limit on the sql query because find_each/find_in_batches uses
+      # a limit that will override any one we provide
+      sent += 1
+      break if sent == limit
+    end
+    Rails.logger.info "Send #{sent} emails to ECTs"
```
kicferk1, line 60: Don't know what's going on with whitespace here, but it is not pretty. `send_ects_to_add_validation_information` should be indented in relation to `).find_each do |profile|`, right?

ethanmills: Rubymine is convinced this is correct, and I can't find a way to change it's mind. Corrected manually

kicferk1 (approved):

PR #1143, opened 2021-09-27
Handling school funded fip in admin

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-936


app/components/admin/schools/cohorts/other_info.rb
```diff
@@ -0,0 +1,30 @@
...
+        MESSAGES = {
+          design_your_own: "designing own induction course",
+          school_funded_fip: "school funded full induction programme",
+          no_early_career_teachers: "no ECTs this year",
+        }.freeze
+
+        def message
+          return "No programme" if school_cohort.nil?
+
+          ["Not using service", MESSAGES[school_cohort.induction_programme_choice.to_sym]].compact.join(" - ")
```
kicferk1, line 25: `.compact.join(" - ")` is nice and all, but would 
```
"Not using service - #{MESSAGES[school_cohort.induction_programme_choice.to_sym]}"
```
be a bit easier to read?

kicferk1: Also, dies it deserve a component unit test?

app/components/admin/schools/cohorts/other_info.html.erb
```diff
@@ -4,13 +4,13 @@
...
-      No programme
+      <%= message %>
     </dd>
     <dd class="govuk-summary-list__actions">
       <% if FeatureFlag.active?(:admin_change_programme) %>
         <%= govuk_link_to "Change", admin_school_change_programme_path(id: cohort.start_year) %>
-      <% end %>
-      <span class="govuk-visually-hidden"> induction programme</span>
+        <span class="govuk-visually-hidden"> induction programme</span>
+    <% end %>
```
asmega, line 13: minor thing, indentation is wrong

app/components/admin/schools/cohorts/other_info.rb
BroiSatse: There's small difference in behaviour though - in case `MESSAGES[key]` returns nil, `.compact.join` will return `"Not using service"`, while string interpolation will yield `Not using service - ". Probably not a big issue though, but the idea is for that component to handle all sort of future types as well.

As per component test - probably should! Will add :)

kicferk1: Hmmm, good point. I wonder if there is some way we could guard against merging code which adds a new option without adding a message here.

Oh well, that might be reason enough to keep the `compact.join` setup.

ethanmills:
Looks like we're just missing a component test, then we'll be good to go

app/components/admin/schools/cohorts/other_info.rb
```diff
@@ -0,0 +1,30 @@
...
+          @cohort = cohort
+          @school_cohort = school_cohort
+        end
+
+      private
+
+        attr_reader :school_cohort, :cohort
+
+        MESSAGES = {
+          design_your_own: "designing own induction course",
```
ethanmills, line 17:
```suggestion
          design_our_own: "designing own induction course",
```

spec/components/admin/schools/cohorts/other_info_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  context "without school cohort" do
+    let(:school_cohort) { nil }
+
+    it { is_expected.to have_content "No programme" }
+  end
+
+  context "with design your own school cohort" do
+    let(:school_cohort) { instance_double SchoolCohort, induction_programme_choice: "design_your_own" }
```
ethanmills, line 29: If we had used a real class here this would have caught the `design_your_own` vs `design_our_own` issue

tonyheadford (commented):
@BroiSatse We've spotted another issue in that if a SIT chooses "View details" of their programme from the dashboard it shows them the "programme_choice" page that says they are funded by DfE which is wrong.  Not sure if you want to handle that here too?

BroiSatse:
@tonyheadford Ye, I've spotted it a while back as well - this is caused by `if school_cohort.cori_induction_programme? / else ` type of statement - there are quite a few issues like that there (for example, we keep displaying "View your ECTs and Mentors" link for opt-out cohorts, including "No ECTs") so I'd fix it in a separate PR

BroiSatse:
@tonyheadford - Ah, I see that one has been fixed already! Anyway, I'll fix this in the next PR

PR #1144, opened 2021-09-27
CPDTP-500: Add npq schedules

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-500

We want to have schedules for npq courses.

I will need to migrate schedules in a bunch of environments after merging this change. We will also need some extra logic around which schedule can exist on which profile, but that comes in a latter PR. 

### Code quality checks


spec/cypress/app_commands/scenarios/admin/school_participants.rb
```diff
@@ -53,7 +55,8 @@
                                           date_of_birth: Date.parse("10/12/1982"),
                                           nino: "NI123456",
                                           teacher_reference_number: "9780824",
-                                          school_urn: school.urn
+                                          school_urn: school.urn,
+                                          npq_course: npq_course
 
   NPQ::CreateOrUpdateProfile.new(npq_validation_data: npq_validation_data).call
```
plisovin, line 61: I guess it will be an issue when I merge my PR into develop because CreateOrUpdateProfile got removed

kicferk1: Maybe, will fix it when you merge your PR

app/services/npq/create_or_update_profile.rb
```diff
@@ -39,5 +39,16 @@ def teacher_profile
     def user
       @user ||= npq_validation_data.user
     end
+
+    def schedule
+      case npq_validation_data.npq_course.identifier
```
jamesrose, line 44: Do we want this schedule to be static? Could an NPQ course belong to a schedule?

spec/cypress/app_commands/scenarios/admin/school_participants.rb
plisovin: It's been merged now

app/services/npq/create_or_update_profile.rb
kicferk1: Ehhh, I think I would prefer it to be static for time being. Really, what this is is a 'default schedule for a course' - which will change every couple months. And I prefer changing it in one place in code, rather than updating each course in each environment. We might revisit this idea soon enough though - changing npq schedule will need to know which schedules a course can have, so maybe we will store that in a database.

plisovin (approved):

app/models/finance/schedule.rb
```diff
@@ -4,7 +4,15 @@ class Finance::Schedule < ApplicationRecord
   has_many :milestones, -> { order(milestone_date: :asc) }
   has_many :participant_profiles
 
-  def self.default
+  def self.default_ecf
```
blackrat, line 7: Would it not be better to index these by course identifier like it's done in the following "accept" and then just call default (or default_for()) with a course identifier.

blackrat, line 7: If these are going to be different between ecf and npq, that sounds like we should start namespacing more. Thoughts?

kicferk1: Soo, have one schedule class for ECF and another schedule class for NPQ? 

Ehh, I am not sure I like it, cause `ParticipantProfiles` would need to point at two different schedule models. And, well, the underlying structure is "a schedule has an identifier and a list of milestones" for both ECF and NPQ. And I can't think of a meaningful behaviour difference. 

The only thing I can think of is that when we are changing schedules we don't want ecf participants to get NPQ schedules and vice versa, but I think scopes are a more appropriate tool to handle that than namespaces.

PR #1147, opened 2021-09-28
Ruby 3 upgrade

ethanmills:
Upgrade to ruby 3.0.2

[new features](https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/)
[the big breaking change](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/)


Gemfile
```diff
@@ -166,7 +166,7 @@ group :test do
   gem "percy-capybara"
   gem "pundit-matchers", "~> 1.7.0"
   gem "rails-controller-testing", ">= 1.0.5"
-  gem "rspec-default_http_header", ">= 0.0.4"
+  gem "rspec-default_http_header", github: "ethanmills/rspec-default_http_header", ref: "b72a803"
```
ethanmills, line 169: PR against original https://github.com/kenchan/rspec-default_http_header/pull/6

app/services/invite_schools.rb
```diff
@@ -261,16 +262,16 @@ def invite_sitless_not_opted_out_schools_for_nqt_plus_one
       .not_opted_out
       .where.missing(:induction_coordinators)
       .find_each do |school|
-        if school.contact_email.blank?
```
ethanmills, line 264: No idea what's going on here

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3.6"
 services:
   db:
-    image: postgres:9.6-alpine
```
ethanmills, line 4: Noticed this as I was going through

.ruby-version
```diff
@@ -1 +1 @@
-2.7.2
+3.0.2
```
kicferk1, line 1: I find it funny how many places we don't use that file in 😂 

Should we just axe it?

app/services/invite_schools.rb
kicferk1: I think `School` is indented with 4 spaces, each of the scopes is indented with 6, and body of the block is indented 2 deeper than the start of previous expression, rather than start of previous line. For some reason.

I guess if we made the school scope into a single line it would look reasonable...

kicferk1 (approved):

.ruby-version
asmega: rbenv, which i use, uses this file. i think it's super useful personally

PR #1153, opened 2021-09-28
Added request_for_details email status to participant_status_tag component

BroiSatse:
https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-873

### DIFF
--- a/app/models/email.rb
+++ b/app/models/email.rb
@@ -11,7 +11,11 @@ class Email < ApplicationRecord
   scope :associated_with, ->(object) { where(id: Association.where(object: object).select(:email_id)) }
   scope :tagged_with, ->(*tags) { tags.inject(self) { |scope, tag| scope.where("? = ANY (tags)", tag) } }
 
-  FAILED_STATUSES = %w[pernament-failure temporary-failure technical-failure].freeze
+  FAILED_STATUSES = %w[permanent-failure temporary-failure technical-failure].freeze
+
+  TAG_RELEASE_TIME = {
+    request_for_details: Time.zone.parse("28/09/2021 10:30"),
+  }.freeze
 
   def create_association_with(*objects, as: nil) # rubocop:disable Naming/MethodParameterName
     objects.each do |object|
@@ -35,7 +39,7 @@ class Email < ApplicationRecord
     order(:created_at).last
   end
 
-  def self.released_at
-    @released_at ||= Time.zone.parse("28/09/2021 10:30")
+  def self.released_at(tag:)
+    [Email.tagged_with(tag).minimum(:created_at), TAG_RELEASE_TIME[tag.to_sym]].compact.min
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/email.rb
```diff
@@ -20,4 +22,20 @@ def create_association_with(*objects, as: nil) # rubocop:disable Naming/MethodPa
...
+    status.in? FAILED_STATUSES
+  end
+
+  def self.latest
+    order(:created_at).last
+  end
+
+  def self.released_at
+    @released_at ||= Time.zone.parse("28/09/2021 10:30")
+  end
```
kicferk1, line 40: The way we are using it has less to do with 'when have we released the emails' and more to do with 'when have we started tagging the email in question with the tag we search by'. If we released the email functionality, but didn't tag the requests for details with anything, then the code finding latest email simply wouldn't logically work. 

And luckily for the case of requesting details, that date of starting to tag them and releasing the email tracking is the same.

Buuuut, there are a lot more emails that we send, and we will probably want to make similar checks for some of them. Should we go through the places that send them and make sure we tag them with something appropriately distinct?

BroiSatse: Ah, that is interesting. You are correct, this should really be probably `self.release_at(tag: nil)`?

I'm very happy with the idea of tagging all the existing emails. We should also convert the exisinig email records (PartnershipEmail / NotificationEmail) into Email objects

kicferk1: I think tagging all existing emails is much better than having a release date for each tag. And if we add tagging to all emails quickly enough, we can just bump the `released_at` date by 1 day and not really lose much.

IMO the existing email records being switched to Email objects should be its own PR, and like you say, as part of the migration there we can tag the emails anyway. 

But I think it would be nice to have the change to tag all emails done in the same PR that adds a release date, cause I think they are quite closely related. Happy to have them in a separate small PR if you prefer though.

BroiSatse: Ok, Updated this to take a tag into an account. But that is now a bit funny - ideally, the release data is the date the first email with given tag has been created... If none has been sent, we could assume it just wasn't released yet. But that messes up the tests unless I explicitly create an email with given tag in the past :D Better ideas?

kicferk1: I think hard-coding it to 29th of September sounds like a great idea, so long as we deploy a PR tagging all emails tomorrow :)

kicferk1 (approved):

app/models/email.rb
BroiSatse: @kicferk1 After chatting with James, we are both very happy with dropping the `Email.released_at` completely, allowing some transient inconsistency - each participant should receive a reminder request every 7 days, so this would solve itself within a week

app/models/email.rb
```diff
@@ -11,6 +11,12 @@ class Association < ApplicationRecord
   scope :associated_with, ->(object) { where(id: Association.where(object: object).select(:email_id)) }
   scope :tagged_with, ->(*tags) { tags.inject(self) { |scope, tag| scope.where("? = ANY (tags)", tag) } }
 
+  FAILED_STATUSES = %w[permanent-failure temporary-failure technical-failure].freeze
+
+  TAG_RELEASE_TIME = {
+    request_for_details: Time.zone.parse("28/09/2021 10:30"),
+  }.freeze
+
```
kicferk1, line 19: We don't need it anymore, right?

app/models/email.rb
kicferk1: Sure, that works for me too.

PR #1154, opened 2021-09-29
Add email format check for sign in

xswoodsx:
## Ticket and context

Ticket: [921](https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?assignee=5f3543bb3e9e2e004db3189d&selectedIssue=CPDRP-921)

The regex in Devise initializer hasn't been used as it didn't cover many invalid formats i tried, e.g "invalid@email,com"

### Code quality checks

### HTML Checks

### Gotchas

2. Try to log in with an incorrect email format 

## External API changes

https://user-images.githubusercontent.com/69353998/135220212-c39613f8-92d5-46fd-99ca-2f6f254e9437.mov

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -40,11 +40,6 @@ class Users::SessionsController < Devise::SessionsController
   end
 
   def create
-    self.resource = validate_email
-    if resource.errors.any?
-      render :new and return
-    end
-
     super
   rescue Devise::Strategies::PasswordlessAuthenticatable::Error
     render :login_email_sent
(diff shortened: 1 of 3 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -86,4 +91,29 @@ def mock_login
...
+
+    user = User.find_by(email: params[:user][:email])
+
+    if user.blank?
+      user = User.new(email: email)
+      user
+    end
+
+    user
+  end
```
kicferk1, line 118: There is a rather more sophisticated email format checker in our code already. `NotifyEmailValidator`.

You use it by making a model, and in that model, writing something like
```
  validates :email, notify_email: true
```

I wonder if we want to make a form object for sign in to wire it up nicely.

ethanmills: We're using `User` here, which already has ` validates :email, presence: true, uniqueness: true, notify_email: true`, so we probably hijack that

ethanmills: Although maybe not, since that would cause issues with uniqueness

kicferk1: Oh, interesting, that would be even nicer :)

kicferk1 (approved):

ethanmills (approved):

PR #1156, opened 2021-09-29
CPDRP-996: add task to schedule automatic validation chasers

ethanmills:
## Ticket and context

Ticket: CPDRP-996

Lots of participants don't have automatic validation chasers scheduled. We need to fill that gap

### Code quality checks


spec/support/delayed_job.rb
```diff
@@ -102,7 +102,9 @@ def query_jobs(object, method_name)
     match do |job_class|
       job_arguments = find_job_with_arguments(job_class)
 
-      job_arguments.any? do |job, args|
+      method = @once ? :"one?" : :"any?"
+
+      job_arguments.send(method) do |job, args|
```
kicferk1, line 107: `@once ? :"one?" : :"any?"` gets me dizzy. Is this the same as
```
      method = @once ? "one?" : "any?"

      job_arguments.send(method.to_sym) do |job, args|
```
?

Also, sending method to object by its name is... not that nice.

Should we extract a private method, something like `check_job_arguments` or something like that, and have an if statement here? So that we can have
```
if @once
  job_arguments.one? check_job_arguments
else
  job_arguments.any? check_job_arguments
end
```

kicferk1 (approved):
The code looks good, the custom matcher might want some love

PR #1164, opened 2021-09-29
Nqt+1 template mail

ethanmills:
Replace the blank template with a standard template


xswoodsx (approved):
LGTM, thanks very much for the changes.

app/mailers/school_mailer.rb
```diff
@@ -19,6 +19,7 @@ class SchoolMailer < ApplicationMailer
   BASIC_TEMPLATE = "b1ab542e-a8d5-4fdf-a7aa-f0ce49b98262"
```
FriedSock, line 19: I think we can remove this constant now as well since nothing will be using it

FriedSock (approved):
Looks good if this does work with notify

app/mailers/school_mailer.rb
ethanmills, line 19:
```suggestion
```

PR #1170, opened 2021-09-30
CPDRP-1000 Participant eligibility logic update

tonyheadford:
## Ticket and context

Ticket: [Jira](https://dfedigital.atlassian.net/browse/CPDRP-1000)
Updates participant eligibility logic to account for the participant profile type.
Adds reason enum so that we can easily see the reason for the given status.

We will want to run this manually against prod, primarily to update the status for mentors that have a previous induction. Currently these are flagged for manual check but we should ignore the previous induction flag for mentors.

```ruby
ECFParticipantEligibility.where(manually_validated: false).find_each { |eligibility| eligibility.determine_status; eligibility.save! }
```

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/db/migrate/20210930125828_add_reason_to_ecf_participant_eligibilities.rb
+++ b/db/migrate/20210930125828_add_reason_to_ecf_participant_eligibilities.rb
@@ -2,6 +2,6 @@
 
 class AddReasonToECFParticipantEligibilities < ActiveRecord::Migration[6.1]
   def change
-    add_column :ecf_participant_eligibilities, :reason, :string, null: false, default: "none" 
+    add_column :ecf_participant_eligibilities, :reason, :string, null: false, default: "none"
   end
 end
--- a/db/migrate/20210930125828_add_reason_to_ecf_participant_eligibilities.rb
+++ b/db/migrate/20210930125828_add_reason_to_ecf_participant_eligibilities.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddReasonToECFParticipantEligibilities < ActiveRecord::Migration[6.1]
+  def change
+    add_column :ecf_participant_eligibilities, :reason, :string, null: false, default: "none" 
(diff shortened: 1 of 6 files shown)


db/migrate/20210930125828_add_reason_to_ecf_participant_eligibilities.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddReasonToECFParticipantEligibilities < ActiveRecord::Migration[6.1]
+  def change
+    add_column :ecf_participant_eligibilities, :reason, :string, null: false, default: "none"
```
kicferk1, line 5: Is this reason for the participant not being eligible?

If so, can we call it `ineligible_reason`? `reason_ineligible`?

jamesrose: Not quite. I think it covers any status reason.

tonyheadford: I guess it's the reason for the status, there aren't any reasons for `eligible` but you'd have to put something in hence the `none` reason.

ethanmills:
I think it would be safer to run 

```ruby
ECFParticipantEligibility.where(manually_validated: false).find_each { |eligibility| eligibility.determine_status; eligibility.save! }
```

otherwise we risk reversing any manual override to eligible

tonyheadford:
> I think it would be safer to run
> 
> ```ruby
> ECFParticipantEligibility.where(manually_validated: false).find_each { |eligibility| eligibility.determine_status; eligibility.save! }
> ```
> 
> otherwise we risk reversing any manual override to eligible

Yes, good idea (although I wasn't aware that there had been any as of yet?)

ethanmills (approved):

PR #1171, opened 2021-10-01
CPDRP-974: Send email to DIY schools

ethanmills:
## Ticket and context

Ticket: CPDRP-974

Send email to DIY schools without pending partnership telling them the NERO site is going

I've not written tests for this, but I've tested it manually. It's a one-off email to 76 people

I will remove this once sent


app/mailers/school_mailer.rb
```diff
@@ -326,4 +327,14 @@ def nqt_plus_one_sit_invite(recipient:, start_url:, school:)
...
   end
+
+  def diy_wordpress_notification(user:)
+    template_mail(
+      DIY_WORDPRESS_NOTIFICATION_TEMPLATE,
+      to: user.email,
+      rails_mailer: mailer_name,
+      rails_mail_template: action_name,
+      personalisation: {},
+    ).tag(:diy_wordpress_notification).associate_with(user, as: :user)
```
BroiSatse, line 338: No need to associate with the user, it will be automatically associated as recipient (unless he changes his email between email being scheduled and sent).

However we might want to associate this email with a school as well?

app/services/invite_schools.rb
```diff
@@ -274,6 +274,18 @@ def invite_sitless_not_opted_out_schools_for_nqt_plus_one
     end
   end
 
+  def inform_diy_schools_of_wordpress
+    User.where(
+      id: diy_school_cohorts_without_pending_partnerships
+            .joins(school: :induction_coordinators)
+            .pluck(:user_id),
```
BroiSatse, line 281: `select(:user_id)` instead of pluck - this will create a single SQL query with a subquery allowing PG to optimize it.

app/services/invite_schools.rb
```diff
@@ -274,6 +274,18 @@ def invite_sitless_not_opted_out_schools_for_nqt_plus_one
...
+    ).distinct.find_each do |user|
```
BroiSatse, line 282: No need for distinct - there's no join on the top relation.

app/mailers/school_mailer.rb
```diff
@@ -326,4 +327,14 @@ def nqt_plus_one_sit_invite(recipient:, start_url:, school:)
...
+  end
```
BroiSatse, line 339: Why does it have `wordpress` in the name - it makes me want to just delete it :D

app/services/invite_schools.rb
ethanmills, line 281:
```suggestion
            .select(:user_id),
```

app/services/invite_schools.rb
ethanmills: Of course
```suggestion
    ).find_each do |user|
```

app/mailers/school_mailer.rb
ethanmills: There isn't necessarily a single school here, the SIT may be controlling two relevant schools - we would just send one email

app/mailers/school_mailer.rb
ethanmills: The good news is this email is telling them we're getting rid of the wordpress site

app/mailers/school_mailer.rb
BroiSatse: Ah, makes sense then. However I keep thinking about that linking - there's usually quite a lot of objects we want to link email to! For example, this email might be linked to all the schools of given SIT and to his SIT profile as well. I try to imagine that there is a small link in the admin panel called "related emails" and wonder what emails should be displayed there for each object....

app/mailers/school_mailer.rb
BroiSatse: :partying_face:

app/services/invite_schools.rb
```diff
@@ -436,4 +448,14 @@ def send_section_41_invite_email(nomination_email)
...
+
+  def diy_school_cohorts_with_pending_partnerships
+    SchoolCohort.where(cohort_id: Cohort.current.id, induction_programme_choice: "design_our_own").joins(school: :partnerships).where(school: { partnerships: { challenge_reason: nil } })
+  end
+
+  def diy_school_cohorts_without_pending_partnerships
+    SchoolCohort
+      .where(cohort_id: Cohort.current.id, induction_programme_choice: "design_our_own")
+      .where.not(id: diy_school_cohorts_with_pending_partnerships)
+  end
```
kicferk1, line 460: Should these be scopes on SchoolCohort?

kicferk1 (approved):

app/services/invite_schools.rb
ethanmills: Some of them might be scopes already, and this is likely the only time we will ever use the others

app/mailers/school_mailer.rb
ethanmills, line 338:
```suggestion
    ).tag(:diy_wordpress_notification)
```

ethanmills: I've removed the explicit link

PR #1172, opened 2021-10-01
Steal first version of tech docs from teacher training api

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-477

I had a little dig, and it seemed to me that what the docs looked in the ticket was eerily similar to what TeacherTraining API were doing. Which is using a gem to generate their documentation.

Vast majority of changes are caused by a different way of automatically presenting API operations - I cut out everything we stole from Apply for Teacher Training, and brought in everything from TeacherTraining API.

The proof of the pudding is in the eating, so dig in - https://ecf-review-pr-1172.london.cloudapps.digital/api-reference

The technical gist of how the docs are handled: We use `govuk_tech_docs` to harness power of `middleman` to generate a static site from the stuff we put in `docs` directory. We output that static site into `public/api-reference` at build time - so when we run our application, it is served from `/api-reference` path. One drawback of it which I haven't found a solution to is that running percy and / or axe on it seems harder. Since it would require the static site to be built before we ran the feature specs :/ I added some basic smoke tests but maybe they are not quite enough. Open to suggestions. 

Getting that `api-reference` path to work was harder than expected, cause the gem is a bit opinionated - I needed to override some methods generating headings on left hand side, because without it they thought everything should be in one big family under `api-reference`.

Open to suggestions of what I missed :)


config/routes.rb
```diff
@@ -376,5 +376,8 @@
 
   get "/how-to-set-up-your-programme", to: "step_by_step#show", as: :step_by_step
 
+  get "/assets/govuk/assets/fonts/:name.:extension", to: redirect("/api-reference/assets/govuk/assets/fonts/%{name}.%{extension}")
```
asmega, line 379: not sure we want this? every asset will require an extra request assuming it's not cached. I feel like this is a workaround and doesn't solve the actual problem we have introduced.

docs/.tool-versions
```diff
@@ -0,0 +1,3 @@
+nodejs 12.15.0
```
asmega, line 1: i don't know if anyone of the team uses this? if not in use we can delete

config/routes.rb
kicferk1: The problem is that the gem `govuk_tech_docs` hard-codes its assets to be in `/assets/govuk/assets/`, and I am not sure it's a good place for something that's a small part of our application.

`govuk_tech_docs` in general doesn't seem to believe it should be just a small part of an application. But I think it is a pretty good idea to have the tech docs be part of what we release, and this seemed like the simplest solution. I guess we could make it a separate mini-service in paas, but that would require a lot of dev ops fun :/

kicferk1: Edit: We could get rid of it when https://github.com/alphagov/tech-docs-gem/pull/197 goes in, or something similar to it.

asmega (approved):

PR #1174, opened 2021-10-01
Email tags

BroiSatse:
Adding tags to most of the emails - together with some much needed clean-up.

Notes:
Current form of email tagging requires tagging and associateion to live inside the mailer - for that purpose we need to pass the actual obejct rather than computed values.

Additionally, some of the templates has been used in two different mailers - we had ParticipantMailer and ParticipantValidationMailer, both sending the same email. As the former was only used in the BetaValidationService, and those emails don't need manual triggering (as those emails ar enow being sent automatically) I've decided to prune that mailer in this PR - it took me few hours of chatting to different people to figure out how they are being used and whehter we expect them to be used in the future (hence some of the email comments) ...


app/mailers/participant_validation_mailer.rb
```diff
@@ -1,63 +0,0 @@
-# frozen_string_literal: true
```
kicferk1, line 1: Is this the right PR to remove it?

Even if we think it's a good enough PR to remove it... can we mention it in the description?

app/mailers/school_mailer.rb
```diff
@@ -22,6 +22,7 @@ class SchoolMailer < ApplicationMailer
   DIY_WORDPRESS_NOTIFICATION_TEMPLATE = "e1067a2f-b027-45a6-8e51-668e170090d1"
   PARTNERED_SCHOOL_INVITE_SIT_EMAIL_TEMPLATE = "8cac177e-b094-4a00-9179-94fadde8ced0"
 
+  # This email is currently (30/09/2021) only used for manually sent chaser emails
```
kicferk1, line 25: I don't really like the idea of having code comments, and I like the idea of dating them even less :P 

If it is not clear what the email is meant to do, we should rename the method sending it to make it more explicit. And I guess we could alias methods to deal with delayed jobs for a while.

app/services/create_induction_tutor.rb
```diff
@@ -24,14 +24,12 @@ def call
         InductionCoordinatorProfile.create!(user: user, schools: [school])
       end
 
-      # TODO: This should really be using deliver_later, but this can't be tested via Cypress
-      # After discussion leaving this as deliver_now with  this comment
```
kicferk1, line 28: Have we resolved this?

app/services/validation_beta_service.rb
```diff
@@ -25,122 +25,6 @@ def remind_sits_to_add_participants
     end
   end
 
-  # This timestamp was generated by finding the last user with request_for_details_sent_at: nil,
```
kicferk1, line 28: Why are we deleting it here? Again, PR description please

spec/cypress/support/commands.js
```diff
@@ -52,7 +52,8 @@ export const computeHeadersFromEmail = (email) =>
       [element.name]: element.unparsed_value,
     }),
     {}
-  );
+  );q
+
```
kicferk1, line 56: I think something will break here :D

app/mailers/school_mailer.rb
BroiSatse: Agree, I hate comments as well - the reason I added those was that it was extremly difficult to figure out when this mailer is being used due to a large part of apparantly unused code existing only to be run manully from the console. I dated this as well so we could make the decision later on to delete this based on how stale this comment is. It makes no sense to refactor this as it is to be deleted.

app/mailers/school_mailer.rb
```diff
@@ -66,33 +69,27 @@ def nomination_confirmation_email(user:, school:, start_url:, step_by_step_url:)
...
       to: recipient,
       rails_mailer: mailer_name,
       rails_mail_template: action_name,
       personalisation: {
-        lead_provider_name: lead_provider_name,
-        delivery_partner_name: delivery_partner_name,
-        school_name: school_name,
+        lead_provider_name: partnership.lead_provider.name,
+        delivery_partner_name: partnership.delivery_partner.name,
+        school_name: partnership.school.name,
```
ethanmills, line 86: How much coupling are we comfortable introducing here? This is probably fine, but I don't think I'd want any more

app/mailers/school_mailer.rb
```diff
@@ -66,33 +69,27 @@ def nomination_confirmation_email(user:, school:, start_url:, step_by_step_url:)
...
         nominate_url: nominate_url,
         challenge_url: challenge_url,
-        challenge_deadline: challenge_deadline,
+        challenge_deadline: partnership.challenge_deadline.strftime("%d/%m/%Y"),
         subject: "FAO: NQT coordinator. Training provider confirmed.",
       },
-    )
+    ).tag(:partnership_created, :request_to_nominate_sit).associate_with(partnership, partnership.school)
```
ethanmills, line 92: Out of curiosity, can this `associate_with` be called where the mailer is called? As in `SchoolMailer.school_partnership_notification_email(...).associate_with(...).deliver_later`? Or is the mailer magic in a way that prevents that?

app/mailers/participant_validation_mailer.rb
BroiSatse: Description updated - this whole mailer was a duplicate of PariticpantMailer, sending same templates with quite a different personalisation. It took me a while to figure this out, so decided to remove it in this PR, especially given that most of the tags added in this PR have no backing in required features making this PR pure TechDebt

app/services/create_induction_tutor.rb
BroiSatse: Now it is :) All the emails are being delivered inline, while all other jobs are being delayed.

app/mailers/school_mailer.rb
BroiSatse: Agree - I'd much rather prefer having associations and tagging available outside of the mailer, so more like:

```
SchoolMailer.some_email().associate_with(school).deliver_later
```
While possible, that requires quite a lot of monkey patching

app/mailers/school_mailer.rb
BroiSatse: Ha! Just commented the same thing :D

So yes, it is theratically possible, but the way mailer deliver_later works is that it stores a ActionMailer::MailDeliveryJob (serialized and wrapped two time), which just keeps trakc of the mailer class, email name and arguments to the mailer. When executed in the backend it just invokes `MailerClass.email_name(**arguments).deliver_now` - so to make tagging and associate_with working outside of the mailer I'd need to extend this job class - and extending Mail::Message already brought quite a lot of pushback...

app/services/validation_beta_service.rb
```diff
@@ -25,122 +25,6 @@ def remind_sits_to_add_participants
...
-  # after automation this request for details is sent automatically
-  AUTOMATION_LAUNCH_TIME = Time.zone.parse("2021-09-17 09:45:33.284979000 +0000").freeze
-  VALIDATION_CHASER_BATCH_LIMIT = 5000
-
```
kicferk1, line 32: We are deleting all unit tests for this servise, but not all the contents of this service. Does that sound right? :P

kicferk1 (approved):
Only question I have is about tests for the service we purge in this PR

app/services/validation_beta_service.rb
BroiSatse: Not all of them, but quite a lot of them - enough for github to truncate the diff for the file. You can click on "Load the diff" - there are about 300 lines of tests deleted for the deleted methods, the remaining tests are still there.

kicferk1: Ahh, my bad. Alright then :)

PR #1177, opened 2021-10-04
Scope rejections when Accepting NPQ applications

asmega:
## Ticket and context

Ticket: https://trello.com/c/yOJFfO9p/555-build-aso-journey

- When a provider accepts an NPQ application at the moment it rejects all other applications
- This change makes it so when the accept comes in, it only rejects other applications from the user on the same course. 

### Code quality checks

### HTML Checks

### Gotchas

- ssh in to the box
- find or create a lead provider
- submit 3 NPQ applications all from the same user - 2 (a+b) on the same course and 1 (c) on another course
- in the block of 2 (a+b), accept `a` and `b` should be rejected. `c` should remain unchanged 

## External API changes

- There is a minor impact on the API on accepting. However do we need to document this as this feature?

### DIFF
--- a/app/services/npq/accept.rb
+++ b/app/services/npq/accept.rb
@@ -15,7 +15,6 @@ module NPQ
     end
 
     def call
-      other_applications = NPQValidationData.where(user_id: user.id).where.not(id: npq_application.id)
       ApplicationRecord.transaction do
         teacher_profile.update!(trn: npq_application.teacher_reference_number) if npq_application.teacher_reference_number_verified?
         create_profile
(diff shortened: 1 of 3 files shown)


kicferk1 (commented):
Should we scope it to the provider too? Or is it ok for provider A accepting causing a rejection of provider B applications?

asmega:
> Should we scope it to the provider too? Or is it ok for provider A accepting causing a rejection of provider B applications?

No. We only want them to have the course with one provider. If later down the line there is an issue with which provider provided the training that will be resolved offline. If we do your suggestion it will cause people to apply for the same course to multiple providers which can then all be accepted and then DfE would have to fund them all.

kicferk1 (approved):

PR #1182, opened 2021-10-04
Add npq dumy data seeding

kicferk1:
## Ticket and context

As we move towards testing with NPQ providers, I think we will find this pretty useful.

### DIFF
--- a/db/seeds/test_data.rb
+++ b/db/seeds/test_data.rb
@@ -501,5 +501,6 @@ ParticipantProfile::Mentor.find_or_create_by!(teacher_profile: teacher_profile) 
 end
 
 ["Capita", "Teach First", "UCL Institute of Education", "Best Practice Network", "Ambition Institute", "Education Development Trust"].each do |provider|
-  ValidTestDataGenerator::LeadProviderPopulater.delay.call(name: provider, total_schools: 10, participants_per_school: 10)
+  ValidTestDataGenerator::LeadProviderPopulater.delay.call(name: provider, total_schools: 1, participants_per_school: 10)
+  ValidTestDataGenerator::NPQLeadProviderPopulater.delay.call(name: provider, total_schools: 1, participants_per_school: 10)
 end
--- a/db/seeds/test_data.rb
+++ b/db/seeds/test_data.rb
@@ -500,7 +500,10 @@ ParticipantProfile::Mentor.find_or_create_by!(teacher_profile: teacher_profile) 
   ParticipantProfileState.find_or_create_by!(participant_profile: mentor_profile)
 end
 
-["Capita", "Teach First", "UCL Institute of Education", "Best Practice Network", "Ambition Institute", "Education Development Trust"].each do |provider|
+LeadProvider.all.map(&:name).each do |provider|
   ValidTestDataGenerator::LeadProviderPopulater.delay.call(name: provider, total_schools: 1, participants_per_school: 10)
+end
+
+NPQLeadProvider.all.map(&:name).each do |provider|
   ValidTestDataGenerator::NPQLeadProviderPopulater.delay.call(name: provider, total_schools: 1, participants_per_school: 10)
 end
--- a/lib/tasks/sandbox_data.rake
+++ b/lib/tasks/sandbox_data.rake
@@ -15,4 +15,16 @@ namespace :lead_providers do
     end
     ValidTestDataGenerator::AmbitionSpecificPopulater.call(name: "Ambition Institute", total_schools: 3, participants_per_school: 1500)
   end
+
+  desc "create NPQ seed schools and participants for API testing"
+  task seed_schools_and_participants_npq: :environment do
+    logger = Logger.new($stdout)
+    logger.formatter = proc do |_severity, _datetime, _progname, msg|
+      "#{msg}\n"
+    end
+
+    ["Ambition Institute", "Capita", "Teach First", "UCL Institute of Education", "Best Practice Network", "Education Development Trust"].each do |provider|
+      ValidTestDataGenerator::NPQLeadProviderPopulater.call(name: provider, total_schools: 100, participants_per_school: 100)
+    end
+  end
 end
--- a/lib/tasks/sandbox_data.rake
+++ b/lib/tasks/sandbox_data.rake
@@ -10,7 +10,7 @@ namespace :lead_providers do
       "#{msg}\n"
     end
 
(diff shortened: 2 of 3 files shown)


lib/tasks/valid_test_data_generator.rb
```diff
@@ -151,4 +153,80 @@ def random_or_nil_trn
...
   end
+
+  class NPQLeadProviderPopulater
+    class << self
+      def call(name:, total_schools: 10, participants_per_school: 10)
+        new(name: name).call(total_schools: total_schools, participants_per_school: participants_per_school)
+      end
+    end
+
+    def call(total_schools: 10, participants_per_school: 100)
```
FriedSock, line 164: This `participants_per_school` variable is passed down through 4 different methods. We could just assign it to an instance variable to avoid that.

lib/tasks/valid_test_data_generator.rb
```diff
@@ -151,4 +153,80 @@ def random_or_nil_trn
...
+    def generate_new_schools(count:, participants_per_school:)
+      count.times { create_fip_school_with_cohort(urn: SchoolURNGenerator.next, participants_per_school: participants_per_school) }
+    end
+
+    def find_or_create_participants(school:, number_of_participants:)
+      generate_new_participants(school: school, count: number_of_participants)
+    end
+
+    def generate_new_participants(school:, count:)
+      while count.positive?
```
FriedSock, line 185: `count.times do ` ?

lib/tasks/valid_test_data_generator.rb
```diff
@@ -151,4 +153,80 @@ def random_or_nil_trn
...
+    def create_fip_school_with_cohort(urn:, participants_per_school:)
+      school = School.find_or_create_by!(urn: urn) do |s|
+        s.name = Faker::Company.name
+        s.address_line1 = Faker::Address.street_address
+        s.postcode = Faker::Address.postcode
+      end
+      find_or_create_participants(school: school, number_of_participants: participants_per_school)
+    end
+
+    def random_or_nil_trn
```
FriedSock, line 222: This doesn't look like it is being used anywhere, did you mean to use it on line 203?

lib/tasks/valid_test_data_generator.rb
```diff
@@ -151,4 +153,80 @@ def random_or_nil_trn
...
+      [true, false].sample ? nil : TRNGenerator.next
+    end
+
+    def weighted_choice(selection:, odds:)
```
FriedSock, line 226: This looks unused as well?

lib/tasks/valid_test_data_generator.rb
asmega: if you use this suggestion make sure to add the argument as line 187 will need it, probably don't call it `count` again though

db/seeds/test_data.rb
```diff
@@ -501,5 +501,6 @@
 end
 
 ["Capita", "Teach First", "UCL Institute of Education", "Best Practice Network", "Ambition Institute", "Education Development Trust"].each do |provider|
-  ValidTestDataGenerator::LeadProviderPopulater.delay.call(name: provider, total_schools: 10, participants_per_school: 10)
+  ValidTestDataGenerator::LeadProviderPopulater.delay.call(name: provider, total_schools: 1, participants_per_school: 10)
+  ValidTestDataGenerator::NPQLeadProviderPopulater.delay.call(name: provider, total_schools: 1, participants_per_school: 10)
```
blackrat, line 505: Surely we should only do this for NPQ lead providers?

lib/tasks/sandbox_data.rake
```diff
@@ -15,4 +15,16 @@ namespace :lead_providers do
...
+
+  desc "create NPQ seed schools and participants for API testing"
+  task seed_schools_and_participants_npq: :environment do
+    logger = Logger.new($stdout)
+    logger.formatter = proc do |_severity, _datetime, _progname, msg|
+      "#{msg}\n"
+    end
+
+    ["Ambition Institute", "Capita", "Teach First", "UCL Institute of Education", "Best Practice Network", "Education Development Trust"].each do |provider|
+      ValidTestDataGenerator::NPQLeadProviderPopulater.call(name: provider, total_schools: 100, participants_per_school: 100)
```
blackrat, line 27: Again. These should match up with the generated NPQ lead providers rather than the ECF ones.

lib/tasks/valid_test_data_generator.rb
blackrat: Yes. It was like this for ECF because of the variable number of ECT/Mentor and isn't required here.

lib/tasks/valid_test_data_generator.rb
```diff
@@ -151,4 +153,80 @@ def random_or_nil_trn
...
+    end
+
+    def find_or_create_participants(school:, number_of_participants:)
+      generate_new_participants(school: school, count: number_of_participants)
+    end
+
+    def generate_new_participants(school:, count:)
+      while count.positive?
+        count -= 1
+        create_participant(school: school, accepted: count.even?)
```
blackrat, line 187: `accepted: count.even?` Other variables are deliberately using a representative sampling. Do we really expect a 50% rejection ratio?

lib/tasks/valid_test_data_generator.rb
blackrat: Probably better to use this for the `accepted:` than just to have the `.even`

lib/tasks/valid_test_data_generator.rb
kicferk1: I think I meant not to have it at all :)

lib/tasks/valid_test_data_generator.rb
kicferk1: I like the certainty of having some accepted and some pending applications even with small numbers (in review apps or locally).

We might come back to it when we know for certain what dummy data the providers want.

db/seeds/test_data.rb
kicferk1: Ah, good spot. And we should do it for all of them probably.

lib/tasks/valid_test_data_generator.rb
kicferk1: Technically speaking the rejection ratio is 0%. The other 50% are pending.

I have no idea what we are expecting. When we know for sure, we can code it in. In the meantime, deterministic 50% seems better than non-deterministic percentages, because at least we can use it to comfortably seed our local db and review app db with something that has some pending and some approved applications.

PR #1187, opened 2021-10-04
CPDTP-250 implement states for declarations

blackrat:
## Ticket and context

There is a lifecycle for declarations that follow a number of states (submitted, payable (after checking), paid, voided being the initial ones). These states don't conflict and are separate, but are currently implemented as flags and separated.

In order to implement payment, this cleans up the implementation and switches existing code to use them as states as opposed to flags. More PRs will follow to actually implement the payment and aggregations once paid declarations are triggered and the update to payment breakdowns to actually handle only new declaration aggregations (and any late payment requirements)

Ticket:  https://dfedigital.atlassian.net/jira/software/projects/CPDTP/boards/87?selectedIssue=CPDTP-250

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/app/models/declaration_state.rb
+++ b/app/models/declaration_state.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class DeclarationState < ApplicationRecord
+  class << self
+    def submit!(participant_declaration:)
+      create!(participant_declaration: participant_declaration)
+    end
+
+    def void!(participant_declaration:)
+      create!(participant_declaration: participant_declaration, state: "voided")
+    end
+
+    def pay!(participant_declaration:)
+      create!(participant_declaration: participant_declaration, state: "paid")
+    end
+  end
+
+  belongs_to :participant_declaration
+
+  enum state: {
+    submitted: "submitted",
+    payable: "payable",
+    paid: "paid",
+    # awaiting_clawback: "awaiting_clawback",
+    # clawed_back: "clawed_back"
+    voided: "voided",
+  }
+end
--- a/app/models/participant_declaration.rb
+++ b/app/models/participant_declaration.rb
@@ -1,39 +1,44 @@
 # frozen_string_literal: true
 
 class ParticipantDeclaration < ApplicationRecord
-  has_many :profile_declarations
-  has_one :current_profile_declaration, -> { order(created_at: :desc) }, class_name: "ProfileDeclaration"
-  has_one :participant_profile, through: :current_profile_declaration
+  has_one :profile_declaration
+  has_one :participant_profile, through: :profile_declaration
+  has_many :declaration_states
+  has_one :current_state, -> { order(created_at: :desc).limit(1) }, class_name: "DeclarationState"
   belongs_to :cpd_lead_provider
   belongs_to :user
 
-  delegate :payable, to: :current_profile_declaration, allow_nil: true
+  delegate :submitted?, :payable?, :voided?, :paid?, to: :current_state, allow_nil: false
+  delegate :fundable?, to: :participant_profile, allow_nil: true
 
   validates :course_identifier, :user, :cpd_lead_provider, :declaration_date, :declaration_type, presence: true
 
   # Helper scopes
   scope :for_lead_provider, ->(cpd_lead_provider) { where(cpd_lead_provider: cpd_lead_provider) }
   scope :for_declaration, ->(declaration_type) { where(declaration_type: declaration_type) }
+  scope :for_profile, ->(profile) { joins(:profile_declaration).joins(:participant_profile).where(participant_profile: profile) }
   scope :started, -> { for_declaration("started").order(declaration_date: "desc").unique_id }
-  scope :uplift, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.uplift) }
-  scope :ect, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.ect_profiles) }
-  scope :mentor, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.mentor_profiles) }
-  scope :npq, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.npq_profiles) }
-  scope :payable, -> { joins(:current_profile_declaration).merge(ProfileDeclaration.where(payable: true)) }
+  scope :uplift, -> { joins(:profile_declaration).merge(ProfileDeclaration.uplift) }
+  scope :ect, -> { joins(:profile_declaration).merge(ProfileDeclaration.ect_profiles) }
+  scope :mentor, -> { joins(:profile_declaration).merge(ProfileDeclaration.mentor_profiles) }
+  scope :npq, -> { joins(:profile_declaration).merge(ProfileDeclaration.npq_profiles) }
+  scope :submitted, -> { joins(:current_state).where(current_state: { state: "submitted" }) }
+  scope :payable, -> { joins(:current_state).where(current_state: { state: "payable" }) }
+  scope :paid, -> { joins(:current_state).where(current_state: { state: "paid" }) }
+  scope :voided, -> { joins(:current_state).where(current_state: { state: "voided" }) }
+  scope :changeable, -> { joins(:current_state).where(current_state: { state: %w[submitted payable] }) }
   scope :unique_id, -> { select(:user_id).distinct }
-  scope :voided, -> { where.not(voided_at: nil) }
-  scope :not_voided, -> { where(voided_at: nil) }
 
   # Time dependent Range scopes
   scope :declared_as_between, ->(start_date, end_date) { where(declaration_date: start_date..end_date) }
   scope :submitted_between, ->(start_date, end_date) { where(created_at: start_date..end_date) }
 
   # Declaration aggregation scopes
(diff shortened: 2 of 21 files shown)


app/models/declaration_state.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    def void!(participant_declaration:)
+      create!(participant_declaration: participant_declaration, state: "voided")
+    end
+
+    def pay!(participant_declaration:)
+      create!(participant_declaration: participant_declaration, state: "paid")
+    end
+  end
```
kicferk1, line 16: Why would we not have this on an instance of `ParticipantDeclaration`?

app/models/participant_declaration.rb
```diff
@@ -1,68 +1,71 @@
 # frozen_string_literal: true
 
 class ParticipantDeclaration < ApplicationRecord
-  has_many :profile_declarations
-  has_one :current_profile_declaration, -> { order(created_at: :desc) }, class_name: "ProfileDeclaration"
-  has_one :participant_profile, through: :current_profile_declaration
+  has_one :profile_declaration
```
kicferk1, line 4: We will need to cut out the old ones (if there are any) in sandbox and prod.

app/models/participant_declaration.rb
```diff
@@ -1,68 +1,71 @@
...
+  has_one :participant_profile, through: :profile_declaration
+  has_many :declaration_states
+  has_one :current_state, -> { order(created_at: :desc).limit(1) }, class_name: "DeclarationState"
```
kicferk1, line 7: I'd think `has_one` does the `limit(1)` thing by default?

app/models/participant_declaration.rb
```diff
@@ -1,68 +1,71 @@
...
-        participant_profile: participant_profile,
-        participant_declaration: self,
-        payable: currently_payable,
-      )
-      reload
-    end
+    new_state = fundable? ? "payable" : "submitted"
+    return if current_state.state == new_state
+
+    DeclarationState.create!(participant_declaration: self, state: new_state) and reload if %w[submitted payable].include?(current_state.state)
```
kicferk1, line 53: We already have one early return, can we have another for this if condition?

app/models/participant_declaration.rb
```diff
@@ -1,68 +1,71 @@
...
   end
 
-  def voided
-    voided_at.present?
+  def submit
```
kicferk1, line 56: Let's add exclamation mark here too?

app/models/participant_declaration.rb
```diff
@@ -1,68 +1,71 @@
...
 
   def void!
-    # TODO: Prevent voiding a processed declaration - that requires clawbacks
-    update!(voided_at: Time.zone.now)
+    DeclarationState.void!(participant_declaration: self) and reload unless voided?
+  end
+
+  def pay!
+    DeclarationState.pay!(participant_declaration: self) and reload if payable?
+  end
```
kicferk1, line 66: ParticipantDeclaration is already accutely aware of the states that `DeclarationState` has - I would be tempted to make `DeclarationState` a dumb data class and handle all the logic with updating `ParticipantDeclaration` state in here.

app/serializers/participant_declaration_serializer.rb
```diff
@@ -1,22 +1,28 @@
...
     declaration.declaration_date.rfc3339
   end
 
+  attribute :voided, &:voided?
+
   attribute(:participant_id, &:user_id)
+
+  attribute :state do |declaration|
+    declaration.current_state.state
+  end
```
kicferk1, line 27: Should we call it `declaration_state` to distinguish it from other states in other objects?

app/services/record_declarations/base.rb
```diff
@@ -61,33 +61,34 @@ def parsed_date
...
-          cpd_lead_provider: cpd_lead_provider,
-          user: user,
-        ) do |participant_declaration|
-          profile_declaration = ProfileDeclaration.create!(
+      ApplicationRecord.transaction do
+        self.class.declaration_model.find_or_create_by!(declaration_parameters) do |participant_declaration|
+          DeclarationState.create!(
+            participant_declaration: participant_declaration,
+            state: user_profile.fundable? ? "payable" : "submitted",
+          )
```
kicferk1, line 73: So we sometimes start in one state and sometimes in another state? I guess it's alright.

app/services/record_declarations/retained/base.rb
```diff
@@ -21,34 +21,8 @@ def valid_evidence_types
...
-          ) do |participant_declaration|
-            profile_declaration = ProfileDeclaration.create!(
-              participant_declaration: participant_declaration,
-              participant_profile: user_profile,
-            )
-            profile_declaration.update!(payable: participant_declaration.currently_payable)
-          end
-        end
+      def declaration_parameters
+        super.merge({ evidence_held: evidence_held })
```
kicferk1, line 25: Nice

kicferk1: Although I know James hates the extra curly braces, `super.merge(evidence_held: evidence_held)`?

swagger/v1/component_schemas/ParticipantDeclarationCsvRow.yml
```diff
@@ -54,7 +54,13 @@ properties:
     nullable: false
     example: true
   voided:
-    description: "Indicates whether this declaration has been voided"
+    description: "[Deprecated - use state instead] Indicates whether this declaration has been voided"
```
kicferk1, line 57: Have you run swaggerise with those changes? I don't think I can see any changes to our api schema file.

kicferk1 (commented):
Do we need to add release notes (and / or usage guide) for it?

app/models/declaration_state.rb
blackrat: At some point I want to move this to state transitions and this is a more logical place to handle those. Also the "reload" is a bit of a code smell to me, but that's probably more down to how the has_one and has_many linkages work.

app/models/participant_declaration.rb
blackrat: There weren't when I last checked, but yes. There should only be a maximum of one when this is deployed.

app/models/participant_declaration.rb
blackrat: I looked at doing that, but decided it was clearer to have it as a definitive include rather than a negative early return.

app/models/participant_declaration.rb
blackrat: There (probably) isn't going to be any way to hit this state. Once payable, it should never go back to submitted.

app/models/participant_declaration.rb
blackrat: As I mentioned earlier, I'm probably going to make this more transition aware. Ideally, I'd like to remove the dodgy has_one, has_many that we're using for historical transition recording.

app/serializers/participant_declaration_serializer.rb
blackrat: Originally it was "state.state". :)

app/services/record_declarations/base.rb
blackrat: Yes. If it's deemed payable at the point of creation, then it's set that way. I actually want the eligibility checks to trigger this state change so we don't need to recheck it later.

app/services/record_declarations/retained/base.rb
blackrat: Yes. Was annoying me that these were in 4 places.

db/schema.rb
```diff
@@ -173,6 +173,15 @@
...
   end
 
+  create_table "declaration_states", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "participant_declaration_id", null: false
+    t.string "state"
+    t.string "submitted"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.index ["participant_declaration_id"], name: "index_declaration_states_on_participant_declaration_id"
+  end
```
kicferk1, line 183: `t.string "submitted"` looks odd, why is it in the schema?

app/models/participant_declaration.rb
kicferk1: Ehhh, who knows. Participant getting removed from DQT in a very timely manner sounds like it could make them non-fundable.

app/serializers/participant_declaration_serializer.rb
kicferk1: I mean the attribute we surface to providers, what we call it internally, ehh, less fussed about that

app/services/record_declarations/actions/with_milestone.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      private
+      attr_reader :milestone
+
+      def initialize(milestone:)
+        @milestone=milestone
+      end
+    end
+  end
+end
```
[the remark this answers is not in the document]
blackrat: Oops. This is for the next PR. There are multiples.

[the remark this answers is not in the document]
blackrat: There will be other instances of its use, but I'll pull it from this PR, since it doesn't really belong with this one.

kicferk1 (approved):
🚢

PR #1189, opened 2021-10-05
Only return 2021 participants to lead provider in API

ethanmills:
The addition of NQT+1 as 2020 participants caused them to appear in the participants results for lead providers where they are at a partnered school. This change limits results to participants in the current cohort (for ECF only)

### DIFF
--- a/config/initializers/preload_stis.rb
+++ b/config/initializers/preload_stis.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+# Taken from https://github.com/rails/rails/blob/main/guides/source/autoloading_and_reloading_constants.md
+# Makes ParticipantProfile::ECF behave properly in dev and test
+unless Rails.application.config.eager_load
+  Rails.autoloaders.main.on_load("ParticipantProfile") do
+    ParticipantProfile.connection.select_values(<<~SQL).each(&:constantize)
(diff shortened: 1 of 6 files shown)


config/initializers/preload_stis.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Taken from https://github.com/rails/rails/blob/main/guides/source/autoloading_and_reloading_constants.md
+# Makes ParticipantProfile::ECF behave properly in dev and test
+unless Rails.application.config.eager_load
+  Rails.autoloaders.main.on_load("ParticipantProfile") do
+    ParticipantProfile.connection.select_values(<<~SQL).each(&:constantize)
```
kicferk1, line 7: I just love how github tells me "Possible SQL injection". 

But we control everything that goes into this query, right?

kicferk1: (I guess we could only enable it in dev and test if we really needed to)

kicferk1 (approved):

config/initializers/preload_stis.rb
ethanmills: Yes, you'd have to inject something into the database in the type column to do anything malicious I think. This will already only run on dev/test because of the `unless Rails.application.config.eager_load`

PR #1190, opened 2021-10-05
CPDRP-999: Get previous induction data from DQT API

ethanmills:
## Ticket and context

Ticket: CPDPR-999

We should take information on which ECTs have had a previous induction from the DQT API. This informs which participants are eligible for funding.

- Removed old temporary DQT API
- changed ineligible participant semantics. I will follow up with a migration to simplify the data stored here to a boolean for previous participation once we're happy this works well

### Code quality checks


Gemfile.lock
```diff
@@ -678,7 +678,7 @@ DEPENDENCIES
   with_model
 
 RUBY VERSION
-   ruby 3.0.2p107
+   ruby 2.7.2p137
```
ethanmills, line 681: Looks like this got missed before - my mistake

app/services/participant_validation_service.rb
```diff
@@ -19,75 +19,60 @@ def validate
...
   end
 
   def check_first_name_only?
     config[:check_first_name_only]
   end
 
   def dqt_record(trn, nino)
-    if FeatureFlag.active?(:full_dqt_api)
-      object = full_dqt_client.get_record(trn: trn, birthdate: date_of_birth, nino: nino)
-      FullDqt::OldRecordMapper.translate(object)
```
kicferk1, line 56: Should we cut out `FullDqt::OldRecordMapper` too?

kicferk1 (commented):
Is most of the changes in test caused by switching to only using new dqt api?

app/services/participant_validation_service.rb
ethanmills: Yep, good spot. Had intended to

ethanmills:
> Is most of the changes in test caused by switching to only using new dqt api?

Yes mostly. There are a couple of new tests on the induction start date logic, but mostly just changing the stubbing

ethanmills:
We have induction start date in prod now, so this is unblocked

kicferk1 (approved):

app/services/participant_validation_service.rb
```diff
@@ -19,75 +19,60 @@ def validate
...
-    validation_data
+    validation_data["induction"]["start_date"] < Date.new(Cohort.current.start_year, 9, 1)
   end
 
   def check_first_name_only?
     config[:check_first_name_only]
   end
 
   def dqt_record(trn, nino)
-    if FeatureFlag.active?(:full_dqt_api)
```
asmega, line 54: not important but we should able to remove the flag itself, can go in a later patch. Theres also a boat load of tests that can also be go as I just copied the tests, one with the flag and one without.

asmega: looks like you already removed the tests 👍

asmega (approved):

app/services/participant_validation_service.rb
ethanmills: Yep, I'll remove the dependency on the flag, release, deactivate the flag, then remove from the code

PR #1205, opened 2021-10-07
CPDTP-526: Add N+1 queries, but ensure the participant state is correct

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-526

This seems to be a Rails issue. Order is not respected when you preload relations. https://github.com/rails/rails/issues/6769

There are a few ways we could fix it - use our own sql to load the ecf data, or store the current state on profile. But in the meantime the easiest is N+1 queries 😢


kicferk1:
No longer needed, we have a better solution!

PR #1208, opened 2021-10-07
CPDTP-526: Add training status to participant profile

kicferk1:
## Ticket and context

https://github.com/DFE-Digital/early-careers-framework/pull/1205 has a temporary solution to our problem.

Permanent solution is to cache the training status on participant profile, so we don't need to do a nested ordered query.

This is first step of that - adding the field in the db. After I merge and deploy that, I will run a task to populate the value on existing profiles. After I do that I will push another PR which switches to using the field in the serializer.

### DIFF
--- a/app/services/participants/withdraw/validate_and_change_state.rb
+++ b/app/services/participants/withdraw/validate_and_change_state.rb
@@ -16,7 +16,7 @@ module Participants
       def perform_action!
         ActiveRecord::Base.transaction do
           ParticipantProfileState.create!(participant_profile: user_profile, state: ParticipantProfileState.states[:withdrawn], reason: reason)
-          user_profile.training_status_withdrawn
+          user_profile.training_status_withdrawn!
         end
         user_profile
       end


app/models/participant_profile_state.rb
```diff
@@ -27,4 +27,8 @@ def valid?(_context = nil)
   }
 
   scope :most_recent, -> { order("created_at desc").limit(1) }
+
+  after_create do |profile_state|
```
StupidCodeFactory, line 31: Suggestion: I would strongly advise no to use active record callback, this usually leads to spaghetti code and hide the complexity deep into the database layer.
A more streamlined solution to active record callbacks are to do this inside the transaction were the record is created or updated.

db/migrate/20211007131604_add_training_status_to_participant_profile.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddTrainingStatusToParticipantProfile < ActiveRecord::Migration[6.1]
+  def change
+    add_column :participant_profiles, :training_status, :string
```
StupidCodeFactory, line 5: you'd probably want to have a default and a not null constraint

spec/support/shared_examples/participant_actions_defer_support.rb
```diff
@@ -11,6 +11,7 @@
   it "creates a deferred state and makes the profile deferred" do
     expect { described_class.call(params: participant_params) }.to change { ParticipantProfileState.count }.by(1)
     expect(user_profile.participant_profile_state.deferred?)
+    expect(user_profile.training_status_deferred?)
```
StupidCodeFactory, line 14: that's not going to do anything

`expect(user_profile).to be_training_status_deferred`

PR #1230, opened 2021-10-12
Couple of chores

StupidCodeFactory:
## Tech review

1. add spring rspec command for faster command and test run
2. update the rails bin to be inline with the rails version
3. generate the rspec binstub

Note that all commands in the `bin` directory have been updated with `bin/rails app:update:bin`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -155,7 +155,6 @@ group :development do
 
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem "spring"
-  gem "spring-commands-rspec"
   gem "spring-watcher-listen", "~> 2.0.0"
   # State machine diagrams - https://github.com/Katee/aasm-diagram
   gem "aasm-diagram"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -525,8 +525,6 @@ GEM
     simplecov_json_formatter (0.1.3)
     socksify (1.7.1)
     spring (2.1.1)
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -155,8 +155,8 @@ group :development do
 
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem "spring"
+  gem "spring-commands-rspec"
```
kicferk1, line 158: Could you add a note to the readme on how to use it?

Is it compatible with running rspec in parallel? If not, which do we want to pick?

StupidCodeFactory: ran parallel test and it all works fine!

bin/spring
```diff
@@ -1,10 +1,17 @@
 #!/usr/bin/env ruby
-if !defined?(Spring) && [nil, "development", "test"].include?(ENV["RAILS_ENV"])
-  # Load Spring without loading other gems in the Gemfile, for speed.
-  require "bundler"
-  Bundler.locked_gems.specs.find { |spec| spec.name == "spring" }&.tap do |spring|
+
+# This file loads Spring without using Bundler, in order to be fast.
+# It gets overwritten when you run the `spring binstub` command.
+
+unless defined?(Spring)
```
kicferk1, line 6: So I think this is actually what breaks your deploy. The condition used to check that the environment is something that we have spring in - but it does not anymore. So it will attempt to load it, well, everywhere.

jamesrose:
Closing as stale.

PR #1240, opened 2021-10-13
CPDTP-538: NPQ Application name cleanup part 1, ValidationData -> Application

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-538

There will be some more renaming, but I think to avoid too many conflicts and this thing becoming too complicated, I will make 1PR per naming cleanup.

In this particular case, the business name we use to talk about the data that a teacher sends us is "NPQ application". The code name for it is "NPQValidationData". We should switch to latter in code I think. 

### Code quality checks

### DIFF
--- a/app/components/admin/participants/details/npq.html.erb
+++ b/app/components/admin/participants/details/npq.html.erb
@@ -28,11 +28,11 @@
     </tr>
     <tr class="govuk-table__row">
       <th scope="row" class="govuk-table__header">Provider</th>
-      <td class="govuk-table__cell"><%= profile.npq_application&.npq_lead_provider&.name %></td>
+      <td class="govuk-table__cell"><%= profile.npq_application.npq_lead_provider.name %></td>
     </tr>
     <tr class="govuk-table__row">
       <th scope="row" class="govuk-table__header">NPQ course</th>
-      <td class="govuk-table__cell"><%= profile.npq_application&.npq_course&.name %></td>
+      <td class="govuk-table__cell"><%= profile.npq_application.npq_course.name %></td>
     </tr>
     <tr class="govuk-table__row">
       <th scope="row" class="govuk-table__header">Last updated</th>
--- a/app/components/admin/participants/details/npq_pending.html.erb
+++ b/app/components/admin/participants/details/npq_pending.html.erb
@@ -36,11 +36,11 @@
     </tr>
     <tr class="govuk-table__row">
       <th scope="row" class="govuk-table__header">Provider</th>
-      <td class="govuk-table__cell"><%= profile.npq_application&.npq_lead_provider&.name %></td>
+      <td class="govuk-table__cell"><%= profile.npq_application.npq_lead_provider.name %></td>
     </tr>
     <tr class="govuk-table__row">
       <th scope="row" class="govuk-table__header">NPQ course</th>
-      <td class="govuk-table__cell"><%= profile.npq_application&.npq_course&.name %></td>
+      <td class="govuk-table__cell"><%= profile.npq_application.npq_course.name %></td>
     </tr>
     <tr class="govuk-table__row">
       <th scope="row" class="govuk-table__header">Last updated</th>


app/models/npq_application.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class NPQValidationData < ApplicationRecord
+class NPQApplication < ApplicationRecord
   has_paper_trail only: %i[user_id npq_lead_provider_id npq_course_id created_at updated_at lead_provider_approval_status]
   # TODO: Rename table
   self.table_name = "npq_profiles"
```
StupidCodeFactory, line 6: Let's rename the table as well!

app/models/participant_profile/npq.rb
```diff
@@ -5,7 +5,7 @@ class ParticipantProfile::NPQ < ParticipantProfile
   belongs_to :school, optional: true
   belongs_to :npq_course, optional: true
 
-  has_one :validation_data, class_name: "NPQValidationData", foreign_key: :id, dependent: :destroy
+  has_one :validation_data, class_name: "NPQApplication", foreign_key: :id, dependent: :destroy
```
StupidCodeFactory, line 8:
```suggestion
  has_one :npq_application, dependent: :destroy
```

and the subsequent calls to `profile.validation_data` => `profile.npq_application`

StupidCodeFactory (changes requested):

app/models/participant_profile/npq.rb
blackrat: Personally, I'd like to see the actual validation data extracted since that's not really part of the application and the application record pointing to it.

app/controllers/api/v1/npq_profiles_controller.rb
```diff
@@ -6,7 +6,7 @@ class NPQProfilesController < Api::ApiController
       include ApiTokenAuthenticatable
 
       def create
-        validation_data = NPQValidationData.new(validation_data_params)
+        validation_data = NPQApplication.new(validation_data_params)
```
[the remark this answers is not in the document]
kicferk1: I don't want to split it into multiple models. I just missed that name cause I searched globally for `npq_validation_data`, to avoid the ecf validation data variables :P

app/models/npq_application.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class NPQValidationData < ApplicationRecord
+class NPQApplication < ApplicationRecord
```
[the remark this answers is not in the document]
kicferk1: This is just to capture what the user sent us, I think.

Also, it's definitely beyond the scope of this PR to change any db models.

app/controllers/api/v1/npq_profiles_controller.rb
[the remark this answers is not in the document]
blackrat: Possibly not in this PR, but tracing through it, it really cries out to be done at some point.

app/components/admin/participants/details/npq.html.erb
```diff
@@ -28,11 +28,11 @@
     </tr>
     <tr class="govuk-table__row">
       <th scope="row" class="govuk-table__header">Provider</th>
-      <td class="govuk-table__cell"><%= profile.validation_data&.npq_lead_provider&.name %></td>
+      <td class="govuk-table__cell"><%= profile.npq_application&.npq_lead_provider&.name %></td>
```
blackrat, line 31: Can this be nil if there's an application?

app/models/npq_lead_provider.rb
```diff
@@ -3,6 +3,6 @@
 class NPQLeadProvider < ApplicationRecord
   belongs_to :cpd_lead_provider, optional: true
 
-  has_many :npq_profiles, class_name: "NPQValidationData"
+  has_many :npq_profiles, class_name: "NPQApplication"
```
StupidCodeFactory, line 6: That one as well is confusing. However maybe let's do it in its own separate little PR.

Profiles are the `ParticipantProfile::(ECF||NPQ)` 

@kicferk1 would you mind creating a ticket in the backlog please?

kicferk1: I might spend some time in the afternoon fixing it :D 

I think https://dfedigital.atlassian.net/browse/CPDTP-538 mentions this too, but I agree, separate PR for this is a good idea :)

PR #1247, opened 2021-10-13
Rename npq_profiles relation to npq_applications where appropriate

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-538

This PR was simpler than expected, so I decided to also use it as the first step of renaming the `npq_profiles` table. The new name: `npq_applications`

Once this gets in, I will need to run a task on sandbox and prod to populate the data in the new table. And then we can remove the old table.

### DIFF
--- a/app/models/npq_application.rb
+++ b/app/models/npq_application.rb
@@ -9,7 +9,9 @@ class NPQApplication < ApplicationRecord
   # Step 3 (Not done): Run a script to migrate all existing data to new table
   # Step 4 (Not done): Push a new change removing the old table
   self.table_name = "npq_profiles"
-  after_commit do |application|
+  after_save { |application| synchronise_temporary_table(application) }
+  after_touch { |application| synchronise_temporary_table(application) }
+  def synchronise_temporary_table(application)
     NPQApplicationTemporary.find_or_initialize_by(id: application.id).update!(application.attributes)
   end
 


app/models/npq_application.rb
```diff
@@ -2,8 +2,16 @@
...
   has_paper_trail only: %i[user_id npq_lead_provider_id npq_course_id created_at updated_at lead_provider_approval_status]
+
   # TODO: Rename table
+  # Step 1 (Done): Make a new table, called npq_applications
+  # Step 2 (Done): Write to both tables
+  # Step 3 (Not done): Run a script to migrate all existing data to new table
+  # Step 4 (Not done): Push a new change removing the old table
   self.table_name = "npq_profiles"
+  after_commit do |application|
+    NPQApplicationTemporary.find_or_initialize_by(id: application.id).update!(application.attributes)
```
ethanmills, line 13: Do we ever rely on the `created_at`? If so, is this included in `attributes`?

ethanmills (approved):

app/models/npq_lead_provider.rb
```diff
@@ -3,6 +3,6 @@
 class NPQLeadProvider < ApplicationRecord
   belongs_to :cpd_lead_provider, optional: true
 
-  has_many :npq_profiles, class_name: "NPQApplication"
-  has_many :npq_participants, through: :npq_profiles, source: :user
+  has_many :npq_applications, class_name: "NPQApplication"
+  has_many :npq_participants, through: :npq_applications, source: :user
```
blackrat, line 7: Is this right? Shouldn't we separate who are actually npq_participants and what is npq_applications?

kicferk1: I think you are right - this should be something like 
```
  has_many :npq_applications, class_name: "NPQApplication"
  has_many :npq_participant_profiles, class_name: "ParticipantProfile::NPQ", through: :npq_applications, source: :profile
  has_many :npq_participants, through: :npq_participant_profiles, source: :user
```

blackrat: Logic on those associations feels a little bit backwards, but I get that the profile is created from the application and not the other way around.

app/controllers/api/v1/npq_participants_controller.rb
```diff
@@ -19,7 +19,6 @@ def npq_lead_provider
       end
 
       def npq_participants
-        # TODO: filter out on approved participants
```
StupidCodeFactory, line 22: do we not need to do that anymore?

app/models/npq_application.rb
StupidCodeFactory: @kicferk1 you want to do this within the transaction. if one fail you want to rollback both records

PR #1249, opened 2021-10-14
Fix accidentally showing some users as 2020 participants to providers

kicferk1:
## Ticket and context

Turns out there are people who are ECTs in year 2020, and mentors in 2021, and we are surfacing some of them as 2020 ects. Which is incorrect.

### DIFF
--- a/app/models/teacher_profile.rb
+++ b/app/models/teacher_profile.rb
@@ -10,8 +10,8 @@ class TeacherProfile < ApplicationRecord
   has_one :early_career_teacher_profile, -> { active_record }, class_name: "ParticipantProfile::ECT"
   has_one :mentor_profile, -> { active_record }, class_name: "ParticipantProfile::Mentor"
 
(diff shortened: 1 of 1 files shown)


app/models/teacher_profile.rb
```diff
@@ -9,7 +9,9 @@ class TeacherProfile < ApplicationRecord
   # TODO: Legacy associations, to be removed
   has_one :early_career_teacher_profile, -> { active_record }, class_name: "ParticipantProfile::ECT"
   has_one :mentor_profile, -> { active_record }, class_name: "ParticipantProfile::Mentor"
-  has_one :ecf_profile, -> { active_record }, class_name: "ParticipantProfile::ECF"
+
+  has_many :ecf_profiles, -> { active_record.includes(:cohort).order("cohort.start_year DESC") }, class_name: "ParticipantProfile::ECF"
```
ethanmills, line 13: I hope the answer is no, but does the fact that this table has a cohort column confuse matters?

kicferk1: I think it could. I made it more explicit, so hopefully not anymore.

PR #1257, opened 2021-10-14
Switch to use npq_applications table for NPQApplication model, drop npq_profiles table

kicferk1:
## Ticket and context

I have copied the data over from `npq_profiles` to `npq_applications` on sandbox and in prod, and we have a hook updating it live.

Time to axe the old table.

### DIFF
--- a/spec/requests/api/v1/npq_participants_spec.rb
+++ b/spec/requests/api/v1/npq_participants_spec.rb
@@ -77,6 +77,17 @@ RSpec.describe "NPQ Participants API", type: :request, with_feature_flags: { par
         end
 
         context "filtering" do
+          before do
+            current_time = Time.zone.now
+            travel_to 10.days.ago
+            list = create_list(:npq_application, 3, npq_lead_provider: npq_lead_provider, school_urn: "123456", npq_course: npq_course)
+
+            list.each do |npq_application|
+              NPQ::Accept.new(npq_application: npq_application).call
+            end
+            travel_to current_time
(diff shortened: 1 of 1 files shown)


db/migrate/20211014133150_drop_npq_profiles.rb
```diff
@@ -0,0 +1,26 @@
...
+      t.references :npq_lead_provider, null: false, foreign_key: true, type: :uuid
+      t.references :npq_course, null: false, foreign_key: true, type: :uuid
+
+      t.date :date_of_birth
+      t.text :teacher_reference_number
+      t.boolean :teacher_reference_number_verified, default: false
+      t.text :school_urn
+      t.text :headteacher_status
+
+      t.boolean :active_alert, default: false
```
blackrat, line 16: Same. As we're doing it for the eligible_for_funding why have the possiblliity of 3 state booleans elsewhere.

spec/requests/api/v1/npq_participants_spec.rb
```diff
@@ -77,10 +77,6 @@
         end
 
         context "filtering" do
-          before do
-            create_list(:npq_application, 2, npq_lead_provider: npq_lead_provider, updated_at: 10.days.ago, school_urn: "123456", npq_course: npq_course)
```
StupidCodeFactory, line 81: I don't quite understand why this is no longer needed? 
is it because now we filter on `users.updated_at`?

PR #1261, opened 2021-10-14
Add a sandbox job to generate test data for EDT regularly

kicferk1:
## Ticket and context

Slack channel conversation. EDT would like a regular new applications and ecf participants, so they can check their code importing new records on a regular basis.

### DIFF
--- a/app/jobs/create_new_fake_sandbox_data_job.rb
+++ b/app/jobs/create_new_fake_sandbox_data_job.rb
@@ -3,22 +3,16 @@
 class CreateNewFakeSandboxDataJob < CronJob
   self.cron_expression = "0 2 * * *"
 
+  EDT_NAME = "Education Development Trust"
   def perform
-    return unless Rails.env.sandbox? || Rails.env.test?
+    return if Rails.env.production?
 
-    provider_name = "Education Development Trust"
-
-    edt_ecf_lead_provider = LeadProvider.find_by(name: provider_name)
-    edt_npq_lead_provider = NPQLeadProvider.find_by(name: provider_name)
-
-    school = edt_ecf_lead_provider.schools.first
-    school_cohort = SchoolCohort.find_by(school: school, cohort: Cohort.current)
     10.times do
       name = Faker::Name.name
       EarlyCareerTeachers::Create.call(
         full_name: name,
         email: Faker::Internet.email(name: name),
-        school_cohort: school_cohort,
+        school_cohort: random_edt_school_cohort,
         mentor_profile_id: nil,
         year_2020: false,
       )
@@ -34,7 +28,7 @@ class CreateNewFakeSandboxDataJob < CronJob
         funding_choice: "",
         headteacher_status: "",
         nino: "",
-        school_urn: school.urn,
+        school_urn: random_edt_school.urn,
         teacher_reference_number: TRNGenerator.next,
         teacher_reference_number_verified: true,
         npq_course: NPQCourse.all.sample,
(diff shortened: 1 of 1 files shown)


app/jobs/create_new_fake_sandbox_data_job.rb
```diff
@@ -0,0 +1,46 @@
...
+    provider_name = "Education Development Trust"
+
+    edt_ecf_lead_provider = LeadProvider.find_by(name: provider_name)
+    edt_npq_lead_provider = NPQLeadProvider.find_by(name: provider_name)
+
+    school = edt_ecf_lead_provider.schools.first
+    school_cohort = SchoolCohort.find_by(school: school, cohort: Cohort.current)
+    10.times do
+      name = Faker::Name.name
+      EarlyCareerTeachers::Create.call(
```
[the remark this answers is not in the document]
kicferk1: > Hi , We have flows running nightly looking for updates and new data. We would like to see these working picking up changes to records as well as pulling in new ones. Happy to discuss different timings/approaches.

I guess I could also update some existing resords, but I don't have a clue what with. New faker emails? :P

app/jobs/create_new_fake_sandbox_data_job.rb
```diff
@@ -0,0 +1,46 @@
...
+  def perform
+    return unless Rails.env.sandbox? || Rails.env.test?
+
+    provider_name = "Education Development Trust"
+
+    edt_ecf_lead_provider = LeadProvider.find_by(name: provider_name)
+    edt_npq_lead_provider = NPQLeadProvider.find_by(name: provider_name)
+
+    school = edt_ecf_lead_provider.schools.first
+    school_cohort = SchoolCohort.find_by(school: school, cohort: Cohort.current)
```
ngw, line 15: Maybe you can extract `edt_ecf_lead_provider`, `edt_nqp_lead_provider`, `school` and `school_cohort` in private methods?

app/jobs/create_new_fake_sandbox_data_job.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def perform
+    return unless Rails.env.sandbox? || Rails.env.test?
+
+    provider_name = "Education Development Trust"
+
+    edt_ecf_lead_provider = LeadProvider.find_by(name: provider_name)
+    edt_npq_lead_provider = NPQLeadProvider.find_by(name: provider_name)
+
+    school = edt_ecf_lead_provider.schools.first
```
ngw, line 14: If I understand this correctly, you just want a random `LeadProvider` school here, right? First should order by id, and schools use uuid. In case I misunderstood and order is important you probably want something like `self.implicit_order_column = "created_at"` in your `School` model.

app/jobs/create_new_fake_sandbox_data_job.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+class CreateNewFakeSandboxDataJob < CronJob
+  self.cron_expression = "0 2 * * *"
+
+  def perform
+    return unless Rails.env.sandbox? || Rails.env.test?
+
+    provider_name = "Education Development Trust"
```
ngw, line 9: Maybe this can be a constant?

lib/tasks/cron.rake
```diff
@@ -7,5 +7,6 @@ namespace :cron do
     ImportGiasDataJob.schedule
     ValidationRetryJob.schedule
     SchoolAnalyticsJob.schedule
+    CreateNewFakeSandboxDataJob.schedule if Rails.env.sandbox?
```
ngw, line 10: Looks like you have a guard (a different guard) at the beginning of `perform`

kicferk1: Yeah, cause I am paranoid that a dev will accidentally run this job where it shouldn't be ran 😂

PR #1264, opened 2021-10-15
Make sure NQT+1s are immune from providers API

kicferk1:
## Ticket and context

Providers should not be able to do anything with NQT+1s.
Turns out by a lucky coincidence (well, not that lucky, part of it is by design) they can't. But we should add tests showing that that's the case, and preserving that behaviour for the future.


kicferk1:
> Have we covered off the weird case of the NQT+1 & 2021 mentor?

The provider needs to tell us which 'course_identifier' they are submitting a declaration for. That is used to fetch the relevant participant profile, and the service goes from there.

If they submit one for 'ecf-induction', it's gonna be treated like in the tests in this PR. 

If they submit it for 'ecf-mentor' then it's gonna be treated like one of the existing tests. 

I can't really see any interaction between the two that we need to worry about.

PR #1269, opened 2021-10-15
Correct redirect when no programme is chosen

ethanmills:
Fixes https://sentry.io/organizations/dfe-bat/issues/2721505226/?project=5748989&query=is%3Aunresolved+is%3Afor_review+assigned_or_suggested%3A%5Bme%2C+none%5D&sort=inbox

### DIFF
--- a/app/controllers/schools/base_controller.rb
+++ b/app/controllers/schools/base_controller.rb
@@ -40,6 +40,6 @@ private
       school: @school,
     )
 
-    redirect_to advisory_schools_choose_programme_path unless @school_cohort
+    redirect_to schools_choose_programme_path(cohort_id: Cohort.current.start_year) unless @school_cohort
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/schools/base_controller.rb
```diff
@@ -40,6 +40,6 @@ def set_school_cohort
       school: @school,
     )
 
-    redirect_to advisory_schools_choose_programme_path unless @school_cohort
+    redirect_to schools_choose_programme_path(cohort_id: Cohort.current.start_year) unless @school_cohort
```
ngw, line 43: By just reading the code I'm not 100% sure of what happens if @school_cohort, and specs don't help me. Is there a #set_school_cohor view anywhere? I see the method gets overridden in other controllers without calling super, what is the app flow here?

PR #1272, opened 2021-10-16
CPDRP-883: make roles page accessible to all

ethanmills:
Fixes https://sentry.io/organizations/dfe-bat/issues/2714281077

## Ticket and context

Ticket: CPDRP-883

Make the roles page publicly accessible

### DIFF
--- a/app/controllers/schools/cohorts_controller.rb
+++ b/app/controllers/schools/cohorts_controller.rb
@@ -21,8 +21,6 @@ module Schools
       end
     end
 
-    def roles
-      render "shared/roles"
-    end
+    def roles; end
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/schools/cohorts_controller.rb
```diff
@@ -21,8 +21,6 @@ def add_participants
       end
     end
 
-    def roles
-      render "shared/roles"
-    end
+    def roles; end
```
BroiSatse, line 24: Do we need it?

app/views/shared/_roles.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <span class="govuk-caption-xl"><%= @school.name %></span>
+    <span class="govuk-caption-xl"><%= @school&.name %></span>
```
BroiSatse, line 7: Should this be wrapped in `if` to avoid empty html tag?

PR #1274, opened 2021-10-16
Remove old feature flags

ethanmills:
I've gone through and removed a load of references to dead feature flags. Once this goes in, I'll disable them (to keep the db clean) then remove them from the list in the service


app/views/admin/participants/show.html.erb
```diff
@@ -9,12 +9,10 @@
 <%= render Admin::Participants::Details.new(profile: @participant_profile) %>
 <%= render Admin::Participants::ValidationTasks.new(profile: @participant_profile) %>
 
-<% if FeatureFlag.active?(:admin_delete_participants) %>
-  <% unless @participant_profile.npq? %>
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
-        <%= govuk_link_to "Delete participant", remove_admin_participant_path %>
-      </div>
+<% unless @participant_profile.npq? || @participant_profile.completed_validation_wizard? %>
```
ethanmills, line 12: This is a behaviour change

PR #1278, opened 2021-10-18
Allowed sending blank personalisations to notify

BroiSatse:
When implementing EmailRedirector there was an assumption that Notify does not accept blank personalisation values. This however was proven not to be the case and it was the `mail-notify` gem that had internally removed all the blank personalisations before sending the request to notify.

This PR monkey patches the gem to allow sending blank personalisations via `template_mail` method. Additionally, it introduces a new personalisation to be sent with all the emails - `subject_tags`. This personalisation will be a blank string when redirector is disabled (production environment), in all other environments it will contain information about the environment and the original recipient of the email.

Note: as this might be potentially dangerous to allow blank personalisations to go through I have decided to make them explicit. After this PR, all the email deliveries with blank personalisation will still fail unless it was explicitly said in the code that this is expected to be blank - I have also added a convenience `blank_allowed` method to our mailer:

```
  personalisation: {
    trn: teacher_profile.trn #=> raises if blank
    note: blank_allowed(note) #=> explicitly blank
}
```

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -3,7 +3,7 @@
 class ApplicationMailer < Mail::Notify::Mailer
 private
 
-  def blank
-    Mail::Notify::Personalisation::BLANK
+  def blank_allowed(value)
+    value.presence || Mail::Notify::Personalisation::BLANK
   end
 end


app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
 
 class ApplicationMailer < Mail::Notify::Mailer
+private
+
+  def blank_allowed(value)
+    value.presence || Mail::Notify::Personalisation::BLANK
+  end
```
kicferk1, line 8: Can we open a PR in the gem that adds similar functionality? Or an issue?

lib/extensions/notify_delivery_method.rb
```diff
@@ -0,0 +1,19 @@
...
+    class Personalisation
+      BLANK = Object.new
+
+      module BlankFix
+        def to_h
+          super.transform_values do |value|
+            value == BLANK ? "" : value
+          end
+        end
+      end
```
kicferk1, line 14: I am not quite sure why we need a constant for blank. The reasonable behaviour, to me, is that when you pass something empty or falsy as a value in personalisation, it gets to be an empty string when sent to notify. 

So something like 
```
      def to_h
        merged_options.map { |_k, v| v.blank? ? "" : v }
      end
```

Thoughts?

app/mailers/application_mailer.rb
BroiSatse: Done: https://github.com/dxw/mail-notify/pull/30

lib/extensions/notify_delivery_method.rb
BroiSatse: I definitely prefer `fail-by-default` here. Yes, it is up for us to handle our blanks, which is why we need to mark those personalisation as potentially blank if we really want them to be blank. All the other blank has to fail, as otherwise we're sending incomplete emails.

There are not gona be a lot of places to do that mapping either - this would only happen in mailer as it is closely related to the structure of the template. I've added `blank_allowed` method to make it easier and to encapsulate the existence of `BLANK` constant. I had to do manual mapping in the Redirector as this happens outside of the mailer, but I don't think there are any other places we'd need to do this.

PR #1311, opened 2021-10-20
CPDTP-562: Add Apply setup for tracing requests

kicferk1:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDTP-562

Apply fire off a background job to store the api requests. I can look into doing that as well, if we think it would cause a noticable hit in performance.

This stores significantly more than `ApiRequestAudit`, so I intend to remove that in a later PR. 

### Code quality checks

### DIFF
--- a/app/jobs/api_request_job.rb
+++ b/app/jobs/api_request_job.rb
@@ -9,7 +9,8 @@ class ApiRequestJob
 
   def perform(request_data, response_data, status_code, created_at)
     request_headers = request_data.fetch(:headers, {})
-    provider = provider_from_auth_token(request_headers.delete("HTTP_AUTHORIZATION"))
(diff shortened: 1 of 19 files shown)


app/jobs/api_request_job.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    ApiRequest.create!(
+      request_path: request_data[:path],
+      request_headers: request_headers,
+      request_body: request_body(request_data),
+      request_method: request_data[:method],
+      response_headers: response_headers,
+      response_body: response_hash(response_body, status_code),
+      status_code: status_code,
+      user_description: provider.present? ? "CpdLeadProvider" : "Other - NPQ or E&L",
```
jamesrose, line 25: Not sure this should be in the model — seems more suited to the view.

ethanmills (approved):

PR #1312, opened 2021-10-20
Do not send ineligibility emails about ERO mentors

ethanmills:
We don't want to send these emails any more because we're not sure what's going on with them

### DIFF
--- a/app/services/store_participant_eligibility.rb
+++ b/app/services/store_participant_eligibility.rb
@@ -40,6 +40,8 @@ private
   def send_ineligible_notification_email
     participant_profile.school_cohort.school.induction_coordinators.each do |induction_tutor|
       mailer_name = "#{participant_profile.participant_type}_#{@participant_eligibility.reason}_email"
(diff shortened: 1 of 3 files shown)


app/services/store_participant_eligibility.rb
```diff
@@ -42,6 +42,8 @@ def send_ineligible_notification_email
       mailer_name = "#{participant_profile.participant_type}_#{@participant_eligibility.reason}_email"
       if IneligibleParticipantMailer.respond_to? mailer_name
         IneligibleParticipantMailer.send(mailer_name, **{ induction_tutor_email: induction_tutor.email, participant_profile: participant_profile }).deliver_later
+      elsif mailer_name == "mentor_previous_participation_email"
```
tonyheadford, line 45: I'd be tempted to short circuit it before the `If IneligibleParticipantMailer.respond_to?` line i.e.

```ruby
next if mailer_name == "mentor_previous_participation_email"

If IneligibileParticipantMailer.respond_to? mailer_name
...
```

tonyheadford (approved):
Non-blocking comment just to clean up the conditional block

PR #1313, opened 2021-10-20
CPDRP-1034 Parallalize rspec and cypress

FriedSock:
## Ticket and context

First stab at speeding up the test suite. The rspec and cypress steps take the longest to run, so even if we are moving away from cypress, it makes sense to speed up the step so that the overall test suite run is actually sped up in the interim.

Use 6 machines to run the rspec and cypress tests in parallel rather than the one currently being used. There are diminishing returns for each extra node we add due to the set up time that needs to be run on each node.

Ticket: https://dfedigital.atlassian.net/browse/CPDRP-1034

### Code quality checks

### HTML Checks

### Gotchas

## External API changes


spec/cypress.json
```diff
@@ -1,4 +1,5 @@
 {
   "baseUrl": "http://localhost:5017",
-  "testFiles": ["**/*.js", "**/*.feature"]
+  "testFiles": ["**/*.js", "**/*.feature"],
+  "projectId": "v7d8uc"
```
ethanmills, line 4: Should this be secret?

PR #1322, opened 2021-10-20
API docs: Link request bodies and more parameters

kicferk1:
## Ticket and context

We were missing some links from our api docs. This adds request bodies and tweaks display of schemas so that more of nested schemas are linked together.


jamesrose:
Screenshot 2021-10-21 at 08 53 06

"This returns a ECFParticipantChangeSchedule schema." doesn't seem quite accurate to me. That request accepts a body conforming to that schema - it doesn't return it.

kicferk1:
> "This returns a ECFParticipantChangeSchedule schema." doesn't seem quite accurate to me. That request accepts a body conforming to that schema - it doesn't return it.

When we are presenting schemas it is not very easy to know whether it is returned or consumed. I will think of some neutral wording.

PR #1327, opened 2021-10-21
CPDRP-1038: Allow replacing a SIT who was previously a mentor

ethanmills:
## Ticket and context

Ticket: CPDRP-1038
Sentry: https://sentry.io/organizations/dfe-bat/issues/2734624887/activity/?project=5748989&query=is%3Aunresolved+is%3Afor_review+assigned_or_suggested%3A%5Bme%2C+none%5D&sort=inbox

This induction tutor has a withdrawn mentor profile, which has validation data on it


kicferk1 (approved):
I wonder if we should not delete those users directly, and just have a job that deletes users without roles that were created/updated more than a day ago. Thoughts?

PR #1343, opened 2021-10-25
New validation journey

BroiSatse:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDRP/boards/102?selectedIssue=CPDRP-1004


app/views/participants/validations/no_match.html.erb
```diff
@@ -0,0 +1,62 @@
...
+          <dt class="govuk-summary-list__key">TRN</dt>
+          <dd class="govuk-summary-list__value">
+            <% if validation_form.no_trn %>
+              Not provided
+            <% else %>
+              <%= validation_form.trn %>
+            <% end %>
+          </dd>
+          <dd class="govuk-summary-list__actions">
+            <%#= govuk_link_to "Change", { action: :show, step: :trn } %>
```
ethanmills, line 21: ?

app/views/participants/validations/dob.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= title %></h1>
+    <%= form_for validation_form, url: { action: :update }, method: :patch do |f| %>
+      <%= f.govuk_date_field :dob,
+        date_of_birth: true,
+        width: "three-quarters",
+        legend: { text: "Your date of birth", class: "govuk-visually-hidden" },
+        hint: { text: "For example 12 11 1990"}
```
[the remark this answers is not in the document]
Lloyd-K: To meet the guidance, please use '14 7 1990'. Thanks

app/views/participants/validations/eligible.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_panel title_text: title, text: nil, classes: "govuk-!-margin-bottom-8" %>
+    <h2 class="govuk-heading-m">What happens next</h2>
+    <% if school_cohort.cip? %>
+      <p class="govuk-body">We’ll email you a link to access your materials within the next 24 hours.</p>
+    <% elsif school_cohort.fip? %>
+      <p class="govuk-body">You will not need to use this service again during your training.</p>
+      <% if school_cohort.delivery_partner %>
+        <p class="govuk-body">
+          <%= school_cohort.school.name %> has chosen <%= school_cohort.delivery_partner.name %> and <% school_cohort.lead_provider.name %> to deliver your training programme.
```
Lloyd-K, line 14: In the review app, user 'fip-mentor@...' only included a delivery partner, not a lead provider. The 'and' still appeared however. Is it possible to code this so the 'and' is removed if only the delivery partner or lead provider name is present?

app/views/participants/validations/nino.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<% content_for :before_content, govuk_back_link(text: "Back", href: back_link_path) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= title %></h1>
+    <%= form_for validation_form, url: { action: :update }, method: :patch do |f| %>
+      <%= f.govuk_text_field :nino,
+        width: "three-quarters",
+        label: { text: "National Insurance Number", class: "govuk-visually-hidden" },
+        hint: { text: "It’s on your National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’."}
```
Lloyd-K, line 12: If I enter the example NINO, I get an error message telling me it's in the wrong format - but it plays back the text as we display it in the example. Just checking that's the correct response please? Thanks

app/views/participants/validations/no_match.html.erb
```diff
@@ -0,0 +1,62 @@
+<% title = validation_form.step_completed?(:no_match) ? "We still cannot find your details" : "We cannot find your details" %>
+<% content_for :title, title %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= title %></h1>
+    <p class="govuk-body">Check the information you entered is correct</p>
```
Lloyd-K, line 7:
```suggestion
    <p class="govuk-body">Check the information you entered is correct.</p>
```

config/locales/participants/validation_form.en.yml
```diff
@@ -0,0 +1,56 @@
...
+              invalid_year: The year must be 4 digits long. For example, 1990
+
+  participants:
+    validation_form:
+      extra_steps:
+        nino:
+          note: Tell us your National Insurance number to help us to find your details.
+          button: Add your National Insurance number
+        name_changed:
+          note: Confirming the full name we have for you may help us find your details
```
Lloyd-K, line 49:
```suggestion
          note: Confirming the full name we have for you may help us find your details.
```

Lloyd-K (changes requested):
2 queries and 2 punctuation suggestions from me. Thanks

lib/tasks/cron.rake
```diff
@@ -5,7 +5,6 @@ namespace :cron do
   task schedule: :environment do
     SessionTrimJob.schedule
     ImportGiasDataJob.schedule
-    ValidationRetryJob.schedule
```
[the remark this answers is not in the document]
BroiSatse: This was quite a bit of a gray region - this job would only be scheduled if the DQT call fails for whatever reason (while still displaying error message to the user). We sort of could do this before as we were collecting whole set of details at once, but with the new journey we might only have a small subset of data and re-running the validation based on that is not sufficient, but also we cannot really continue to gather more information - all we can do is to display the error to the user so they come back later.

app/forms/participants/participant_validation_form.rb
```diff
@@ -2,134 +2,140 @@
...
-
-        if trn !~ /\A\d+\z/
-          errors.add(:trn, :invalid)
-        elsif trn.length < 5
-          errors.add(:trn, :too_short, count: 5)
-        elsif trn.length > 7
-          errors.add(:trn, :too_long, count: 7)
-        end
-      end
+    def check_eligibility!
```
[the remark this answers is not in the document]
BroiSatse: It is now back in

app/views/participants/validations/nino.html.erb
BroiSatse: Yes, the example NINO is not a valid NINO number - which is in accordance with  govuk styleguide for nino: https://design-system.service.gov.uk/patterns/national-insurance-numbers/

app/services/participant_validation_service.rb
```diff
@@ -55,7 +56,9 @@ def full_dqt_client
   end
 
   def matching_record(trn:, nino:, full_name:, dob:)
-    return if trn.blank?
+    return if trn.blank? && nino.blank?
+
+    trn ||= "1"
```
jamesrose, line 61: Can you explain why you need to default this? As far as I can see, you will never set this because you return if trn is blank.

app/services/participant_validation_service.rb
```diff
@@ -55,7 +56,9 @@ def full_dqt_client
   end
 
   def matching_record(trn:, nino:, full_name:, dob:)
-    return if trn.blank?
+    return if trn.blank? && nino.blank?
```
jamesrose, line 59: Are you sure we want to return here if the `nino` is blank? It contradicts a later line that handles it:

```ruby
nino_matches = nino.present? && nino.downcase == dqt_record["ni_number"]&.downcase
matches += 1 if nino_matches
```

app/views/participants/validations/manual_check.html.erb
```diff
@@ -0,0 +1,28 @@
...
+<% content_for :title, title %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_panel(title_text: title) %>
+
+    <h2 class="govuk-heading-m">What happens next</h2>
+    <p class="govuk-body">We may need to contact you for more information to complete your registration.</p>
+
+    <%# TODO: THIS SEEMS TO BE IDENTICAL TO ELIGIBLE COPY - confirm and dry! %>
```
jamesrose, line 11: Do you want to address this TODO?

db/schema.rb
```diff
@@ -691,6 +691,17 @@
     t.string "type", default: "Finance::Schedule::ECF"
   end
 
+  create_table "school_access_tokens", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
jamesrose, line 694: What is this? It doesn't seem related to the PR.

app/services/validate_participant.rb
```diff
@@ -1,104 +0,0 @@
-# frozen_string_literal: true
```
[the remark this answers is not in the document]
jamesrose: I agree.

What was the issue with always creating a validation record?

lib/tasks/cron.rake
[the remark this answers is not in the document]
jamesrose: Why isn't re-running the validation based on the data we _do_ have sufficient? What _does_ happen in this PR if the DQT fails?

app/services/participant_validation_service.rb
jamesrose: Sorry, @ethanmills pointed out this is an `&&` not `||`

db/schema.rb
BroiSatse: Sry, migration from the other branch, removing now

lib/tasks/cron.rake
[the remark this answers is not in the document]
BroiSatse: If DQT fails, the user sees 500 page and returns later. Previously user also would see a 500 page, return later to see his validation to be executed. With the previous journey, we asked the users for all the information at once, now we are executing the validation multiple time having more and more data - what shall we do when the validation fails and we need more data? User is very likely to come back later anyway, and given how edge case this scenario is I'd say it is not worth to reattempt validation in a job.

spec/features/participants/validation_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    sign_in_as participant_profile.user
+    expect(page).to have_current_path participants_validation_step_path(:trn)
+    expect(page).to be_accessible
+    page.percy_snapshot("Participant validation journey: trn")
+
+    click_on "Continue"
+    expect(page).to have_content "Enter your teacher reference number"
+
+    fill_in "Teacher reference number (TRN)", with: "1234567"
+    click_on "Continue"
```
tonyheadford, line 21: Why isn't this in the style of feature tests that we've adopted?

lib/tasks/cron.rake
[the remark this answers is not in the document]
tonyheadford: Surely if we are validating their data we have enough info to do that?  So if the DQT request errors we should queue the validation retry job?

app/services/validate_participant.rb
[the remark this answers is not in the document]
tonyheadford: I would really like to keep this in, it's been a huge help dealing with support queries on the console.  At the very least rework it so it calls your new stuff but the single service interface remains the same.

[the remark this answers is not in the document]
BroiSatse: This has now been added: `ParticipantValidationForm.call(profile, save_validation_data_without_match: false)`

lib/tasks/cron.rake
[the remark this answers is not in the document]
BroiSatse: We might have enough info, but we might not. In the new journey we try to valdiate them with miniaml amount of information possible first, only asking for more details if that fails. Retrying validation behind their back without giving them any feedback on the result is not great, it is better for them to come back to the page and fill their details again.

app/views/participants/validations/eligible.html.erb
BroiSatse: That was a typo on my side, now fixed.

app/views/participants/validations/manual_check.html.erb
BroiSatse: Addressed

Lloyd-K (approved):

app/services/validate_participant.rb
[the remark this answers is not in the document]
tonyheadford: Hmmm but now the naming isn't an action?  I really dislike using "form" objects for anything other than gathering/validating form data.  The naming/purpose then doesn't make sense in other contexts.  It's much nicer to have a form object gather data and call service/command objects to do the work and keep the responsibilities simpler.

ethanmills (approved):
This PR is too big for me to have confidence that it won't break anything. That said, I can't see anything wrong with it.

Has Nicola given this the ok?

PR #1365, opened 2021-10-27
[CPDRP-1033] Move from delayed_job to Sidekiq

FriedSock:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDRP-1033

### Code changes

The bulk of the changes here are removing any `delayed_job` specific code, and just use the `ActiveJob` interface. This allows us to just change the `ActiveJob` adapter, then new jobs created will be created by `sidekiq` and stored in redis. Also replace the cron job implementation with `sidekiq-cron` which will handle the rescheduling for us.

#### Redis
Adds a redis instance that `sidekiq` relies on. All app start commands have been changed to additionally run `sidekiq` as well as `delayed_job`.

The "tiny-ha" redis instance isn't actually that tiny for our purposes: https://docs.cloud.service.gov.uk/deploying_services/redis/#set-up-a-redis-service

If we generously estimate that each job takes 1KB of data to store in redis, 1.5GB of space will let us queue 1.5 million jobs.

#### Sidekiq workers
I originally tried running sidekiq on the existing worker boxes but it didn't work, detailed explanation here: https://github.com/DFE-Digital/early-careers-framework/pull/1365/commits/d4ae41d93da1c0a1d421840fa5f25628e276a97b

### Migration Plan

- Deploy these changes. From this point new jobs will be created in redis and worked off by `sidekiq`
- Manually delete any Cron jobs from `delayed_job`. Otherwise they will continue to reschedule themselves and we will have them running on `sidekiq` and `delayed_job`.
- Monitor the size of the `delayed_job` queue, it may take a week or two for all the jobs to get worked off or fail
- Make a follow up deploy to remove the workers from terraform
- Drop the delayed_jobs table and uninstall the gem etc

### Rollback Plan
Change the queue adapter back to `:delayed_job` https://github.com/DFE-Digital/early-careers-framework/pull/1365/files#diff-c1fd91cb1911a0512578b99f657554526f3e1421decdb9e908712beab57e10f9R54

Then depending on the reason for rolling back, we may need to manually run some jobs (or recreate them in delayed job) to recover.

### Semantic changes

With delayed job, we had the ability to make a database change and create an corresponding job atomically, as they were both stored in Postgres. With sidekiq it is possible for a transaction to commit and then the job fail to be created afterwards. For some jobs we will want to create a job inside a transaction and in others we wouldn't want that to block the app functionality. Below is an analysis of all the jobs and how their behaviour will change when migrated to sidekiq.

#### Analytics / notifications / misc jobs

Jobs that have potential problems with atomicity here I would argue we are fine to migrate with and add some tech debt tickets to strengthen the behaviour of anything marked as "Questionable". The application will continue working just fine if some of these jobs are dropped, and we can manually run them to recover in the unlikely scenario of redis falling over, which we will see in sentry.

| Job | Verdict | Notes |
| --- | -------  | ------- |
| StreamBigQueryParticipantDeclarationsJob | OK | runs on a cron schedule |
| NPQ::StreamBigQueryEnrollmentJob | Questionable | runs after save on npq applications being changed. Previously ran in a transaction. We could change this to work on a cron schedule like StreamBigQueryParticipantDeclarationsJob. Otherwise we'd need to backfill jobs that tried to run if redis falls over. | 
| Analytics::UpsertParticipantProfileJob | Questionable | new job replacing implicit use of `delay` previously. In many places this was done outside a transaction but was done inside in one place. Again this could be changed to work like StreamBigQueryParticipantDeclarationsJob |
| ParticipantDetailsReminderJob | OK | always created in a transaction and schedules emails for later |
| Analytics::RecordValidationJob | OK | new job replacing implicit use of `delay` previously. Only used in one place that was outside of a transaction |
| PartnershipNotificationJob | OK | done in a transaction and schedules for later |
| PartnershipReminderJob | OK | done in a transaction and schedules for later |
| ImportGiasDataJob | OK | done on a cron schedule |
| SchoolAnalyticsJob | OK | done on a cron schedule |
| SessionTrimJob | OK | done on a cron schedule |
| ApiRequestJob | OK | previously ran outside of a transaction. Doesn't matter that it is to be run immediately as there are no db records in the args |

#### Jobs that change the database

If these jobs are not recorded application functionality breaks, so we want to fail if we can't create the job.

| Job | Verdict | Notes |
| --- | -------  | ------- |
| PartnershipActivationJob | OK | Is created in a database transaction, and creates the job to be executed far in the future |

#### Emails

Mixture of these are delayed inside/outside database transactions it looks like all "one time" emails eg `account_created_email` are inside transactions.

#### Don't matter

These jobs only run offline in non-prod environments.

SchoolDataImporterJob, CreateNewFakeSandboxDataJob

### Code quality checks

### HTML Checks

### Gotchas

3. 

## External API changes


terraform/workspace-variables/production.tfvars
```diff
@@ -12,7 +12,8 @@ paas_web_app_deployment_strategy = "blue-green-v2"
 paas_web_app_instances = 4
 paas_web_app_memory = 8192
 paas_worker_app_instances = 1
-paas_worker_app_start_command = "/app/bin/delayed_job --pool=mailers --pool=*:2 start && bundle exec rake jobs:work"
+paas_worker_app_start_command = "/app/bin/delayed_job --pool=mailers --pool=*:2 start && (bundle exec rake jobs:work &) && bundle exec sidekiq -C config/sidekiq.yml"
+paas_redis_service_plan = "tiny-ha-5_x"
```
jamesrose, line 16:
```suggestion
paas_redis_service_plan = "tiny-ha-6_x"
```

or even

```suggestion
paas_redis_service_plan = "small-ha-6_x"
```

What do you think? If you accept this, please use 6_x for all environments.

config/initializers/sentry.rb
```diff
@@ -31,6 +31,8 @@
       end
     when /delayed_job/
       0.001
```
jamesrose, line 33:
```suggestion
      end
```

app/services/invite_schools.rb
```diff
@@ -34,200 +34,13 @@ def reached_limit(school)
...
-  end
-
   def send_ministerial_letters
     School.eligible.each do |school|
       recipient = school.contact_email
       delay(queue: "mailers", priority: 1).send_ministerial_letter(recipient) if recipient.present?
     end
   end
 
-  def invite_mats(school_urns)
```
jamesrose, line 64: 🙏

app/services/invite_schools.rb
```diff
@@ -34,200 +34,13 @@ def reached_limit(school)
     end
   end
 
-  def send_chasers
```
jamesrose, line 37: 🙏

config/initializers/sentry.rb
FriedSock: We still need this whilst we are running delayed job and sidekiq in parallel

ethanmills:
We wipe the dev/review db on every deploy. That removed all the delayed jobs. Do we need something similar for redis?

README.md
```diff
@@ -131,7 +131,7 @@ These are deployed using terraform. See the documentation for [details on terraf
 
 In addition to a web app, some environments have a worker app for running background jobs. For details, see the terraform code.
 
-The `/heathcheck` endpoint on each deployed app will give details on things like version number, commit SHA, delayed jobs, and database migrations.
+The `/heathcheck` endpoint on each deployed app will give details on things like version number, commit SHA, background jobs, and database migrations.
```
ethanmills, line 134: Do we need to add a redis dependency right at the top of this file?

FriedSock:
Closing temporarily so that the review app gets deleted so the redis instance type can be changed

app/jobs/analytics/upsert_participant_profile_job.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Analytics::UpsertParticipantProfileJob < ApplicationJob
```
ethanmills, line 3:
```suggestion
class Analytics::UpsertECFParticipantProfileJob < ApplicationJob
```
Really pedantic, but might be worth being explicit here. What do you think?

app/controllers/participants/validations_controller.rb
```diff
@@ -254,7 +254,7 @@ def store_form_and_redirect_to_step(step)
     end
 
     def store_analytics(matched:)
-      Analytics::ECFValidationService.record_validation(
+      Analytics::RecordValidationJob.perform_later(
```
ethanmills, line 257: Does this work? I didn't think sidekiq jobs could take AR models? https://github.com/mperham/sidekiq/wiki/The-Basics#client

ethanmills: https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple

FriedSock: Yes it does https://github.com/mperham/sidekiq/wiki/Active-Job#using-global-id

We could change all the jobs to find by ID instead of using the active job functionality, but maybe that is more of a tech debt task as it would make this PR even larger

terraform/workspace-variables/review.tfvars
```diff
@@ -10,4 +10,13 @@ paas_app_stopped = false
 paas_web_app_deployment_strategy = "blue-green-v2"
 paas_web_app_instances = 1
 paas_web_app_memory = 1024
-paas_web_app_start_command = "bundle exec rake cf:on_first_instance db:migrate && /app/bin/delayed_job --pool=mailers --pool=priority_mailers --pool=*:2 start && rails db:safe_reset && rails cron:schedule && rails s"
\ No newline at end of file
+paas_web_app_start_command = "bundle exec rake cf:on_first_instance db:migrate db:safe_reset redis:flushall && /app/bin/delayed_job --pool=mailers --pool=priority_mailers --pool=* start && (bundle exec sidekiq -C config/sidekiq.yml &) && rails s"
+paas_redis_service_plan = "micro-6_x"
+
+# Temporarily added for testing purposes, delete before merging
```
jamesrose, line 16: Don't forget to delete these!

jamesrose (approved):
Awesome. Nice work on getting this 🚢 !

app/models/participant_profile/ecf.rb
```diff
@@ -46,6 +48,12 @@ def fundable?
   def policy_class
     ParticipantProfile::ECFPolicy
   end
+
+private
+
+  def update_analytics
+    Analytics::UpsertECFParticipantProfileJob.perform_later(participant_profile: self) if saved_change_to_training_status?
+  end
```
StupidCodeFactory, line 56: Why has this moved to only `ParticipantProfile::ECF` this would stream anymore NPQ participant changes.

lib/tasks/flush_redis.rake
```diff
@@ -0,0 +1,13 @@
...
+namespace :redis do
+  desc "Reset the redis db, used on deploys in the dev and review envs"
+  task flushall: :environment do
+    service_config = JSON.parse(ENV["VCAP_SERVICES"])
+    redis_config = service_config["redis"]
+    redis_worker_config = redis_config.select { |r| r["instance_name"].include?("worker") }.first
+    redis_credentials = redis_worker_config["credentials"]
+    redis = Redis.new(url: redis_credentials["uri"])
+    redis.flushall
+  end
```
StupidCodeFactory, line 12: Is the Redis server shared across review apps?

FriedSock: No, there is a new redis for each app

app/models/participant_profile/ecf.rb
FriedSock: I found that this analytics job was erroring when running for NPQ participants during the tests when the participants were being created. Which makes sense as `ECFValidationService` doesn't sound like it should be running for NPQ participants. I don't have a good explanation as to why we don't see lots of errors in sentry on production for this, maybe because we aren't changing the training status column?

I think it was breaking on this line https://github.com/DFE-Digital/early-careers-framework/blob/CPDRP-1033-sidekiq-migration/app/services/analytics/ecf_validation_service.rb#L47 because NPQ profiles don't have a `ecf_participant_eligibility` relation

app/controllers/healthcheck_controller.rb
```diff
@@ -13,7 +13,7 @@ def check
         migration_version: migration_version,
       },
       delayed_job: {
```
ethanmills, line 15: This should either become sidekiq, or we should add a new section. Probably the latter, while we run them in parallel

config/sidekiq_cron_schedule.yml
```diff
@@ -0,0 +1,24 @@
+session_trim:
+  cron: "0 1 * * *"
+  class: "SessionTrimJob"
+  queue: default
+create_new_fake_sandbox_data:
+  cron: "0 2 * * *"
+  class: "CreateNewFakeSandboxDataJob"
+  queue: default
```
ethanmills, line 8: I know this early returns on prod, but this still looks a little scary. Would it be worth having one schedule per environment do you think?

app/jobs/stream_big_query_participant_declarations_job.rb
```diff
@@ -1,10 +1,6 @@
 # frozen_string_literal: true
 
-class StreamBigQueryParticipantDeclarationsJob < CronJob
-  self.cron_expression = "0 * * * *"
-
-  queue_as :big_query_participant_declarations
-
+class StreamBigQueryParticipantDeclarationsJob < ApplicationJob
```
ethanmills, line 3: I think this might fail in non-production environments?

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "sidekiq/web"
+
+if ENV.key?("VCAP_SERVICES")
```
ethanmills, line 5: How is this configured locally?

ethanmills (approved):

config/initializers/sidekiq.rb
FriedSock: If nothing is configured it defaults to http://127.0.0.1:6379/, which is the port redis defaults to when you run `redis-server` locally as well, might stick that in a comment or the readme.

app/jobs/stream_big_query_participant_declarations_job.rb
FriedSock: From what I remember, bigquery will give you a 200 response saying something like "didn't work" if there is a problem 😄. So it is probably "failing" like this without erroring in all environments right now, but yes it will benefit from some short circuiting based on the env

PR #1372, opened 2021-10-28
Moved start and privacy tests to feat specs

ngw:
Refactor cypress scenario into feature spec

### DIFF
--- a/spec/cypress/integration/Privacy.feature
+++ b/spec/cypress/integration/Privacy.feature
@@ -1,8 +1,0 @@
-Feature: Privacy page
-  Scenario: Visiting the privacy policy page
-    Given scenario "current_privacy_policy" has been run
-    And I am on "start" page
-    When I click on "link" containing "Privacy"
-    Then I should be on "privacy" page
-    And the page should be accessible
-    And percy should be sent snapshot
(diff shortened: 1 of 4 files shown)


ngw:
> Maybe have a look at what I have been doing https://github.com/DFE-Digital/early-careers-framework/pull/1325/files
> 
>     * I think having a separate steps file (with some organisation so methods are easy to find) instead of putting steps under `private` means that they can be re-used
> 
>     * Avoid using 'Ands' when 'Thens' are more appropriate for expectations, I think we should reserve 'Ands' for setup

And should just inherit from the Given/When/Then it's under, and shouldn't have any specific meaning. Configuration should be "Given", actually: https://www.agilealliance.org/glossary/gwt/#q=~(infinite~false~filters~(postType~(~'page~'post~'aa_book~'aa_event_session~'aa_experience_report~'aa_glossary~'aa_research_paper~'aa_video)~tags~(~'given*20when*20then))~searchTerm~'~sort~false~sortDirection~'asc~page~1) (nice url agilealliance...)
I've followed the previous test's style mainly, this methods have very low reusability and the steps I see are IMVHO a bit problematic, as they keep redefining instance variables instead of using dependency injection.
We should probably catch up on this all together

ethanmills:
`Avoid using 'Ands' when 'Thens' are more appropriate for expectations, I think we should reserve 'Ands' for setup`
Don't agree with this one actually. The first statement of a "phase" should be one of given/when/then, then all subsequent lines should be And

ethanmills (approved):

PR #1379, opened 2021-11-01
CPDRP-850 Track major school changes

tonyheadford:
## Ticket and context

Automatically process the more straight-forward major changes to school data received from GIAS.

__Note:__ I've not added this to the `ImportGiasDataJob` as yet as I want to run it manually initially.

This will process `DataStage::SchoolChange` records created during the GIAS data import process.
It will attempt to handle school status changes (opening and closing) and migrate data to the successor school if present.  It does not attempt to handle more complex splits and merges that potentially need more interaction or communication with the schools.
When migrating a school it will add successor and predecessor links similar to the `DataStage::SchoolLink` links imported from GIAS so we can track what has happened. These are not currently surfaced in the UI.

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/app/services/data_stage/process_school_changes.rb
+++ b/app/services/data_stage/process_school_changes.rb
@@ -19,6 +19,8 @@ module DataStage
 
         if change.attribute_changes.key? "school_status_code"
           handle_status_code_change(change)
+        else
+          change.school.create_or_sync_counterpart!
         end
       end
     end
@@ -26,17 +28,19 @@ module DataStage
   private
 
     def has_unhandled_changes?(school_change)
-      unhandled_attrs = (unhandled_attributes & change.attribute_changes.keys)
+      unhandled_attrs = (unhandled_attributes & school_change.attribute_changes.keys)
+      counterpart = school_change.school.counterpart
+
       has_unhandled_changes = false
 
-      if unhandled_attrs.any?
-        counterpart = school_change.school.counterpart
+      if counterpart.present? && unhandled_attrs.any?
         unhandled_attrs.each do |attr|
           next if counterpart.send(attr).blank?
           has_unhandled_changes = true
           break
         end
       end
+
       has_unhandled_changes
     end
 
@@ -75,17 +79,17 @@ module DataStage
         return
       end
 
-      last_link = change.school.school_links.find_by(link_type: "Successor")
-      if last_link.present?
+      successor_link = change.school.school_links.find_by(link_type: "Successor")
+      if successor_link.present?
         ActiveRecord::Base.transaction do
-          successor = find_or_create_successor!(last_link.link_urn)
+          successor = find_or_create_successor!(successor_link.link_urn)
           move_assets_from!(school: change.school.counterpart, successor: successor)
-          change.school.counterpart.school_links.successor.create!(link_school: successor)
-          successor.school_links.predecessor.create!(link_school: change.school.counterpart)
-          change.school.create_or_sync_counterpart!
-          change.update!(handled: true)
+          change.school.counterpart.school_links.successor.simple.create!(link_urn: successor.urn)
+          successor.school_links.predecessor.simple.create!(link_urn: change.school.urn)
         end
       end
+      change.school.create_or_sync_counterpart!
+      change.update!(handled: true)
     end
 
     def find_or_create_successor!(urn)
--- a/app/services/data_stage/process_school_changes.rb
+++ b/app/services/data_stage/process_school_changes.rb
@@ -77,9 +77,9 @@ module DataStage
           move_assets_from!(school: change.school.counterpart, successor: successor)
           change.school.counterpart.school_links.successor.simple.create!(link_urn: successor.urn)
           successor.school_links.predecessor.simple.create!(link_urn: change.school.urn)
+          change.school.create_or_sync_counterpart!
+          change.update!(handled: true)
         end
-        change.school.create_or_sync_counterpart!
-        change.update!(handled: true)
       end
     end
 
--- a/app/services/data_stage/process_school_changes.rb
+++ b/app/services/data_stage/process_school_changes.rb
@@ -97,7 +97,9 @@ module DataStage
       SchoolCohort.where(school: school).each do |school_cohort|
         school_cohort.update!(school: successor)
         school_cohort.ecf_participant_profiles.each do |profile|
-          RectifyParticipantSchool.call(participant_profile: profile, school: successor)
+          RectifyParticipantSchool.call(participant_profile: profile,
+                                        school: successor,
+                                        transfer_pupil_premium_and_sparsity: false)
         end
       end
 
--- a/app/services/data_stage/process_school_changes.rb
+++ b/app/services/data_stage/process_school_changes.rb
@@ -36,6 +36,7 @@ module DataStage
         unhandled_attrs.each do |attr|
           current_value = counterpart.send(attr)
           next if current_value.blank? || current_value == school.send(attr)
+
           has_unhandled_changes = true
           break
         end
--- a/app/services/data_stage/process_school_changes.rb
+++ b/app/services/data_stage/process_school_changes.rb
@@ -1,0 +1,109 @@
+# frozen_string_literal: true
+
+module DataStage
+  class ProcessSchoolChanges < ::BaseService
+    include GiasTypes
+
+    UNHANDLED_ATTRIBUTES = %w[
+      administrative_district_code
+      administrative_district_name
+      school_type_code
+      school_type_name
+      section_41_approved
+      ukprn
+    ].freeze
+
+    def call
+      DataStage::SchoolChange.includes(:school).unhandled.status_changed.find_each do |change|
+        next if has_unhandled_changes?(change)
+
+        if change.attribute_changes.key? "school_status_code"
+          handle_status_code_change(change)
+        end
+      end
+    end
+
+  private
+
+    def has_unhandled_changes?(school_change)
+      unhandled_attrs = (unhandled_attributes & change.attribute_changes.keys)
+      has_unhandled_changes = false
+
+      if unhandled_attrs.any?
+        counterpart = school_change.school.counterpart
+        unhandled_attrs.each do |attr|
+          next if counterpart.send(attr).blank?
+          has_unhandled_changes = true
+          break
+        end
+      end
+      has_unhandled_changes
+    end
+
+    def has_unhandled_attributes?(attributes)
+      (unhandled_attributes & attributes.keys).any?
+    end
+
+    def unhandled_attributes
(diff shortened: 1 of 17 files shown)


ethanmills:
Will take a look at this tomorrow, not enough left in my head today

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,117 @@
...
+    UNHANDLED_ATTRIBUTES = %w[
+      school_type_code
+      school_type_name
+      section_41_approved
+      ukprn
+    ].freeze
+
+    def call
+      DataStage::SchoolChange.includes(:school).unhandled.status_changed.find_each do |change|
+        next if has_unhandled_changes?(change)
```
ethanmills, line 16: Does the word "unhandled" on these two lines mean different things?

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    def move_assets_from!(school:, successor:)
+      raise ActiveRecord::Rollback if successor.school_cohorts.any?
+
+      Partnership.active.where(school: school).each { |partnership| partnership.update!(school: successor) }
+
+      SchoolCohort.where(school: school).each do |school_cohort|
+        school_cohort.update!(school: successor)
+        school_cohort.ecf_participant_profiles.each do |profile|
+          RectifyParticipantSchool.call(participant_profile: profile, school: successor)
```
ethanmills, line 110: Is this the right thing here? This sets the pp and sparsity to the new school. Are we sure we want this?

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,117 @@
...
+        end
+      end
+
+      InductionCoordinatorProfilesSchool.where(school: school).each { |profile| profile.update!(school: successor) }
```
ethanmills, line 114:
```suggestion
      InductionCoordinatorProfilesSchool.where(school: school).each { |link| link.update!(school: successor) }
```
These aren't profiles

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,117 @@
...
+      section_41_approved
+      ukprn
+    ].freeze
+
+    def call
+      DataStage::SchoolChange.includes(:school).unhandled.status_changed.find_each do |change|
+        next if has_unhandled_changes?(change)
+
+        if change.attribute_changes.key? "school_status_code"
+          handle_status_code_change(change)
```
ethanmills, line 19: How do we detect mergers/splits?

app/services/set_school_local_authority.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class SetSchoolLocalAuthority < BaseService
+  def call
+    if school.local_authority&.code != la_code
+      ActiveRecord::Base.transaction do
+        school.latest_school_authority&.update!(end_year: start_year)
+        SchoolLocalAuthority.create!(school: school,
+                                     local_authority: LocalAuthority.find_by(code: la_code),
+                                     start_year: start_year)
```
ethanmills, line 10: I wonder if a year is granular enough, if we're updating this daily instead of yearly. Would be good to check in with the LPDOB folks to see what they need here

app/services/set_school_local_authority_district.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class SetSchoolLocalAuthorityDistrict < BaseService
+  def call
+    if school.local_authority_district&.code != administrative_district_code
+      ActiveRecord::Base.transaction do
+        school.school_local_authority_districts.latest.first&.update!(end_year: start_year)
```
ethanmills, line 7: Pretty inconsistent with LA

ethanmills: We should change the `latest` scope from a `where` to `find_by`. I know I wrote it, but it doesn't make any sense

app/services/set_school_local_authority_district.rb
```diff
@@ -0,0 +1,39 @@
...
+  attr_reader :administrative_district_code, :school, :start_year
+
+  def initialize(school:, administrative_district_code: nil, start_year: Time.zone.now.year)
+    @school = school
+    @administrative_district_code = administrative_district_code || school.administrative_district_code
+    @start_year = start_year
+  end
+
+  def update_participants_sparsity_flag!
+    unless school.partnerships.any?
```
ethanmills, line 28: I would guess this wants an `unchallenged` or maybe an `active`, but I'm not sure which

app/services/data_stage/process_school_changes.rb
tonyheadford: Yes, bit confusing. `SchoolChange` has a handled flag on the record to indicate whether it's been dealt with.  The `has_unhandled_changes?` means it contains changes to attributes that we don't want/know how to deal with yet

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,117 @@
...
+          break
+        end
+      end
+
+      has_unhandled_changes
+    end
+
+    def has_unhandled_attributes?(attributes)
+      (unhandled_attributes & attributes.keys).any?
+    end
```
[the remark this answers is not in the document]
tonyheadford: Good spot, perhaps not now. I changed my mind and decided to allow unhandled attributes if they hadn't been set already in the destination school

app/services/set_school_local_authority_district.rb
ngw: @ethanmills it's not the best idea to use `find_by` in a scope as it fires a db query. Scopes should only return scopes (using `where`, `order` and other methods that actually return scopes).
I might be missing some context here, looks `latest` could be better served by a method that executes the query or by a scope using `order` + `limit`.

ngw (commented):
For these kinds of tasks one of the things I like to do is add a fixture file with anonymized data full of worst case scenarios which I know, try to import/process it and check on the numbers after importing, as some kind of stress/black box test.
This normally helps to test the system based on a specification, because you create the fixture file with data you know will trigger some behavior.
Not saying that you must do it of course, just saying what I would do :)

app/services/data_stage/process_school_changes.rb
ngw: Maybe we should rename one of the 2?

tonyheadford: I'll rename the scope to `unprocessed`

app/services/data_stage/process_school_changes.rb
tonyheadford: In these cases this is effectively the same school, so treating it as moving the participants to the correct school rather than changing schools. I _think_ that's the right thing to do?

app/services/data_stage/process_school_changes.rb
tonyheadford: That is done in the `close_school` if we don't have a simple "Successor" then we bail

app/services/set_school_local_authority.rb
tonyheadford: LPDOB?
This is only likely to happen where we've had revisions to LAs like splitting Northamptonshire, there are a few awaiting to be processed I think but unlikely to be a common occurrence.

app/services/data_stage/process_school_changes.rb
ethanmills: I don't really follow. In a split, wouldn't there be two successors?

ethanmills: And it a merger, each of the schools have a single successor (the same one)

app/services/set_school_local_authority_district.rb
tonyheadford: I'm not sure the only steer was if there's a partnership leave sparsity alone, so thought it safest to just test whether there was or had been any?

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    def unhandled_attributes
```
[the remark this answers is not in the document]
tonyheadford: Nothing now, good spot

app/services/data_stage/process_school_changes.rb
```diff
@@ -0,0 +1,117 @@
...
+      MAJOR_CHANGE_ATTRIBUTES - %w[school_status_code school_status_name]
+    end
+
+    def handle_status_code_change(change)
+      from_code = change.school.school_status_code
+
+      if from_code.in? [1, 3]
```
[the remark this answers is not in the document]
ethanmills: What happens if it changes from 1 to 3? From Open to Open but Proposed to close?

app/services/data_stage/process_school_changes.rb
ethanmills: I have some reservations about this (quite possibly because I've misunderstood it). It seems to me that this will:

- Open the successor school even when we can't handle the change and close the predecessor
- Close the predecessor and mark the change as handled even when we cannot move the assets etc

tonyheadford: The intention is that it will always open the new school, but only close the old one if we can handle the change.

app/services/data_stage/process_school_changes.rb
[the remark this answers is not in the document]
tonyheadford: Ah - we don't actually care but I intended for it to sync and it doesn't so I've refactored that thanks!

app/services/data_stage/process_school_changes.rb
tonyheadford: Mergers and splits as far as I've been able to tell have different school link reasons in GIAS, not simply "Successor" it would be something like "Successor - merged"

tonyheadford: I don't think the change gets marked as handled on a close if we can't perform the change?

app/services/set_school_local_authority.rb
jamesrose: Lead Partner team.

ethanmills: yes, the catchily named Lead Providers, DfE, and Other Bodies team

tonyheadford: I asked the question but have had no reply so I guess yearly is still OK.

app/services/set_school_local_authority_district.rb
tonyheadford: I'm going to remove updating the participants sparsity as it should only be set at the point in time they are added.

ethanmills (approved):
Happy with this, looks great! I asked one question in slack dev channel, but I don't think that will affect this

PR #1384, opened 2021-11-01
CPDRP-1091: Set SITs an email if a ppt email has bounced

ethanmills:
## Ticket and context

Ticket: CPDRP-1091

Sent induction tutors an email for each participant where an email has bounced in the last month.

No tests for this as it's a one off, but I've tested it manually


app/services/utm_service.rb
```diff
@@ -2,42 +2,43 @@
...
-    year2020_nqt_invite_sit_validated: "year2020-nqt-invite-sit-validated",
-    year2020_nqt_invite_sit_no_participants: "year2020-nqt-invite-sit-no-participants",
     partnered_invite_sit_reminder: "partnered-invite-sit-reminder",
+    partnership_notification: "partnership-notification",
     remind_fip_sit_to_complete_steps: "remind-fip-sit-to-complete-steps",
-    add_participants_unpartnered_cip: "add-participants-unpartnered-cip",
+    sign_in: "sign-in",
+    sign_in_reminder: "sign-in-reminder",
+    sit_mentor_validation_info_2709: "sit-mentor-validation-info-2709",
+    sit_participant_email_bounced: "sit-participant-email-bounced",
```
FriedSock, line 32: Do we not need to add this into the `SOURCE` as well as the `CAMPAIGN` hash?

FriedSock (approved):

app/services/utm_service.rb
ethanmills: I don't think so, the GA campaigns now seem to be showing up as intended

PR #1385, opened 2021-11-02
CPDLP-508 Add schema validation which logs results

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-508

- This reintroduces schema validations for one particular set of endpoints, participant declarations
- From conversations I have had there is hesitance to make failing schema validations prevent api calls from being made 
- Therefore the outcome of this is to log the data instead
- From the logging we should be able to determine the ratio of api calls between passing and failing schema validation
- Also if something did fail validation, what that payload looks like and why it failed. this way we can either fix our end or ask api users to change their calls to conform
- The previous version of the schema validation used handcrafted json schemas to validate against. This change uses the swagger open api 3.0.1 spec which is close to being a json schema. My understanding is we may need to use open api 3.1 in order to have a fully valid json schema. But from my testing it seems to be working

## Future changes

- Analysis of logs to determine failure/success rate to leading changes to api and/or informing api users to change
- Change these validations checks into a blocking calls and return errors to client
- Remove handcrafted json schemas and associated code

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

- Changes have been made in this area but there is nothing that directly impacts api users that needs to be documented

### DIFF
--- a/app/controllers/api/v1/participant_declarations_controller.rb
+++ b/app/controllers/api/v1/participant_declarations_controller.rb
@@ -11,6 +11,9 @@ module Api
 
       def create
         params = HashWithIndifferentAccess.new({ cpd_lead_provider: cpd_lead_provider }).merge(permitted_params["attributes"] || {})
+
+        log_schema_validation_results
+
         render json: RecordParticipantDeclaration.call(params)
       end
 
@@ -74,6 +77,19 @@ module Api
         else
           raise
         end
+      end
+
+      def log_schema_validation_results
+        errors = SchemaValidator.call(raw_event: request.raw_post)
+
+        if errors.blank?
+          Rails.logger.info "Passed schema validation"
+        else
+          Rails.logger.info "Failed schema validation for #{request.raw_post}"
+          Rails.logger.info errors
+        end
+      rescue StandardError => e
+        Rails.logger.info "Error on schema validation, #{e}"
       end
     end
   end
--- a/app/services/schema_validator.rb
+++ b/app/services/schema_validator.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require "json_schema/validation"
+
+class SchemaValidator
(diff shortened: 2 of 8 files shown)


app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -75,6 +78,19 @@ def permitted_params
...
+      def log_schema_validation_results
+        errors = SchemaValidator.call(raw_event: request.raw_post)
+
+        if errors.blank?
+          Rails.logger.info "Passed schema validation"
+        else
+          Rails.logger.info "Failed schema validation for #{request.raw_post}"
+          Rails.logger.info errors
+        end
+      rescue StandardError => e
```
asmega, line 91: I'm not hugely confident of this code so have added extra catch to prevent it from blocking calls

swagger/v1/component_schemas/ECFParticipantRetainedDeclaration.yml
```diff
@@ -1,5 +1,6 @@
 description: "An ECF retained participant declaration"
 type: object
+additionalProperties: false
```
asmega, line 3: i tried setting these to `additional_properties` but does not output `additionalProperties` which is the correct openapi key

app/services/schema_validator.rb
```diff
@@ -0,0 +1,40 @@
...
+    @raw_event = raw_event
+  end
+
+  def schema
+    # TODO: need to pull in correct schema depending on request
+    hash = YAML.safe_load(File.read(Rails.root.join("swagger/v1/component_schemas/ParticipantDeclaration.yml")))
+
+    hash["components"] = {}
+    hash["components"]["schemas"] = {}
+    hash["components"]["schemas"] = CACHED_SCHEMAS
```
asmega, line 32: i spent some brief time sourcing diving the `json-schema` library. I am sure there is a better way to implement but for the moment this works.

jamesrose (approved):
Nice one :)

PR #1397, opened 2021-11-04
CPDLP-267 Add payable view to payment breakdown

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?assignee=5e4e50b03df51b0c9374c1e4&selectedIssue=CPDLP-267

- Added LHS nav when view in payment breakdown
- On top of the current view there is now a payable view for those items in `payable` state

## Screenshots

Screenshot 2021-11-04 at 12 48 39

### Code quality checks

### HTML Checks

### Gotchas

- Visit https://ecf-review-pr-1397.london.cloudapps.digital
- Login as `finance@example.com`
- View ECF payment breakdowns
- Should see current and payable views
- Unfortunately insufficient data has been setup to differentiate the data between the views

## External API changes

- There are no external API changes


app/controllers/finance/ecf/payment_breakdowns_controller.rb
```diff
@@ -11,14 +11,64 @@ def show
           contract: @ecf_lead_provider.call_off_contract,
           event_type: :started,
         )
-        @cutoff_date = "On #{helpers.cutoff_date}"
```
asmega, line 14: I've removed as many helpers as I reasonably can, there are probably more for another time. I don't think putting view helpers in controllers in a great idea.

app/controllers/finance/ecf/payment_breakdowns_controller.rb
```diff
@@ -11,14 +11,64 @@ def show
...
+        render :show
       end
 
     private
 
       def lead_provider_scope
         policy_scope(LeadProvider, policy_scope_class: FinanceProfilePolicy::Scope)
       end
+
+      def payable_milestone
```
asmega, line 41: so there does not seem to be a concept of "bucketing declaration to an invoice" or at least yet. So as far as I can tell when something is paid we simply know it is in the paid state but not which invoice it belongs to.

This is why these milestones and the dates are separate (for) now and hardcoded. till these concepts are worked in and we can query these in a correct manner

app/helpers/finance_helper.rb
```diff
@@ -1,20 +1,6 @@
...
-    "31.01.2022",
-    "30.04.2022",
-    "30.09.2022",
-    "31.01.2023",
-    "30.04.2023",
-    "31.10.2023",
-    "31.01.2024",
-    "30.04.2024",
-    "30.09.2024",
-  ].freeze
```
asmega, line 16: I'm not sure why we have these hardcoded milestone dates so have replaced them with the data from the database where previously used

app/helpers/finance_helper.rb
```diff
@@ -24,27 +10,14 @@ def float_to_percentage(number)
   end
 
   def total_payment(breakdown)
-    service_fee = breakdown[:service_fees].map { |params| params[:monthly] }.inject(&:+)
-    output_payment = breakdown[:output_payments].map { |params| params[:subtotal] }.inject(&:+)
-    other_fees = breakdown[:other_fees].values.map { |other_fee| other_fee[:subtotal] }.inject(&:+)
```
asmega, line 29: I'm not a fan of this method either and think it should be removed in a future change. I think we want to be asking the breakdown (or other object) for the total payment rather than reach in summing all these values

app/controllers/finance/ecf/payment_breakdowns_controller.rb
jamesrose: I don't see invoices as being within the current scope of the service right now. In some ways we don't need to know. We'd just expect to just be told when something has been paid by finance.

spec/features/finance/payment_breakdown_spec.rb
```diff
@@ -22,15 +23,29 @@
...
     when_i_click_on_view_contract_link
     then_the_page_should_be_accessible
     and_percy_should_be_sent_a_snapshot_named("Contract breakdown for an ECF provider")
   end
 
+  scenario "Can get to ECF payment breakdown page for payable declarations" do
+    given_i_am_logged_in_as_a_finance_user
+    and_there_is_ecf_provider_with_contract
+    and_there_is_a_schedule
+    visit "/finance/ecf/payment_breakdowns/cffd2237-c368-4044-8451-68e4a4f73369/payable"
```
jamesrose, line 37:
```suggestion
    visit "/finance/ecf/payment_breakdowns/cffd2237-c368-4044-8451-68e4a4f73369/payable"
```

Can you share this ID across the spec?

jamesrose (approved):
🚢

spec/features/finance/payment_breakdown_spec.rb
asmega: I've rolled this test into the previous one to remove the overhead of performing the setup again. also no need to do the `visit`

PR #1413, opened 2021-11-09
Enforce TRN requirement on participant validation service

jamesrose:
## Ticket and context

- Don't raise an [ambiguous](https://sentry.io/organizations/dfe-bat/issues/2699607064/?environment=production&project=5748989&query=is%3Aunresolved) exception if the TRN provided is nil.
- Remove duplicated padding method from DQT client. We already do this further up the chain.

If the review thinks it would be useful I'd be happy to report log when we receive nil TRNs. I don't think a nil TRN warrants an exception in this instance because it's expected.

### DIFF
--- a/app/services/participant_validation_service.rb
+++ b/app/services/participant_validation_service.rb
@@ -55,6 +55,8 @@ private
   end
 
   def matching_record(trn:, nino:, full_name:, dob:)
+    return if trn.blank?
+
     padded_trn = trn.rjust(7, "0")
     dqt_record = dqt_record(padded_trn, nino)
     return if dqt_record.nil?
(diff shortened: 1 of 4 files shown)


asmega:
NPQ isn't padding the TRN so you may want to make that change first before doing this one. Otherwise users that enter a shorter TRN can't possibly get auto-validated. alternatively we could pad on the incoming api endpoint as another option

jamesrose:
@asmega `ParticipantValidationService` is the entry point for NPQ validation as well. That's the service that we use in the `ParticipantValidationController`. 

Previous to this change we were padding in the service object _and_ the DQT client unnecessarily.

asmega:
> @asmega `ParticipantValidationService` is the entry point for NPQ validation as well. That's the service that we use in the `ParticipantValidationController`.
> 
> Previous to this change we were padding in the service object _and_ the DQT client unnecessarily.

Ok, I see it now :)

asmega (approved):

ethanmills (approved):

PR #1414, opened 2021-11-09
Enqueue sign in emails to be delivered

jamesrose:
## Ticket and context

We can deliver these out of out-of-band. This will automatically give us [redelivery](https://sentry.io/organizations/dfe-bat/issues/2756441853/?environment=production&project=5748989&query=is%3Aunresolved) were Notify be down.

### DIFF
--- a/app/services/invite_schools.rb
+++ b/app/services/invite_schools.rb
@@ -45,7 +45,7 @@ class InviteSchools
                  .uniq
 
       emails.each do |email|
-        delay(queue: "mailers", priority: 1).create_and_send_nomination_email(email, school)
+        delay(queue: "mailers").create_and_send_nomination_email(email, school)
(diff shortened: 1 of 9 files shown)


ethanmills (approved):
We need to be really careful to schedule all other emails with a lower priority

ethanmills:
It might be good to have a lower default priority on the mailer queue, then set this higher e39504b0-9543-438a-aca3-b441c3b768bf

tonyheadford:
> It might be good to have a lower default priority on the mailer queue, then set this higher https://github.com/collectiveidea/delayed_job#named-queues

Could we perhaps have a `priority_mailer` queue and lower the default on the mailer queue?

ethanmills (approved):
Looks good 🙂

PR #1430, opened 2021-11-11
Changes to admins creating NQT+1s

ethanmills:
Allow them to be creating using emails already in use
 - Not allowed for active ECTs
 - Allowed for other user types
 - Name should not be changed for active participants

### DIFF
--- a/app/controllers/admin/schools/cohort_2020_controller.rb
+++ b/app/controllers/admin/schools/cohort_2020_controller.rb
@@ -19,8 +19,15 @@ module Admin
       end
 
       def create
-        @user = User.new(params.require(:user).permit(:full_name, :email))
+        @user = User.find_or_initialize_by(params.require(:user).permit(:email))
+        @user.full_name = params.dig(:user, :full_name)
+
         render :new and return unless @user.valid?
+
+        if @user.early_career_teacher?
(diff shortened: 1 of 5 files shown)


app/controllers/admin/schools/cohort_2020_controller.rb
```diff
@@ -19,9 +19,16 @@ def new
...
 
       def create
-        @user = User.new(params.require(:user).permit(:full_name, :email))
+        @user = User.find_or_initialize_by(params.require(:user).permit(:email))
+        @user.full_name = params.dig(:user, :full_name)
+
         render :new and return unless @user.valid?
 
+        if @user.early_career_teacher?
+          @user.errors.add(:base, "A user with this email address is currently participating as an ECT at school with urn #{@user.teacher_profile.current_ecf_profile.school.urn}")
```
tonyheadford, line 28: Should error messages be in the i18n file now?

ethanmills: Yes, probably. What does that give us again?

tonyheadford: All the error messages in one place primarily, should a content person ever need to review again, @ericaporter recently did some refactoring to move them all together. I know this is an admin page though.

tonyheadford (approved):

PR #1433, opened 2021-11-12
add rails console command to history

asmega:
## Ticket and context

Ticket: none

- I got bored of copying and pasting the various commands to fire up a rails console
- I've now added the needed command to the history 

### Code quality checks

### HTML Checks

### Gotchas

- NOTE: I've actually tested this so unfortunately due to my testing it will have affected the history
```sh
cf target -s earlycareers-framework-dev
cf ssh ecf-review-pr-1433
````
- hit up arrow key
- hit enter
- should now be in a rails console

## External API changes

- there are no external api changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -94,6 +94,8 @@ COPY --from=assets-precompile /app /app
 COPY --from=assets-precompile /usr/local/bundle/ /usr/local/bundle/
 COPY --from=middleman /public/ /app/public/
 
+RUN echo "cd /app && /usr/local/bundle/bin/bundle exec rails c" > /root/.ash_history
+
 WORKDIR /app
 
 # Use this for development testing

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -94,6 +94,8 @@ COPY --from=assets-precompile /app /app
 COPY --from=assets-precompile /usr/local/bundle/ /usr/local/bundle/
 COPY --from=middleman /public/ /app/public/
 
+RUN echo "cd /app && /usr/local/bundle/bin/bundle exec rails c" > /root/.ash_history
```
StupidCodeFactory, line 97:
```suggestion
RUN echo "cd /app && /usr/local/bundle/bin/bundle exec rails c" > /root/.bash_history
```

ethanmills:
Would it be better to set an alias for this instead?

Dockerfile
StupidCodeFactory, line 97: also to remove the painful /usr/local/bin
```suggestion
ENV BUNDLE_PATH "/usr/local/bundle"
ENV PATH "$PATH:/usr/local/bin"
```

StupidCodeFactory: I think this is a better fix. echo'ing the command in the history, is probably helpful but does not address the root of the problem which is that `bundle` in not in the `PATH`

StupidCodeFactory (approved):

Dockerfile
asmega: i'm pretty sure alpine uses `.ash_history`, it worked when i tested it

asmega:
> Would it be better to set an alias for this instead?

i'm not fussed theres multiple ways of fixing this:
- alias as you suggested
- adding to path to the profile so it's available when prompt is loaded as @StupidCodeFactory suggested

Dockerfile
StupidCodeFactory: faire enough, I thought this was a typo

jamesrose:
This could save history of PII, passwords, tokens, etc. Are we sure we want this?

jamesrose (approved):

PR #1445, opened 2021-11-16
BUGFIX: ensure withdrawn profiles are always shown as such in API

ethanmills:
Ensure that when a participant has an active profile not associated with the calling provider, the API response displays the correct withdrawn profile

### DIFF
--- a/app/controllers/api/v1/ecf_participants_controller.rb
+++ b/app/controllers/api/v1/ecf_participants_controller.rb
@@ -14,11 +14,11 @@ module Api
       def index
         respond_to do |format|
           format.json do
-            participant_hash = ParticipantSerializer.new(paginate(participants)).serializable_hash
+            participant_hash = ParticipantSerializer.new(paginate(participants), { params: { mentor_ids: mentor_ids } }).serializable_hash
             render json: participant_hash.to_json
           end
           format.csv do
-            participant_hash = ParticipantSerializer.new(participants).serializable_hash
+            participant_hash = ParticipantSerializer.new(participants, { params: { mentor_ids: mentor_ids } }).serializable_hash
             render body: to_csv(participant_hash)
           end
         end
@@ -28,7 +28,7 @@ module Api
 
       def serialized_response(profile)
         ParticipantSerializer
-          .new(profile.user)
+          .new(profile)
           .serializable_hash.to_json
       end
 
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/ecf_participants_controller.rb
```diff
@@ -41,19 +41,41 @@ def lead_provider
...
+                                                        :school,
+                                                        :ecf_participant_eligibility,
+                                                        :ecf_participant_validation_data,
+                                                        :schedule,
+                                                        teacher_profile: { current_ect_profile: { mentor_profile: :user } },
+                                                      )
+
+        if updated_since.present?
+          participant_profiles = participant_profiles.where(user: { updated_at: updated_since.. })
+                                                     .order("\"user\".updated_at, \"user\".id")
```
jamesrose, line 58:
```suggestion
                                                     .order("user.updated_at, user.id")
```

Or did you mean `order("users.updated_at, users.id")` ?

app/controllers/api/v1/ecf_participants_controller.rb
```diff
@@ -41,19 +41,41 @@ def lead_provider
...
+        participant_profiles.order("\"user\".created_at")
+      end
+
+      def participant_profile_ids
+        @participant_profile_ids ||= lead_provider.ecf_participant_profiles
+                                                  .select("DISTINCT ON (participant_profiles.teacher_profile_id) teacher_profile_id, participant_profiles.status, participant_profiles.id")
+                                                  .joins(:school_cohort)
+                                                  .where(school_cohort: { cohort_id: Cohort.current.id })
+                                                  .order(:teacher_profile_id, status: :asc)
+                                                  .map(&:id)
```
jamesrose, line 70:
```suggestion
                                                  .pluck(:id)
```

ethanmills: This breaks with the distinct on I think, because pluck reSelects

app/controllers/api/v1/ecf_participants_controller.rb
ethanmills: No, AR is a pain:
`INNER JOIN "users" "user" ON "user"."id" = "teacher_profiles"."user_id"`

app/controllers/api/v1/ecf_participants_controller.rb
```diff
@@ -41,19 +41,41 @@ def lead_provider
...
+                                                        :cohort,
+                                                        :school,
+                                                        :ecf_participant_eligibility,
+                                                        :ecf_participant_validation_data,
+                                                        :schedule,
+                                                        teacher_profile: { current_ect_profile: { mentor_profile: :user } },
+                                                      )
+
+        if updated_since.present?
+          participant_profiles = participant_profiles.where(user: { updated_at: updated_since.. })
```
StupidCodeFactory, line 57: What's this `updated_since..` syntax? Never seen this? 
is this a shorthand for something in which case could we use a more idiomatic syntax?

spec/requests/api/v1/ecf_participants_spec.rb
```diff
@@ -188,6 +188,24 @@
...
+          end
+
+          it "does not include personal information of the participant" do
+            get "/api/v1/participants/ecf"
+            matching_records = parsed_response["data"].select { |record| record["id"] == active_profile_with_other_provider.user.id }
+            expect(matching_records.first["attributes"]["full_name"]).to be_nil
+            expect(matching_records.first["attributes"]["email"]).to be_nil
+            expect(matching_records.first["attributes"]["teacher_reference_number"]).to be_nil
+          end
+        end
```
ethanmills, line 208: I've checked that these tests fail on develop

app/controllers/api/v1/ecf_participants_controller.rb
StupidCodeFactory: also I believe the order didn't work because this probably should be:

```ruby
.where(users: { updated_at: updated_since.. })
.order("users.updated_at")
`

StupidCodeFactory: no need to order by `users.id` since those `id` are `UUID` and the order of those is pretty random

jamesrose: It's an endless range — it means updated from that time. 

Re: ordering. That's how it works at the moment so let's keep it consistent for now.

jamesrose (approved):

PR #1447, opened 2021-11-17
Updates to analytics queries

ngw:
- On the school table, is it possible to have a flag where an induction tutor has also been added as a mentor? Is it possible to also include the participant ID of that linked mentor record?
- On the participant table, TRN was missing as a field in the dummy data.

### DIFF
--- a/analytics_queries/participants.sql
+++ b/analytics_queries/participants.sql
@@ -12,8 +12,9 @@ SELECT u.id                                                    as participant_id
        c.start_year                                            as cohort,
        pp.training_status                                      as training_status,
        s.name                                                  as schedule,
+       tp.trn                                                  as trn,
        epvd.created_at                                         as trn_provided_at,
-       (epe.status IN ('eligible', 'matched'))                 AS trn_validated,
+       (epe.status IN ('eligible', 'matched'))                 as trn_validated,
        (epe.manually_validated OR epe.status = 'manual_check') as manual_validation_required,
        (CASE
             WHEN epe.status = 'eligible' THEN true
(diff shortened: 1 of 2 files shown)


analytics_queries/participants.sql
```diff
@@ -12,6 +12,7 @@ SELECT u.id                                                    as participant_id
        c.start_year                                            as cohort,
        pp.training_status                                      as training_status,
        s.name                                                  as schedule,
+       epvd.trn                                                as trn,
```
ethanmills, line 15: We should use the TRN on the teacher profile instead of this one. This one won't be retained forever, and also if they made a mistake in entering it that we could automatically correct with the other data then the teacher profile will be correct and this will be as-entered

analytics_queries/schools.sql
```diff
@@ -7,14 +7,19 @@ SELECT DISTINCT s.urn,
...
                 (u.current_sign_in_at IS NOT NULL) as tutor_signed_in,
                 u.full_name                        as tutor_name,
-                u.email                            as tutor_email
+                u.email                            as tutor_email,
+                (pp.id IS NOT NULL)                as sit_mentor,
+
 FROM schools s
          LEFT OUTER JOIN school_cohorts sc on s.id = sc.school_id
-         LEFT OUTER JOIN cohorts c on sc.cohort_id = c.id
+         LEFT OUTER JOIN cohorts c on sc.cohort_id = c.id 1
```
ethanmills, line 15: What is this for?

ngw: I'd say me trying to switch buffer in vi. Removed.

jamesrose (approved):

ethanmills (approved):

PR #1450, opened 2021-11-17
CPDRP-1137: Prevent record withdrawal of participants with declarations

ethanmills:
## Ticket and context

Ticket: CPDRP-1137

## Tech review
Stronger policy driven enforcement of when participant profiles can be withdrawn by admins and SITs.
The criteria here are a little complex, which is why I went with a policy


tonyheadford (approved):
Nice

PR #1454, opened 2021-11-18
CPDLP-672 Prevent voiding certain declarations

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-672

- At API level prevent voided, payable and paid declarations from being voidable
- I have followed the crowd for the way the error is raised. However i feel raising exceptions for validation messages is not the way forward

### Code quality checks

### HTML Checks

### Gotchas

- still somewhat difficult given current state of dummy data

## External API changes

- Minor impact on API. I don't think it needs documenting at this moment in time

### DIFF
--- a/app/models/participant_declaration.rb
+++ b/app/models/participant_declaration.rb
@@ -64,12 +64,17 @@ class ParticipantDeclaration < ApplicationRecord
   scope :paid_npqs_for_lead_provider, ->(lead_provider) { paid_for_lead_provider(lead_provider).npq }
   scope :paid_uplift_for_lead_provider, ->(lead_provider) { paid_for_lead_provider(lead_provider).uplift }
 
+  # TODO: Voiding paid should trigger clawbacks, but currently OOS
+  def voidable?
+    !voided? && !payable? && !paid?
+  end
+
   def make_submitted!
     DeclarationState.submitted!(self) if eligible?
   end
 
   def make_voided!
-    DeclarationState.voided!(self) unless voided? || payable? || paid? # TODO: Voiding paid should trigger clawbacks, but currently OOS
+    DeclarationState.voided!(self) if voidable?
   end
 
   def make_eligible!
--- a/app/services/void_participant_declaration.rb
+++ b/app/services/void_participant_declaration.rb
@@ -3,22 +3,30 @@
 class VoidParticipantDeclaration
   attr_accessor :cpd_lead_provider, :id
 
+  def initialize(cpd_lead_provider:, id:)
+    @cpd_lead_provider = cpd_lead_provider
+    @id = id
+  end
+
   def call
-    declaration = ParticipantDeclaration.for_lead_provider(cpd_lead_provider).find(id)
-
+    # TODO: these should be moved to activemodel validations
+    # and not controller level exceptions
     raise Api::Errors::InvalidTransitionError, I18n.t(:declaration_already_voided) if declaration.voided?
-
-    latest_declaration = declaration.participant_profile.participant_declarations.order(declaration_date: :desc).first
     raise Api::Errors::InvalidTransitionError, I18n.t(:void_last_declaration_only) if latest_declaration != declaration
+    raise Api::Errors::InvalidTransitionError, I18n.t(:declaration_not_voidable) unless declaration.voidable?
 
-    declaration.voided!
+    declaration.make_voided!
+
     ParticipantDeclarationSerializer.new(declaration).serializable_hash.to_json
   end
 
 private
 
-  def initialize(cpd_lead_provider:, id:)
-    @cpd_lead_provider = cpd_lead_provider
-    @id = id
+  def declaration
+    @declaration ||= ParticipantDeclaration.for_lead_provider(cpd_lead_provider).find(id)
+  end
+
+  def latest_declaration
+    @latest_declaration ||= declaration.participant_profile.participant_declarations.order(declaration_date: :desc).first
   end
 end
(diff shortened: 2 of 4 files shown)


app/models/participant_declaration.rb
```diff
@@ -64,12 +64,17 @@ class ParticipantDeclaration < ApplicationRecord
   scope :paid_npqs_for_lead_provider, ->(lead_provider) { paid_for_lead_provider(lead_provider).npq }
   scope :paid_uplift_for_lead_provider, ->(lead_provider) { paid_for_lead_provider(lead_provider).uplift }
 
+  # TODO: Voiding paid should trigger clawbacks, but currently OOS
+  def voidable?
+    !voided? && !payable? && !paid?
+  end
```
ericaporter, line 70: We definitely want &&'s here, not ||'s?

app/services/void_participant_declaration.rb
```diff
@@ -3,22 +3,30 @@
...
-  def call
-    declaration = ParticipantDeclaration.for_lead_provider(cpd_lead_provider).find(id)
+  def initialize(cpd_lead_provider:, id:)
+    @cpd_lead_provider = cpd_lead_provider
+    @id = id
+  end
 
+  def call
+    # TODO: these should be moved to activemodel validations
+    # and not controller level exceptions
```
StupidCodeFactory, line 13: This does not seem to be a massive undertaking. Perhaps this could be done as part of the PR, as it is touching the validation.
However maybe the validations are fine here, this is not a controller, it is a service class and the validations here are tied to the process of voiding a declaration, there is an argument to be made for the validation to remain here I think.
if we'd move this validation to `active record` (I'm assuming here that's what you meant by `active model` we'd be coupling our database layer to the application logic, which is sometime more of pain than a helper.

StupidCodeFactory: however I'm with you 💯 raising an error is a not ideal a simple error message added to the service class would be a bit cleaner

config/locales/en.yml
```diff
@@ -50,6 +50,7 @@ en:
   withdrawn_participant: Cannot perform actions on a withdrawn participant
   declaration_on_incorrect_state: "The declaration on withdrawn or deferred participant can not be accepted"
   declaration_already_voided: "Declaration is already voided"
+  declaration_not_voidable: "Declaration not voidable"
```
StupidCodeFactory, line 53: not entirely sure but I think the format of the error message should be something the follows that convention:

`The property '#/state' is not voidable`

?

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -7,10 +7,10 @@
   include_context "lead provider profiles and courses"
   let(:parsed_response) { JSON.parse(response.body) }
 
-  describe "post" do
-    let(:token) { LeadProviderApiToken.create_with_random_token!(cpd_lead_provider: cpd_lead_provider) }
-    let(:bearer_token) { "Bearer #{token}" }
+  let(:token) { LeadProviderApiToken.create_with_random_token!(cpd_lead_provider: cpd_lead_provider) }
+  let(:bearer_token) { "Bearer #{token}" }
 
+  describe "post" do
```
StupidCodeFactory, line 13:
```suggestion
  describe "POST /api/v1/pariticipant-declarations" do
```

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -414,6 +414,43 @@ def build_params(attributes)
     end
   end
 
+  describe "#void" do
```
StupidCodeFactory, line 417:
```suggestion
  describe "PUT /api/v1/pariticipant-declarations/:id/void" do
```

app/models/participant_declaration.rb
```diff
@@ -64,12 +64,17 @@ class ParticipantDeclaration < ApplicationRecord
   scope :paid_npqs_for_lead_provider, ->(lead_provider) { paid_for_lead_provider(lead_provider).npq }
   scope :paid_uplift_for_lead_provider, ->(lead_provider) { paid_for_lead_provider(lead_provider).uplift }
 
+  # TODO: Voiding paid should trigger clawbacks, but currently OOS
```
StupidCodeFactory, line 67: Let's wait for this https://ukgovernmentdfe.slack.com/archives/C01BE7G56F7/p1637577636468600

and remove this comment. 

I do really think,  that `TODO` should not live in the code, as the PMs have no visibility on it hence ignore this is a thing and can't really factor this into any sprint, leaving the code with lots of potentially never addressed TODOs.

jamesrose: @StupidCodeFactory We'll already have a ticket for this — it's more an FYI to ourselves that we have not yet solved this problem :)

app/services/void_participant_declaration.rb
asmega: nah, i'm doing this now. it would improve the code quality but makes zero difference in outcome of this ticket. I agree it should be done but I need to ship this change so we can do it later

config/locales/en.yml
asmega: where is that convention found? I'm only keeping parity with what's there now. I feel we'd want to change all those in one go rather than have one snowflaked error message

StupidCodeFactory: I was under the impression this was the format used for properties, there are plenty of examples throughout the yaml file, i.e line 56, 57. 
I'll add error message format consistency to our spreadsheet of API improvements.

StupidCodeFactory (approved):

PR #1457, opened 2021-11-19
Fix STI autoloading fix, preventing CI from migrating the database

StupidCodeFactory:
There was apparently some problems with autoloading of ParticipantProfile::ECF
The problem was not described, but the previous fix was not working on CI, preventing the schema from being loaded when a PR was rebase/refreshed against develop with new migrations.


asmega:
I think the code is fine. but I don't see how this would make a difference?

StupidCodeFactory:
> I think the code is fine. but I don't see how this would make a difference?

this delay the query to be run when the class is reference, so later then at boot time in the initialisers

asmega (approved):

PR #1461, opened 2021-11-22
ECF participants API performance

ethanmills:
Rework creation of the mentor_ids hash to eliminate the map

Before and after flamegraphs, viewable with https://www.speedscope.app/
[flamegraphs.zip](https://github.com/DFE-Digital/early-careers-framework/files/7582967/flamegraphs.zip)

I've left the perf tools in, they're probably useful. Happy to remove if wanted though

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -159,6 +159,11 @@ group :development do
 
   # State machine diagrams - https://github.com/Katee/aasm-diagram
   gem "aasm-diagram"
+
+  # Profiling
+  gem "memory_profiler"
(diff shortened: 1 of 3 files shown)


jamesrose (approved):
Thank you. Is it worth doing something similar with `participant_profile_ids` as well?

ethanmills:
> Thank you. Is it worth doing something similar with `participant_profile_ids` as well?

Yes, but that's a little more complex, and I want to make the changes one at a time to measure impact

PR #1468, opened 2021-11-23
Show schema component examples in API docs

ethanmills:
While looking into a query, I noticed that the API docs didn't include the same examples that the swagger docs did, making it really hard to figure out how to call some endpoints.

I've changes the docs templates to add the examples in where they are JSON. I've also updated a couple of examples. The docs are still inconsistent (and sometimes wrong) in a few places, but this should help

### DIFF
--- a/swagger/v1/api_spec.json
+++ b/swagger/v1/api_spec.json
@@ -1482,7 +1482,8 @@
           "participant_type",
           "cohort",
           "status",
-          "training_status"
+          "training_status",
+          "updated_at"
         ],
         "properties": {
           "id": {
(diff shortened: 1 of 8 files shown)


swagger/v1/api_spec.json
```diff
@@ -1466,8 +1482,7 @@
           "participant_type",
           "cohort",
           "status",
-          "training_status",
-          "updated_at"
+          "training_status"
```
ethanmills, line 1485: I'm not sure why this has changed. I ran `bundle exec rake rswag:specs:swaggerize` to generate this.

jamesrose (approved):

PR #1473, opened 2021-11-25
[CPDLP-736] - fix documentation for declaration

StupidCodeFactory:
## Ticket and context

[Ticket:](https://dfedigital.atlassian.net/browse/CPDLP-736)

Fixes the API docs about creating a started or retained participant declaration


swagger/v1/component_schemas/ECFParticipantRetainedDeclaration.yml
```diff
@@ -1,50 +1,13 @@
-description: "An ECF retained participant declaration"
+description: "An participant retain declaration request body"
```
FriedSock, line 1: It seems like we mix "participant" and "ECF participant" a lot in the docs. I think it is probably better to always explicitly state ECF or NPQ, what do you think?

Also minor, but I think this should be "retained" instead of "retain" as well?

swagger/v1/component_schemas/ECFParticipantRetainedDeclarationAttributes.yml
```diff
@@ -0,0 +1,50 @@
+description: "An ECF retained participant declaration"
```
FriedSock, line 1: ECF participant retained declaration?

swagger/v1/component_schemas/ECFParticipantStartedDeclaration.yml
```diff
@@ -1,38 +1,13 @@
-description: "An ECF started and completed participant declaration"
+description: "An participant started declaration request body"
```
FriedSock, line 1: Just one more "An participant" needs fixing here, then looks good.

FriedSock (approved):

PR #1476, opened 2021-11-25
Prevent SITs from editing validated participant details

ethanmills:
Prevent SITs from being able to edit names or emails of participants after they have validated or had any declarations made against them. We're going quite restrictive to start with, then will revisit this later. Context: https://ukgovernmentdfe.slack.com/archives/C01J1J4FM50/p1637772622072000


jamesrose (approved):

tonyheadford (approved):
LGTM

PR #1495, opened 2021-12-01
Ignore Pundit::NotAuthorizedError exceptions in Sentry

jamesrose:
## Ticket and context

Ignore `Pundit::NotAuthorizedError` exceptions in Sentry. I don't think that they are notable or exceptional — do we gain anything from having them in [Sentry](https://sentry.io/organizations/dfe-bat/issues/2735113167/?environment=production&project=5748989&query=is%3Aunresolved)?

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -16,6 +16,8 @@ Sentry.init do |config|
     filter.filter(event.to_hash)
   end
 
+  config.excluded_exceptions += ["Pundit::NotAuthorizedError"]
+
   config.traces_sampler = lambda do |sampling_context|
     transaction_context = sampling_context[:transaction_context]
     op = transaction_context[:op]


ethanmills:
When do we expect to see them? I think we only see them at the moment due to a bug in the impersonation feature?

jamesrose:
@ethanmills I noticed a few where the user doesn't have the right permissions to see that particular page: https://sentry.io/organizations/dfe-bat/issues/2809804517/?environment=production&project=5748989&query=is%3Aunresolved Is that related to the impersonation feature as well?

ethanmills:
Hmm, perhaps not. It was James Hamilton. Perhaps we should make those pages visible to admins though? I think at the moment it's interesting to know who is trying to see what - maybe they should be able to?

jamesrose:
@ethanmills It contains sensitive contract information that we probably don't want to expose to everyone who has access to the admin area.

I agree it's interesting, but I don't think it belongs in Sentry 🤷

ethanmills (approved):

PR #1500, opened 2021-12-01
School access tokens

BroiSatse:
This PR, removes PartnershipNotificationEmail and NominationEmail records and replaces them with a single, more generic SchoolAccessToken.

SchoolAccessToken only goal is to provide authorization for the school user without an account via the email links (just like both email records did previously).

The reason - we have introduced Email audit some while ago now, meaning that PNE and NE records become almost obsolete - the only thing those records were adding on top of Email record were the access tokens to perform specific actions against the School model without need to sign in. Once that role is taken by SAT, those records can be safely deleted.

AccessToken is a generic concept and could be reused for any school related action, without unnecessary coupling to the action implementation. It is a single place to manage token access, further separating concerns of authorisation from the controller.

How it works
------------------

When sending an email with links to the generic school user, we create new SchoolAccessToken with expected permissions against that School record. When user clicks the link, token is taken from the params and stored in the session for next requests - so there is no need to pass the token around in hidden fields and links.

Using a token does not modify `current_user`, as it performs authorisation without authentication (which I am not 100% sure if that's a good idea, but that's how it was designed...). Token authorization must be done explicitly in the controller.


app/mailers/school_mailer.rb
```diff
@@ -79,52 +79,35 @@ def nomination_confirmation_email(sit_profile:, school:, start_url:, step_by_ste
 
   # This email is sent to the school which has been reported by the lead provider and which hasn't yet nominated SIT
   # It also contains request to nominate SIT.
-  def school_partnership_notification_email(recipient:, partnership:, challenge_url:, nominate_url:)
-    template_mail(
+  def school_partnership_notification_email(partnership:, access_token:, reminder: false, campaign: nil)
+    email = template_mail(
       SCHOOL_PARTNERSHIP_NOTIFICATION_EMAIL_TEMPLATE,
-      to: recipient,
+      to: partnership.school.contact_email.presence || "ecf-tech@digital.education.gov.uk",
```
jamesrose, line 85: What's the motivation for sending to us here?

BroiSatse: I'm not 100% sure, but I wanted to avoid as many changes in behaviour as possible. I think the reasoning here might be that we want to be notified if the email fails? Or we want to prevent delivery errors in the service? I'll try to find out :)

ethanmills:
Could I have a description please? If I'm going to read almost 3000 lines of code I'd like to know what it's going to do and why first

app/controllers/challenge_partnerships_controller.rb
```diff
@@ -25,39 +27,34 @@ def success; end
...
-    elsif partnership_id.present?
-      @partnership = Partnership.find(partnership_id)
-      authorize(@partnership, :update?)
+    if access_token.present?
+      if access_token.school != partnership.school || !access_token.permits?(:challenge_partnership)
+        raise Pundit::NotAuthorizedError
+      end
     else
-      raise ActionController::RoutingError, I18n.t(:not_found)
+      authorize(partnership, :update?)
```
ethanmills, line 35: I'm not sure what's going on here, what is `current_user` going to be?

BroiSatse: `current_user` returns a user who is signed in. When accessing any page with a token, `current_user` is nil, the same as with NominationEmail tokens. I've considered creating `AnonymousSchoolUser` object to use as a `current_user` for token access, however it would require quite a strange interface to mimic some of the association required in policies.

But maybe it would not be that hard, happy to revisit?

app/forms/nominate_induction_tutor_form.rb
```diff
@@ -16,19 +16,17 @@ def attributes
...
-    if school_id
-      School.find school_id
-    else
-      NominationEmail.find_by(token: token).school
-    end
+    School.find school_id
   end
 
   def email_already_taken?
+    # byebug
```
tonyheadford, line 29: Can we remove this comment please?

app/lib/mail/delivery_recorder.rb
```diff
@@ -14,16 +14,16 @@ def initialize(enabled:)
     def delivered_email(mail)
       return unless enabled?
 
-      response = mail.delivery_method.response
+      response = mail.delivery_method.response if mail.delivery_method.respond_to?(:response)
```
tonyheadford, line 17: In what circumstances would this not be expected to respond to `:response`?  Perhaps add a comment to explain?

app/services/invite_schools.rb
```diff
@@ -366,26 +374,19 @@ def find_school(urn)
...
-      expiry_date: email_expiry_date,
-    ).deliver_now.delivery_method.response.id
-
-    nomination_email.update!(notify_id: notify_id)
+  def send_nomination_email(to:, access_token:)
+    SchoolMailer.nomination_email(
+      recipient: to,
+      school: access_token.school,
+      access_token: access_token,
+    ).deliver_later
```
tonyheadford, line 389: Just a thought - could the creation of the access token be moved into the school mailer to clean the method calls up a bit as it seems to be passed around a bit? Also is there a risk here that if background jobs are blocked for some reason the access_token could expire before being sent? (I'm guessing that's highly unlikely)

spec/requests/nominations/choose_how_to_continue_spec.rb
```diff
@@ -76,43 +72,40 @@
...
-        { nominate_how_to_continue_form: { how_to_continue: "no", token: token } }
+        { nominate_how_to_continue_form: { how_to_continue: "no" } }
       end
 
       context "when making a choice for the first time" do
         let(:nomination_email) { create :nomination_email, opened_at: nil }
         let(:now) { rand(-100..100).hours.ago }
 
-        it "records opened_at on the nomination_email" do
+        pending "records opened_at on the nomination_email" do
```
tonyheadford, line 108: why pending?

spec/requests/nominations/choose_how_to_continue_spec.rb
```diff
@@ -123,7 +116,7 @@
       context "when opening nomination page for another time" do
         let(:nomination_email) { create :nomination_email, opened_at: 10.days.ago }
 
-        it "does not record opened_at on the nomination_email" do
+        pending "does not record opened_at on the nomination_email" do
```
tonyheadford, line 119: why pending?

tonyheadford (changes requested):
One comment to remove and a couple of questions, otherwise I think it looks fine

PR #1503, opened 2021-12-02
Split out web and worker crash alert rules

jamesrose:
## Ticket and context

We currently alert when any instance type crashes. This PR splits out web and worker crash alert rules so that we can better identify what type of instance has crashed.

### DIFF
--- a/terraform/monitoring/config/production/alert.rules.yml
+++ b/terraform/monitoring/config/production/alert.rules.yml
@@ -70,10 +70,18 @@ groups:
   - name: Crashed-apps
     rules:
       - alert: AppsCrashed
-        expr: rate(crash{app=~"ecf-production"}[1m])*60 > 1
+        expr: rate(crash{app="ecf-production"}[1m])*60 > 1
         annotations:
           summary: Crashed Apps non-zero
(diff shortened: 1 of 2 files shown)


terraform/monitoring/config/development/alert.rules.yml
```diff
@@ -70,10 +70,18 @@ groups:
...
-        expr: rate(crash{app=~"ecf-dev"}[1m])*60 > 1
+        expr: rate(crash{app="ecf-dev"}[1m])*60 > 1
         annotations:
           summary: Crashed Apps non-zero
           description: "ecf-dev: At least 1 crashed app (current value: {{ $value }})"
         labels:
           environment: development
           severity: low
+      - alert: AppsCrashed
+        expr: rate(crash{app="ecf-dev-worker"}[1m])*60 > 1
```
ethanmills, line 81: This doesn't exist

ethanmills (approved):
Only hitch here is we lose anything about the `venerable` versions, but I think that's fine

terraform/monitoring/config/development/alert.rules.yml
jamesrose: Great spot, thank you.

PR #1508, opened 2021-12-03
CPDLP-764 cohort on schedule

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-764

- Previously a schedule was not directly associated with a cohort
- This was implied through the name of the schedule
- This adds the association between schedules and cohorts
- The  schedule name can now therefore be free from containing the cohort year
- This change at the moment is somewhat superficial as this association is not utilised but will pave the way for future changes
- After this is deployed i will ensure all environments contain this extra data

### Code quality checks

### HTML Checks

### Gotchas

- `cf ssh ecf-review-pr-1508`
- start rails console
- `pp Finance::Schedule.all`
- all schedules should belong to a cohort

## External API changes

- There are no external api changes

### DIFF
--- a/spec/factories/finance/schedules.rb
+++ b/spec/factories/finance/schedules.rb
@@ -21,6 +21,8 @@ FactoryBot.define do
       end
     end
 
+    cohort { Cohort.find_or_create_by(start_year: 2021) }
+
     factory :ecf_schedule, class: "Finance::Schedule::ECF", parent: :schedule do
       name { "ECF September standard 2021" }
       schedule_identifier { "ecf-september-standard-2021" }
--- a/spec/factories/finance/schedules.rb
+++ b/spec/factories/finance/schedules.rb
@@ -21,7 +21,7 @@ FactoryBot.define do
       end
     end
 
-    cohort { Cohort.find_or_create_by(start_year: 2021) }
+    cohort { Cohort.find_or_create_by!(start_year: 2021) }
 
     factory :ecf_schedule, class: "Finance::Schedule::ECF", parent: :schedule do
       name { "ECF September standard 2021" }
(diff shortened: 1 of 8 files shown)


spec/factories/finance/schedules.rb
```diff
@@ -21,6 +21,8 @@
       end
     end
 
+    cohort { Cohort.find_or_create_by(start_year: 2021) }
```
StupidCodeFactory, line 24:
```suggestion
    cohort { Cohort.find_or_create_by!(start_year: 2021) }
```

StupidCodeFactory: _if_ we add validation later, that might be easier to spot the failure.

spec/models/cohort_spec.rb
```diff
@@ -3,6 +3,16 @@
...
 
 RSpec.describe Cohort, type: :model do
+  describe "#schedules" do
+    subject { described_class.create!(start_year: 3000) }
+
+    let!(:schedule) { create(:ecf_schedule, cohort: subject) }
+
+    it "returns associated scheules" do
+      expect(subject.schedules).to include(schedule)
+    end
```
StupidCodeFactory, line 13: This test seems a bit redundant, that basically boils down to test active record. We can leave it, I'm not a massive fan of this. An alternative would be to use the shoulda matchers 

```ruby
is_expected.to bleong_to(:schedule)
```
The intent to test that the relationship is setup reads a bit clearer. But Im not fussed as it's essentially the same.

StupidCodeFactory (approved):
Could you add the

spec/models/cohort_spec.rb
asmega: sure. i only added as `Schedule` is namespaced under `Finance::Schedule` so wanted to confirm it

jamesrose (approved):

PR #1517, opened 2021-12-06
CST-112 Multiple email addresses

tonyheadford:
## Ticket and context

Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-112)

## Tech review

Enable multiple email addresses to be stored against a user to aid with duplicate management
Uses `ParticipantIdentity` to hold original `User.id` and `User.email` and an origin (`ecf` or `npq`)
`ParticipantIdentity` associates a user with their participant profiles.
A `User` can have many `ParticipantProfile`s
A `ParticipantProfile` can have many `ParticipantProfile`s

Currently the link from `ParticipantProfile` to `ParticipantIdentity` is `optional` in order to roll this out.
We will need to add a migration to update the field to a foreign key and remove the `optional` clause once rolled out.

### Code quality checks

### HTML Checks

### Gotchas

## External API changes


app/controllers/api/v1/ecf_participants_controller.rb
```diff
@@ -42,8 +42,9 @@ def lead_provider
 
       def participants
         participant_profiles = ParticipantProfile::ECF.where(id: participant_profile_ids)
-                                                      .joins(:user)
+                                                      .joins(:participant_identity, :user)
```
ethanmills, line 45: Will this change not break the API response before we backfill the data?

app/controllers/api/v1/ecf_participants_controller.rb
```diff
@@ -82,6 +75,21 @@ def mentor_ids
...
       end
+
+      def fetch_participant_profile_ids
+        # this retrieves the list of ECF participant profiles for the LeadProvider, one per
+        # teacher_profile (now participant_identity).
+        # Withdrawn profiles are included unless there is also an active one.
+        # The DISTINCT ON clause chooses the first record after ordering by status.
+        inner_query = lead_provider
+          .ecf_participant_profiles
+          .select("DISTINCT ON (participant_profiles.participant_identity_id) participant_identity_id, participant_profiles.status, participant_profiles.id AS id")
```
ethanmills, line 86: ditto

ethanmills:
I think this breaks our check on email usage everywhere we create users

app/controllers/api/v1/ecf_participants_controller.rb
tonyheadford: Yes, good catch. Will need to pull some of this out and add in after backfilling ...

tonyheadford:
> I think this breaks our check on email usage everywhere we create users

Could you elaborate @ethanmills ?

lib/devise/strategies/passwordless_authenticatable.rb
```diff
@@ -18,7 +18,16 @@ def valid?
 
       def authenticate!
         if params[:user].present?
-          user = User.find_by(email: params[:user][:email])
+          email = params.dig(:user, :email)
+
+          identity = ParticipantIdentity.find_by(email: email)
+          if identity.present?
+            confirmed_email = identity.email
```
ethanmills, line 25: Why do we need `confirmed_email`? Won't it always just be the same as `email`, or am I missing something?

ethanmills (approved):

lib/devise/strategies/passwordless_authenticatable.rb
tonyheadford: The supplied email could be assorted case, there was an existing spec that tested the email used in the sign in email was the "normalized" version

PR #1535, opened 2021-12-08
Add reasons to papertrail

ethanmills:
Add a reason column to papertrail, for recording notes about changes made on the console

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,6 +239,22 @@ per participant service fee £398 (40%) >> monthly service fee £27k >> total se
 * "Output payments" are payments made based on the performance of the training provider (i.e. their output).
 * "Payment type" for start/retention_x/completion output payments.
 
+## Storing reasons in PaperTrail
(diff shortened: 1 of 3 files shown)


jamesrose (approved):

tonyheadford (commented):
That is going to take some memory/discipline as often in the console for hours. It would be easy to forget to update the reason when working on multiple things.  Is occasionally having the wrong reason better/worse than not having a reason?

ethanmills:
> That is going to take some memory/discipline as often in the console for hours. It would be easy to forget to update the reason when working on multiple things. Is occasionally having the wrong reason better/worse than not having a reason?

I suggest using the block syntax where possible

tonyheadford:
> > That is going to take some memory/discipline as often in the console for hours. It would be easy to forget to update the reason when working on multiple things. Is occasionally having the wrong reason better/worse than not having a reason?
> 
> I suggest using the block syntax where possible

Yes, means you'd need to write all your changes outside of the console and copy paste them back in a block though. May help up write tooling quicker though!

PR #1580, opened 2022-01-06
Start induction term listing changes 

xswoodsx:
## Ticket and context

Ticket: 
[246](https://dfedigital.atlassian.net/jira/software/projects/CST/boards/119?assignee=5f3543bb3e9e2e004db3189d&selectedIssue=CST-246) 
[247](https://dfedigital.atlassian.net/jira/software/projects/CST/boards/119?assignee=5f3543bb3e9e2e004db3189d&selectedIssue=CST-247)

246
Screenshot 2022-01-06 at 08 15 56

247
Screenshot 2022-01-06 at 08 15 11

### Code quality checks

### HTML Checks

### Gotchas

2. Login in as `cpd-test+tutor-10@example.com`
3. go to the dashboard. All users who haven't been assigned a start date will be grouped as Autumn 2021. 
4. Add an ECT or Mentor, follow the journey and Autumn 2021 should be removed as an option. 
5. Complete the journey and that ppt should be on your dashboard with the term you selected.


tonyheadford (approved):

app/models/participant_profile.rb
```diff
@@ -116,6 +114,13 @@ def sit_mentor?
   def policy_class
     ParticipantProfilePolicy
   end
+
+  def start_term_human
+    # Shouldn't we just update records in the database instead?
+    return "Autumn 2021" if start_term.blank?
```
jamesrose, line 120: Can `start_term` be blank? It looks like we have a default set and `null: false`

app/models/participant_profile.rb
```diff
@@ -55,6 +49,10 @@ class ParticipantProfile < ApplicationRecord
 
   self.ignored_columns = %w[user_id]
 
+  def self.humanize_start_term(start_term)
+    start_term.to_s.split("_").map(&:capitalize).join(" ")
```
jamesrose, line 53:
```suggestion
    start_term.to_s.humanize
```

I don't think this belongs in the model. It's a view-layer helper, if anything!

app/forms/schools/add_participant_form.rb
```diff
@@ -89,6 +90,10 @@ def mentor
       @mentor = (User.find(mentor_id) if mentor_id.present? && mentor_id != "later")
     end
 
+    def start_term_options
+      @start_term_options ||= %w[spring_2022 summer_2022]
```
jamesrose, line 94: My view is that the applicable values here belong in the model. What was the motivation for moving them out?

app/models/participant_profile.rb
jamesrose: That is, I don't think you need to update the model because it's already set, and this method becomes unnecessary.

app/models/participant_profile.rb
BroiSatse: Very much agree and was really unhappy about adding this in. helper does not feel right either here (they never do tbh), but I didn't want to repeat the same logic in all the places it is required. It is however sort-of a view concern, so I will, with heavy heart, move this to helper

app/models/participant_profile.rb
BroiSatse: Ah interesting - I wonder why we had all the `start_term || "Autumn 2021"` in the code. But - in any case, the default is incorrect - it should be `"autumn_2021"` to fit all the other values...

xfiddyx: I thought there was a bunch of records that didn't have any start_term date set? I might have that wrong but i thought we added the attribute and select a term date page after we had users. That's why I had the `start_term || "Autumn 2021"`

app/forms/schools/add_participant_form.rb
BroiSatse: Ok, that's the interesting bit. We have two types of values here:

  1. model values - containing all the possible values that might have been assigned to models in the past (and ideally in the future), which we probably should have format validator for.
  2. form values - containing all the possible values that can be assigned at the time of using the form.

This list was previously defined as enum on the model (and also used for translations into humanized string). We could not remove values from this enum, as this would break the existing models. As a result we had a lovely cryptic `ParticipantProfile.start_terms.drop(1).each` here to skip the first option when displaying the form, which would shortly turn into `drop(2)`. This would not scale at all - in few years that enum would grow to 20 and more elements while updating the magic number in the form. In the ideal scenario those values would be generated rather than hardcoded, but I did not want to bring too much into a bug-fix PR.

app/models/participant_profile.rb
jamesrose: @BroiSatse Although on second pass this is for a _human_ version of the `start_term` — e.g., not the form value. I _think_ the entire method here is not needed: can just depend on the default and use `humanize_start_term(start_term)` ?

PR #1605, opened 2022-01-10
[CPDLP-757] - add rejection reason to participant declaration

StupidCodeFactory:
## Ticket and context

Tickets: [CPDLP-757](https://dfedigital.atlassian.net/browse/CPDLP-757), [CPDLP-758](https://dfedigital.atlassian.net/browse/CPDLP-758)

## Tech review

I've slightly changed the acceptance criteria of CPDLP-757 - instead of recording a rejection reason I'm marking the participant record as ineligible and store the original participant declaration, this ensure we know why the declaration is ineligible and also to which declaration it is the duplicate of. Also ensures with do not allow `app/services/record_declarations/actions/make_declarations_payable.rb` to change its state to `payable`

When a declaration is made and 
### Code quality checks

### HTML Checks

### Gotchas

## External API changes


app/models/participant_declaration.rb
```diff
@@ -89,7 +92,24 @@ def make_paid!
     DeclarationState.paid!(self) if payable?
   end
 
+  def make_ineligible!
+    DeclarationState.ineligible!(self)
```
jamesrose, line 96: I think this should only be possible from `submitted?` for now.

app/models/participant_declaration.rb
```diff
@@ -5,13 +5,18 @@ class ParticipantDeclaration < ApplicationRecord
   belongs_to :cpd_lead_provider
   belongs_to :user
   belongs_to :participant_profile
+  belongs_to :original_participant_declaration, class_name: "ParticipantDeclaration", optional: true
+  has_many :duplicate_participant_declarations, class_name: "ParticipantDeclaration", foreign_key: :original_participant_declaration_id, inverse_of: :original_participant_declaration
+
+  include PGEnum(ineligibility_reason: %w[duplicate])
```
jamesrose, line 11: How about `state_reason` ?

app/models/participant_declaration.rb
```diff
@@ -89,7 +94,30 @@ def make_paid!
...
+
+  def duplicate_declarations
+    self.class.joins(participant_profile: :teacher_profile)
+      .where(participant_profiles: { teacher_profiles: { trn: participant_profile.teacher_profile.trn } })
+      .where.not(participant_profiles: { teacher_profiles: { trn: nil } })
+      .where.not(user_id: user_id)
+      .where(
+        declaration_type: declaration_type,
+        course_identifier: course_identifier,
+        declaration_date: declaration_date,
```
jamesrose, line 116: Do we need to check this?

app/services/record_declarations/base.rb
```diff
@@ -64,10 +83,11 @@ def create_declaration_attempt!
     end
 
     def find_or_create_record!
-      self.class.declaration_model.find_or_create_by!(declaration_parameters).tap do |participant_declaration|
-        DeclarationState.submitted!(participant_declaration)
-        participant_declaration.make_eligible! if fundable?
-      end
+      self.class.declaration_model.find_or_create_by!(declaration_parameters)
```
jamesrose, line 86: This should still `DeclarationState.submitted!(participant_declaration)`

spec/services/record_declarations/base_spec.rb
```diff
@@ -50,7 +126,7 @@ def matches_lead_provider?
     end
 
     it "does not have errors on milestone_date" do
-      expect { subject.call }.not_to raise_error(ActionController::ParameterMissing)
```
StupidCodeFactory, line 53: This was creating a false positive as the warned by RSpec.
The error was the example model was missing `def self.declaration_model`

spec/factories/participant_identity.rb
```diff
@@ -10,12 +10,5 @@
     trait :npq_origin do
       origin { "npq" }
     end
-
-    initialize_with do
-      ParticipantIdentity.find_or_initialize_by(email: email) do |identity|
-        identity.user = user
-        identity.external_identifier = user.id
-      end
-    end
```
StupidCodeFactory, line 19: This now uses the service class in the participant_profile factory. Also removes the need to maintain the same logic in the production code and test code which further enhance the class' test coverage.
We could be quite aggressive and delete `spec/services/identity/create_spec.rb` since it this class would be used all over the place by the `:participant_profile` factory.

spec/factories/lead_provider.rb
```diff
@@ -4,7 +4,6 @@
   factory :lead_provider do
     name  { "Lead Provider" }
     cohorts { Cohort.all }
-    cpd_lead_provider
```
StupidCodeFactory, line 7: This was creating an extra record causing test to have false positive

app/services/record_declarations/base.rb
StupidCodeFactory: this has been moved further up into `#call` 
https://github.com/DFE-Digital/early-careers-framework/pull/1605/files#diff-ec96a6b08aefdf2afd683dd457777d9d18ca2029614ad570f21bb4b25f0df667R44

app/services/record_declarations/base.rb
```diff
@@ -18,7 +20,7 @@ class Base
     validate :date_has_the_right_format
 
     validates :declaration_type, inclusion: { in: :valid_declaration_types, message: I18n.t(:invalid_declaration_type) }
-    delegate :schedule, :participant_declarations, :fundable?, to: :user_profile, allow_nil: true
```
StupidCodeFactory, line 21: remove an extra later of indirection with reduces the cognitive load needed to understand this class

app/services/record_declarations/base.rb
```diff
@@ -55,6 +60,19 @@ def initialize(params:)
       self.declaration_type = params[:declaration_type]
     end
 
+    def set_eligibility
+      if participant_declaration.duplicate_declarations.count > 1
+        raise MultipleParticipantDeclarationDuplicate
```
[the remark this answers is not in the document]
StupidCodeFactory: This could occur if we had a teacher registering with 3 different email address. In which we might want to take a closer look. This was a mean to sort of red flag something is probably a bit off. 
I'm happy to remove this check, I agree that is a bit defensive.

app/models/declaration_state.rb
```diff
@@ -3,19 +3,27 @@
 class DeclarationState < ApplicationRecord
   belongs_to :participant_declaration
 
+  include PGEnum(state_reason: %w[duplicate])
```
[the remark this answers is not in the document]
StupidCodeFactory: it is using Rails enum under the hood, this is to also enforce the data consistency at database level, this maps the rails enum to a PG enumerated type. Just good practice to keep the application layer and database layer consistent. 
Happy to remove if we don't care about enforcing the consistency at the database level, we do add foreign key contraints so I figured might as well do it for enums.

spec/factories/lead_provider.rb
StupidCodeFactory: I'd rather use the higher level factory `create(:cpd_lead_provider, :with_lead_provider).lead_provider`, that avoids creating extra db records in both directions. But I'll ✂️ this.

app/models/declaration_state.rb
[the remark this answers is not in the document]
StupidCodeFactory: https://github.com/alassek/activerecord-pg_enum

app/services/record_declarations/base.rb
```diff
@@ -130,5 +149,9 @@ def milestone
     def valid_declaration_types
       self.class.valid_declaration_types
     end
+
+    def original_participant_declaration
+      @original_participant_declaration ||= participant_declaration.duplicate_declarations.first
```
jamesrose, line 154: Is `first` consistent here?

app/models/participant_declaration.rb
```diff
@@ -5,13 +5,16 @@ class ParticipantDeclaration < ApplicationRecord
   belongs_to :cpd_lead_provider
   belongs_to :user
   belongs_to :participant_profile
+  belongs_to :superseded_by, class_name: "ParticipantDeclaration", optional: true
+  has_many :duplicate_participant_declarations, class_name: "ParticipantDeclaration", foreign_key: :superseded_by_id, inverse_of: :superseded_by
```
jamesrose, line 9: What do you think about this association name? Does this read better than something like `supersedes`?

app/services/record_declarations/base.rb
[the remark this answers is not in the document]
jamesrose: Yeah, I think we should remove it. It seems to me like we might care about this, but it's not a show stopper. As long as we make subsequent declarations ineligible.

app/models/declaration_state.rb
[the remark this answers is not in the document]
jamesrose: I can see the advantages but I'd rather be consistent until we make a conscious decision to migrate over to it throughout the service. It is a bit obtuse, but not a show stopper.

spec/services/record_declarations/base_spec.rb
```diff
@@ -3,10 +3,82 @@
...
+
+      let(:original_participant_declaration) { ParticipantDeclaration.find(JSON.parse(record_original_declaration).dig("data", "id")) }
+
+      before { record_original_declaration }
+
+      it "transitions the declaration to ineligible", :aggregate_failures do
+        duplicate_participant_declaration = ParticipantDeclaration.find(JSON.parse(record_declaration).dig("data", "id"))
+
+        expect(original_participant_declaration.duplicate_participant_declarations).to eq([duplicate_participant_declaration])
+        expect(duplicate_participant_declaration.declaration_states.pluck(:state)).to eq(%w[submitted ineligible])
```
jamesrose, line 76: Nice, thanks for checking the history :)

app/models/declaration_state.rb
```diff
@@ -3,18 +3,21 @@
...
```
jamesrose, line 6: Do we want to expose this in the serializer yet, or are we going to hold off for now?

StupidCodeFactory: We've agreed with Nathan to hold off for now, until a provider actually ask for something.

app/models/participant_declaration.rb
```diff
@@ -89,7 +92,26 @@ def make_paid!
...
+
+  def duplicate_declarations
+    self.class.joins(participant_profile: :teacher_profile)
+      .where(participant_profiles: { teacher_profiles: { trn: participant_profile.teacher_profile.trn } })
+      .where.not(participant_profiles: { teacher_profiles: { trn: nil } })
+      .where.not(user_id: user_id, id: id)
+      .where(
+        declaration_type: declaration_type,
+        course_identifier: course_identifier,
+        superseded_by_id: nil,
```
jamesrose, line 111: I think we'll want `voided: false` here as well — we should allow them to post dups of voided declarations.

app/services/record_declarations/base.rb
```diff
@@ -55,6 +60,15 @@ def initialize(params:)
       self.declaration_type = params[:declaration_type]
     end
 
+    def set_eligibility
```
jamesrose, line 63: This method is great — very clear!

app/services/record_declarations/base.rb
```diff
@@ -82,13 +97,15 @@ def declaration_parameters
...
     def record_exists_with_different_declaration_date?
-      declaration = self.class.declaration_model.find_by(
-        user: user,
-        course_identifier: course_identifier,
-        declaration_type: declaration_type,
-      )
-
-      declaration.present? && !declaration.voided? && declaration.declaration_date != Time.zone.parse(declaration_date)
+      declaration = self.class.declaration_model
+                      .where.not(state: self.class.declaration_model.states[:voided])
```
jamesrose, line 101: What's the advantage of this vs below?

```suggestion
                      .where.not(state: ParticipantDeclaration.states[:voided])
```

app/models/participant_declaration.rb
```diff
@@ -89,7 +92,27 @@ def make_paid!
     DeclarationState.paid!(self) if payable?
   end
 
+  def make_ineligible!(reason: nil)
```
jamesrose, line 95:
```suggestion
  def make_ineligible!(reason:)
```

This one is up to you: whether we enforce a reason or leave it flexible.

jamesrose (approved):
@StupidCodeFactory Super work 👏  I'm approving this now because I don't think these comments are show stoppers.

PR #1636, opened 2022-01-18
[CPDLP-899] Fix change schedule with null identifier

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-899

- First commit is a refactor so you may want to review on commit basis
- Fixes passing in a `null` `schedule_identifier` when changing schedule. Now throws an error instead of choosing an incorrect schedule
- This is a regression that was introduced when `identifier_alias` was added
- Added `Schedule#schedule_identifier` presence validation to prevent any bad data being introduced 
- Improved some error messages, the rails default messages don't quite cut it for our needs

### Code quality checks

### HTML Checks

### Gotchas

- Find a participant
- Make API call to change their schedule to a `null` `schedule_identifier`
- Alternatively an other incorrectly named schedule
- Should return an error
- Schedule should not be changed

## External API changes

- None


app/controllers/api/v1/participants_controller.rb
```diff
@@ -9,7 +9,9 @@ class ParticipantsController < Api::ApiController
       include ParticipantActions
 
       def change_schedule
-        perform_action(service_namespace: ::Participants::ChangeSchedule)
+        service = recorder(service_namespace: ::Participants::ChangeSchedule).new(params: params_for_recorder)
+        result = service.call
+        render json: serialized_response(result)
       end
```
jamesrose, line 15: Has this changed from what `perform_action` is doing?

asmega: I think I see where you're going. `perform_action` has not changed and continues to work for the other actions. The internal API for `ChangeSchedule` has changed however this looks to be the only placed where it is used

StupidCodeFactory: This should also be available under the `/ecf` namespace
Whenever I've touched this part of the code I've ported as well under `/ecf` namespace

app/services/participants/change_schedule/base.rb
```diff
@@ -1,16 +1,89 @@
...
+
+      def cohort_object
+        @cohort_object ||= if @cohort
+                             Cohort.find_by(start_year: @cohort)
+                           else
+                             Cohort.current
+                           end
+      end
+
+      def schedule
```
StupidCodeFactory, line 57: Suggestion: the idiomatic way to do this and slightly, less contrived in my option.
```suggestion
        def schedule
          @schedule ||= begin
            alias_search_query = Finance::Schedule
              .where("identifier_alias IS NOT NULL")
              .where(identifier_alias: schedule_identifier, cohort: cohort_object)

            Finance::Schedule
              .where(schedule_identifier: schedule_identifier, cohort: cohort_object)
              .or(alias_search_query)
              .first
          end
        end
```

app/services/participants/change_schedule/base.rb
```diff
@@ -1,16 +1,89 @@
...
+          .or(alias_search_query)
+          .first
+      end
+
+      def not_already_withdrawn
+        errors.add(:participant_id, I18n.t(:withdrawn_participant)) if participant_profile_state&.withdrawn?
+      end
+
+      def schedule_valid_with_pending_declarations
+        user_profile&.participant_declarations&.each do |declaration|
```
StupidCodeFactory, line 75: Suggestion: prefer early returns:

```ruby 
return unless user_profile&.participant_declarations
```

app/services/participants/change_schedule/early_career_teacher.rb
```diff
@@ -3,7 +3,15 @@
 module Participants
   module ChangeSchedule
     class EarlyCareerTeacher < ECF
-      include Participants::EarlyCareerTeacher
```
StupidCodeFactory, line 6: 🏆

app/services/participants/change_schedule/ecf.rb
```diff
@@ -3,8 +3,13 @@
 module Participants
   module ChangeSchedule
     class ECF < Base
-      include Participants::ECF
-      include ValidateAndChangeSchedule
```
StupidCodeFactory, line 7: 🏆

app/services/participants/change_schedule/ecf.rb
```diff
@@ -3,8 +3,13 @@
...
+      delegate :school_cohort, to: :user_profile, allow_nil: true
+
+    private
+
+      def matches_lead_provider?
+        cpd_lead_provider == school_cohort&.lead_provider&.cpd_lead_provider
```
StupidCodeFactory, line 11: Suggestion: it think this reads much better
```ruby
return false unless school_cohort&.lead_provider

cpd_lead_provider == school_cohort.lead_provider.cpd_lead_provider
```

app/services/participants/change_schedule/mentor.rb
```diff
@@ -3,7 +3,15 @@
 module Participants
   module ChangeSchedule
     class Mentor < ECF
-      include Participants::Mentor
```
StupidCodeFactory, line 6: 🏆

app/services/participants/change_schedule/npq.rb
```diff
@@ -3,8 +3,19 @@
 module Participants
   module ChangeSchedule
     class NPQ < Base
-      include Participants::NPQ
-      include ValidateAndChangeSchedule
```
StupidCodeFactory, line 7: 🏆

StupidCodeFactory (approved):
looks good to me. Left some stylistic comments, not massively fussed about any of those but maybe this one https://github.com/DFE-Digital/early-careers-framework/pull/1636/files#diff-dfc2406ed84c175f294531cf175ed2057f1c713dfb3bd614231a8271e0f0ed32R57

Up to you

app/controllers/api/v1/participants_controller.rb
asmega: It is already available under the /ecf namespace. the routing just maps here as at the moment the behaviour is identical

app/services/participants/change_schedule/base.rb
asmega: sure we can change that next time around

app/services/participants/change_schedule/ecf.rb
asmega: I've mostly left these as is for now

app/services/participants/change_schedule/base.rb
asmega: I've mostly left these as is for now

PR #1642, opened 2022-01-19
[CPDLP-825] - adapt script to transition NPQ declarations

StupidCodeFactory:
## Ticket and context

[Ticket:](https://dfedigital.atlassian.net/browse/CPDLP-825) 

Adapt make declaration payable to be able to handle both ECF and NPQ declarations

### DIFF
--- a/app/services/record_declarations/actions/make_declarations_payable.rb
+++ b/app/services/record_declarations/actions/make_declarations_payable.rb
@@ -4,13 +4,13 @@ module RecordDeclarations
   module Actions
     class MakeDeclarationsPayable
       class << self
-        def call(start_date: "2021-09-01 00:00:00", end_date: "2021-11-01 00:00:00")
-          new.call(start_date: start_date, end_date: end_date)
+        def call(declaration_class:, start_date: "2021-09-01 00:00:00", end_date: "2021-11-01 00:00:00")
+          new.call(declaration_class: declaration_class, start_date: start_date, end_date: end_date)
         end
       end
 
-      def call(start_date:, end_date:)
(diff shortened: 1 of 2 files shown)


app/services/record_declarations/actions/make_declarations_payable.rb
```diff
@@ -4,13 +4,13 @@ module RecordDeclarations
   module Actions
     class MakeDeclarationsPayable
       class << self
-        def call(start_date: "2021-09-01 00:00:00", end_date: "2021-11-01 00:00:00")
-          new.call(start_date: start_date, end_date: end_date)
+        def call(declaration_type:, start_date: "2021-09-01 00:00:00", end_date: "2021-11-01 00:00:00")
```
jamesrose, line 7: I think this is a bit confusing as `ParticipantDeclaration` has a `declaration_type` field. `declaration_klass` ?

StupidCodeFactory: @jamesrose it's stored as the `type` in the the db. also could just be `declaration_class`

tonyheadford (commented):
This smells like it could be a concern instead of a service?

StupidCodeFactory:
> This smells like it could be a concern instead of a service?

Where would you inject this ?

StupidCodeFactory:
> > > This smells like it could be a concern instead of a service?
> > 
> > 
> > Where would you inject this ?
> 
> Into the `declaration_type` thing that you're manipulating? (Admittedly I've only looked at this PR not the wider code so feel free to ignore if I've got the wrong end of the stick)

Oh I see, I don't think the database layer should know anything about the application business other then its data model.

asmega (approved):

PR #1644, opened 2022-01-19
Fix for term editing

ngw:
Had to move the start_term options back to ParticipantProfile.
These changes fix https://sentry.io/organizations/dfe-bat/issues/2942896307/?project=5748989&referrer=slack

### DIFF
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -42,7 +42,7 @@ module Schools
 
       validates :start_term,
                 presence: { message: I18n.t("errors.start_term.blank") },
-                inclusion: { in: ParticipantProfile::START_TERM_OPTIONS }
+                inclusion: { in: ParticipantProfile::ECF::CURRENT_START_TERM_OPTIONS }
 
       next_step do
         if type == :ect && mentor_options.any?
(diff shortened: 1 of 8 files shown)


app/forms/schools/add_participant_form.rb
```diff
@@ -90,10 +90,6 @@ def mentor
       @mentor = (User.find(mentor_id) if mentor_id.present? && mentor_id != "later")
     end
 
-    def start_term_options
-      @start_term_options ||= %w[spring_2022 summer_2022]
-    end
```
BroiSatse, line 95: Uh, no, why?

app/models/participant_profile.rb
```diff
@@ -49,6 +49,8 @@ class ParticipantProfile < ApplicationRecord
 
   self.ignored_columns = %w[user_id]
 
+  START_TERM_OPTIONS = %w[spring_2022 summer_2022].freeze
```
BroiSatse, line 52: This should not live in the model - this is the collection of **current** possibilities for this field and as such is a concern of the form rather than the model. Model has potentially infinite amount of start_term_options

jamesrose: Agreed. I agree that those values probably shouldn’t be on the the profile, but if you do want to do that and return to improving the approach, put them on `ParticipantProfile::ECF` as they are not relevant to NPQ.

ngw: Correct me if I'm wrong, but these are also the values for mentors?

jamesrose (approved):

PR #1651, opened 2022-01-20
Fix issue with jobs being picked up before transaction completes

tonyheadford:
## Ticket and context

(Re-creating because of build issues in the previous PR)
Issue reported where the partnership challenge reason in the email has the text `translation missing: en.partnerships.challenge_reasons` which would indicate that the reason has not been set.
However the code to schedule the mailer is within the transaction block that updates the reason, and so it is thought that the mailer job is being picked up (in some cases, not all) before the transaction has been written to the db.

This moves the mail sending code outside the transaction block

### Code quality checks

### HTML Checks

### Gotchas

## External API changes


xswoodsx (approved):
LGTM :)

PR #1670, opened 2022-01-25
Ensuring that all API actions are recorded against the correct participant profile

BroiSatse:
As most of the duplicate teacher profiles has been now merged together, we should be always using participant identifier to identify the profile providers are acting upon. The old logic was trying to find the profile by going identity => user => find_profile, which is not guaranteed to return the correct profile, as users can now have multiple profiles at the same time (this on it own is a result of my mistake during de-dup process and will be handled separately). For example -  `identity.user.early_career_teacher_profile`, as a `has_one` profile would return one of the profiles if two matching profiles are present. However, since providers use identifier we recorded on participant identity, we can actually find the correct profile from `identity` record with `identity.participant_profiles.ect.active_record.first`.

### DIFF
--- a/app/services/participants/early_career_teacher.rb
+++ b/app/services/participants/early_career_teacher.rb
@@ -6,12 +6,14 @@ module Participants
     extend ActiveSupport::Concern
     included do
       extend EarlyCareerTeacherClassMethods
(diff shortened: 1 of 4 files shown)


tonyheadford (approved):
Good spot

PR #1686, opened 2022-01-31
Add identity external identifier to analytics database

jamesrose:
## Ticket and context

This will allow people who consume the analytics database to determine whether a profile has been deduplicated. 

Post-deploy I'll need to re-insert all records so that they have this field.


db/analytics_migrate/20220131150652_add_external_id_to_ecf_participant.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddExternalIdToECFParticipant < ActiveRecord::Migration[6.1]
+  def change
+    add_column :ecf_participants, :external_id, :string
```
asmega, line 5: probably not worth changing but i usually just use `text` for postgres

asmega (approved):

PR #1705, opened 2022-02-08
Improve search and displayed information on finance participant search

jamesrose:
## Ticket and context

- Allow searching by User ID or the Participant Identity's external ID
- Add declaration state history to information displayed


app/controllers/finance/participants_controller.rb
```diff
@@ -14,8 +14,18 @@ def index
...
+  private
+
+    def find_user_by_id_or_external_id
+      id = params[:query] || params[:id]
+
+      scope = User.includes(:participant_profiles, :participant_identities)
+      @user = scope
+      .where(id: id)
+      .or(scope.where(participant_identities: { external_identifier: id }))
+      .first
```
tonyheadford, line 28: If the users you're interested in will always have at least one profile they will have an identity record so you don't need to use `user_id` just `external_identifier` - also this might help https://github.com/DFE-Digital/early-careers-framework/blob/8ab57a8d5bf1caf2ac45a09d65fabe4641305f67/app/services/identity.rb

tonyheadford (approved):

app/controllers/finance/participants_controller.rb
jamesrose: @tonyheadford I wanted to be able to search by the internal user ID or the ID we display to providers. Thanks for the helper, though — that tidies it up!

PR #1712, opened 2022-02-09
[CPDLP-999] - Address pull request comments

StupidCodeFactory:
## Ticket and context

Ticket:

1. fix double link to current statement
image

2. Fix NPQ payment breakdown
3. minor template tweaks - thanks @ericaporter!

Of that back of this work, we're in the process of creating tickets to address the future NPQ m2-release (early march) in the next sprint. That will also address m3-release and fully automate the process for the next m_x_-releases


app/services/finance/npq/participant_eligible_and_payable_aggregator.rb
```diff
@@ -15,18 +15,18 @@ def self.aggregation_types
...
+        self.statement         = statement
+        self.cpd_lead_provider = statement.cpd_lead_provider
+        self.recorder          = recorder
+        self.course_identifier = course_identifier
       end
 
       def aggregate(aggregation_type:, event_type:)
         scope = recorder.public_send(self.class.aggregation_types[event_type][aggregation_type], cpd_lead_provider, course_identifier)
-        scope = scope.where(statement_id: statement.id)
+        scope = scope.where(statement_id: statement.id) unless statement.open?
```
asmega, line 29: if you plan to not to use declarations against this statement till after the payment date. this coupled with fact with another change in this PR to use `ParticipantEligibleAndPayableAggregator`. will mean that the statement will change in value after the cutoff but before the payment data. as it will pick up on new eligible declarations that are not assigned to this statement.

StupidCodeFactory, line 29: Edit: I accidentally delete @asmega's comment: 
```
if you plan to not to use declarations against this statement till after the payment date. this coupled with another change in this PR to use ParticipantEligibleAndPayableAggregator. will mean that the statement will change in value after the cutoff but before the payment date. as it will continue to pick up on new eligible declarations that are not assigned to this statement that come in
```

app/controllers/finance/npq/course_payment_breakdowns_controller.rb
```diff
@@ -6,20 +6,12 @@ class CoursePaymentBreakdownsController < BaseController
...
-        if @statement.name == "January 2022"
-          @statement.id = nil
-        end
-
         @npq_course        = NPQCourse.find_by!(identifier: params[:id])
+        @statement         = @cpd_lead_provider.npq_lead_provider.statements.find(params[:statement_id])
         @breakdown         = Finance::NPQ::CalculationOrchestrator.new(
           statement: @statement,
           contract: @npq_lead_provider.npq_contracts.find_by!(course_identifier: params[:id]),
+          aggregator: ParticipantEligibleAndPayableAggregator,
```
asmega, line 14: only using `ParticipantEligibleAndPayableAggregator` here will cause several bugs:
- when the declarations become `paid`, those declarations will drop out of scope and the output fees fees will drop to zero
- when viewing this page for a statement that has declarations assigned to it eg after payment freeze, it will ignore all those assigned declarations and therefore show incorrect values

StupidCodeFactory: @asmega we'll address this in the next sprint, Conor and I have created the stories and we'll refine them tomorrow

app/controllers/finance/ecf/statements_controller.rb
```diff
@@ -9,18 +9,9 @@ def show
         @ecf_lead_provider = lead_provider_scope.find(params[:payment_breakdown_id])
         @cpd_lead_provider = @ecf_lead_provider.cpd_lead_provider
 
-        @statements = Finance::Statement::ECF.where(cpd_lead_provider: @cpd_lead_provider).order(deadline_date: :desc).to_a
-        @statements.prepend Finance::Statement::ECF.new(cpd_lead_provider: @cpd_lead_provider, name: "Current")
```
asmega, line 13: Maybe it's intended but should be noted, that removing this will remove the current statement. I don't see anywhere an equivalent of this. So this will result in one fewer statement in the ECF payment breakdown pages

StupidCodeFactory: @asmega it is in here 
https://github.com/DFE-Digital/early-careers-framework/pull/1712/files#diff-ea0aa868734a91ebb8c48393d7521f7823427657f792995bb997677e9de41022R12

PR #1717, opened 2022-02-09
Allow finance users to drilldown by NPQ application ID

jamesrose:
Allow finance users to drilldown by NPQ application ID

This area of the service is quickly becoming very useful — more than just the temporary page that it was initially. I'm going to pad out the testing of this area before merging this.


jamesrose:
Waiting for https://github.com/DFE-Digital/early-careers-framework/pull/1716/files to be merged so that I can re-use some of the test setup.

PR #1720, opened 2022-02-10
[CPDLP-962] Able to void payable declarations

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-962

- Providers can void payable declarations
- Voided payables change state to `voided`
- These voided declarations continue to remain attached to a statement
- Therefore these voided declaration will be excluded from the statement during aggregation
- Which means the frozen statement will drop in value
- When viewing a statement. There is now a link to see the voided declarations that are associated to the statement
- The calculation orchestrators now take an instance of an aggregator instead of a class. This means it is now possible to inject a `recorder` (a fancy name for a `scope`)

### Code quality checks

### HTML Checks

### Gotchas

- For a given statement in play
- Assign some payable declarations to it if there aren't any
- Note the value of the statement
- Should be able to void a payable declaration via the api
- Value of statement should drop
- When viewing the statement at the bottom should be able to click through to the declaration(s) that have been voided

## External API changes

- Kind of although I don't think it needs documenting


spec/features/finance/payment_breakdown_spec.rb
```diff
@@ -45,13 +46,24 @@
     then_percy_should_be_sent_a_snapshot_named("Contract breakdown for an ECF provider")
 
     when_i_click_on("Back")
-    when_i_click_on("November 2021")
     then_the_page_should_be_accessible
-    then_percy_should_be_sent_a_snapshot_named("Payment breakdown for an ECF provider (latest)")
+    and_percy_should_be_sent_a_snapshot_named("Payment breakdown for an ECF provider (latest)")
+
+    when_i_click_on("November 2021")
```
StupidCodeFactory, line 52:
```suggestion
    click_link("November 2021")
```

spec/features/finance/payment_breakdown_spec.rb
```diff
@@ -45,13 +46,24 @@
...
 
 private
 
+  def then_i_see_voided_declarations
+    first("table") do
+      expect(page).to have_css("tr", count: 3)
+    end
+  end
+
   def when_i_click_on(string)
```
StupidCodeFactory, line 67: ✂️

spec/features/finance/payment_breakdown_spec.rb
```diff
@@ -91,6 +103,13 @@ def and_multiple_declarations_are_submitted
     multiple_retained_declarations_are_submitted_jan_statement
   end
 
+  def and_voided_payable_declarations_are_submitted
+    participant_profiles = create_list(:ect_participant_profile, 2, school_cohort: school_cohort, cohort: cohort, sparsity_uplift: true)
+    participant_profiles.map { |participant| ParticipantProfileState.create!(participant_profile: participant) }
+    participant_profiles.map { |participant| ECFParticipantEligibility.create!(participant_profile_id: participant.id).eligible_status! }
+    participant_profiles.map { |participant| create_voided_declarations_nov(participant) }
+  end
```
StupidCodeFactory, line 111: Should probably use `EarlyCareerTeachers::Create` for a realistic test setup. 
Probably would not be more concise as the you'd have to setup a school on the school pupil premium and so on... But it would be worth investing the time to get some school factories that allow for this.

spec/features/finance/npq/course_payment_breakdown_spec.rb
```diff
@@ -182,11 +204,11 @@ def then_i_should_see_correct_breakdown_summary(npq_lead_provider, npq_contract)
...
       expect(page).to have_content(npq_contract.recruitment_target)
       expect(page).to have_content("Current participants")
-      expect(page).to have_content(ParticipantDeclaration::NPQ.neither_paid_nor_voided_lead_provider_and_course(npq_lead_provider, npq_contract.course_identifier).count)
+      expect(page).to have_content(ParticipantDeclaration::NPQ.where(statement: statement, cpd_lead_provider: cpd_lead_provider, course_identifier: npq_contract.course_identifier, state: "payable").count)
       expect(page).to have_content("Total paid")
-      expect(page).to have_content(expected_total_paid(npq_contract))
+      expect(page).to have_content(ParticipantDeclaration::NPQ.where(statement: statement, cpd_lead_provider: cpd_lead_provider, course_identifier: npq_contract.course_identifier, state: "payable").count)
       expect(page).to have_content("Total not paid")
-      expect(page).to have_content(ParticipantDeclaration::NPQ.submitted_for_lead_provider_and_course(npq_lead_provider, npq_contract.course_identifier).count)
+      # TODO: to be implemented later
```
StupidCodeFactory, line 211: @asmega what's to implement later? Would you be make this a bit more descriptive please?

app/models/finance/statement/ecf.rb
```diff
@@ -15,9 +15,16 @@ def cache_original_value!
...
       statement: self,
     )
   end
+
+  def aggregator
+    Finance::ECF::ParticipantAggregator.new(
+      statement: self,
+      recorder: ParticipantDeclaration::ECF.where.not(state: %w[voided]),
+    )
+  end
```
StupidCodeFactory, line 29: really not a big fan of coupling the database with the business logic, passing `self` is a bit of a code smell (in both orchestrator and aggregator). 
I'd rathe keep the instantiation higher up at the controller level to avoid this two ways dependencies.

spec/features/finance/npq/course_payment_breakdown_spec.rb
asmega: This is referring the previous line `Total not paid` which is not implemented as it is not currently possible

app/models/finance/statement/ecf.rb
asmega: We can extract this out to a service class and delegate. I've not done that here as the payment breakdown classes are still in flux and would rather have something throw away here.

spec/features/finance/payment_breakdown_spec.rb
asmega: I'm jumping to on the bandwagon of what was already here. I don't think it's worth the effort tidying. I think in the long term all these tests should be ripped out. The view should be getting some kind of view object that we can stub and test. We should not be doing huge and complex data setups just to test a view.

StupidCodeFactory: Well that's sort of our "end to end" test. So a realistic test setup provides much great assurance the test is valuable. Otherwise you just have a test that you think might work if your setup mirrors the production code. But we won't have this screen for long, might point was just to re-use production code for test setup which is far more reliable than having hand crafted test setup which intrinsic correctness will drift as the production code evolves.

PR #1721, opened 2022-02-11
[CPDLP-912] Changing schedule invalid for course returns an error

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-912

- Changing schedule not valid for the course now returns a validation error
- also introduced schedules only permitted for mentors
- "Replacement" eg `ecf-replacement-april` schedules are for mentors only. This data change will happen after deployment

### Code quality checks

### HTML Checks

### Gotchas

- Find or create an ECT participant
- Change their schedule to a replacement schedule
- Operation should fail, relevant error message returned
- Find or create a Mentor participant
- Change their schedule to a replacement schedule
- Operation should succeed

## External API changes

- Probably nothing that needs documenting

### DIFF
--- a/spec/services/participants/change_schedule/base_spec.rb
+++ b/spec/services/participants/change_schedule/base_spec.rb
@@ -50,12 +50,7 @@ RSpec.describe Participants::ChangeSchedule::Base do
       let(:user) { profile.user }
       let(:profile) { create(:mentor_participant_profile) }
       let(:schedule) do
-        create(
-          :schedule,
-          type: "Finance::Schedule::ECFMentor",
-          name: "Schedule for mentors only",
-          schedule_identifier: "ecf-replacement-april",
-        )
+        create(:ecf_mentor_schedule)
       end
 
       subject do
@@ -76,12 +71,7 @@ RSpec.describe Participants::ChangeSchedule::Base do
       let(:user) { profile.user }
       let(:profile) { create(:ect_participant_profile) }
       let(:schedule) do
-        create(
-          :schedule,
-          type: "Finance::Schedule::ECFMentor",
-          name: "Schedule for mentors only",
-          schedule_identifier: "ecf-replacement-april",
-        )
+        create(:ecf_mentor_schedule)
       end
 
       subject do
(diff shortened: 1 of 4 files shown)


app/models/finance/schedule/npq_leadership.rb
```diff
@@ -11,6 +11,10 @@ class NPQLeadership < Schedule
         npq-executive-leadership
       ].freeze
 
+      def self.permitted_course_identifiers
+        IDENTIFIERS
+      end
```
StupidCodeFactory, line 16: So we either move everything to class methods or constants, not both. For this sort of things I prefer constants.

spec/services/participants/change_schedule/base_spec.rb
```diff
@@ -45,6 +45,58 @@ def matches_lead_provider?
...
       end
     end
+
+    context "when changing to schedule suitable for the course" do
+      let(:user) { profile.user }
+      let(:profile) { create(:mentor_participant_profile) }
+      let(:schedule) do
+        create(
+          :schedule,
+          type: "Finance::Schedule::ECFMentor",
```
StupidCodeFactory, line 55:
```
create(:ecf_mentor_schedule, #...)
```

StupidCodeFactory (commented):
I think we're missing a spec for ECFMentor and ECT, would you mind adding those to `spec/services/participants/change_schedule/ecf_spec.rb`

asmega:
> I think we're missing a spec for ECFMentor and ECT, would you mind adding those to `spec/services/participants/change_schedule/ecf_spec.rb`

But this wouldn't increase coverage, as no extra code has been added to `app/services/participants/change_schedule/ecf.rb`. All the extra code has been added the base class which has been covered. I don't see why I'd then copy and paste the same tests for every single subclass?

spec/services/participants/change_schedule/base_spec.rb
asmega: I've just pushed a commit to swap this out

app/models/finance/schedule/npq_leadership.rb
asmega: These are now all constants

StupidCodeFactory (approved):

PR #1731, opened 2022-02-14
[CPDLP-1013] - Move Sidekiq jobs outsides of transactions

StupidCodeFactory:
## Ticket and context

Whilst trying to see a lot of participant record, wrapped the seed script in a transactions for speed purposes mainly but also to ensure a clean state of the database if the seed script failed. 

This created a bunch of jobs that failed because the transaction would take a while before being committed. Causing the jobs to failed and being retried. Effectively spamming Sidekiq.

[Ticket:](https://dfedigital.atlassian.net/browse/CPDLP-1013)


FriedSock (approved):

app/services/early_career_teachers/create.rb
```diff
@@ -17,22 +17,20 @@ def call
         end
 
         ParticipantProfile::ECT.create!({
-          teacher_profile: teacher_profile,
-          schedule: Finance::Schedule::ECF.default,
-          participant_identity: Identity::Create.call(user: user),
-        }.merge(ect_attributes)) do |profile|
+                                          teacher_profile: teacher_profile,
```
FriedSock, line 20: Now that we have removed after_save for `ParticipantProfile::ECF`. Will update_analytics still run when this transaction completes? Or do we need to change the transaction this is inside to be `ParticipantProfile::ECT.transaction do`?

FriedSock: They appear to be the same method so I would guess it will work the same either way:
```
[1] pry(main)> $ User.transaction

From: /Users/jackbracewell/Code/early-careers-framework/.gem/ruby/2.7.4/gems/activerecord-6.1.4.4/lib/active_r
ecord/transactions.rb:207:
Owner: ActiveRecord::Transactions::ClassMethods
Visibility: public
Signature: transaction(?, &block)
Number of lines: 3

def transaction(**options, &block)
  connection.transaction(**options, &block)
end
[2] pry(main)> $ ActiveRecord::Base.transaction

From: /Users/jackbracewell/Code/early-careers-framework/.gem/ruby/2.7.4/gems/activerecord-6.1.4.4/lib/active_r
ecord/transactions.rb:207:
Owner: ActiveRecord::Transactions::ClassMethods
Visibility: public
Signature: transaction(?, &block)
Number of lines: 3

def transaction(**options, &block)
  connection.transaction(**options, &block)
end
```

PR #1737, opened 2022-02-15
[CPDLP-1001] Validate declaration type for schedule

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1001

- It was assumed `declaration_type` is tied to the the groupings ECF/NPQ
- However this is not quite the case and can be gathered from the schedule they are on
- Remove the hardcoded declaration type from `RecordDeclarations` as we now want to delegate this to the schedule
- Therefore we can now validate the incoming `declaration_type` against the current schedule

### Code quality checks

### HTML Checks

### Gotchas

- Attempt to submit a `retained-7` declaration for any participant as this milestones does not exist
- Should see a relevant validation error message returned
- Another test is to find/create a participant on one of the `npq-specialist` schedules
- Attempt to send a `retained-2` declaration which does not exist
- Should see a relevant validation error message returned

## External API changes

- Nothing that needs to be documented

### DIFF
--- a/app/services/record_declarations/ecf.rb
+++ b/app/services/record_declarations/ecf.rb
@@ -4,13 +4,6 @@ module RecordDeclarations
   module ECF
     extend ActiveSupport::Concern
 
-    STARTED      = "started"
-    COMPLETED    = "completed"
-    RETAINED_ONE = "retained-1"
-    RETAINED_TWO = "retained-2"
-    RETAINED_THREE = "retained-3"
-    RETAINED_FOUR = "retained-4"
-
     included do
       extend ECFClassMethods
     end
@@ -18,10 +11,6 @@ module RecordDeclarations
     module ECFClassMethods
       def declaration_model
         ParticipantDeclaration::ECF
-      end
-
-      def valid_declaration_types
-        [STARTED, COMPLETED, RETAINED_ONE, RETAINED_TWO, RETAINED_THREE, RETAINED_FOUR]
       end
     end
   end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -36,7 +36,7 @@ en:
   invalid_declaration_type: "The property '#/declaration_type' must be available for '#/course_identifier'"
   invalid_declaration_date: "The property '#/declaration_date' must be a valid RCF3339 date"
   future_declaration_date: "The property '#/declaration_date' can not declare a future date"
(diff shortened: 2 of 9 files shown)


app/services/record_declarations/ecf.rb
```diff
@@ -4,13 +4,6 @@ module RecordDeclarations
...
     extend ActiveSupport::Concern
 
-    STARTED      = "started"
-    COMPLETED    = "completed"
-    RETAINED_ONE = "retained-1"
-    RETAINED_TWO = "retained-2"
-    RETAINED_THREE = "retained-3"
-    RETAINED_FOUR = "retained-4"
-
     included do
```
ericaporter, line 7: Really like that this has gone!

spec/services/record_declarations/base_spec.rb
```diff
@@ -132,12 +133,36 @@ def matches_lead_provider?
...
+            declaration_type: "retained-4",
+            participant_id: user.id,
+          },
+        )
+      end
+
+      it "returns an error" do
+        expect { subject.call }.to raise_error(ActionController::ParameterMissing, /Declaration type does not exist for this schedule/)
+      end
     end
```
ericaporter, line 166: What happens if the declaration type is missing, out of interest?

config/locales/en.yml
```diff
@@ -36,6 +36,7 @@ en:
   invalid_declaration_type: "The property '#/declaration_type' must be available for '#/course_identifier'"
   invalid_declaration_date: "The property '#/declaration_date' must be a valid RCF3339 date"
   future_declaration_date: "The property '#/declaration_date' can not declare a future date"
+  mismatch_declaration_type_for_schedule: "Declaration type does not exist for this schedule"
```
StupidCodeFactory, line 39:
```suggestion
  mismatch_declaration_type_for_schedule: "#/declaration_type does not exist for this schedule"
```

spec/services/record_declarations/base_spec.rb
asmega: it's covered by https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/services/record_declarations/base.rb#L18 on mainline

config/locales/en.yml
asmega: I've tweaked this now

StupidCodeFactory (approved):

PR #1739, opened 2022-02-17
[CPDLP-962] Can void not last declaration

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-962

- Previously limiting voids to last declaration only may have made sense
- However now this does not seem a reasonable constraint to have
- So has been removed

### Code quality checks

### HTML Checks

### Gotchas

- For a given statement in play
- Assign some payable declarations to it if there aren't any
- Note the value of the statement
- Should be able to void a payable declaration via the api
- Value of statement should drop
- When viewing the statement at the bottom should be able to click through to the declaration(s) that have been voided

## External API changes

- Nothing that needs documenting


ericaporter (approved):

StupidCodeFactory (approved):
🏆

ericaporter (approved):

PR #1746, opened 2022-02-18
[CPDLP-1040] Stop payable declarations reverting to eligible

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1040

- This fixes a bug where by a provider could submit a declaration for an existing payable declaration causing it to revert back to eligible
- All incoming declarations now create a new record regardless if another similar one exists or not
- We now create a DeclarationState record when a ParticipantDeclation is created as an ActiveRecord callback

### Code quality checks

### HTML Checks

### Gotchas

- Find a declaration in payable or paid state
- Submit the same declaration again as the provider
- NOTE: timestamps MUST be identical otherwise it will be considered as a different declaration and you get some kind of error which doesn't test this edge case
- It will create another declaration that is ineligible
- The original declaration will remain in original state of payable or paid

## External API changes

- Nothing that needs documenting

### DIFF
--- a/app/models/participant_declaration.rb
+++ b/app/models/participant_declaration.rb
@@ -81,7 +81,7 @@ class ParticipantDeclaration < ApplicationRecord
   scope :paid_npqs_for_lead_provider, ->(lead_provider) { paid_for_lead_provider(lead_provider).npq }
   scope :paid_uplift_for_lead_provider, ->(lead_provider) { paid_for_lead_provider(lead_provider).uplift }
 
-  after_create :create_initial_declaration_state
+  before_create :build_initial_declaration_state
 
   # TODO: Voiding paid should trigger clawbacks, but currently OOS
   def voidable?
@@ -131,8 +131,8 @@ class ParticipantDeclaration < ApplicationRecord
 
 private
 
-  def create_initial_declaration_state
-    declaration_states.create(state: state)
+  def build_initial_declaration_state
+    declaration_states.build(state: state)
   end
 end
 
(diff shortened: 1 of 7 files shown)


app/models/participant_declaration.rb
```diff
@@ -81,6 +81,8 @@ class ParticipantDeclaration < ApplicationRecord
   scope :paid_npqs_for_lead_provider, ->(lead_provider) { paid_for_lead_provider(lead_provider).npq }
   scope :paid_uplift_for_lead_provider, ->(lead_provider) { paid_for_lead_provider(lead_provider).uplift }
 
+  after_create :create_initial_declaration_state
```
jamesrose, line 84: Do you want this build as part of the create? (e.g., `before_create` and `declaration_states.build(...)`)

asmega: - i'm not sure `before_create` would work as it needs to reference the persisted record.
- `build` i think could work but where would you stick it? and also what would you set the state to when you call `build` as it could be `submitted` or `ineligible`

jamesrose:
```
  before_create :build_initial_declaration_state
  
  private
  
  def build_initial_declaration_state
    declaration_states.build(state: state)
  end
```

I _think_ this should work if `inverse_of` is correctly set on each side of the association. The save on the declaration will save the state.

asmega: Worked without `inverse_of `. Saving the parent seems to also save the child

jamesrose (approved):

PR #1759, opened 2022-02-25
Storing NPQApplication employment answers

BroiSatse:
npq-application is now also sending us employment details which we need to store: https://github.com/DFE-Digital/npq-registration/pull/330


asmega:
should there be an accompanying documentation changes within this PR?

tonyheadford (commented):
Could we have spec(s) for this?

tonyheadford (approved):

PR #1770, opened 2022-02-28
Changing school scenario tests

mattonfoot:
## Context

A feature branch to explore end-to-end tests checking the inputs form one user and the outputs for all other users.

For a full list of potential testing scenarios identified to be covered see <https://docs.google.com/document/d/13fgyZdADb8B2bFxzmDePvJTQloqgM5eav8ZL7On8XpQ/edit?usp=sharing>

To run just the added tests use the command

`bundle exec rspec ./spec/features/transfering_a_participant --format documentation`

### test scenarios catalog

Output in test run:
https://github.com/DFE-Digital/early-careers-framework/runs/5534683013?check_suite_focus=true#step:8:16

## Tech review
- [ ] The purpose of the tests make sense to the reader without having to read the detailed code
- [ ] Tests are using the right interfaces ( APIs, UIs, object methods etc. )


spec/factories/provider_relationships.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :provider_relationship do
+    cohort { Cohort.find_or_create_by!(start_year: 2021) }
```
StupidCodeFactory, line 5: Could we have something dynamic like `Date.current.month > 8 ?  Date.current.year : Date.current.year + 1`? 
this is assuming the start of the next cohort is each September

spec/features/after_transfering_an_ect/transfer_participant_steps.rb
```diff
@@ -0,0 +1,140 @@
...
+  def when_sit_takes_on_the_participant(sit, participant)
+    school = sit.schools.first
+    school_cohort = school.school_cohorts.first
+
+    participant.update!(school_cohort: school_cohort)
+    participant.teacher_profile.update!(school: school)
+  end
+
+  def then_participant_can_be_seen_by_lead_provider(lead_provider, lead_provider_token, participant)
+    assert_equal true, lead_provider.ecf_participant_profiles.include?(participant)
```
StupidCodeFactory, line 74: Would be idea to avoid a mix of minitest/rspec assertions 
Could include  `include RSpec::Matchers` and then swap `assert_*` to the their RSpec equivalent: 
```suggestion
    expect(lead_provider.ecf_participant_profiles).to include(participant
```

mattonfoot: Changed in new commit

spec/factories/provider_relationships.rb
mattonfoot: Follow the school_cohort initialiser so we can set it or just use the current

asmega:
From the very top of you description, a question:

```
after Transferring the only ECT from another school onto a FIP
  from a FIP with a different provider, before a started declaration has occurred
    the ECT cannot be seen by another Lead Provider
    the ECT can be seen by the new SIT
    the ECT cannot be seen by the original Lead Provider # They can be but nulled out?
    the declarations cannot be seen by the original Lead Provider # I think this is misleading. The declarations can be seen but they don't have any
```

mattonfoot:
> From the very top of you description, a question:
> 
> ```
> after Transferring the only ECT from another school onto a FIP
>   from a FIP with a different provider, before a started declaration has occurred
>     the ECT cannot be seen by another Lead Provider
>     the ECT can be seen by the new SIT
>     the ECT cannot be seen by the original Lead Provider # They can be but nulled out?
>     the declarations cannot be seen by the original Lead Provider # I think this is misleading. The declarations can be seen but they don't have any
> ```

very good point I am currently only checking the ID is present and had not got around to looking for the details and checking they are obfuscated. On the other one I have started using the `skip` option on scenarios so that we can turn them on when we implement them allowing us to record future expectations and identify when the service does not support them yet

spec/support/matchers/finance_portal/can_find_participant_training_status_in_finance_portal.rb
```diff
@@ -0,0 +1,48 @@
...
+        @error = :id unless drilldown.can_see_participant?(user.id)
+        @error = :status unless drilldown.can_see_training_status?(training_status)
+
+        sign_out
+
+        if @error.nil?
+          true
+        else
+          false
+        end
```
jamesrose, line 32:
```suggestion
        @error.nil?
```

spec/support/matchers/finance_portal/can_find_training_declarations_in_finance_portal.rb
```diff
@@ -0,0 +1,55 @@
...
+          @error = :declarations unless drilldown.can_see_declaration?(declaration_type, "ect-induction", "payable")
+        end
+
+        sign_out
+
+        if @error.nil?
+          true
+        else
+          false
+        end
```
jamesrose, line 35:
```suggestion
        @error.nil?
```

spec/support/matchers/finance_portal/can_find_participant_lead_provider_in_finance_portal.rb
```diff
@@ -0,0 +1,49 @@
...
+        @error = :id unless drilldown.can_see_participant?(user.id)
+        @error = :lead_provider unless drilldown.can_see_lead_provider_urn?(lead_provider_name)
+
+        sign_out
+
+        if @error.nil?
+          true
+        else
+          false
+        end
```
jamesrose, line 33:
```suggestion
        @error.nil?
```

spec/support/matchers/finance_portal/can_find_participant_status_in_finance_portal.rb
```diff
@@ -0,0 +1,48 @@
...
+        @error = :id unless drilldown.can_see_participant?(user.id)
+        @error = :status unless drilldown.can_see_status?(status)
+
+        sign_out
+
+        if @error.nil?
+          true
+        else
+          false
+        end
```
jamesrose, line 32:
```suggestion
        @error.nil?
```

spec/support/features/pages/admin_support_portal.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Pages
+  class AdminSupportPortal
+    include Capybara::DSL
+
+    def view_participant_list
```
asmega, line 7: not important but random opinion. i'd probably delete this class and choose to inline. it's not used elsewhere. the method also feels like it conflated trying to both interact with the page and do the page object (which is not quite a page object)

spec/support/matchers/finance_portal/can_find_participant_lead_provider_in_finance_portal.rb
asmega: just to expand. `nil` and `false` are falsey everything else is truthy

spec/support/matchers/lead_provider_api/can_retrieve_participant_details_from_the_ecf_participants_endpoint.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module Support
+  module CanRetrieveParticipantDetailsFromTheEcfParticipantsEndpoint
+    extend RSpec::Matchers::DSL
+
+    RSpec::Matchers.define :be_able_to_retrieve_the_details_of_the_participant_from_the_ecf_participants_endpoint do |participant_name, participant_type, options = {}|
+      match do |lead_provider_name|
+        user = User.find_by(full_name: participant_name)
+        raise "Could not find User for #{participant_name}" if user.nil?
```
asmega, line 10: this seems to work so leave it. but it looks you've written your own little pattern for assertions thru exceptions as opposed to using patterns defined by rspec. see https://relishapp.com/rspec/rspec-expectations/docs/custom-matchers/define-a-custom-matcher

asmega: oh i see, you've done above on line 8 then stuck it in a huge block

spec/spec_helper.rb
```diff
@@ -63,6 +63,7 @@
   # aliases for `it`, `describe`, and `context` that include `:focus`
   # metadata: `fit`, `fdescribe` and `fcontext`, respectively.
   config.filter_run_when_matching :focus
+  config.filter_run_excluding end_to_end_scenario: true
```
asmega, line 66: 👍

asmega (approved):
Besides the fact theres loads here. i don't see any problems with it. I'd rather have the extra coverage. Nice one Matt!

PR #1781, opened 2022-03-03
BigQuery should accept rows that don't match schema

jamesrose:
## Ticket and context

Sometimes we might change the schema on declarations but forget to update the BigQuery schema. Previously this would fail the BigQuery participant stream silently because we push _all_ attributes on that model. 

This change accepts rows that contain values that do not match the schema. The unknown values are ignored.


asmega (approved):

asmega:
nice! probably a few other places that could do with borrowing this

PR #1818, opened 2022-03-17
[CPDLP-1121] Nullify when training_status is withdrawn

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1121

- only email is now nullified
- all other fields are left intact
- this affects participant serializers
- this is so when provider withdraws they no see their details
 
### Code quality checks

### HTML Checks

### Gotchas

- Find an ECF user that is present via `/ecf/participants` endpoint
- Withdraw them either `status` or `training_status` set to `withdrawn`
- `email` should now be nullified via same api call
- Perform the same tests for CSV version
- Perform the same tests for NPQ + extra tests
- Additional feature for NPQ only. If you have 2 enrollments with the same provider, you must withdraw both enrollments in order for the email to be nullified

## External API changes

- Yes, to some degree. Some fields that could be previously nullified can no longer be nullified. Docs have been updated accordingly

### DIFF
--- a/app/services/npq/is_user_withdrawn.rb
+++ b/app/services/npq/is_user_withdrawn.rb
@@ -17,7 +17,6 @@ module NPQ
 
     def active_npq_profiles
       scope = user.npq_profiles
-      scope = scope.includes(npq_application: [:npq_course])
       scope = scope.joins(npq_application: { npq_lead_provider: [:cpd_lead_provider] })
       scope = scope.where(npq_applications: { npq_lead_providers: { cpd_lead_provider: cpd_lead_provider } })
       scope = scope.where.not(training_status: "withdrawn")
--- a/app/services/npq/is_user_withdrawn.rb
+++ b/app/services/npq/is_user_withdrawn.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module NPQ
+  class IsUserWithdrawn
+    attr_reader :user, :cpd_lead_provider
+
+    def initialize(user:, cpd_lead_provider:)
+      @user = user
+      @cpd_lead_provider = cpd_lead_provider
+    end
+
+    def call
+      active_npq_profiles.none?
+    end
+
+  private
+
+    def active_npq_profiles
+      scope = user.npq_profiles
+      scope = scope.includes(npq_application: [:npq_course])
+      scope = scope.joins(npq_application: { npq_lead_provider: [:cpd_lead_provider] })
+      scope = scope.where(npq_applications: { npq_lead_providers: { cpd_lead_provider: cpd_lead_provider } })
+      scope = scope.where.not(training_status: "withdrawn")
+      scope.where.not(status: "withdrawn")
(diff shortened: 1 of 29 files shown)


app/services/npq/is_user_withdrawn.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def call
+      active_npq_profiles.none?
+    end
+
+  private
+
+    def active_npq_profiles
+      scope = user.npq_profiles
+      scope = scope.includes(npq_application: [:npq_course])
```
StupidCodeFactory, line 20: Do we actually need this line? no `where` clause references this table. I think the `join` at the line below should be sufficient. 
We also don't need this include as we are not going to load them in memory. This query will just generate a "SELECT 1 from ..... WHERE ..."

spec/serializers/api/v2/participant_serializer_spec.rb
```diff
@@ -0,0 +1,325 @@
...
+
+        context "when the participant is eligible for funding" do
+          before do
+            ECFParticipantEligibility.create!(participant_profile: ect_profile).eligible_status!
+            ECFParticipantEligibility.create!(participant_profile: mentor_profile).eligible_status!
+          end
+
+          it "outputs correctly formatted serialized Mentors" do
+            expected_json_string = "{\"data\":{\"id\":\"#{mentor.id}\",\"type\":\"participant\",\"attributes\":{\"email\":\"#{mentor.email}\",\"full_name\":\"#{mentor.full_name}\",\"mentor_id\":null,\"school_urn\":\"#{mentor.mentor_profile.school.urn}\",\"participant_type\":\"mentor\",\"cohort\":\"#{mentor_cohort.start_year}\",\"status\":\"active\",\"teacher_reference_number\":\"#{mentor.teacher_profile.trn}\",\"teacher_reference_number_validated\":true,\"eligible_for_funding\":true,\"pupil_premium_uplift\":false,\"sparsity_uplift\":false,\"training_status\":\"active\",\"schedule_identifier\":\"ecf-standard-september\",\"updated_at\":\"#{mentor.updated_at.rfc3339}\"}}}"
+            expect(ParticipantSerializer.new(mentor_profile).serializable_hash.to_json).to eq expected_json_string
```
StupidCodeFactory, line 24: you don't really need to check against a JSON string. Just assert against the #serializable_hash. That will make things a tad easier to read

asmega: Yeah i agree. It's copy/pasted so i'm not changing it as I've not touched it. next person to touch it can swap to a hash, i've done this for the tests I have modified

app/services/npq/is_user_withdrawn.rb
asmega: good spot, i must of extracted this from elsewhere that needed it

mattonfoot (approved):

PR #1862, opened 2022-03-28
[CPDLP-1128] Change schedule updates induction record

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1128

- There is now a schedule fields on the induction record
- When a provider changes schedule we now also update the relevant induction record
- Once we swap over the schedule on the profile will become defunct
- However I am leaving the profile schedule present as there will be existing code that utilises it therefore maintain backwards compatibility

### Code quality checks

### HTML Checks

### Gotchas

- assuming you have a participant with an induction record otherwise no induction record will get updated
- change their schedule
- should also change the schedule on the induction record on top of the profile schedule
- NPQ should not be affected

## External API changes

- None

### DIFF
--- a/spec/services/participants/change_schedule/early_career_teacher_spec.rb
+++ b/spec/services/participants/change_schedule/early_career_teacher_spec.rb
@@ -3,4 +3,125 @@
 require "rails_helper"
 
 RSpec.describe Participants::ChangeSchedule::EarlyCareerTeacher do
+  describe "validations" do
+    context "when null schedule_identifier given" do
+      let(:cpd_lead_provider) { create(:cpd_lead_provider, :with_lead_provider) }
+      let(:lead_provider) { cpd_lead_provider.lead_provider }
+      let(:user) { profile.user }
+      let(:profile) { create(:ect_participant_profile, school_cohort: school_cohort) }
+      let(:school_cohort) { create(:school_cohort) }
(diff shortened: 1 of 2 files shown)


spec/services/participants/change_schedule/early_career_teacher_spec.rb
```diff
@@ -3,4 +3,150 @@
...
+    let(:user) { profile.user }
+    let(:profile) { create(:ect_participant_profile, school_cohort: school_cohort) }
+    let(:school_cohort) { create(:school_cohort) }
+    let(:cohort) { school_cohort.cohort }
+    let(:school) { school_cohort.school }
+    let!(:partnership) { create(:partnership, school: school, lead_provider: lead_provider, cohort: school_cohort.cohort) }
+    let(:schedule) { create(:ecf_mentor_schedule) }
+
+    let(:schedule) do
+      Finance::Schedule.create!(
```
StupidCodeFactory, line 121: Suggestion: use `FactoryBot` to be consistent with the rest of the code

StupidCodeFactory (approved):

spec/services/participants/change_schedule/early_career_teacher_spec.rb
asmega: done

PR #1863, opened 2022-03-28
[CPDLP-1170] Refactor deferring tests

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1170

- Refactor only
- Removes shared tests for defer services
- This increases test duplication code but far easier to make changes rather than tests shared across multiple with aren't very localised
- This will be needed as NPQ and ECF code will diverge soon with the introduction of induction records
- Noticed a bug for deferring withdrawn folk. this already exists and not in scope for this PR for fixing
- These tests now run in about a quarter of the time the original tests took due to the fact we no longer load in so many unrelated test setups due to the shared nature

### Code quality checks

### HTML Checks

### Gotchas

n/a

## External API changes

- None

### DIFF
--- a/spec/services/participants/defer/early_career_teacher_spec.rb
+++ b/spec/services/participants/defer/early_career_teacher_spec.rb
@@ -2,36 +2,103 @@
 
 require "rails_helper"
 
-require_relative "../../../shared/context/lead_provider_profiles_and_courses"
-
 RSpec.describe Participants::Defer::EarlyCareerTeacher do
-  include_context "lead provider profiles and courses"
-  let(:participant_params) do
-    {
-      cpd_lead_provider: cpd_lead_provider,
-      participant_id: ect_profile.user.id,
-      course_identifier: "ecf-induction",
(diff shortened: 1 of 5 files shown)


spec/services/participants/defer/early_career_teacher_spec.rb
```diff
@@ -2,36 +2,103 @@
...
+    context "when status is withdrawn" do
+      before do
+        ParticipantProfileState.create!(participant_profile: profile, state: "withdrawn")
+        profile.update!(status: "withdrawn")
+      end
+
+      xit "returns an error and does not update training_status" do
+        # TODO: there is a gap and bug here
+        # it should return a useful error
+        # but throws an error as we scope to active profiles only and therefore never find the record
```
StupidCodeFactory, line 67: Could you add a ticket for this? sounds like something we want to address (with a link to this PR for context)

StupidCodeFactory (approved):

PR #1844, opened 2022-03-23
[CPDLP-1149] - Make statements cohort sensitive

StupidCodeFactory:
## Ticket and context

[Ticket](https://dfedigital.atlassian.net/browse/CPDLP-1149)

TODO: remove optionals and foreign_key constraints in a subsequent PR **to be deployed separately**

### Code quality checks
- [ ] All commit messages are meaningful and true
- [ ] Added enough automated tests

### HTML Checks
- [ ] All new pages have automated accessibility checks
- [ ] All new pages have visual tests via Percy

### Gotchas
- [ ] All `School` queries are correctly scoped - `eligible` when they need to be

### How can someone see it it review app?
1. Click the link to review app (posted by a `github-actions` bot below)
2. Follow the steps from the ticket.

## External API changes

Does this change anything in our external APIs? If so, did you remember to update the CHANGELOG for Lead Providers? (docs/source/release-notes)

PR #1868, opened 2022-03-28
[CPDLP-1149] - make statements cohort sensitive two

StupidCodeFactory:
## Ticket and context

This is a follow up of this [PR](https://github.com/DFE-Digital/early-careers-framework/pull/1844)
Add database constrains and foreign keys

The tasks to populate the cohort for statements and contracts have already been run in all environments


app/services/importers/seed_statements.rb
```diff
@@ -9,11 +9,13 @@ def call
         statement = Finance::Statement::ECF.find_or_create_by!(
           name: statement_data.name,
           cpd_lead_provider: cpd_lead_provider,
+          cohort: Cohort.find_by(start_year: 2021),
```
asmega, line 12: you defined a private method for this. why do you not use it here?

asmega (approved):

PR #1871, opened 2022-03-29
Migrate to Shakapacker and Webpack 5

leoapost:
## Ticket and context

Ticket: N/A

Our current assets pipeline uses the deprecated Webpacker and dependabot is raising issues for both our Webpacker and Webpack versions that we can't fix unless we migrate them. Moving to Shakapacker and Webpack 5 seems to be the quickest from all the available solutions.

Changes:
- Bumps up webpack to v5.70
- Replaces webpacker with shakapacker v6.2
- Updates the webpacker.yml config
- Adds the browserslist in the package.json
- Uses a single webpack config as it doesn't seem we need a config per env
- Moves the extensions from webpacker to webpack config
- Adds webpack css and scss loaders
- Renames the webpack binstubs to webpacker
- Updates the paths of the static files in the application layout
- Updates the pre-commit hook to avoid picking node_module paths as AWS secrets
- Fixes a bug in the session specs that loads the sentry js script

## Tech review
Notes: 
- Both the `application` and `sentry` scripts will use defer, because Shakapacker currently doesn't have a solution to create script tags with some using defer and some not.
- Shakapacker has renamed the webpack binstubs to `bin/webpacker` and `bin/webpacker-dev-server`
- The assets folders remain as they were before. Source files in `app/webpacker`, compiled files in `public/packs`
- Run `rails webpacker:clobber` to remove the old packs


app/views/layouts/_application.html.erb
```diff
@@ -1,29 +1,31 @@
...
-    <%= favicon_link_tag asset_pack_path('media/images/govuk-apple-touch-icon-180x180.png'), rel: 'apple-touch-icon', type: 'image/png', size: '180x180' %>
+    <%= favicon_link_tag asset_pack_path('static/govuk-frontend/govuk/assets/images/favicon.ico') %>
+    <%= favicon_link_tag asset_pack_path('static/govuk-frontend/govuk/assets/images/govuk-mask-icon.svg'), rel: 'mask-icon', type: 'image/svg', color: "#0b0c0c" %>
+    <%= favicon_link_tag asset_pack_path('static/govuk-frontend/govuk/assets/images/govuk-apple-touch-icon.png'), rel: 'apple-touch-icon', type: 'image/png' %>
+    <%= favicon_link_tag asset_pack_path('static/govuk-frontend/govuk/assets/images/govuk-apple-touch-icon-152x152.png'), rel: 'apple-touch-icon', type: 'image/png', size: '152x152' %>
+    <%= favicon_link_tag asset_pack_path('static/govuk-frontend/govuk/assets/images/govuk-apple-touch-icon-167x167.png'), rel: 'apple-touch-icon', type: 'image/png', size: '167x167' %>
+    <%= favicon_link_tag asset_pack_path('static/govuk-frontend/govuk/assets/images/govuk-apple-touch-icon-180x180.png'), rel: 'apple-touch-icon', type: 'image/png', size: '180x180' %>
```
lazydevorg, line 26: often static assets are served by web servers instead of Rails itself.
I'm not sure whether it's our case however, if it is then I guess we should check whether this could break anything.

bin/yarn
```diff
@@ -1,16 +1,15 @@
...
 
-    normalized_path != __dir__ && File.executable?(Pathname.new(normalized_path).join('yarn'))
-  end
-
-  if executable_path
-    exec File.expand_path(Pathname.new(executable_path).join('yarn')), *ARGV
+  if yarn
+    exec yarn, *ARGV
   else
     $stderr.puts "Yarn executable was not detected in the system."
```
lazydevorg, line 14: I think it would be good to print the yarn path, if any, to help debug any issue.

lazydevorg (commented):
@leoapost this PR looks good to me.
I don't fill I have enough knowledge of Webpack to approve this PR though.
It would be better to get an additional pair of eyes 😊

bin/yarn
leoapost: This is a very simple binstub copied over by shakapacker (see the [source file](https://github.com/shakacode/shakapacker/tree/master/lib/install/bin/yarn)). 

You should be good to use it if you have yarn installed as [mentioned](https://github.com/DFE-Digital/early-careers-framework#prerequisites) in the README. If not, it will ask you to download it.

app/views/layouts/_application.html.erb
leoapost: That all it does is linking the static assets really. It will know how to build the path to the static assets, no matter how we serve them.

jamesrose (approved):

PR #1878, opened 2022-03-29
Remove deprecated payment breakdown task

jamesrose:
## Ticket and context

This Rake task was used in the very early stages of development of the payment calculators to demonstrate that they worked correctly, and to provide an easy way of viewing the information.

Now we have much better testing and UI interfaces that people can use to see better information.


asmega (approved):

asmega:
I've rebased this PR so we can get this in

PR #1889, opened 2022-03-31
[CPDLP-1170] InductionRecord#training_status updates via API

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1170

- Lead providers deferring, resuming, withdrawn now affect induction records for `training_status`
- The AR callback for syncing `training_status` has been removed
- Seeds have been updated to support this. I will continue to admit the state of the seeds is appalling and needs a re-vamp

### Code quality checks

### HTML Checks

### Gotchas

- Assuming induction records are setup
- As a lead provider defer/resume/withdraw a participant
- Relevant induction record should be updated

## External API changes

- None

### DIFF
--- a/db/seeds/test_data.rb
+++ b/db/seeds/test_data.rb
@@ -90,9 +90,14 @@ School.find_or_create_by!(urn: "000005") do |school|
   InductionCoordinatorProfile.find_or_create_by!(user: user) do |profile|
     profile.update!(schools: [school])
   end
-  SchoolCohort.find_or_create_by!(cohort: Cohort.current, school: school, induction_programme_choice: "full_induction_programme")
+  school_cohort = SchoolCohort.find_or_create_by!(cohort: Cohort.current, school: school, induction_programme_choice: "full_induction_programme")
   delivery_partner = DeliveryPartner.find_or_create_by!(name: "Test Delivery Partner")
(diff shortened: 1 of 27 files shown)


db/seeds/test_data.rb
```diff
@@ -408,81 +444,117 @@
...
 teacher_profile = TeacherProfile.find_or_create_by!(user: user)
-ParticipantProfile::Mentor.find_or_create_by!(teacher_profile: teacher_profile) do |mentor_profile|
+participant_profile = ParticipantProfile::Mentor.find_or_create_by!(teacher_profile: teacher_profile) do |mentor_profile|
   mentor_profile.school_cohort = School.find_by(urn: "000105").school_cohorts.find_by(cohort: Cohort.current)
   mentor_profile.schedule = Finance::Schedule::ECF.default
   mentor_profile.participant_identity = Identity::Create.call(user: user, origin: :ecf)
   ParticipantProfileState.find_or_create_by!(participant_profile: mentor_profile)
 end
 
+induction_programme = participant_profile.school_cohort.induction_programmes.first
```
StupidCodeFactory, line 454: Should we add an order close, just for preventive good measure? seems there will be only one at this point so maybe not necessary?

StupidCodeFactory (approved):
:shipit:

db/seeds/test_data.rb
asmega: i checked these, there only one. i ran the script 20 to makes there weren't any oddities.

PR #1911, opened 2022-04-05
Kaminari replaced by Pagy for Schools dashboard

lazydevorg:
## Ticket and context

Ticket: CTS-587

### DIFF
--- a/app/views/schools/dashboard/index.html.erb
+++ b/app/views/schools/dashboard/index.html.erb
@@ -22,4 +22,4 @@
     <% end %>
   </tbody>
 </table>
-<%= govuk_paginate @schools, @pagy %>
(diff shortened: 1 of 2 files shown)


app/views/schools/dashboard/index.html.erb
```diff
@@ -22,4 +22,4 @@
     <% end %>
   </tbody>
 </table>
-<%= govuk_paginate @schools %>
+<%= govuk_paginate @schools, @pagy %>
```
asmega, line 25: one of you other changes made it `<%==`. do you need the double equals here? or do you not need the double equals in the other place?

lazydevorg: yes it is needed 👍
nice catch

PR #1912, opened 2022-04-05
[CPDLP-1159] v1 participant endpoint uses induction records

asmega:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1159

- there was another endpoint that was run in parallel with existing fetch participant endpoint that was using induction records
- the old endpoint has been swapped out for this parallel endpoint

### Code quality checks

### HTML Checks

### Gotchas

- Assuming induction records are seeded/created
- Fetch ECF/NPQ participants as JSON/CSV
- Should see requested participants

## External API changes

- None, changes should be transparent


spec/requests/api/v1/ecf_participants_spec.rb
```diff
@@ -12,21 +12,41 @@
   let(:school_cohort) { create(:school_cohort, school: partnership.school, cohort: cohort, induction_programme_choice: "full_induction_programme") }
   let(:token) { LeadProviderApiToken.create_with_random_token!(cpd_lead_provider: cpd_lead_provider) }
   let(:bearer_token) { "Bearer #{token}" }
-  let!(:mentor_profile) { create(:mentor_participant_profile, school_cohort: school_cohort) }
+  let!(:mentor_profile) do
+    create(:mentor_participant_profile, school_cohort: school_cohort)
+      .tap { |profile| Induction::Enrol.call(participant_profile: profile, induction_programme: induction_programme) }
```
StupidCodeFactory, line 17: Should we be doing this in the factory (in a later PR as this could break more specs, may be worth a try just to assess the amount), like it is done here:
https://github.com/DFE-Digital/early-careers-framework/blob/6403cd5c25977810619ebad6c26bbb85378f17c3/spec/factories/participant_profiles.rb#L86

StupidCodeFactory (approved):

spec/requests/api/v1/ecf_participants_spec.rb
asmega: yes, that would make sense. i can do that for future ones

PR #1962, opened 2022-04-19
E2e scenario fixes

mattonfoot:
E2E scenarios should pass for FIP 2 FIP transfers

### DIFF
--- a/spec/features/scenarios/changes_of_circumstance/onboarding_a_deferred_participant_spec.rb
+++ b/spec/features/scenarios/changes_of_circumstance/onboarding_a_deferred_participant_spec.rb
@@ -64,7 +64,10 @@ RSpec.feature "Onboard a deferred participant", type: :feature, end_to_end_scena
           and_sit_at_pupil_premium_school_reported_programme "New SIT", "CIP"
         end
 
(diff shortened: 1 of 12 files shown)


spec/features/scenarios/changes_of_circumstance/onboarding_a_deferred_participant_spec.rb
```diff
@@ -92,13 +107,25 @@ def when_context(scenario)
         context "Then the Original SIT" do
           subject(:original_sit) { "Original SIT" }
 
-          it { should_not find_participant_details_in_school_induction_portal "the Participant" }
+          it do
+            should_not find_participant_details_in_school_induction_portal "the Participant",
```
asmega, line 111: don't bother changing this now but for future reference my understanding is the `should` syntax is the old way and will be removed at some point in the future. https://rspec.rubystyle.guide/#use-expect

asmega (approved):

lazydevorg (approved):

spec/features/scenarios/changes_of_circumstance/onboarding_a_deferred_participant_spec.rb
mattonfoot: changed to expected_to

PR #1965, opened 2022-04-19
[CST-583] Update the lead provider dashboard to report partnerships for the new cohort

leoapost:
## Ticket and context

Ticket: https://dfedigital.atlassian.net/browse/CST-583

We want the lead providers to view their school partnerships from multiple cohorts and allow them to begin reporting any new partnerships to DfE ahead of the start of the 2022/23 cohort.

The early registrations for the new cohort will start early May, so a feature flag is required.

Changes:
- Add the `multiple_cohorts` feature flag to enable the changes when we the new cohort opens for early registrations
- Add multiple cohort support to the lead provider dashboard by using the `cohort` param in the urls
- Scope the displayed schools, school exports and search results to the selected cohort
- Use better filenames when exporting the school partnerships
- Display the `confirm more schools` button only for the cohort that's open for registrations
- Display the `download schools` button only when there are schools for the selected cohort
- Update the nav item component to take url params under consideration when applying the css styles to the selected item

### Code quality checks

### HTML Checks

### Gotchas

## External API changes

### DIFF
--- a/app/controllers/lead_providers/report_schools/base_controller.rb
+++ b/app/controllers/lead_providers/report_schools/base_controller.rb
@@ -13,11 +13,7 @@ module LeadProviders
       def start
         clean_form!
 
-        report_schools_form.cohort_id = if FeatureFlag.active?(:multiple_cohorts)
-                                          Cohort.next.id
-                                        else
-                                          Cohort.current.id
-                                        end
+        report_schools_form.cohort_id = Cohort.next.id
         report_schools_form.lead_provider_id = current_user.lead_provider_profile.lead_provider.id
       end
 
--- a/app/controllers/lead_providers/report_schools/base_controller.rb
+++ b/app/controllers/lead_providers/report_schools/base_controller.rb
@@ -13,7 +13,7 @@ module LeadProviders
       def start
         clean_form!
 
-        report_schools_form.cohort_id = Cohort.current.id
+        report_schools_form.cohort_id = Cohort.next.id
         report_schools_form.lead_provider_id = current_user.lead_provider_profile.lead_provider.id
       end
 
--- a/app/controllers/lead_providers/report_schools/base_controller.rb
+++ b/app/controllers/lead_providers/report_schools/base_controller.rb
@@ -13,7 +13,11 @@ module LeadProviders
       def start
         clean_form!
 
-        report_schools_form.cohort_id = Cohort.next.id
+        report_schools_form.cohort_id = if FeatureFlag.active?(:multiple_cohorts)
+                                          Cohort.next.id
+                                        else
+                                          Cohort.current.id
+                                        end
         report_schools_form.lead_provider_id = current_user.lead_provider_profile.lead_provider.id
       end
 
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -13,7 +13,9 @@ class Cohort < ApplicationRecord
   end
 
   def self.next
-    find_by(start_year: 2022)
+    year = FeatureFlag.active?(:multiple_cohorts) ? 2022 : 2021
+
+    find_by(start_year: year)
   end
 
   def display_name
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -12,6 +12,10 @@ class Cohort < ApplicationRecord
     find_by(start_year: 2021)
   end
 
+  def self.next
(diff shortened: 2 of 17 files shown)


db/seeds/initial_seed.rb
```diff
@@ -17,7 +18,7 @@
   { provider_name: "UCL Institute of Education", cip: ucl_cip },
 ].each do |seed|
   provider = LeadProvider.find_or_create_by!(name: seed[:provider_name])
-  provider.update!(cohorts: [cohort_2021]) unless provider.cohorts.any?
+  provider.update!(cohorts: [cohort_2021, cohort_2022]) unless provider.cohorts.any?
```
jamesrose, line 21: This won't create the cohorts on production so you'll want to make sure they exist before flipping on the feature flag. I'd hold off until later though, as there might be unintended consequences of having multiple cohorts.

app/models/cohort.rb
```diff
@@ -12,6 +12,10 @@ def self.current
     find_by(start_year: 2021)
   end
 
+  def self.next
+    find_by(start_year: 2022)
+  end
```
jamesrose, line 17: Certainly not a deal-breaker, but have you considered this? You might be able to omit most of your feature conditions elsewhere.

```suggestion
  def self.next
    find_by(start_year: FeatureFlag.active?(:multiple_cohorts) ? 2022 : 2021)
  end
```

jamesrose (approved):

db/seeds/initial_seed.rb
leoapost: Yes, my understanding is that we need to manually create both the cohort and the feature flag on production before we flip it on.

app/models/cohort.rb
leoapost: Good suggestion. I pushed the changes.

lazydevorg (approved):

app/views/dashboard/_navigation.html.erb
```diff
@@ -3,8 +3,14 @@
     <%= component.nav_item(path: dashboard_path) do %>
       Overview
     <% end %>
-    <%= component.nav_item(path: lead_providers_your_schools_path) do %>
-      Schools
+    <% if FeatureFlag.active?(:multiple_cohorts) %>
```
jamesrose, line 6: Do you still need this conditional? It should always point to the right place.

app/controllers/lead_providers/report_schools/base_controller.rb
```diff
@@ -13,7 +13,11 @@ class BaseController < ::LeadProviders::BaseController
       def start
         clean_form!
 
-        report_schools_form.cohort_id = Cohort.current.id
+        report_schools_form.cohort_id = if FeatureFlag.active?(:multiple_cohorts)
```
jamesrose, line 16: I think after my suggestion this can always be `Cohort.next`

app/views/dashboard/show.html.erb
```diff
@@ -9,7 +9,11 @@
         </h2>
         <p>Choose a delivery partner and upload schools using a CSV.</p>
         <h2 class="govuk-heading-m">
-          <%= govuk_link_to "Check your schools for #{Cohort.current.display_name}", lead_providers_your_schools_path %>
+          <% if FeatureFlag.active?(:multiple_cohorts) %>
+            <%= govuk_link_to "Check your schools", lead_providers_your_schools_path(cohort: Cohort.next) %>
```
jamesrose, line 13: If this is just about the copy change I'd drop this condition entirely and just change it to "Check your schools" immediately.

jamesrose (approved):

app/views/dashboard/_navigation.html.erb
leoapost: We can drop the conditional, but keep the param in the URL as we want the cohort 2022 schools to be the default view when we flip on multiple cohorts.

app/controllers/lead_providers/report_schools/base_controller.rb
leoapost: I removed the conditional.

app/views/dashboard/show.html.erb
leoapost: I agree, that should be fine. I've pushed the changes.

jamesrose (approved):
🚢

PR #1977, opened 2022-04-20
Move Cypress features to Capybara Specs

mattonfoot:
Extends Spec/Support/Features/Pages to cover original scenarios built with Cypress and converts feature files to spec files


spec/support/features/pages/base.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Pages
+  class Base
```
asmega, line 4: i feel like you're writing you own page object library. would you consider using something like https://github.com/site-prism/site_prism ?

mattonfoot: that would indeed be simpler

mattonfoot: except main has been broken for 5 months

mattonfoot: done, nice library

bin/rspec_ci
```diff
@@ -5,10 +5,11 @@ ci_node_index = ENV["CI_NODE_INDEX"] || "0"
 github_sha = ENV['GITHUB_SHA'] || "fake-sha-123"
 
 tests = Dir["spec/**/*_spec.rb"].
```
jamesrose, line 7:
```suggestion
tests = Dir["spec/{[!features/]**/*_spec.rb"].
```

and remove the select below

spec/features/admin/schools/challenging_partnerships_spec.rb
```diff
@@ -6,20 +6,20 @@
   # feature removed while the new journeys are developed
   xscenario "Admin challenges school partnership" do
```
jamesrose, line 7: This one is disabled — should it be?

spec/features/admin/notes/participant_note_spec.rb
```diff
@@ -7,7 +7,7 @@
   include ParticipantDetailsSteps
   before do
     given_there_is_a_school_that_has_chosen_fip_for_2021_and_partnered
-    and_i_am_signed_in_as_an_admin
```
jamesrose, line 10: I don't think this rename was the right way to go — the user action is that they sign in.

spec/features/admin/schools/managing_2020_participants_spec.rb
```diff
@@ -5,15 +5,15 @@
 RSpec.feature "Admin managing 2020 participants", js: true, rutabaga: false do
   scenario "Admin views participants and adds a new one" do
     given_there_is_a_school_with_2020_cohort
-    and_i_am_signed_in_as_an_admin
+    and_i_authenticate_as_an_admin
     when_i_visit the_school_2020_cohort_page
-    then_the_page_should_be_accessible
-    and_percy_should_be_sent_a_snapshot_named("Admin NQT+1 listing page")
+    then_the_page_is_accessible
```
jamesrose, line 10: This should be `and_the_page_should_be_accessible` and/or `then_the_page_should_be_accessible` because they are matchers — they can fail.

spec/features/challenge_partnership_page_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+  let!(:cohort) { create :cohort, start_year: 2021 }
+
+  let!(:privacy_policy) do
+    create :privacy_policy
+    PrivacyPolicy::Publish.call
+  end
+
+  describe "when using an email link" do
+    scenario "Can see challenge options from an email link" do
+      given_a_fip_school_with_a_partnership_that_can_be_challenged "test-sit@example.com", "Test school", "111111-test-school", "abc1234"
```
jamesrose, line 15: I don't think you need to input these variables at all because they are all the same — can you just put them in the method?

PR #2008, opened 2022-04-29
Layout content tweaks

etheya:
Content tweaks and correct layout for some of the elements including H1s with wrong class style etc. Also changed out some of the dashboard column headers to be correct.


etheya:
Merged into 1 ticket, so closing.

PR #2011, opened 2022-04-29
content updates and layout fixes for nominate sit journey

etheya:
Futher content and layout changes to bring the live into line with prototype.


etheya:
Merged into 1 ticket, so closing.

PR #2051, opened 2022-05-06
[Design updates] -  confirm which programme first time through service

etheya:
Screens to follow. 
Includes, correct styling, headers, changed content.

### DIFF
--- a/app/views/privacy_policies/show.html.erb
+++ b/app/views/privacy_policies/show.html.erb
@@ -1,23 +1,22 @@
 <% content_for :title, "Privacy policy" %>
 
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-three-quarters">
-    <h1 class="govuk-heading-xl">Privacy policy</h1>
+    <div class="govuk-grid-column-two-thirds">
 
-    <% if !@policy.acceptance_required?(current_user) %>
-      <% content_for :before_content, govuk_back_link(text: "Back", href: :back) %>
-    <% else %>
-      <p>Before you continue, please check our privacy policy</p>
-      <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
-    <% end %>
+        <h1 class="govuk-heading-xl">Privacy policy</h1>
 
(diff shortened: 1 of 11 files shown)


app/views/privacy_policies/show.html.erb
```diff
@@ -1,23 +1,22 @@
...
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-three-quarters">
-    <h1 class="govuk-heading-xl">Privacy policy</h1>
+    <div class="govuk-grid-column-two-thirds">
 
-    <% if !@policy.acceptance_required?(current_user) %>
-      <% content_for :before_content, govuk_back_link(text: "Back", href: :back) %>
-    <% else %>
-      <p>Before you continue, please check our privacy policy</p>
-      <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
```
jamesrose, line 11: Are we ok removing this condition?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_04_29_144309) do
```
jamesrose, line 13: Can you run `rake db:migrate` locally and commit the changes? I don't think this change is relevant to the PR.

app/views/privacy_policies/show.html.erb
etheya: Well the condition is still there other than the line ( Before you continue and the <hr> ) the rest should still be there. 

Not sure what the line of content saying that does for a user to be fair. Condition for acceptance is still in place isnt it @jamesrose

db/schema.rb
etheya: No it shouldnt be.. i did do it - not sure why its pushing up that file. Can we not remove it from the PR @jamesrose

jamesrose (approved):

jamesrose:
@etheya I fixed this one for you. Merge when you're ready.

etheya:
Done - thanks

PR #2065, opened 2022-05-10
[582] - removed content no longer needed as multi cohort going live

etheya:
We no longer need this content - with go live of Multi cohort

image


etheya:
If @jamesrose or @xswoodsx  could review :-)

jamesrose (approved):

PR #2068, opened 2022-05-10
Display blank LP and DP names when the partnership is challenged

leoapost:
### Context

We should not display the Lead Provider and Delivery Partner names when the partnership is challenged.

Retrieve the latest partnership for the cohort and display:
- the LP and DP names, when it's not challenged
- "To be confirmed", when the partnership does not exist yet (pending)
-  no LP and DP names, when it's challenged

### DIFF
--- a/app/views/schools/dashboard/_school_cohort_details.html.erb
+++ b/app/views/schools/dashboard/_school_cohort_details.html.erb
@@ -1,7 +1,7 @@
 <% lead_provider = school_cohort.default_induction_programme&.lead_provider %>
 <% delivery_partner =  school_cohort.default_induction_programme&.delivery_partner %>
 <% ects_count = school_cohort.current_induction_records.count %>
-<% partnership = school_cohort.school.partnerships.active.find_by(cohort: school_cohort.cohort) %>
+<% latest_partnership = school_cohort.school.partnerships.where(cohort: school_cohort.cohort, relationship: false).order(created_at: :desc).limit(1).first %>
 
 <% if school_cohort.school_chose_cip? && school_cohort.default_induction_programme&.core_induction_programme.blank? %>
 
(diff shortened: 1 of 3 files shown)


lazydevorg (approved):

app/views/schools/dashboard/_school_cohort_details.html.erb
```diff
@@ -37,12 +37,20 @@
   <% if school_cohort.full_induction_programme? %>
     <% list.row do |row| %>
       <% row.key(text: "Training provider") %>
-      <% row.value(text: lead_provider&.name ? lead_provider.name : "To be confirmed") %>
+      <% if latest_partnership&.challenged? == true %>
```
jamesrose, line 40:
```suggestion
      <% if latest_partnership&.challenged? %>
```

Do you need the condition here (and below)?

app/views/schools/dashboard/_school_cohort_details.html.erb
```diff
@@ -1,7 +1,7 @@
 <% lead_provider = school_cohort.default_induction_programme&.lead_provider %>
 <% delivery_partner =  school_cohort.default_induction_programme&.delivery_partner %>
 <% ects_count = school_cohort.current_induction_records.count %>
-<% partnership = school_cohort.school.partnerships.active.find_by(cohort: school_cohort.cohort) %>
+<% latest_partnership = school_cohort.school.partnerships.where(cohort: school_cohort.cohort).order(created_at: :desc).limit(1).first %>
```
jamesrose, line 4: Can you explain this one? I read this and thought about whether ordering by `created_at` enough? Is it possible that you select the wrong partnership here if there are multiple?

app/views/schools/dashboard/_school_cohort_details.html.erb
leoapost: Both are required because there are different behaviours for true, false and nil values.

`false`: the LP and DP names, when it's not challenged
`true`: no LP and DP names, when it's challenged
`nil`: "To be confirmed", when the partnership does not exist yet (pending)

app/views/schools/dashboard/_school_cohort_details.html.erb
leoapost: I don't know how we create partnerships, so I had to assume the latest one will be always the one we need. @tonyheadford may know more?

leoapost: @jamesrose Tony suggested we should be ok if we included `relationship: false` in the `where` statement. I've already pushed it.

jamesrose (approved):

app/views/schools/dashboard/_school_cohort_details.html.erb
jamesrose: Thanks! 🚢

PR #2074, opened 2022-05-10
[CPDLP-1316] NPQ cohort can be injected

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1316
- NPQ applications are not cohort aware
- This PR gets us a step closer

- Default all existing NPQ applications to 2021
- Allow NPQ rails app to inject the cohort when applying

### Guidance to review

- Client work for this change is yet to be written
- So you could hand craft an API call and application should be assigned to set cohort

### DIFF
--- a/db/migrate/20220510115548_add_cohort_npq_applications.rb
+++ b/db/migrate/20220510115548_add_cohort_npq_applications.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddCohortNPQApplications < ActiveRecord::Migration[6.1]
+  def change
+    safety_assured do
+      add_reference :npq_applications, :cohort, null: true
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


db/migrate/20220510115548_add_cohort_npq_applications.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddCohortNPQApplications < ActiveRecord::Migration[6.1]
+  def change
+    add_column :npq_applications, :cohort, :integer, default: 2021, null: false
```
jamesrose, line 5: Should this not be a join to `Cohort`?

asmega: ok, i've changed this so it is now a `Cohort` object

asmega:
Superseded by https://github.com/DFE-Digital/early-careers-framework/pull/2082 due to failing rake task on build

PR #2079, opened 2022-05-10
Extra conditional on sending added_and_validated_email

xswoodsx:
When a SIT adds themselves as a mentor it's erroring because there is no template for this instance. We're adding a temporary conditional to check if they're adding selves so that sentry doesn't error. We'll decide later whether to stick with this or add a new template.

[Error in sentry](https://sentry.io/organizations/dfe-teacher-services/issues/3262680691/?project=5748989&query=is%3Aunresolved)

### DIFF
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -333,8 +333,9 @@ module Schools
 
       participant_validation_record = validation_record(profile)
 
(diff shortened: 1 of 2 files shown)


jamesrose:
Could you add a spec please?

jamesrose:
Lovely, thanks!

PR #2080, opened 2022-05-10
Fix prod issue when partnership challenged

tonyheadford:
### Context

A prod issue has been found where if there is a FIP school with a training_status withdrawn participant and the partnership has been challenged that the participants list attempts to use the `@school.lead_provider(<year>).name` which in this case generates an error.


spec/features/schools/participant_dashboard/manage_fip_partnered_participants_spec.rb
```diff
@@ -24,7 +24,7 @@
       and_i_am_signed_in_as_an_induction_coordinator
     end
 
-    context "in to a school" do
+    xcontext "in to a school" do
```
jamesrose, line 27: Should this still be disabled?

tonyheadford: Rebased 👍

jamesrose (approved):

PR #2100, opened 2022-05-12
[Bug fix] Handle partnership change

tonyheadford:
### Context

When a partnership is challenged and subsequently a new partnership declared we should:
 * Add a new `InductionProgramme`
 * Add any participants from the previous (default) induction programme to the new induction programme
 * Set the new induction programme as the default for the school cohort


xswoodsx (approved):
LGTM,  it need a rebase for the failing tests?

PR #2102, opened 2022-05-13
Allow school inviter to run in the background

jamesrose:
### Context

We're about to invite schools without a SIT for 22/23. This PR allows the mailer to run in the background so that we can depend on Sidekiq to automatically retry invitations if they were to fail.

### DIFF
--- a/app/forms/nomination_request_form.rb
+++ b/app/forms/nomination_request_form.rb
@@ -36,7 +36,7 @@ class NominationRequestForm
       ActiveRecord::Base.transaction do
         raise TooManyEmailsError if reached_email_limit
 
-        invite_schools_service.run([school.urn])
+        invite_schools_service.perform([school.urn])
       end
     end
   end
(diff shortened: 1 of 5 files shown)


app/forms/nomination_request_form.rb
```diff
@@ -36,7 +36,7 @@ def save!
       ActiveRecord::Base.transaction do
         raise TooManyEmailsError if reached_email_limit
 
-        invite_schools_service.run([school.urn])
+        invite_schools_service.perform([school.urn])
```
asmega, line 39: are you sure this change makes it async? from https://www.rubydoc.info/github/mperham/sidekiq/Sidekiq/Worker
should it not be `perform_async` on the class?

jamesrose: @asmega No, it doesn't but I intentionally kept this one synchronous so that we can keep the transaction above it. I only wanted to make it async so that I can run this in the background when I send out emails to many 1000s of schools manually.

asmega (approved):

PR #2113, opened 2022-05-16
[CPDLP-1326] Expose api targeted flag

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1326
- We want to surface `targeted_delivery_funding_eligibility` to lead providers in the API
- So they are clear on what funding they will receive

- Expose `targeted_delivery_funding_eligibility` in API v1 on NPQ applications
- Updated docs
- Updated change log 

### Guidance to review

- Fetch an NPQ application via API
- Should see `false` on `targeted_delivery_funding_eligibility`
- Although may `true` if correct criteria met
- View docs on review app
- Should see extra field on NPQ applications
- Should see updated change log

### DIFF
--- a/app/models/npq_application.rb
+++ b/app/models/npq_application.rb
@@ -3,7 +3,7 @@
 class NPQApplication < ApplicationRecord
   has_paper_trail only: %i[user_id npq_lead_provider_id npq_course_id created_at updated_at lead_provider_approval_status]
 
-  self.ignored_columns = %w[user_id]
+  self.ignored_columns = %w[user_id targeted_support_funding_eligibility]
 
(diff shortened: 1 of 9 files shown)


app/models/npq_application.rb
```diff
@@ -3,7 +3,7 @@
 class NPQApplication < ApplicationRecord
   has_paper_trail only: %i[user_id npq_lead_provider_id npq_course_id created_at updated_at lead_provider_approval_status]
 
-  self.ignored_columns = %w[user_id]
+  self.ignored_columns = %w[user_id targeted_support_funding_eligibility]
```
StupidCodeFactory, line 6: Why is that ignored?

jamesrose (approved):

asmega:
this has now been done elsewhere

PR #2115, opened 2022-05-16
[CPDLP-1353] add class to encapsulate seeding of NPQ apps

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1353

- Add a class than can then be called to generate test NPQ applications for environments

### Guidance to review

- Fire up a rails console
- Execute the following
```
require_relative "db/seeds/npq_applications.rb"
cpd = CpdLeadProvider.all.sample
Seeds::NPQApplication.new(cpd_lead_provider: cpd).call
```
- Should have a new NPQ application against the provider

### DIFF
--- a/db/seeds/npq_applications.rb
+++ b/db/seeds/npq_applications.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+require "securerandom"
+
+module Seeds
+  class NPQApplication
+    attr_reader :cpd_lead_provider
+
+    def initialize(cpd_lead_provider:)
+      @cpd_lead_provider = cpd_lead_provider
+    end
+
+    def call
+      return if Rails.env.production?
+
+      NPQ::BuildApplication.call(
+        npq_application_params: {
+          active_alert: false,
+          date_of_birth: rand(60.years.ago.to_date..23.years.ago.to_date),
+          teacher_reference_number: trn,
+          eligible_for_funding: [true, false].sample,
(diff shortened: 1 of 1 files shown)


db/seeds/npq_applications.rb
```diff
@@ -0,0 +1,84 @@
...
+module Seeds
+  class NPQApplication
+    attr_reader :cpd_lead_provider
+
+    def initialize(cpd_lead_provider:)
+      @cpd_lead_provider = cpd_lead_provider
+    end
+
+    def call
+      NPQ::BuildApplication.call(
```
jamesrose, line 14:
```suggestion
      return if Rails.env.production?
      
      NPQ::BuildApplication.call(
```

What do you think?

asmega: sure. it's not loaded so, one has to be explicitly require it for it to be available

db/seeds/npq_applications.rb
```diff
@@ -0,0 +1,86 @@
...
+      @works_in_school ||= [true, false].sample
+    end
+
+    def employment_role
+      @employment_role ||= [
+        "Teacher",
+        "Manager",
+        "Support Staff",
+        "Business Owner",
+        "Admin Staff",
```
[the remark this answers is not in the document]
asmega: nah, it's just some random string I came up with

jamesrose (approved):

PR #2120, opened 2022-05-17
[CPDLP-1335] Internal API NPQ funding

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1335

- Added service class to calculate NPQ funding eligibility based on TRN and course
- Added private api endpoint so it is accessible via NPQ rails app api call

### Guidance to review

- Create a private api token
- Using this token request NPQ funding with `trn` and `npq_course_identifier`
- Should return appropriate result

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -52,6 +52,9 @@ Rails.application.routes.draw do
 
     namespace :v1 do
       concern :participant_actions, Participants::Routing.new
+
+      resources :npq_funding, only: [:index], path: "npq-funding"
+
       resources :ecf_participants, path: "participants/ecf", only: %i[index] do
         concerns :participant_actions
       end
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -52,6 +52,9 @@
 
     namespace :v1 do
       concern :participant_actions, Participants::Routing.new
+
+      resources :npq_funding, only: [:index], path: "npq-funding"
```
Lockyy, line 56: Is this more of a show? We're looking up data on a single TRN and getting one hash of data back.

Lockyy (approved):
LGTM

config/routes.rb
asmega: sure, I've changed this to a `show` now

Lockyy (approved):
Approving, Phil has let me know that the end-to-end failures are unrelated and being fixed by Matt

PR #2122, opened 2022-05-18
[CPDLP] - 1162 banding tracker not displaying correctly

StupidCodeFactory:
### Context

- [CPDLP-1162](https://dfedigital.atlassian.net/browse/CPDLP-1162): 

Bands are now calculated per declaration type across the entire cohort.


app/controllers/finance/banding_tracker/providers_controller.rb
```diff
@@ -21,7 +21,9 @@ def paid_participant_count_per_bands
       end
 
       def payable_participant_count_per_bands
-        @payable_participant_count_per_bands ||= lead_provider.participant_declarations.payable
+        @payable_participant_count_per_bands ||= lead_provider
```
asmega, line 24: while we're here can we also code in the current cohort to prevent any panic attacks?

asmega:
I think I would have preferred to use the payment calculator for the banding tracker. Because if there is a bug in the payment calc you won't see it here as it's an independent calculation

StupidCodeFactory:
> I think I would have preferred to use the payment calculator for the banding tracker. Because if there is a bug in the payment calc you won't see it here as it's an independent calculation

@asmega, I understand your concerns, we are going to remove the payment calculator as it is no longer needed, it's quite obscure to work with (and should have been gotten rid off when we we re-did the payment breakdown page). 
However we'll address this in our firebreak sprint.
We now have the necessary associations in place to use simple scopes + group by.
I've go a branch which demonstrate this. https://github.com/DFE-Digital/early-careers-framework/pull/2007/files#diff-0b86ca4c4ed52cc02af637cef0fc084ba54814b8b37542b713eee75a8316b7efL63-L95

PR #2123, opened 2022-05-18
CST-723 - change LD DP issue

lazydevorg:
### Context

- Ticket: CST-723

When a school set up the next cohort and choose to either change the lead provider or the delivery partner the dashboard shows that both of them need to be confirmed.
This is happening because the partnership is not created in the case above.
The solution proposed is to have two flags on the induction programme that store whether the lead provider or the delivery partner need to be confirmed.

### DIFF
--- a/app/models/school_cohort.rb
+++ b/app/models/school_cohort.rb
@@ -59,8 +59,24 @@ class SchoolCohort < ApplicationRecord
     school.lead_provider(cohort.start_year)
   end
 
+  def lead_provider_to_be_confirmed?
+    default_induction_programme.lead_provider_to_be_confirmed
+  end
+
+  def previous_lead_provider
+    school.lead_provider(cohort.start_year - 1)
+  end
+
   def delivery_partner
     school.delivery_partner_for(cohort.start_year)
+  end
+
+  def delivery_partner_to_be_confirmed?
+    default_induction_programme.delivery_partner_to_be_confirmed
+  end
+
+  def previous_delivery_partner
+    school.delivery_partner_for(cohort.start_year - 1)
   end
 
   def school_chose_cip?
--- a/app/models/school_cohort.rb
+++ b/app/models/school_cohort.rb
@@ -39,6 +39,7 @@ class SchoolCohort < ApplicationRecord
   scope :for_year, ->(year) { joins(:cohort).where(cohort: { start_year: year }) }
 
   delegate :description, :academic_year, to: :cohort
+  delegate :delivery_partner_to_be_confirmed, to: :default_induction_programme
 
   after_save do |school_cohort|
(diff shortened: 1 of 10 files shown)


tonyheadford (commented):
A bit unsure why we need 2 flags, perhaps I misunderstood but I thought that is the SIT want to change the LP we handle that there is no LP/DP, it's only when the SIT wants to change the DP do we have the situation when we want to display the LP but not the DP from the partnership in the previous cohort?

app/models/school_cohort.rb
```diff
@@ -59,10 +59,22 @@ def lead_provider
     school.lead_provider(cohort.start_year)
   end
 
+  def previous_lead_provider
+    school.lead_provider(cohort.start_year - 1)
+  end
+
```
tonyheadford, line 65: Is this used anywhere? If it is could we move this somewhere else please, or remove if not?

app/models/school_cohort.rb
```diff
@@ -59,10 +59,22 @@ def lead_provider
...
   end
 
+  def delivery_partner_to_be_confirmed?
+    default_induction_programme.delivery_partner_to_be_confirmed
+  end
+
+  def previous_delivery_partner
+    school.delivery_partner_for(cohort.start_year - 1)
+  end
+
```
tonyheadford, line 77: Same as above, can we move this somewhere else or delete it please if it's not used?

app/models/school_cohort.rb
```diff
@@ -59,10 +59,22 @@ def lead_provider
...
+    school.lead_provider(cohort.start_year - 1)
+  end
+
   def delivery_partner
     school.delivery_partner_for(cohort.start_year)
   end
 
+  def delivery_partner_to_be_confirmed?
+    default_induction_programme.delivery_partner_to_be_confirmed
+  end
```
tonyheadford, line 72: Could we delegate this to `default_induction_programme` ?

app/models/school_cohort.rb
lazydevorg: not used and removed

app/models/school_cohort.rb
lazydevorg: This is used in the DashBoardController to show the previous lead provider when a partnership is not present.
I have put this method in the SchoolCohort class because it is related to the School and Cohort to which the SchoolCohort belongs to.
I'm happy to move it somewhere else. Any idea?

tonyheadford: I would probably handle this in the DashboardController, perhaps just find the previous school_cohort and then call the `lead_provider` method on that, if that makes sense?  Just feels like navigating different cohorts and providers should be on the model?

PR #2124, opened 2022-05-18
[CPDLP-1360] Improve InductionRecord serializer updated_at

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1360
- Providers are not updating/refreshing records as the `updated_at` field is not accurately reflecting when a record has been updated

- `ParticipantFromInductionRecordSerializer` `updated_at` now considers more info to more accurately reflect when record was updated at

### Guidance to review

- Find ECF user via API
- Touch their induction record
- `updated_at` should have been affected

### DIFF
--- a/app/serializers/api/v1/participant_from_induction_record_serializer.rb
+++ b/app/serializers/api/v1/participant_from_induction_record_serializer.rb
@@ -116,7 +116,10 @@ module Api
       end
 
       attribute :updated_at do |induction_record|
-        induction_record.participant_profile.user.updated_at.rfc3339
+        [
+          induction_record.participant_profile.user.updated_at,
+          induction_record.updated_at,
+        ].max.rfc3339
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/serializers/api/v1/participant_from_induction_record_serializer.rb
```diff
@@ -116,7 +116,10 @@ def eligible_for_funding?(profile)
       end
 
       attribute :updated_at do |induction_record|
-        induction_record.participant_profile.user.updated_at.rfc3339
+        [
```
jamesrose, line 119: I think we would want to update the `updated_since` flag too, otherwise LPs will be only querying against the user `updated_at`

asmega: Good point, I can add that. I'm hoping an extra `or` query won't cost too much

jamesrose (approved):

PR #2129, opened 2022-05-18
[CPDLP-1243] - allow a df e user to change training status from the support page

StupidCodeFactory:
### Context

- [CPDLP-1243](https://dfedigital.atlassian.net/browse/CPDLP-1243): 

Allow to update a participant's training status with a simple drop down. 
I added a service class to handle the update. It's seems overkill for now but we're going to add more to this and this will come in handy and cleaner.


app/services/finance/participants/update.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Finance
+  module Participants
+    class Update
+      def initialize(participant_profile)
+        self.participant_profile = participant_profile
+      end
+
+      def call(attributes)
```
StupidCodeFactory, line 10: This probably appears for now overkill. We have however more fields to update in the pipeline, and this may become more relevant. If it remains overkill i.e `participant_profile.update!(attributes)` is the only thing we need to do then we can get rid of this. 
But if someone feels strongly about it I'm happy to remove it and have the controller perform the update.

app/controllers/finance/participants/profiles_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+        @profile = user.participant_profiles.find(params[:id])
+      end
+
+      def update
+        profile = user.participant_profiles.find(params[:id])
+        Participants::Update.new(profile).call(profile_attributes)
+
+        return redirect_to finance_participant_path(user) if profile.errors.none?
+
+        @profile = profile
```
asmega, line 16: you can delete this line if you set in on 11 then use it throughout the rest of this method

spec/factories/participant_profiles/ects.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :ect, class: "ParticipantProfile::ECT" do
```
asmega, line 4: there is already a `ect_participant_profile` factory. is that insufficient for your needs?

app/views/finance/participants/_ecf_profile.html.erb
```diff
@@ -1,57 +1,70 @@
...
     <% row.key { "Status" } %>
     <% row.value { pp.status } %>
+    <% row.action %>
   <% end %>
 
   <% summary_list.row do |row| %>
     <% row.key { "Training status" } %>
     <% row.value { pp.training_status } %>
+    <% row.action text: "Change", visually_hidden_text: "Change training status", href: edit_finance_participant_profile_path(@user, pp) %>
   <% end %>
```
ericaporter, line 44: You could put this and the other text in en.yml but I can see this isn't being done already in these views

asmega (approved):

spec/features/finance/participant_drilldown_spec.rb
```diff
@@ -38,6 +38,28 @@
...
+        .to have_sibling("dd.govuk-summary-list__value", text: "withdrawn")
+
+      actions = page
+                  .find("dt.govuk-summary-list__key", text: "Training status")
+                  .sibling("dd.govuk-summary-list__actions")
+      within(actions) { click_on "Change" }
+
+      expect(page).to have_select("Change training status", selected: "Withdrawn")
+    end
   end
```
ericaporter, line 63: This is a bit of a change in the style of the feature tests, is it worse to have to abstract everything or to have the inconsistency? 😄

PR #2133, opened 2022-05-19
[CPDLP-1049] Introduce statement line items

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1049
- In order to accomplish clawbacks we need to associate a declaration with multiple statements
- One for where we payout and potentially another for when we do a refund
- At the moment a declaration can only be associated with a single statement
- So a join table has been introduced to overcome this limitation

- Introduce statement line items
- This joins statements and declarations together
- The old mechanism has been left in place as it is still used whilst new mechanism is slotted into position and missing data backfilled
- Added class to backfill line items for existing declarations associated with a statement
- When declarations are auto attached to a statement the line item is also created
- When voiding a declaration the line item is also updated to reflect the new `voided` state
- When transitioning a statement to `payable`, the state of line items are also transitioned 
- There is a slight gap as code to transition to `paid` does not exist yet

### Guidance to review

- `Importers::StatementLineItems.new.call` should backfill line items

### DIFF
--- a/app/models/finance/statement.rb
+++ b/app/models/finance/statement.rb
@@ -32,20 +32,7 @@ class Finance::Statement < ApplicationRecord
     end
 
     def attach(participant_declaration)
-      statement = statement_for(participant_declaration)
-      statement.participant_declarations << participant_declaration
-    end
-
-  private
-
-    def statement_for(participant_declaration)
-      cohort = participant_declaration.participant_profile.schedule.cohort
-      case participant_declaration
-      when ParticipantDeclaration::ECF
-        participant_declaration.cpd_lead_provider.lead_provider.next_output_fee_statement(cohort)
-      when ParticipantDeclaration::NPQ
-        participant_declaration.cpd_lead_provider.npq_lead_provider.next_output_fee_statement(cohort)
-      end
+      Finance::DeclarationStatementAttacher.new(participant_declaration: participant_declaration).call
     end
   end
 
--- a/app/models/finance/statement.rb
+++ b/app/models/finance/statement.rb
@@ -7,7 +7,10 @@ class Finance::Statement < ApplicationRecord
 
   belongs_to :cpd_lead_provider
   belongs_to :cohort
+
   has_many :participant_declarations
+  has_many :statement_line_items, class_name: "Finance::StatementLineItem"
+
   scope :payable,                   -> { where("deadline_date < DATE(NOW()) AND payment_date >= DATE(NOW())") }
   scope :closed,                    -> { where("payment_date < ?", Date.current) }
   scope :with_future_deadline_date, -> { where("deadline_date >= DATE(NOW())") }
--- a/app/models/finance/statement_line_item.rb
+++ b/app/models/finance/statement_line_item.rb
@@ -5,4 +5,9 @@ class Finance::StatementLineItem < ApplicationRecord
 
   belongs_to :statement
   belongs_to :participant_declaration
+
+  scope :eligible, -> { where(state: "eligible") }
+  scope :payable, -> { where(state: "payable") }
+  scope :paid, -> { where(state: "paid") }
+  scope :voided, -> { where(state: "voided") }
 end
--- a/app/models/finance/statement_line_item.rb
+++ b/app/models/finance/statement_line_item.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class Finance::StatementLineItem < ApplicationRecord
+  self.table_name = "statement_line_items"
+
+  belongs_to :statement
+  belongs_to :participant_declaration
+end
--- a/app/models/finance/statement_line_item.rb
+++ b/app/models/finance/statement_line_item.rb
@@ -6,8 +6,12 @@ class Finance::StatementLineItem < ApplicationRecord
   belongs_to :statement
   belongs_to :participant_declaration
 
-  scope :eligible, -> { where(state: "eligible") }
-  scope :payable, -> { where(state: "payable") }
-  scope :paid, -> { where(state: "paid") }
-  scope :voided, -> { where(state: "voided") }
+  enum state: {
+    submitted: "submitted",
+    eligible: "eligible",
+    payable: "payable",
+    paid: "paid",
+    voided: "voided",
+    ineligible: "ineligible",
+  }
 end
--- a/app/models/finance/statement_line_item.rb
+++ b/app/models/finance/statement_line_item.rb
@@ -14,4 +14,19 @@ class Finance::StatementLineItem < ApplicationRecord
     voided: "voided",
     ineligible: "ineligible",
   }
+
+  scope :billable, -> { where(state: %w[eligible payable paid]) }
+
+  validate :validate_single_billable_relationship, on: [:create]
+
+private
+
+  def validate_single_billable_relationship
+    if Finance::StatementLineItem
+      .where(participant_declaration: participant_declaration)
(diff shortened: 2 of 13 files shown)


app/models/finance/statement.rb
```diff
@@ -7,7 +7,10 @@ class Finance::Statement < ApplicationRecord
 
   belongs_to :cpd_lead_provider
   belongs_to :cohort
+
   has_many :participant_declarations
+  has_many :statement_line_items, class_name: "Finance::StatementLineItem"
```
StupidCodeFactory, line 12: Note for later: we'll need to apply the correct scope to the subclasses. So that we can query and aggregate on those on the payment breakdowns

db/migrate/20220518120740_create_statement_line_items.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateStatementLineItems < ActiveRecord::Migration[6.1]
+  def change
+    create_table :statement_line_items do |t|
+      t.references :statement
+      t.references :participant_declaration
+
+      t.text :state
```
StupidCodeFactory, line 9:
```suggestion
      t.text :state, null: false # possibly default: "eligible"?
```

app/models/finance/statement_line_item.rb
```diff
@@ -0,0 +1,13 @@
...
+class Finance::StatementLineItem < ApplicationRecord
+  self.table_name = "statement_line_items"
+
+  belongs_to :statement
+  belongs_to :participant_declaration
+
+  scope :eligible, -> { where(state: "eligible") }
+  scope :payable, -> { where(state: "payable") }
+  scope :paid, -> { where(state: "paid") }
+  scope :voided, -> { where(state: "voided") }
```
StupidCodeFactory, line 12: can we have an `enum :state {}` to be consistent with the rest of the app and obvious what are the possible states?
That would also generate those scopes automatically.

app/services/void_participant_declaration.rb
```diff
@@ -14,13 +14,20 @@ def call
     raise Api::Errors::InvalidTransitionError, I18n.t(:declaration_already_voided) if declaration.voided?
     raise Api::Errors::InvalidTransitionError, I18n.t(:declaration_not_voidable) unless declaration.voidable?
 
-    declaration.make_voided!
+    ApplicationRecord.transaction do
+      declaration.make_voided!
+      line_item.update!(state: "voided") if line_item
```
StupidCodeFactory, line 19: suggestion: 
```suggestion
      line_item.voided! if line_item
```

app/models/finance/statement_line_item.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class Finance::StatementLineItem < ApplicationRecord
```
StupidCodeFactory, line 3: I think would add a validation that there can not be another line items may not existing for that declaration except `pending_withdrawal` and `withdrawn`
It's a bit defensive but I'd rather proceed with caution

spec/services/finance/declaration_statement_attacher_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  let(:statement) { create(:ecf_statement, output_fee: true, deadline_date: 2.months.from_now) }
+  let(:cpd_lead_provider) { statement.cpd_lead_provider }
+
+  subject { described_class.new(participant_declaration: declaration) }
+
+  describe "#call" do
+    it "creates line item" do
+      expect {
+        subject.call
+      }.to change(Finance::StatementLineItem, :count).by(1)
```
StupidCodeFactory, line 14: slightly more robust, checks it is attached to the correct statement:
```suggestion
      }.to change(statement.line_items, :count).by(1)
```

spec/services/finance/declaration_statement_attacher_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    it "create line item with same state as declaration" do
+      subject.call
+
+      line_item = Finance::StatementLineItem.last
+      expect(line_item.state).to eql(declaration.state)
```
StupidCodeFactory, line 21: I'm assuming this is only temporary and the state will move off the declaration to the line item.

spec/services/void_participant_declaration_spec.rb
```diff
@@ -85,5 +85,39 @@
...
+      let(:line_item) { declaration.statement_line_items.first }
+
+      before do
+        Finance::DeclarationStatementAttacher.new(participant_declaration: declaration).call
+      end
+
+      it "update line item state to voided" do
+        subject.call
+        expect(declaration.reload).to be_voided
+        expect(line_item.reload.state).to eql("voided")
```
StupidCodeFactory, line 119:
```suggestion
        expect(line_item.reload).to be_voided
```

db/migrate/20220518120740_create_statement_line_items.rb
asmega: i can stick the null protection in but don't think setting a default would be wise in case someone forgets to set it

app/models/finance/statement_line_item.rb
asmega: sorry, i don't follow your sentence, could you re-phrase? or give an example

spec/services/finance/declaration_statement_attacher_spec.rb
asmega: > the state will move off the declaration to the line item.

no, state will exist on both the declaration and line item. they have different purposes.

- the line item state represents the relationship between the declaration and the statement
- the declaration state represents the state of the declaration in the lifecycle (state machine) of the declaration independent of any statements

app/models/finance/statement_line_item.rb
StupidCodeFactory: 2. when creating an `awaiting_clawback` it must have a similar "paid" line item 
3. when creating a `clawedback` line items it should check it has the corresponding `awaiting_clawback`

asmega: the concepts of clawbacks have not been added yet. they do not exist yet in the system. these concepts will be introduced in the next iteration. therefore I cannot at the moment add any validation pertaining to clawbacks.

asmega: i can however add validation to prevent a declaration from being attached twice to a statement(s)

spec/services/finance/declaration_statement_attacher_spec.rb
StupidCodeFactory: so we could have a paid declaration that's been clawed back or a paid declaration that's been awaiting for claw back?

StupidCodeFactory: what's would that be useful for?

asmega: > so we could have a paid declaration that's been clawed back

a clawed back declaration must have been paid at some point by definition. otherwise it would have been voided. so not sure what your point is?

> what problem is that trying to solve?

if hypothetically state were moved to the line item. how does one query the current state of a declaration?
- when a provider fetch all their declarations, we must provide the current state of declarations. how do you do that at scale?
- another example is one shouldn't be able to void something is clawed back. how do you discern the current state of a declaration to prevent the invalid state change?

StupidCodeFactory (approved):

PR #2140, opened 2022-05-20
Fix for bad requests

ericaporter:
### Context

- Ticket: n/a

Prevent no method error when users make bad requests with missing / incorrect user_profile

### DIFF
--- a/app/services/participants/change_schedule/ecf.rb
+++ b/app/services/participants/change_schedule/ecf.rb
@@ -42,6 +42,8 @@ module Participants
       attr_reader :schedule_identifier, :cohort_year
 
       def relevant_induction_record
+        return if user_profile.nil?
+
         user_profile
           .induction_records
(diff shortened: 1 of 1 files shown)


app/services/participants/change_schedule/ecf.rb
```diff
@@ -42,6 +42,8 @@ def call
       attr_reader :schedule_identifier, :cohort_year
 
       def relevant_induction_record
+        return if user_profile.nil?
```
jamesrose, line 45:
```suggestion
        raise ActionController::ParameterMissing, I18n.t(:invalid_participant) if user_profile.nil?
```

Might be better?

jamesrose (approved):

StupidCodeFactory:
@ericaporter we can close this now that it's been fixed in https://github.com/DFE-Digital/early-careers-framework/pull/2158

PR #2142, opened 2022-05-20
[CPDLP-1258] DP participants filters/search feature

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?issueParent=39631&selectedIssue=CPDLP-1258

* Add search and filter by role, academic year and status

### Guidance to review

* Login as `delivery-partner@example.com` user

### DIFF
--- a/app/components/delivery_partners/participants/table_row.rb
+++ b/app/components/delivery_partners/participants/table_row.rb
@@ -5,11 +5,18 @@ module DeliveryPartners
     class TableRow < BaseComponent
       with_collection_parameter :participant_profile
 
-      delegate :user, :teacher_profile, :school_cohort, :schedule, :training_status, :ecf_participant_eligibility,
+      delegate :user,
+               :teacher_profile,
+               :school,
+               :cohort,
                to: :participant_profile
 
       delegate :full_name, :email, :user_description,
                to: :user
+
+      delegate :training_status,
+               to: :induction_record,
+               allow_nil: true
 
       def initialize(participant_profile:)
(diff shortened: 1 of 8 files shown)


app/components/delivery_partners/participants/table_row.rb
```diff
@@ -5,7 +5,11 @@ module Participants
     class TableRow < BaseComponent
       with_collection_parameter :participant_profile
 
-      delegate :user, :teacher_profile, :school_cohort, :schedule, :training_status, :ecf_participant_eligibility,
+      delegate :user,
+               :teacher_profile,
+               :training_status,
```
asmega, line 10: you want to be using `training_status` on the `InductionRecord`. The profile one will start to no longer accurately reflect the user circumstances

asmega: so if you are a DP you should be able to work out the lead provider and therefore work out the correct induction record

app/services/delivery_partners/participants_filter.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+
+    def filter_role(scoped, role)
+      case role
+      when "ect"
+        scoped.where(type: "ParticipantProfile::ECT")
+      when "mentor"
+        scoped.where(type: "ParticipantProfile::Mentor")
+      else
+        scope
```
asmega, line 61: should this be `scoped`? otherwise infinite loop?

PR #2153, opened 2022-05-24
CST-702 Remove start term step from add participant form

lazydevorg:
### Context

- Ticket: CST-702

When adding participants the start term page has been removed as not more needed having multiple cohorts.


tonyheadford (approved):
LGTM

PR #2158, opened 2022-05-25
[CPDLP-1371] - fix change schedule 500 errors

StupidCodeFactory:
### Context

- [CPDLP-1371](https://dfedigital.atlassian.net/browse/CPDLP-1371): 

Is the user profile is nil because the participant is withdrawn then flag as invalid participant.

### DIFF
--- a/spec/features/inducation_nomination_spec.rb
+++ b/spec/features/inducation_nomination_spec.rb
@@ -78,6 +78,7 @@ RSpec.feature "Nominating tutors", :js do
       expect(page).to have_css(".govuk-panel--confirmation", text: "Induction tutor nominated")
 
       and_the_page_should_be_accessible
+      and_percy_should_be_sent_a_snapshot_named "Start nominations email already used again"
 
       expect(SchoolMailer)
         .to have_received(:nomination_confirmation_email)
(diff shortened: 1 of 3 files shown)


spec/features/inducation_nomination_spec.rb
```diff
@@ -78,7 +78,7 @@
       expect(page).to have_css(".govuk-panel--confirmation", text: "Induction tutor nominated")
 
       and_the_page_should_be_accessible
-      and_percy_should_be_sent_a_snapshot_named "Start nominations email already used"
```
StupidCodeFactory, line 81: For some reason this fails on my branch, there are two of them in the file.

asmega (approved):

spec/services/participants/change_schedule/ecf_spec.rb
```diff
@@ -24,6 +24,25 @@ def matches_lead_provider?
...
+          schedule_identifier: schedule.schedule_identifier,
+          participant_id: user.id,
+          course_identifier: "ecf-induction",
+          cpd_lead_provider: CpdLeadProvider.new,
+        })
+      end
+
+      it "should have an error" do
+        expect { subject.call }.to raise_error(ActionController::ParameterMissing, /must be a valid Participant ID/)
+      end
```
ericaporter, line 43: I would question whether the error message could be a bit more descriptive but I think we have a ticket for that

ericaporter (approved):

spec/services/participants/change_schedule/ecf_spec.rb
StupidCodeFactory: yes the way the validations are order dependent is making this nearly impossible to tell the provider that this participant is withdrawn, which really you be the error.

PR #2163, opened 2022-05-26
[CPDLP-1259] Set up UI for dfe support to manage users

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1259

* New "Delivery partner" admin section
* List of all delivery partner users
* Add new delivery partner user

### Guidance to review

* Login as admin, go to "Delivery partner" section

### DIFF
--- a/app/controllers/admin/delivery_partners/users_controller.rb
+++ b/app/controllers/admin/delivery_partners/users_controller.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module Admin
+  module DeliveryPartners
+    class UsersController < Admin::BaseController
+      skip_after_action :verify_authorized
+      skip_after_action :verify_policy_scoped
+
+      def index
+        sorted_users = User.delivery_partner_users.name_order
+        @pagy, @users = pagy(sorted_users, page: params[:page], items: 20)
+      end
+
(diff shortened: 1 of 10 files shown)


app/controllers/admin/delivery_partners/users_controller.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Admin
+  module DeliveryPartners
+    class UsersController < Admin::BaseController
```
asmega, line 5: I don't think we want to inherit from `Admin::BaseController` or be in the `Admin` module. This area is not an "admin" area. its just an area for DP users. so could we split this out?

app/forms/delivery_partners/user_form.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module DeliveryPartners
+  class UserForm
+    include ActiveModel::Model
+    include ActiveModel::Serialization
```
asmega, line 6: do you need `ActiveModel::Serialization`?

asmega (approved):

PR #2165, opened 2022-05-27
[CPDLP-1377] - Expose private child care URN for NPQ application response

StupidCodeFactory:
### Context

- [CPDLP-1377](https://dfedigital.atlassian.net/browse/CPDLP-1377)

### DIFF
--- a/app/serializers/api/v1/npq_application_serializer.rb
+++ b/app/serializers/api/v1/npq_application_serializer.rb
@@ -8,24 +8,25 @@ module Api
       include JSONAPI::Serializer
       include JSONAPI::Serializer::Instrumentation
 
-      attributes :participant_id,
-                 :full_name,
+      attributes :course_identifier,
+                 :created_at,
+                 :eligible_for_funding,
                  :email,
                  :email_validated,
+                 :employer_name,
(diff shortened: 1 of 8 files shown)


app/serializers/api/v1/npq_application_serializer.rb
```diff
@@ -8,24 +8,25 @@ class NPQApplicationSerializer
...
-                 :funding_choice,
-                 :course_identifier,
-                 :works_in_school,
-                 :employer_name,
-                 :employment_role,
                  :status,
-                 :created_at,
-                 :updated_at
+                 :updated_at,
+                 :works_in_school
```
StupidCodeFactory, line 30: Note: Alphabetically rendered the attributes

app/serializers/api/v2/npq_application_serializer.rb
```diff
@@ -4,60 +4,9 @@
 
 module Api
   module V2
-    class NPQApplicationSerializer
```
asmega, line 7: I wouldn't worry about V2 it's not released and fallen so far behind at the moment

asmega (approved):

PR #2166, opened 2022-05-27
Partnership analytics bug fixes

lazydevorg:
### Context

- Ticket: no ticket


db/analytics_migrate/20220526112409_create_analytics_ecf_partnership.rb
```diff
@@ -2,7 +2,9 @@
 
 class CreateAnalyticsECFPartnership < ActiveRecord::Migration[6.1]
   def change
-    create_table :ecf_partnership_analytics do |t|
+    create_table :ecf_partnerships do |t|
+      t.uuid :partnership_id
```
asmega, line 6: just a note that you are modifying an already merged therefore deployed migration (for dev). therefore dev will not be in the correct state unless you do this by hand

asmega (approved):

PR #2167, opened 2022-05-27
[CPDLP-1272] - Fetch NPQ applications for cohort 2022

StupidCodeFactory:
### Context

- Ticket:


asmega (approved):

spec/requests/api/v1/npq_applications_spec.rb
```diff
@@ -4,6 +4,7 @@
 require "csv"
 
 RSpec.describe "NPQ Applications API", :with_default_schedules, type: :request do
+  let(:cohort) { Cohort.current || create(:cohort, :current) }
   let(:npq_lead_provider) { create(:npq_lead_provider) }
```
ericaporter, line 8: Just wondering why we need the option of either here

ericaporter (approved):
Looks good, tiny query on lets

PR #2168, opened 2022-05-27
[CST-775] Enable withdrawn participants to transfer to FIP school

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-775)

Currently we prevent withdrawn participants from transferring in the FIP to FIP transfer journey because of complications regarding making their `training_status` "active" with use of the `ParticipantProfileState`.  Work is underway to enable this in the LP team, in the meantime we can enable transfers and set the `training_status` without using the `ParticipantProfileState`.

There have been some assumptions made in the transfer journey, given that the participant has been withdrawn:
1. The participant is treated as if they are at the same LP and DP as the school (no options to choose different programme)
2. Only the new LP is notified of the transfer in

I've added a placeholder to reactivate the `ParticipantProfileState` that is currently commented out to the `Induction::Enrol` service

### DIFF
--- a/app/services/induction/enrol.rb
+++ b/app/services/induction/enrol.rb
@@ -2,15 +2,21 @@
 
 class Induction::Enrol < BaseService
   def call
-    participant_profile.induction_records.create!(
-      induction_programme: induction_programme,
-      start_date: start_date,
-      induction_status: :active,
-      schedule: participant_profile.schedule,
-      preferred_identity: preferred_identity,
-      mentor_profile: mentor_profile,
-      school_transfer: school_transfer,
-    )
+    ActiveRecord::Base.transaction do
+      participant_profile.induction_records.create!(
+        induction_programme: induction_programme,
+        start_date: start_date,
+        training_status: :active,
+        induction_status: :active,
+        schedule: participant_profile.schedule,
+        preferred_identity: preferred_identity,
+        mentor_profile: mentor_profile,
+        school_transfer: school_transfer,
+      )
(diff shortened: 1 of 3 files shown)


app/services/induction/enrol.rb
```diff
@@ -2,15 +2,21 @@
...
+        induction_status: :active,
+        schedule: participant_profile.schedule,
+        preferred_identity: preferred_identity,
+        mentor_profile: mentor_profile,
+        school_transfer: school_transfer,
+      )
+      # TODO: here would be a convenient place to reactivate a withdrawn / deferred profile
+      # However this needs some checking/work by the LP team
+      # reactivate_profile! unless participant_profile.active_training_status?
+    end
```
ericaporter, line 19: I think the best plan is for us to remove the validation so that we don't have to check if there is an active state on the participant_profile and we just allow it

app/services/induction/enrol.rb
```diff
@@ -41,4 +47,10 @@ def preferred_identity
   def schedule_start_date
     participant_profile.schedule.milestones.first.start_date
   end
+
+  def reactivate_profile!
+    ParticipantProfileState.create!(participant_profile: participant_profile,
```
ericaporter, line 52: If we remove the validations, I might consider renaming this method because we are technically not reactivating anything - we are creating a new induction record and record the new participant_profile_state - will consider this is another PR

ericaporter (approved):
Looks good, a few things to look at for the LP team in subsequent PR

PR #2190, opened 2022-06-06
NPQ 2022 acceptance and schedule

asmega:
### Context

- Ticket: n/a
- When accepting a 2022 NPQ application it was assigning a 2021 schedule

- Accepting a 2022 NPQ application now assigns a 2022 schedule

### Guidance to review

- none

### DIFF
--- a/app/models/npq_course.rb
+++ b/app/models/npq_course.rb
@@ -7,16 +7,16 @@ class NPQCourse < ApplicationRecord
     pluck(:identifier)
   end
 
-  def self.schedule_for(npq_course)
+  def self.schedule_for(npq_course:, cohort: Cohort.find_by!(start_year: 2021))
     case npq_course.identifier
     when *Finance::Schedule::NPQLeadership::IDENTIFIERS
-      Finance::Schedule::NPQLeadership.default
+      Finance::Schedule::NPQLeadership.find_by!(cohort: cohort)
     when *Finance::Schedule::NPQSpecialist::IDENTIFIERS
-      Finance::Schedule::NPQSpecialist.default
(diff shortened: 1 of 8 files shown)


app/models/npq_course.rb
```diff
@@ -7,16 +7,16 @@ def self.identifiers
     pluck(:identifier)
   end
 
-  def self.schedule_for(npq_course)
+  def self.schedule_for(npq_course:, cohort: Cohort.find_by!(start_year: 2021))
```
asmega, line 10: i have left this as class method. i'm not sure why this wasn't implemented an instance method on the course in the first place as opposed to a class method that takes an instance of itself

StupidCodeFactory: historically this was on the Schedule class because it was doing lots of lookup on the constants inside the Schedule namespace, then it got moved to this class. It would indeed be a slightly nicer

app/models/participant_profile/npq.rb
```diff
@@ -34,7 +34,7 @@ def participant_type
     end
 
     def fundable?
-      npq_application&.eligible_for_funding
+      npq_application&.eligible_for_dfe_funding
```
StupidCodeFactory, line 37: should this change be in this PR if it's in  https://github.com/DFE-Digital/early-careers-framework/pull/2186 ?

StupidCodeFactory (approved):

app/models/participant_profile/npq.rb
asmega: the commit is the same so don't think it will cause an issue

PR #2202, opened 2022-06-08
E2E Scenario fixes due to StatementLineItem data model

mattonfoot:
### Context

Changes statement name for checks in finance portal to October 2022 and removes the statement attribution step which is now redundant due to the changes in the declaration API.

### Guidance to review

PR should be green


spec/support/features/steps/changes_of_circumstance_steps.rb
```diff
@@ -288,14 +288,10 @@ def when_developers_transfer_the_deferred_participant(sit_name, participant_name
       end
     end
 
-    def and_eligible_training_declarations_are_made_payable(statement_name)
+    def and_eligible_training_declarations_are_made_payable
       next_ideal_time @timestamp + 3.days
       travel_to(@timestamp) do
```
StupidCodeFactory, line 293: I'm wondering whether it wouldn't have been a bit more handy to to pass the timestamp at which we want the actions to be performed as an argument so that it is explicit, as opposed to have to  keep a mental pointer to an instance variable that may be mutated at any point in time.

StupidCodeFactory (approved):

spec/support/features/steps/changes_of_circumstance_steps.rb
mattonfoot: so move next ideal time out in to the actual GIVEN WHEN in the Scenario rather than in the steps?

i can do that, I have considered it

PR #2205, opened 2022-06-08
[CPDLP-1386] Enable support user to manage (edit name/email and remove) existing DP users

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1386

* Edit DP user
* Delete DP user

### Guidance to review

* Login as admin user and go to DP from top menu

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -292,7 +292,11 @@ Rails.application.routes.draw do
     end
 
     namespace :delivery_partners, path: "delivery-partners" do
-      resources :users, only: %i[index new create]
+      resources :users, only: %i[index new create edit update destroy] do
+        member do
+          get "delete", action: :delete
+        end
+      end
     end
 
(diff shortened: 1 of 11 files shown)


config/routes.rb
```diff
@@ -292,7 +292,11 @@
     end
 
     namespace :delivery_partners, path: "delivery-partners" do
-      resources :users, only: %i[index new create]
+      resources :users, only: %i[index new create edit update destroy] do
+        member do
+          get "delete", action: :delete
```
asmega, line 297: why is this a `get`?

asmega: oh ignore me i see it now, its show the delete page

asmega (approved):

config/routes.rb
mooktakim: @asmega this is a confirmation page because destroy is done

PR #2209, opened 2022-06-09
[CPDLP-1055] ECF Statement handles clawbacks

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1055

- ECF statements are now more interested in `Finance::StatementLineItem` rather than the declaration itself
- Removed `cache_original_value` as it does not appear to be used, I can introduce it back at a later date
- For ECF statements no longer use aggregators nor orchestrators
- The above has been re-implemented from the ground up with calculators instead
- These new classes have been plumbed into the frontend interface
- Uplift cap has been rounded upwards to the nearest £100 to prevent any partial uplift payments this makes clawback logic substantially easier
- I have not pruned the dead/dormant code of orchestrators and aggregators to prevent bloat of this PR

### Guidance to review

- Only via the backend can clawbacks can be introduced. There is no mechanism for this yet. The reason being the system must be able to handle clawbacks before they are introduced

### DIFF
--- a/app/models/finance/statement_line_item.rb
+++ b/app/models/finance/statement_line_item.rb
@@ -13,16 +13,23 @@ class Finance::StatementLineItem < ApplicationRecord
     paid: "paid",
     voided: "voided",
     ineligible: "ineligible",
+    awaiting_clawback: "awaiting_clawback",
+    clawed_back: "clawed_back",
   }
 
   scope :billable, -> { where(state: %w[eligible payable paid]) }
+  scope :refundable, -> { where(state: %w[awaiting_clawback clawed_back]) }
 
   validate :validate_single_billable_relationship, on: [:create]
+
+  def billable?
+    %w[eligible payable paid].include?(state)
+  end
 
 private
 
   def validate_single_billable_relationship
-    if Finance::StatementLineItem
+    if billable? && Finance::StatementLineItem
       .where(participant_declaration:)
       .billable
       .exists?
--- a/app/models/finance/statement_line_item.rb
+++ b/app/models/finance/statement_line_item.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class Finance::StatementLineItem < ApplicationRecord
+  BILLABLE_STATES = %w[eligible payable paid].freeze
+  REFUNDABLE_STATES = %w[awaiting_clawback clawed_back].freeze
+
   self.table_name = "statement_line_items"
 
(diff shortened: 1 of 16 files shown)


app/models/finance/statement_line_item.rb
```diff
@@ -13,16 +13,23 @@ class Finance::StatementLineItem < ApplicationRecord
...
+    clawed_back: "clawed_back",
   }
 
   scope :billable, -> { where(state: %w[eligible payable paid]) }
+  scope :refundable, -> { where(state: %w[awaiting_clawback clawed_back]) }
 
   validate :validate_single_billable_relationship, on: [:create]
 
+  def billable?
+    %w[eligible payable paid].include?(state)
```
jamesrose, line 26:
```suggestion
    BILLABLE_STATES.include?(state)
```

How about we push this up so it can be shared by this and:

```
scope :billable, -> { where(state: BILLABLE_STATES) }
```

app/models/finance/statement_line_item.rb
```diff
@@ -13,16 +13,23 @@ class Finance::StatementLineItem < ApplicationRecord
...
+  end
+
 private
 
   def validate_single_billable_relationship
-    if Finance::StatementLineItem
+    if billable? && Finance::StatementLineItem
```
jamesrose, line 32: What do you need to check for `Finance::StatementLineItem`?

asmega: i want to make sure a declaration does not get associated to more that one statement for each direction ie:
- if money going out (eligible, payable, paid) we at most only have one line item
- if money coming in (awaiting_clawback, clawed_back) we at most only have one line item

jamesrose: Ah, sorry, I misread that. I didn't see it was a multi-line condition.

app/services/finance/ecf/output_calculator.rb
```diff
@@ -0,0 +1,245 @@
...
+      attr_reader :statement
+
+      def initialize(statement:)
+        @statement = statement
+      end
+
+      def banding_breakdown
+        return @banding_breakdown if @banding_breakdown
+
+        array = declaration_types.each.map do |declaration_type|
```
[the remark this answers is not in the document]
asmega: I've renamed the local variable to `bandings`

jamesrose (approved):

PR #2210, opened 2022-06-09
Add nil safety to full name on form.

xswoodsx:
This was causing an error in [sentry](https://sentry.io/organizations/dfe-teacher-services/issues/3280306034/?project=5748989) likely caused by the user manually entering the url and there being no name in the form.

### DIFF
--- a/app/views/schools/add_participants/cannot_add.html.erb
+++ b/app/views/schools/add_participants/cannot_add.html.erb
@@ -5,7 +5,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <span class="govuk-caption-xl"><%= @school.name %></span>
-    <h1 class="govuk-heading-xl"> You cannot add <%= add_participant_form.full_name.titleize %></h1>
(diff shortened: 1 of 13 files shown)


tonyheadford (commented):
I wonder if this is just burying the actual problem?  I think we should perhaps check/track the form data and prevent jumping around by redirecting to the first step or something similar if the session data isn't there, what do you think?

xswoodsx:
> I wonder if this is just burying the actual problem? I think we should perhaps check/track the form data and prevent jumping around by redirecting to the first step or something similar if the session data isn't there, what do you think?

Yeah, it feels like a lazy solution, there is something in the custom MultiStep controller to track the session and steps. i.e if you complete the journey and the use the back button it redirects back to the dashboard. However when you manually enter the url, it behaves differently. Looks like this has always happened, it's just me titleizing the name has shown the error. 

I'll look into it more and see what it can do.

jamesrose:
@xswoodsx Would you like to keep this PR open?

asmega:
I'm closing this PR as seems to be stale. Please re-open if needed

PR #2218, opened 2022-06-10
[CPDLP-1411] Do not filter schools in admin for sandbox

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1411

* Updated `SchoolPolicy` to not filter when in `sandbox` environment as admin.

### DIFF
--- a/app/policies/school_policy.rb
+++ b/app/policies/school_policy.rb
@@ -17,6 +17,7 @@ class SchoolPolicy < ApplicationPolicy
 
   class Scope < Scope
     def resolve
+      return scope if Rails.env.sandbox? && user.admin?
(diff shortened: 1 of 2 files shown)


asmega (approved):

app/policies/school_policy.rb
```diff
@@ -17,6 +17,7 @@ def show?
 
   class Scope < Scope
     def resolve
+      return scope if Rails.env.sandbox? && user.admin?
```
jamesrose, line 20: Could we not just make the sandbox filters eligible?

PR #2222, opened 2022-06-13
[Bugfix] Use InductionRecord.training_status when available for status tag component

tonyheadford:
### Context

The admin ui is inconsistent with the view the SIT sees when a participant has withdrawn training status.  The status tag component uses the `training_status` from the participant profile not the induction record so this PR adds the induction record to the call when available.

### DIFF
--- a/app/components/admin/participants/table_row.rb
+++ b/app/components/admin/participants/table_row.rb
@@ -15,7 +15,7 @@ module Admin
 
     private
 
-      attr_reader :profile, :induction_record
+      attr_reader :profile
 
       def induction_record
         return unless @profile.ect? || @profile.mentor?
--- a/app/components/admin/participants/table_row.rb
+++ b/app/components/admin/participants/table_row.rb
@@ -7,6 +7,7 @@ module Admin
 
       def initialize(profile:)
         @profile = profile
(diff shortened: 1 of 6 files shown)


app/components/admin/participants/table_row.rb
```diff
@@ -7,6 +7,7 @@ class TableRow < BaseComponent
 
       def initialize(profile:)
         @profile = profile
+        @induction_record = get_induction_record
```
jamesrose, line 10:
```suggestion
```

app/components/participant_status_tag_component.rb
```diff
@@ -35,6 +36,10 @@ def tag_attributes
     { text: "Contacting for information", colour: "grey" }
   end
 
+  def training_status_withdrawn?
+    (induction_record || profile).training_status_withdrawn?
```
lazydevorg, line 40: induction_record seems to be used with `&` operator almost everywhere. Should it be used here as well to make it nil safe?

tonyheadford: If induction_record is nil here it will default to using the profile

lazydevorg (approved):

PR #2225, opened 2022-06-14
[CPDLP-1405] - refactor error messages and API service classes

StupidCodeFactory:
### Context

- [Ticket](https://dfedigital.atlassian.net/browse/CPDLP-1405):

1. add active model validation to the service class, than collect errors and serialise the errors if the service class is invalid.
  this allow to collect several errors at once a lead provider may have made when submitting the declaration.
  This will also allow us to do further introspection of the errors and possibly improve the error messages.
2. Introduction of few new factories `ect` and `mentor` that now use the services class which drastically reduces the manual setup and make the test ever so slightly reliable and more maintainable, when the logic in those service classes will change the tests will automatically follow the same logic.
3. update to `participant_declarations` factory. Those now use the service class create then and then transition them to the desired state using various traits. The thing that I'm not massively keen on are the checks made to automatically create the upcoming output fee statement that they'll be attached to. But I will revisit this later in another PR where I'll be touching those.


lib/cpdlp/model.rb
```diff
@@ -0,0 +1,12 @@
+module CPDLP
```
asmega, line 1: can we just embed this file? seems like a pointless module? also i think we should stop using `/lib` with `require`

app/controllers/api/api_controller.rb
```diff
@@ -8,7 +8,6 @@ class ApiController < ActionController::API
     rescue_from ActionController::UnpermittedParameters, with: :unpermitted_parameter_response
     rescue_from ActionController::BadRequest, with: :bad_request_response
     rescue_from ActiveRecord::StatementInvalid, with: :bad_request_response
-    rescue_from ArgumentError, with: :bad_request_response
```
StupidCodeFactory, line 11: this will hide genuine application repose and not log them to Sentry...

app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -10,11 +10,18 @@ class ParticipantDeclarationsController < Api::ApiController
       include ApiFilter
 
       def create
-        params = HashWithIndifferentAccess.new({ cpd_lead_provider: }).merge(permitted_params["attributes"] || {})
+        if permitted_params.dig("attributes", "declaration_type")
```
StupidCodeFactory, line 13: used this to get path to run some `spec/request/api/v1/participants_spec.rb` to craft the errors message response serialisation.

app/models/participant_profile/npq.rb
```diff
@@ -8,6 +8,8 @@ class NPQ < ParticipantProfile
 
     has_one :npq_application, foreign_key: :id
 
+    has_many :participant_declarations, class_name: "ParticipantDeclaration::NPQ"
```
StupidCodeFactory, line 11: ignore that

app/serializers/api/v1/active_model_errors_serializer.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      def self.from(service)
+        {
+          errors: service
+            .errors
+            .messages
+            .map { |title, detail| new(title:, detail: detail.join(", ")).serializable_hash },
+        }
+      end
+    end
```
StupidCodeFactory, line 22: This allows to render `all` the possible errors with given request. Before we could only render one at a time., due to the errors being handle with exceptions.

app/validators/induction_record_validator.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    record.errors.add(:cpd_lead_provider, I18n.t(:invalid_participant))
+  end
+
+  def relevant_induction_record?(record)
+    InductionRecord
+      .joins(:participant_profile, induction_programme: { school_cohort: [:cohort], partnership: [:lead_provider] })
+      .where(participant_profile: record.participant_profile)
+      .where(induction_programme: { partnerships: { lead_provider: record.cpd_lead_provider.lead_provider } })
+      .where(induction_programme: { school_cohorts: { cohort: Cohort.current } })
```
asmega, line 21: i think there will be a bug if the scope to `Cohort.current` when next cohort kicks in. do we need to include cohort in this query?

app/validators/declaration_date_validator.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+class DeclarationDateValidator < ActiveModel::Validator
```
StupidCodeFactory, line 3: a custom validator, to make separate the validation of a component of the request request (i.e declaration_date) so that it does not clutter the service class.

app/validators/induction_record_validator.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class InductionRecordValidator < ActiveModel::Validator
```
StupidCodeFactory, line 3: Probably rename this `HasInductionRecordValidator`

spec/factories/early_career_teachers/create.rb
```diff
@@ -0,0 +1,17 @@
...
+    school_cohort
+
+    initialize_with do
+      EarlyCareerTeachers::Create.call(
+        email: user.email,
+        full_name: user.full_name,
+        mentor_profile_id: mentor.id,
+        school_cohort:,
+      )
+    end
```
StupidCodeFactory, line 15: Use production code to create realistic test setup and not have to maintain complicated factories that will drift away from production has the time goes not. It's not a silver bullet but it's still more maintainable.

app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -10,11 +10,18 @@ class ParticipantDeclarationsController < Api::ApiController
...
+          service = RecordDeclarations::Started::EarlyCareerTeacher
```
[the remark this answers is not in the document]
StupidCodeFactory: @asmega that's just for demo purposes, see comment above ^^ - that's not the final implementation

app/services/record_declarations/base.rb
```diff
@@ -1,164 +1,63 @@
...
+    attribute :declaration_date, :datetime
+    attribute :declaration_type
+    attribute :participant_id
 
+    validates :participant_identity, presence: { message: I18n.t(:invalid_participant, scope: model_name.i18n_key) }
     validates :declaration_date, :declaration_type, presence: true
-    validates :parsed_date, future_date: true, allow_blank: true
-    validate :date_has_the_right_format
-    validate :validate_schedule_present
-    validate :validate_milestone_exists
```
asmega, line 22: did we lose some validations are a result of this change? eg `validate_milestone_exists` seems to have gone?

app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -10,11 +10,18 @@ class ParticipantDeclarationsController < Api::ApiController
...
+                      .new(permitted_params["attributes"].merge(cpd_lead_provider:))
 
-        log_schema_validation_results
+          render_from_service(service, ParticipantDeclarationSerializer)
+        else
```
asmega, line 18: why do we want/need this `else` branch? are we not able to delegate to the service?

app/services/record_declarations/base.rb
```diff
@@ -1,164 +1,63 @@
...
-    def validate_milestone!
-      if parsed_date < milestone.start_date.beginning_of_day
-        raise ActionController::ParameterMissing, I18n.t(:declaration_before_milestone_start)
-      end
-
-      if milestone.milestone_date.present? && (milestone.milestone_date.end_of_day <= parsed_date)
-        raise ActionController::ParameterMissing, I18n.t(:declaration_after_milestone_cutoff)
-      end
+    def participant_identity
+      @participant_identity ||= ParticipantIdentity.find_by(external_identifier: participant_id)
```
StupidCodeFactory, line 42: replace with `Identity.find_user_by(:id participant_id)`

db/migrate/20220616131226_add_unique_index_to_statement_line_items.rb
```diff
@@ -0,0 +1,11 @@
+class AddUniqueIndexToStatementLineItems < ActiveRecord::Migration[6.1]
+  def change
+    safety_assured do
+      remove_index :statement_line_items, :statement_id
```
asmega, line 4: are we sure we want to remove this? if we want to grab all the line items for a particular statement it now needs to scan the whole table?

app/services/record_declaration.rb
```diff
@@ -0,0 +1,142 @@
...
+  attribute :declaration_date, :datetime
+  attribute :declaration_type
+  attribute :participant_id
+
+  before_validation :declaration_attempt
+
+  validates :declaration_date, :declaration_type, presence: true
+  validates :declaration_date, future_date: true, declaration_date: true, allow_blank: true
+  validates :participant_id, participant_not_withdrawn: true
+  validates :cpd_lead_provider, induction_record: true
```
StupidCodeFactory, line 19: TODO: find a better name for this

app/controllers/api/v2/participant_declarations_controller.rb
```diff
@@ -49,8 +49,12 @@ def show
       def query_scope
         scope = ParticipantDeclaration.for_lead_provider(cpd_lead_provider)
         scope = scope.where("user_id = ?", participant_id_filter) if participant_id_filter.present?
-        scope = scope.where("updated_at > ?", updated_since) if updated_since.present?
-        scope
+
+        if updated_since.present?
+          scope.where("updated_at > ?", updated_since)
```
asmega, line 54: `scope` needs to be mutated here otherwise as it stands this line won't do anything

app/models/participant_declaration.rb
```diff
@@ -150,6 +150,7 @@ def duplicate_declarations
       .where.not(user_id:, id:)
       .where(state: %w[submitted eligible payable paid])
       .where(
+        cpd_lead_provider:,
```
asmega, line 153: does this not now change the behaviour and after a transfer lead provider B can be post for the same declarations and we end up double paying?

app/models/finance/statement_line_item.rb
```diff
@@ -1,16 +1,18 @@
...
-  self.table_name = "statement_line_items"
-
   belongs_to :statement
   belongs_to :participant_declaration
 
+  scope :billable, -> { eligible.or(payable).or(paid) }
+  scope :refundable, -> { awaiting_clawback.or(clawed_back) }
+
   enum state: {
-    submitted: "submitted",
```
StupidCodeFactory, line 13: removed this as no submitted declaration can be added to a statement

app/models/participant_declaration/ecf.rb
```diff
@@ -2,20 +2,4 @@
...
-    previous_induction_record = participant_profile.relevant_induction_record(lead_provider: cpd_lead_provider.lead_provider)
-
-    return if previous_induction_record.nil?
-    return if previous_induction_record.end_date.nil?
-    return if participant_profile.current_induction_record == previous_induction_record
-
-    if previous_induction_record.end_date < declaration_date
-      errors.add(:base, I18n.t("declaration_must_be_before_end_date"))
-    end
-  end
```
StupidCodeFactory, line 20: this was an undesirable validation, this fixes https://dfedigital.atlassian.net/browse/CPDLP-1657

app/services/api/participant_declarations/index.rb
```diff
@@ -1,43 +1,45 @@
 # frozen_string_literal: true
 
 module Api
-  class ParticipantDeclarations::Index
```
StupidCodeFactory, line 4: nested the scope as this created class autoloading errors

app/services/identity.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "identity/create"
```
StupidCodeFactory, line 3: Fixes class autoloading problem (particularly in factories)

app/services/participant_profile_resolver.rb
```diff
@@ -0,0 +1,29 @@
...
+  class << self
+    def call(participant_identity, course_identifier)
+      return unless participant_identity
+
+      if ParticipantProfile::ECT::COURSE_IDENTIFIERS.include?(course_identifier)
+        participant_identity
+          .participant_profiles
+          .active_record
+          .ects
+          .first
```
StupidCodeFactory, line 13: this code was exisiting in other modules. This is however wrong! this randomly chooses a profile we don't know which one it is. Participants are supposed to have only one ECF profile of each type at most but we know this is not the case.

app/services/record_declaration.rb
```diff
@@ -0,0 +1,151 @@
...
+
+      Finance::Statement.attach(participant_declaration) unless participant_declaration.submitted?
+
+      declaration_attempt.update!(participant_declaration:)
+    end
+
+    participant_declaration
+  end
+
+  def declaration_date=(raw_declaration_date)
```
StupidCodeFactory, line 56: We do some defensive parsing as I suspect lead provider don't always respect the date format. Possibly we could remove this in the future?

db/migrate/20220616131226_add_unique_index_to_statement_line_items.rb
StupidCodeFactory: this is covered in those indices https://github.com/DFE-Digital/early-careers-framework/pull/2225/files#diff-8d6085c3e87baff35cf7da33c61ceb84b635522fe5b34f0b4e518a03e9c7478eR9-R10

app/services/participant_declarations/mark_as_paid.rb
```diff
@@ -0,0 +1,25 @@
...
+
+module ParticipantDeclarations
+  class MarkAsPaid
+    def initialize(statement)
+      self.statement = statement
+    end
+
+    def call(participant_declaration)
+      ParticipantDeclaration.transaction do
+        participant_declaration.make_paid!
```
asmega, line 11: as this has been extracted out and can therefore be called by anyone i personally would opt to add some guards here against someone who accidentally calls this class with a declaration which may not be in the right state to use this class. or maybe it's not a problem and has already been guarded elsewhere if let's say someone passed in a `submitted` or `awaiting_clawback` declaration?

app/services/participant_profile_resolver.rb
```diff
@@ -0,0 +1,29 @@
...
+          .participant_profiles
+          .active_record
+          .mentors
+          .first
+      elsif NPQCourse.identifiers.include?(course_identifier)
+        participant_identity
+          .participant_profiles
+          .joins(participant_identity: { npq_applications: :npq_course })
+          .where(npq_courses: { identifier: course_identifier })
+          .first
```
asmega, line 25: there are some workarounds that have been implemented to handle bad data aka duplicates. this appears to be missing here which i foresee will add more problems to the pile if they are removed. For example there exists code elsewhere where the calling provider is passed in, this extra information can be used to determine the correct profile from their duplicates. without this we may end up attaching declarations to the incorrect profile?

app/services/participant_declarations/mark_as_paid.rb
mooktakim: We don't really have proper statemachine here, normally this would raise an error if you aren't allowed this state transition, but unfortunately, these fail silently (`make_paid!` only does it when `payable?`)

app/services/record_declaration.rb
```diff
@@ -0,0 +1,151 @@
...
+    return unless schedule
+
+    schedule.milestones.find_by(declaration_type:)
+  end
+
+  def schedule
+    participant_profile&.schedule
+  end
+
+private
```
asmega, line 79: https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/services/record_declarations/base.rb#L94

I think the code has diverged and missing the above validation

app/validators/induction_record_validator.rb
asmega: I've changed this on the branch i'm working on from `Cohort.current` to `Cohort.where(start_year: 2021..)`

app/services/record_declaration.rb
```diff
@@ -0,0 +1,151 @@
+# frozen_string_literal: true
+
+class RecordDeclaration
```
asmega, line 3: So it likes this PR only covers the endpoint for recording declarations. Will there be further PRs to cover off the other endpoints such as defer/resume and the classes that power those?

StupidCodeFactory: Yes there were the stories for the rests of class.

app/services/record_declaration.rb
StupidCodeFactory: well spotted! re-added

app/services/participant_declarations/mark_as_paid.rb
StupidCodeFactory: @asmega I've added a `line_item.payable?` guard close and some specs  around that

app/services/participant_profile_resolver.rb
StupidCodeFactory: @asmega this class is mere copy past of what is already existing:
https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/services/participants/early_career_teacher.rb#L11-L15

https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/services/participants/mentor.rb#L11-L15

https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/services/participants/npq.rb#L20-L31

I think this what has been causing us a bunch of pain. Hence I've collected those in a class, we will need to address those.

asmega: See this PR for more detail

> this class is mere copy past of what is already existing

> https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/services/participants/npq.rb#L20-L31

It's not though, as you see there is an additional `where` clause in the code you referenced but it's not present in the version of this PR

app/services/participant_profile_resolver.rb
peteryates: Is this problem logged in a ticket for a follow up?

app/services/record_declaration.rb
```diff
@@ -0,0 +1,170 @@
...
+
+  def validate_evidence_held?
+    declaration_type.present? && declaration_type != "started"
+  end
+
+  def original_participant_declaration
+    @original_participant_declaration ||= participant_declaration.duplicate_declarations.order(created_at: :asc).first
+  end
+
+  def validates_billable_slot_available
```
peteryates, line 157: Is this method called? I can't see from where.

app/validators/declaration_date_validator.rb
```diff
@@ -0,0 +1,65 @@
...
```
peteryates, line 3: Worth covering this validator with specs?

app/validators/induction_record_validator.rb
```diff
@@ -0,0 +1,33 @@
...
```
peteryates, line 3: I think this validator should have some specs too.

peteryates (commented):
Overall I think this is a big step in the right direction. Happy to see this go in once we've had some more eyes on it.

app/services/record_declaration.rb
asmega: the implementation has also changed. i'm not sure the new implementation does what it was intended to do? `participant_declaration` looks to return an `ActiveRecord` object, the code then chains `where` which I think is not valid

app/services/record_declaration.rb
```diff
@@ -0,0 +1,170 @@
...
+
+  attr_writer :raw_declaration_date
+
+  def participant_profile_for_course_identifier
+    return unless participant_identity
+  end
+
+  def declaration_attempt
+    if participant_id
+      @declaration_attempt ||= ParticipantDeclarationAttempt.create!(
```
cwrw, line 94: Do we risk creating an invalid `ParticipantDeclarationAttempt` record or one with empty attributes if we run this before validating missing params except the participant id?

app/services/record_declaration.rb
```diff
@@ -0,0 +1,170 @@
...
+              in: ->(record_declaration) { record_declaration.participant_profile.class::VALID_EVIDENCE_HELD },
+              if: :validate_evidence_held?,
+              message: I18n.t(:invalid_evidence_type),
+            }
+
+  validate :validate_milestone_exists
+
+  attr_reader :raw_declaration_date
+
+  DECLARATION_TYPE_WITH_EVIDENCE_HELD = %w[
```
cwrw, line 34: Are we using this anywhere I might have missed?

app/services/record_declaration.rb
```diff
@@ -0,0 +1,170 @@
...
+
+  validates :declaration_date, :declaration_type, presence: true
+  validates :declaration_date, future_date: true, declaration_date: true, allow_blank: true
+  validates :participant_id, participant_not_withdrawn: true
+  validates :cpd_lead_provider, induction_record: true
+  validates :course_identifier, course: true
+  validates :evidence_held,
+            presence: { message: I18n.t(:missing_evidence_held), if: :validate_evidence_held? },
+            inclusion: {
+              in: ->(record_declaration) { record_declaration.participant_profile.class::VALID_EVIDENCE_HELD },
```
cwrw, line 25: Could we ensure that a participant profile exists before running this validation? An example to this would be: If we leave out `course_identifier` from the params we get a 500 with the error: `NameError (uninitialized constant NilClass::VALID_EVIDENCE_HELD` where previously we would get `"The property '#/course_identifier' must be an available course to '#/participant_id'"`

app/validators/declaration_date_validator.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def date_has_the_right_format(record)
+    return if record.declaration_date.blank?
+    return if record.raw_declaration_date.match?(RFC3339_DATE_REGEX)
+
+    record.errors.add(:declaration_date, I18n.t(:invalid_declaration_date))
+  end
+
+  def declaration_within_milestone(record)
+    return unless record.milestone
```
cwrw, line 22: Could we also guard against `declaration_date` missing as well?

app/models/finance/statement_line_item.rb
```diff
@@ -1,16 +1,18 @@
...
+
   BILLABLE_STATES = %w[eligible payable paid].freeze
   REFUNDABLE_STATES = %w[awaiting_clawback clawed_back].freeze
 
-  self.table_name = "statement_line_items"
-
   belongs_to :statement
   belongs_to :participant_declaration
 
+  scope :billable, -> { eligible.or(payable).or(paid) }
```
cwrw, line 12: these seem to be duplicates of https://github.com/DFE-Digital/early-careers-framework/pull/2225/files#diff-b12d4a990e9638e969b742c993f93cbdf7fec6dba18934aaa85136a137207c7cL23

app/models/induction_programme.rb
```diff
@@ -22,7 +22,8 @@ class InductionProgramme < ApplicationRecord
   has_many :transferred_induction_records, -> { transferred }, class_name: "InductionRecord"
   has_many :participant_profiles, through: :active_induction_records
   has_many :current_participant_profiles, through: :current_induction_records, source: :participant_profile
-
+  has_one :lead_provider, through: :partnership
```
cwrw, line 25: do we still need https://github.com/DFE-Digital/early-careers-framework/pull/2225/files#diff-3e4c2359566ad9ebf06892b48edf98cbf3cb2296b4d3b37b292e2a7527e4e52fL27 now we have the `has_one` through set?

app/models/participant_declaration/ecf.rb
cwrw: could we remove the `declaration_must_be_before_end_date`  since it is no longer used?

app/serializers/api/v1/active_model_errors_serializer.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      attribute :title
+      attribute :detail
+
+      def self.from(service)
+        {
+          errors: service
+            .errors
+            .messages
+            .map { |title, detail| new(title:, detail: detail.uniq.join(", ")).serializable_hash },
```
cwrw, line 18: I just wanted to confirm we are happy moving away from the error structure/titles we have now such as `"Bad or missing parameters"` to the attribute name instead?

app/services/npq/accept.rb
```diff
@@ -74,13 +72,9 @@ def participant_profile
...
         participant_identity: npq_application.participant_identity,
-      )
-
-      find_or_create_initial_profile_state
-    end
-
-    def find_or_create_initial_profile_state
-      ParticipantProfileState.find_or_create_by(participant_profile:)
+      ).tap do |pp|
+        pp.participant_profile_states.create!
```
cwrw, line 76: do we risk creating duplicates as we are no longer finding or creating by?

db/migrate/20220616131226_add_unique_index_to_statement_line_items.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddUniqueIndexToStatementLineItems < ActiveRecord::Migration[6.1]
+  def change
+    safety_assured do
+      remove_index :statement_line_items, :statement_id
+      remove_index :statement_line_items, :participant_declaration_id
+
+      add_index :statement_line_items, %i[statement_id participant_declaration_id state], unique: true, name: :unique_statement_declaration_state
```
cwrw, line 9: do we need a uniqueness validation in the model for those?

app/services/record_declaration.rb
StupidCodeFactory: Fixed it https://github.com/DFE-Digital/early-careers-framework/pull/2225/files#diff-2fc4a0a9043f3924b5b01515d5102b99560a6498e3418b6abd82a7c985a7d36dR144 and added  test.

app/services/record_declaration.rb
StupidCodeFactory: re-ad

spec/factories/finance/schedules.rb
```diff
@@ -4,13 +4,14 @@
...
-          { name: "Output 2 - Retention Point 1", start_date: Date.new(2021, 11, 1), milestone_date: Date.new(2022, 1, 31), payment_date: Date.new(2022, 2, 28), declaration_type: "retained-1" },
-          { name: "Output 3 - Retention Point 2", start_date: Date.new(2022, 2, 1), milestone_date: Date.new(2022, 4, 30), payment_date: Date.new(2022, 5, 31), declaration_type: "retained-2" },
-          { name: "Output 4 - Retention Point 3", start_date: Date.new(2022, 5, 1), milestone_date: Date.new(2022, 9, 30), payment_date: Date.new(2022, 10, 31), declaration_type: "retained-3" },
-          { name: "Output 5 - Retention Point 4", start_date: Date.new(2022, 10, 1), milestone_date: Date.new(2023, 1, 31), payment_date: Date.new(2023, 2, 28), declaration_type: "retained-4" },
-          { name: "Output 6 - Participant Completion", start_date: Date.new(2023, 2, 1), milestone_date: Date.new(2023, 4, 30), payment_date: Date.new(2023, 5, 31), declaration_type: "completed" },
+          { name: "Output 1 - Participant Start", start_date: Date.new(start_year, 9, 1), milestone_date: Date.new(2021, 11, 30), payment_date: Date.new(2021, 11, 30), declaration_type: "started" },
```
cwrw, line 9: could the `milestone_date` and `payment_date` also be relative to `start_year`?

spec/requests/api/v1/ecf_participants_spec.rb
```diff
@@ -322,15 +305,15 @@
...
           expect(ect_row["participant_type"]).to eql "ect"
-          expect(ect_row["cohort"]).to eql partnership.cohort.start_year.to_s
-          expect(ect_row["teacher_reference_number"]).to eql ect.teacher_profile.trn
-          expect(ect_row["teacher_reference_number_validated"]).to eql "false"
-          expect(mentor_row["eligible_for_funding"]).to be_empty
-          expect(mentor_row["pupil_premium_uplift"]).to eql "false"
-          expect(mentor_row["sparsity_uplift"]).to eql "false"
-          expect(mentor_row["training_status"]).to eql "active"
+          expect(ect_row["cohort"]).to eql mentor.mentor_profile.current_induction_record.schedule.cohort.start_year.to_s
+          expect(ect_row["teacher_reference_number"]).to be_empty
```
cwrw, line 311: I'm curious why this and the below are now expected to be empty across this spec and other specs?

cwrw:
I've tested creating/voiding declarations for the current cohort and previous cohort and all is well. A few minor comments left, after that happy to ✅ Worth getting one more round of eyes/testing on it from @asmega @peteryates

app/services/finance/declaration_statement_attacher.rb
```diff
@@ -5,16 +5,13 @@ module Finance
...
-      @participant_declaration = participant_declaration
+    def initialize(participant_declaration)
+      self.participant_declaration = participant_declaration
     end
 
     def call
       ApplicationRecord.transaction do
-        StatementLineItem.create!(
-          statement:,
+        statement.statement_line_items.create_or_find_by(
```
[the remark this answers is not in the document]
StupidCodeFactory: The statement should always exist we've already seeded it until 2024. So I think it is fine. Actually a 500 is probably the more accurate response, since the lead provider params might be valid. So I think this is the correct behaviour

spec/requests/api/v1/ecf_participants_spec.rb
StupidCodeFactory: That was the `teacher_profile` factory that was setting a random TRN. Which I've removed because now it's passed in  in the service classes that creates the `teacher_profile` associated with the that participant.

StupidCodeFactory: and the participants in this test aren't `:eligible_for_funding` which is the trait that would set that up.

cwrw (approved):
Thanks @StupidCodeFactory immense work! A few things we discussed in person as out of scope for this PR but worth revisiting later:
- Cohorts set up in specs (ongoing issue)
- Base in error messages title
- Revisiting when we show certain error messages as could be confusing

spec/components/admin/npq/applications/analysis/table_row_spec.rb
```diff
@@ -1,38 +1,26 @@
...
   end
 
-  component { described_class.new application: }
+  component { described_class.new application: npq_application }
 
-  it { is_expected.to have_content "Full name\n    #{application.user.full_name}" }
-  it { is_expected.to have_link application.user.full_name, href: admin_participant_path(application.profile) }
-  it { is_expected.to have_content "Application Status\n    rejected" }
-  it { is_expected.to have_content npq_course.name }
+  it { is_expected.to have_content "Full name\n    #{npq_application.user.full_name}" }
```
peteryates, line 18: I know this wasn't added in this PR but here we can make it more readable using [normalize_ws](https://www.rubydoc.info/gems/capybara/Capybara%2FNode%2FMatchers:assert_text):

```
it { is_expected.to have_css(
  ".nhsuk-table__cell",
  text: "Full name #{npq_application.user.full_name}",
  normalize_ws: true
) }
```

PR #2241, opened 2022-06-17
[CPDLP-1439] Whitelist voidable states

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1439
- Using a blacklist strategy for this kind of thing is a recipe for bugs. As when new states are introduced this code must be changed to keep up

- Use whitelist to work out if a state is voidable instead

### Guidance to review

- none

### DIFF
--- a/spec/models/participant_declaration_spec.rb
+++ b/spec/models/participant_declaration_spec.rb
@@ -436,4 +436,25 @@ RSpec.describe ParticipantDeclaration, type: :model do
       end
     end
   end
+
+  describe "#voidable?" do
+    [
+      { state: "submitted", voidable: true },
+      { state: "eligible", voidable: true },
+      { state: "payable", voidable: true },
+      { state: "paid", voidable: false },
+      { state: "voided", voidable: false },
+      { state: "ineligible", voidable: true },
+      { state: "awaiting_clawback", voidable: false },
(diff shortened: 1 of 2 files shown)


spec/models/participant_declaration_spec.rb
```diff
@@ -436,4 +436,25 @@
...
+      { state: "submitted", voidable: true },
+      { state: "eligible", voidable: true },
+      { state: "payable", voidable: true },
+      { state: "paid", voidable: false },
+      { state: "voided", voidable: false },
+      { state: "ineligible", voidable: true },
+      { state: "awaiting_clawback", voidable: false },
+      { state: "clawed_back", voidable: false },
+    ].each do |hash|
+      context "when declaration is #{hash[:state]}" do
```
StupidCodeFactory, line 451: not a big fan of dynamically generated examples, becomes a bit tricky if you need to debug one in the middle. 
Would you be able to example those? Thanks.

StupidCodeFactory (approved):

spec/models/participant_declaration_spec.rb
asmega: if you need to debug a single failing test you can change/filter the array to the exact one test you want to run. alternatively you can simply create a new test with the exact setup you want to test.

PR #2249, opened 2022-06-20
[CPDLP-1448] Statements been counting unbillable uplift declarations

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1448
- We have and are counting unbillable unlifts eg those in `ineligible` states

- this fixes the scope so we do not count unbillable declarations for uplifts

### Guidance to review

- n/a

### DIFF
--- a/app/models/participant_declaration.rb
+++ b/app/models/participant_declaration.rb
@@ -80,6 +80,7 @@ class ParticipantDeclaration < ApplicationRecord
   scope :unique_ects,    -> { unique_id.ect }
   scope :unique_mentors, -> { unique_id.mentor }
   scope :unique_uplift,  -> { unique_id.uplift }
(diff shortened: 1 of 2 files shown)


app/models/participant_declaration.rb
```diff
@@ -80,6 +80,7 @@ class ParticipantDeclaration < ApplicationRecord
   scope :unique_ects,    -> { unique_id.ect }
   scope :unique_mentors, -> { unique_id.mentor }
   scope :unique_uplift,  -> { unique_id.uplift }
+  scope :billable_unique_uplift, -> { unique_id.uplift.where(state: %w[eligible payable paid]) }
```
asmega, line 83: this will work for now, but will soon be obsoleted as we should be working off the line item not the declaration

jamesrose (approved):

asmega:
closing as the change are not long applicable with the rollout of new statement calculators build from the ground up

PR #2253, opened 2022-06-20
[CPDLP-1447] Introduce reconciliation to statements

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1447
- Paid statements in the past have not been generated correctly and providers have been overpaid
- We need to be able to account for the overpayments so we can underpay later to balance the books 

- This allows us to reconcile a statement by an arbitrary amount
- This number can be positive or negative
- This gives the ability to ensure all money is accounted for correctly when there has been a mistake
- It has been decided to not render these in the statements for now until designed
- This does mean if someone pulled out a calculator the numbers may not add up as the reconcile amount is not shown

### Guidance to review

- none


jamesrose (approved):
Looks good to me!

PR #2258, opened 2022-06-22
[CPDLP-1055] fix previous statements

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1055
- this is need as there are many subclasses of a statement

- when looking a previous limits we hardcode the subclass we want to use since there are many subclasses
- otherwise does not form correct query to find previous statements

### Guidance to review

- none

### DIFF
--- a/app/models/finance/statement/ecf.rb
+++ b/app/models/finance/statement/ecf.rb
@@ -18,6 +18,13 @@ class Finance::Statement::ECF < Finance::Statement
   def calculator
     @calculator ||= Finance::ECF::StatementCalculator.new(statement: self)
   end
+
+  def previous_statements
+    Finance::Statement::ECF
+      .where(cohort:)
+      .where(cpd_lead_provider:)
+      .where("payment_date < ?", payment_date)
+  end
 end
 
 require "finance/statement/ecf/payable"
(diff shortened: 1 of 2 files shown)


jamesrose (approved):

app/models/finance/statement/ecf.rb
```diff
@@ -18,6 +18,13 @@ def payable!
   def calculator
     @calculator ||= Finance::ECF::StatementCalculator.new(statement: self)
   end
+
+  def previous_statements
+    Finanace::Statement::ECF
```
StupidCodeFactory, line 23:
```suggestion
    cpd_lead_provider.where(cohort:).where("payment_date < ?", payment_date)
```

app/models/finance/statement/npq.rb
```diff
@@ -6,6 +6,13 @@ class Finance::Statement::NPQ < Finance::Statement
   def payable!
     update!(type: "Finance::Statement::NPQ::Payable")
   end
+
+  def previous_statements
+    Finanace::Statement::NPQ
```
StupidCodeFactory, line 11:
```suggestion
    cpd_lead_provider.where(cohort:).where("payment_date < ?", payment_date)
```

Could these be moved up the main `Finance::Statement` class?

StupidCodeFactory: Oh I see that there was already that method there. 
Correct me if I'm wrong, but it would have picked up the both the ECF and NPQ statements correct?

StupidCodeFactory (approved):

PR #2261, opened 2022-06-22
Appropriate bodies

lazydevorg:
### Context

- Ticket: CST-797

* Script to scrape appropriate bodies, create a CSV file and import it
* AppropriateBody reference added to SchoolCohort
* SchoolCohort.appropriate_body_unknown is used to store the choice about not knowing the current appropriate body. The field is used to hide the appropriate body details in the school dashboard.
* The copies of the school cohort setup final steps have all changed.
* The schools dashboard shows an "Add" link in the appropriate body details row that links to a static temporary page.


app/services/induction/set_school_cohort_appropriate_body.rb
```diff
@@ -0,0 +1,28 @@
...
+  def initialize(school_cohort:, appropriate_body_id:, appropriate_body_appointed:)
+    @school_cohort = school_cohort
+    @appropriate_body_id = appropriate_body_id
+    @appropriate_body_appointed = appropriate_body_appointed
+  end
+
+  def set_appropriate_body
+    if !appropriate_body_appointed
+      school_cohort.appropriate_body_unknown = true
+      school_cohort.appropriate_body = nil
```
tonyheadford, line 22: Could we not just use the presence of the `appropriate_body` to indicate whether the appropriate body is known or not?  Or is there something subtle I'm missing?

app/controllers/schools/setup_school_cohort_controller.rb
```diff
@@ -168,7 +195,19 @@ def set_cohort_induction_programme!(programme_choice, opt_out_of_updates: false)
...
-                                                  delivery_partner_to_be_confirmed: delivery_partner_to_be_confirmed?)
+                                                  delivery_partner_to_be_confirmed: delivery_partner_to_be_confirmed?,
+                                                  appropriate_body_appointed: appropriate_body_appointed?,
+                                                  appropriate_body: @setup_school_cohort_form.appropriate_body)
+    end
+
+    def appropriate_body_appointed?
+      @setup_school_cohort_form.appropriate_body_appointed == "yes"
+    end
+
```
tonyheadford, line 206: Could we move this method onto the form?

lazydevorg: 👍

app/services/induction/set_school_cohort_appropriate_body.rb
lazydevorg: @tonyheadford we need to show the "Appropriate body" detail line in the dashboard only for the school cohorts that went through the appropriate body selection. 

To recap:
* Existing school cohorts: don't show the appropriate body detail
* School cohorts with appropriate body unknown: show a detail line with an empty value and an "Add" action
* School cohorts with an appropriate body selected: show the name of the appropriate body with no action

In order to achieve that we need to differentiate the school cohorts where an appropriate body was never selected and the ones where the appropriate body is unknown.

tonyheadford: Gotcha, thought there might be something I was missing 👍

app/services/induction/set_cohort_induction_programme.rb
```diff
@@ -55,4 +63,10 @@ def programme_attrs
 
     attrs
   end
+
+  def set_appropriate_body
+    Induction::SetSchoolCohortAppropriateBody.call(school_cohort:,
+                                                   appropriate_body_id: appropriate_body,
+                                                   appropriate_body_appointed:)
```
tonyheadford, line 70: Do you think it might make sense to remove this call from here and invoke it from the caller?  Would keep the concerns a bit more focussed and clean up the number of params being passed in?
Fine to leave it if you think we'll always need to set the AB info when we are setting the programme choice

tonyheadford (approved):
Looks good, could you check with Jo around smart quotes there seem to be a few of `'` in the views and I suspect they should be `‘` and `’` instead

PR #2270, opened 2022-06-24
[CPDLP-1464] Permit declarations after a clawback

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1464
- After something has clawed back we want providers to be able to submit again for this event

- After a declaration is `awaiting_clawback` or `clawed_back` then a subsequent new declaration can be sent
- also no longer creates an additional declaration if submitted (or onwards) event is found and returns an error instead

### Guidance to review

- none

### DIFF
--- a/app/services/record_declarations/base.rb
+++ b/app/services/record_declarations/base.rb
@@ -20,6 +20,7 @@ module RecordDeclarations
     validate :date_has_the_right_format
     validate :validate_schedule_present
     validate :validate_milestone_exists
+    validate :validates_billable_slot_available
 
     delegate :schedule, :participant_declarations, to: :user_profile, allow_nil: true
 
@@ -38,8 +39,6 @@ module RecordDeclarations
       validate_provider!
       validate_milestone!
       validate_withdrawals!
-
-      raise ActiveRecord::RecordNotUnique, "Declaration with given participant ID already exists" if record_exists_with_different_declaration_date?
 
       ParticipantDeclaration.transaction do
         set_eligibility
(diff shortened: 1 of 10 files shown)


app/services/record_declarations/base.rb
```diff
@@ -95,16 +94,20 @@ def declaration_parameters
...
+        .where(state: %w[submitted eligible payable paid])
+        .find_by(
+          user:,
+          course_identifier:,
+          declaration_type:,
+        )
 
-      declaration.present? && declaration.declaration_date != Time.zone.parse(declaration_date)
+      if declaration.present?
+        errors.add(:base, "There already exists a declaration that will be or has been paid for this event")
```
StupidCodeFactory, line 109: Please use `I18n` like the rest of the file.

spec/services/record_declarations/base_spec.rb
```diff
@@ -198,19 +198,37 @@ def matches_lead_provider?
...
       end
 
-      it "does not change the state of the original declaration" do
-        expect { subject.call }.not_to change { existing_declaration.reload.state }
+      it "does add another declaration with correct eligibility" do
+        expect { subject.call }.to change { ParticipantDeclaration.count }.by(1)
+
+        declaration = ParticipantDeclaration.order(created_at: :asc).last
+
+        expect(declaration.state).to eql("eligible")
```
StupidCodeFactory, line 231:
```suggestion
        expect(declaration).to be_eligible
```

StupidCodeFactory (approved):

PR #2275, opened 2022-06-27
migration to delete statement from declaration

asmega:
- this is now done though a join table of statement line items
- therefore this column can now be removed as no longer used

### Context

- Ticket:


app/models/participant_declaration.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ParticipantDeclaration < ApplicationRecord
-  self.ignored_columns = %w[statement_type]
+  self.ignored_columns = %w[statement_type statement_id]
```
StupidCodeFactory, line 4: why is this necessary? the column will be removed so there is nothing to ignore. 🤔

asmega: the strong migration thing complains at me. if it weren't there personally I wouldn't add it

PR #2285, opened 2022-06-29
[CPDLP-1481] Delegate NPQ type totals to each course

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1481
- This resolve a bug where on NPQ statement overview the total retained could be lower than actual value
- This is because the query de-duped
- So a user that has the same declaration type across multiple courses (likely NPQH + ASO) would only be counted once whey they should all be counted

- Delegate these total values to each course/contract then sum them all up

### Guidance to review

- none

### DIFF
--- a/app/services/finance/npq/course_statement_calculator.rb
+++ b/app/services/finance/npq/course_statement_calculator.rb
@@ -42,17 +42,19 @@ module Finance
       end
 
       def billable_declarations_count_for_declaration_type(declaration_type)
-        if declaration_type == "retained"
-          declaration_type = %w[retained-1 retained-2]
-        end
-
-        statement
+        scope = statement
           .billable_statement_line_items
           .joins(:participant_declaration)
           .where(participant_declarations: { course_identifier: course.identifier })
-          .where(participant_declarations: { declaration_type: })
           .merge(ParticipantDeclaration.select("DISTINCT (user_id, declaration_type)"))
-          .count
+
+        scope = if declaration_type == "retained"
+                  scope.where("participant_declarations.declaration_type LIKE ?", "retained-%")
+                else
+                  scope.where(participant_declarations: { declaration_type: })
+                end
+
+        scope.count
       end
 
       def clawback_payment


spec/services/finance/npq/course_statement_calculator_spec.rb
```diff
@@ -12,6 +12,35 @@
 
   subject { described_class.new(statement:, contract:) }
 
+  describe "#billable_declarations_count_for_declaration_type" do
+    before do
+      declarations = []
+
+      6.times do
```
asmega, line 19: can't use `create_list` here. otherwise we end up with all declaration of the same `declaration_type` from the `sample` call

spec/services/finance/npq/course_statement_calculator_spec.rb
```diff
@@ -12,6 +12,35 @@
...
+      declarations.each do |dec|
+        Finance::StatementLineItem.create!(
+          statement:,
+          participant_declaration: dec,
+          state: dec.state,
+        )
+      end
+    end
+
+    it "can count different declaration types", :aggregate_failures do
```
asmega, line 37: `aggregate_failures` to save on setup costs therefore time

spec/features/finance/npq/course_payment_breakdown_spec.rb
```diff
@@ -171,8 +169,8 @@ def then_i_should_see_correct_output_payment_breakdown
     within first(".app-application__card") do
       expect(page).to have_css("tr:nth-child(1) td:nth-child(1)", text: "Output payment")
       expect(page).to have_css("tr:nth-child(1) td:nth-child(2)", text: total_declarations(npq_leading_behaviour_culture_contract))
-      expect(page).to have_css("tr:nth-child(1) td:nth-child(3)", text: number_to_pounds(160))
-      expect(page).to have_css("tr:nth-child(1) td:nth-child(4)", text: number_to_pounds(1440))
+      expect(page).to have_css("tr:nth-child(1) td:nth-child(3)", text: number_to_pounds(162))
```
asmega, line 172: i've checked these by hand. these were different as we were lazy loading the contracts and were never evaluated therefore `first(".app-application__card")` would have been for a different contract/course

spec/features/finance/npq/course_payment_breakdown_spec.rb
```diff
@@ -305,19 +303,31 @@ def statement_declarations_per_contract(contract)
   end
 
   def total_starts
-    statement_declarations.started.count
+    statement
```
asmega, line 306: these are still bad as we are asserting against a moving target but i'm not fixing in this PR

StupidCodeFactory: agreed that's not ideal we should add that to let(:expected_(started|retaine|completed)_count

spec/services/finance/npq/course_statement_calculator_spec.rb
StupidCodeFactory: whenever there are multiple `expect` we should enforce adding `:aggregate_failures`. I think  there is a cop for that. I usually do but may skip it sometimes

app/services/finance/npq/course_statement_calculator.rb
```diff
@@ -41,6 +41,20 @@ def refundable_declarations_by_type_count
           .count
       end
 
+      def billable_declarations_count_for_declaration_type(declaration_type)
+        if declaration_type == "retained"
+          declaration_type = %w[retained-1 retained-2]
```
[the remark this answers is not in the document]
asmega: NPQ only has at most 2 retained milestones afaik

app/services/finance/npq/course_statement_calculator.rb
```diff
@@ -41,6 +41,20 @@ def refundable_declarations_by_type_count
...
+        end
+
+        statement
+          .billable_statement_line_items
+          .joins(:participant_declaration)
+          .where(participant_declarations: { course_identifier: course.identifier })
+          .where(participant_declarations: { declaration_type: })
+          .merge(ParticipantDeclaration.select("DISTINCT (user_id, declaration_type)"))
```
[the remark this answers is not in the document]
asmega: Yes, i think would work. However I feel pushing this logic downwards into `CourseStatementCalculator` is better than having queries in `StatementCalculator`. The downside is there will more queries and therefore a bit slower

app/services/finance/npq/course_statement_calculator.rb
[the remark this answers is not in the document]
StupidCodeFactory: That's hardly scalable, also this seems misplaced the caller should be responsible to know this sort of things, there will be no need for an `if` statement, the signature should be :

```ruby
def billable_declarations_count_for_declaration_type(declaration_types)
```

app/services/finance/npq/statement_calculator.rb
```diff
@@ -47,30 +47,21 @@ def total_payment
...
 
       def total_starts
-        statement
-          .billable_statement_line_items
-          .joins(:participant_declaration)
-          .where(participant_declarations: { declaration_type: "started" })
-          .merge(ParticipantDeclaration.select("DISTINCT (user_id, declaration_type)"))
-          .count
+        contracts.sum do |contract|
+          CourseStatementCalculator.new(statement:, contract:).billable_declarations_count_for_declaration_type("started")
```
StupidCodeFactory, line 51: also seem we'd do just fine with a single instance of the `CourseStatementCalculator` 
just pass the contract as an argument to the methods. Would remove the need to instantiate a bunch of objects per contract / per declaration type

StupidCodeFactory: after checking the class it appears we don't even use the contract. Can you ✂️  that please.

asmega: not sure I follow:
- `CourseStatementCalculator` does use the `contract` that is passed so I can't remove that. Just because the method doesn't use it does not mean i can remove it from initialisation. 
- There are multiple courses so I expect `CourseStatementCalculator` to be instantiated multiple times, not once

StupidCodeFactory (approved):

PR #2288, opened 2022-06-29
[CPDLP-1429] Track changes using touch and updated_at

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CPDLP-1429)

Track changes to induction records, induction programmes and participant identities.

See also #2287

### DIFF
--- a/app/models/induction_programme.rb
+++ b/app/models/induction_programme.rb
@@ -25,11 +25,19 @@ class InductionProgramme < ApplicationRecord
 
   delegate :school, to: :school_cohort
 
+  after_commit :touch_induction_records
+
   def lead_provider
     partnership&.lead_provider
   end
 
   def delivery_partner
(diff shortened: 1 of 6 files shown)


app/models/induction_programme.rb
```diff
@@ -25,11 +25,19 @@ class InductionProgramme < ApplicationRecord
...
   end
 
   def delivery_partner
     partnership&.delivery_partner
   end
+
+private
+
+  def touch_induction_records
+    current_induction_records.each(&:touch)
```
jamesrose, line 41: I think you can even do this in Rails 6 with makes it not an N+1:

```suggestion
    current_induction_records.touch_all
```

jamesrose (approved):

jamesrose (approved):

PR #2294, opened 2022-07-01
[CPDLP-1475] users with 2+ NPQ profiles can change schedule

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1475
- this is a workaround for bad data where users have multiple NPQ profiles
- providers are failing to change schedule as it is not picking the correct profile

- users with multiple NPQ profiles with different providers can now change schedule
- we observe the calling provider and scope profiles to that provider

### Guidance to review

- none

### DIFF
--- a/app/services/participants/change_schedule/npq.rb
+++ b/app/services/participants/change_schedule/npq.rb
@@ -42,7 +42,17 @@ module Participants
       attr_reader :schedule_identifier, :cohort_year
 
       def user_profile
-        user&.npq_profiles&.active_record&.includes({ npq_application: [:npq_course] })&.where('npq_courses.identifier': course_identifier)&.first
+        user
+          .npq_profiles
+          .active_record
+          .includes({ npq_application: [:npq_course] })
+          .where('npq_courses.identifier': course_identifier)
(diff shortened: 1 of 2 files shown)


app/services/participants/change_schedule/npq.rb
```diff
@@ -42,7 +42,17 @@ def call
       attr_reader :schedule_identifier, :cohort_year
 
       def user_profile
-        user&.npq_profiles&.active_record&.includes({ npq_application: [:npq_course] })&.where('npq_courses.identifier': course_identifier)&.first
+        user
+          &.npq_profiles
+          &.active_record
+          &.includes({ npq_application: [:npq_course] })
+          &.where('npq_courses.identifier': course_identifier)
+          &.where({ npq_application: { npq_lead_provider: } })
```
StupidCodeFactory, line 50: None of those `&.` is necessary. this will always return and `ActiveRecord::Relationships` which will never be nil

asmega: good shout, I've removed all these now

lazydevorg (approved):

PR #2296, opened 2022-07-01
[CPDLP-1468] When profile becomes eligible, attach declarations

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1468
- There is a case where a participant is not yet eligible but a declaration has been submitted
- Later the participant is found to be eligible as a result their declaration(s) are changed to payable
- However when this happens we are not attaching them to a relevant statement therefore will not be paid for

- When participant becomes eligible also attach declarations to relevant statement

### Guidance to review

- after this is deployed will need to find eligible declarations left behind and attach them to the next statement


mooktakim (approved):
👍

PR #2308, opened 2022-07-05
[CN-261] Add admin interface for exporting NPQApplications

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CN-261

To allow self service of data export requests this PR adds in an interface for requesting an export. 

As an added layer of security, exports are not provided directly to the person requesting the export. Instead they are exported to google drive with its own layer of access control.

- Adds in admin interface for scheduling export job
- Adds export job that creates CSV of NPQ Applications and uploads to google drive
- Adds NPQApplicationExport model to track export requests and provide an audit log

### Guidance to review

When you pull this down locally scheduling will work correctly but actually running the background job will fail as you need to set up the environment variables to allow the Google Service Account to run, see the .env.development changes to see what you need to put in place, these can be obtained from the google cloud interface.


app/services/admin/application_csv_generator.rb
```diff
@@ -0,0 +1,63 @@
...
+      :funding_choice,
+      :employer_name,
+      :employment_role,
+      :targeted_delivery_funding_eligibility,
+      :created_at,
+      :user_id,
+      :user_email,
+      :user_full_name,
+      :teacher_reference_number_verified,
+      # :teacher_reference_number # Possibly not needed
```
Lockyy, line 35: This is pending a response from Chesca on the Jira ticket.

app/services/admin/application_csv_generator.rb
```diff
@@ -0,0 +1,63 @@
...
+  class ApplicationCsvGenerator
+    CSV_EXPORT_COLUMNS = [
+      :id,
+      :cohort_start_year,
+      :npq_course_id,
+      :npq_lead_provider_id,
+      :npq_course_name,
+      :npq_lead_provider_name,
+      # :teacher_catchment, # Not currently sent to ECF from NPQ
+      # :teacher_catchment_country, # Not currently sent to ECF from NPQ
```
Lockyy, line 15: These will need uncommenting and including in the export once we get these sending from NPQ into ECF.

PR #2309, opened 2022-07-06
[CPDLP-1491] Add class to encapsulate declaration extracts

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1491
- There is a block of SQL that gets passed around to run "extracts"

- Make a home for this SQL but sticking it in a class for future reference and use

### Guidance to review

- none

### DIFF
--- a/app/services/finance/extract_report.rb
+++ b/app/services/finance/extract_report.rb
@@ -26,7 +26,7 @@ module Finance
               clp.name                                               AS lead_provider_name,
                 na.school_urn                                        AS school_urn,
                 na.eligible_for_funding                              AS eligible_for_funding,
-                (SELECT name FROM schools WHERE urn = na.school_urn) AS school_name,
+                schools.name                                         AS school_name,
                 pd.user_id                                           AS participant_id,
                 tp.trn                                               AS participant_trn,
                 pp.id                                                AS application_id,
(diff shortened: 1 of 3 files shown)


app/services/finance/extract_report.rb
```diff
@@ -0,0 +1,65 @@
...
+      @raw_connection ||= connection.raw_connection
+    end
+
+    def npq_sql
+      <<-SQL
+        COPY (SELECT
+              clp.name                                               AS lead_provider_name,
+                na.school_urn                                        AS school_urn,
+                na.eligible_for_funding                              AS eligible_for_funding,
+                (SELECT name FROM schools WHERE urn = na.school_urn) AS school_name,
```
peteryates, line 29: Any reason to do a subquery here rather than `join schools`?

I _think_ this might result in a separate query per row (unconfirmed, not sure how PostgreSQL will deal with it internally).

asmega: - It's copy paste job so didn't look too much into it, as it seems to work
- if you do a join and it doesn't return anything the row will be empty. 
- here the subquery will either return a string or null
- but more importantly if it is null it will still return the row unlike the join

asmega: you could probably do so some kind of outer join to yield the same result

peteryates: Yeah, `left outer` will do the trick.

peteryates (approved):
Looks good to me

PR #2312, opened 2022-07-06
Extract appropriate body selection from school cohort setup

lazydevorg:
### Context

- Ticket: CST-800


app/controllers/concerns/appropriate_body_selection/controller.rb
```diff
@@ -0,0 +1,109 @@
...
+      render "/appropriate_body_selection/body_appointed"
+    end
+
+    def update_appropriate_body_appointed
+      if @appropriate_body_form.valid? :body_appointed
+        store_appropriate_body_form
+        if @appropriate_body_form.body_appointed?
+          redirect_to action: :appropriate_body_type
+        else
+          method(appropriate_body_submit_action).call
```
xswoodsx, line 23: Just wondering what's the difference between doing the method().call and just calling the method `appropriate_body_submit_action`

app/forms/appropriate_body_selection_form.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+class AppropriateBodySelectionForm
+  include ActiveModel::Model
+  include ActiveRecord::AttributeAssignment
+  include ActiveModel::Serialization
+
+  attr_accessor :body_appointed, :body_type, :body_id
+
```
xswoodsx, line 9: Think it needs some error messages when no options etc are selected.

app/controllers/concerns/appropriate_body_selection/controller.rb
xswoodsx: So is that to stop data from a different journey using data in another session? and then causing problems with what's submitted?

PR #2320, opened 2022-07-08
remove participant_declaration#voided_at column

asmega:
- this column was never used

### Context

- Ticket:


db/schema.rb
```diff
@@ -504,7 +504,6 @@
     t.boolean "works_in_school"
     t.string "employer_name"
     t.string "employment_role"
-    t.boolean "targeted_support_funding_eligibility", default: false
```
mooktakim, line 507: should this be included?

mooktakim (approved):

db/schema.rb
asmega: no, i'll remove it

PR #2326, opened 2022-07-12
Confirm or choose appropriate body when adding ECTs

lazydevorg:
### Context

- Ticket: CST-800

- When adding ECTs the user get asked to confirm whether the school's appropriate body is the same as the ECT.
  - If confirmed to be the same then the appropriate body is assigned to the ECT too.
  - If not confirmed then the user can select an appropriate body for the ECT.
- The appropriate body selection is allowed only for ECTs (not mentors) and only when the school has previously selected an appropriate body.
- The user can edit ECTs appropriate body from detail page if the school has previously selected an appropriate body.

### DIFF
--- a/app/controllers/schools/setup_school_cohort_controller.rb
+++ b/app/controllers/schools/setup_school_cohort_controller.rb
@@ -82,6 +82,8 @@ module Schools
                                        school_name: @school.name
     end
 
+    def appropriate_body_type; end
+
     def what_changes_submitted; end
 
     def change_fip_programme_choice; end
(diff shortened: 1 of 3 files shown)


app/services/early_career_teachers/create.rb
```diff
@@ -33,10 +33,11 @@ def call
...
         if school_cohort.default_induction_programme.present?
-          Induction::Enrol.call(participant_profile: profile,
-                                induction_programme: school_cohort.default_induction_programme,
-                                mentor_profile:,
-                                start_date:)
+          induction_record = Induction::Enrol.call(participant_profile: profile,
+                                                   induction_programme: school_cohort.default_induction_programme,
+                                                   mentor_profile:,
+                                                   start_date:)
+          induction_record.update!({ appropriate_body_id: })
```
tonyheadford, line 40: Could we change `Induction::Enrol` to take this as an optional param so that we can keep the appearance that induction records are "immutable" ?

tonyheadford (approved):
LGTM!

ltello (approved):

ltello (approved):

PR #2328, opened 2022-07-13
NPQ application seeder attaches a cohort

asmega:
### Context

- Ticket: n/a
- NPQ Applications now require a cohort

- Update seeding mechanism to attach next cohort to new NPQ applications

### Guidance to review

- none


peteryates (commented):
This looks good to me. As per the discussion on Slack I think the non-production guard in `npq_applications.rb` gives the necessary safety.

It might be worth ensuring it can't be run in prod via a spec?

PR #2331, opened 2022-07-14
Fix tests around NPQ and withdrawal

asmega:
### Context

- Ticket: n/a

- Fixes tests around NPQ and withdrawal
- This is due to a started declaration must exist now in order to withdraw

### Guidance to review

- none

### DIFF
--- a/spec/requests/api/v2/ecf_participants_spec.rb
+++ b/spec/requests/api/v2/ecf_participants_spec.rb
@@ -407,14 +407,62 @@ RSpec.describe "Participants API", type: :request do
     end
   end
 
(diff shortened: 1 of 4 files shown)


spec/requests/api/v2/ecf_participants_spec.rb
```diff
@@ -407,14 +407,62 @@
     end
   end
 
-  it_behaves_like "JSON Participant Deferral endpoint", "participant" do
```
asmega, line 410: i would really like for these rspec shared examples groups to be removed in the long term

peteryates (approved):
Nice, all looks clear and concise to me 👌

PR #2332, opened 2022-07-15
[CPDLP-1531] Return error when operating on withdrawn profile

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1531
- When a provider attempts to change schedule and the profile is withdrawn
- This currently explodes as there is a scope that returns only active profiles so therefore returns `nil`
- We then attempt to send messages to this `nil` object 

- Add validation so that if no active profile is found and there are other profiles, it adds an error that the participant is withdrawn
- This short circuits reaching the code that sends messages to `nil` 
- This fixes change schedule when attempting to operate on a withdrawn profile

### Guidance to review

- none

### DIFF
--- a/spec/services/participants/early_career_teacher_spec.rb
+++ b/spec/services/participants/early_career_teacher_spec.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+RSpec.describe Participants::EarlyCareerTeacher do
+  let(:klass) do
+    Class.new do
+      include ActiveModel::Model
+      include ActiveModel::Validations
+      include Participants::EarlyCareerTeacher
+
+      attr_accessor :participant_identity
+    end
+  end
+
+  subject { klass.new(participant_identity:) }
+
+  describe "validation" do
+    context "when single profile is withdrawn" do
+      let(:participant_identity) { participant_profile.participant_identity }
+      let(:participant_profile) { create(:ect_participant_profile, :withdrawn) }
+
+      it "is not valid" do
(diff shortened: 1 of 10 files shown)


spec/services/participants/early_career_teacher_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+RSpec.describe Participants::EarlyCareerTeacher do
+  let(:klass) do
+    Class.new do
```
peteryates, line 5: Never seen it done like this, nice!

I usually just end up creating a class outside the `describe` block.

peteryates (approved):
Looks good.

There's a lot of overlap between `early_career_teacher_spec.rb` and `mentor_spec.rb`, but while there's just two places this is probably more sensible than splitting it out into shared examples.

asmega:
> Looks good.
> 
> There's a lot of overlap between `early_career_teacher_spec.rb` and `mentor_spec.rb`, but while there's just two places this is probably more sensible than splitting it out into shared examples.

I'm hoping we can delete most of this code at some point as this inheritance is rather painful. @StupidCodeFactory is refactoring these at the moment so we should get some improvements soon

asmega:
Blocking merge as after PO review did not seem to behave as desired by not fixing the intended problem

asmega:
Closing as did not seem to be an issue. It's dropped down the line for things to fix

PR #2336, opened 2022-07-18
[CST-849]: new journey for a tutor to update an ect or mentor's name

ltello:
### Context
A new journey for a tutor to change a participant's name has been defined so that:
- A new view to ask for the reason to change the name is presented to the tutor before typing the new name.
- If the reason is that the participant should have not been registered in the service, then the change is not allowed and a view explaining the reason is displayed. From it, the tutor can go to the feature to remove the participant from the cohort.
- If the reason is that the participant should be replaced with a different person, then the change is not allowed and a view explaining the reason is displayed. The tutor needs to add a new ect and remove the existing one instead of changing the name.
- Any other reason will allow the change to happen and a form is displayed asking for the new details. After the change is effective, the tutor will see a view with links to go either to the participants profile page or the participants dashboard.

- Ticket: [As a SIT, I can edit a participant's name](https://dfedigital.atlassian.net/browse/CST-849)


app/controllers/schools/base_controller.rb
```diff
@@ -20,30 +20,26 @@ def ensure_school_user
   end
 
   def active_school
-    return if params[:school_id].blank?
-
-    School.friendly.find(params[:school_id])
+    School.friendly.find(params[:school_id]) if params[:school_id].present?
```
lazydevorg, line 23: I think this way if the school_id param is not present it won't fail although I think the initial intention was for it to fail. Furthermore this is return nil that might raise bugs difficult to spot.
The urls in the school namespace all have school_id param

app/controllers/schools/base_controller.rb
```diff
@@ -20,30 +20,26 @@ def ensure_school_user
...
   end
 
   def active_cohort
-    return if params[:cohort_id].blank?
+    Cohort.find_by(start_year: params[:cohort_id]) if params[:cohort_id].present?
```
lazydevorg, line 27: same as above (https://github.com/DFE-Digital/early-careers-framework/pull/2336/files#diff-98d04464325117a46a438fb6b315b490b1c83703e41d81fad3fecf61f96c3d4dR23)

lazydevorg (changes requested):

app/controllers/schools/base_controller.rb
ltello: Sorry but I don't get what you mean. 
I think both approaches would return exactly the same for any value coming in params. I don't mind to revert it if you prefer the old syntax though

lazydevorg: Sorry what I mean is that before adding `if params[:school_id].present?` this method would raise a record not found error when the school_id param was not present or the School was not found.
The way it is in the PR it will make the method return nil instead.
That means the nil value returned have to be handled in all places using the method.
I think if an error is raised instead, it will be better as the school_id should always be present in the school namespace

lazydevorg: `School.friendly.find(nil)` is raises a ActiveRecord::RecordNotFound.
When the param :school_id is not present `params[:school_id]` returns nil and so the `.find` raises the error.
The `active_school` method will not return and raise an error.
Adding the `if params[:school_id].present?` will make the method return nil when `params[:school_id]` is not present.
This changes the current behaviour making the caller method have to deal with the nil.
Below the `active_school` method there is:
```ruby
def set_school_cohort
    @school = active_school
    @cohort = active_cohort
    @school_cohort = policy_scope(SchoolCohort).find_by(cohort: @cohort, school: @school)
```
...
The `@school_cohort` line will fail having `@cohort` and/or `@school` possibly nil.
Furthermore the BaseController class uses `active_school` and `active_cohort` to set up `@school` and `@cohort` for all controller based upon it.
All controllers suppose that `@school` and `@cohort` are set and can't be nil.
If we introduce this change `@school` and `@cohort` might be nil and this should not be possible.
A nil `@school` or a nil `@cohort` should never happen at this point that's why I think it's better that it fails then return a nil.

PR #2342, opened 2022-07-18
remove BPN seeds

asmega:
- this does not appear to be used
- also means that MakeDeclarationsPayable and MakeDeclarationsPaid can
  now also be removed which should no longer be used

### Context

- Ticket:


asmega:
gone stale

PR #2346, opened 2022-07-20
[CST-188]: add confirmation page when tutor changes a participant's email address

ltello:
### Context

- Ticket: [As a SIT, I can edit a participant's email address](https://dfedigital.atlassian.net/browse/CST-188)

  Added a confirmation view once a tutor has been able to successfully update the email address of one of their participants.

It also includes a few tweaks:
- Keep marked the option selected by a SIT when they click back in one of the views to change a participant's name 
- Do not pre-fill the input box to edit a participant's name or email address

### DIFF
--- a/app/views/schools/participants/edit_email.html.erb
+++ b/app/views/schools/participants/edit_email.html.erb
@@ -7,7 +7,7 @@
     <%= form_for @induction_record.preferred_identity, url: schools_participant_update_email_path, method: :put do |f| %>
     <%= f.govuk_error_summary %>
 
-      <%= f.govuk_text_field :email, label: { text: "Change #{@profile.ect? ? "ECT’s" : "mentor’s"} email address", tag: 'h1', size: 'xl' } %>
+      <%= f.govuk_text_field :email, label: { text: "What’s #{@profile.full_name}’s correct email address?", tag: 'h1', size: 'xl' } %>
       <%= f.govuk_submit %>
     <% end %>
   </div>
--- a/app/views/schools/participants/edit_email.html.erb
+++ b/app/views/schools/participants/edit_email.html.erb
@@ -5,9 +5,12 @@
     <span class="govuk-caption-xl"><%= @school.name %></span>
 
     <%= form_for @induction_record.preferred_identity, url: schools_participant_update_email_path, method: :put do |f| %>
-    <%= f.govuk_error_summary %>
-
-      <%= f.govuk_text_field :email, label: { text: "What’s #{@profile.full_name}’s correct email address?", tag: 'h1', size: 'xl' } %>
+      <%= f.govuk_error_summary %>
+      <%= f.govuk_text_field :email,
+                             class: 'govuk-!-width-two-thirds',
+                             hint: -> { tag.p(t('participant.edit_email_hint')) },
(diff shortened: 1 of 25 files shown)


lazydevorg (approved):

tonyheadford (approved):
LGTM!

app/views/schools/participants/edit_email.html.erb
```diff
@@ -4,10 +4,17 @@
...
+                  scope: '',
+                  url: schools_participant_update_email_path,
+                  method: :put do |f| %>
+      <%= f.govuk_error_summary %>
+      <%= f.govuk_text_field :email,
+                             value: nil,
+                             class: 'govuk-!-width-two-thirds',
+                             hint: -> { tag.p(t('participant.edit_email_hint')) },
+                             label: -> { tag.h1(t('participant.edit_email', name: @profile.full_name),
+                                                class: 'govuk-heading-xl') } %>
```
peteryates, line 17: I don't think you need the proc here:

```suggestion
                             label: { text: t('participant.edit_email', name: @profile.full_name), size: 'xl', tag: 'h1' }
```
```

File: app/views/schools/participants/edit_name.html.erb [hunk 2]
```diff
@@ -1,19 +1,19 @@
...
+    <%= form_with model: @profile.user, scope: '', url: schools_participant_update_name_path, method: :put do |f| %>
       <%= f.govuk_error_summary %>
       <%= f.govuk_text_field :full_name,
-                             label: {
-                               text: "What should we edit #{@profile.full_name}’s name to?",
-                               tag: 'h1',
-                               size: 'xl'
-                             } %>
+                             value: nil,
+                             class: 'govuk-!-width-two-thirds',
```
peteryates, line 14:
```suggestion
                             width: 'two-thirds',
```

app/views/schools/participants/edit_email.html.erb
ltello: Initially I wrote it like your suggestion. 
However, that left a very short space between the label and the input which should be much bigger as per design.
Using a proc is the only way I could figure out to achieve it. 

Now that you say that, I have checked and it is the css class what I missed, so I have tried to add it to your approach and it works!. Thanks I will change it!

app/views/schools/participants/edit_name.html.erb
ltello: The prototype uses `govuk-!-width-two-thirds` class and if I change it to `two-thirds` the input box gets wider

peteryates: It's odd that it's wider with the arg than with the class being set by name, in the form builder it [_should_ be doing the same thing](https://github.com/DFE-Digital/govuk-formbuilder/blob/842cd45df3619d9b439a729d114a1caa9d6a3928/lib/govuk_design_system_formbuilder/traits/input.rb#L79).

ltello: You are absolutely right. Sorry, my bad.
I just tried `class: 'two-thirds'` instead of `width: 'two-thirds'`

ltello: Applied the same change to edit_email template

PR #2347, opened 2022-07-20
[CN-308] Retrospectively remove ECHO from targeted support funding eligibility

Lockyy:
### Context

https://dfedigital.atlassian.net/browse/CN-308

Adds rake task that updates `targeted_delivery_funding_eligibility` to false for all Application records with the EHCO course, `targeted_delivery_funding_eligibility` set to true, and where created after the relaunch.

### Guidance to review

1. Create an Application record with these attributes:
```
  course == Course.ehco.first
  created_at >= 6 June 2022 12:00
  targeted_delivery_funding_eligibility == true
```
2. Run ehco:update_targeted_delivery_funding_eligibility:dry_run 
3. See that it has shown that this record will be updated
4. Check that targeted_delivery_funding_eligibility change has not been persisted.
5. Run ehco:update_targeted_delivery_funding_eligibility:run
6. Check that targeted_delivery_funding_eligibility change has been persisted.

### DIFF
--- a/app/services/npq/ehco/targeted_delivery_funding_eligibility_updater.rb
+++ b/app/services/npq/ehco/targeted_delivery_funding_eligibility_updater.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module NPQ
+  module Ehco
+    class TargetedDeliveryFundingEligibilityUpdater
+      REOPEN_DATE = Time.zone.parse("6 June 2022 12:00")
+
+      class << self
+        delegate :run, to: :new
+      end
+
+      def run
+        logger = Logger.new($stdout)
+        logger.info "Updating EHCO NPQ Applications, this may take a couple of minutes..."
+
+        ehco_npq_course = NPQCourse.find_by(id: "66dff4af-a518-498f-9042-36a41f9e8aa7")
+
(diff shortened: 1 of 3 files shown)


app/services/npq/ehco/targeted_delivery_funding_eligibility_updater.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      class << self
+        delegate :run, to: :new
+      end
+
+      def run
+        logger = Logger.new($stdout)
+        logger.info "Updating EHCO NPQ Applications, this may take a couple of minutes..."
+
+        ehco_npq_course = NPQCourse.find_by(id: "66dff4af-a518-498f-9042-36a41f9e8aa7")
```
leoapost, line 16: This will return the [NPQEYL](https://github.com/DFE-Digital/early-careers-framework/blob/develop/db/seeds/initial_seed.rb#L55) course, not the [EHCO](https://github.com/DFE-Digital/early-careers-framework/blob/develop/db/seeds/initial_seed.rb#L54).

Also, it may worth finding the course by its `identifier` instead of its `id`?

leoapost (changes requested):

leoapost (approved):

PR #2353, opened 2022-07-25
Participant history viewer

mattonfoot:
### Context

It should be possible to see the historical events or changes to a CPD record over time in order to help determine whay a record is in the state is is and where we might go to correct or undo any actions that misreported information.

### Guidance to review

login as an admin user and proceed to the participants dashboard, from here you can bring up any participants screen and view the events tab.

bring up the rails console in your chosen environment and run a query to find a `User` or `ParticipantProfile` object. Call `ParticipantEventHistory.from_user` or `ParticipantEventHistory.from_profile` and access the `events` attribute to see a full list of events created for a record.

image

### DIFF
--- a/app/controllers/admin/participants_controller.rb
+++ b/app/controllers/admin/participants_controller.rb
@@ -8,6 +8,7 @@ module Admin
     before_action :load_participant, except: :index
     before_action :historical_induction_records, only: :show, unless: -> { @participant_profile.npq? }
     before_action :latest_induction_record, only: :show, unless: -> { @participant_profile.npq? }
+    before_action :event_list, only: :show
 
     def show; end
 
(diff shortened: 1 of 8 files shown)


app/controllers/admin/participants_controller.rb
```diff
@@ -8,6 +8,7 @@ class ParticipantsController < Admin::BaseController
     before_action :load_participant, except: :index
     before_action :historical_induction_records, only: :show, unless: -> { @participant_profile.npq? }
     before_action :latest_induction_record, only: :show, unless: -> { @participant_profile.npq? }
+    before_action :event_list, only: :show
 
     def show; end
```
peteryates, line 13: If this is just being set for `:show` it probably makes more sense to assign inside the `show` method.

mattonfoot: Seem to have linter issues when I try to do this

PR #2360, opened 2022-07-25
[CST-857] bug: participant show breaks when mentor has been removed by a sit

ltello:
### Context

- Ticket: [BUG - Participant show breaks when mentor has been removed ](https://dfedigital.atlassian.net/browse/CST-857)

This happens because there is some inconsistency in the models for this situations:

```
induction_record.mentor_profile.status #=> ‘withdrawn’
induction_record.mentor_profile.user.mentor_profile #=> nil  !!!
```

The  reason for that is: 

```
class TeacherProfile
  has_one :mentor_profile, -> { active_record }, class_name: "ParticipantProfile::Mentor"
end
```

So from an induction_record we can get to the mentor’s user associated with this record.

However if we go backwards, from that mentor user we can’t get to that user’s mentor profile or their associated induction_record.

 - Make the view not to require to go backwards and receive @mentor_profile instead of @mentor to retrieve the  mentor_profile.id
 - Move to a service object all the logic for a SIT to remove a participant from a school cohort.
 - Reuse the extracted logic in the admin dashboard to remove participants exactly the same way as a SIT does.

### DIFF
--- a/db/migrate/20220725205758_dissociate_withdrawn_mentors_from_induction_records.rb
+++ b/db/migrate/20220725205758_dissociate_withdrawn_mentors_from_induction_records.rb
@@ -1,9 +1,0 @@
-# frozen_string_literal: true
-
-class DissociateWithdrawnMentorsFromInductionRecords < ActiveRecord::Migration[6.1]
-  def change
-    InductionRecord.joins(:mentor_profile)
-                   .where(participant_profiles: { status: :withdrawn })
-                   .update_all(mentor_profile_id: nil)
-  end
-end
--- a/db/migrate/20220725205758_dissociate_withdrawn_mentors_from_induction_records.rb
+++ b/db/migrate/20220725205758_dissociate_withdrawn_mentors_from_induction_records.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class DissociateWithdrawnMentorsFromInductionRecords < ActiveRecord::Migration[6.1]
   def change
     InductionRecord.joins(:mentor_profile)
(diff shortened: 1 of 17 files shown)


db/migrate/20220725205758_dissociate_withdrawn_mentors_from_induction_records.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class DissociateWithdrawnMentorsFromInductionRecords < ActiveRecord::Migration[6.1]
+  def change
+    InductionRecord.joins(:mentor_profile)
+                   .where(participant_profiles: { status: :withdrawn })
+                   .update_all(mentor_profile_id: nil)
+  end
+end
```
tonyheadford, line 9: I don't think we need this?

tonyheadford (changes requested):
Looks good except I don't think we want that migration?

db/migrate/20220725205758_dissociate_withdrawn_mentors_from_induction_records.rb
ltello: Ok, this was when I thought we were going to nullify all the induction records mentored by the mentor deleted. 
I will remove it now

tonyheadford (approved):

db/migrate/20220725205758_dissociate_withdrawn_mentors_from_induction_records.rb
ltello: Re-added the migration for current induction records still linked to withdrawn mentor profiles.

PR #2363, opened 2022-07-27
[CST-799] Schools who have already set up their programme can report their AB

lazydevorg:
### Context

- Ticket: CST-799

Schools that have not appointed an appropriate body can now add one from the dashboard.
The appropriate body detail line shows an empty value, as previously, and the Add action link.
The link was taking the user to a static page showing a work in progress message.
The link now takes the user to the appropriate body selection journey.
After the journey is completed the appropriate body is applied to the cohort and all the related ECTs.


app/views/schools/cohorts/appropriate_body/confirm.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    <ul class="govuk-list govuk-list--bullet">
+      <li>you’ve already added to this service</li>
+      <li>starting their training in the 2022 to 2023 academic year</li>
+    </ul>
+    <p>We do this because most schools work with only one appropriate body.</p>
+
+    <h2 class="govuk-heading-m">How to report a different appropriate body for an ECT</h2>
+    <p class="govuk-body">
+      If you’ve already added the ECT to this service, select their name on the
+      <%= govuk_link_to "ECTs and mentors", schools_participants_path, class: "govuk-link govuk-link--no-visited-state" %>
```
peteryates, line 20: `#govuk_link_to` supports doing this with args:

```suggestion
      <%= govuk_link_to "ECTs and mentors", schools_participants_path, no_visited_state: true %>
```

app/services/induction/set_school_cohort_appropriate_body.rb
```diff
@@ -2,27 +2,35 @@
...
     @school_cohort = school_cohort
     @appropriate_body_id = appropriate_body_id
     @appropriate_body_appointed = appropriate_body_appointed
+    @update_induction_records = update_induction_records
   end
 
-  def set_appropriate_body
-    if !appropriate_body_appointed
+  def set_school_cohort_appropriate_body
+    if appropriate_body_appointed == false
```
peteryates, line 23: Is there a reason for the `== false` here, rather than just doing `if appropriate_body_appointed` and inverting the 'then/else'? I guess it's explicitly guarding against nils?

peteryates (commented):
Looks fine to me overall, feels like it makes sense. Left a couple of minor comments/questions.

app/services/induction/set_school_cohort_appropriate_body.rb
lazydevorg: the appropriate_body_appointed field is used when the journey ask the user whether the appropriate body has been appointed or not.
There are some journeys where the user just chooses an appropriate body.
When the journey requires to skip the appropriate body appointed step then the value is nil (never used).
We need to differentiate when the value is explicitly set to false that's when the user explicitly state they don't know the appropriate body.
To answer your question, yes it is guarding against nil.

lazydevorg: @peteryates is that ok to leave it like this?
Otherwise I'm open to suggestions.

peteryates (approved):
Looks fine to me.

PR #2365, opened 2022-07-27
[CPDLP-1532] Update change schedule

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1532
- This rolls up several changes into one but gets us a step closer to providers being able to re-post declarations after they have been clawed back

- There is no longer a change schedule class for ECTs vs mentors. This has now been rolled into one class for `ECF`
- The mixin `ProfileAttributes` has been removed and now inlined for comprehend-ability
- When changing schedule this now creates a new `InductionRecord` rather than mutating the relevant one
- It does still technically mutate the relevant induction record but only to track the `changed`
- We now select the correct induction record based on the last `created_at`
- When attempting to change schedule we now validate existing declarations against the new schedule. But only those that are `paid` or at some point could become `paid` ie `submitted`, `eligible`, `payable`, `paid` so they do not break the criteria of the new milestones
- Otherwise provider must void or clawback the required declarations in order to change schedule
- Provider are now no longer able to change schedule that belongs to a different cohort
- Requirement for the above bullet point may changed if it becomes too burdensome
- Appropriate nil guards have been added to validations so they do not error out due to nil pointers

### Guidance to review

- none


app/services/participants/change_schedule/ecf.rb
```diff
@@ -3,17 +3,31 @@
...
   module ChangeSchedule
     class ECF
-      include ProfileAttributes
       include ActiveModel::Validations
 
+      attr_accessor :course_identifier, :participant_id, :cpd_lead_provider
+
       delegate :school_cohort, to: :user_profile, allow_nil: true
 
+      validates :course_identifier, presence: { message: I18n.t(:missing_course_identifier) }
```
peteryates, line 12: Would it make sense to have an `inclusion` validation rule here?

peteryates (approved):
Looks good, left one minor comment on the validation of `course_identifier` (vs doing it in `#valid_courses`), but there's a chance I'm just missing the why.

app/services/participants/change_schedule/ecf.rb
asmega: > Looks good, left one minor comment on the validation of course_identifier (vs doing it in #valid_courses), but there's a chance I'm just missing the why.

what you said is perfectly reasonable and this class still needs more refining. i feel now at least things start to become more sane so things like this are more visible and changes like your suggestions can be made as the code becomes easier to reason with

PR #2366, opened 2022-07-28
[CPDLP-1552] Show the correct participant role in the delivery partners check data service

leoapost:
### Context

- Ticket: [CPDLP-1552](https://dfedigital.atlassian.net/browse/CPDLP-1552)

The check data service shows wrong participant roles in the UI and csv extract, when users are both Mentors and Induction Tutors.

- I created the `role` method in the `ParticipantProfile::ECF` model to determine the participant's role based on their participant type
- Replaced the `User#user_description` method that shows the wrong role in both the UI and CSV serializer, with the `ParticipantProfile::ECF#role`

### DIFF
--- a/app/models/participant_profile/ecf.rb
+++ b/app/models/participant_profile/ecf.rb
@@ -79,15 +79,6 @@ class ParticipantProfile < ApplicationRecord
         .first
     end
 
-    def role
-      case participant_type
-      when :ect
-        "Early career teacher"
-      when :mentor
-        "Mentor"
-      end
-    end
-
   private
 
     def update_analytics
--- a/app/models/participant_profile/ecf.rb
+++ b/app/models/participant_profile/ecf.rb
@@ -79,6 +79,15 @@ class ParticipantProfile < ApplicationRecord
         .first
     end
 
+    def role
+      case participant_type
+      when :ect
+        "Early career teacher"
+      when :mentor
+        "Mentor"
+      end
+    end
+
   private
 
     def update_analytics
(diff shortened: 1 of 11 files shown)


app/models/participant_profile/ecf.rb
```diff
@@ -79,6 +79,15 @@ def relevant_induction_record(lead_provider:)
         .first
     end
 
+    def role
```
asmega, line 82: fyi if you implemented `role` on the subclasses there would be no need for a case statement

spec/models/participant_profile/ecf_spec.rb
```diff
@@ -210,4 +210,22 @@
       end
     end
   end
+
+  describe "#role" do
+    context "when participant is Mentor" do
+      let(:profile) { create(:mentor_participant_profile) }
```
asmega, line 216: i probably would have chosen to use `build` here to do it in memory rather than make the db call

asmega (approved):

app/models/participant_profile/ecf.rb
leoapost: I moved it in the ParticipantProfile class and overriding it in the subclasses.

spec/models/participant_profile/ecf_spec.rb
leoapost: I went with the `create` because the `build(:mentor_participant_profile)` is failing

asmega (approved):

mooktakim (approved):

PR #2370, opened 2022-07-29
[Bug fix] Prevent duplicate start_dates in induction records

tonyheadford:
### Context

The change process for an `InductionRecord` involves duplicating the existing record and applying the changes to the duplicate and updating the existing one with a `changing` status and an `end_date`.  When a record has a future `start_date` we need to preserve it and so this creates records with duplicate `start_date`s. This becomes an issue when trying to sort on `start_date` or when trying to traverse InductionRecords by start and end dates. As there is no real need to preserve the history of future events that haven't happened, we can prevent this and just update a future record instead of creating a new successor record.

Update `Induction::ChangeInductionRecord` to handle future induction records.

### DIFF
--- a/app/services/induction/change_induction_record.rb
+++ b/app/services/induction/change_induction_record.rb
@@ -5,23 +5,24 @@ class Induction::ChangeInductionRecord < BaseService
     ActiveRecord::Base.transaction do
       time_now = Time.zone.now
 
-      new_record = induction_record.dup
-      induction_record.changing!(time_now)
-
-      default_attrs = {
-        start_date: time_now,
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Looks good, feels like it makes sense. Did you intend to commit the change to secure headers here?

tonyheadford:
> Looks good, feels like it makes sense. Did you intend to commit the change to secure headers here?

Yes, it was the root of weird feature spec failures we were having - @StupidCodeFactory spotted it and I'd thought he'd beat me to the fix

PR #2373, opened 2022-08-02
Better clawback service with logging of exceptions

asmega:
### Context

- Ticket: n/a

- This improves the class that processes clawbacks in csv format
- If there are any exceptions they are caught and logged and no longer exit the process part way thru

### Guidance to review

- none


peteryates (approved):
Makes sense

PR #2380, opened 2022-08-03
[CN-422] Add NPQ Application Eligibility import system

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CN-422

#### Changes to existing systems
 
- Adds NPQApplications and Admin::NPQApplications namespaces
- Move existing code into these namespaces
- Add NPQApplication export index page

#### New systems

- Add Eligibility Imports system
- Add GOOGLE_DRIVE_NPQ_DOWNLOAD_FOLDER_ID env variable for controlling where imports should be pulled from

### Guidance to review

1. Add GOOGLE_DRIVE_NPQ_DOWNLOAD_FOLDER_ID env variable (re-using the dev test  GOOGLE_DRIVE_NPQ_UPLOAD_FOLDER_ID allows you to easily obtain a value for this)
2. Upload file to connected Google Drive folder following format of example file from schedule page to change an NPQApplication record in your database.
3. Ensure sidekiq is running locally
4. Schedule upload via schedule form.
5. Check that NPQApplication has changed.

#### Export Schedule Page

This is the page that currently exists within the CPD admin interface. Currently when you click “NPQ Applications” you just go right to this form and can schedule an export.

This page has been moved to be reached via the new “Schedule Export” button on the exports index page.

Export schedule page

#### Exports Index Page

This page gives us a quick audit trail of exports, it doesn’t surface who did what export but it shows that we’re storing that data and lets people know what has been exported and when.

Exports index page

#### Imports index Page

This is the main new page to the new import system.

Reached via a tab on the NPQ applications section this provides a log of imports and their status, the screenshot shows the possible states that imports can be in and shows how they are presented to the user. 

Once an import is scheduled the import itself will happen in the background and progress can be checked by refreshing this page.

Imports index page

#### Imports Show Page

Once an import is complete or has failed you can go to this page where you can see the results of an import. This one in particular is a page for one that completed with errors, it will display a list of errors along with details of what was imported. It will display errors that prevented specific records from saving for example.

Import show page

#### Imports Schedule Page

This is the imports page mirror of the export schedule page. It takes just a name of the file that you’re uploading, it will then go to google drive, find the file, download it, and process it.

Screenshot 2022-08-03 at 12 03 56


app/models/npq_applications/eligibility_import.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    scope :new_to_old, -> { order(created_at: :desc) }
+
+    def perform_later
+      Admin::NPQApplications::EligibilityImportJob.perform_later(self)
+    end
+
+    def done?
+      completed? || failed? || completed_with_errors?
+    end
```
peteryates, line 32: The word 'done' suggests success to me, could `#finished?` work here?

peteryates (approved):
Looks great. Left one comment on the naming of a method but it's not really important.

Perhaps a few bits of the views could be shortened using [govuk_summary_list](https://govuk-components.netlify.app/components/summary-list) and [govuk_table](https://govuk-components.netlify.app/components/table/), but it's fine as is.

app/models/npq_applications/eligibility_import.rb
Lockyy: I went with processed to mirror the processed_at attribute.

PR #2383, opened 2022-08-04
[CPDLP-1577] V3 API docs

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1577

- Add v3 api docs
- Implementation is not present therefore not out the documented endpoints actually function
- Tech docs renderer has been updated to make shared schemas better

### Guidance to review

- https://ecf-review-pr-2383.london.cloudapps.digital/api-reference


StupidCodeFactory (approved):

docs/lib/govuk_tech_docs/open_api/templates/api_reference_full.html.erb
```diff
@@ -2,7 +2,7 @@
   <%= Rouge::Themes::Github.render(scope: '.highlight') %>
 </style>
 
-<h1 id="<%= info.title.parameterize %>">Endpoints <%= document.info.version %></h1>
+<h1 id="<%= info.title.parameterize %>"><%= document.info.title %></h1>
```
peteryates, line 5:
```suggestion
<%= tag.h1(document.info.title, id: info.title.parameterize) %>
```

docs/lib/govuk_tech_docs/open_api/templates/responses.html.erb
```diff
@@ -53,7 +53,29 @@
           end
         end %>
 
-        <% if request_body.present? %>
+        <% example_done = false %>
+
+        <% if response.content['application/json'] %>
+          <% if response.content['application/json']["examples"] %>
+            <% response.content['application/json']["examples"].each do |name, hash| %>
```
peteryates, line 60: Would using a `response.content.dig&.each` make this easier?

docs/lib/govuk_tech_docs/open_api/templates/api_reference_full.html.erb
asmega: i'll work this into the next PR

docs/lib/govuk_tech_docs/open_api/templates/responses.html.erb
asmega: maybe i'll take a look in the next PR

docs/lib/govuk_tech_docs/open_api/templates/api_reference_full.html.erb
asmega: i just tried this and don't have access to `tag`. I'm guessing this is a rails view helper? If so unfortunately we're not in a rails app not sure where to grab this helper from?

docs/lib/govuk_tech_docs/open_api/templates/responses.html.erb
asmega: tbh i'm hoping this is throw away code. all this view code is really not ideal and should be abstracted elsewhere. i don't think this tech docs swagger combo is good and we should aim to replace it with something else. both from a technical maintenance and user perspective. I think it has great negatives for all stake holders in play.

PR #2384, opened 2022-08-04
[CST-891] Accept slashes and letters in TRN input

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-891)

Allow SITs and participants to enter TRNs as per the guidance on the forms and handle the stripping of the unwanted characters.
TRN could contain letters and slashes e.g. RP99/12345
Once stripped of everything except numerical digits:
* TRN should be between 5 and 7 digits long
* We should pad with zeros on the left to make the TRN 7 digits long if not e.g. "123456" becomes "0123456"

Adds a `TeacherReferenceNumber` class that handles the formatting
Adds a `TeacherReferenceNumberValidator` to simplify validation of TRN entry.
Modify forms that add, transfer or validate participants to use these new classes

### DIFF
--- a/app/controllers/schools/transferring_participants_controller.rb
+++ b/app/controllers/schools/transferring_participants_controller.rb
@@ -286,7 +286,7 @@ module Schools
 
     def check_against_dqt?
       @transferring_participant_form.full_name.present? &&
-        @transferring_participant_form.trn.present? &&
+        @transferring_participant_form.formatted_trn.present? &&
         @transferring_participant_form.date_of_birth.present?
     end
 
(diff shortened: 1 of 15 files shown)


peteryates:
The code looks great but the change makes me a bit uneasy. [We describe a TRN as](https://www.gov.uk/guidance/teacher-reference-number-trn):

> A TRN is a 7 digit number that uniquely identifies you in the education sector in England.

It worries me that:

* by allowing 5 digits rather than 7, people confused about their TRN might enter the wrong thing
* if we're going to use a subset of the number they entered, will it be confusing if they see it in its numeric form anywhere else (in the rest of the journey, in correspondence, etc)
* are we confusing the matter by having different guidance on what TRN is than elsewhere in DfE

If this is all accounted for and the team _definitely_ want this I'll approve tomorrow 😅

tonyheadford:
> The code looks great but the change makes me a bit uneasy. [We describe a TRN as](https://www.gov.uk/guidance/teacher-reference-number-trn):
> 
> > A TRN is a 7 digit number that uniquely identifies you in the education sector in England.
> 
> It worries me that:
> 
> * by allowing 5 digits rather than 7, people confused about their TRN might enter the wrong thing
> * if we're going to use a subset of the number they entered, will it be confusing if they see it in its numeric form anywhere else (in the rest of the journey, in correspondence, etc)
> * are we confusing the matter by having different guidance on what TRN is than elsewhere in DfE
> 
> If this is all accounted for and the team _definitely_ want this I'll approve tomorrow 😅

Good questions, we have been allowing 5 - 7 digit TRNs as long as I can remember. They are zero padded to 7 digits and not always provided to us as 7 digits (I assume if cut-n-pasted from a spreadsheet). The slashes and alpha chars are also common and not needed. This change is really about making it easier for SITs and participants to give us their TRN without insisting on digits only and making it inline with the guidance we give when entering their TRN.  It's being asked for by design and content colleagues.  However, if it's not the correct thing to do then we should check further

peteryates:
@tonyheadford - thanks for the additional context.

peteryates (approved):
Looks good.

PR #2393, opened 2022-08-11
CST-808: make trn mandatory for newly registered participants

ltello:
### Context

- Tickets: 
  [SITs must enter a TRN in order to register their ECT](https://dfedigital.atlassian.net/browse/CST-808)
  [SITs must enter a TRN in order to register their Mentor](https://dfedigital.atlassian.net/browse/CST-870)

   - Copy changes in the add participant journey to state TRN is now mandatory for SITs to register new participants
   - If a participant record with the provided trn already exists in the system, the SIT can continue transferring the participant instead of duplicating it. (Before this PR, participant's name, trn and dob musted match an existing participant to deviate the SIT to the transfer journey).
   - If participant TRA record is not found with participant's name, trn and dob, National Insurance number can now be used to try again and find it.
   - If no TRA record is found with nino either, a new screen is displayed to the user and the registration journey stops.
   
### Guidance to review

**Add ECT with National Insurance Number - Success**
Screenshot 2022-08-12 at 13 33 46
Screenshot 2022-08-12 at 13 38 05
Screenshot 2022-08-12 at 13 38 25
Screenshot 2022-08-12 at 13 38 59
Screenshot 2022-08-12 at 13 50 31
Screenshot 2022-08-12 at 13 51 13
Screenshot 2022-08-12 at 13 51 26
Screenshot 2022-08-12 at 13 51 37
Screenshot 2022-08-12 at 13 51 54
Screenshot 2022-08-12 at 13 52 04
Screenshot 2022-08-12 at 13 52 23
Screenshot 2022-08-12 at 13 52 29
Screenshot 2022-08-12 at 13 52 37

**Add ECT with National Insurance Number - Fail**
Screenshot 2022-08-12 at 13 33 46
Screenshot 2022-08-12 at 13 38 05
Screenshot 2022-08-12 at 13 38 25
Screenshot 2022-08-12 at 13 38 59
Screenshot 2022-08-12 at 13 50 31
Screenshot 2022-08-12 at 13 51 13
Screenshot 2022-08-12 at 13 51 26
Screenshot 2022-08-12 at 13 51 37
Screenshot 2022-08-12 at 14 36 21


app/forms/schools/add_participant_form.rb
```diff
@@ -218,18 +217,18 @@ def check_records
...
     def check_for_existing_profile
-      self.existing_participant_profile = ParticipantProfile::ECF.joins(:ecf_participant_validation_data)
-                                                        .where("LOWER(full_name) = ? AND trn = ? AND date_of_birth = ?",
-                                                               full_name.downcase,
-                                                               formatted_trn,
-                                                               date_of_birth).first
+      self.existing_participant_profile = ParticipantProfile::ECF
+                                            .joins(:ecf_participant_validation_data)
+                                            .where(ecf_participant_validation_data: { trn: formatted_trn })
+                                            .first
```
StupidCodeFactory, line 223: What happens when we have more than one profile with the same TRN?

spec/lib/national_insurance_number_spec.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe NationalInsuranceNumber do
+  describe "valid?" do
```
StupidCodeFactory, line 6:
```suggestion
  describe "#valid?" do
```

spec/support/features/steps/changes_of_circumstance_steps.rb
```diff
@@ -71,34 +71,23 @@ def and_sit_reported_participant(sit_name, participant_name, participant_email,
                                            .add_participant_details
                                            .continue
                                            .choose_to_add_an_ect_or_mentor
+        # participant_dob = Date.new(1990, 9, 1)
```
StupidCodeFactory, line 74: Do we still need that?

tonyheadford (commented):
Looks great, however a couple of things I noticed - we should allow upper and lower case characters in the national insurance number to be entered, I don't _think_ you're doing that, and also in the places where you provide an example NINO it should be the one on the GDS guidelines https://design-system.service.gov.uk/patterns/national-insurance-numbers/

app/forms/schools/add_participant_form.rb
ltello: The SIT will be asked if this new participant actually is a school transfer. 
If they answer YES they will be deviated to the transfer journey.
Otherwise they will see a screen pointing them to contact Support

spec/support/features/steps/changes_of_circumstance_steps.rb
ltello: Of course not. Thanks!

ltello:
> 
I have just reused the existing NINO REGEXP we have here:
https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/forms/participants/participant_validation_form.rb#L31

tonyheadford:
> @tonyheadford I have just reused the existing NINO REGEXP we have here: https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/forms/participants/participant_validation_form.rb#L31
> 
> Also, the sample NINO, I can change it to the one in the guides, the thing is that NINO is invalid according to the rules of NINO format and that's the reason I changed it

I think that's why it's used because it doesn't belong to anyone. I think you would just need to force upper case on the input before testing against the REGEX

ltello:
@tonyheadford 
I understand the intention but it feels to me a bit weird to give an example which is invalid by definition. 
Anyway, example NINO reverted and lowercase letters allowed.  Please, review again

config/locales/en.yml
```diff
@@ -140,7 +140,7 @@ en:
       blank: "Please select a provider"
     national_insurance_number:
       blank: Enter the national insurance number (NINo)
-      invalid: National Insurance number is not the correct format (for example, AB 12 34 56 A or AB123456A)
+      invalid: National Insurance number is not the correct format (for example, QQ 12 34 56 A or QQ123456A)
```
peteryates, line 143: I think in this case the example format should be in the hint rather than the error message. There's a similar example in [the design system](https://design-system.service.gov.uk/components/error-message/#label).

ltello: @peteryates  the hint already have it. See [nino.html](https://github.com/DFE-Digital/early-careers-framework/pull/2393/files#diff-3ee2729e3caa6215a96ae52f457f8cb06749138a5c3a3aca1281994c559ac9e6R20) template. 
I am just sticking to the same pattern as the existing [TRN](https://github.com/DFE-Digital/early-careers-framework/pull/2393/files#diff-44438ce218f5287c58d0017f965d888715635d94280669896f75841fbd7b4cd7L145).
But happy to remove the `(for example, QQ 12 34 56 A or QQ123456A)` if you think it is better that way

ltello: @peteryates Invalid nino error message updated after talking with @JoDimbleby. Thanks!!

lazydevorg (changes requested):

app/forms/schools/add_participant_form.rb
```diff
@@ -218,18 +217,18 @@ def check_records
...
 
     def check_for_existing_profile
-      self.existing_participant_profile = ParticipantProfile::ECF.joins(:ecf_participant_validation_data)
-                                                        .where("LOWER(full_name) = ? AND trn = ? AND date_of_birth = ?",
-                                                               full_name.downcase,
-                                                               formatted_trn,
-                                                               date_of_birth).first
+      self.existing_participant_profile = ParticipantProfile::ECF
+                                            .joins(:ecf_participant_validation_data)
+                                            .where(ecf_participant_validation_data: { trn: formatted_trn })
```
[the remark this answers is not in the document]
ltello: Yes, it is after our discussions to make TRN mandatory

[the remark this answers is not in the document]
ltello: Yes, it is after our discussions to make TRN mandatory

app/lib/national_insurance_number.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+private
+
+  def parse_nino
+    parsed = nino.to_s.upcase.gsub(/\s*/, "")
+    return [nil, :blank]   if parsed.blank?
+    return [nil, :invalid] if parsed !~ REGEXP
+
+    [parsed, nil]
```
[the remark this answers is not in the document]
ltello: We might have as well raised specialized NationalIinsuranceNumber::Error subclass instances but for consistency, I was just following the pattern based on keys used for the TRN.
Please [see](https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/lib/teacher_reference_number.rb#L3)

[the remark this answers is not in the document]
lazydevorg: @ltello I get your point. Furthermore, since the class is handling this privately I think it's ok 👍

lazydevorg (approved):

PR #2397, opened 2022-08-15
Content updates for TRN error messages 

tonyheadford:
### Context

Content tweaks to TRN validation error messages.


peteryates (approved):
Definitely an improvement

PR #2401, opened 2022-08-16
[CPDLP-1534] Providers can trigger a clawback via the API

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1534

* Added extension to `Finance::ClawbackDeclaration` to make clawback when the declaration is paid.

### DIFF
--- a/app/services/void_participant_declaration.rb
+++ b/app/services/void_participant_declaration.rb
@@ -1,34 +1,46 @@
 # frozen_string_literal: true
 
 class VoidParticipantDeclaration
-  attr_accessor :cpd_lead_provider, :id
+  attr_reader :participant_declaration
 
-  def initialize(cpd_lead_provider:, id:)
-    @cpd_lead_provider = cpd_lead_provider
-    @id = id
+  def initialize(participant_declaration:)
+    @participant_declaration = participant_declaration
   end
 
   def call
-    # TODO: these should be moved to activemodel validations
-    # and not controller level exceptions
-    raise Api::Errors::InvalidTransitionError, I18n.t(:declaration_already_voided) if declaration.voided?
-    raise Api::Errors::InvalidTransitionError, I18n.t(:declaration_not_voidable) unless declaration.voidable?
-
-    ApplicationRecord.transaction do
-      declaration.make_voided!
-      line_item.voided! if line_item
+    if participant_declaration.paid?
+      make_awaiting_clawback
+    else
+      make_voided
     end
 
-    Api::V1::ParticipantDeclarationSerializer.new(declaration).serializable_hash.to_json
+    participant_declaration
+  end
+
+  def make_awaiting_clawback
+    clawback = Finance::ClawbackDeclaration.new(participant_declaration:)
+    clawback.call
+
+    if clawback.errors.any?
+      raise Api::Errors::InvalidTransitionError, clawback.errors.full_messages.join(", ")
+    end
+  end
+
+  def make_voided
+    # TODO: these should be moved to activemodel validations
+    # and not controller level exceptions
+    raise Api::Errors::InvalidTransitionError, I18n.t(:declaration_already_voided) if participant_declaration.voided?
+    raise Api::Errors::InvalidTransitionError, I18n.t(:declaration_not_voidable) unless participant_declaration.voidable?
+
+    ApplicationRecord.transaction do
+      participant_declaration.make_voided!
+      line_item.voided! if line_item
+    end
   end
 
 private
 
   def line_item
-    declaration.statement_line_items.find_by(state: %w[eligible payable])
-  end
-
-  def declaration
-    @declaration ||= ParticipantDeclaration.for_lead_provider(cpd_lead_provider).find(id)
+    participant_declaration.statement_line_items.find_by(state: %w[eligible payable submitted])
   end
 end
--- a/app/services/void_participant_declaration.rb
+++ b/app/services/void_participant_declaration.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class VoidParticipantDeclaration
-  attr_reader :participant_declaration
-
   def initialize(participant_declaration:)
     @participant_declaration = participant_declaration
   end
@@ -39,6 +37,8 @@ class VoidParticipantDeclaration
   end
 
 private
+
+  attr_reader :participant_declaration
 
   def line_item
     participant_declaration.statement_line_items.find_by(state: %w[eligible payable submitted])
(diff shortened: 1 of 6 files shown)


app/services/void_participant_declaration.rb
```diff
@@ -1,34 +1,46 @@
...
   end
 
   def call
+    if participant_declaration.paid?
+      make_awaiting_clawback
+    else
+      make_voided
+    end
+
+    Api::V1::ParticipantDeclarationSerializer.new(participant_declaration).serializable_hash.to_json
```
asmega, line 17: while we're here any chance we can make this return `participant_declaration` instead and let the controller decide what serialiser to use?

app/services/void_participant_declaration.rb
```diff
@@ -1,34 +1,46 @@
...
 
 private
 
   def line_item
-    declaration.statement_line_items.find_by(state: %w[eligible payable])
-  end
-
-  def declaration
-    @declaration ||= ParticipantDeclaration.for_lead_provider(cpd_lead_provider).find(id)
+    participant_declaration.statement_line_items.find_by(state: %w[eligible payable])
```
asmega, line 44: on a side note possible bug there. this should probably cover `submitted` too

asmega (approved):

StupidCodeFactory:
@mooktakim  I think this acceptance criteria isn't fulfilled, would you mind adding a test that covers this acceptance? 

`A provider can post a declaration of a type already associated with the participant and course and the declaration be eligible, providing the existing declaration is awaiting_clawback or clawed_back state`

StupidCodeFactory:
I'd also expect to see a few test around the change schedule post clawback, if you could also add those that'd be wonderful, thanks.

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -375,6 +375,48 @@ def build_params(attributes)
...
+                cpd_lead_provider:,
+                participant_profile: ect_profile,
+                course_identifier: valid_params[:course_identifier],
+              )
+            end
+
+            it "returns 200" do
+              params = build_params(valid_params)
+              post "/api/v1/participant-declarations", params: params
+              expect(response.status).to eq 200
```
StupidCodeFactory, line 396: @mooktakim here you want to assert that the a new declaration has been created and that it also is in the correct state alongside with a statement line item in the correct test.

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -375,6 +375,87 @@ def build_params(attributes)
...
+            expect(ParticipantDeclaration.first).to be_awaiting_clawback
+
+            params = build_params(valid_params)
+            post "/api/v1/participant-declarations", params: params
+            expect(response.status).to eq 200
+
+            declarations = ParticipantDeclaration.all
+            expect(declarations.count).to eq(2)
+            expect(declarations[0]).to be_awaiting_clawback
+            expect(declarations[1]).to be_submitted
```
StupidCodeFactory, line 403: I'd personally prefer something a tad more explicit:

```ruby
params = build_params(valid_params)
expect {
  post "/api/v1/participant-declarations", params: params
  expect(response.status).to eq 200
}.to change { ect_profile.participant_declaration.submitted.count }.from(0).to(1)

# sanity check that the initial declaration is still on the correct state
expect(existing_participant_declaration).to be_awaiting_clawback
```

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -375,6 +375,87 @@ def build_params(attributes)
...
+
+          it "returns 200" do
+            expect(ParticipantDeclaration.count).to eq(1)
+            expect(ParticipantDeclaration.first).to be_awaiting_clawback
+
+            params = build_params(valid_params)
+            post "/api/v1/participant-declarations", params: params
+            expect(response.status).to eq 200
+
+            declarations = ParticipantDeclaration.all
```
StupidCodeFactory, line 400: the ordering isn't guaranteed that's usually yields flaky specs

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -375,6 +375,87 @@ def build_params(attributes)
...
+              :awaiting_clawback,
+              user: ect_profile.user,
+              cpd_lead_provider:,
+              participant_profile: ect_profile,
+              course_identifier: valid_params[:course_identifier],
+            )
+          end
+
+          it "returns 200" do
+            expect(ParticipantDeclaration.count).to eq(1)
```
StupidCodeFactory, line 393: To add an extra layer of assurance you always want to be a bit more precise in those types of assertions and you achieve those two lines in one assertion as well:

```ruby
expect(ect_profile.participant_profiles.awaiting_clawback.count).to eq(1)

File: spec/requests/api/v1/participant_declarations_spec.rb [hunk 7]
```diff
@@ -375,6 +375,87 @@ def build_params(attributes)
...
+            expect(ParticipantDeclaration.first).to be_clawed_back
+
+            params = build_params(valid_params)
+            post "/api/v1/participant-declarations", params: params
+            expect(response.status).to eq 200
+
+            declarations = ParticipantDeclaration.all
+            expect(declarations.count).to eq(2)
+            expect(declarations[0]).to be_clawed_back
+            expect(declarations[1]).to be_submitted
```
StupidCodeFactory, line 430: same remark as above ^^

spec/requests/api/v1/participant_declarations_spec.rb
StupidCodeFactory: Those could be passing but not not for the reason you think, another factory may have created that record as a side effect, which happens a lot with FactoryBot when you're not careful.

app/services/void_participant_declaration.rb
```diff
@@ -1,34 +1,46 @@
 # frozen_string_literal: true
 
 class VoidParticipantDeclaration
-  attr_accessor :cpd_lead_provider, :id
+  attr_reader :participant_declaration
```
StupidCodeFactory, line 4: This should be private, we're not accessing it from anywhere

spec/services/void_participant_declaration_spec.rb
```diff
@@ -9,14 +9,13 @@
...
   let(:cpd_lead_provider) { profile.school_cohort.school.partnerships[0].lead_provider.cpd_lead_provider }
-  let(:another_cpd_lead_provider) { create(:cpd_lead_provider) }
 
   before do
     create(:partnership, school:)
   end
 
   describe "#call" do
-    let(:declaration) do
+    let(:participant_declaration) do
```
StupidCodeFactory, line 18: 🎖️ thanks for renaming this properly

StupidCodeFactory (approved):

PR #2403, opened 2022-08-17
CPDLP-1655-npqs-training-status

leoapost:
The drilldown page is using the cached training_status value from the
participant profile record to display the participant training status
for NPQ participants, when the API is using the latest participant profile
state record. These two can go out of sync and cause issues and confusion.

Replacing the ParticipantProfile training_status in the drilldown page with the 
ParticipantProfile state method ensures the correct state is always displayed.

Eventually the InductionRecord will replace both the training_status and the 
ParticipantProfileState. ECF participants already using the InductionRecord,
but the NPQ ones not just yet.

### Context

- Ticket: CPDLP-1655

I changed the drilldown page to display the `ParticipantProfile.state` instead of the `training_status`, but only for the NPQs

### DIFF
--- a/app/views/finance/change_training_statuses/new.html.erb
+++ b/app/views/finance/change_training_statuses/new.html.erb
@@ -13,7 +13,11 @@
       <p>
         <strong>Training status</strong>
         <br/>
-        <%= @participant_profile.training_status %>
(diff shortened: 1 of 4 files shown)


asmega (approved):
It's not ideal but think this will do for now

PR #2408, opened 2022-08-19
CST-949: show school funded fip participants on sit participants dash…

ltello:
…board

### Context

- Ticket: [ECTs/Mentors unable to view full participant record](https://dfedigital.atlassian.net/browse/CST-949)

Make school funded fip participants visible in the participants dashboard

### DIFF
--- a/app/models/induction_record.rb
+++ b/app/models/induction_record.rb
@@ -47,8 +47,9 @@ class InductionRecord < ApplicationRecord
     withdrawn: "withdrawn",
   }, _prefix: "training_status"
 
-  scope :fip, -> { joins(:induction_programme).merge(InductionProgramme.full_induction_programme) }
   scope :cip, -> { joins(:induction_programme).merge(InductionProgramme.core_induction_programme) }
+  scope :fip, -> { joins(:induction_programme).merge(InductionProgramme.full_induction_programme) }
+  scope :sffip, -> { joins(:induction_programme).merge(InductionProgramme.school_funded_fip) }
 
   scope :end_date_null, -> { where(end_date: nil) }
   scope :end_date_in_past, -> { where(end_date: ...Time.zone.now) }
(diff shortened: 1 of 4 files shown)


peteryates (commented):
While the changes here are inkeeping with the style, I think this class could do with some attention. The test setup is incredibly involved.

Happy to take a closer look next week. Perhaps we let this through and the class for some attention or a refactor.

ltello:
> While the changes here are inkeeping with the style, I think this class could do with some attention. The test setup is incredibly involved.
> 
> Happy to take a closer look next week. Perhaps we let this through and the class for some attention or a refactor.

I agree. The thing is that this dashboard is being redesigned by the UX and content people, so not sure if it is worth a refactor at this stage.

app/models/induction_record.rb
```diff
@@ -47,8 +47,9 @@ class InductionRecord < ApplicationRecord
     withdrawn: "withdrawn",
   }, _prefix: "training_status"
 
-  scope :fip, -> { joins(:induction_programme).merge(InductionProgramme.full_induction_programme) }
   scope :cip, -> { joins(:induction_programme).merge(InductionProgramme.core_induction_programme) }
+  scope :fip, -> { joins(:induction_programme).merge(InductionProgramme.full_induction_programme) }
+  scope :sffip, -> { joins(:induction_programme).merge(InductionProgramme.school_funded_fip) }
```
tonyheadford, line 52:
```suggestion
  scope :school_funded_fip, -> { joins(:induction_programme).merge(InductionProgramme.school_funded_fip) }
```
Think this would help make it easier to understand

tonyheadford (commented):
Could we replace occurrences of `sffip` with `:school_funded_fip` to make it easier to understand?

app/models/induction_record.rb
ltello: Sure

PR #2410, opened 2022-08-22
Just log percy snapshot error when build finalized

peteryates:
### Context and changes

We have a few flaky tests and occasionally need to re-run part of the suite. While this isn't ideal, sometimes they pass on the second run but are reported as failed because Percy can't add new screenshots to a build that's already been finalized. [Here's an example](https://github.com/DFE-Digital/early-careers-framework/actions/runs/2903242442) where the build should be green but isn't.

In this instance Percy shouldn't cause the build to fail, so we'll detect finalized build errors and just log them, but `raise` on everything else.

The error message will still be logged.


peteryates:
Rerunning `RSpec (6,3)` to test

PR #2414, opened 2022-08-23
[CST-965] update registration journeys for cip only schools setting up their 2022 23 cohort

ltello:
### Context

- Ticket: [Update registration journeys for CIP-only schools setting up their 2022/23 cohort](https://dfedigital.atlassian.net/browse/CST-965)

- Remove "ects and mentors" row from the school dashboard for CIP-only schools, so they can't manage participants.
- Add copy saying it at the botom.
- Do not show FIP choice when a CIP-only school is choosing an induction programme
- Customize the confirmation screen for schools that choose a school funded fip programme.

### DIFF
--- a/app/helpers/schools/dashboard_helper.rb
+++ b/app/helpers/schools/dashboard_helper.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module Schools
+  module DashboardHelper
+    def manage_ects_and_mentors?(school_cohort)
+      !(school_cohort.school.cip_only? || school_cohort.school_chose_diy?)
+    end
+  end
+end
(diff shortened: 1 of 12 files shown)


app/helpers/schools/dashboard_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Schools
+  module DashboardHelper
+    def manage_ects_and_mentors?(school_cohort)
+      !(school_cohort.school.cip_only? || school_cohort.school_chose_diy?)
```
tonyheadford, line 6:
```suggestion
      school_cohort.full_induction_programme? || school_cohort.core_induction_programme?
```

tonyheadford: This would make more sense to me but may not fix the UI for all of the ones that currently have selected the wrong programme despite until the underlying data is fixed (if necessary?)

tonyheadford (approved):
Looks great, made one optional suggestion

PR #2415, opened 2022-08-23
Stop full name and date of birth from being logged

peteryates:
### Context

I noticed some detail we probably shouldn't be logging while hunting down an error in the logs.

Added the immediate offenders here:

* [x] `full_name`
* [x] `date_of_birth`

Others that _might_ be worth filtering:

* [x] `trn` (_potentially_ identifiable)


tonyheadford (commented):
Yes, it feels `trn` should be filtered too

tonyheadford (approved):

PR #2419, opened 2022-08-25
Move notification email sending to its own service class

peteryates:
### Context

The process of sending notification emails was tricky to follow and only partially covered by tests. This change moves it to its own class and adds some specs, hopefully making it easier to follow.

### Review

Wasn't sure if `Induction` is the right namespace for the class, I put it in there for consistency with the other service classes in use but originally had it in `services/schools/transfers`

### DIFF
--- a/app/controllers/schools/transferring_participants_controller.rb
+++ b/app/controllers/schools/transferring_participants_controller.rb
@@ -123,46 +123,18 @@ module Schools
     #   b. Send email to participant to notify them.
     #
     def send_notification_emails!(induction_record, was_withdrawn_participant)
-      current_lead_provider_users = current_lead_provider&.users || []
-      target_lead_provider_users = participant_lead_provider&.users || []
+      current_lead_provider_profiles = current_lead_provider&.users&.map(&:lead_provider_profile) || []
+      target_lead_provider_profiles = participant_lead_provider&.users&.map(&:lead_provider_profile) || []
 
-      if was_withdrawn_participant
-        current_lead_provider_users.each do |user|
-          ParticipantTransferMailer.provider_transfer_in_notification(
-            induction_record:,
-            lead_provider_profile: user.lead_provider_profile,
-          ).deliver_later
-        end
-      elsif matching_lead_provider_and_delivery_partner?
-        current_lead_provider_users.each do |user|
-          ParticipantTransferMailer.provider_existing_school_transfer_notification(
-            induction_record:,
-            lead_provider_profile: user.lead_provider_profile,
-          ).deliver_later
-        end
-      elsif @transferring_participant_form.switch_to_schools_programme?
-        current_lead_provider_users.each do |user|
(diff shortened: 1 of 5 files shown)


app/controllers/schools/transferring_participants_controller.rb
```diff
@@ -123,46 +123,18 @@ def cannot_add; end
...
-          ParticipantTransferMailer.provider_new_school_transfer_notification(
-            induction_record:,
-            lead_provider_profile: user.lead_provider_profile,
-          ).deliver_later
-        end
-      end
-
-      ParticipantTransferMailer.participant_transfer_in_notification(induction_record:).deliver_later
+      current_lead_provider_profiles = current_lead_provider&.users&.map(&:lead_provider_profile) || []
+      target_lead_provider_profiles = participant_lead_provider&.users&.map(&:lead_provider_profile) || []
```
peteryates, line 127: Not especially fond of chaining like this, feels like this could be achieved through relationships and `through:`

app/services/induction/send_notification_emails.rb
```diff
@@ -0,0 +1,112 @@
...
+  #   a. Send email to incoming lead provider.
+  #   b. Send email to outgoing lead provider, requesting that they withdraw them.
+  #   c. Send email to participant to notify them.
+  #
+  # 3) Moving to target school, but continuing with current lead provider.
+  #
+  #   a. Send email to current lead provider, notifying them to expect a new school.
+  #   b. Send email to participant to notify them.
+  #
+  class SendNotificationEmails < BaseService
```
tonyheadford, line 24: Could this be named something like `SendTransferNotificationEmails` to make it a bit clearer?

app/controllers/schools/transferring_participants_controller.rb
```diff
@@ -123,46 +123,18 @@ def cannot_add; end
...
+
+      Induction::SendTransferNotificationEmails.new(
+        induction_record:,
+        was_withdrawn_participant:,
+        same_delivery_partner: with_the_same_delivery_partner?,
+        same_provider: with_the_same_provider?,
+        switch_to_schools_programme: @transferring_participant_form.switch_to_schools_programme?,
+        current_lead_provider_profiles:,
+        target_lead_provider_profiles:,
+      ).call
```
tonyheadford, line 137:
```suggestion
      Induction::SendTransferNotificationEmails.call(
        induction_record:,
        was_withdrawn_participant:,
        same_delivery_partner: with_the_same_delivery_partner?,
        same_provider: with_the_same_provider?,
        switch_to_schools_programme: @transferring_participant_form.switch_to_schools_programme?,
        current_lead_provider_profiles:,
        target_lead_provider_profiles:,
      )
```

tonyheadford (approved):
One suggestion, the services derived from `BaseService` can just be invoked with `.call` so no need to explicitly instantiate them first, unless you prefer to do it that way of course.

PR #2420, opened 2022-08-25
Prune legacy payment calculator code

asmega:
### Context

- Ticket: n/a

- This removes legacy payment calculator code that is no longer used
- There is some old code that is not used but this is a start to removing it

### Guidance to review

- None


peteryates (approved):
Looks good.

cwrw (approved):

PR #2422, opened 2022-08-26
[CPDLP-1538] Return all cohorts past 2021 when querying participant induction records

cwrw:
### Context

A provider is using the header: `X-With-Server-Date` to time travel on sandbox to test different declarations.
They are getting a 422 when attempting to set milestones that occur in the subsequent year (eg `retain-3`)

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1538

For an ECF participant, when retrieving induction records, we are only returning records from the current cohort. The current cohort is dependant on the time of year if the `multiple_cohorts` flag has been activated. This means when we attempt to return induction records for a cohort with milestones in the next year, we end up querying the incorrect cohort (current cohort will return 2022 although the one we need is 2021). Trying to tie induction records to the wrong cohort will return no records, meaning a 422 is returned incorrectly.

Allow multiple cohorts to be queried to return the correct records instead of only looking at current which could be incorrect.

### Guidance to review
Did I miss anything? Also keen to hear about where is the best place to test this if anyone has any better ideas

To test this POST to `api/v1/participant-declarations/` with a declaration type past started and declaration date in the future where the milestone is in the next year. Also set the `X-With-Server-Date` as well to after that milestone date.

### DIFF
--- a/spec/requests/api/v1/participant_declarations_spec.rb
+++ b/spec/requests/api/v1/participant_declarations_spec.rb
@@ -450,6 +450,62 @@ RSpec.describe "participant-declarations endpoint spec", type: :request do
           end
         end
       end
+
+      context "when participant has been retained", with_feature_flags: { multiple_cohorts: "active" } do
+        let!(:cohort) { create(:cohort, :next) }
+        let!(:started_declaration) { create(:participant_declaration, user: ect_profile.user, cpd_lead_provider:, course_identifier: "ecf-induction", participant_profile: ect_profile) }
+        let(:milestone_start_date) { ect_profile.schedule.milestones.find_by(declaration_type:).start_date }
+        let(:valid_params) do
+          {
+            participant_id: ect_profile.user.id,
+            declaration_type:,
+            declaration_date: milestone_start_date.rfc3339,
+            course_identifier: "ecf-induction",
+            evidence_held: "other",
+          }
+        end
+
+        before do
+          travel_to milestone_start_date + 6.months
+        end
+
+        context "with milestone in the same year as the cohort start year" do
+          let(:declaration_type) { "retained-1" }
+          it "creates a declaration record" do
+            params = build_params(valid_params)
+            expect { post "/api/v1/participant-declarations", params: }.to change(ParticipantDeclaration, :count).by(1).and change(ParticipantDeclarationAttempt, :count).by(1)
+          end
+
+          it "sets the correct declaration type on the declaration record" do
+            params = build_params(valid_params)
+            post "/api/v1/participant-declarations", params: params
+
+            expect(response.status).to eq 200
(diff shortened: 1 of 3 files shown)


asmega (approved):

asmega:
> Also keen to hear about where is the best place to test this if anyone has any better ideas

I guess an alternative would be a spec for `app/services/participants/ecf.rb` (make it if it does not exist?) That way the test can be a little lower down the stack.

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -450,6 +450,62 @@ def build_params(attributes)
...
+            post "/api/v1/participant-declarations", params: params
+
+            expect(response.status).to eq 200
+            declaration = ParticipantDeclaration::ECF.find(JSON.parse(response.body).dig("data", "id"))
+            expect(declaration.declaration_type).to eq(declaration_type)
+          end
+        end
+
+        context "with milestone in the year after the cohort start year" do
+          let!(:reatined_1_declaration) { create(:participant_declaration, user: ect_profile.user, cpd_lead_provider:, course_identifier: "ecf-induction", participant_profile: ect_profile, declaration_type: "retained-1") }
```
StupidCodeFactory, line 490:
```suggestion
          let!(:retained_1_declaration) { create(:participant_declaration, user: ect_profile.user, cpd_lead_provider:, course_identifier: "ecf-induction", participant_profile: ect_profile, declaration_type: "retained-1") }
```

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -450,6 +450,62 @@ def build_params(attributes)
...
+          let!(:reatined_2_declaration) { create(:participant_declaration, user: ect_profile.user, cpd_lead_provider:, course_identifier: "ecf-induction", participant_profile: ect_profile, declaration_type: "retained-2") }
```
StupidCodeFactory, line 491:
```suggestion
          let!(:retained_2_declaration) { create(:participant_declaration, user: ect_profile.user, cpd_lead_provider:, course_identifier: "ecf-induction", participant_profile: ect_profile, declaration_type: "retained-2") }
```

spec/requests/api/v1/participant_declarations_spec.rb
StupidCodeFactory: Suggestion: if we're not using the `let` for any later assertion I think it's clearer to create them in a `before` block.

StupidCodeFactory (approved):
@cwrw approved, there was just a couple of typos

cwrw:
@asmega @StupidCodeFactory addressed comments, thanks for the review. I kept the other higher level tests as they seemed also useful in testing declarations, however good shout around adding lower level concern tests. Will amend the test setup again once the other service class PR is merged

asmega (approved):

PR #2424, opened 2022-08-26
[CPDLP-1721]: Fix bug in cohorts filter on ECF participants endpoint

cwrw:
### Context
When filtering by cohorts while retrieving ECF participants we are getting a 500

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1721

Return a list of records rather than one record to allow mapping the id in the cohorts query. 

### Guidance to review
Is the test setup correct or am I missing something?

### DIFF
--- a/spec/requests/api/v1/ecf_participants_spec.rb
+++ b/spec/requests/api/v1/ecf_participants_spec.rb
@@ -200,6 +200,26 @@ RSpec.describe "Participants API", type: :request do
           end
         end
 
+        context "when cohort parameter is supplied" do
+          it "returns participants only within that cohort" do
+            next_cohort = create(:cohort, :next)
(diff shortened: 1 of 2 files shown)


StupidCodeFactory (approved):

spec/requests/api/v1/ecf_participants_spec.rb
```diff
@@ -200,6 +200,26 @@
           end
         end
 
+        context "when cohort parameter is supplied" do
+          it "returns participants only within that cohort" do
+            next_cohort = create(:cohort, :next)
```
cwrw, line 205: Chatted to @StupidCodeFactory in person but for posterity: I'll revisit this test setup after https://github.com/DFE-Digital/early-careers-framework/pull/2225 is merged to simplify and improve using the new factories

PR #2431, opened 2022-09-01
hardcode Cohort.current to 2021

asmega:
### Context

- Ticket: n/a
- `Cohort.current` is now referring to `2022` whereas there is reference code that expects this to be `2021`

- Hardcode `Cohort.current` to `2021`
- This is in place whilst we fix references to `Cohort.current` to select the desired cohort
- Once that work is complete we should be able to revert this change

### Guidance to review

- none

### DIFF
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -8,12 +8,7 @@ class Cohort < ApplicationRecord
   has_many :npq_contracts
 
   def self.current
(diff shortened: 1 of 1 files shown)


StupidCodeFactory (approved):

peteryates (approved):
Makes sense in the short term.

asmega:
Closing in favour of https://github.com/DFE-Digital/early-careers-framework/pull/2436

PR #2438, opened 2022-09-02
[CPDNPQ-581] Accept employment_type data from NPQ and expose in NPQ application exports

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-581

- Accept employment_type from NPQ when persisting application data
- Add employment_type to NPQ application export


db/migrate/20220901105322_add_employment_type_to_npq_applications.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddEmploymentTypeToNPQApplications < ActiveRecord::Migration[6.1]
+  def change
+    add_column :npq_applications, :employment_type, :string
```
StupidCodeFactory, line 5: @Lockyy is this field a free text field or should we add an enum to constrain its possible values?

Lockyy: This one is a free text field. 

However, even for fixed sets of values within NPQ we've not been applying enums on the ECF side for npq_applications to allow the NPQ app to freely change options without needing PRs on both sides.

lazydevorg (approved):

PR #2439, opened 2022-09-05
Intermittent Failure - Use correct updated_at for expected_json

Lockyy:
### Context

The serializer was using the user and the expectation was using the profile. This lead to intermittent failures due to possible rounding differences between the two.

Use the correct updated_at column

https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/serializers/api/v1/participant_serializer.rb#L88


peteryates (approved):
Nicely done!

PR #2442, opened 2022-09-05
Fix cohort issue for NQT+1 check

tonyheadford:
### Context

The participant validation service was using `Cohort.current.start_year` in its check for a previous induction. This check is only meant to catch pre-Sept 2021 start dates and is now wrongly flagging participants as being NQT+1

Fix the date being checked against to 1/9/2021

### DIFF
--- a/app/services/participant_validation_service.rb
+++ b/app/services/participant_validation_service.rb
@@ -41,7 +41,8 @@ private
     return true if validation_data["induction"]["completion_date"].present?
     return false if validation_data["induction"]["start_date"].nil?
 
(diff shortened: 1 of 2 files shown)


cwrw (approved):
👍 Might be worth also checking `app/services/analytics/ecf_school_service.rb` as that might also need to change

tonyheadford:
> 👍 Might be worth also checking `app/services/analytics/ecf_school_service.rb` as that might also need to change

Think that's doing something different, I'm assuming that is only publishing the latest years info into analytics.

PR #2448, opened 2022-09-06
Upgrade to latest version of `govuk-frontend`

peteryates:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CST-897 (groundwork)

The version of GOV.UK frontend we're using is quite a long way behind current, and it's out of sync with govuk-components and govuk-formbuilder, both of which are tied to a single version of the design system.

This PR brings them all up to date.

- Use the `govuk_header` component for the main header
- Upgrade from govuk-frontend 3.14.0 to 4.3.1
- Replace deprecated govuk-main-wrapper with spacing
- Upgrade to most recent version of `govuk-components`
- Switch to [new link styles](https://designnotes.blog.gov.uk/2021/07/07/making-links-easier-to-see-and-read-on-gov-uk/)
- Upgrade to version 3.2.1 of govuk-components
- Upgrade to 3.1.0 of the form builder

### Guidance to review

This will need quite a detailed look through, especially on pages where we've deviated from the design system or are using hardcoded markup instead of the components library.

I suspect there will be minor changes to padding and spacing here and there, they're worth highlighting here so we can see if they're big enough to warrant a fix. For example, if you look at the blue masthead screenshots below there is a _minor_ difference in height due to the design system dropping a helper class and I had to re-implement it using the spacing helpers.

### Preview

| Before (`3.14.0`) | After (`4.3.1` |
| ------ | -------|
| Screenshot from 2022-09-07 13-02-05 | Screenshot from 2022-09-07 13-02-08  |
| Screenshot from 2022-09-07 13-02-45 | Screenshot from 2022-09-07 13-03-04 |
| Screenshot from 2022-09-07 13-09-12 | Screenshot from 2022-09-07 13-09-10 | 
| Screenshot from 2022-09-07 13-10-22 | Screenshot from 2022-09-07 13-10-20 |

### DIFF
--- a/app/webpacker/styles/application.scss
+++ b/app/webpacker/styles/application.scss
@@ -28,6 +28,7 @@ $govuk-image-url-function: frontend-image-url;
 @import "lead_providers/contents_list";
 @import "lead_providers/syntax_highlighting";
 @import "lead_providers/statements";
+@import "pagination";
 @import "subnav";
 @import "primary-nav";
 @import "search_box";
--- a/app/webpacker/styles/pagination.scss
+++ b/app/webpacker/styles/pagination.scss
@@ -1,0 +1,3 @@
+nav.govuk-pagination {
+  justify-content: center;
+}


Gemfile
```diff
@@ -73,9 +73,6 @@ gem "strong_migrations", "~> 1.2"
 # Acts as State Machine for participant and declaration states
 gem "aasm"
 
-# Pagination
-gem "pagy", "~> 5.10", ">= 5.10.1"
```
asmega, line 77: are you sure you can remove pagy? It's used here for example https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/controllers/concerns/api_pagination.rb#L5

peteryates: Pagy is automatically pulled in by the latest version of `govuk-components`, I've switched over to using the [pagination component](https://govuk-components.netlify.app/components/pagination/).

asmega (approved):

peteryates:
Going to hold off merging this for a while as we have other priorities; it's a wide-reaching change that affects every page.

leodfe (approved):

app/components/admin/participants/table.html.erb
```diff
@@ -17,4 +17,4 @@
   </table>
 </div>
 
-<%== govuk_paginate profiles, @pagy %>
+<%== govuk_pagination(pagy: @pagy) %>
```
cwrw, line 20: something I noticed about pagination is that it's no longer centered within the page, is this intentional to adhere to the component version?

cwrw (approved):
went through a few users and looks good, just subtle table alignment changes or dropdown width which is negligible, just a question around the pagination otherwise 👍

peteryates:
Originally I wondered about the pagination too, but left it as technically the old one was left aligned (although you could only tell when you weren't on the first page)

I've made it resemble the old one a bit more:

| Original | Before centring | After centring |
| --------- | ---------------- | -------------- |
| Screenshot from 2022-10-11 13-53-40 | Screenshot from 2022-10-11 13-45-07 | Screenshot from 2022-10-11 13-49-25 |

cc @cwrw

cwrw (approved):
Thanks @peteryates looks great 👍

PR #2449, opened 2022-09-06
[CPDLP-1547] - ensure no double posting

StupidCodeFactory:
### Context

- Ticket:


db/migrate/20220906152545_add_unique_index_participant_declaration.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddUniqueIndexParticipantDeclaration < ActiveRecord::Migration[6.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :participant_declarations, %i[cpd_lead_provider_id participant_profile_id declaration_type course_identifier state], unique: true, name: :unique_declaration_index, algorithm: :concurrently, where: "state NOT IN ('voided', 'ineligible')"
```
asmega, line 7: would it not be better to whitelist rather than blacklist on the `where`?

asmega: although not likely is is technically possible to clawback multiple times there having multiple `clawed_back` should be permitted

StupidCodeFactory: 👍 sounds sensible

app/services/record_declaration.rb
```diff
@@ -111,16 +111,29 @@ def validate_milestone_exists
...
+  rescue ActiveRecord::RecordNotUnique
+    participant_profile
+      .participant_declarations
+      .submitted
+      .or(
+        participant_profile
+          .participant_declarations
+          .billable,
+      )
+      .find_by!(declaration_parameters)
```
peteryates, line 136: I'm not sure I see the need for the query to be repeated after the `rescue`. If the initial `find_or_create_by!` doesn't find a match the `find_by!` can't either. Should we just `fail`/`raise` instead?

StupidCodeFactory: @peteryates that was to manage a race condition where it appears the a lead provider was able to double post a declaration. 
I've actually replaced that with `create_or_find_by!` which actually does  the `ActiveRecord::RecordNotUnique` exception handling and issue the find.

spec/models/participant_declaration_spec.rb
```diff
@@ -300,4 +300,52 @@
...
+    before { described_class.create!(attributes) }
+
+    it "raises an not unique error" do
+      expect { described_class.create!(attributes) }.to raise_error ActiveRecord::RecordNotUnique
+    end
+
+    context "when the declaration state id voided" do
+      let(:state) { :voided }
+
+      it "raises an not unique error" do
```
cwrw, line 330:
```suggestion
      it "does not raises a unique error" do
```
here and below

cwrw (approved):

spec/models/participant_declaration_spec.rb
peteryates: "does not raise a unique error"

PR #2451, opened 2022-09-06
CST-1000: service object to amend participants cohort

ltello:
### Context

 Support is receiving loads of requests to change participants to a different cohort after wrong registration by the SITs.

- Ticket: [CST-1000](https://dfedigital.atlassian.net/browse/CST-1000)

Backend code (service + form) to serve the frontend when this functionality is added to the admin console:
Given a list of email addresses and the start year of the cohorts to move the participants from/to,
return a list emails of participants the service was able to move and a list of emails with errors for those that couldn't be moved.


app/forms/schools/amend_participant_cohort_form.rb
```diff
@@ -0,0 +1,109 @@
...
+                                                      .find_by(participant_identity:)
+    end
+
+    def persist
+      ActiveRecord::Base.transaction do
+        induction_record.update!(induction_programme:, start_date:, schedule:)
+        participant_profile.update!(school_cohort: target_school_cohort, schedule:)
+      rescue ActiveRecord::RecordInvalid
+        false
+      end
```
peteryates, line 90: Perhaps we should be logging an error here, although with the `#valid?` check preceding it in `#save` I'd expect not to reach here.

peteryates (approved):
This seems sound to me. @lazydevorg - is it doing all you'd expect from the process?

app/forms/schools/amend_participant_cohort_form.rb
ltello: You are right. I forgot to add the error to the form. 
Fixed

app/forms/schools/amend_participant_cohort_form.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    def participant_profile
+      return unless participant_identity
+
+      @participant_profile ||= ParticipantProfile::ECF.training_status_active
+                                                      .active_record
+                                                      .find_by(participant_identity:)
+    end
+
+    def persist
```
lazydevorg, line 89: this method persist changes in the DB while the service (below) uses the form to run the business logic.
Should the service run the logic and the form use the service instead?
It seems weird that a service is using a form.
Is the form used in any page?

ltello: I know what you mean. I wouldn't mind to change it but this is the rationale I built it this way:
To me Support people (via the Admin UI) should have an option on any participant to change it to a different cohort as they have to change a participant's name or email, etc. That is, individually at participant level. 
If we added it, that controller should consume this Induction::AmendParticipantCohortForm directly.

The service however, is something meant to replicate what we where doing manually at Zendesk (change a bunch of emails to a different cohort) and therefore it is consuming the form once per email.

lazydevorg: I get your point @ltello . The form will be needed in the admin and that's ok.
I think we should avoid that a service depends on a form.
I think it should be the opposite though.
The services we have all run business login called by controllers or forms.
Is there any issue in making the form call the service for the persist part?

ltello: Form renamed to Service

lazydevorg (approved):

PR #2452, opened 2022-09-07
Remove now deprecated shared context in favour of new test setup

cwrw:
### Context
Testing strategy recently changed removing large setup contexts.

- Ticket: N/A

Remove old setup as it no longer exists, in favour of new strategy.

### Guidance to review
V3 tests are ignored as they fail anyway, but get the test to the point it fails on the endpoints not existing and not the setup


StupidCodeFactory (approved):
Thanks! 👏

PR #2453, opened 2022-09-07
[CPDLP-1707] Can post 2022 declarations

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1707
- Also covers https://dfedigital.atlassian.net/browse/CPDLP-1644
- This enables providers to post 2022 declarations
- Able to import NPQ contracts from CSV

- When viewing statements, added cohort filter so finance users can switch between cohorts for statements
- Added new method `schedule_for` to be able to find schedule from a given profile
- Added class to be able to import `NPQContract`s from CSV
- Seeding process now automatically adds 2022 statements, ecf contracts and npq contracts
- Pruned `Finance::Statement.attach` in favour of calling `DeclarationStatementAttacher` directly
- `DeclarationStatementAttacher` now correctly determines the cohort and therefore attaches declarations to correct statements

### Guidance to review

- When viewing a statement these do not conform to the current designs. Extra work needs to be applied to add LHS filtering system
- When viewing a statement the filtering system at the moment does not use javascript. Therefore it is possible to select an invalid statement. Extra work needs to be applied so when cohort is changed this updates the dropdown for possible statements that are selectable

### DIFF
--- a/app/controllers/finance/payment_breakdowns_controller.rb
+++ b/app/controllers/finance/payment_breakdowns_controller.rb
@@ -46,17 +46,19 @@ module Finance
     end
 
     def choose_npq_statement
+      cohort = Cohort.find_by(start_year: params[:cohort_year])
       npq_lead_provider = NPQLeadProvider.find(params[:npq_lead_provider])
       statement_name = params[:statement].humanize.gsub("-", " ")
-      statement = npq_lead_provider.statements.find_by(name: statement_name)
+      statement = npq_lead_provider.statements.find_by(cohort:, name: statement_name)
 
       redirect_to finance_npq_lead_provider_statement_path(npq_lead_provider.id, statement)
     end
 
     def choose_ecf_statement
+      cohort = Cohort.find_by(start_year: params[:cohort_year])
       lead_provider = LeadProvider.find(params[:lead_provider])
       statement_name = params[:statement].humanize.gsub("-", " ")
-      statement = lead_provider.statements.find_by(name: statement_name)
+      statement = lead_provider.statements.find_by(cohort:, name: statement_name)
 
       redirect_to finance_ecf_payment_breakdown_statement_path(lead_provider.id, statement)
     end
--- a/db/seeds/npq_contracts/fake-2021.csv
+++ b/db/seeds/npq_contracts/fake-2021.csv
@@ -1,0 +1,78 @@
+provider_name,cohort_year,course_name,recruitment_target,per_participant,service_fee_installments
+Ambition Institute,2021,NPQ Leading Teaching ,1000,900,12
+Ambition Institute,2021,NPQ Leading Behaviour and Culture,1001,901,13
+Ambition Institute,2021,NPQ Leading Teacher Development,1002,902,14
+Ambition Institute,2021,NPQ for Senior Leadership,1003,903,15
+Ambition Institute,2021,NPQ for Headship,1004,904,16
+Ambition Institute,2021,NPQ for Executive Leadership,1005,905,17
+Ambition Institute,2021,NPQ Leading Literacy,1006,906,18
+Ambition Institute,2021,NPQ Early Years Leadership,1007,907,19
+Best Practice Network,2021,NPQ Leading Teaching ,1008,908,20
+Best Practice Network,2021,NPQ Leading Behaviour and Culture,1009,909,21
+Best Practice Network,2021,NPQ Leading Teacher Development,1010,910,22
+Best Practice Network,2021,NPQ for Senior Leadership,1011,911,23
+Best Practice Network,2021,NPQ for Headship,1012,912,24
+Best Practice Network,2021,NPQ for Executive Leadership,1013,913,25
+Church of England,2021,NPQ Leading Teaching ,1014,914,26
+Church of England,2021,NPQ Leading Behaviour and Culture,1015,915,27
+Church of England,2021,NPQ Leading Teacher Development,1016,916,28
+Church of England,2021,NPQ for Senior Leadership,1017,917,29
+Church of England,2021,NPQ for Headship,1018,918,30
+Church of England,2021,NPQ for Executive Leadership,1019,919,31
+Education Development Trust,2021,NPQ Leading Teaching ,1020,920,32
+Education Development Trust,2021,NPQ Leading Behaviour and Culture,1021,921,33
+Education Development Trust,2021,NPQ Leading Teacher Development,1022,922,34
+Education Development Trust,2021,NPQ for Senior Leadership,1023,923,35
+Education Development Trust,2021,NPQ for Headship,1024,924,36
+Education Development Trust,2021,NPQ for Executive Leadership,1025,925,37
+Education Development Trust,2021,NPQ Leading Literacy,1026,926,38
+Education Development Trust,2021,NPQ Early Years Leadership,1027,927,39
+School-Led Network,2021,NPQ Leading Teaching ,1028,928,40
+School-Led Network,2021,NPQ Leading Behaviour and Culture,1029,929,41
+School-Led Network,2021,NPQ Leading Teacher Development,1030,930,42
+School-Led Network,2021,NPQ for Senior Leadership,1031,931,43
+School-Led Network,2021,NPQ for Headship,1032,932,44
+School-Led Network,2021,NPQ for Executive Leadership,1033,933,45
+School-Led Network,2021,NPQ Leading Literacy,1034,934,46
+School-Led Network,2021,NPQ Early Years Leadership,1035,935,47
+Leadership Learning South East,2021,NPQ Leading Teaching ,1036,936,48
+Leadership Learning South East,2021,NPQ Leading Behaviour and Culture,1037,937,49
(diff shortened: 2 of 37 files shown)


spec/services/finance/npq/statement_calculator_spec.rb
```diff
@@ -191,4 +191,32 @@
...
     end
   end
+
+  context "when there exists contracts over multiple cohorts", with_feature_flags: { multiple_cohorts: "active" } do
+    let!(:cohort_2022) { Cohort.next || create(:cohort, :next) }
+    let!(:contract_2022) { create(:npq_contract, npq_lead_provider:, cohort: cohort_2022) }
+    let!(:statement_2022) { create(:npq_statement, cpd_lead_provider:, cohort: cohort_2022) }
+
+    before do
+      declarations = create_list(
```
cwrw, line 200: do we need `create_list` if its only one we need?

spec/components/finance/statements/npq_statement_selector_spec.rb
```diff
@@ -7,8 +7,9 @@
   let(:npq_lead_provider)    { cpd_lead_provider.npq_lead_provider }
   let!(:npq_statement)       { create(:npq_statement, cpd_lead_provider:) }
   let!(:other_npq_statement) { create(:npq_statement) }
+  let(:cohorts) { Cohort.where(start_year: 2021..) }
```
cwrw, line 10: Maybe we should add this as a scope on `Cohort` as seems like we're adhering to this rule in many places, what do you think?

spec/requests/finance/payment_breakdowns_spec.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
cwrw, line 3: maybe we can add feature tests as well 🤔 ?

db/seeds/test_data.rb
```diff
@@ -919,3 +919,8 @@ def create_participant_with_eligibility(urn, name, participant_class, options =
     count: 10,
   ]
 end
+
+service = Importers::NPQContracts.new(
```
cwrw, line 923: To fix the tests after we rolled into cohort 2022, we amended the cohorts to 2022 for the test data (see https://github.com/DFE-Digital/early-careers-framework/blob/develop/db/seeds/test_data.rb#L369). So I think we'll be generating a lot of those fake ones now, with the old and new fixtures. Maybe we can rip the old creation out completely from `test_data` and only use the new way to import contracts, or create a new test csv for 2021 as well to create both?

cwrw (commented):
some initial thoughts after a quick review, will do another/test once out of draft

spec/services/record_declaration_spec.rb
```diff
@@ -205,6 +205,26 @@
       it_behaves_like "validates the course_identifier, cpd_lead_provider, participant_id"
       it_behaves_like "validates existing declarations"
       it_behaves_like "validates the participant milestone"
+
+      context "for 2022 cohort", :with_default_schedules, with_feature_flags: { multiple_cohorts: "active" } do
+        let!(:schedule) { create(:ecf_schedule, cohort:) }
+        let!(:statement) { create(:ecf_statement, :output_fee, deadline_date: 6.weeks.from_now, cpd_lead_provider:, cohort:) }
+
+        let(:cohort) { Cohort.next || create(:cohort, :next) }
```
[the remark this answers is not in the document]
asmega: i feel like i'm kind of doing that already unless i'm missing something?
```
let(:participant_profile) { create(:ect, :eligible_for_funding, school_cohort:, lead_provider:) }
```
the above overrides the original `let`. this factory by default creates it own `school_cohort`, but instead we pass our own here. we could define a default for `school_cohort` in this spec and use that as the default instead if that is what you're suggesting? but has the downside of not using the default specified in the factory?

spec/services/finance/npq/statement_calculator_spec.rb
asmega: yep, this now uses `create`

db/seeds/test_data.rb
asmega: i've deleted all that code. it now imports 2021 npq contracts via csv import mechanism for which i've generated some fake data for

spec/components/finance/statements/npq_statement_selector_spec.rb
asmega: if we did not sure what we would name it though?

spec/services/record_declaration_spec.rb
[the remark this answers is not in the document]
cwrw: chatted in person (as was the latter and a suggestion)

app/controllers/finance/payment_breakdowns_controller.rb
```diff
@@ -46,17 +46,19 @@ def choose_provider_npq
...
 
       redirect_to finance_npq_lead_provider_statement_path(npq_lead_provider.id, statement)
     end
 
     def choose_ecf_statement
+      cohort = Cohort.find_by(start_year: params[:cohort_year])
       lead_provider = LeadProvider.find(params[:lead_provider])
       statement_name = params[:statement].humanize.gsub("-", " ")
-      statement = lead_provider.statements.find_by(name: statement_name)
+      statement = lead_provider.statements.find_by(cohort:, name: statement_name)
```
cwrw, line 61: whilst we don't want to implement any js to hide statements not available to lead providers, the current journey is a bit broken and returns a 500 if a statement is missing. Can we instead redirect and show a flash message saying statement not found if the statement is `nil` rather than getting a 500?

db/seeds/npq_contracts/fake-2021.csv
```diff
@@ -0,0 +1,78 @@
...
+Teacher Development Trust,2021,NPQ Early Years Leadership,1059,959,71
+UCL Institute of Education,2021,NPQ Leading Teaching ,1060,960,72
+UCL Institute of Education,2021,NPQ Leading Behaviour and Culture,1061,961,73
+UCL Institute of Education,2021,NPQ Leading Teacher Development,1062,962,74
+UCL Institute of Education,2021,NPQ for Senior Leadership,1063,963,75
+UCL Institute of Education,2021,NPQ for Headship,1064,964,76
+UCL Institute of Education,2021,NPQ for Executive Leadership,1065,965,77
+UCL Institute of Education,2021,NPQ Leading Literacy,1066,966,78
+UCL Institute of Education,2021,NPQ Early Years Leadership,1067,967,79
+Ambition Institute,2021,The Early Headship Coaching Offer,300,800,0
```
cwrw, line 70: Could you add the following course name to the npq courses locale as it's missing

spec/components/finance/statements/npq_statement_selector_spec.rb
cwrw: naming is hard 😅 , `active_start_years`? depends what we see 2020 as and what term we use to name that cohort

db/seeds/npq_contracts/fake-2021.csv
asmega: are you referring to these? https://github.com/DFE-Digital/early-careers-framework/blob/develop/config/locales/en.yml#L393 I don't see these been used anywhere and is probably dead code?

cwrw: Sorry meant https://github.com/DFE-Digital/early-careers-framework/blob/develop/config/locales/en.yml#L385 when I seeded my local env and tried accessing payment breakdowns for NPQ I got multiple missing translations, eg:`translation missing: en.npq_courses.npq-early-headship-coaching-offer`

app/controllers/finance/payment_breakdowns_controller.rb
asmega: we could do this but I feel we wouldn't get much value from doing this. if there will be some work to align to designs and add the JS this change would be made somewhat redundant

db/seeds/npq_contracts/fake-2021.csv
asmega: nice spot, i'll add those now

spec/components/finance/statements/npq_statement_selector_spec.rb
asmega: i guess it would be something like `inauguration_onwards` or `inception_onwards`?

app/controllers/finance/payment_breakdowns_controller.rb
cwrw: I found it really hard to navigate/test and kept hitting an error locally, I wonder if others will hit the same on other test envs when deployed. The fix is pretty small with if/else and showing a flash message. We can implement if we get feedback regarding that I guess and for now can keep as is

spec/components/finance/statements/npq_statement_selector_spec.rb
cwrw: Maybe, let's do this later when we revisit cohorts further and agree a name with the team

cwrw (approved):

cwrw:
Since we cannot commit the real statement csv and I don't have visibility of what new course names we have, can we make sure all locales are set for those as well, thanks!

PR #2461, opened 2022-09-09
[CPDLP-1669] Cache delivery partner on declaration

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1669
- Providers would like to be able to see delivery partners on declarations. So they are able to work out who gets paid what

- When posting a declaration, cache the delivery partner on the declaration
- There will be future changes to backfill this data, expose via the api and allow support to make amendments

### Guidance to review

- none


db/migrate/20220908105721_add_dp_to_declarations.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddDpToDeclarations < ActiveRecord::Migration[6.1]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :participant_declarations, :delivery_partner, index: { algorithm: :concurrently }
```
cwrw, line 7: Might need to also add a ` null: true`?

app/models/participant_declaration.rb
```diff
@@ -7,6 +7,7 @@ class ParticipantDeclaration < ApplicationRecord
   belongs_to :user
   belongs_to :participant_profile
   belongs_to :superseded_by, class_name: "ParticipantDeclaration", optional: true
+  belongs_to :delivery_partner, optional: true
```
cwrw, line 10: Do we need the opposite relationship on `delivery_partner` for later?

cwrw (approved):
Just a question for me around deployments in this app, we generally can include code/migrations in the same PR and there won't be any issues?

db/migrate/20220908105721_add_dp_to_declarations.rb
asmega: apparently it already defaults to true. see https://api.rubyonrails.org/v6.1.6.1/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference

app/models/participant_declaration.rb
asmega: Maybe. I don't see that happening for some time yet though. At the moment I don't see the use case for a delivery partner to see all their declarations.

asmega:
> Just a question for me around deployments in this app, we generally can include code/migrations in the same PR and there won't be any issues?

I've not faced this issue before. I can understand in some cases where yes you may want to split out because things need to happen in a certain sequence. My understanding here is that the migration will happen and complete first before the app kicks in. And the app is able to boot in order to run the migration in the first place. So I don't think there will be a problem here. Review app was also fine, but if something happens on staging at least we will know that something isn't quite right. I'm fairly confident that won't happen though

PR #2474, opened 2022-09-13
[CPDLP-1539] - expose non uk countries to providers on the participant response across all api versions

StupidCodeFactory:
### Context

- [CPDLP-1539](https://dfedigital.atlassian.net/browse/CPDLP-1539): 

Added the country ISO 3666 alpha3 code to `npq_applications` as `teacher_catchment_iso_country_code` and added it to the API alongside the `teacher_catchment` and `teacher_catchment_country`

### Guidance to review
Screenshot 2022-09-23 at 17 13 30
Screenshot 2022-09-23 at 17 13 44


app/services/npq/build_application.rb
```diff
@@ -39,9 +39,21 @@ def npq_application_attributes
...
       )
     end
 
+    def teacher_catchment_country
+      npq_application_params[:teacher_catchment_country]
+    end
+
+    def teacher_catchment_iso_country_code
+      return if teacher_catchment_country.blank?
+      return unless (country = ISO3166::Country.find_country_by_iso_short_name(teacher_catchment_country))
```
peteryates, line 52: Is it worth logging to sentry if we search for a code but don't find one?

lib/tasks/one_offs.rake
```diff
@@ -1,11 +1,11 @@
...
-  desc "Extend ECF Standard April started milestone date"
-  task extend_ecf_standard_april_started_milestone_date: :environment do
-    schedule = Finance::Schedule.find_by(name: "ECF Standard April")
-    schedule.schedule_milestones.find_by(name: "Output 1.1 - Participant Start").destroy!
-    Finance::Milestone.find_by(name: "Output 1.1 - Participant Start").destroy!
-    schedule.milestones.find_by(declaration_type: "started").update!(milestone_date: Date.new(2022, 5, 31))
+  desc "backfill teacher catchement iso country code"
+  task backfill_teacher_catchment_iso_country_code: :environment do
+    NPQApplication.where.not(teacher_catchment_country: nil).find_each do |npq_application|
+      country = ISO3166::Country.find_country_by_iso_short_name(npq_application.teacher_catchment_country)
```
peteryates, line 7: Again it would be interesting to know which of these don't match, but that's easy enough to do afterwards. There are some funny ones in the gem that might not cleanly match, like `Bolivia (Plurinational State of)`.

peteryates (approved):
Looks good to me overall. Left a really minor suggestion.

PR #2480, opened 2022-09-14
Duplicate profiles

asmega:
### Context

- Ticket:


spec/requests/api/v2/ecf_participants_spec.rb
```diff
@@ -267,19 +285,13 @@
...
-                  .user
+          ect = ect_profile.user
           ect_row = parsed_response.find { |row| row["id"] == ect.id }
+
           expect(ect_row).not_to be_nil
           expect(ect_row["email"]).to eql ect.email
           expect(ect_row["full_name"]).to eql ect.full_name
-
-          expect(ect_row["mentor_id"]).to be_blank
+          expect(ect_row["mentor_id"]).to eql(mentor.id)
```
cwrw, line 294: 🌟

app/services/api/v1/ecf_participants/index.rb
```diff
@@ -4,53 +4,60 @@ module Api
...
-                      :preferred_identity,
-                      :schedule,
-                      induction_programme: { school_cohort: %i[school cohort] },
-                      mentor_profile: [:participant_identity],
-                      participant_profile: %i[ecf_participant_eligibility ecf_participant_validation_data participant_identity teacher_profile user],
-                    )
-                    .joins("JOIN (#{join.to_sql}) AS latest_induction_records ON latest_induction_records.id = induction_records.id AND latest_induction_records.created_at_precedence = 1")
+            .where(id: induction_record_ids_with_deduped_profiles)
+            .joins(schedule: [:cohort], participant_profile: { school_cohort: [:cohort] })
+            .where(schedule: { cohort: with_cohorts })
```
cwrw, line 18: out of curiosity when would the `schedule.cohort` be different than `participant_profile . school_cohorts .cohort`?

cwrw (commented):
Looks good 👍 can you re-enable this test https://github.com/DFE-Digital/early-careers-framework/blob/develop/spec/requests/api/v1/test_ecf_participants_spec.rb#L138?

asmega:
> Looks good 👍 can you re-enable this test https://github.com/DFE-Digital/early-careers-framework/blob/develop/spec/requests/api/v1/test_ecf_participants_spec.rb#L138?

Good spot, I've just re-enabled that. Seems to pass verbatim

cwrw (approved):

app/services/api/v1/ecf_participants/index.rb
StupidCodeFactory: The correct way to look the cohort `induction_records.schedule.cohort`. We've seen last week that there are inconsistencies between `participant_profile#schedule` and `induction_record#schedule`

asmega: That's what you think to be correct, but you're only looking at part of the picture. On the schools side of things they don't care about schedules and hardly look at them. Yet somehow what are able to determine the cohort of a participant. Therefore is `induction_records.schedule.cohort` the correct way to determine's someone cohort?

With regards to `participant_profile#schedule` that is not present in the above snippet therefore I believe has no bearing on this part of the conversation.

asmega: I've removed the joining and where on `participant_profile` for cohort. This only goes down the schedule part here. elsewhere in this class we also use the cohort from the induction programme. so two routes are used but this one has been removed

cwrw (approved):

PR #2484, opened 2022-09-15
[CPDLP-1770] Remove different declaration date validation

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1770
- There is a regression whereby providers cannot re-post for a declaration after it has been clawed back. ie the original declaration is in `awaiting_clawback` or `clawed_back` state

- This removes the validation `record_does_not_exists_with_different_declaration_date`
- I'm not sure why we have this validation and what purposes it serves so I have removed it as it blocking providers from reposting
- There already exists the validation `validates_billable_slot_available` which checks there is open slot available otherwise prevents the declaration from getting through
- This should be enough for our needs to my understanding

### Guidance to review

- none

### DIFF
--- a/spec/services/record_declaration_spec.rb
+++ b/spec/services/record_declaration_spec.rb
@@ -68,7 +68,7 @@ RSpec.shared_examples "validates the course_identifier, cpd_lead_provider, parti
 end
 
 RSpec.shared_examples "validates existing declarations" do
-  context "when an exisiting declaration already exists" do
+  context "when an existing declaration already exists" do
     let!(:existing_declaration) do
       described_class.new(params).call
     end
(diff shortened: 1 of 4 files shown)


spec/services/record_declaration_spec.rb
```diff
@@ -293,6 +293,36 @@
...
+
+      statement = participant_declaration.statements[0]
+
+      service = ParticipantDeclarations::MarkAsPayable.new(statement)
+      service.call(participant_declaration)
+
+      service = ParticipantDeclarations::MarkAsPaid.new(statement)
+      service.call(participant_declaration)
+
+      Finance::ClawbackDeclaration.new(participant_declaration).call
```
cwrw, line 318: could we use the `clawed_back` trait for declarations instead of setting this up manually?

cwrw (approved):

spec/services/record_declaration_spec.rb
asmega: @StupidCodeFactory do you have a preference here of going through the service classes or setting up a record manually?

PR #2485, opened 2022-09-15
[CPDLP-1772] Add new service to transfer an NPQ participant between lead providers

cwrw:
### Context
A task to move an NPQ participant between lead providers prompted creating a service to wrap the logic and test it 
- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1772

Using the external identifier of the participant and the ID of the new NPQ lead provider, the participant's application should move to the new NPQ lead provider. The new provider can now make declarations to this participant

### Guidance to review
I think(?) this isn't the right place to add this service, as this hierarchy is used differently for inheritance purposes. Happy to move to `services/npq` or as a standalone service class, any thoughts welcome

### DIFF
--- a/app/services/transfers/npq_participants.rb
+++ b/app/services/transfers/npq_participants.rb
@@ -26,7 +26,7 @@ module Transfers
     end
 
     def npq_lead_provider
-      @npq_lead_provider ||= NPQLeadProvider.find_by(id: new_npq_lead_provider_id)
+      @npq_lead_provider ||= NPQLeadProvider.find_by!(id: new_npq_lead_provider_id)
     end
 
     def participant_profile
@@ -34,11 +34,11 @@ module Transfers
         ParticipantProfile::NPQ
           .active_record
           .joins(participant_identity: { npq_applications: :npq_course })
-          .find_by(participant_identity:, npq_courses: { identifier: course_identifier })
+          .find_by!(participant_identity:, npq_courses: { identifier: course_identifier })
     end
 
     def participant_identity
-      @participant_identity ||= ParticipantIdentity.find_by(external_identifier:)
+      @participant_identity ||= ParticipantIdentity.find_by!(external_identifier:)
     end
 
     def npq_application
--- a/app/services/transfers/npq_participants.rb
+++ b/app/services/transfers/npq_participants.rb
@@ -2,9 +2,12 @@
 
 module Transfers
   class NPQParticipants
-    def initialize(external_identifier:, new_npq_lead_provider_id:)
+    attr_reader :external_identifier, :new_npq_lead_provider_id, :course_identifier
+
+    def initialize(external_identifier:, new_npq_lead_provider_id:, course_identifier:)
       @external_identifier = external_identifier
       @new_npq_lead_provider_id = new_npq_lead_provider_id
+      @course_identifier = course_identifier
     end
 
     def call
@@ -14,8 +17,6 @@ module Transfers
     end
 
   private
-
-    attr_reader :external_identifier, :new_npq_lead_provider_id
 
     def transfer_participant
       return unless participant_profile && npq_application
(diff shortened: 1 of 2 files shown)


app/services/participants/transfer/npq.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module Participants
```
cwrw, line 3: I think this should probably be under `services` directly (see review guidance), thoughts?

app/services/participants/transfer/npq.rb
```diff
@@ -0,0 +1,50 @@
...
+        @npq_lead_provider ||= NPQLeadProvider.find_by(id: new_npq_lead_provider_id)
+      end
+
+      def participant_profile
+        @participant_profile ||=
+          ParticipantProfile::NPQ
+            .where(participant_identity:)
+            .npqs
+            .active_record
+            .first
```
peteryates, line 38: Worth using a `find_by` here instead of the `where`/`first` combo?

```suggestion
          ParticipantProfile::NPQ
            .npqs
            .active_record
            .find_by(participant_identity:)
```

peteryates:
I'd agree that this doesn't feel like the right spot for it. I think being in the `app/services` would make sense, but I get the impression there could be a several more of these `Transfer`-type services added so perhaps it's worth creating `app/services/transfers`?

app/services/participants/transfer/npq.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      def npq_lead_provider
+        @npq_lead_provider ||= NPQLeadProvider.find_by(id: new_npq_lead_provider_id)
+      end
+
+      def participant_profile
+        @participant_profile ||=
+          ParticipantProfile::NPQ
+            .npqs
```
asmega, line 35: i think `ParticipantProfile::NPQ` should cover the scope of `npqs`?

app/services/participants/transfer/npq.rb
```diff
@@ -0,0 +1,49 @@
...
+            .active_record
+            .find_by(participant_identity:)
```
asmega, line 37: for this particular query we may want to pass the course/course_identifier as it is possible for someone to be doing two NPQs at the same time. We want to be able to discern between which one we want to transfer

app/services/participants/transfer/npq.rb
StupidCodeFactory: I'd put it under `NPQ` namespace

app/services/participants/transfer/npq.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      def transfer_participant
+        return unless participant_profile && npq_application
+
+        npq_application.update!(npq_lead_provider:)
+        Rails.logger.info "Participant (#{external_identifier}) was transferred to (#{npq_lead_provider.name}) successfully"
+      end
+
+      def npq_lead_provider
+        @npq_lead_provider ||= NPQLeadProvider.find_by(id: new_npq_lead_provider_id)
```
StupidCodeFactory, line 29: I'd use a find_by! or add some validation otherwise we'd never never now the participant has not been transferred, unless you stare at the logs and look for that one line in the logs, which in prod might a challenging task.
```suggestion
        @npq_lead_provider ||= NPQLeadProvider.find_by!(id: new_npq_lead_provider_id)
```

app/services/participants/transfer/npq.rb
StupidCodeFactory, line 37: Same suggestion as my previous comment:
```suggestion
            .find_by!(participant_identity:)
```

app/services/participants/transfer/npq.rb
cwrw: I've moved to under `transfers/` as per Peter's comment as we might have other transfers to add

StupidCodeFactory (approved):

PR #2505, opened 2022-09-21
[CPDLP-1578] Access management

mooktakim:
### Context

- Ticket:
  - https://dfedigital.atlassian.net/browse/CPDLP-1578
  - https://dfedigital.atlassian.net/browse/CPDLP-1572
  - https://dfedigital.atlassian.net/browse/CPDLP-1767
  - https://dfedigital.atlassian.net/browse/CPDLP-1571

* Access management pages
* Appropriate body role added
* Added seed to create appropriate-body@example.com user for AB

### DIFF
--- a/app/controllers/appropriate_bodies/appropriate_bodies_controller.rb
+++ b/app/controllers/appropriate_bodies/appropriate_bodies_controller.rb
@@ -8,7 +8,7 @@ module AppropriateBodies
       @choose_organisation_form = ChooseOrganisationForm.new
       @choose_organisation_form.user = current_user
 
-      if @choose_organisation_form.only_one?
+      if @choose_organisation_form.only_one
         redirect_to appropriate_body_participants_path(@choose_organisation_form.appropriate_body)
       end
     end
--- a/app/controllers/appropriate_bodies/appropriate_bodies_controller.rb
+++ b/app/controllers/appropriate_bodies/appropriate_bodies_controller.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module AppropriateBodies
+  class AppropriateBodiesController < BaseController
+    layout "application"
+
+    def index
+      @choose_organisation_form = ChooseOrganisationForm.new
+      @choose_organisation_form.user = current_user
+
+      if @choose_organisation_form.only_one?
+        redirect_to appropriate_body_participants_path(@choose_organisation_form.appropriate_body)
+      end
+    end
+
+    def create
+      @choose_organisation_form = ChooseOrganisationForm.new(choose_organisation_form_params)
+      @choose_organisation_form.user = current_user
+
+      if @choose_organisation_form.valid?
+        redirect_to appropriate_body_participants_path(@choose_organisation_form.appropriate_body)
+      else
+        render :index
+      end
+    end
+
+  private
+
+    def choose_organisation_form_params
+      params.fetch(:appropriate_bodies_choose_organisation_form, {}).permit(
+        :appropriate_body_id,
+      )
+    end
+  end
+end
--- a/app/controllers/appropriate_bodies/base_controller.rb
+++ b/app/controllers/appropriate_bodies/base_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module AppropriateBodies
+  class BaseController < ApplicationController
+    include Pundit::Authorization
+
+    before_action :authenticate_user!
+    before_action :ensure_appropriate_body
+    before_action :set_appropriate_body
+
+    layout "appropriate_bodies"
+
+  private
+
+    def ensure_appropriate_body
+      raise Pundit::NotAuthorizedError, I18n.t(:forbidden) unless current_user.appropriate_body?
+    end
+
+    def set_appropriate_body
+      return if params[:appropriate_body_id].blank?
+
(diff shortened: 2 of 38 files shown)


app/controllers/appropriate_bodies/appropriate_bodies_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      if @choose_organisation_form.only_one?
+        redirect_to appropriate_body_participants_path(@choose_organisation_form.appropriate_body)
+      end
+    end
+
+    def create
+      @choose_organisation_form = ChooseOrganisationForm.new(choose_organisation_form_params)
+      @choose_organisation_form.user = current_user
+
+      if @choose_organisation_form.valid?
```
peteryates, line 20: Should it be saving here in addition to checking it's valid?

mooktakim: @peteryates we don't need to save here, we only want to redirect to the correct organisation (appropriate body).

peteryates: Ah - fair enough, thanks @mooktakim. It being in `#create` threw me 😅

cwrw (commented):
looking good, just some minor suggestions/comments

app/forms/appropriate_bodies/choose_organisation_form.rb
```diff
@@ -0,0 +1,38 @@
...
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :user
+    attribute :appropriate_body_id
+
+    validates :appropriate_body_id, inclusion: { in: :valid_ids }
+
+    def appropriate_body_options
+      @appropriate_body_options ||= user.appropriate_bodies.each_with_object({}) do |ab, sum|
```
cwrw, line 14: maybe we can name this something different than sum?

spec/features/appropriate_bodies/appropriate_bodies_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
cwrw, line 3: do we want to add a similar feature test for choosing roles, or is this covered by other specific feature tests?

app/forms/choose_role_form.rb
```diff
@@ -0,0 +1,68 @@
...
+      helpers.participant_start_path(user)
+    when "no_role"
+      helpers.dashboard_path
+    end
+  end
+
+  def only_one_role?
+    return false unless user.user_roles.count == 1
+
+    self.role = user.user_roles.first
```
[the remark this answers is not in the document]
mooktakim: @cwrw 
In this case when only one role, we want to set the role so we can then redirect to it in the controller, `choose_role_form.redirect_path` will return the correct path.

app/models/appropriate_body_profile.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class AppropriateBodyProfile < ApplicationRecord
+  has_paper_trail
+
+  belongs_to :user
+  belongs_to :appropriate_body
+
+  def self.create_appropriate_body_user(full_name, email, appropriate_body)
```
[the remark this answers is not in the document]
mooktakim: We would use this in the support admin pages when we want to create/edit a AB users
also likely to be used for the initial batch of users that would be created.

app/controllers/appropriate_bodies/base_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    layout "appropriate_bodies"
+
+  private
+
+    def ensure_appropriate_body
+      raise Pundit::NotAuthorizedError, I18n.t(:forbidden) unless current_user.appropriate_body?
+    end
+
+    def set_appropriate_body
```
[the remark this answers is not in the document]
mooktakim: @cwrw 
this is for the other controllers that'll be inheriting from `BaseController`, eg `ParticipantsController`, (which at the moment is blank, there's another ticket for that)

app/controllers/appropriate_bodies/appropriate_bodies_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module AppropriateBodies
+  class AppropriateBodiesController < BaseController
+    layout "application"
+
+    def index
+      @choose_organisation_form = ChooseOrganisationForm.new
+      @choose_organisation_form.user = current_user
```
cwrw, line 9: could we set user through the initialiser as well?

app/controllers/choose_roles_controller.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class ChooseRolesController < ApplicationController
+  before_action :authenticate_user!
+  before_action :set_choose_role_form, only: %i[show create]
+
+  def show
+    if @choose_role_form.only_one_role || @choose_role_form.has_no_role
+      redirect_to @choose_role_form.redirect_path(helpers:)
```
cwrw, line 9: something maybe in the next AB prs: we need to watch out for how many redirects we're doing to get the single AB role page, and to avoid too many

cwrw (approved):
👍

PR #2507, opened 2022-09-21
[CPDLP-1564] - Assurance report download links

StupidCodeFactory:
### Context
For easy access by SRM to assurance reports a link has been added to download the report in CSV format.
- [Ticket: ](https://dfedigital.atlassian.net/browse/CPDLP-1564)


peteryates (commented):
The idea of using views does worry me a bit here as when we need to make changes to the database, PostgreSQL will prevent it with `cannot drop column abc of table xyz because other objects depend on it`.

Ordinarily, on stable schemas, it's something I'd be eager to use but I think we might be better served by query objects here for the time being, while we're likely to be moving stuff around.

Eager to hear other opinions from the team.

StupidCodeFactory:
> The idea of using views does worry me a bit here as when we need to make changes to the database, PostgreSQL will prevent it with `cannot drop column abc of table xyz because other objects depend on it`.
> 
> Ordinarily, on stable schemas, it's something I'd be eager to use but I think we might be better served by query objects here for the time being, while we're likely to be moving stuff around.
> 
> Eager to hear other opinions from the team.

I'm happy to move this to ParticipantDeclaration::(ECF|NPQ).find_by_sql(). One would both way have to fix the query, either before dropping the column or after. I like views for this kind of complicated queries, but I don't feel like `cannot drop column abc of table xyz because other objects depend on it` is sufficient enough constraint to not use views as we've already got foreign key constraints that would produce similar error.
I was actually hoping we'd use views for payment breakdowns rather than having a tone of complicated ruby code when we could achieve it with a single SQL query.

cwrw:
I agree with postponing the use of views for now. The state of our model and data isn't stable yet, and is shifting quite a bit, where just within a month we've had to rethink the way we query things many times in the code. 

Moving this logic and it's queries away from tested code will add hidden complexity for us, and I suspect we'll be amending this quite a bit as well for the short term while we discover more anomalies. 

However when we do get to a more stable place with the data, I agree that it would be great for us to move to using views then.

app/services/finance/ecf/assurance_report/query.rb
```diff
@@ -0,0 +1,89 @@
...
+        def participant_declarations
+          ParticipantDeclaration::ECF.find_by_sql(sql)
+        end
+
+      private
+
+        attr_accessor :statement
+
+        def sql
+          <<~EOSQL
```
cwrw, line 20: could we add a test for this query service, as we're setting values through `case`?

app/services/finance/npq/assurance_report/query.rb
```diff
@@ -0,0 +1,70 @@
...
+        def participant_declarations
+          @participant_declarations ||= ParticipantDeclaration::NPQ.find_by_sql(sql)
+        end
+
+      private
+
+        attr_accessor :statement
+
+        def sql
+          <<~EOSQL
```
cwrw, line 20: same here, can we add a test for this service?

app/services/finance/ecf/assurance_report/presenter.rb
```diff
@@ -0,0 +1,74 @@
...
+          self.statement_id = statement_id
+          self.query        = query
+        end
+
+        def to_csv
+          participant_declarations
+        end
+
+        def participant_declarations
+          @participant_declarations ||= query.participant_declarations.map do |participant_declaration|
```
cwrw, line 17: I think there might be performance implications if we were to map through a large number of participant declarations and hold in memory, maybe we should invoke things lazily?

app/services/finance/ecf/assurance_report/query.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+module Finance
+  module ECF
+    class AssuranceReport
+      class Query
+        def initialize(statement)
```
cwrw, line 7: Out of curiosity did you run those queries in prod, just curious about it's performance?

app/services/finance/ecf/assurance_report/presenter.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+module Finance
+  module ECF
+    class AssuranceReport
+      class Presenter
```
cwrw, line 6: Suggestion: I would prefer if we had a CSV serialiser, for instance `AssuranceRerpotCsvSerializer` which accepts the scope from the query and generates the CSV headers and rows (without the need to include the lib version). This means all the CSV logic lives in that serialiser, and we follow the established pattern we have for other existing CSV serialisers in the API, what do you think?

app/services/finance/ecf/assurance_report/query.rb
StupidCodeFactory: yes I do run those every output fee statement. Pretty fast.

app/services/finance/npq/assurance_report/query.rb
StupidCodeFactory: Fair point, yes the `CASE WHEN` could certainly do with a spec.

cwrw (approved):
Tested in prod with the worst statement and all good, thanks for addressing all comments, looks great 👍

PR #2521, opened 2022-09-26
[CPDLP-1613] ECF statements: uplift fees in own separate table

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1613
- Uplift fees are not considered adjustments

- Uplift fees has it's own row in the summary breakdown
- Uplift fees no longer in the adjustments table and now lives in it's own table

### Guidance to review

- none

### DIFF
--- a/app/services/finance/ecf/statement_calculator.rb
+++ b/app/services/finance/ecf/statement_calculator.rb
@@ -132,7 +132,7 @@ module Finance
       end
 
       def adjustments_total
-        total_for_uplift - clawback_deductions
+        -clawback_deductions
       end
 
       def clawback_deductions
@@ -142,7 +142,7 @@ module Finance
       end
 
       def total(with_vat: false)
-        sum = service_fee + output_fee + adjustments_total + statement.reconcile_amount
+        sum = service_fee + output_fee + total_for_uplift + adjustments_total + statement.reconcile_amount
         sum += vat if with_vat
         sum
       end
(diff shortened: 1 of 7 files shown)


app/services/finance/ecf/statement_calculator.rb
```diff
@@ -132,7 +132,7 @@ def total_for_uplift
       end
 
       def adjustments_total
-        total_for_uplift - clawback_deductions
+        -clawback_deductions
```
cwrw, line 135: Adjustments will now appear as negative in the summary table, do we want that value to change to that?

spec/support/features/pages/finance/finance_payment_breakdown_report.rb
```diff
@@ -51,7 +53,7 @@ def has_retained_1_declaration_payment_table?(num_ects: 0, num_mentors: 0, num_d
     end
 
     def has_other_fees_table?(num_ects: 0, num_mentors: 0)
-      adjustments_panel.has_uplift_payments? num_ects + num_mentors
```
cwrw, line 54: could we add a step for the adjustments panel now we switched this up to uplifts?

app/services/finance/ecf/statement_calculator.rb
asmega: if its negative zero this now shows it as unsigned zero with the new commit on the end

PR #2527, opened 2022-09-27
Fix bug that excluded data on final day of export

Lockyy:
### Context

The NPQ Application Exports allowed you to set the end date for exports. The export was wrongly ending at the start of that day rather than the end, entirely missing out a day of data.

Set the end date to midnight of the end_date.


peteryates (approved):
Looks good.

PR #2528, opened 2022-09-27
[CPDLP-1580] Access management - Switch between deliver partners

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1580

* DP participants page updated to allow for multiple organisations and switching between them
* DP choose organisation page
* Login as `delivery-partner@example.com`
  * If the user only has one DP partner role, it will redirect to the participants
  * If more than one, will redirect to choose organisation page

### DIFF
--- a/app/controllers/admin/delivery_partners/users_controller.rb
+++ b/app/controllers/admin/delivery_partners/users_controller.rb
@@ -5,11 +5,11 @@ module Admin
     class UsersController < Admin::BaseController
       skip_after_action :verify_authorized
       skip_after_action :verify_policy_scoped
-      before_action :set_user, only: %i[edit update delete destroy]
+      before_action :set_delivery_partner_profile_and_authorize, only: %i[edit update delete destroy]
 
       def index
-        sorted_users = User.delivery_partner_users.name_order
-        @pagy, @users = pagy(sorted_users, page: params[:page], items: 20)
+        scoped = DeliveryPartnerProfile.includes(:user).order("users.full_name asc")
+        @pagy, @delivery_partner_profiles = pagy(scoped, page: params[:page], items: 20)
       end
 
       def new
@@ -28,11 +28,11 @@ module Admin
       end
 
       def edit
-        @delivery_partners_user_form = ::DeliveryPartners::UpdateUserForm.new(@user)
+        @delivery_partners_user_form = ::DeliveryPartners::UpdateUserForm.new(delivery_partner_profile: @delivery_partner_profile)
       end
 
       def update
-        @delivery_partners_user_form = ::DeliveryPartners::UpdateUserForm.new(@user)
+        @delivery_partners_user_form = ::DeliveryPartners::UpdateUserForm.new(delivery_partner_profile: @delivery_partner_profile)
 
         if @delivery_partners_user_form.update(permitted_params)
           set_success_message(content: "Changes saved successfully.", title: "Success")
@@ -45,7 +45,7 @@ module Admin
       def delete; end
 
       def destroy
-        @user.delivery_partner_profile.destroy!
+        @delivery_partner_profile.destroy!
         set_success_message(content: "Delivery partner user deleted.", title: "Success")
         redirect_to admin_delivery_partners_users_path
       end
@@ -60,9 +60,9 @@ module Admin
         )
       end
 
-      def set_user
-        @user = User.delivery_partner_users.find(params[:id])
-        authorize @user
+      def set_delivery_partner_profile_and_authorize
+        @delivery_partner_profile = DeliveryPartnerProfile.find(params[:id])
+        authorize @delivery_partner_profile.user
       end
     end
   end
--- a/app/controllers/appropriate_bodies/participants_controller.rb
+++ b/app/controllers/appropriate_bodies/participants_controller.rb
@@ -2,6 +2,14 @@
 
 module AppropriateBodies
   class ParticipantsController < BaseController
-    def index; end
+    def index
+      appropriate_body
+    end
+
+  private
+
+    def appropriate_body
+      @appropriate_body ||= current_user.appropriate_bodies.find(params[:appropriate_body_id])
+    end
   end
 end
--- a/app/forms/delivery_partners/choose_organisation_form.rb
+++ b/app/forms/delivery_partners/choose_organisation_form.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module DeliveryPartners
+  class ChooseOrganisationForm
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :user
+    attribute :delivery_partner_id
+
+    validates :delivery_partner_id, inclusion: { in: :valid_ids }
+
+    def delivery_partner_options
+      @delivery_partner_options ||= user.delivery_partners.each_with_object({}) do |ab, sum|
+        sum[ab.id] = ab.name
+      end
+    end
+
+    def delivery_partner
+      return if delivery_partner_id.blank?
+
+      user.delivery_partners.find(delivery_partner_id)
+    end
+
+    def only_one
+      return false if user.delivery_partners.count > 1
+
+      self.delivery_partner_id = user.delivery_partners.first.id
+      true
+    end
+
+  private
+
+    def valid_ids
+      delivery_partner_options.keys
+    end
+  end
+end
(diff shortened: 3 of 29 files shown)


app/controllers/appropriate_bodies/participants_controller.rb
```diff
@@ -2,6 +2,16 @@
...
+      appropriate_body
+    end
+
+  private
+
+    def appropriate_body
+      @appropriate_body ||= current_user.appropriate_bodies.find(params[:appropriate_body_id])
+    end
+
+    helper_method :appropriate_body
```
asmega, line 15: how come a change in approach here? I personally prefer this method over `before_action` as they are lazy

app/forms/delivery_partners/choose_organisation_form.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+    end
+
+    def delivery_partner
+      return if delivery_partner_id.blank?
+
+      user.delivery_partners.find(delivery_partner_id)
+    end
+
+    def only_one
```
asmega, line 25: predicate?

app/views/admin/delivery_partners/users/edit.html.erb
```diff
@@ -8,13 +8,13 @@
...
 
-    <%= form_with model: @delivery_partners_user_form, scope: :delivery_partners_user_form, url: admin_delivery_partners_user_path(@user) do |f| %>
+    <%= form_with model: @delivery_partners_user_form, scope: :delivery_partners_user_form, url: admin_delivery_partners_user_path(@delivery_partner_profile) do |f| %>
       <%= f.govuk_collection_select(:delivery_partner_id, DeliveryPartner.name_order, :id, :name, label: { text: "Delivery partner" }, options: { include_blank: true }, class: ["js-location-select"]) %>
       <%= f.govuk_text_field(:full_name, label: { text: "Full name" }) %>
       <%= f.govuk_email_field(:email, label: { text: "Email" }) %>
       <div class="govuk-button-group">
         <%= f.govuk_submit "Save" %>
-        <%= govuk_button_link_to "Delete", delete_admin_delivery_partners_user_path(@user), class: "govuk-button--warning", "data-test"=> "delete-button" %>
+        <%= govuk_button_link_to "Delete", delete_admin_delivery_partners_user_path(@delivery_partner_profile), class: "govuk-button--warning", "data-test"=> "delete-button" %>
```
asmega, line 17: RE `data-test`,  personally not a fan of adding cruft to the dom. and pretty sure it can be targeted with a decent selector. is this being used?

app/forms/delivery_partners/update_user_form.rb
```diff
@@ -3,47 +3,46 @@
...
-      @email = user.email
-      @delivery_partner_id = user.delivery_partner_profile&.delivery_partner_id
-    end
+    delegate :user, to: :delivery_partner_profile
 
-    def delivery_partner
-      @delivery_partner ||= DeliveryPartner.find(delivery_partner_id)
+    def initialize(*args)
+      super
+      self.full_name = user.full_name
```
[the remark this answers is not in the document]
mooktakim: they won't be preloaded from the user model. Only at attribute assignment.
In this case, we want to preload the existing name, email and DP so the edit form can display it.

app/controllers/appropriate_bodies/participants_controller.rb
mooktakim: @asmega 
I guess this is more to stay consistent with what we do in other controllers, we don't use `before_action` much.

app/views/admin/delivery_partners/users/edit.html.erb
mooktakim: @asmega 
I don't think we need `data-test` here. I'll remove it

app/forms/delivery_partners/choose_organisation_form.rb
mooktakim: @asmega 
As we're setting the `delivery_partner_id` here, I've removed the `?`

app/controllers/appropriate_bodies/appropriate_bodies_controller.rb
```diff
@@ -26,6 +24,10 @@ def create
 
   private
 
+    def set_choose_organisation_form
```
cwrw, line 27: instead of using `before_action`, can we try set the attribute memoized?
```
def choose_organisation_form
    @choose_organisation_form ||= ChooseOrganisationForm.new(user: current_user)
end
```

app/models/delivery_partner_profile.rb
```diff
@@ -6,6 +6,8 @@ class DeliveryPartnerProfile < ApplicationRecord
   belongs_to :user
   belongs_to :delivery_partner
 
+  scope :with_users, -> { includes(:user) }
```
cwrw, line 9: I wonder if we do need this scope, as we're using it only once 🤔

mooktakim: @cwrw 
sure, I'll get rid of it

app/controllers/admin/delivery_partners/users_controller.rb
```diff
@@ -5,11 +5,10 @@ module DeliveryPartners
     class UsersController < Admin::BaseController
       skip_after_action :verify_authorized
       skip_after_action :verify_policy_scoped
-      before_action :set_user, only: %i[edit update delete destroy]
+      before_action :set_delivery_partner_profile_and_authorize, only: %i[edit update delete destroy]
 
       def index
-        sorted_users = User.delivery_partner_users.name_order
-        @pagy, @users = pagy(sorted_users, page: params[:page], items: 20)
+        @pagy, @delivery_partner_profiles = pagy(DeliveryPartnerProfile.includes(:user), page: params[:page], items: 20)
```
cwrw, line 11: could we sort by name here as well?

app/forms/delivery_partners/choose_organisation_form.rb
```diff
@@ -0,0 +1,38 @@
...
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :user
+    attribute :delivery_partner_id
+
+    validates :delivery_partner_id, inclusion: { in: :valid_ids }
+
+    def delivery_partner_options
+      @delivery_partner_options ||= user.delivery_partners.each_with_object({}) do |ab, sum|
```
cwrw, line 14: ➕

app/views/admin/delivery_partners/users/delete.html.erb
```diff
@@ -1,9 +1,9 @@
 <% content_for :title, "Delete user" %>
 <h1>Do you want to delete this user?</h1>
-<p class="govuk-!-font-weight-bold">Supplier user: <span class="govuk-!-font-weight-regular"><%= @user.full_name %></span></p>
```
cwrw, line 3: I think we are missing the summary error message box in this form

app/views/admin/delivery_partners/users/edit.html.erb
```diff
@@ -8,13 +8,13 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-l">Edit user details</h1>
 
-    <%= form_with model: @delivery_partners_user_form, scope: :delivery_partners_user_form, url: admin_delivery_partners_user_path(@user) do |f| %>
```
cwrw, line 11: same here regarding missing the summary error box (not part of this pr)

config/locales/en.yml
```diff
@@ -343,6 +343,10 @@ en:
           attributes:
             appropriate_body_id:
               inclusion: Choose an organisation
+        delivery_partners/choose_organisation_form:
+          attributes:
+            delivery_partner_id:
+              inclusion: Choose an organisation
```
cwrw, line 349: should this be choose a delivery partner 🤔 ?

cwrw (commented):
tested and looks great, some minor comments inline

config/locales/en.yml
mooktakim: @cwrw 
In the UI design we ask them to choose an organisation. The error message should stay consistent

app/views/admin/delivery_partners/users/delete.html.erb
mooktakim: @cwrw 
We're not rendering this page to show the errors. We're deleting and then redirecting to index page

app/views/admin/delivery_partners/users/edit.html.erb
mooktakim: @cwrw 
Added, same for new page

cwrw (approved):
👍 Thanks for addressing all the comments

PR #2529, opened 2022-09-27
Allow admins to change induction start date

peteryates:
### Context

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CST/boards/119?assignee=5e1dba26b5771b0ca44142ee&selectedIssue=CST-871

* Add a 'change' link to the details tab so the induction start date can be changed
* Add a form object to validate the change and a regular service object
* Create a form based on the designs in the ticket to update the induction start date

| Details tab | Change induction date form | Success |
| ----------- | -------------------------- | ------------------- |
| Screenshot from 2022-09-27 11-35-43 |  image | image |

### Guidance to review

Do we need additional validation on the start date field? I assume we do but am not sure exactly how it should work (hence the `TODO`)

### DIFF
--- a/app/forms/admin/participants/change_induction_start_date_form.rb
+++ b/app/forms/admin/participants/change_induction_start_date_form.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Admin::Participants
+  class ChangeInductionStartDateForm
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+    include ::ActiveRecord::AttributeAssignment
+
+    attribute :induction_start_date, :date
+
+    validates :induction_start_date, presence: true
+
+    def to_h
+      { induction_start_date: }
+    end
+  end
+end
--- a/app/controllers/admin/participants/change_induction_start_date_controller.rb
+++ b/app/controllers/admin/participants/change_induction_start_date_controller.rb
@@ -12,10 +12,13 @@ module Admin::Participants
     def update
       @participant_profile = retrieve_participant_profile
       @participant_profile_form = ChangeInductionStartDateForm.new(change_induction_start_date_params)
+      induction_start_date = @participant_profile_form.induction_start_date
 
-      if @participant_profile_form.valid?
-        induction_start_date = @participant_profile_form.induction_start_date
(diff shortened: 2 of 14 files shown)


app/forms/admin/participants/change_induction_start_date_form.rb
```diff
@@ -0,0 +1,18 @@
...
+module Admin::Participants
+  class ChangeInductionStartDateForm
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+    include ::ActiveRecord::AttributeAssignment
+
+    attribute :induction_start_date, :date
+
+    validates :induction_start_date, presence: true
+    # TODO: other validation?
```
peteryates, line 12: Not sure what/if there's anything we should be checking here, especially as it's on the admin side.

leodfe: I see some specs in the participant_validation_service_spec.rb related to the `induction_start_date`. Any chances we need to run the ParticipantValidationService when changing the induction_start_date?

peteryates: Yeah - I saw that too. From what I could see it looks unrelated, the [`induction_start_date`](https://github.com/DFE-Digital/early-careers-framework/blob/develop/spec/services/participant_validation_service_spec.rb#L25) variable just clashes with the name of the attribute, but actually represents an `InductionRecord#start_date`.

Looking through the code I can't even see where `#induction_start_date is used, other than displaying it in the admin panel. It was only added in July (407a705bd2).

Also, having just looked at the ticket comments, Mili said:

> confirmed that there are no contractual constraints on induction start date so we don’t need to apply any validation/constraints on changing this

Looks like this should be ok, providing it's targetting the right date. Should it really be changing the start date of the latest induction record? @tonyheadford might have a better idea.

tonyheadford: Yes thats fine as long as there is understanding that, at the moment at least, the `start_date` on the first induction record isn't necessarily the same as the `induction_start_date`.  There's already plenty of scope for misunderstanding :)

tonyheadford (approved):

PR #2543, opened 2022-09-28
[CST-1016] Add mentors who are also ECTs

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-1016)

Enable an ECT to become a mentor.  The mentor profile should be exposed to providers with the same ID as the ECT profile.
This means that there will be 2 participants with the same ID in the participants API endpoint, they will however be different types and (presumably) be in a different cohort year.
There are a small number of requests for this coming through support so this is the initial step to help us get those mentors in and available for training (and iron out any issues that we may not have thought of). Adding the mentor profile will be a developer task for now (using the `Mentors::AddProfileToECT` service in this PR).

Adds a new service that:
 * Adds a mentor profile under the same `ParticipantIdentity` as the ECT profile
 * Enrols the mentor profile into the `default_induction_programme` of the `school_cohort` if available
 * Adds the mentor profile to the school's mentor pool
 * Copies the validation data from the ECT and evaluates the eligibility as a mentor

### Guidance to review

* You should be able to add a mentor profile to any existing ECT from the console using the new service.
* In the UI you should see 2 profiles for the user
* You should be able to select the mentor to mentor ECTs at the selected school
* In the participants API endpoint you should see 2 entries for the external_identifier for the user
* You should be able to post declarations to the ECT and mentor profiles individually.

### DIFF
--- a/app/services/mentors/add_profile_to_ect.rb
+++ b/app/services/mentors/add_profile_to_ect.rb
@@ -17,7 +17,6 @@ module Mentors
           sparsity_uplift: sparsity_uplift?(start_year),
           pupil_premium_uplift: pupil_premium_uplift?(start_year),
           school_cohort:,
-          start_term:,
         )
 
         ParticipantProfileState.create!(participant_profile: mentor_profile,
@@ -40,12 +39,11 @@ module Mentors
 
   private
 
-    attr_reader :ect_profile, :preferred_email, :start_term, :school_cohort, :start_date
+    attr_reader :ect_profile, :preferred_email, :school_cohort, :start_date
 
-    def initialize(ect_profile:, school_cohort:, preferred_email: nil, start_term: nil, start_date: nil)
+    def initialize(ect_profile:, school_cohort:, preferred_email: nil, start_date: nil)
       @ect_profile = ect_profile
       @preferred_email = preferred_email || ect_profile.participant_identity.email
-      @start_term = start_term || school_cohort.cohort.start_term_options.first
       @start_date = start_date
       @school_cohort = school_cohort
     end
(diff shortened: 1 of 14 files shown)


app/services/mentors/add_profile_to_ect.rb
```diff
@@ -0,0 +1,71 @@
...
+      check_no_mentor_profiles_exist!
+
+      mentor_profile = nil
+
+      ActiveRecord::Base.transaction do
+        mentor_profile = ParticipantProfile::Mentor.create!(
+          teacher_profile: ect_profile.teacher_profile,
+          participant_identity: ect_profile.participant_identity,
+          schedule: Finance::Schedule::ECF.default_for(cohort: school_cohort.cohort),
+          sparsity_uplift: sparsity_uplift?(start_year),
```
asmega, line 17: slight digression, can this change from one year to the next? if so, does this need to be specific to the `SchoolCohort` and not `School`

tonyheadford: They can change yes, we will get a new list of uplifts every year from policy.  These are now both stored in the `PupilPremium` records attached to the `School` for each year. I _guess_ it might make sense for them to be on `SchoolCohort` but these could be added before any cohort programme decision is made at a given school so there would be nowhere to put them at the point of importing them.

ltello (approved):

lazydevorg (approved):

peteryates:
Closing this as it's been untouched for ages. Happy to reopen if we want it later.

tonyheadford:
Reopening as we're looking at this now

tonyheadford:
Closing this and reopening as a new one as the review app doesn't want to play nice

PR #2547, opened 2022-09-29
[CPDLP-1810] Return validation error if wrong cohort passed in

cwrw:
### Context
If the wrong cohort is passed in on change schedule, make sure validation completes successfully rather than raising a 500 error due to the schedule not being found

https://sentry.io/organizations/dfe-teacher-services/issues/3491823308/?environment=sandbox&project=5748989&query=is%3Aunresolved

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1810?focusedCommentId=49783&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-49783

Return early in validation if schedule not present to avoid hitting a 500

### DIFF
--- a/app/services/participants/change_schedule/ecf.rb
+++ b/app/services/participants/change_schedule/ecf.rb
@@ -66,7 +66,7 @@ module Participants
 
       def validate_cannot_change_cohort
         if relevant_induction_record &&
-            relevant_induction_record.schedule.cohort.start_year != cohort.start_year
+            relevant_induction_record.schedule.cohort.start_year != cohort&.start_year
           errors.add(:cohort, I18n.t("cannot_change_cohort"))
         end
       end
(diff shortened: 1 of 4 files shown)


asmega (approved):

app/services/participants/change_schedule/ecf.rb
```diff
@@ -159,6 +159,7 @@ def not_already_withdrawn
 
       def validate_new_schedule_valid_with_existing_declarations
         return if user.blank? || user_profile.blank?
+        return unless schedule
```
StupidCodeFactory, line 162: I really don't like how those validations now rely on certain validations to have happened beforehand. 
I'd like to potentially experiment with the chain of responsibility pattern so chain validations. I don't really know if that would come out handy.

PR #2548, opened 2022-09-29
[CST-1063] Cannot add participant from different school message

lazydevorg:
### Context

- Ticket: CST-1063

The "cannot add" logic checks whether the existing participant is in the same school of the new participant.
The message shown changes based on whether the above condition is true or not.

### DIFF
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -360,7 +360,7 @@ module Schools
     end
 
     def existing_participant_profile_same_school?
-      existing_participant_profile&.school == school_cohort.school
+      existing_participant_profile&.current_induction_record&.school == school_cohort.school
     end
   end
 end
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -360,7 +360,7 @@ module Schools
     end
 
     def existing_participant_profile_same_school?
-      existing_participant_profile&.induction_records.latest&.school == school_cohort.school
+      existing_participant_profile&.induction_records&.latest&.school == school_cohort.school
     end
   end
 end
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -360,7 +360,7 @@ module Schools
     end
 
     def existing_participant_profile_same_school?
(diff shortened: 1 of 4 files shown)


app/forms/schools/add_participant_form.rb
```diff
@@ -358,5 +358,9 @@ def appropriate_body_selected
     rescue ActiveRecord::RecordNotFound
       nil
     end
+
+    def existing_participant_profile_same_school?
+      existing_participant_profile&.school == school_cohort.school
```
ltello, line 363: Surely ParticipantProfile::ECF.school is legacy. 
Shouldn't we use the participant_profile's latest_induction_record's school instead ?

tonyheadford: Yes I agree, lets use the induction records for this :+1:

app/views/schools/add_participants/cannot_add.html.erb
```diff
@@ -7,7 +7,11 @@
     <span class="govuk-caption-xl"><%= @school.name %></span>
     <h1 class="govuk-heading-xl"> You cannot add <%= add_participant_form.full_name.titleize %></h1>
 
-    <p class="govuk-body">Our records show this person is already registered on an ECF-based training programme at a different school</p>
+    <% if add_participant_form.existing_participant_profile_same_school? %>
+      <p class="govuk-body">Our records show this person is already registered on an ECF-based training programme at this school</p>
+    <% else %>
+      <p class="govuk-body">Our records show this person is already registered on an ECF-based training programme at a different school</p>
+    <% end %>
```
tonyheadford, line 14: Could we DRY this up a bit?

app/forms/schools/add_participant_form.rb
```diff
@@ -358,5 +358,9 @@ def appropriate_body_selected
     rescue ActiveRecord::RecordNotFound
       nil
     end
+
+    def existing_participant_profile_same_school?
+      existing_participant_profile&.current_induction_record&.school == school_cohort.school
```
ltello, line 363: I wonder if it'd be better to use the latest induction record no matter its induction_status:
`existing_participant_profile&.induction_records&.latest&.school == school_cohort.school`

ltello (approved):

ltello (approved):

PR #2549, opened 2022-09-29
[CPDLP-1798] Inject monthly service fee into contracts

asmega:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1798
- Stakeholders wish to be able to specify ad-hoc monthly service fees that are not derived from other numbers such as recruitment targets

- On a contract one can now set `monthly_service_fee`
- If this is not set it uses the previous mechanism for calculating this value
- If set we use this figure instead when presenting thru statements
- This feature is available to both ECF and NPQ

### Guidance to review

- none

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_09_15_142750) do
+ActiveRecord::Schema.define(version: 2022_09_29_102011) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -536,7 +537,6 @@
     t.boolean "works_in_school"
     t.string "employer_name"
     t.string "employment_role"
-    t.boolean "targeted_support_funding_eligibility", default: false
```
cwrw, line 539: ✂️

spec/services/finance/npq/course_statement_calculator_spec.rb
```diff
@@ -201,4 +201,49 @@
...
+          :npq_contract,
+          :with_monthly_service_fee,
+          npq_lead_provider:,
+          course_identifier: npq_course.identifier,
+          recruitment_target: 438,
+        )
+      end
+
+      it "returns value calulated from monthly_service_fee from contract" do
+        expected = contract.monthly_service_fee / contract.recruitment_target
```
cwrw, line 243: could we return an explicit number instead like the other specs?

leandroalemao (approved):
Agree with @cwrw comments 👍 other than that LGTM 👍 😄

PR #2554, opened 2022-09-30
Ensure we time travel when setting up test scenarios

cwrw:
### Context
Scenarios started failing as we've passed the deadline date on one of the statements (30 September)
Although we are time travelling in the scenarios, we aren't time travelling when creating the statements, which means we are still attaching the declaration to the latest statement which keeps shifting as we go forward.
We fixed that by changing the statement name and pushed it forward.

However, we need to fix this permanently to avoid those failures again after we pass another deadline.

- Ticket: n/a

Create statements when time travelling, and change expected statement to the past, so it never changes again.
Also shift the timestamp we travel to by 2 months to avoid declarations being attached to two statements, and to adhere to the current test setup and expectations.

### DIFF
--- a/spec/support/features/steps/changes_of_circumstance_steps.rb
+++ b/spec/support/features/steps/changes_of_circumstance_steps.rb
@@ -101,10 +101,10 @@ module Steps
       milestone = participant_profile.schedule.milestones
                                      .where(declaration_type: declaration_type.to_s.gsub("_", "-"))
                                      .first
-
-      next_ideal_time milestone.milestone_date - 2.days
+      before_milestone_timestamp = milestone.milestone_date - 2.months
+      next_ideal_time before_milestone_timestamp
       travel_to(@timestamp) do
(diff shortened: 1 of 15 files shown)


spec/support/features/steps/changes_of_circumstance_steps.rb
```diff
@@ -101,10 +101,10 @@ def and_lead_provider_has_made_training_declaration(lead_provider_name, particip
       milestone = participant_profile.schedule.milestones
                                      .where(declaration_type: declaration_type.to_s.gsub("_", "-"))
                                      .first
-
-      next_ideal_time milestone.milestone_date - 2.days
+      before_miletsone_timestamp = milestone.milestone_date - 2.months
```
peteryates, line 104: Typo here, `milestone`

peteryates (commented):
Looks great. One minor typo to fix.

peteryates (approved):

PR #2561, opened 2022-09-30
Improve the admin participant search

peteryates:
### Context

The search/index page wasn't eager loading the results causing a large number of queries. These have been optimised into a couple and some of the scope logic reworked. Participants can now be found by email address too.

- Add scopes for sorting induction records by create
- Eager load relations used in participants table
- Extend the search so it covers email addresses


tonyheadford (approved):
Nice!

PR #2564, opened 2022-10-03
[CPDLP-1692] Add API V3 Statements endpoint

cwrw:
### Context
Add API V3 statements endpoint

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1692

- Add Statements serialiser
- Add `paid?` methods to subclasses to make it easier to calculate paid statements
- Add Statements query service to parse params and return correct statements
- Add API v3 statements route and controller

### Guidance to review
To test:

enable the feature flag: `FeatureFlag.activate("api_v3")`
GET the following endpoint: `/api/v3/statements`
To use different filters: `/api/v3/statements?filter[type]=npq&filter[cohort]=2021,2022`

### DIFF
--- a/app/controllers/api/v3/finance/statements_controller.rb
+++ b/app/controllers/api/v3/finance/statements_controller.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module Api
+  module V3
+    class Finance::StatementsController < Api::ApiController
+      include ApiTokenAuthenticatable
+      include ApiPagination
+      include ApiFilter
+
+      def index
+        render json: Finance::StatementSerializer.new(paginate(finance_statements)).serializable_hash.to_json
+      end
+
+    private
+
+      def finance_statements
+        @finance_statements ||= finance_statements_query.statements
+      end
+
+      def finance_statements_query
+        Finance::Statements::Index.new(
+          cpd_lead_provider: current_user,
+          params: statement_params,
+        )
+      end
+
+      def statement_params
+        params.permit(filter: %i[cohort type])
+      end
+    end
+  end
+end
--- a/app/models/finance/statement/ecf/paid.rb
+++ b/app/models/finance/statement/ecf/paid.rb
@@ -4,4 +4,8 @@ class Finance::Statement::ECF::Paid < Finance::Statement::ECF
   def open?
     false
   end
+
+  def paid?
+    true
+  end
 end
--- a/app/serializers/api/v3/finance/statement_serializer.rb
+++ b/app/serializers/api/v3/finance/statement_serializer.rb
@@ -21,12 +21,7 @@ module Api
         end
 
         attribute :type do |statement|
-          case statement
-          when ::Finance::Statement::ECF
-            "ecf"
-          when ::Finance::Statement::NPQ
-            "npq"
-          end
+          statement.class::STATEMENT_TYPE
         end
 
         attribute :cohort do |statement|
--- a/app/serializers/api/v3/finance/statement_serializer.rb
+++ b/app/serializers/api/v3/finance/statement_serializer.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+require "jsonapi/serializer/instrumentation"
+
+module Api
+  module V3
+    module Finance
+      class StatementSerializer
+        include JSONAPI::Serializer
+        include JSONAPI::Serializer::Instrumentation
+
+        set_id :id
+        set_type :statement
+
+        attribute :month do |statement|
+          statement.name[/([^\s]+)/]
+        end
+
+        attribute :year do |statement|
+          statement.name[/(\d+)/]
+        end
+
+        attribute :type do |statement|
+          case statement
+          when ::Finance::Statement::ECF
+            "ecf"
+          when ::Finance::Statement::NPQ
+            "npq"
+          end
+        end
+
+        attribute :cohort do |statement|
+          statement.cohort.start_year.to_s
+        end
+
+        attribute :cut_off_date do |statement|
+          statement.deadline_date.rfc3339
+        end
+
+        attribute :payment_date do |statement|
+          statement.payment_date.rfc3339
+        end
+
+        attribute :paid, &:paid?
+      end
+    end
+  end
+end
--- a/app/services/api/v3/finance/statements/index.rb
+++ b/app/services/api/v3/finance/statements/index.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module Api
+  module V3
+    module Finance
+      module Statements
+        class Index
+          attr_reader :cpd_lead_provider, :params
+
+          def initialize(cpd_lead_provider:, params:)
+            @cpd_lead_provider = cpd_lead_provider
+            @params = params
+          end
+
+          def statements
+            statement_class
+              .includes(:cohort)
+              .where(
+                cpd_lead_provider:,
+                cohort_id: with_cohorts.map(&:id),
+              ).order(created_at: :asc)
+          end
+
+        private
+
+          def filter
+            params[:filter] ||= {}
+          end
+
+          def with_cohorts
+            return Cohort.where(start_year: filter[:cohort].split(",")) if filter[:cohort].present?
+
+            Cohort.where("start_year > 2020")
+          end
+
+          def statement_class
+            return ::Finance::Statement if filter[:type].blank?
+
+            case filter[:type]
+            when "ecf"
+              ::Finance::Statement::ECF
+            when "npq"
+              ::Finance::Statement::NPQ
+            else
+              ::Finance::Statement.none
+            end
+          end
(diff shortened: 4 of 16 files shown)


app/controllers/api/v3/finance/statements_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Api
+  module V3
+    class Finance::StatementsController < Api::ApiController
```
cwrw, line 5: I've kept to the structure of `finance/statement` in the models

app/models/finance/statement/ecf/paid.rb
```diff
@@ -4,4 +4,8 @@ class Finance::Statement::ECF::Paid < Finance::Statement::ECF
   def open?
     false
   end
+
+  def paid?
```
cwrw, line 8: is this correct, do we trust that all  records have the type` Finance::Statement::ECF::Paid` when they are paid, or is this calculated via the output fee/payment and deadline date?

app/serializers/api/v3/finance/statement_serializer.rb
```diff
@@ -0,0 +1,48 @@
...
+    module Finance
+      class StatementSerializer
+        include JSONAPI::Serializer
+        include JSONAPI::Serializer::Instrumentation
+
+        set_id :id
+        set_type :statement
+
+        attribute :month do |statement|
+          statement.name[/([^\s]+)/]
```
cwrw, line 16: I've used the name to get this value, as the payment date can change manually in the db

app/models/finance/statement/ecf/paid.rb
StupidCodeFactory: @cwrw the `Finance::Statement::ECF::Paid` are manually flipped to paid. So you should be able to trust that all the statement line items should be either `paid`, `clawed_back` or `voided`. Although I doubt this has any kind of  validation as it was done hastily.

app/services/api/v3/finance/statements/index.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module Api
+  module V3
+    module Finance
+      module Statements
+        class Index
+          attr_reader :cpd_lead_provider, :params
+
+          def initialize(cpd_lead_provider:, params:)
```
StupidCodeFactory, line 10: Do we really need keyword arguments for only to arguments? I'd argue that up to 2 arguments it's not necessarily useful.
Another point, I'd rather have to params have passed to `def statements(filter)` so that you could apply different filters to the same instance, debugging in the console.

app/services/api/v3/finance/statements/index.rb
```diff
@@ -0,0 +1,52 @@
...
+            params[:filter] ||= {}
+          end
+
+          def with_cohorts
+            return Cohort.where(start_year: filter[:cohort].split(",")) if filter[:cohort].present?
+
+            Cohort.where("start_year > 2020")
+          end
+
+          def statement_class
```
StupidCodeFactory, line 36: Suggestion: inject this value at runtime rather than trying to guess what is the correct one in this class, gives flexibility to lookup `Finance::Statement::ECF::Paid`, `Finance::Statement::ECF::Payable` etc..

config/routes.rb
```diff
@@ -128,6 +128,7 @@
     end
 
     namespace :v3, constraints: ->(_request) { FeatureFlag.active?(:api_v3) } do
+      resources :statements, only: %i[index], controller: "finance/statements"
```
StupidCodeFactory, line 131: I think this might work and is slightly shorter, but that's really just stylistic so feel free to ignore
```suggestion
      resources :statements, only: %i[index], module: "finance"
```

spec/factories/finance/statements.rb
```diff
@@ -21,6 +21,10 @@
 
     factory :ecf_statement, class: "Finance::Statement::ECF" do
       cpd_lead_provider { association :cpd_lead_provider, :with_lead_provider }
+
+      factory :ecf_paid_statement, class: "Finance::Statement::ECF::Paid" do
```
StupidCodeFactory, line 25: I'd argue that `create(:ecf_statement, :paid)` should be sufficient and adding an extract factory isn't necessarily worth it.

spec/models/finance/statement/ecf/paid_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Finance::Statement::ECF::Paid do
+  subject { create(:ecf_paid_statement) }
+
+  describe "#paid?" do
+    it "returns true" do
```
StupidCodeFactory, line 9: 🤠 
```suggestion
    it { is_expected.to be_paid }
```

spec/requests/api/v3/statements_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+      end
+
+      context "with cohort filter" do
+        let(:params) { { filter: { cohort: "2021" } } }
+        it "returns statements within filter cohort" do
+          get("/api/v3/statements", params:)
+
+          expect(parsed_response["data"].size).to eql(2)
+        end
+      end
```
StupidCodeFactory, line 113: should we creating an RSpec shared helper for those type of specs that we'll have across our APIs? I.e one for pagination and one for unauthorised and not found responses?

spec/serializers/api/v3/finance/statement_serializer_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+require "rails_helper"
+
+module Api
+  module V3
+    module Finance
+      RSpec.describe StatementSerializer do
+        describe "serialization" do
+          let(:cohort) { create(:cohort, :next) }
+          let(:statement) { create(:ecf_statement, name: "January 2022", cohort:) }
+          let(:subject) { described_class.new(statement).serializable_hash }
```
StupidCodeFactory, line 12: I'd decompose that differently:

```ruby
subject(:serialiser) { described_class.new(statement) }

 it "returns the month" do
  expect(serialiser.serializable_hash[:data][:attributes][:month]).to eq("January")
end
```
I think it's a bit less confusing to have the subject be the instance of what is in between `RSpec.describe` _the_thing_we_want_to_test_the_behaviour_of_ `do`, which is the default behaviour. 
And in the body of the test be explicit about what method / behaviour we are testing. 
This way the `subject` never changes and you don't have to jump around the specs to keep track of what the subject is.

spec/serializers/api/v3/finance/statement_serializer_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+            expect(subject[:data][:attributes][:cohort]).to eq("2022")
+          end
+
+          it "returns the cut off dates" do
+            expect(subject[:data][:attributes][:cut_off_date]).to eq(statement.deadline_date.rfc3339)
+          end
+
+          it "returns the payment dates" do
+            expect(subject[:data][:attributes][:payment_date]).to eq(statement.payment_date.rfc3339)
+          end
```
StupidCodeFactory, line 32: Could we collapse those attributes in one single test? ie. `have_including` or `have_attributes`

cwrw: If there was no logic attached to those maybe we could have, but since the logic differs in calculating those values my preference is to keep them separate in their own expectation so can test them under different scenarios and see if they fail

spec/requests/api/v3/statements_spec.rb
cwrw: yes that's a good shout, I'll do that in a separate PR after this is merged so it doesn't pollute this one

spec/factories/finance/statements.rb
cwrw: I added this factory to be inline with the `npq_paid_statement`, however I think we need it due to the class specified, where the `paid` trait does not supply that

app/services/api/v3/finance/statements/index.rb
cwrw: I think using keyword arguments, generally makes our code more explicit, and adds a lot more flexibility to extending classes in the future without having to rely on order and being able to set defaults etc. This service is still likely to change/be extended with further statement endpoints, so I prefer to define keyword arguments. If this was unlikely to ever change then yes I would tend to agree.

Re filter passed through to statements, this service will likely be extended to serve a single statement, I'll revisit this suggestion then to see how we it all fits in

app/services/api/v3/finance/statements/index.rb
cwrw: addressed in https://github.com/DFE-Digital/early-careers-framework/pull/2564/commits/4cbecd97524261228f5dcd155ff0986cc6392314

PR #2567, opened 2022-10-03
[CPDLP-1781] appropriate body participants page

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1781
- Depends on PR: https://github.com/DFE-Digital/early-careers-framework/pull/2528

* Participants page added for appropriate bodies, similar to the delivery partner participants page
* Login as `appropriate-body@example.com`


app/components/appropriate_bodies/participants/table_row.rb
```diff
@@ -0,0 +1,60 @@
...
+        return unless status_name
+
+        title = t("participant_profile_status.status.#{status_name}.title")
+        description = t("participant_profile_status.status.#{status_name}.description")
+
+        if description.present?
+          content_tag(:strong, title) +
+            content_tag(:p, description, class: "govuk-body-s")
+        else
+          content_tag(:strong, title)
```
peteryates, line 41: Not really important here, but if you know the tag you can call it as a method directly. So instead of `content_tag(:strong, title)` it'd be `tag.strong(title)`.

app/views/appropriate_bodies/participants/index.html.erb
```diff
@@ -1,2 +1,62 @@
...
+    <% else %>
+      <tr class="govuk-table__row">
+        <td colspan="11" class="govuk-table__cell">
+          No results found
+        </td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
+<%== govuk_paginate @induction_records, @pagy %>
```
peteryates, line 62:
```suggestion
<%= govuk_paginate @induction_records, @pagy %>
```

app/views/appropriate_bodies/participants/index.html.erb
```diff
@@ -1,2 +1,62 @@
...
+    {
+      field: :status,
+      value: params[:status],
+      label: "status",
+      options: @filter.status_options,
+    },
+  ],
+) %>
+
+<table class="govuk-table">
```
peteryates, line 33: Could this have been made easier using `govuk_table`?

https://govuk-components.netlify.app/components/table/#building-a-table-from-arrays

The only difference is that it doesn't support adding a 'no results' row that spans the whole table, but from an accessibility point of view we should probably not be rendering a table at all by that point.

peteryates (commented):
Looks good overall. Left a few minor comments.

app/components/appropriate_bodies/participants/table_row.rb
mooktakim: @peteryates 
Cool, will make those changes.

app/views/appropriate_bodies/participants/index.html.erb
mooktakim: @peteryates will take a look

peteryates: Cheers @mooktakim. If it's a big job feel free to skip it this time, just wanted to let you know it's there and already included in the project 👍🏽

peteryates (approved):
Looks good

PR #2568, opened 2022-10-04
Move NPQ-specific components into NPQ namespace

peteryates:

This confused me earlier as there are NPQ-specific operations (like `edit_admin_participant_npq_change_full_name_path`) in the markup but there's no indiciation based on the filename/path that it's an NPQ-specific component.

To make it easier to follow I've moved them inside the NPQ namespace.


app/components/admin/npq/participants/details.rb
```diff
@@ -0,0 +1,54 @@
...
+          updated_at.to_formatted_s(:govuk)
+        end
+
+      private
+
+        attr_reader :profile
+
+        class NPQPending < Details; end
+
+        class NPQ < Details; end
```
cwrw, line 50: Not sure if this might now be confusing 🤔

peteryates: I didn't even notice these down here! Who defines classes at the bottom of files? 😅 

I can't see what purpose they serve so will delete them.

cwrw (approved):

PR #2576, opened 2022-10-05
[CPDLP-1689] Add fetch a specific delivery partner via API v3 endpoint 

leandroalemao:
### Context

Providers can fetch a specific delivery partner given its ID via API

- Ticket: [CPDLP-1689](https://dfedigital.atlassian.net/browse/CPDLP-1689)

Add a new endpoint GET: `/api/v3/delivery-partners/:id`

### Guidance to review
[API-v3-Reference](https://ecf-dev.london.cloudapps.digital/api-reference/reference-v3.html#api-v3-delivery-partners-get)
[API v3 changes](https://docs.google.com/document/d/1AfOoSCCZKIdvFL6sijbi7vmsfZNu7RLHejdAWEUTmqo)


app/services/api/v3/delivery_partners/show.rb
```diff
@@ -0,0 +1,32 @@
...
+          @params = params
+        end
+
+        def delivery_partner
+          lead_provider.delivery_partners.find_by("delivery_partners.id = ?", params[:id])
+        end
+
+      private
+
+        def filter
```
cwrw, line 20: I don't think we need filters/with cohorts here?
So regarding where we put this, let's discuss tomorrow as in the PR Im working on I included it in the existing query service, as keen to get your thoughts

leandroalemao: Yeah I forgot to delete.. all sorted now.. tks @cwrw

app/services/api/v3/delivery_partners/show.rb
```diff
@@ -0,0 +1,20 @@
...
+      class Show
+        attr_reader :lead_provider, :params
+
+        def initialize(lead_provider:, params:)
+          @lead_provider = lead_provider
+          @params = params
+        end
+
+        def delivery_partner
+          lead_provider.delivery_partners.find_by("delivery_partners.id = ?", params[:id])
```
cwrw, line 15:
```suggestion
          lead_provider.delivery_partners.find(params[:id])
```
I think if an ID does not exist, or we don't have access to view that delivery partner, we should return a 404 in the API, currently the response exposes the following error instead `PG::InvalidTextRepresentation` and a bad request.

cwrw (commented):
looking good, a few comments

cwrw (approved):

PR #2578, opened 2022-10-05
Cpdlp 1693 view single statement

cwrw:
### Context
Add API V3 `statements /:id` endpoint

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1693

- Add `statement` to Statements query service to a single statement when passing in an ID
- Add API v3 statements route and controller


app/services/api/v3/finance/statements/index.rb
```diff
@@ -21,6 +21,10 @@ def statements
               ).order(payment_date: :asc)
```
cwrw, line 21: chatted with @leandroalemao, we should rename this service

leandroalemao: Yeah my suggestion is to rename it to `queries.rb` or `actions.rb` because we reusing the same class for different actions on the controller.

leandroalemao (commented):
Other than https://github.com/DFE-Digital/early-careers-framework/pull/2578/files#r988809628 LGTM 👍 😄  tks

app/services/api/v3/finance/statements/index.rb
cwrw: addressed in https://github.com/DFE-Digital/early-careers-framework/pull/2578/commits/24aa353d56321e570ca59571e3cd9ddaeef79f66

leandroalemao (approved):
👍 😄

PR #2585, opened 2022-10-07
allow cohort transfers when the participant declarations are not bill…

ltello:
…able or submitted

### Context
  Admin cohort transfer tool reject transfers when the participant has declarations in their current cohort.
  This should not be the case when there are no billable or submitted declarations.
  This PR fixes it.

- Ticket: [1075](https://dfedigital.atlassian.net/browse/CST-1075)

### DIFF
--- a/spec/services/induction/amend_participant_cohort_spec.rb
+++ b/spec/services/induction/amend_participant_cohort_spec.rb
@@ -190,21 +190,41 @@ RSpec.describe Induction::AmendParticipantCohort do
           end
         end
 
-        it "returns true and set no errors" do
-          expect(form.save).to be_truthy
-          expect(form.errors).to be_empty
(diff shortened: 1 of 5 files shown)


cwrw (approved):

spec/services/induction/amend_participant_cohort_spec.rb
```diff
@@ -126,15 +126,22 @@
         end
       end
 
-      context "when the participant has declarations" do
-        before do
-          allow_any_instance_of(described_class).to receive(:participant_declarations).and_return(true)
-        end
+      %i[submitted eligible payable paid].each do |declaration_state|
```
cwrw, line 129: could you add a test for an existing declaration that's clawed back or voided just to show that it goes through?

PR #2589, opened 2022-10-07
Adds basic participant event viewer that include papertrail versions

mattonfoot:
### Context

It should be possible to see the historical events or changes to a CPD record over time in order to help determine whay a record is in the state is is and where we might go to correct or undo any actions that misreported information.

Refactor admin participant views into separate pages per section to allow for non-performant queries such as papertrail versions.

### Guidance to review

login as an admin user and proceed to the participants dashboard, from here you can bring up any participants screen and view the events tab.

bring up the rails console in your chosen environment and run a query to find a `User` or `ParticipantProfile` object. Call `ParticipantEventHistory.from_user` or `ParticipantEventHistory.from_profile` and access the `events` attribute to see a full list of events created for a record.

image


app/views/admin/participants/_events.html.erb
```diff
@@ -0,0 +1,22 @@
...
+        <th scope="row" class="govuk-table__header">Reporter</th>
+      </tr>
+      <% event_list.each do |event| %>
+        <tr>
+          <td class="govuk-table__cell"><%= event.date.to_s(:govuk) %></td>
+          <td class="govuk-table__cell"><%= event.description %></td>
+          <td class="govuk-table__cell"><%= event.reporter %></td>
+        </tr>
+      <% end %>
+    </tbody>
```
peteryates, line 18:
```suggestion
  <table class="govuk-table">
    <thead class="govuk-table__head">
      <tr class="govuk-table__row">
        <th scope="row" class="govuk-table__header">Date</th>
        <th scope="row" class="govuk-table__header">Description</th>
        <th scope="row" class="govuk-table__header">Reporter</th>
      </tr>
    </thead>
    <tbody class="govuk-table__body">
      <% event_list.each do |event| %>
        <tr class="govuk-table__row">
          <td class="govuk-table__cell"><%= event.date.to_s(:govuk) %></td>
          <td class="govuk-table__cell"><%= event.description %></td>
          <td class="govuk-table__cell"><%= event.reporter %></td>
        </tr>
      <% end %>
    </tbody>
```

tonyheadford (commented):
I'm a bit worried about performance here. Our versions table is ridiculously huge and only indexed on the id (I think). I've noticed a bit of slow down with the current tabs and I expect that trawling version info on top during the page render will make it worse. Could this be a page linked from of an existing tab perhaps?

PR #2599, opened 2022-10-10
[CPDLP-1723] Enable admin users to manage AB users on the service

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1723

* New admin section "Appropriate bodies" to manage AB users
* Login with "admin@example.com"
* AB admin console hidden behind feature flag `appropriate_bodies`

### DIFF
--- a/app/controllers/admin/appropriate_bodies/users_controller.rb
+++ b/app/controllers/admin/appropriate_bodies/users_controller.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+module Admin
+  module AppropriateBodies
+    class UsersController < Admin::BaseController
+      skip_after_action :verify_authorized, only: :index
+      skip_after_action :verify_policy_scoped, except: :index
+      before_action :set_appropriate_body_profile_and_authorize, only: %i[edit update delete destroy]
+
+      def index
+        authorize AppropriateBodyProfile
+        scoped = policy_scope(AppropriateBodyProfile).includes(:user).order("users.full_name asc")
+        @pagy, @appropriate_body_profiles = pagy(scoped, page: params[:page], items: 20)
+      end
+
+      def new
+        authorize AppropriateBodyProfile
+        @appropriate_bodies_user_form = ::AppropriateBodies::CreateUserForm.new
+      end
+
+      def create
+        authorize AppropriateBodyProfile
+        @appropriate_bodies_user_form = ::AppropriateBodies::CreateUserForm.new(permitted_params)
+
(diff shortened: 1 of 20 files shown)


app/views/admin/appropriate_bodies/users/delete.html.erb
```diff
@@ -0,0 +1,9 @@
+<% content_for :title, "Delete user" %>
+<h1>Do you want to delete this user?</h1>
+<p class="govuk-!-font-weight-bold">Supplier user: <span class="govuk-!-font-weight-regular"><%= @appropriate_body_profile.user.full_name %></span></p>
+<%= form_for :confirmation, url: admin_appropriate_bodies_user_path(@appropriate_body_profile), method: :delete do |f| %>
+  <div class="govuk-button-group">
+    <%= f.govuk_submit "Delete", classes: "govuk-button--warning", "data-test": "delete-button" %>
```
peteryates, line 6:
```suggestion
    <%= f.govuk_submit "Delete", warning: true, "data-test": "delete-button" %>
```

peteryates:
Code wise it looks ok to me. I notice that in the review app there are two user records listed that contain the same data (and they both change when one is edited). Should there be a distinct in there somewhere?

mooktakim:
> Code wise it looks ok to me. I notice that in the review app there are two user records listed that contain the same data (and they both change when one is edited). Should there be a distinct in there somewhere?

Its the same user with multiple AB roles.  This list is not users but AB profiles for users.
There's larger work that we need to do to combine all users, AB's and DP's in one place we add/edit/delete user and then add/remove roles for that user.

peteryates:
Could we add an ID or something to differentiate them? I think the rows being identical is likely to confuse people.

app/controllers/admin/appropriate_bodies/users_controller.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+module Admin
+  module AppropriateBodies
+    class UsersController < Admin::BaseController
+      skip_after_action :verify_authorized
```
cwrw, line 6: should `verify_authorized` and `verify_policy_scoped` be scoped to only/except index 🤔 ?

mooktakim: @cwrw 
Corrected the pundit policies 👍

cwrw (approved):

peteryates (approved):
Looks good!

PR #2600, opened 2022-10-10
Add partnership challenging to the admin pages

peteryates:
### Context

- Ticket: 

### Summary

* make the existing partnership challenging functionality partner-centric
* display the partnership details on the form (so it's clear what we're challening)
* display all the partnerships for each cohort beneath the listing with a 'challenge' link

The partnership challenging controller appeared to be written in a way that assumes there's only one partnership per cohort per school. I don't think this is necessarily always the case (it certainly isn't in the seed data).

Now instead of grabbing the partnership based on the school and cohort, the partnership id is a URL param.

Because multiple partnerships might be available to edit I've added the relevant details to the top of the challenge form too.

| Parterships list | Challenge form | Confirmation screen | Notification |
| ------------- | --------------- | ----------------- | -------------------- |
| image | image | Screenshot from 2022-10-10 16-58-53 | image |

### DIFF
--- a/app/components/admin/schools/cohorts/cip_info.html.erb
+++ b/app/components/admin/schools/cohorts/cip_info.html.erb
@@ -26,4 +26,6 @@
       <span class="govuk-visually-hidden"> training materials</span>
     </dd>
   </div>
+
+  <%= content %>
 </dl>
--- a/app/components/admin/schools/cohorts/cohort.html.erb
+++ b/app/components/admin/schools/cohorts/cohort.html.erb
@@ -1,4 +1,20 @@
 <h3 class="govuk-heading-m govuk-!-margin-bottom-2">
   <%= cohort.display_name %> cohort
 </h3>
+
 <%= render cohort_info %>
+
+<% if partnerships.present? %>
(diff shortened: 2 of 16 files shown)


tonyheadford (commented):
Looks great.  In the service there will only be 1 reported unchallenged `Partnership` per cohort. This is the default `Partnership` record.  We added the ability to have multiple "relationship" `Partnership` records per cohort to support transfers continuing different programmes.  I don't know if this changes how you might want to present these?  I think it would be really useful to be able to differentiate between the "reported by the provider" `Partnership` and the "created by the service" relationships.

peteryates:
ok @tonyheadford - I'll add some kind of indicator on

peteryates:
I've updated the partnerships list a bit so it displays the _type_ for each (Default partnership vs Relationship):

Screenshot from 2022-10-11 15-22-16

peteryates:
@contentdesign101 and I also rejigged the confirmation screen a bit:

Screenshot from 2022-10-11 15-26-08

tonyheadford (approved):

PR #2602, opened 2022-10-10
CST-1077: sit participants in the right cohort on school transfers

ltello:
### Context

- Ticket: [CST-1077](https://dfedigital.atlassian.net/browse/CST-1077)

- "Calculate" the right cohort to transfer a participant (based on the induction taken or not by the participant in previous schools) ASAP in the add participant and transfer journeys
- Redirect the user to the "school setup cohort" journey if the cohort to add/transfer a participant is not been set up yet.
- Redirect the user back to their initial journey after setting up the missing cohort for the school.
- Finally, sit the participant in the right cohort.


app/views/schools/choose_programme/show.html.erb
```diff
@@ -9,39 +9,39 @@
...
-                    "https://www.gov.uk/government/publications/early-career-framework-reforms-overview/early-career-framework-reforms-overview",
-                    target: :_blank,
-                    rel: "noopener noreferrer",
-                    class: "govuk-link govuk-link--no-visited-state" %>
-                </p>
-                <p class="govuk-body">You can also contact your
+                                  "https://www.gov.uk/government/publications/early-career-framework-reforms-overview/early-career-framework-reforms-overview",
+                                  target: :_blank,
+                                  rel: "noopener noreferrer",
+                                  class: "govuk-link govuk-link--no-visited-state" %>
```
peteryates, line 25: Not really important here but you can pass `no_visited_state: true` to `govuk_link_to` for the same effect.

app/views/schools/add_participants/target_school_cohort_not_set.html.erb
```diff
@@ -0,0 +1,32 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-xl"><%= @school.name %></span>
+    <h1 class="govuk-heading-xl">
+      <%= "We need some more information" %>
+    </h1>
+
+    <p class="govuk-body">
+      Before you can register <%= add_participant_form.full_name.titleize %> for ECF-based training at your school,
```
peteryates, line 13: I'm not sure we shold be titleizing peoples names here. It might work for some but will break others (McDonald vs Mcdonald)

peteryates (commented):
Logic all seems sensible to me. Left a couple of minor comments.

It feels like there's a lot going on in `AddParticipantsController` and `TransferringParticipantsController`. Perhaps this could be reworked slightly into the service layer as a follow up?

app/views/schools/add_participants/target_school_cohort_not_set.html.erb
ltello: You are right. I copied it from a similar view. There must be loads across the code.
I will change at least this one. Thanks!

app/views/schools/choose_programme/show.html.erb
ltello: Great thanks. Much less verbose!.

peteryates: After Christmas it will support `new_tab: true` which means you'd be able to drop ` target: :_blank, rel: "noopener noreferrer"` too, but I [need to wait for Ruby 2.7 support to drop before merging](https://github.com/DFE-Digital/govuk-components/pull/363).

app/views/schools/add_participants/target_school_cohort_not_set.html.erb
peteryates: Ok, I'll have a look for others. Thanks for the heads up.

lazydevorg (approved):

PR #2604, opened 2022-10-11
Use InductionRecords for Support ECTs service user details

mattonfoot:
### Context

The support ECTs service uses the /ecf-user API endpoint which still looks at the User object for email address and the ParticipantProfile::ECF for the school_cohort.

These can, after a transfer, now be totally different to the current InductionRecord.

Change the `ECFUserSerializer` so that it gets the current `InductionRecord` from the most recent `ParticipantProfile::ECF` and then get the relevant attributes from the InductionProgramme referenced here

Additionally the relevant spec tests need to have the setup also create induction_records on an induction_programme

### Guidance to review

modify `lib/api_smoke_test_ecf_induction_records.rb` so it contains the correct `api_token` and `domain` then run the script. It will request both `ecf-users` and `ecf-induction-records` and check each entry for differences.

Example of using the script to call the app within the environment for this PR:

```bash
bundle exec ruby ./lib/api_smoke_test.rb "HAPPY FEET" https://ecf-review-pr-2604.london.cloudapps.digital/api/v1/ecf-users https://ecf-review-pr-2604.london.cloudapps.digital/api/v1/ecf-induction-records
```

To run the smoke test locally simply change the EngageAndLearnApiToken to something you know and call the script like so 

```bash
$ bundle exec rails c
$ EngageAndLearnApiToken.create_with_known_token!("HAPPY FEET")
$ exit
$ bundle exec ruby ./lib/api_smoke_test.rb "HAPPY FEET" https://localhost:3000/api/v1/ecf-users https://localhost:3000/api/v1/ecf-induction-records
```

Example output when differences are found:

```bash
ERROR: The API responses are different
Blue records: 69
Green records: 69
Duplicates found: 0
Matches: 63
Changed: 6
New: 0
Lost: 0


Expected to find
    /attributes/core_induction_programme: "none"

within:
===
{
  "id": "8a7077d9-9c46-4ada-baf7-2ac1646cc020",
  "type": "user",
  "attributes": {
    "cohort": 2021,
    "core_induction_programme": "edt",
    "email": "rp-ect-edt@example.com",
    "full_name": "John Doe",
    "induction_programme_choice": "core_induction_programme",
    "registration_completed": false,
    "user_type": "early_career_teacher"
  }
}
```


app/serializers/api/v1/ecf_user_serializer.rb
```diff
@@ -25,10 +25,15 @@ class ECFUserSerializer
...
+      attributes :full_name
+
+      attributes :email do |user|
+        preferred_identity = find_current_induction_record(user)&.preferred_identity
+        preferred_identity&.email || user.email
+      end
 
       attributes :user_type do |user|
-        case user.teacher_profile.ecf_profiles.last&.type
+        case find_ecf_profile(user)&.type
```
peteryates, line 36: Really minor, but these `attributes` can be singular as they're defining just one attribute.

spec/requests/api/v1/ecf_users_spec.rb
```diff
@@ -8,16 +8,27 @@
...
+    let(:mentor_profile) { create :mentor_participant_profile, school_cohort:, core_induction_programme:, cohort: }
+
+    let!(:npq_participant) { create :npq_participant_profile, school: }
+    let!(:npq_mentor_participant) { create :npq_participant_profile, school:, teacher_profile: mentor_profile.teacher_profile }
+    let!(:ecf_participants) { create_list :ect_participant_profile, 2, school_cohort:, core_induction_programme:, cohort: }
+
+    before do
+      Induction::Enrol.call(participant_profile: mentor_profile, induction_programme:, start_date: 2.months.ago)
+      Induction::Enrol.call(participant_profile: ecf_participants[0], induction_programme:, start_date: 2.months.ago)
+      Induction::Enrol.call(participant_profile: ecf_participants[1], induction_programme:, start_date: 2.months.ago)
```
peteryates, line 31: I think it'd be more readable to have `first_ecf_participant` and `second_ecf_participant` defined separately and referred to by their name rather than index.

peteryates (commented):
Looks fine to me overall, left a couple of really minor comments. I'd approve but I think we need someone with a better knowledge of the API to have a look too.

app/serializers/api/v1/ecf_user_serializer.rb
mattonfoot: These have now been removed as I have moved to an induction_record focused serialiser

lib/api_smoke_test.rb
```diff
@@ -0,0 +1,62 @@
...
+if new_records.count.zero? && lost_records.count.zero?
+  puts "The API responses are the same"
+else
+  puts "ERROR: The API responses are different"
+  puts "blue_records: #{blue_data.count}"
+  puts "green_records: #{green_data.count}"
+  puts "duplicates_found: #{green_data.count - deduped_records.count}"
+  puts "matches: #{same_records.count}"
+  puts "new: #{new_records.count}"
+  puts "lost: #{lost_records.count}"
```
peteryates, line 60: I think this can be simplified a bit by creating an array of ids:

```ruby
blue_ids = blue_data.map { |r| r[:id] }
green_ids = green_data.map { |r| r[:id] }

duplicate_records = green_ids.tally.filter { |_, v| v > 1 } # assuming we're counting the number of duplications rather than duplicated records
same_records      = (blue_ids & green_ids)
new_records       = green_ids - blue_ids
lost_records      = blue_ids - green_ids

```

peteryates:
Again, it looks good to me. Left another minor suggestion for making the smoke test easier to follow.

lib/api_smoke_test.rb
mattonfoot: this is much simpler

app/serializers/api/v1/ecf_induction_record_serializer.rb
```diff
@@ -0,0 +1,76 @@
...
+          CIP_TYPES[:teach_first]
+        when "UCL Institute of Education"
+          CIP_TYPES[:ucl]
+        else
+          CIP_TYPES[:none]
+        end
+      end
+
+      attributes :induction_programme_choice do |induction_record|
+        induction_record.school_cohort.induction_programme_choice
```
tonyheadford, line 64:
```suggestion
        induction_record.induction_programme.training_programme
```

tonyheadford (commented):
One comment, is this ready to go out now or are we waiting a review from a wider audience?

PR #2606, opened 2022-10-12
Ensure that school_cohort combinations are unique

peteryates:
### Context

We had a case recently where a school had two 2021 cohort records. This should never happen. This change enforces it both at the database and model level.

* add a unique index to `school_cohorts` on `school_id` and `cohort_id`
* add a unique constraint to the `SchoolCohort` model

### Before merging

**Don't merge this until we have no more duplicates.**

This query should return no results in prod.

```
with dupes as (
  select school_id, cohort_id
  from school_cohorts
  group by school_id, cohort_id
  having count(*) > 1
)
select urn, name, postcode
from schools
join dupes on schools.id = dupes.school_id;
```

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_09_29_104505) do
+ActiveRecord::Schema.define(version: 2022_10_12_103024) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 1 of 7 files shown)


db/schema.rb
```diff
@@ -631,7 +631,7 @@
     t.boolean "sparsity_uplift"
     t.boolean "pupil_premium_uplift"
     t.uuid "delivery_partner_id"
-    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY ((ARRAY['submitted'::character varying, 'eligible'::character varying, 'payable'::character varying, 'paid'::character varying, 'clawed_back'::character varying])::text[]))"
+    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY (ARRAY[('submitted'::character varying)::text, ('eligible'::character varying)::text, ('payable'::character varying)::text, ('paid'::character varying)::text, ('clawed_back'::character varying)::text]))"
```
peteryates, line 634: My version here is from a clean `be rails db:setup`, this shouldn't change. I suspect someone has a local dev database slightly out of sync with the `schema.rb` definition.

StupidCodeFactory: @peteryates  I've got the same as well after running db:migrate I usually nuke the diff but it's looks like the way the schema is regenerated has changed since. But it does have the same values on the partial index

peteryates: Yeah it's an odd one. I'll dig into it when I've got a min. Will repush this with the diff removed.

peteryates:
Closing for now as there'll be data to clean up first. Will revisit.

PR #2610, opened 2022-10-13
[CPLD-1730] Refactor deferring a participant

cwrw:
### Context

To move to the new world of API rendering and services, refactor Defer API endpoint.

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1730

- Add new defer participant service
- Add methods between NPQ and ECF profiles to expose a shared interface, to simplify the service
- Use new API methods to surface response or validation errors
- Remove old defer participant implementation

### Guidance to review

Since we have to live in both worlds of `participant_actions` vs new actions, we will need to reuse some of those components such as `params_for_recorder` etc. however when we move all actions to the new world, we can get rid of participant actions and have have explicit methods for both NPQ and ECF


app/controllers/api/v1/npq_participants_controller.rb
```diff
@@ -29,6 +29,12 @@ def withdraw
         end
       end
 
+      def defer
+        service = DeferParticipant.new(params_for_recorder)
```
cwrw, line 33: Since we have to live in both worlds of `participant_actions` vs new actions, we will need to reuse some of those components such as `params_for_recorder` etc. however when we move all actions to the new world, we can get rid of participant actions and have have explicit methods for both NPQ and ECF

leandroalemao (approved):

app/models/participant_profile/ecf.rb
```diff
@@ -93,7 +93,19 @@ def schedule_for(cpd_lead_provider:)
...
+
+    def withdrawn_for?(cpd_lead_provider:)
+      !!latest_induction_record_for(cpd_lead_provider:)&.training_status_withdrawn?
+    end
+
+    def deferred_for?(cpd_lead_provider:)
+      !!latest_induction_record_for(cpd_lead_provider:)&.training_status_deferred?
+    end
+
+    def serialize_for(lead_provider:)
```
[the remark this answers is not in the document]
peteryates: I'd agree the method name doesn't look like it's lining up with the intent. It's also just wrapping `#relevant_induction_record` - could it be an alias?

peteryates (approved):
This is a **huge** improvement. The deferral process is already way easier to follow.

cwrw:
@peteryates @StupidCodeFactory @leandroalemao thanks for your review. I've addressed all comments in a separate un-blocker PR: https://github.com/DFE-Digital/early-careers-framework/pull/2627 as there are multiple refactors happening at once, to unblock others while this is tested, if you don't mind having a look there. I'll rebase this as soon as that is merged. Thanks!

app/services/defer_participant.rb
```diff
@@ -0,0 +1,89 @@
...
+  def participant_profile
+    @participant_profile ||=
+      ParticipantProfileResolver.call(
+        participant_identity:,
+        course_identifier:,
+        cpd_lead_provider:,
+      )
+  end
+
+  def schedule
```
leandroalemao, line 48: Is this being used anywhere?

StupidCodeFactory (approved):

app/services/defer_participant.rb
cwrw: We need the surface schedule for the induction record validator, see: https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/validators/induction_record_validator.rb#L31

PR #2614, opened 2022-10-14
Fix relevant induction record lookup

cwrw:
### Context

ECF Action participant endpoints such as defer/resume/withdraw, find the relevant induction record for a lead provider to serialise a response in the API: https://github.com/DFE-Digital/early-careers-framework/blob/7ab4cbc54985cbcdcb85e0736c1bd11bc56a71f5/app/controllers/concerns/api/v1/participant_actions.rb#L32

When we lookup those induction records we are currently only scoping them to the current cohort i.e. 2022. So if an action is attempted for participants with induction record in older cohorts they get an empty body in the API response as no induction record is returned

- Ticket: n/a

We should return induction records for all cohorts, and not just the current cohort for a lead provider.

### Guidance to review
Did I miss anything?


mooktakim (approved):
👍

PR #2615, opened 2022-10-17
[CPDLP-1744] - add guard clauses so that double profile can not be created

StupidCodeFactory:
### Context

- [Ticket](https://dfedigital.atlassian.net/browse/CPDLP-1744) 

This change aims at preventing creation of duplicate profile whilst we resolve and soft delete the existing duplicate profiles

### DIFF
--- a/app/services/mentors/create.rb
+++ b/app/services/mentors/create.rb
@@ -3,22 +3,16 @@
 module Mentors
   class Create < BaseService
     include SchoolCohortDelegator
+    ParticipantProfileExistsError = Class.new(RuntimeError)
 
     def call
       mentor_profile = nil
       ActiveRecord::Base.transaction do
-        # NOTE: This will not update the full_name if the user has an active participant profile,
-        # the scenario I am working on is enabling a NPQ user to be added as a mentor
-        # Not matching on full_name means this works more smoothly for the end user
-        # and they don't get "email already in use" errors if they spell the name differently
-        user = User.find_or_create_by!(email:) do |mentor|
-          mentor.full_name = full_name
-        end
         user.update!(full_name:) unless user.teacher_profile&.participant_profiles&.active_record&.any?
 
-        teacher_profile = TeacherProfile.find_or_create_by!(user:) do |profile|
-          profile.school = school_cohort.school
-        end
+        create_teacher_profile
+
+        raise ParticipantProfileExistsError if participant_profile_exists?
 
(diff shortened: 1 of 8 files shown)


spec/services/identity/create_spec.rb
```diff
@@ -47,7 +47,7 @@
       context "when the existing profiles belong to another identity" do
         let(:ect)                { create(:mentor, :eligible_for_funding, user:) }
         let(:exisiting_identity) { ect.participant_identities.first }
-        let!(:mentor_profile)    { create(:mentor, :eligible_for_funding, trn: ect.teacher_profile.trn, user:) }
+        let!(:mentor_profile)    { create(:mentor, :eligible_for_funding, trn: ect.teacher_profile.trn).tap { |pp| pp.update!(teacher_profile: ect.teacher_profile) } }
```
peteryates, line 50: Is there a reason this can't be 

```ruby
create(:mentor,
       :eligible_for_funding,
       trn: ect.teacher_profile.trn,
       teacher_profile: ect.teacher_profile)
```
I'm guessing it's an oddity of our factory setup?

app/services/mentors/create.rb
```diff
@@ -67,5 +61,26 @@ def mentor_attributes
...
+
+    def teacher_profile
+      @teacher_profile ||= TeacherProfile.find_or_create_by!(user:) do |profile|
+        profile.school = school_cohort.school
+      end
+    end
+    alias_method :create_teacher_profile, :teacher_profile
+
+    def participant_profile_exists?
+      teacher_profile.participant_profiles.mentors.exists? || user.participant_identities.joins(:participant_profiles).where(participant_profiles: { type: "ParticipantProfile::Mentor" }).exists?
```
peteryates, line 83: I have a feeling we do this elsewhere too. Think it's worth having a `Participants::AlreadyExists` service?

peteryates (approved):
Looks good. Left a couple of minor comments.

spec/services/identity/create_spec.rb
StupidCodeFactory: @peteryates yes the factory does not allow for it, hence the odd setup. Which I think is ok, since this should not happen and denotes we're dealing with legacy data.

app/services/mentors/create.rb
StupidCodeFactory: @peteryates I'd be happy to implement such service, I couldn't find where something similar was being done.
But I'll keep that one mind.

PR #2620, opened 2022-10-18
Remove any remaining calls to #titleize names

peteryates:
### Context

- Ticket: 

Follows on from #2616. Didn't pick these up on the first sweep asI only looked in views.

I also removed a couple of overlapping apostrophe-adding methods, everything now uses the `#possessive_name` helper.

### DIFF
--- a/app/forms/schools/add_participant_form.rb
+++ b/app/forms/schools/add_participant_form.rb
@@ -322,7 +322,7 @@ module Schools
     end
 
     def display_name
-      type == :self ? "your" : "#{full_name&.titleize}’s"
+      type == :self ? "your" : "#{full_name}’s"
     end
 
     def formatted_nino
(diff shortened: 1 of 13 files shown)


app/forms/schools/add_participant_form.rb
```diff
@@ -322,7 +322,7 @@ def store_validation_result!(profile)
     end
 
     def display_name
-      type == :self ? "your" : "#{full_name&.titleize}’s"
+      type == :self ? "your" : "#{full_name}’s"
```
tonyheadford, line 325: Should this use `possessive_name` ?

tonyheadford (approved):
Looks good, one comment but feel free to ignore

app/forms/schools/add_participant_form.rb
peteryates: I left this one as we can't see helpers from the form object and moving the logic would be a bit more involved. I agree it should though.

PR #2621, opened 2022-10-18
[CPDLP-1854] Refactor change schedule endpoints

leandroalemao:
### Context

Refactor `change_schedule` endpoints in preparation for the new API v3

- Ticket: [CPDLP-1854](https://dfedigital.atlassian.net/browse/CPDLP-1854)

Refactor API v1 & v2 `change_schedule` endpoints in preparation for the new API v3 endpoint


app/services/valid_courses_resolver.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class ValidCoursesResolver
```
cwrw, line 3: could we use the `CourseValidator` instead as I think it's doing something similar?

app/services/change_schedule.rb
```diff
@@ -0,0 +1,174 @@
+# frozen_string_literal: true
+
+class ChangeSchedule
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Validations::Callbacks
+  include Participants::ProfileAttributes
```
cwrw, line 7: could we avoid including profile attributes, as we will be deprecating all those when this work is complete?

app/services/change_schedule.rb
```diff
@@ -0,0 +1,174 @@
...
+                  Cohort.find_by(start_year: super)
+                else
+                  Cohort.current
+                end
+  end
+
+  def relevant_induction_record
+    return if user.blank? || participant_profile.blank?
+
+    participant_profile
```
cwrw, line 89: could we reuse the methods we have already on ECF profiles to look this up?

app/services/change_schedule.rb
```diff
@@ -0,0 +1,174 @@
...
+      .induction_records
+      .joins(induction_programme: { partnership: [:lead_provider] })
+      .where(induction_programme: { partnerships: { lead_provider: } })
+      .order(created_at: :asc)
+      .last
+  end
+
+  def not_already_withdrawn
+    if ParticipantProfile::ECF::COURSE_IDENTIFIERS.include?(course_identifier)
```
cwrw, line 98: let's catchup around this as I've done this differently on my PR

app/services/change_schedule.rb
```diff
@@ -0,0 +1,174 @@
...
+        Induction::ChangeInductionRecord.call(
+          induction_record: relevant_induction_record,
+          changes: {
+            schedule:,
+          },
+        )
+      end
+    end
+
+    relevant_induction_record_for_profile(participant_profile)
```
cwrw, line 45: let's also have a chat about this, I introduced a method particularly for serialisation, as I think this will return `nil` now for NPQ participants when changing schedule

spec/services/participants/ecf_spec.rb
```diff
@@ -1,63 +0,0 @@
-# frozen_string_literal: true
```
cwrw, line 1: I think we can keep this spec as well as the one below as we're still using those in the remaining action methods, we can delete those in the end?

app/services/change_schedule.rb
```diff
@@ -0,0 +1,187 @@
+# frozen_string_literal: true
+
+class ChangeSchedule
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Validations::Callbacks
+
+  attribute :cpd_lead_provider
+  attribute :participant_id
+  attribute :reason
```
cwrw, line 10: Could you remove `reason` as I don't think it's required for changing a schedule

app/services/user_profile_resolver.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class UserProfileResolver
```
cwrw, line 3: is this used anywhere, can it be deleted?

app/services/valid_courses_resolver.rb
```diff
@@ -0,0 +1,35 @@
...
```
cwrw, line 3: same here, is this used anywhere and can we delete?

app/services/change_schedule.rb
```diff
@@ -0,0 +1,187 @@
...
+  attribute :course_identifier
+  attribute :schedule_identifier
+  attribute :cohort
+
+  alias_attribute :user_profile, :participant_profile
```
cwrw, line 15: I was wondering why we needed to alias participant profile 🤔, could we always instead use `participant_profile` to avoid confusion ?

app/services/change_schedule.rb
```diff
@@ -0,0 +1,187 @@
...
+  delegate :participant_profile_state, to: :participant_profile, allow_nil: true
+  delegate :lead_provider, to: :cpd_lead_provider, allow_nil: true
+
+  validates :course_identifier, course: true, presence: { message: I18n.t(:missing_course_identifier) }
+  validates :participant_id, presence: { message: I18n.t(:missing_participant_id) }
+  validates :participant_id, format: { with: /\A[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\Z/, message: I18n.t("errors.participant_id.invalid") }, allow_blank: true
+  validates :cpd_lead_provider, presence: { message: I18n.t(:missing_cpd_lead_provider) }
+  validate :participant_has_user_profile
```
cwrw, line 23: can we use this validation instead as it will validate the participant profile and induction record, as well as the cpd lead provider is linked to the profile?`validates :cpd_lead_provider, induction_record: true`

app/services/change_schedule.rb
```diff
@@ -0,0 +1,187 @@
...
+  attribute :course_identifier
+  attribute :schedule_identifier
+  attribute :cohort
+
+  alias_attribute :user_profile, :participant_profile
+  delegate :participant_profile_state, to: :participant_profile, allow_nil: true
+  delegate :lead_provider, to: :cpd_lead_provider, allow_nil: true
+
+  validates :course_identifier, course: true, presence: { message: I18n.t(:missing_course_identifier) }
+  validates :participant_id, presence: { message: I18n.t(:missing_participant_id) }
```
cwrw, line 20: I'm not sure we need the next 3 validations, if we add the induction record validator, let me know your thoughts

app/services/change_schedule.rb
```diff
@@ -0,0 +1,187 @@
...
+    end
+  end
+
+  def participant_has_user_profile
+    return if errors.any?
+
+    errors.add(:participant_id, I18n.t(:invalid_participant)) if user_profile.blank?
+  end
+
+  def relevant_induction_record_for_profile(participant_profile)
```
cwrw, line 184: ✂️

app/services/change_schedule.rb
```diff
@@ -0,0 +1,187 @@
...
+
+  def validate_provider
+    return if user.blank? || participant_profile.blank?
+
+    unless participant_profile && matches_lead_provider?
+      errors.add(:participant_id, I18n.t(:invalid_participant))
+    end
+  end
+
+  def matches_lead_provider?
```
cwrw, line 135: maybe we can add a shared interface for NPQ/ECF profiles similar to withdrawn etc.?

cwrw (commented):
looking good, a few comments, thanks for addressing the others!

app/services/change_schedule.rb
leandroalemao: @cwrw All done tks 👍 😄

app/services/change_schedule.rb
leandroalemao: @cwrw All done tks 👍 😄

app/services/change_schedule.rb
```diff
@@ -0,0 +1,174 @@
...
+  def valid_courses
+    ValidCoursesResolver.call(
+      course_identifier:,
+    )
+  end
+
+private
+
+  def cohort
+    @cohort ||= if super
```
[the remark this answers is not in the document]
leandroalemao: @cwrw It seems not be working as expected so left as It was in the initial commit.

app/services/change_schedule.rb
leandroalemao: For some reason some pre-existing tests started to fail after I added the `induction_record: true` validation. So kept how it was in the first place.

app/services/change_schedule.rb
leandroalemao: @cwrw All done tks

app/models/participant_profile/ecf.rb
```diff
@@ -105,6 +105,10 @@ def deferred_for?(cpd_lead_provider:)
       !!latest_induction_record_for(cpd_lead_provider:)&.training_status_deferred?
     end
 
+    def matches_lead_provider?(cpd_lead_provider:)
+      relevant_induction_record(lead_provider: cpd_lead_provider&.lead_provider).present?
```
StupidCodeFactory, line 109: @leandroalemao don't use this method, it's sorting by `InductionRecord#start_date` when it should be `InductionRecord#created_at`

app/services/change_schedule.rb
```diff
@@ -56,6 +56,19 @@ def participant_profile
                                )
   end
 
+  def alias_search_query
+    Finance::Schedule
+      .where("identifier_alias IS NOT NULL")
```
peteryates, line 61:
```suggestion
      .where.not(identifier_alias: nil)
```

app/models/participant_profile/ecf.rb
peteryates: We should really have one way to do this. If `#relevant_induction_record` isn't doing the right thing can we fix/delete it?

(not one for this PR @leandroalemao, more of a general comment - perhaps worth adding to the backlog)

app/services/change_schedule.rb
leandroalemao: Addressed tks 👍 😄 🔝

cwrw (approved):
Thanks for addressing all the comments, looks great and tested. Just some final comments around a duplicate validation but past that 👍

app/services/change_schedule.rb
```diff
@@ -0,0 +1,153 @@
...
+  end
+
+  def validate_cannot_change_cohort
+    if relevant_induction_record &&
+        relevant_induction_record.schedule.cohort.start_year != cohort&.start_year
+      errors.add(:cohort, I18n.t("cannot_change_cohort"))
+    end
+  end
+
+  def schedule_valid_with_pending_declarations
```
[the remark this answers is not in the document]
leandroalemao: Hi @cwrw I'll wait for Tom when he's back then so we can ask him about it. 👍 😄

cwrw (approved):
⭐

app/services/change_schedule.rb
```diff
@@ -131,4 +132,10 @@ def validate_cannot_change_cohort
       errors.add(:cohort, I18n.t("cannot_change_cohort"))
     end
   end
+
+  def change_with_a_different_schedule
+    return unless new_schedule && participant_profile && new_schedule == participant_profile.schedule
+
+    errors.add(:schedule_identifier, I18n.t(:invalid_schedule))
```
cwrw, line 139: should we return an error message that indicates this schedule is already on the profile rather than invalid?

spec/services/change_schedule_spec.rb
```diff
@@ -117,8 +127,8 @@
 
   context "ECT participant profile" do
     let(:cpd_lead_provider) { induction_record.cpd_lead_provider }
-    let(:schedule_identifier) { "ecf-standard-september" }
```
cwrw: why did those need to change? can we also use `schedule_identifier:,` instead of setting the same value twice?

cwrw (commented):
a few minor comments, happy to chat in person

app/services/change_schedule.rb
```diff
@@ -0,0 +1,141 @@
...
+
+  def user
+    @user ||= participant_identity&.user
+  end
+
+  def cohort
+    @cohort ||= if super
+                  Cohort.find_by(start_year: super)
+                else
+                  Cohort.current
```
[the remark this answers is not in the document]
StupidCodeFactory: That being said you'd have to go and change it in other similar service classes. Maybe more something to keep for the ultimate pass once everything has been written and we can extract a couple abstractions to make things a bit easier to test and end our reliance on `Cohort.current` 🙈

cwrw (approved):
Third approval :D, thanks @leandroalemao

PR #2624, opened 2022-10-18
CPDLP-1869-manually-fix-x6-retained-1-declarations-from-ecf-to-npq-for-teach-first-and-ambition

StupidCodeFactory:
### Context

- [Ticket](https://dfedigital.atlassian.net/browse/CPDLP-1869)

Fix the data participant declaration not having the correct type and not attached to the correct statement


cwrw (commented):
looks good, could you also update the participant profile id on the declaration, as I think they are attached to the ECF profiles rather than the NPQ ones?

PR #2627, opened 2022-10-19
Expose shared interface for Participant profiles

cwrw:
### Context
This is an unblocker PR as we are refactoring multiple endpoints at the same time

- Ticket: n/a

- Add shared methods for both NPQ and ECF profiles to be used in action endpoints refactor

### Guidance to review

based off https://github.com/DFE-Digital/early-careers-framework/pull/2610

### DIFF
--- a/spec/models/participant_profile/npq_spec.rb
+++ b/spec/models/participant_profile/npq_spec.rb
@@ -27,7 +27,7 @@ RSpec.describe ParticipantProfile::NPQ, :with_default_schedules, type: :model do
     end
   end
 
-  describe "#withdrawn_for", :with_default_schedules do
+  describe "#withdrawn_for" do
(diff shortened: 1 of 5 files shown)


spec/models/participant_profile/npq_spec.rb
```diff
@@ -26,4 +26,56 @@
...
+  end
+
+  describe "#deferred_for" do
+    let(:cpd_lead_provider) { subject.npq_application.npq_lead_provider.cpd_lead_provider }
+
+    context "when participant is deferred" do
+      subject { create(:npq_participant_profile, :deferred) }
+
+      it "returns true" do
+        expect(subject.reload.deferred_for?(cpd_lead_provider:)).to be true
```
cwrw, line 57: I kept this format as we need to reload subject, so couldn't use predicate matchers

leandroalemao (approved):
LGTM 😄 👍

PR #2628, opened 2022-10-19
Display all participant's identities for ECTs in their admin page

leoapost:
### Context

Add a new tab in the participant's profile to display all the identities of ECT users. This will allow easier troubleshooting from the admin page.

- Ticket: [CPDNPQ-640](https://dfedigital.atlassian.net/browse/CPDNPQ-640)

I created a new partial to display key info from the participant identities and a label to know if it's an original or transferred identity. The partial is rendered in a new tab in the admin participant profile page.

### Guidance to review
1. visit the participants admin pages
2. open the profile of an ECT participant. The `identities` tab should be displayed
3. open the profile of an NPQ participant. The `identities` tab should not be displayed

| NPQ      | ECF           |
| ------------- |-------------|
| npq-identities | ecf-identities |


app/views/admin/participants/_identities.html.erb
```diff
@@ -0,0 +1,25 @@
...
+      <% row.key(text: "Email" ) %>
+      <% row.value(text: identity.email ) %>
+    <% end %>
+
+    <% summary_list.row do |row| %>
+      <% row.key(text: "Origin" ) %>
+      <% row.value(text: identity.origin ) %>
+    <% end %>
+  <% end %>
+<% end %>
```
tonyheadford, line 25: Could this use your new component?

tonyheadford (commented):
Couple of comments but looks great.

app/helpers/admin_helper.rb
```diff
@@ -46,4 +46,12 @@ def format_address(*parts)
 
     safe_join(parts.compact_blank, tag.br)
   end
+
+  def identity_transferred_label(identity)
+    if identity.user_id == identity.external_identifier
+      "Original"
+    else
+      "Transferred"
```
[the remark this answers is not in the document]
leoapost: How do we identify these "Additional" or "Secondary" ones?

app/components/admin/npq/participants/identities.html.erb
```diff
@@ -0,0 +1,29 @@
+<h2 class="govuk-heading-m govuk-!-margin-top-9">
+  Identities
+</h2>
+
+<% identities.each_with_index do |identity, i| %>
+  <h3><%= "Identity #{i + 1} (#{identity_transferred_label(identity)})" %></h3>
```
peteryates, line 6:
```suggestion
<% identities.each.with_index(1) do |identity, i| %>
  <h3><%= "Identity #{i} (#{identity_transferred_label(identity)})" %></h3>
```

app/views/admin/participants/_identities.html.erb
leoapost: I moved the component outside the NPQ namespace to make it more generic and refactored the view to re use it.

app/helpers/admin_helper.rb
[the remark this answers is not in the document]
tonyheadford: Ah, wasn't quite what I meant but a really good question! So we can identify an original easy enough.  If the identity was transferred due to de-duping it would most likely have one or more participant_profiles.  The additional identites used to give participants additional email addresses would most likely not have any participant profiles.  But all I was suggesting was that was there something more inclusive/helpful to call the non-original ones rather than "Transferred".  It is only a suggestion, please don't let this hold you up

app/views/admin/participants/_identities.html.erb
tonyheadford: Nice!

tonyheadford (approved):

app/helpers/admin_helper.rb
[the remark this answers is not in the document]
leoapost: Oh I see. If it's OK, I think I prefer to keep it consistent with the rest of the code for now and review/refactor everything together at a later stage.

PR #2635, opened 2022-10-20
[CPLD-1736] Refactor resuming a participant

cwrw:
### Context
To move to the new world of API rendering and services, refactor Resume API endpoint.
- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1736

- Add new resume participant service
- Add methods between NPQ and ECF profiles to expose a shared interface, to simplify the service
- Use new API methods to surface response or validation errors
- Remove old resume participant implementation

### Guidance to review
The other refactor prs have not been merged, due to testing, so things might be refactored further when that happens. Especially around moving out shared validations into validators, as well as rethinking the form implementations


app/forms/finance/ecf/change_training_status_form.rb
```diff
@@ -34,16 +34,20 @@ def save
...
-            force_training_status_change: true,
-          },
-        )
+        params = {
+          cpd_lead_provider:,
+          course_identifier:,
+          participant_id: participant_profile.participant_identity.external_identifier,
+        }
+
+        if training_status == "active"
```
cwrw, line 43: need to rethink this when other refactor prs are merged, since we are no blocked with testing

app/services/resume_participant.rb
```diff
@@ -0,0 +1,81 @@
...
+
+private
+
+  def not_already_active
+    return unless participant_profile
+
+    errors.add(:participant_profile, I18n.t(:already_active)) if participant_profile.active_for?(cpd_lead_provider:)
+  end
+
+  def not_already_withdrawn
```
cwrw, line 53: this should move to a shared validator once other prs are merged

StupidCodeFactory: I don't think we should event have this validation in the first place. We could simply just return a 200.

spec/services/resume_participants_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+  it_behaves_like "resuming a participant"
+
+  it "creates a new active induction record" do
+    expect { service.call }.to change { InductionRecord.count }
+  end
+
+  it "adds the correct attributes to the new induction_record" do
+    service.call
+
+    expect(participant_profile.induction_records.latest.training_status).to eq("active")
```
StupidCodeFactory, line 102: 🙈 
```suggestion
    expect(participant_profile.induction_records.latest).to be_training_status_active
```

StupidCodeFactory (approved):
👍  LGTM

One comment on the "not already withdrawn" error which to me seems pointless

app/services/resume_participant.rb
cwrw: Good question, I think resume only applies to defer and not withdraw. My understanding is that you cannot re activate someone already withdrawn, but can resume a deferred participant. We allow this only in the finance drilldown tool as it's internal. Will confirm with Tom, but this is maintaining previous business logic

leandroalemao (approved):
LGTM 👍 😄

PR #2637, opened 2022-10-21
Fix flaky serializer spec

tonyheadford:
### Context

Noticed that a intermittently failing spec was due to a timing inconsistency where the wrong object `created_at` and `updated_at` were being compared in this spec.


cwrw (approved):
👍

PR #2638, opened 2022-10-21
Use explicit name to avoid failure in regex expectation

cwrw:
### Context
When checking for the following in the page:
```
/What is #{ect_profile.user.full_name}.*/
```
Names such as: Elza O'Reilly will fail that regex because they are formatted differently on the page. 
- Ticket: n/a

Set the user name explicitly to avoid failures on random names.

### Guidance to review
example of failure: https://github.com/DFE-Digital/early-careers-framework/actions/runs/3295940164/jobs/5435027414

### DIFF
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :user do
     email { Faker::Internet.unique.safe_email }
-    full_name { Faker::Name.name }
+    sequence(:full_name) { |n| "John Doe #{n}" }
     login_token { Faker::Alphanumeric.alpha(number: 10) }
     login_token_valid_until { 12.hours.from_now }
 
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Looks good.

Think it's worth changing the user factory full_name to be a sequence by default? Maybe with a `random_name` trait.

spec/factories/users.rb
```diff
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :user do
     email { Faker::Internet.unique.safe_email }
-    full_name { Faker::Name.name }
+    sequence(:full_name) { "John Doe #{n}" }
```
peteryates, line 6:
```suggestion
    sequence(:full_name) { |n| "John Doe #{n}" }
```

cwrw: oops, addressed 😅

PR #2642, opened 2022-10-21
Add RSpec filter to avoid running API v3 specs

cwrw:
### Context
V3 API tests are excluded in our CI runs, however locally they fail unless manually excluded

- Ticket: n/a

- Add filter to config
- Add this new filter to all specs to ignore
- Remove exclusion from rspec ci now its no longer needed

### Guidance to review
When we add endpoints we will start removing those filters one by one, whilst also setting the API feature flag to true
This is based off @leoapost work in https://github.com/DFE-Digital/early-careers-framework/pull/2640

### DIFF
--- a/spec/docs/v3/statements_spec.rb
+++ b/spec/docs/v3/statements_spec.rb
@@ -2,7 +2,7 @@
 
 require "swagger_helper"
 
-RSpec.describe "API", type: :request, swagger_doc: "v3/api_spec.json" do
+RSpec.describe "API", type: :request, swagger_doc: "v3/api_spec.json", api_v3: true do
   path "/api/v3/statements" do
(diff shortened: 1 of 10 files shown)


spec/docs/v3/statements_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require "swagger_helper"
 
-RSpec.describe "API", type: :request, swagger_doc: "v3/api_spec.json" do
+RSpec.describe "API", type: :request, swagger_doc: "v3/api_spec.json", api_v3: true do
```
cwrw, line 5: we've already added some of those endpoints. Will fix/enable in another PR as specs need updating

PR #2643, opened 2022-10-21
Add methods in ParticipantIdentity to determine its origin

leoapost:
### Context
The ParticipantIdentity model has two scopes to retrieve the 
1. original identity which is the one created with the user 
2. transferred identities which can be either 
2.1. just more identities of the same user
2.2. transferred identities from other user records of the same participant, caused by the de-dupping process

The transferred identities returning two different things, confuses things and can lead to more issues. 

So, this is a follow up from my previous PR #2628 as an attempt to identify the non original identities correctly and display them in the participants admin profile for easier troubleshooting. 

ps. it is inspired by the [discussion](https://github.com/DFE-Digital/early-careers-framework/pull/2628#discussion_r1000412139) with Tony on how the origin of the identities can be potentially determined.

- Ticket: N/A

- Rename the `transferred` scope to something more inclusive
- Add method to get identify the origin of the identity
- Update the identities component to display a better label for the origin of the identity

### DIFF
--- a/app/components/admin/participants/identities.rb
+++ b/app/components/admin/participants/identities.rb
@@ -10,10 +10,12 @@ module Admin
       end
 
       def identity_transferred_label(identity)
-        if identity.user_id == identity.external_identifier
+        if identity.original_identity?
           "Original"
-        else
+        elsif identity.transferred_identity?
           "Transferred"
+        elsif identity.additional_identity?
+          "Additional"
         end
       end
     end
--- a/app/controllers/concerns/current_user.rb
+++ b/app/controllers/concerns/current_user.rb
@@ -9,7 +9,7 @@ module CurrentUser
     original = super
     return if original.nil?
 
-    transferred_identity = ParticipantIdentity.transferred.find_by(external_identifier: original.id)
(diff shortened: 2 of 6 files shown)


leoapost:
@StupidCodeFactory I see your concerns about identities with their external_identifier not matching the user id, but the sole purpose of this PR is to make it easier to troubleshoot participants from the admin pages, but also give some more context when reading the code, as currently it's misleading.

tonyheadford:
> I'm wondering if we're doing the right thing here: https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/services/identity/create.rb#L43-L47
> 
> I don't have enough knowledge around this but external identities where supposed to always reference a user_id, they would have been an `id` that was a lead provider would have had but we'd back track through the external_identifier to always know for which main use a declaration was made.

A `ParticipantIdentity` is also used for an additional email addresses but those external identifiers are not (necessarily) exposed to the LP.  There is a unique constraint on `external_identifier` to prevent duplicates which may or may not help in these scenarios.

tonyheadford (approved):
LGTM

PR #2654, opened 2022-10-24
Reduce unfound country Sentry messages from `:error` to `:warning`

peteryates:
### Context

The default [error level is :error](https://docs.sentry.io/platforms/ruby/usage/set-level/), which triggers an alert. In this case we do want to know about the missing/wrong countries because it'll show a mismatch between our countries lists in the NPQ and ECF apps, but it's not important enough to be considered an error.

The originally selected value will still be saved in `teacher_catchment_country` making the problem rows easy to identify and fix later if needed.


cwrw (approved):
Past fixing the spec: `spec/services/npq/build_application_spec.rb:87`  🙌

PR #2660, opened 2022-10-25
Improve duplicate query

cwrw:
co-authored with @StupidCodeFactory 

### Context
After manually inspecting records we found the following:
- Some induction records don't have lead providers
- Transferred participants can have induction records belonging to multiple lead providers
- Participants have multiple identities

- Ticket: n/a

Amend duplicate query to:
- Account for induction records not linked to lead providers
- Group declarations by profile only to avoid not linking to them in the UI if we have multiple lead providers
- Join on user_id rather than external id on profiles to include ones with multiple identities
- Account for transferred participants 
- Surface timestamps and induction programme types in UI


StupidCodeFactory (approved):
🚢 🇮🇹

PR #2661, opened 2022-10-25
Fix manual validation exporter

tonyheadford:
### Context

Manual validation exporter can fail if the validation data is `nil`.  This prevents it from generating an error.


peteryates (approved):
Looks good.

PR #2664, opened 2022-10-26
Explicity define cohort years in school cohorts

cwrw:
### Context
The Factory when creating a cohort can create cohorts in years 2055, which might override the `active_registration_cohort` scope.

In the flakey spec we are creating a school cohort for a participant using the current year, so when the validation check kicks in this record isn't returned as the `active_registration_cohort` is different.

Explicitly set cohort years so we don't get failures.

- Ticket: 

### Guidance to review

example of failure: https://github.com/DFE-Digital/early-careers-framework/actions/runs/3328033619/jobs/5503552423

### DIFF
--- a/spec/requests/admin/schools/cohort_2020_spec.rb
+++ b/spec/requests/admin/schools/cohort_2020_spec.rb
@@ -9,8 +9,8 @@ RSpec.describe "Admin::Schools::Cohort2020", type: :request do
   let(:cohort_2020) { create(:cohort, start_year: 2020) }
   let(:school_cohort) { create(:school_cohort, :cip, cohort: cohort_2020, school:, core_induction_programme: cip) }
   let!(:participants) { create_list(:ect_participant_profile, 5, school_cohort:) }
-  let(:other_cohort) { create(:cohort) }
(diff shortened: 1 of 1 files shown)


StupidCodeFactory (approved):

spec/requests/admin/schools/cohort_2020_spec.rb
```diff
@@ -9,8 +9,8 @@
   let(:cohort_2020) { create(:cohort, start_year: 2020) }
   let(:school_cohort) { create(:school_cohort, :cip, cohort: cohort_2020, school:, core_induction_programme: cip) }
   let!(:participants) { create_list(:ect_participant_profile, 5, school_cohort:) }
-  let(:other_cohort) { create(:cohort) }
-  let(:other_school_cohort) { create(:school_cohort, :cip, cohort: other_cohort, school: school_cohort.school, core_induction_programme: cip) }
+  let(:cohort_2021) { Cohort.find_by(start_year: 2021) || create(:cohort, :current) }
```
peteryates, line 12: I don't think we can be in a position where 2021 doesn't exist as it's explicitly created in the `initial_seed.rb`

peteryates (approved):

PR #2672, opened 2022-10-26
Ensure all node-versions are set to 16.17.x

peteryates:
### Context

We've had a couple of failures here and there where the node version hasn't matched up with the one that's on the alpine image. Hopefully this will fix it.


peteryates:
Closing, we'll jump straight to 16.18.0

PR #2674, opened 2022-10-26
[CPDLP-1850] Refactor Accept and Reject NPQ application endpoints

leandroalemao:
### Context

Refactor `accept` and `reject` NPQ application endpoints in preparation for the new API v3

- Ticket: [CPDLP-1850](https://dfedigital.atlassian.net/browse/CPDLP-1850)

Refactor API v1 & v2 `accept` and `reject` NPQ application endpoints in preparation for the new API v3 endpoint.


app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -26,19 +26,15 @@ def show
...
+        render_from_service(service, json_serializer_class)
       end
 
       def accept
-        if NPQ::Accept.call(npq_application:)
-          render json: json_serializer_class.new(npq_application).serializable_hash
-        else
-          render json: { errors: Api::ErrorFactory.new(model: npq_application).call }, status: :bad_request
-        end
+        service = AcceptNPQApplication.new(npq_application:)
```
cwrw, line 35: can we retain the `NPQ::AcceptApplication` structure, as this only relates to NPQ applications and no ECF? same for the reject, what do you think?

app/services/accept_npq_application.rb
```diff
@@ -0,0 +1,105 @@
...
+      npq_application
+    end
+  end
+
+private
+
+  def not_already_accepted
+    return if npq_application.blank?
+
+    errors.add(:npq_application, I18n.t("activerecord.errors.models.npq_application.attributes.lead_provider_approval_status.has_already_been_accepted")) if npq_application.accepted?
```
cwrw, line 31: could you move the error nesting in the locale to avoid the long path?

app/services/accept_npq_application.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  def call
+    return self unless valid?
+
+    ApplicationRecord.transaction do
+      teacher_profile.update!(trn: npq_application.teacher_reference_number) if npq_application.teacher_reference_number_verified?
+      participant_profile
+      npq_application.update(lead_provider_approval_status: "accepted") && other_applications_in_same_cohort.update(lead_provider_approval_status: "rejected")
+      deduplicate_by_trn!
+      npq_application
```
cwrw, line 22: can this move outside the transaction?

app/services/accept_npq_application.rb
```diff
@@ -0,0 +1,105 @@
...
+  def other_applications_in_same_cohort
+    @other_applications_in_same_cohort ||= NPQApplication
+      .joins(:participant_identity)
+      .where(participant_identity: { user_id: user.id })
+      .where(npq_course:)
+      .where(cohort:)
+      .where.not(id: npq_application.id)
+  end
+
+  def participant_profile
```
cwrw, line 68: could this be renamed to indicate a profile is being created?

app/services/reject_npq_application.rb
```diff
@@ -0,0 +1,34 @@
...
+    npq_application.update!(lead_provider_approval_status: "rejected")
+    npq_application
+  end
+
+private
+
+  def not_already_rejected
+    return if npq_application.blank?
+
+    errors.add(:npq_application, I18n.t("activerecord.errors.models.npq_application.attributes.lead_provider_approval_status.has_already_been_rejected")) if npq_application.rejected?
```
cwrw, line 26: same question here around error message, let me know your thoughts

cwrw, line 26: same question here around error message, let me know your thoughts

spec/services/reject_npq_application_spec.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.shared_examples "validating an application to be rejected" do
```
cwrw, line 5: do we need the shared examples as this is only for NPQ, same for other spec?

cwrw (commented):
looking good, some minor comments inline. Also could you delete the old implementation please?

app/controllers/api/v1/npq_applications_controller.rb
leandroalemao: Hi @cwrw tks for your review. Addressed in https://github.com/DFE-Digital/early-careers-framework/pull/2674/commits/8c9fde545669f063b8a66cd2c524248a8025fdb5

app/services/accept_npq_application.rb
leandroalemao: Hi @cwrw tks for your review. Addressed in https://github.com/DFE-Digital/early-careers-framework/pull/2674/commits/8c9fde545669f063b8a66cd2c524248a8025fdb5

app/services/accept_npq_application.rb
leandroalemao: Hi @cwrw tks for your review. Addressed in https://github.com/DFE-Digital/early-careers-framework/pull/2674/commits/8c9fde545669f063b8a66cd2c524248a8025fdb5

app/services/accept_npq_application.rb
leandroalemao: Hi @cwrw tks for your review. Addressed in https://github.com/DFE-Digital/early-careers-framework/pull/2674/commits/8c9fde545669f063b8a66cd2c524248a8025fdb5

app/services/reject_npq_application.rb
leandroalemao: Hi @cwrw tks for your review. Addressed in https://github.com/DFE-Digital/early-careers-framework/pull/2674/commits/8c9fde545669f063b8a66cd2c524248a8025fdb5

spec/services/reject_npq_application_spec.rb
leandroalemao: Hi @cwrw tks for your review. Addressed in https://github.com/DFE-Digital/early-careers-framework/pull/2674/commits/8c9fde545669f063b8a66cd2c524248a8025fdb5

cwrw (approved):
Nicely done, thanks @leandroalemao

app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -26,19 +26,15 @@ def show
       end
 
       def reject
-        if NPQ::Reject.call(npq_application:)
-          render json: json_serializer_class.new(npq_application).serializable_hash
-        else
-          render json: { errors: Api::ErrorFactory.new(model: npq_application).call }, status: :bad_request
-        end
+        service = NPQ::RejectApplication.new(npq_application:)
```
StupidCodeFactory, line 29: I would not have tied the action to the namespace but rather would have added classes as such:
```ruby
module NPQ
  module Application
    class Reject; end
    class Accept; end
  end
end
```

app/services/npq/accept_application.rb
```diff
@@ -0,0 +1,112 @@
...
+                                                          .where(lead_provider_approval_status: "accepted")
+                                                          .where.not(id: npq_application.id)
+    end
+
+    def other_applications_in_same_cohort
+      @other_applications_in_same_cohort ||= NPQApplication
+        .joins(:participant_identity)
+        .where(participant_identity: { user_id: user.id })
+        .where(npq_course:)
+        .where(cohort:)
```
StupidCodeFactory, line 66: Could we collapse all these `#where` together (same for the method above):
```suggestion
        .where(cohort:, npq_course:, participant_identity: { user_id: user.id })
```

app/services/npq/accept_application.rb
```diff
@@ -0,0 +1,112 @@
...
+        ParticipantProfileState.find_or_create_by(participant_profile: pp)
+      end
+    end
+
+    def participant_profile
+      @participant_profile ||= ParticipantProfile.find(npq_application.id)
+    end
+
+    def teacher_profile
+      @teacher_profile ||= user.teacher_profile || user.build_teacher_profile
```
StupidCodeFactory, line 89: ⚠️ ⚠️  Please change: Here you if the user does not have a teacher profile. 
You want to lookup teacher profile by by TRN and attach this `teacher_profile` to the user. In _theory_ there should not be two teacher profile with the same TRN. 
Another thing that needs to be though of is what if the user attach to the teacher profile is different than the one attached to the `NPQApplication` there should be some kind of error thrown as this has the potential to create duplicate profile!

@cwrw @peteryates @tonyheadford would you mind having a think about that, we've some some hacks in place to prevent double paying via the `TRN` routes but this looks like the place where duplicates could be unintentionally created?

At this point if it didn't exists you'd also want to add the TRN from `NPQApplication#teacher_reference_number` onto the `teacher_profile`

app/services/npq/accept_application.rb
```diff
@@ -0,0 +1,112 @@
...
+    def deduplicate_by_trn!
+      return if participant_profile.teacher_profile.trn.blank?
+
+      same_trn_user = TeacherProfile
+        .where(trn: participant_profile.teacher_profile.trn)
+        .where.not(id: participant_profile.teacher_profile.id)
+        .first
+        &.user
+
+      Identity::Transfer.call(from_user: participant_profile.user, to_user: same_trn_user) if same_trn_user
```
StupidCodeFactory, line 109: I've always been suspicious about this class and some potential unintended consequences. @cwrw could double check that this is ok. Lots of side effects. ⚠️

app/services/npq/reject_application.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module NPQ
+  class RejectApplication
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+    include ActiveModel::Validations::Callbacks
```
StupidCodeFactory, line 7: ✂️

app/services/npq/reject_application.rb
```diff
@@ -0,0 +1,36 @@
...
+    def call
+      return self unless valid?
+
+      npq_application.update!(lead_provider_approval_status: "rejected")
+      npq_application
+    end
+
+  private
+
+    def not_already_rejected
```
StupidCodeFactory, line 24: I think this validation is superfluous. If the Application is already rejected no need to return and error just skip the 
` npq_application.update!(lead_provider_approval_status: "rejected")`

spec/requests/api/v1/npq_applications_spec.rb
```diff
@@ -267,20 +267,16 @@
...
       it "returns error in repsonse" do
         post "/api/v1/npq-applications/#{npq_profile.id}/reject"
 
         expect(parsed_response.key?("errors")).to be_truthy
-
-        expect(parsed_response["errors"][0].key?("title")).to be_truthy
-        expect(parsed_response["errors"][0].key?("detail")).to be_truthy
-        expect(parsed_response["errors"][0]["title"]).to eql("Status cannot change from accepted")
-        expect(parsed_response["errors"][0]["detail"]).to eql("Once accepted an application cannot change state")
+        expect(parsed_response.dig("errors", 0, "detail")).to eql("Once accepted an application cannot change state")
```
StupidCodeFactory, line 279: No biggy but we're losing coverage on the `title` attribute of the response.

app/services/npq/accept_application.rb
cwrw: chatted with @StupidCodeFactory in person but for posterity: We'll check this in detail and create a ticket incase to address as it's out of scope for this PR (I'll link the ticket after investigation in the comments below)

app/services/npq/accept_application.rb
cwrw: same here, will create a ticket after investigation to tackle issues raised in that class

leandroalemao:
Hi @StupidCodeFactory you ok? I've addressed some of the comments in https://github.com/DFE-Digital/early-careers-framework/pull/2674/commits/a3e05e5f8c1aa74a42d3d7e06bf6e86cc2340f4d tkssss

PR #2679, opened 2022-10-28
Move the users query to a query service object

peteryates:
### Context

To support being able to test different querying approaches (started in #2604) we want to move queries to service objects where possible.

This will allow us to consume them from multiple places, giving us the ability to compare queries/serializers without having to add new endpoints to the API.


app/services/api/v1/users/ecf_users_query.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Api::V1::Users
+  class ECFUsersQuery
```
cwrw, line 4: maybe instead of nesting it as such `api/v1/users/ecf_users_query.rb` we could nest this query as `api/v1/ecf/users_query.rb`. Let me know your thoughts

cwrw (approved):
👍 just a question around the structure of modules/classes

app/services/api/v1/users/ecf_users_query.rb
peteryates: Good call, done.

cwrw (approved):

PR #2682, opened 2022-10-28
Add rake task for comparing query object outputs

peteryates:
### Context

**This is a WIP.**

We want to compare the results from two queries with production data. Originally the plan was to add a new endpoint but this was seen as a risk.

Instead we migrated the actual queries to service objects which can be called from anywhere. This sample rake task implements the comparison logic without needing a new API endpoint.

### Note

* this is forked from #2604 and was rebased from `develop`
* the commits here that aren't in #2604 are f7031ff56c945ee34ab1504d5 and cb1796a2d0bbb720b7

### Tasks

* [x] Remove the induction records controller


lib/tasks/compare/ecf_users_and_induction_records.rake
```diff
@@ -0,0 +1,43 @@
...
+      blue_data = serialized_users[:data].map { |u| u[:id] }
+      green_data = serialized_induction_records[:data].map { |u| u[:id] }
+
+      duplicate_records = green_data.tally.filter { |_, v| v > 1 } # assuming we're counting the number of duplications rather than duplicated records
+      same_records      = (blue_data & green_data)
+      new_records       = green_data - same_records
+      lost_records      = blue_data - same_records
+
+      if new_records.count.zero? && lost_records.count.zero?
+        puts "The API responses are the same"
```
peteryates, line 22: "The API responses are both empty"

.rubocop.yml
```diff
@@ -42,6 +42,11 @@ Rails/Output:
   Exclude:
     - "app/services/importers/*"
     - "app/services/finance/extract_report.rb"
+    - "lib/api_smoke_test.rb"
```
cwrw, line 45: ✂️

app/controllers/concerns/api_pagination.rb
```diff
@@ -12,6 +12,12 @@ def paginate(scope)
     paginated_records
   end
 
+  def paginate_array(array)
```
cwrw, line 15: ✂️

app/controllers/api/v2/ecf_induction_records_controller.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module Api
```
cwrw, line 3: ✂️

config/routes.rb
```diff
@@ -112,6 +112,7 @@
       resources :npq_enrolments, only: %i[index], path: "npq-enrolments"
       resources :users, only: %i[index create]
       resources :ecf_users, only: %i[index create], path: "ecf-users"
+      resources :ecf_induction_records, only: %i[index create], path: "ecf-induction-records"
```
cwrw, line 115: ✂️

app/serializers/api/v1/ecf_induction_record_serializer.rb
```diff
@@ -0,0 +1,76 @@
...
+      CIP_TYPES = {
+        ambition: "ambition",
+        edt: "edt",
+        teach_first: "teach_first",
+        ucl: "ucl",
+        none: "none",
+      }.freeze
+
+      set_id :id do |induction_record|
+        induction_record.preferred_identity.external_identifier
```
cwrw, line 28: we currently get the external id from the participant profile, however the email from the preferred identity

lib/tasks/compare/ecf_users_and_induction_records.rake
```diff
@@ -0,0 +1,43 @@
...
+      if new_records.count.zero? && lost_records.count.zero?
+        puts "The API responses are both empty"
+      else
+        puts "ERROR: The API responses are different"
+        puts "Blue records: #{blue_data.count}"
+        puts "Green records: #{green_data.count}"
+        puts "Duplicates found: #{duplicate_records.count}"
+        puts "Matches: #{same_records.count}"
+        puts "Changed: #{(new_records - lost_records).count}"
+        puts "New: #{(new_records - (new_records - lost_records)).count}"
```
cwrw, line 30: not sure about those two calculations for new and lost, can we just output `new_records.count` and `lost_records.count`?

lib/tasks/compare/ecf_users_and_induction_records.rake
```diff
@@ -0,0 +1,43 @@
...
+        puts "Lost: #{(lost_records - (lost_records - new_records)).count}"
+
+        new_records.each do |before|
+          id = before[:id]
```
cwrw, line 34: I think `new_records` is an array of ids and is no longer in it's serliaized so we cannot do this?

lib/tasks/compare/ecf_users_and_induction_records.rake
```diff
@@ -0,0 +1,43 @@
...
+          after = lost_records.filter { |record| record[:id] == id }.first
+          puts "\n\nExpected to find\n"
+          puts JsonDiff.diff(before, after, include_was: true).map { |el| "    #{el['path']}: \"#{el['was']}\"" }.join("\n")
```
cwrw, line 37: I wonder if we should diff the whole data set to include new and lost records rather than checking the lost records include the new record id 🤔 happy to have a chat as I think I missed something

lib/tasks/compare/ecf_users_and_induction_records.rake
```diff
@@ -10,32 +10,53 @@ namespace :compare do
...
-      same_records      = (blue_data & green_data)
-      new_records       = green_data - same_records
-      lost_records      = blue_data - same_records
+      duplicate_records = ids_from_serialized_induction_records.tally.filter { |_, v| v > 1 } # assuming we're counting the number of duplications rather than duplicated records
+      same_records      = (ids_from_serialized_users & ids_from_serialized_induction_records)
+      ir_but_not_users  = ids_from_serialized_induction_records - ids_from_serialized_users
+      users_but_not_ir  = ids_from_serialized_users - ids_from_serialized_induction_records
 
-      if new_records.count.zero? && lost_records.count.zero?
+      if ids_from_serialized_users.count.zero? && ids_from_serialized_induction_records.count.zero?
```
cwrw, line 21: Should this check `ir_but_not_users` and `users_but_not_ir` instead? As we don't want to return an error if we have records in the initial serialised responses?

cwrw (approved):
Just one comment otherwise works as expected and generates a diff file correctly 🙌

PR #2689, opened 2022-11-01
Allow sync of email when only 1 identity

tonyheadford:
### Context

Recently the sync between `User.email` and the original `ParticipantIdentity.email` was removed to facilitate easier changes to emails typically for CIP users when multiple identities are in play.  However this is causing issues when the `User` only has a single `ParticipantIdentity` and there is an expectation that they will be in sync.

Reintroduce the sync when there is only a single identity

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -36,6 +36,7 @@ class User < ApplicationRecord
   has_many :npq_application_exports
 
   before_validation :strip_whitespace
+  after_update :sync_email_with_identity
 
   validates :full_name, presence: true
   validates :email, presence: true, uniqueness: true, notify_email: true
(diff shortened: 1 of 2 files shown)


peteryates (approved):
Looks good.

This kind of change makes me think that we should replace the `users.email` column with a `users.default_participant_identity_id` and delegate `User#email` to `ParticipantIdentity`.

tonyheadford:
> Looks good.
> 
> This kind of change makes me think that we should replace the `users.email` column with a `users.default_participant_identity_id` and delegate `User#email` to `ParticipantIdentity`.

Yes something like that would be good. I imagine that we might need to look at how Devise is configured maybe and also add `ParticipantIdentity` records to non-participants first though.

PR #2694, opened 2022-11-01
[CPDLP-1866] Fix inconsistent email usage detection in users API

leandroalemao:
### Context

Fix inconsistent email usage detection in users API

- Ticket: [CPDLP-1866](https://dfedigital.atlassian.net/browse/CPDLP-1866)

There is a disconnect between the user lookup and user creation API endpoints. When attempting to sync a user from NPQ to the CPD database an “Email already in use” error occurred during user creation after a user lookup via that email returned no user. 

This is happening because GET /api/v1/users (Api::V1::UsersController#index) is using .where(email:) to find the relevant users instead of looking up via Identity. This is causing issues when a user is looked up via an email that isn’t in the direct email field on the User record but is attached to it elsewhere.

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -50,8 +50,8 @@ module Api
       def users
         users = User.all
         users = users.changed_since(updated_since) if updated_since.present?
-        users = users.where(email:) if email.present?
-        users
+        users = users.joins(:participant_identities).where(participant_identities: { email: }).or(users.where(email:)) if email.present?
+        users.distinct
       end
     end
   end
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -48,10 +48,7 @@ module Api
       end
 
       def users
-        users = User.all
-        users = users.changed_since(updated_since) if updated_since.present?
-        users = users.joins(:participant_identities).where(participant_identities: { email: }).or(users.where(email:)) if email.present?
-        users.distinct
+        Api::V1::UsersQuery.new(updated_since:, email:).all
       end
     end
   end
--- a/spec/requests/api/v1/users_spec.rb
+++ b/spec/requests/api/v1/users_spec.rb
@@ -81,6 +81,37 @@ RSpec.describe "API Users", :with_default_schedules, type: :request do
           expect(parsed_response["data"].size).to eql(0)
         end
       end
+
+      context "when filtering by identity record email" do
+        let(:user) { create(:user, email: "fred@example.com") }
+        let!(:identity) { create(:participant_identity, user:, email: "charlie@example.com") }
+
+        context "when a matching identity record exists" do
+          it "returns the associated user record" do
+            get "/api/v1/users", params: { filter: { email: "charlie@example.com" } }
+
+            expect(parsed_response["data"].size).to eql(1)
+            expect(parsed_response.dig("data", 0, "attributes", "email")).to eql("fred@example.com")
+          end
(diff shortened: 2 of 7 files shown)


peteryates (approved):
Looks great to me. Would be great to see this moved to a query object ([like this](https://github.com/DFE-Digital/early-careers-framework/pull/2679)) at some point.

cwrw (approved):
👍 +1 for query service

spec/requests/api/v1/users_spec.rb
```diff
@@ -81,6 +81,37 @@
           expect(parsed_response["data"].size).to eql(0)
         end
       end
+
+      context "when filtering by identity record email" do
+        let(:user) { create(:user, email: "fred@example.com") }
+        let!(:identity) { create(:participant_identity, user:, email: "charlie@example.com") }
+
+        context "when a matching identity record exists" do
```
StupidCodeFactory, line 89: could we add a test case where the user has no identity attached to it and verify that it still returns this user (or maybe it should actually not since this user may not be in a valid state?

you'll find a none negligible number of users without participant_identities:
`User.left_outer_joins(:participant_identities).where(participant_identities: {id: nil}).count`

and also recently created:
`ser.left_outer_joins(:participant_identities).where(participant_identities: {id: nil}).order(created_at: :desc).first.created_at` == `today`

StupidCodeFactory: What are the consequences of returning such user to NPQ registration service? @Lockyy @peteryates

app/controllers/api/v1/users_controller.rb
```diff
@@ -50,8 +50,8 @@ def email
       def users
         users = User.all
         users = users.changed_since(updated_since) if updated_since.present?
-        users = users.where(email:) if email.present?
-        users
+        users = users.joins(:participant_identities).where(participant_identities: { email: }).or(users.where(email:)) if email.present?
```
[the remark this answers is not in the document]
leandroalemao: Hi @StupidCodeFactory tks for your review. I've just moved the logic into a query service and also added the left join instead.. so now we're correctly returning the user even when it does not have a `participant_identity`. 👍 😄

spec/requests/api/v1/users_spec.rb
leandroalemao: Test case added. tksss

app/controllers/api/v1/users_controller.rb
```diff
@@ -48,10 +48,7 @@ def email
       end
 
       def users
-        users = User.all
-        users = users.changed_since(updated_since) if updated_since.present?
-        users = users.where(email:) if email.present?
-        users
+        Api::V1::UsersQuery.new(updated_since:, email:).all
```
peteryates, line 51: 👍🏽

spec/requests/api/v1/users_spec.rb
peteryates: I think a `User` can be valid if there are no identities attached because `User` is directly linked to some other kinds of profile (`AppropriateBodyProfile`, `TeacherProfile`). The tests look good like they're providing good coverage though.

Lockyy: It would also lead to further issues if we didn't return a user because then we'd be in a similar situation where a user record is blocking record creation but isn't surfaced by the lookup.

Lockyy (approved):
This looks like it'll solve the issues the NPQ app has had, will need to verify once deployed that there are no further issues for users in this situation but this is great. Thanks for resolving this.

PR #2697, opened 2022-11-01
Use method stubs to avoid relying on Postgres `Date(now())`

cwrw:
### Context
When rendering the finance statements page we rely on this scope:
https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/models/finance/statement.rb#L34
Which uses a Postgres `DATE(NOW())` timestamp. This supersedes time traveling in the specs and returns the latest statement according the time now.

We extract the cohort param from the latest statement, which meant we were defaulting to 2022. This means we didn't find our `November 2021` with that cohort which is why the scenarios fail

- Ticket: n/a

Use method stubs for the scope?

### Guidance to review
Rethink tomorrow

### DIFF
--- a/spec/support/features/steps/changes_of_circumstance_steps.rb
+++ b/spec/support/features/steps/changes_of_circumstance_steps.rb
@@ -457,6 +457,8 @@ module Steps
     end
 
     def then_the_finance_portal_shows_the_lead_provider_payment_breakdown(lead_provider_name, statement_name, total_ects, total_mentors, started, retained, completed, voided, uplift: true)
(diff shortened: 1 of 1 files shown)


StupidCodeFactory:
@cwrw What about using `Time.zone.now` in the scope?:

https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/models/finance/statement.rb#L34

```ruby
 scope :with_future_deadline_date, -> { where("deadline_date >= ?)", Time,zone.now.beginning_of_day) }

```

StupidCodeFactory (approved):

PR #2701, opened 2022-11-02
Tidy up the statement date scopes

peteryates:
### Context and changes

We ran into a few date-related problems where time travel in the tests was expected but the scopes which used PostgreSQL's `date()` function were unaware. This is a quick attempt at tidying and standardising them, plus covering with specs.

- Remove empty Finance::ECFPaymentsHelper
- Use ActiveRecord range syntax for date queries

### DIFF
--- a/app/helpers/finance/ecf_payments_helper.rb
+++ b/app/helpers/finance/ecf_payments_helper.rb
@@ -1,6 +1,0 @@
-# frozen_string_literal: true
-
-module Finance
-  module ECFPaymentsHelper
-  end
-end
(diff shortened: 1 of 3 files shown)


cwrw (approved):
Thanks @peteryates, only comment around this is that I found it hard without the tests and checking locally to parse the ranges, but I think since there are tests showing that 👍

peteryates:
@cwrw yeah I agree. Even when writing the scopes I had to think about `..` vs `...`. I pushed another commit that switches those bits to Arel, which with `lt`/`lteq`/`gt`/`gteq` is more readable.

PR #2702, opened 2022-11-03
Add new Deleted Duplicate table to store serialised profiles

cwrw:
### Context

To make sure we have a record of deleted duplicate participant profiles, add a new table which stores a serialised record of the deleted profile and any related records. Also a link to the primary participant profile.

- Ticket: n/a

- Add new table for deleted duplicates
- Add new serialiser for profiles to be deleted, including all relevant attributes of related models
- 
### Guidance to review
Did I miss anything?

### DIFF
--- a/app/serializers/finance/ecf/duplicate_serializer.rb
+++ b/app/serializers/finance/ecf/duplicate_serializer.rb
@@ -1,0 +1,175 @@
+# frozen_string_literal: true
+
+module Finance
+  module ECF
+    class DuplicateSerializer
+      include JSONAPI::Serializer
+
+      set_id :id
+      set_type :duplicate
+
+      attribute :type, &:type
+      attribute :sparsity_uplift, &:sparsity_uplift
+      attribute :pupil_premium_uplift, &:pupil_premium_uplift
+      attribute :status, &:status
+      attribute :school_urn, &:school_urn
+      attribute :school_ukprn, &:school_ukprn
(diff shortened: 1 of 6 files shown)


app/serializers/finance/ecf/duplicate_serializer.rb
```diff
@@ -0,0 +1,175 @@
...
+            reason: participant_profile_state.reason,
+            state: participant_profile_state.state,
+            created_at: participant_profile_state.created_at.rfc3339,
+            updated_at: participant_profile_state.updated_at.rfc3339,
+          }
+        end
+      end
+
+      attribute(:participant_validation_data) do |object|
+        participant_profile_state = object.ecf_participant_validation_data
```
tonyheadford, line 116: Should that variable be named `participant_validation_data` ?

tonyheadford (commented):
Think that looks good. It might also be really useful to have a simple service that could create a the `DeletedDuplicate` given 2 participant profiles?

cwrw:
@tonyheadford That's an excellent point regarding adding a service for `DeletedDuplicate`. However I was hoping to go through the first batch of deletions to get a sense of what we expect and how things look like especially since we might need to merge things, after that happy to open a new pr for that service which accepts a primary and list of duplicates to delete, and serializes the deleted profiles.

StupidCodeFactory (approved):

PR #2706, opened 2022-11-04
Move json-diff gem out of development/test group

peteryates:
### Context and changes

Also make it `require: false` and explicitly require it in the comparison script as that's the only place it'll be used for the time being.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -105,11 +105,11 @@ gem "net-imap", require: false
 gem "net-pop", require: false
 gem "net-smtp", require: false
 
+gem "json-diff", "~> 0.4.1", require: false
+
 group :development, :test do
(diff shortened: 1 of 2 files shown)


cwrw (approved):

Gemfile
```diff
@@ -105,12 +105,12 @@ gem "net-imap", require: false
 gem "net-pop", require: false
 gem "net-smtp", require: false
 
+gem "json-diff", "~> 0.4.1", require: false
```
cwrw, line 108: do we need  the `"~> 0.4.1"`?

peteryates: Just checked the gem and this release is from 2017 with zero activity on the GitHub repo since 🙈 

As it's just used in a rake task I don't see it being a problem.

PR #2707, opened 2022-11-04
Export path in Dockerfile to avoid having to type full path

cwrw:
### Context
We currently have to include the full path to run /usr/local/bin/bundle exec when sshing into the container

- Ticket: n/a

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -94,7 +94,10 @@ COPY --from=assets-precompile /app /app
 COPY --from=assets-precompile /usr/local/bundle/ /usr/local/bundle/
 COPY --from=middleman /public/ /app/public/
 
-RUN echo "cd /app && /usr/local/bin/bundle exec rails c" > /root/.ash_history
+RUN echo export PATH=/usr/local/bin:\$PATH > /root/.ashrc
+ENV ENV="/root/.ashrc"
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -94,7 +94,10 @@ COPY --from=assets-precompile /app /app
 COPY --from=assets-precompile /usr/local/bundle/ /usr/local/bundle/
 COPY --from=middleman /public/ /app/public/
 
-RUN echo "cd /app && /usr/local/bin/bundle exec rails c" > /root/.ash_history
+RUN echo export PATH=/usr/local/bin:\$PATH > /root/.ashrc
```
StupidCodeFactory, line 97:
```suggestion
RUN echo export PATH=/usr/local/bin:/usr/local/bundle/bin:\$PATH > /root/.ashrc
```

StupidCodeFactory (approved):

Dockerfile
StupidCodeFactory: Could also add an `export BUNDLE_GEMFILE=/app/Gemfile` somewhere for extra safety

cwrw: discussed in person: will move to that in the next step for now going to try this initially

peteryates (approved):
Definitely an improvement 👍🏽

PR #2708, opened 2022-11-04
Fix flakey change schedule spec

cwrw:
### Context
Explicitly set participant profile schedule to avoid validation failures if they are the same with the new schedule
- Ticket:


leandroalemao (approved):
👍 tkss

PR #2709, opened 2022-11-07
Show 'No delivery partner' in email when missing

peteryates:
### Changes and context

Currently when sending a fip_provider_has_withdrawn_a_participant email we try to access the partnership's delivery partner. This is an optional value so not all partnerships are guaranteed to have one.

Currently it fails when one is missing, here I'm proposing that we instead display a string of text that says 'No delivery partner' instead.

### DIFF
--- a/app/mailers/school_mailer.rb
+++ b/app/mailers/school_mailer.rb
@@ -446,7 +446,7 @@ class SchoolMailer < ApplicationMailer
         name: induction_coordinator.user.full_name,
         withdrawn_participant_name: withdrawn_participant.user.full_name,
         school: withdrawn_participant.school.name,
-        delivery_partner: partnership.delivery_partner.name,
+        delivery_partner: partnership&.delivery_partner&.name || "No delivery partner",
         lead_provider: partnership.lead_provider.name,
       },
     )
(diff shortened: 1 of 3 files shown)


app/mailers/school_mailer.rb
```diff
@@ -446,7 +446,7 @@ def fip_provider_has_withdrawn_a_participant(withdrawn_participant:, induction_c
         name: induction_coordinator.user.full_name,
         withdrawn_participant_name: withdrawn_participant.user.full_name,
         school: withdrawn_participant.school.name,
-        delivery_partner: partnership.delivery_partner.name,
+        delivery_partner: partnership&.delivery_partner&.name || "No delivery partner",
```
ltello, line 449: `Partnership` already has:
`delegate :name, to: :lead_provider, allow_nil: true, prefix: true`

We might add similar delegation method for deliver_partner so this line could end up being:
`delivery_partner: partnership.delivery_partner_name || "No delivery partner",`

ltello (approved):

peteryates:
We agreed on these changes in Slack. The email template has been tweaked so there's a 'key' describing the lead provider and delivery partner values:

image

PR #2721, opened 2022-11-08
Rejig the analytics queries for multiple cohorts

peteryates:
### Context

Written under Nathan's supervision! 😅

Main changes:

* we split up the complicated schools query by removing  the induction tutor specific bits
* added a cohort identifier to the partnerships table
* made the schools query combine data from the 2021/2022 cohorts and display them side-by-side

### Notes for review

These scripts aren't used by the app in any way, they power an ad hoc export that's used for reporting purposes.

### DIFF
--- a/analytics_queries/induction_tutors.sql
+++ b/analytics_queries/induction_tutors.sql
@@ -6,7 +6,7 @@
           u.full_name                        AS tutor_name,
           u.email                            AS tutor_email,
           (pp.id IS NOT NULL)                AS sit_mentor,
-          u.id                               AS sit_mentor_id
+          pi.external_identifier             AS sit_mentor_id
 
     FROM schools s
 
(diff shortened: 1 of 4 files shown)


analytics_queries/induction_tutors.sql
```diff
@@ -0,0 +1,21 @@
+\copy (
+  SELECT s.urn,
+          (icp.id IS NOT NULL)               AS tutor_nominated,
+          icp.created_at                     AS tutor_nominated_at,
+          (u.current_sign_in_at IS NOT NULL) AS tutor_signed_in,
+          u.full_name                        AS tutor_name,
+          u.email                            AS tutor_email,
+          (pp.id IS NOT NULL)                AS sit_mentor,
+          u.id                               AS sit_mentor_id
```
tonyheadford, line 9: Just wondering if this needs to be the external identifier from the participant identity or not here?

tonyheadford (approved):
Just a non-blocking question regarding `user.id` vs `participant_identity.external_identifier`

analytics_queries/induction_tutors.sql
peteryates: Good call, Nathan agrees. I'll switch this over before merging.

peteryates: @tonyheadford - ok switched this over, would you mind taking another quick look?

tonyheadford (approved):
LGTM

PR #2724, opened 2022-11-09
[CPDLP-1893] Add a new column to NPQ statements CSV report

leandroalemao:
### Context

We are adding targeted delivery funding to our NPQ statement, we need to update the query when downloading CSV to include the relevant fields.

- Ticket: [CPDLP-1893](https://dfedigital.atlassian.net/browse/CPDLP-1893)

Add a new column (Targeted Delivery Funding) to NPQ statements CSV report.

### DIFF
--- a/app/serializers/finance/npq/assurance_report/csv_serializer.rb
+++ b/app/serializers/finance/npq/assurance_report/csv_serializer.rb
@@ -23,6 +23,7 @@ module Finance
           "Declaration Created At",
           "Statement Name",
           "Statement ID",
+          "Targeted Delivery Funding",
         ].freeze
 
         def initialize(scope, statement)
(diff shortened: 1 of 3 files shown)


app/serializers/finance/npq/assurance_report/csv_serializer.rb
```diff
@@ -68,6 +69,7 @@ def to_row(record)
             record.declaration_created_at.iso8601,
             record.statement_name,
             record.statement_id,
+            record.targeted_delivery_funding_eligibility,
```
cwrw, line 72: I think you'll also need to add:
```
a. targeted_delivery_funding_eligibility
```
here as well: https://github.com/DFE-Digital/early-careers-framework/blob/develop/app/services/finance/npq/assurance_report/query.rb#L28

cwrw (approved):
👍

PR #2725, opened 2022-11-09
Remove unnecessary creation of NPQ default schedule

cwrw:
### Context
With the `with_default_schedules` we create all necessary schedules. When creating the extra schedule the test is flakey as there are two similar default schedules and we can return either one, which fails on equality expectation

- Ticket: n/a

Remove creation of default as its done here: https://github.com/DFE-Digital/early-careers-framework/blob/develop/spec/support/with_default_schedules.rb

### Guidance to review

example failure: https://github.com/DFE-Digital/early-careers-framework/actions/runs/3419197516/jobs/5692417259


leandroalemao (approved):
LGTM 👍 😄 tks

PR #2736, opened 2022-11-14
Update babel, shakapacker

peteryates:
### Context

This is a bit of investigation in to why the babel/preset-env Dependabot PR isn't building.

To make this build succeed I had to explicitly depend on `@babel/plugin-proposal-object-rest-spread`. Without it (even when it's an indirect dependency) the build wouldn't pass.

Fixes #2729, #2719


yarn.lock
```diff
@@ -915,6 +1011,87 @@
     core-js-compat "^3.25.1"
     semver "^6.3.0"
 
+"@babel/preset-env@^7.20.2":
```
cwrw, line 1014: The diff from https://github.com/DFE-Digital/early-careers-framework/pull/2719/files is quite different than here for babel, so not sure why though 🤔 ?

cwrw (approved):
thanks @peteryates 🙌

PR #2740, opened 2022-11-15
Pull out the differences in endpoint service responses for analysis

mattonfoot:
### Context

Generate useful reports for analysis due to 10K records getting new external IDs and so comparison has been made difficult. need to be able to examine the records to work out how they can be eliminated as not requiring approval.

- Ticket: https://dfedigital.atlassian.net/browse/CST-616

write records that are not present in both service responses or have differences so we can analyse them.


peteryates (approved):
Looks good.

Using a file per category makes way more sense 😃

PR #2744, opened 2022-11-15
Clean up participant actions and services

cwrw:
### Context
Work for moving participant actions into services has been completed for all endpoints. Cleanup the rest of the unused code
- Ticket: n/a

- Remove unused participant directory in services
- Tidy up participant actions and remove extra methods
- Move logic for validation out of controller into service class

### Guidance to review
Need to test that we are using serialisers from v1/v2 correctly in the API with new structure


leandroalemao (approved):
LGTM.. great stuff. 👍 😄 

The only thing I can see is some broken references in the seed file below but don't know if we'll ever use it again or not.. looks like it was a one-off task/seeds importer done specifically to BPN.

https://github.com/DFE-Digital/early-careers-framework/blob/5708067577d728c0157b12b9157e21c8a76453e0/app/services/importers/seed_bpn_declarations.rb#L94

app/services/withdraw_participant.rb
```diff
@@ -61,6 +62,21 @@ def not_already_withdrawn
...
+    errors.add(:participant_profile, I18n.t(:no_started_declaration_found)) unless any_participant_declarations_started?
+  end
+
+  def any_participant_declarations_started?
+    participant_profile
+      .participant_declarations
+      .where(
+        course_identifier:,
+        declaration_type: "started",
+      ).any?
```
peteryates, line 77: Would `#exists?` make sense here?

```suggestion
      ).exists?
```

peteryates (approved):
Massive improvement 👌🏽 

Left a minor suggestion.

cwrw:
Thanks all for the review, addressed all comments if you'd like to have another look

peteryates (approved):
🚢

PR #2746, opened 2022-11-15
Attempt at finding correct record in both responses

mattonfoot:
The ECF User query and the IR Query use different UUIDs for the id of the records. It is chance that ids match

This now uses a map of user_id to external_identifier to find matched records which should mean we see actual differences.

### DIFF
--- a/lib/tasks/compare/ecf_users_and_induction_records.rake
+++ b/lib/tasks/compare/ecf_users_and_induction_records.rake
@@ -44,10 +44,9 @@ namespace :compare do
             # skip if the attributes are the same
             next
           else
-            original_attributes = JsonDiff.diff(record_from_users_query[:attributes], record_from_ir_query[:attributes], include_was: true).map do |el|
-              attributes = {}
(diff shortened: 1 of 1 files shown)


peteryates (approved):
Looks fine, runs locally without errors.

lib/tasks/compare/ecf_users_and_induction_records.rake
```diff
@@ -6,67 +6,73 @@ namespace :compare do
...
-            diff_report += "--------------------\n"
-            diff_report += "#{JSON.pretty_generate(record_from_ir_query)}\n"
+            original_attributes = {}
+            JsonDiff.diff(record_from_users_query[:attributes], record_from_ir_query[:attributes], include_was: true).each do |el|
+              original_attributes[el["path"].sub("/", "")] = el["was"]
+            end
+
+            diff_report.push user_id: records["id"],
+                             external_identifier: records["external_identifier"],
+                             old_record: { attributes: original_attributes },
```
peteryates, line 54: is `old_record` the best description here? Aren't these the values that have been modified?

PR #2753, opened 2022-11-17
Support ECT user endpoint reports

mattonfoot:
Small change to ensure we return the same id in the API as Support for ECTs uses that to find the record for a user and if we did not then some people would get emailed another invite but some would not get one at all.


peteryates (approved):
Looks good

PR #2754, opened 2022-11-18
Double DB statement timeout

cwrw:
### Context

Our queries are performing really badly, while we improve performance bump timeout as it's already too low.

- Ticket: n/a

Double timeout to 20s

### Guidance to review
We'll monitor once deployed to see what impact this has on the DB and frequency of timeouts.


peteryates (approved):
👍

PR #2755, opened 2022-11-18
Create outcome model

peteglondon:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1897

- Add `ParticipantDeclarationOutcome` model and table
- ParticipantDeclaration `has_many: :outcomes`
- Basic model-level validation

### Not covered by this PR
- `Sits under the correct model hierarchy (possibly NPQ)` - we don't seem to have a model hierarchy. This should be the subject of a bigger conversation around DDD
- `Include any required scopes on a participant profile and declarations which will allow retrieving all outcomes for a participant, or the latest outcome from the latest completed declaration (there are voided or clawed back declarations` - best to create the scopes when we have a concrete use case for them.


app/models/participant_declaration.rb
```diff
@@ -25,6 +25,8 @@ class ParticipantDeclaration < ApplicationRecord
            through: :billable_statement_line_items,
            source: :statement
 
+  has_many :outcomes, class_name: "ParticipantDeclarationOutcome"
```
cwrw, line 28: Outcomes are always going to be for NPQ declarations, is there a way we can namespace this to ensure that's evident?

db/schema.rb
```diff
@@ -639,7 +648,7 @@
     t.boolean "sparsity_uplift"
     t.boolean "pupil_premium_uplift"
     t.uuid "delivery_partner_id"
-    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY ((ARRAY['submitted'::character varying, 'eligible'::character varying, 'payable'::character varying, 'paid'::character varying, 'clawed_back'::character varying])::text[]))"
+    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY (ARRAY[('submitted'::character varying)::text, ('eligible'::character varying)::text, ('payable'::character varying)::text, ('paid'::character varying)::text, ('clawed_back'::character varying)::text]))"
```
cwrw, line 651: Can we avoid committing the extra schema changes past the new table (here and below)?

app/models/participant_declaration.rb
peteglondon: - moved it to `ParticipantDeclaration::Outcome::NPQ` (https://github.com/DFE-Digital/early-careers-framework/pull/2755/commits/95eaef1ea5c82c20e3500b50f90e3b5c105fdbb5 and https://github.com/DFE-Digital/early-careers-framework/pull/2755/commits/04c5b2128d33f93e8c250cf29db7bd320ff6647a)
- moved the `has_many` to `ParticipantDeclaration::NPQ` (https://github.com/DFE-Digital/early-careers-framework/pull/2755/commits/830d23e254ade372f9137296e22fc01d00e099ef)
- specify `ParticipantDeclaration::NPQ` for the `belongs_to` (https://github.com/DFE-Digital/early-careers-framework/pull/2755/commits/8a9330b7454e8c1680a3ef6b6e56d99caafe525f)

app/models/participant_declaration/npq.rb
```diff
@@ -7,6 +7,8 @@ class ParticipantDeclaration::NPQ < ParticipantDeclaration
 
   has_many :statements, class_name: "Finance::Statement::NPQ", through: :statement_line_items
 
+  has_many :outcomes, class_name: "ParticipantDeclaration::Outcome::NPQ", foreign_key: "participant_declaration_id"
```
cwrw, line 10: In terms of what outcome means, I think it's related to a participant as a whole passing rather than to a declaration which is just to mark that the participant has completed the course. So maybe our model should be named "ParticipantOutcome::NPQ"? Same with renaming the table?

app/models/participant_declaration/outcome/npq.rb
```diff
@@ -0,0 +1,23 @@
...
+  validates :state, presence: true
+  validates :completion_date, presence: true
+  validate :completion_is_not_in_future, if: :completion_date
+
+private
+
+  def completion_is_not_in_future
+    return if completion_date <= Time.zone.today
+
+    errors.add(:completion_date, "Cannot be in the future")
```
cwrw, line 21: could you add the error message to the locale file `en.yml`?

spec/models/participant_declaration/npq_spec.rb
```diff
@@ -8,6 +8,16 @@
       create_list(:npq_participant_declaration, 20)
     end
 
+    describe "associations" do
+      let(:user) { create(:user) }
```
cwrw, line 12: could we add a factory for outcomes?

app/models/participant_declaration/npq.rb
peteglondon: Renamed https://github.com/DFE-Digital/early-careers-framework/pull/2755/commits/92d5dcb8c90931ec2fa37b69128569941b25a60c
Renamed table https://github.com/DFE-Digital/early-careers-framework/pull/2755/commits/5f08a9843854bd6dde4aed43d52a3466d20281ad

spec/models/participant_declaration/npq_spec.rb
peteglondon: Done https://github.com/DFE-Digital/early-careers-framework/pull/2755/commits/85b96f3a568339e19b5dfb04c6c212d45fa737e1

app/models/participant_declaration/outcome/npq.rb
peteglondon: done https://github.com/DFE-Digital/early-careers-framework/pull/2755/commits/64e8d11ae6b578daeadb0a65e10dc4a55bb9d34f

db/migrate/20221121095445_rename_participant_declaration_outcomes.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class RenameParticipantDeclarationOutcomes < ActiveRecord::Migration[6.1]
+  def change
+    safety_assured do
+      rename_table :participant_declaration_outcomes, :participant_outcomes
```
cwrw, line 6: Personal preference/nit: since this is a new table and the first time we're introducing it prefer to only have the original migration with the correct name rather than a rename, but reckon you'll have to close/open the PR as this will upset the review app

cwrw (approved):
thanks for addressing all the comments. Left one final optional comment otherwise 👍

db/migrate/20221121095445_rename_participant_declaration_outcomes.rb
peteglondon: I would prefer that too but the rebase in this case would be horrible

PR #2757, opened 2022-11-18
Minor changes to Support ECTs User Query by Induction Record

mattonfoot:
Changes a few selection rules:

  - group by clause - so it is participant_profiles not preferred_identities
  - ordering - as current API does not filter out leaving and withdrawn
  - fallback to participant_profile to try and get core_induction_programme name if it is not copied to the induction_record

### DIFF
--- a/app/services/api/v1/ecf/induction_records_query.rb
+++ b/app/services/api/v1/ecf/induction_records_query.rb
@@ -10,10 +10,7 @@ module Api::V1::ECF
     end
 
     def all
-      induction_records = InductionRecord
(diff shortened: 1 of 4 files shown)


app/services/api/v1/ecf/induction_records_query.rb
```diff
@@ -10,23 +10,43 @@ def initialize(updated_since: nil, email: nil)
...
+        induction_records = induction_records.joins(:preferred_identity).where(preferred_identity: { email: })
       end
 
       induction_records
-        .group_by(&:preferred_identity_id)
-        .transform_values(&:first)
-        .values
+    end
+
+    def relevant_induction_records
```
tonyheadford, line 26: Could we include a comment about what this does and why it is different ie. not just `current` records any more?

tonyheadford (approved):
Looks good, I left one comment that I think would be helpful, but non-blocking

PR #2766, opened 2022-11-23
Add direct route for new form url and update link in views

tonyheadford:
### Context

The old feedback URL was stored in Microsoft land somewhere and the user who owned the account had left. As a temporary measure a simpler Google form has been put together by our UR team and this PR updates the URL to link to the new form.

Move URL into the routes
Replace hard-coded URL in views with Rails URL helper `feedback_form_url`

### Guidance to review

Feedback link in header banner and lead provider content page now points to new form URL

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,6 +15,10 @@ Rails.application.routes.draw do
     get "/users/link-invalid", to: "users/sessions#link_invalid"
   end
 
+  direct :feedback_form do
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -15,6 +15,10 @@
     get "/users/link-invalid", to: "users/sessions#link_invalid"
   end
 
+  direct :feedback_form do
```
peteryates, line 18: TIL about `direct`! 😬

peteryates (approved):

PR #2773, opened 2022-11-27
Seeding via factories

peteryates:
### Context

This is an experiment in solving our problem of seeding environments with realistic-looking data.

### Background

Having done some experimentation in building objects and scenarios without using factories, it's clear that we were just re-implementing much of what's provided by FactoryBot.

Our own use of FactoryBot in the specs is problematic because it's overcomplicated. Many factories build related objects which build other related objects - it quickly gets out of hand and it's hard to debug.

####  🏭  Factories

* new factories with some ground rules:
  * no associations by default (require traits to create other objects)
  * no `after` callbacks (other than for logging)
* prefixed with `seed_`
* placed in `spec/factories/seeds`

#### 🌱 Seeds 

* each _scenario_ is built in its own file
* scenarios sequentially build records using factories
* each scenario should be wrapped in a transaction
* scenarios should be simple plain Ruby

### Next steps

1. Build some more scenarios
2. Investigate taking the common parts of the scenarios we make and abstracting them into modules (so we might be able to call something like `build_school(early_career_teachers: 3, mentors: 1, cohort: 2021)`
3. Add factories for models as we need them

### Future steps

These aren't decided but this is the path I think we should take.

1. Start using the `seed_` factories in tests
2. Decommission the legacy factories one by one
3. Drop the `seed_` prefix and make replace the old factories completely

### Other ideas

* As we will be able to generate custom data on the fly we might want to add some controls to the admin panel (in non-production environments!) to generate trigger them. This might be useful on a review app where we want 20k schools to test a new search change.

### Guidance to test this locally

I run this a lot while developing. To keep an eye on what's going on I tile two terminal windows side by side, one to run the commands and one to view logs.

Re-seed with:

```
bundle exec rails db:seed:replant
```

Earlier on we also had to use `NEW_SEEDS=true` but now the default is flipped so to get the legacy seeds instead prefix with `LEGACY_SEEDS=true`.

Pull useful messages from the logs (replace `ag` with `rg` or `grep` as needed):

```
tail -f log/development.log | ag "(seeded|seeding|importing|adding)"
```

### Guidance to review

Don't be alarmed by the "Files changed: 140", here's a summary of what's changed:

* renamed old seeds directory from `db/seeds` to `db/legacy_seeds`
* changed some references to seeds we'd made from specs 🤯  fbe935c6763a24cdebf7c
* created new factories for seeding in `spec/factories/seeds` (see [the readme](https://github.com/DFE-Digital/early-careers-framework/blob/seeding-via-factories/spec/factories/seeds/README.md) for the rules they stick to)
* wrote tests to make sure the new factories stick to the rules in `spec/seeds/seed_factories`
* added new seeds in `db/new_seeds`
* added a 'run' file that determines which seeds to run and runs them in `db/new_seeds/run.rb`


spec/factories/seeds/participant_profile_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+      user { build(:seed_user) }
+    end
+
+    trait :with_teacher_profile do
+      teacher_profile { build(:seed_teacher_profile) }
+    end
+
+    trait :with_participant_identity do
+      participant_identity { build(:seed_participant_identity) }
+    end
```
tonyheadford, line 21: Sort of feels wrong to have this in here as it's not directly associated with a `:teacher_profile`?

spec/factories/seeds/teacher_profile_factory.rb
```diff
@@ -0,0 +1,23 @@
...
+  factory(:seed_teacher_profile, class: "TeacherProfile") do
+    trn { Faker::Number.unique.rand_in_range(10_000, 100_000).to_s }
+
+    trait :with_school do
+      school { build(:seed_school) }
+    end
+
+    trait :with_participant_identity do
+      participant_identity { build(:seed_participant_identity) }
+    end
```
tonyheadford, line 13: Sort of feels wrong to have this in here as it's not directly associated with a :teacher_profile?

peteryates: I think that might've been some over enthusiastic copy/pasting!

spec/factories/seeds/user_factory.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory(:seed_user, class: "User") do
+    email { Faker::Internet.unique.safe_email }
+    full_name { Faker::Name.name }
```
cwrw, line 6: Just a note for when we move those back to tests to make sure we don't have names like `O'niel` which makes tests flakey, maybe we can fix that from now?

spec/factories/seeds/teacher_profile_factory.rb
```diff
@@ -0,0 +1,23 @@
...
+    trait :with_school do
+      school { build(:seed_school) }
+    end
+
+    trait :with_mentor_profile do
+      mentor_profile { build(:seed_mentor_profile) }
+    end
+
+    trait :with_user do
+      user { build(:seed_user) }
```
cwrw, line 16: user is required on a teacher profile, could we move this outside the trait?

spec/factories/seeds/teacher_profile_factory.rb
```diff
@@ -0,0 +1,23 @@
...
+
+FactoryBot.define do
+  factory(:seed_teacher_profile, class: "TeacherProfile") do
+    trn { Faker::Number.unique.rand_in_range(10_000, 100_000).to_s }
+
+    trait :with_school do
+      school { build(:seed_school) }
+    end
+
+    trait :with_mentor_profile do
```
cwrw, line 11: ✂️

spec/factories/seeds/participant_identity_factory.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory(:seed_participant_identity, class: "ParticipantIdentity") do
+    email { Faker::Internet.unique.email }
+    external_identifier { SecureRandom.uuid }
+
+    trait :with_user do
+      association(:user, factory: :seed_user)
```
cwrw, line 9: I think user is required here as well, so might need to move it outside the trait

spec/factories/seeds/participant_profile_factory.rb
```diff
@@ -0,0 +1,25 @@
...
+FactoryBot.define do
+  factory(:seed_participant_profile, class: "ParticipantProfile") do
+    factory(:seed_ecf_participant_profile, class: "ParticipantProfile::ECF") do
+      type { "ParticipantProfile::ECF" }
+    end
+
+    schedule { Finance::Schedule.first }
+
+    trait :with_user do
+      user { build(:seed_user) }
```
cwrw, line 12: I think a profile has a user through teacher profile 🤔

spec/factories/seeds/partnership_factory.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory(:seed_partnership, class: "Partnership") do
+    trait(:with_cohort) { cohort { Cohort.first } }
```
cwrw, line 5: first three are required so could we move outside traits?

spec/factories/seeds/induction_programme_factory.rb
```diff
@@ -0,0 +1,17 @@
...
+FactoryBot.define do
+  factory(:seed_induction_programme, class: "InductionProgramme") do
+    training_programme { "full_induction_programme" }
+
+    trait(:fip) { training_programme { "full_induction_programme" } }
+    trait(:cip) { training_programme { "core_induction_programme" } }
+    trait(:design_our_own) { training_programme { "design_our_own" } }
+    trait(:school_funnded_fip) { training_programme { "school_funnded_fip" } }
+
+    trait(:with_school_cohort) { association(:school_cohort, factory: :seed_school_cohort) }
```
cwrw, line 12: could we move this outside a trait as it's required?

spec/factories/seeds/induction_programme_factory.rb
```diff
@@ -0,0 +1,17 @@
...
+    trait(:with_school) { association(:school, factory: :seed_school) }
```
cwrw, line 13: ✂️

cwrw (commented):
looks great, thanks for kicking this off! I think most of my comments revolve around the following ground rule:

> no associations by default (require traits to create other objects)

I think factories, if called on their own without any traits should be valid with the bare minimum set up. Traits can be used to add the extra associations which are not required, and will not cause the factory to fail otherwise. Happy to chat further around this decision and how to go forward, but left the comments around the ones we need to move out of traits in the meantime.

Otherwise it's a great first step, Im really excited to see how this will evolve when we add multiple scenarios and start tying things together.

peteryates:
> I think factories, if called on their own without any traits should be valid with the bare minimum set up. Traits can be used to add the extra associations which are not required, and will not cause the factory to fail otherwise.

This was  a bit of a dilemma for me too! The two options in my head were to:

## Make them minimal by default (current approach)

```ruby
FactoryBot.define do
  factory(:seed_some_object, class: "SomeObject") do
    trait(:with_some_other_object) { association(:some_other_object, factory: :seed_some_other_object) }
    trait(:with_yet_another_object) { association(:yet_another_object, factory: :seed_yet_another_object) }

    trait(:valid) do
      with_some_other_object
      with_yet_another_object
    end
  end
end
```

## Make them valid by default

```ruby
FactoryBot.define do
  factory(:seed_some_object, class: "SomeObject") do
    association(:some_other_object, factory: :seed_some_other_object)
    association(:yet_another_object, factory: :seed_yet_another_object)

    # better name needed!
    trait(:standalone) do
      some_other_object { nil }
      yet_another_object { nil }
    end
  end
end
```

I don't mind them generating their own relations but would we end up in the same situation as we're in now? Especially where some sets of models have triangular relationships and often we end up with a 'spare' object.

These approaches both fail anywhere there are `after(:build)` blocks as there's no way to inhibit them with traits.

spec/factories/seeds/participant_profile_factory.rb
peteryates: Yeah good call, afraid I copied and pasted a few times and that looks like it was meant for `participant_identity` 🤦🏽‍♂️

spec/factories/seeds/school_factory.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory(:seed_school, class: "School") do
+    transient { location { Faker::Address.city } }
+    transient { domain { [location.parameterize, Faker::Internet.domain_suffix].join } }
+
+    urn { Faker::Number.unique.decimal_part(digits: 7).to_s }
+    name { [location, %w[High Grammar Infant Nursary].sample, "School"].join(" ") }
```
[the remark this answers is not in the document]
peteryates: As this was for seeds rather than tests I was aiming for realistic-looking data, so the school would be `Northampton High`, in `Northampton`, and the email address would be `head@northampton-high.net`. I was also thinking about 'cascading' the domain down to the users (`lucy-maths-teacher@northampton-high.net`) but that's one for another day!

I figured that some consistency might make navigating in review apps a bit more cohesive.

spec/factories/seeds/induction_programme_factory.rb
peteryates: But `TeacherProfile` `has_one: :mentor_profile`? That surprised me too!

spec/factories/seeds/school_factory.rb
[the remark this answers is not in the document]
tonyheadford: Could this stuff be set in the scenarios, so basic Fakery by default but known values for scenarios?

spec/factories/seeds/cohort_factory.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory(:seed_cohort, class: "Cohort") do
+    sequence(:start_year) { Faker::Number.unique.between(from: 2025, to: 3025) }
+
```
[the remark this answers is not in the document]
tonyheadford: Would limiting the year to the current one work by default, the ranges could be specified in scenarios perhaps?  It gets tedious having to check if the cohort exists before being able to use it, there was a ticket somewhere I think about making `Cohort.current` take a param, perhaps that would help (but moves outside of factories)

spec/factories/seeds/induction_programme_factory.rb
tonyheadford: Ouch, I think there are a few crappy associations/scopes that could do with sanitising

tonyheadford:
> > I think factories, if called on their own without any traits should be valid with the bare minimum set up. Traits can be used to add the extra associations which are not required, and will not cause the factory to fail otherwise.
> 
> This was a bit of a dilemma for me too! The two options in my head were to:
> 
> ## Make them minimal by default (current approach)
> 
> ## Make them valid by default

Yes tricky. I think if we don't make them minimal by default we'll end up in the same place we are now. At least there's a way out using scenarios to build the more complex collections of objects once we've got them refactored up into smaller components down the line perhaps

peteryates:
Closing to regenerate a new review app with the working statuscake config

db/new_seeds/base/add_mentor_scenarios.rb
```diff
@@ -0,0 +1,93 @@
...
+
+# first some generic mentors
+
+[
+  OpenStruct.new(
+    full_name: "Sally Mentor",
+    email: "rp-mentor.ambition.2021@example.com",
+    core_induction_programme: ambition,
+    school_cohort: school_cohort_one,
+    schedule: Finance::Schedule::ECF.default,
```
tonyheadford, line 37: I think these need to use `Finance::Schedule::ECF.default_for(cohort: xxx)` other wise it will use `Cohort.current`?

peteryates: Good call, will do

spec/factories/seeds/school_factory.rb
```diff
@@ -0,0 +1,26 @@
...
+    transient { domain { [location.parameterize, Faker::Internet.domain_suffix].join } }
+
+    urn { Faker::Number.unique.decimal_part(digits: 7).to_s }
+    name { [location, %w[High Grammar Infant Nursery].sample, "School"].join(" ") }
+    postcode { Faker::Address.postcode }
+    address_line1 { Faker::Address.street_address }
+    address_line3 { location }
+    domains { Array.wrap(domain) }
+    school_website { "www.#{domain}" }
+    primary_contact_email { Faker::Internet.email(domain:) }
```
tonyheadford, line 15: Don't think these are required to create a valid school

spec/factories/seeds/teacher_profile_factory.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory(:seed_teacher_profile, class: "TeacherProfile") do
+    trn { Faker::Number.unique.rand_in_range(10_000, 100_000).to_s }
+
+    trait(:ith_school) do
```
tonyheadford, line 7:
```suggestion
    trait(:with_school) do
```

db/new_seeds/base/add_mentor_scenarios.rb
peteryates: Is this what you mean @tonyheadford? c58af5f0074ea1

spec/factories/seeds/school_factory.rb
peteryates: I added these so the school pages looked fuller (and hopefully more realistic) in seeded environments. Think they'll be problematic?

tonyheadford (approved):
Looks great!

cwrw (approved):
🌟 this is fantastic! Massive step in improving our seeds and process 👍

db/seeds.rb
```diff
@@ -2,23 +2,27 @@
...
+  if Rails.env.development? || Rails.env.deployed_development?
+    FeatureFlag::PERMANENT_SETTINGS.each do |feature|
+      FeatureFlag.activate(feature)
+    end
+    FeatureFlag::TEMPORARY_FEATURE_FLAGS.each do |feature|
+      FeatureFlag.activate(feature)
+    end
   end
+else
+  load(Rails.root.join(*%w[db new_seeds run.rb]).to_s)
```
cwrw, line 25: oops missed this, just a quick q do we need to worry about envs before we run this as in the legacy one we are explicitly setting those? also do we want to enable the feature flags as well (as those aren't included in the new seeds run file)

PR #2788, opened 2022-11-30
scenarios tests for comparing support ECTs query and serialiser

mattonfoot:
Does
------

- add shared examples file to add user record scenarios that have caused issues in migration to IRs
- add tests that check both the queries and the serialisers for consistency for the provided scenarios
- amend the user serialiser to remove flip-flopping on user type in ECT/Mentor scenario - chooses oldest profile for consistency
- build in fallbacks within IR serialiser for badly formed records
- add comments and tests to prove that it is not possible to use `InductionRecord.current` or `InductionRecord.latest` to identify correct records
- use SQL window query to order and group InductionRecords by user id and select the single relevant record to work with for performance

Does not
----------

- add support for more than one email address per user in Support ECTs - requires work in Support ECTs
- add support for newer or multiple `participant_profiles` where for instance an ECT has become a Mentor - assumes access still needed as ECT but does not check if ECT training is finished

### DIFF
--- a/app/serializers/api/v1/ecf_induction_record_serializer.rb
+++ b/app/serializers/api/v1/ecf_induction_record_serializer.rb
@@ -25,58 +25,78 @@ module Api
       }.freeze
 
       set_id :id do |induction_record|
-        induction_record.preferred_identity.user_id
+        find_participant_identity(induction_record)&.user_id
       end
 
-      attributes :email, &:participant_email
+      attributes :email do |induction_record|
+        find_participant_identity(induction_record)&.email
+      end
 
       attributes :full_name do |induction_record|
-        induction_record.preferred_identity.user.full_name
+        find_participant_identity(induction_record)&.user&.full_name
       end
 
       attributes :user_type do |induction_record|
-        case induction_record.participant_profile.type
-        when ParticipantProfile::ECT.name
-          USER_TYPES[:early_career_teacher]
-        when ParticipantProfile::Mentor.name
-          USER_TYPES[:mentor]
-        else
+        if is_consistently_withdrawn(induction_record)
           USER_TYPES[:other]
+        else
+          case induction_record.participant_profile.type
+          when ParticipantProfile::ECT.name
+            USER_TYPES[:early_career_teacher]
+          when ParticipantProfile::Mentor.name
+            USER_TYPES[:mentor]
+          else
+            USER_TYPES[:other]
+          end
         end
       end
 
       attributes :core_induction_programme do |induction_record|
-        core_induction_programme = find_core_induction_programme(induction_record)
-        case core_induction_programme&.name
-        when "Ambition Institute"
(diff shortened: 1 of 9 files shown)


app/serializers/api/v1/ecf_induction_record_serializer.rb
```diff
@@ -25,59 +25,79 @@ class ECFInductionRecordSerializer
...
       end
+
+      def self.find_participant_identity(induction_record)
+        induction_record.preferred_identity ||
+          induction_record.participant_profile.participant_identity
+      end
+
+      # guard against some of the more complex scenarios where it is not possible to determine what is correct behaviour
+      def self.is_consistently_withdrawn(induction_record)
+        induction_record.induction_status == "withdrawn" && induction_record.participant_profile.status == "withdrawn"
```
tonyheadford, line 99:
```suggestion
        induction_record.withdrawn_induction_status? && induction_record.participant_profile.withdrawn_record?
```

app/services/api/v1/ecf/induction_records_query.rb
```diff
@@ -13,40 +13,41 @@ def all
...
+      # we can't use `InductionRecord.current` here, because we need to include records that finished in the past if there is no current one
+      # we can't use `InductionRecord.latest` here because the latest record might not be what they are doing right now
+      #
+      # we need to find the oldest `participant_profile` for the user otherwise we risk unwanted changes to their access
+      # we need to prioritise `end_date` with `null` first because we need to find the record describing what they are doing right now over what they will eventually do after a transfer
+      # we need to order by `start_date` second in case two records have the same `end_date` such as after modifying a transfer to add a mentor
+      InductionRecord.select(
+        <<-SQL,
+          FIRST_VALUE(induction_records.id) OVER (
+            PARTITION BY "users"."id"
```
tonyheadford, line 44: You could probably replace with `teacher_profile.id` and save a join?

app/serializers/api/v1/ecf_induction_record_serializer.rb
mattonfoot: Thanks I was trying to figure that one out

app/services/api/v1/ecf/induction_records_query.rb
```diff
@@ -13,40 +13,41 @@ def all
...
+            ORDER BY induction_records.end_date ASC NULLS FIRST, induction_records.start_date ASC
```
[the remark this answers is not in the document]
mattonfoot: Adding

app/services/api/v1/ecf/induction_records_query.rb
mattonfoot: even simpler to use `"participant_profiles"."teacher_profile_id"` and reduce to one join that already has the right constraints on it.

tonyheadford (approved):

PR #2793, opened 2022-12-01
[CPDLP-1879] Move changing training status feature to the latest induction record level for ECF participant profiles (admin)

leandroalemao:
### Context

We are making changes to the drilldown to help DfE users better manage ECF participants' training status.  

Currently, DfE users cannot view and change a given participant’s training status in the drilldown effectively. This is due to treating training_status as a profile-level field and selecting the state of the participant from one of them.

We will remove the training status field and change functionality from ParticipantProfile::ECF and introduce the change training_status functionality to the latest induction record per lead provider. This should hopefully help DfE users better manage participants training_status including where a participant has links to two providers/schools.

- Ticket: [CPDLP-1879](https://dfedigital.atlassian.net/browse/CPDLP-1879)

Implement modifying training_status on the latest induction record per lead provider for ECF participant profiles.


app/views/finance/participants/_ecf_profile.html.erb
```diff
@@ -72,48 +68,65 @@
...
 
     <% summary_list.row do |row| %>
       <% row.key { "Training status" } %>
       <% row.value { ir.training_status } %>
+      <%
+        if ir == pp.latest_induction_record_for(cpd_lead_provider: ir.cpd_lead_provider)
+          row.action(
+            text: "Change",
+            visually_hidden_text: "training status",
+            href: new_finance_participant_profile_ecf_induction_records_path(pp.id, ir.id))
```
cwrw, line 107: instead of passing in the induction record ID, can we pass in the CPD lead provider ID, which then works well with our services as they already accept a CPD lead provider?

app/views/finance/participants/_ecf_profile.html.erb
```diff
@@ -38,11 +38,7 @@
   <% summary_list.row do |row| %>
     <% row.key { "Training status" } %>
     <% row.value { pp.training_status } %>
```
cwrw, line 40: I think we need to remove surfacing this value as well 🤔 ?

app/views/finance/participants/_ecf_profile.html.erb
```diff
@@ -72,48 +68,65 @@
...
       <% row.key { "Induction status" } %>
       <% row.value { ir.induction_status } %>
+      <% row.action(text: :none) %>
     <% end %>
 
     <% summary_list.row do |row| %>
       <% row.key { "Training status" } %>
       <% row.value { ir.training_status } %>
+      <%
+        if ir == pp.latest_induction_record_for(cpd_lead_provider: ir.cpd_lead_provider)
```
cwrw, line 103: how does this work for CIP induction records (they don't have a CPD lead provider). Do we want to have a fallback for those to change the status? 
Also could we move this method into a helper which can be tested?

app/views/finance/participants/_ecf_profile.html.erb
leandroalemao: Hi Abeer you ok?
Just realised I'm sending the induction record ID up to `app/forms/finance/ecf/change_training_status_form.rb` only to be able to then fetch the correct `cpd_lead_provider` from it and then passing only the `cpd_lead_provider` to the services (Defer, Withdraw , etc..). So it's up for the services to select again the latest induction record for that lead provider.
I hope it's all good
Tks

app/views/finance/participants/_ecf_profile.html.erb
leandroalemao: Addressed tks @cwrw

app/views/finance/participants/_ecf_profile.html.erb
leandroalemao: Addressed tks @cwrw

PR #2794, opened 2022-12-01
[CPDNPQ-723] Expand NPQ Funding API call to include targeted funding

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-723

api/v1/npq-funding/:id wil now return previously_received_targeted_funding_support in its response. This will be true if the user has any accepted applications where targeted_delivery_funding_eligibility is true


shjohnson (approved):
LGTM

PR #2821, opened 2022-12-13
Update Statuscake version to 2.0.5

cwrw:
### Context

Deploys have started failing as we are using a deprecated API version of statuscake

- Update statuscake version in terraform
- Use API token instead of API key and username
- update statuscake required resource variables


peteryates (approved):
Looks good!

PR #2809, opened 2022-12-08
[CPDLP-1686] Add API endpoint to create an outcome for a participant

leandroalemao:
### Context

A provider can create a new outcome when POSTing to the following endpoint  /api/v3/participants/npq/{participant_id}/outcomes

- Ticket: [CPDLP-1686](https://dfedigital.atlassian.net/browse/CPDLP-1686)

### Changes proposed in this pull request

Add a new POST endpoint `/api/v3/participants/npq/{participant_id}/outcomes` (for all versions) which accepts the following body:

```
{
  "data": {
    "type": "npq-outcome-confirmation",
    "attributes": {
      "course_identifier": "npq-leading-teaching",
      "has_passed": false,
      "completion_date": "2021-05-31T02:21:32.000Z"
    }
  }
}
```
 
and creates a new NPQ participant outcome

PR #2822, opened 2022-12-13
[CPDLP-1898] amend existing api post declarations endpoint to support outcomes

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1898
- Depends on PR: https://github.com/DFE-Digital/early-careers-framework/pull/2809

* Added feature flag `participant_outcomes_feature`
* Updated POST `/api/v1/participant-declarations` to include `has_passed`
* Updated POST `/api/v2/participant-declarations` to include `has_passed`
* `ParticipantOutcome::NPQ` is created when `has_passed` is true
* Updated API doc

### Guidance to review

* Enable feature
* Add `completed` declaration with `has_passed` set to `true`

[CPDLP-1898]: https://dfedigital.atlassian.net/browse/CPDLP-1898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/serializers/api/v1/participant_declaration_serializer.rb
+++ b/app/serializers/api/v1/participant_declaration_serializer.rb
@@ -35,7 +35,7 @@ module Api
         declaration.current_state.dasherize
       end
 
-      attribute :has_passed do |declaration|
+      attribute :has_passed, if: FeatureFlag.active?(:participant_outcomes_feature) do |declaration|
         if declaration.npq?
           declaration.outcomes.latest&.passed?
         end
--- a/app/serializers/api/v1/participant_declaration_serializer.rb
+++ b/app/serializers/api/v1/participant_declaration_serializer.rb
@@ -37,7 +37,7 @@ module Api
 
       attribute :has_passed, if: -> { FeatureFlag.active?(:participant_outcomes_feature) } do |declaration|
         if declaration.npq?
-          declaration.outcomes.latest&.passed?
+          declaration.outcomes.latest&.has_passed?
         end
       end
     end
--- a/app/serializers/api/v1/participant_declaration_serializer.rb
+++ b/app/serializers/api/v1/participant_declaration_serializer.rb
@@ -35,7 +35,7 @@ module Api
         declaration.current_state.dasherize
       end
 
-      attribute :has_passed, if: FeatureFlag.active?(:participant_outcomes_feature) do |declaration|
+      attribute :has_passed, if: -> { FeatureFlag.active?(:participant_outcomes_feature) } do |declaration|
         if declaration.npq?
           declaration.outcomes.latest&.passed?
         end
--- a/app/serializers/api/v1/participant_declaration_serializer.rb
+++ b/app/serializers/api/v1/participant_declaration_serializer.rb
@@ -34,6 +34,12 @@ module Api
       attribute :state do |declaration|
         declaration.current_state.dasherize
       end
+
+      attribute :has_passed do |declaration|
+        if declaration.npq?
+          declaration.outcomes.latest&.passed?
+        end
+      end
     end
   end
 end
--- a/app/serializers/api/v2/participant_declaration_serializer.rb
+++ b/app/serializers/api/v2/participant_declaration_serializer.rb
@@ -26,7 +26,7 @@ module Api
         declaration.current_state.dasherize
       end
 
-      attribute :has_passed, if: FeatureFlag.active?(:participant_outcomes_feature) do |declaration|
+      attribute :has_passed, if: -> { FeatureFlag.active?(:participant_outcomes_feature) } do |declaration|
         if declaration.npq?
           declaration.outcomes.latest&.passed?
         end
--- a/app/serializers/api/v2/participant_declaration_serializer.rb
+++ b/app/serializers/api/v2/participant_declaration_serializer.rb
@@ -25,6 +25,12 @@ module Api
       attribute :state do |declaration|
         declaration.current_state.dasherize
       end
+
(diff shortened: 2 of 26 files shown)


spec/serializers/api/v1/participant_declaration_serializer_spec.rb
```diff
@@ -4,13 +4,62 @@
...
+      let(:has_passed) { true }
+
+      it "returns true" do
+        result = described_class.new(participant_declaration).serializable_hash
+        expect(result[:data][:attributes][:has_passed]).to eql(true)
+      end
+    end
+
+    describe "when participant outcome is false" do
+      let(:has_passed) { true }
```
cwrw, line 49: could we set this to false and avoid the setup on line 53?

app/serializers/api/v1/participant_declaration_serializer.rb
```diff
@@ -34,6 +34,12 @@ class ParticipantDeclarationSerializer
       attribute :state do |declaration|
         declaration.current_state.dasherize
       end
+
+      attribute :has_passed do |declaration|
```
cwrw, line 38: is it possible to  only surface this value if feature flag is active?

app/services/record_declaration.rb
```diff
@@ -182,9 +186,31 @@ def validates_billable_slot_available
...
+  def participant_declaration_class
+    if participant_profile.npq?
+      ParticipantDeclaration::NPQ
+    else
+      ParticipantDeclaration::ECF
+    end
+  end
+
+  def validate_has_passed?
+    participant_profile&.npq? && declaration_type == "completed"
```
cwrw, line 198: could we also add the feature flag here as well?

cwrw (commented):
looking good, just some feature flag qs

app/services/record_declaration.rb
```diff
@@ -182,9 +186,33 @@ def validates_billable_slot_available
...
+      ParticipantDeclaration::NPQ
+    else
+      ParticipantDeclaration::ECF
+    end
+  end
+
+  def validate_has_passed?
+    return false unless FeatureFlag.active?(:participant_outcomes_feature)
+
+    participant_profile&.npq? && declaration_type == "completed"
```
cwrw, line 200: could we also ignore ehco and aso NPQ declarations?

mooktakim: @cwrw 
The `NPQ::CreateParticipantOutcome` service class won't create outcome if ehco and aso declarations.
Tom didn't want a validation error here, just not create outcome silently.

cwrw: I think however  we will still be validating has passed is required on those declarations, and might need to add an extra check here to avoid that validation?

app/serializers/api/v1/participant_declaration_serializer.rb
```diff
@@ -34,6 +34,12 @@ class ParticipantDeclarationSerializer
       attribute :state do |declaration|
         declaration.current_state.dasherize
       end
+
+      attribute :has_passed, if: FeatureFlag.active?(:participant_outcomes_feature) do |declaration|
```
cwrw, line 38:
```suggestion
      attribute :has_passed, if: -> { FeatureFlag.active?(:participant_outcomes_feature) } do |declaration|
```

app/serializers/api/v2/participant_declaration_serializer.rb
```diff
@@ -25,6 +25,12 @@ class ParticipantDeclarationSerializer
       attribute :state do |declaration|
         declaration.current_state.dasherize
       end
+
+      attribute :has_passed, if: FeatureFlag.active?(:participant_outcomes_feature) do |declaration|
```
cwrw, line 29:
```suggestion
      attribute :has_passed, if: -> { FeatureFlag.active?(:participant_outcomes_feature) } do |declaration|
```

app/serializers/api/v1/participant_declaration_serializer.rb
```diff
@@ -34,6 +34,12 @@ class ParticipantDeclarationSerializer
...
+        if declaration.npq?
+          declaration.outcomes.latest&.passed?
```
cwrw, line 40: when latest is empty, we are getting an empty association, might need to amend this scope

mooktakim: @cwrw 
Sure, its the reason why i changed `latest` to a class method instead of scope.

app/serializers/api/v1/participant_declaration_serializer.rb
```diff
@@ -34,6 +34,12 @@ class ParticipantDeclarationSerializer
       attribute :state do |declaration|
         declaration.current_state.dasherize
       end
+
+      attribute :has_passed, if: -> { FeatureFlag.active?(:participant_outcomes_feature) } do |declaration|
+        if declaration.npq?
+          declaration.outcomes.latest&.passed?
```
cwrw, line 40: ah I forgot about voided outcomes: if an outcome is voided we will currently return `false`, should we return `nil` instead?

cwrw (approved):
looks great 🙌 just one last comment around the serialiser

app/serializers/api/v1/participant_declaration_serializer.rb
mooktakim: @cwrw 
Updated code to return nil if voided

PR #2823, opened 2022-12-13
Remove data migration data

cwrw:
### Context

Remove code for now as it's breaking the migration in production, and will recreate to run as a one off task


peteryates (approved):
Looks good.

peteryates:
cc @ltello - we might need to find a workaround for this. Worth building it as a rake task or perhaps we just run it manually once?

PR #2831, opened 2022-12-14
[CPDLP-1937] Amend Participant not withdrawn validator

cwrw:
### Context

A participant can be reinstated after being withdrawn. However the validation does not take that into account and finds any withdrawn participant state for a profile and raises a validation error.

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1937

Change the validation to look for the most recent participant state scoped to provider, and then check if it has been withdrawn and the declaration is after the profile has been withdrawn.

Also add tests, we like tests.

### Guidance to review
Open to suggestions for a nicer implementation as always


leandroalemao (approved):
LGTM 👍

PR #2834, opened 2022-12-14
CST-1184: remove :multiple_cohorts feature flag

ltello:
### Context
Currently our cohorts are hard-coded and referenced (usually by start_year throughout the app and tests).
This means that every year there will need to be a manual task where a developer needs to go through and update things for the upcoming cohort.
This is likely to be a chore and we’re probably going to introduce bugs by missing a reference here or there to a previous year.
This is a big job

Two commonly-used methods are Cohort.current and Cohort.next, both check whether FeatureFlag.active?(:multiple_cohorts) is enabled and return either 2021 or 2022 accordingly.

Some of our tests assume the :multiple_cohorts feature flag is enabled.
Some assume the flag is disabled.

We have lots of code in production that hasn't been covered by specs for ~7 months since the flag was toggled
We have code and template content that's in templates but hasn't been accessible since the toggle date
We want to be confident that when we roll over into 2023, 2024 (or 2029)nothing breaks

we want simpler code, templates and tests that won’t break beyond a certain date

- Ticket: [CST-1184](https://dfedigital.atlassian.net/browse/CST-1184)

- remove the :multiple_cohorts feature flag entirely, it hasn’t been toggled for 7 months and will probably blow up in prod if we do switch it.

- remove areas of templates that are no longer accessible without the flag. This is easier said than done because some specs expect the multiple_cohorts: false variant.

- remove hardcoded references to particular cohorts. We might need to leave some where they reference a special year, like 2020.

### Guidance to review

[CST-1184]: https://dfedigital.atlassian.net/browse/CST-1184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/lead_providers/report_schools/base_controller.rb
+++ b/app/controllers/lead_providers/report_schools/base_controller.rb
@@ -13,7 +13,7 @@ module LeadProviders
 
       def start
         clean_form!
-        report_schools_form.cohort_id = @cohort&.id || Cohort.next&.id
+        report_schools_form.cohort_id = @cohort&.id || Cohort.next.id
         report_schools_form.lead_provider_id = current_user.lead_provider_profile.lead_provider.id
       end
 
--- a/app/controllers/lead_providers/report_schools/base_controller.rb
+++ b/app/controllers/lead_providers/report_schools/base_controller.rb
@@ -13,6 +13,7 @@ module LeadProviders
 
       def start
         clean_form!
+
         report_schools_form.cohort_id = @cohort&.id || Cohort.next.id
         report_schools_form.lead_provider_id = current_user.lead_provider_profile.lead_provider.id
       end
--- a/app/controllers/lead_providers/report_schools/base_controller.rb
+++ b/app/controllers/lead_providers/report_schools/base_controller.rb
@@ -13,8 +13,7 @@ module LeadProviders
 
       def start
         clean_form!
-
-        report_schools_form.cohort_id = @cohort&.id || Cohort.next.id
+        report_schools_form.cohort_id = @cohort&.id || Cohort.next&.id
         report_schools_form.lead_provider_id = current_user.lead_provider_profile.lead_provider.id
       end
 
--- a/app/controllers/lead_providers/report_schools/base_controller.rb
+++ b/app/controllers/lead_providers/report_schools/base_controller.rb
@@ -14,7 +14,7 @@ module LeadProviders
       def start
         clean_form!
 
-        report_schools_form.cohort_id = @cohort&.id || Cohort.next.id
+        report_schools_form.cohort_id = @cohort&.id || Cohort.current.id
         report_schools_form.lead_provider_id = current_user.lead_provider_profile.lead_provider.id
       end
 
--- a/app/controllers/schools/dashboard_controller.rb
+++ b/app/controllers/schools/dashboard_controller.rb
@@ -1,8 +1,7 @@
 # frozen_string_literal: true
 
 class Schools::DashboardController < Schools::BaseController
-  before_action :set_school_cohorts, only: :show, unless: -> { FeatureFlag.active?(:multiple_cohorts) }
-  before_action :set_multi_cohorts, only: :show, if: -> { FeatureFlag.active?(:multiple_cohorts) }
+  before_action :set_multi_cohorts, only: :show
   skip_after_action :verify_authorized
   skip_after_action :verify_policy_scoped
 
@@ -25,18 +24,6 @@ class Schools::DashboardController < Schools::BaseController
   end
 
 private
-
-  def set_school_cohorts
-    @school = active_school
-
-    @cohort_list = [Cohort.current]
-    @school_cohorts = @school.school_cohorts.where(cohort: @cohort_list)
-
-    # This will need to be updated when more than one cohort is supported
-    unless @school_cohorts[0]
-      redirect_to schools_choose_programme_path(cohort_id: Cohort.current.start_year)
-    end
-  end
 
   def set_multi_cohorts
     @school = active_school
--- a/app/controllers/schools/dashboard_controller.rb
+++ b/app/controllers/schools/dashboard_controller.rb
@@ -44,7 +44,7 @@ private
   def previous_lead_provider(school_cohort)
     school = school_cohort.school
     previous_start_year = school_cohort.cohort.start_year - 1
-    previous_school_cohort = school.school_cohorts.find_by(cohort: Cohort.find_by(start_year: previous_start_year))
+    previous_school_cohort = school.school_cohorts.find_by(cohort: Cohort[previous_start_year])
     previous_school_cohort.lead_provider
   end
 
--- a/app/controllers/schools/dashboard_controller.rb
+++ b/app/controllers/schools/dashboard_controller.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 class Schools::DashboardController < Schools::BaseController
-  before_action :set_multi_cohorts, only: :show
+  before_action :set_school_cohorts, only: :show
+  before_action :check_school_cohorts, only: :show
   skip_after_action :verify_authorized
   skip_after_action :verify_policy_scoped
 
@@ -25,12 +26,15 @@ class Schools::DashboardController < Schools::BaseController
 
 private
 
-  def set_multi_cohorts
-    @school = active_school
-    @school_cohorts = @school.school_cohorts.dashboard_cohorts
+  def check_school_cohorts
     if @school_cohorts.empty?
       redirect_to schools_choose_programme_path(cohort_id: Cohort.active_registration_cohort.start_year)
     end
+  end
+
+  def set_school_cohorts
+    @school = active_school
+    @school_cohorts = @school.school_cohorts.dashboard_cohorts
   end
 
   def set_up_new_cohort?
--- a/app/controllers/schools/dashboard_controller.rb
+++ b/app/controllers/schools/dashboard_controller.rb
@@ -44,7 +44,7 @@ private
   def previous_lead_provider(school_cohort)
     school = school_cohort.school
     previous_start_year = school_cohort.cohort.start_year - 1
-    previous_school_cohort = school.school_cohorts.find_by(cohort: Cohort[previous_start_year])
+    previous_school_cohort = school.school_cohorts.find_by(cohort: Cohort.find_by(start_year: previous_start_year))
     previous_school_cohort.lead_provider
   end
 
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -8,20 +8,27 @@ class Cohort < ApplicationRecord
   has_many :statements
 
   # Class Methods
+  def self.[](year)
+    find_by_start_year(year)
+  end
+
   def self.active_registration_cohort
     where(registration_start_date: ..Date.current).order(start_year: :desc).first
   end
 
   def self.previous
-    where(academic_year_start_date: ..(Date.current - 1.year)).order(start_year: :desc).first
+    date_range = (Date.current - 2.year + 1.day)..(Date.current - 1.year)
+    where(academic_year_start_date: date_range).order(start_year: :desc).first
   end
 
   def self.current
-    where(academic_year_start_date: ..Date.current).order(start_year: :desc).first
+    date_range = (Date.current - 1.year + 1.day)..Date.current
+    where(academic_year_start_date: date_range).first
   end
 
   def self.next
-    where.not(academic_year_start_date: ..Date.current).order(start_year: :asc).first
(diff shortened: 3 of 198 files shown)


app/models/cohort.rb
```diff
@@ -1,51 +1,64 @@
...
   has_many :call_off_contracts
   has_many :npq_contracts
+  has_many :partnerships
+  has_many :schedules, class_name: "Finance::Schedule"
+  has_many :statements
+
+  # Class Methods
+
+  # Find the cohort with start year matching the given year.
+  def self.[](year)
```
cwrw, line 13: Im not too sure about exposing this new interface for lookups, my preference would be to still use `find_by(start_year:)` or the current/next/previous methods as I prefer the explicitness. But happy to hear the team's thoughts on this. 

Or maybe for 2020 because it's a special year we have scopes instead, let me know your thoughts

app/models/cohort.rb
```diff
@@ -1,51 +1,64 @@
...
 
   def self.current
-    # TODO: Register and Partner 262: Figure out how to update current year
-    if FeatureFlag.active?(:multiple_cohorts)
-      where("academic_year_start_date <= ?", Time.zone.now).order(start_year: :desc).first
-    else
-      find_by(start_year: 2021)
-    end
+    date_range = (Date.current - 1.year + 1.day)..Date.current
+    where(academic_year_start_date: date_range).first
```
cwrw, line 23: Im a bit unsure around our rules for the start/end of a cohort, and how `registration_start_date` and `academic_year_start_date` differ, also if this is similar for both NPQ and ECF, do you mind clarifying?

spec/factories/cohorts.rb
```diff
@@ -10,13 +10,16 @@
     registration_start_date { Date.new(start_year.to_i, 5, 10) }
     academic_year_start_date { Date.new(start_year.to_i, 9, 1) }
 
+    trait :previous do
+      start_year { Date.current.year - (Date.current.month < 9 ? 2 : 1) }
```
cwrw, line 14: Im not too sure about the rules for cohort here, similar to the question above and if we should avoid using September?

cwrw (commented):
Immense work @ltello, I did an initial scan and had a few questions/thoughts, will do a deep dive later today

app/controllers/lead_providers/report_schools/base_controller.rb
```diff
@@ -13,8 +13,7 @@ class BaseController < ::LeadProviders::BaseController
 
       def start
         clean_form!
-
-        report_schools_form.cohort_id = @cohort&.id || Cohort.next.id
+        report_schools_form.cohort_id = @cohort&.id || Cohort.next&.id
```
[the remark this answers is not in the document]
ltello: That's fine by me.
It seems we are creating initial cohorts, schedules, lead providers, etc early on the seeding proccess. 
Should we add this suggestion to Peter to add it to his seed's PR?

app/views/schools/dashboard/_school_cohort_details.html.erb
```diff
@@ -50,7 +50,7 @@
     <% end %>
 
     <% list.row do |row| %>
-      <% row.key(text: "Delivery partner") %>
+s      <% row.key(text: "Delivery partner") %>
```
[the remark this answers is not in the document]
ltello: Thanks

cwrw:
I was also wondering if it was possible to break this PR up, maybe the feature flag removal first, then the scope changes and then the factory fix (not sure if that is possible though)

app/models/cohort.rb
ltello: This is just syntax sugar to shorten typing when you need to find a cohort by start year. 

It is so massively used everywhere in the test suite that I opted for:
 `Cohort[2021] || create(:cohort, start_year: 2021)` 
instead of
 `Cohort.find_by_start_year(2021) || create(:cohort, start_year: 2021)` 

Happy to revert it if people don't find it useful.

app/models/cohort.rb
ltello: For a cohort, I am considering:
- the academic year start date as the start of the cohort. 
- the registration start date, earlier than the academic year start date, and meant to allow the registration of participants for that cohort.

spec/factories/cohorts.rb
ltello: Given that I explicitly set Sept, 1st as the academic_year_start_date in the factory, I am using that same date to compute the previous cohort in the factory. 
Maybe I could rewrite it like this:
```
trait :previous do
      start_year { Date.current.year - (Date.current.month < academic_year_start_date.month ? 2 : 1) }
end
```

app/models/cohort.rb
tonyheadford: Yes,  I agree with @cwrw

app/models/cohort.rb
tonyheadford: The rules we followed when this was originally added were (if we needed to decide on which cohort to show in the UI for example) the `active_registration_cohort` would normally be the `current` except for the period between the `registration_start_date` and the start of the `next` cohort.

tonyheadford (commented):
Epic effort @ltello really well done getting through this, I know you put a lot of effort in over Christmas and it's really appreciated.  Like @cwrw mentioned I would really like to see this separated out into 2 or 3 PRs though. I know it's extra work but I think that's OK given the complexity of this and to help get the changes properly understood and digested and to help get this reviewed by devs across the teams.  From a cross team meeting we had this morning, I think there wants to be a bit more understanding of what's needed with Cohorts so, if you  were going to break this up, perhaps focus on removing the feature flag and fixing the related test first?  I don't know that others think.  Also I still need to spend more time working through the PR but wanted to know your thoughts on this first.

ltello:
I don't think it is possible to split this PR. 
Initially one might think (and that's what I started to do) to first remove the feature flag and then refactor the Cohort staff to remove hardcoded years in it and making it work for the coming cohorts starting in the next future. 

The thing is that everything was very interrelated:
- we had loads of tests for code only under multi_cohort conditions.
- loads of tests for code only under no multi_cohort conditions. That is, we were testing something not happening in production conditions and therefore not testing features as they really happen in production. 
- Cohort.current and Cohort.next were hardcoded to fixed years like 2021 and 2021 under no multi_cohort conditions and something very different under multi_cohort conditions.
- Most test cases were written assuming 2021 and 2022 would be the current or next cohorts. Or even 2021 and 2021.

When I started to remove the feature flag, for example one thing I had to do was to start rewriting dozens of complex tests that were written for non multi_cohort conditions. 
What's the purpose to rewrite them all assuming wrong definitions of Cohort.current and Cohort.next when in the next task (make Cohort dynamic for future years) I would redefine those methods and therefore I would need to rewrite these tests all again? 

So I tried a different approach starting first by making Cohort dynamic with no hardcoded years and leaving the feature flag for the end.
The issue with this strategy was that I would need to rewrite dozens of not useful tests (those under non multi-cohort conditions) to adapt them to dynamic cohorts. After that big effort I would need to remove the feature flag and therefore rewrite the same tests to adapt them to a multi_cohort scenario.

That sounded like a massive double effort, so I decided to do it all at once: 
- remove the feature flag
- make the cohorts dynamic
- and fix all the test suite to adapt it all to the new scenario.

Breaking this down into smaller PRs feels like impossible to me. I understand people want to get a clear idea of the changes in this PR but breaking this down into smaller PRs doesn't seem feasible for that purpose.

In any case, this PR puts us in a much better position than what we had: features not tested (or tested under non real conditions), features tested only for past/present cohorts with no guaranty of work for the next year, lot of legacy code not reachable in production, etc.)

app/controllers/lead_providers/report_schools/base_controller.rb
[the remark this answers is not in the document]
ltello: Oops!, the seeding PR has been merged already so I have added the changes here.

app/models/cohort.rb
ltello: reverted

app/controllers/schools/dashboard_controller.rb
```diff
@@ -1,8 +1,7 @@
 # frozen_string_literal: true
 
 class Schools::DashboardController < Schools::BaseController
-  before_action :set_school_cohorts, only: :show, unless: -> { FeatureFlag.active?(:multiple_cohorts) }
-  before_action :set_multi_cohorts, only: :show, if: -> { FeatureFlag.active?(:multiple_cohorts) }
+  before_action :set_multi_cohorts, only: :show
```
peteryates, line 4: It doesn't feel like `set_multi_cohorts` is a good name for this method any more, perhaps just `set_school_cohorts`?

Perhaps the check to ensure some exist should be split out too, but it's not as important as the method name.

db/legacy_seeds/initial_seed.rb
```diff
@@ -1,9 +1,22 @@
...
+                          registration_start_date: Date.new(2020, registration_month, registration_day),
+                          academic_year_start_date: Date.new(2020, academic_year_start_month, academic_year_start_day))
+
+# Create cohorts since 2021 until Cohort.next
+next_cohort_start_year = Date.current.year + (Date.current.month < academic_year_start_month ? 0 : 1)
+cohorts = (2021..next_cohort_start_year).to_a.map do |start_year|
+  Cohort.find_or_create_by!(start_year:,
+                            registration_start_date: Date.new(start_year, registration_month, registration_day),
+                            academic_year_start_date: Date.new(start_year, academic_year_start_month, academic_year_start_day))
+end
```
peteryates, line 19: does this need amending for the new seed structure?

peteryates: scratch that, have seen below!

app/models/cohort.rb
```diff
@@ -1,51 +1,59 @@
...
-  def description
-    "#{start_year} to #{start_year + 1}"
-  end
+  # Instance Methods
 
-  def next
-    Cohort.find_by(start_year: start_year + 1)
+  # e.g. "2021/22"
+  def academic_year
+    sprintf("#{start_year}/%02d", ((start_year + 1) % 100))
```
peteryates, line 35: Might need to be careful here about displaying a date range with a `/`, the GDS guidelines suggest using `2022 to 2023`.

However, depending on where we display it (ie if there are several in a table) this might be ok.

https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates

peteryates: Actually as we're not changing the behaviour here ignore this, I'll revisit it separately.

spec/factories/cohorts.rb
peteryates: Controversial thought, but should the cohort be in code (ie a config file) rather than in the database?

Clearly not one for now.

peteryates (approved):
I'm happy with these changes, everything feels well thought out. Really glad to see the back of this feature flag.

As it's a big PR I'd like another set of eyes on it too, but from me it looks good to go.

app/controllers/schools/dashboard_controller.rb
ltello: Renamed and split

spec/services/partnership_notification_service_spec.rb
```diff
@@ -2,20 +2,19 @@
 
 RSpec.describe PartnershipNotificationService do
   subject(:partnership_notification_service) { described_class.new }
+
+  let!(:lead_provider) { create(:lead_provider) }
+  let!(:delivery_partner) { create(:delivery_partner) }
+  # let!(:cohort) { Cohort.find_by(start_year: 2021) || create(:cohort, start_year: 2021) }
```
cwrw, line 8: ✂️

cwrw (commented):
Looks great, thanks for addressing all the comments. Just the one comment around next otherwise 👍

tonyheadford (approved):
Well done @ltello this was a big job!  I think once the remaining comments are address from others this looks good to me

spec/services/partnership_notification_service_spec.rb
ltello: Thanks

app/models/cohort.rb
```diff
@@ -1,51 +1,59 @@
...
-      find_by(start_year: 2021)
-    end
+    date_range = (Date.current - 1.year + 1.day)..Date.current
+    where(academic_year_start_date: date_range).first
   end
 
   def self.next
-    year = FeatureFlag.active?(:multiple_cohorts) ? 2022 : 2021
-    find_by(start_year: year)
+    date_range = (Date.current + 1.day)..(Date.current + 1.year)
```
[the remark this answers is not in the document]
ltello: Good point. I will review those `Current.next` and change them.

cwrw (approved):
amazing, nicely done 🌟

PR #2835, opened 2022-12-14
[CPDLP-1936] Check that user exists before creating declaration attempt

cwrw:
### Context
If the participant external ID does not match a user with the same ID we are failing to create record declarations as it fails the validation for creating a declaration attempt which requires a user to exist.

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1936

Do not depend on the user having the same ID as the external ID, look up the user and check they exist before attempting to create the declaration attempt to avoid unnecessary failures.

### DIFF
--- a/app/services/record_declaration.rb
+++ b/app/services/record_declaration.rb
@@ -85,15 +85,19 @@ private
   end
 
   def declaration_attempt
-    if participant_id && cpd_lead_provider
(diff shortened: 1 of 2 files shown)


leandroalemao (approved):
LGTM 👍

PR #2845, opened 2022-12-16
Reorganise the admin console

peteryates:
The admin console is going to be receiving some love in the new year. This PR is laying the groundwork.

### Changes so far:

* replace the participants tabbed overview with separate pages using the `SubnavComponent` (like the Schools section)
* add a currently unused landing page for the admin console, we might use this for an overview
* start moving some overlapping controller fucntionality into concerns, rather than piling it into `Admin::BaseController`

### Remaining tasks

* [x] Work out what to do with NPQ participants
* [x] Write some specs for the new controllers and concerns

### Preview

image


spec/support/features/pages/admin_support/admin_support_participant_list.rb
```diff
@@ -5,7 +5,8 @@
 module Pages
   class AdminSupportParticipantList < ::Pages::BasePage
     set_url "/admin/participants"
-    set_primary_heading "Participants"
+    # this is a hack as the participants name is the page title
+    set_primary_heading(/^.*$/)
```
peteryates, line 9: Forgot I'd left this in, will take another look 🤔

app/controllers/concerns/admin/participants/find_induction_records.rb
```diff
@@ -0,0 +1,39 @@
...
+            :appropriate_body,
+            :preferred_identity,
+            :schedule,
+            induction_programme: {
+              partnership: :lead_provider,
+              school_cohort: %i[cohort school],
+            },
+            mentor_profile: :user,
+          )
+          .order(created_at: :desc)
```
tonyheadford, line 35: We sometimes have issues with this ordering, I'm not sure why `created_at` is used, I think it would be better to do `order(start_date: :desc, created_at: :desc)` to prevent records appearing wrongly if they've been created out of order / de-duped / merged etc.

tonyheadford (approved):
Looks great! I like the addition of the concerns too, nice.

app/controllers/concerns/admin/participants/find_induction_records.rb
ltello: I think that ideally the order should only be `created_at: :desc` but, as Tony mentions, we still have issues with the start-date/end-date of records not matching the creation order.

If we finally order the records according to Tony's suggestion, can we please add the creation date to the list of fields displayed for each induction record in the `induction records` tab? That would help to investigate and solve Zendesk tickets.

ltello (approved):

app/controllers/concerns/admin/participants/find_induction_records.rb
peteryates: Thank you, made changes addressing both requests.

app/views/admin/participants/induction_records/show.html.erb
```diff
@@ -75,6 +83,11 @@
         row.value(text: induction_record.appropriate_body_name)
       end
 
+      sl.row do |row|
+        row.key(text: "Creation date")
+        row.value(text: induction_record.start_date&.to_s(:govuk))
```
ltello, line 88:
```suggestion
        row.value(text: induction_record.created_at&.to_s(:govuk))
```

ltello:
Nice!

spec/support/features/pages/admin_support/admin_support_participant_list.rb
peteryates: Leaving this as it's only a spec.

PR #2848, opened 2022-12-18
Add a .keep file to logs directory

peteryates:
### Context

When setting up the app from fresh we get this error because we try and write to the mail.log file without the log directory existing.

```
No such file or directory @ rb_sysopen - log/mail.log
```

The log directory being present will stop it from tripping people up in the future.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Install dependencies
         run: bundle install
       - name: Create mail log file
-        run: mkdir log && touch log/mail.log
+        run: touch log/mail.log
       - name: Lint ruby
         run: rubocop
       - name: Lint SCSS


.github/workflows/lint.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Install dependencies
         run: bundle install
       - name: Create mail log file
-        run: mkdir log && touch log/mail.log
+        run: touch log/mail.log
```
cwrw, line 26: could we remove this as well from the readme/dockerfile.dev?

cwrw (approved):

.github/workflows/lint.yml
peteryates: so I actually just removed the `touch log/mail.log` commands too, they're regular `Logger` objects so will create the file automatically if it's not there (providing the containing directory is present)

PR #2856, opened 2022-12-20
Remove references to touch `log/mail.log`

peteryates:
Sorry @cwrw I set automerge on the previous PR 🤦🏽 

Here's the commit I meant to add.

> The mail log is instantiated with `Logger.new("log/mail.log", ...)` so will automatically create the file should it be missing, providing the log directory is present.


cwrw (approved):
thanks @peteryates

PR #2880, opened 2023-01-05
[CPDLP-1934] Void an outcome when a provider voids a completion declaration which has an outcome

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1934

* Updated void participant declaration endpoint service to create void participant outcome
* Created `NPQ::VoidParticipantOutcome` service to action void outcome creation.

### DIFF
--- a/app/services/npq/void_participant_outcome.rb
+++ b/app/services/npq/void_participant_outcome.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+module NPQ
+  class VoidParticipantOutcome
+    def initialize(participant_declaration)
+      @participant_declaration = participant_declaration
+    end
+
+    def call
+      return unless create_participant_outcome?
+      return if latest_participant_outcome&.voided?
+
+      ParticipantOutcome::NPQ.create!(
+        participant_declaration:,
+        completion_date: declaration_date,
+        state: "voided",
+      )
+    end
+
+  private
+
+    attr_reader :participant_declaration
+
+    delegate :course_identifier, :participant_profile,
+             :declaration_date, :declaration_type,
+             to: :participant_declaration
+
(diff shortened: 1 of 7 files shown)


app/services/npq/void_participant_outcome.rb
```diff
@@ -0,0 +1,47 @@
...
+  class VoidParticipantOutcome
+    def initialize(participant_declaration)
+      @participant_declaration = participant_declaration
+    end
+
+    def call
+      return unless create_participant_outcome?
+      return if latest_participant_outcome&.voided?
+
+      ParticipantOutcome::NPQ.create!(
```
cwrw, line 13: could we call `CreateParticipantOutcome` instead as it has further validations?

app/services/npq/void_participant_outcome.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def create_participant_outcome?
+      return false unless FeatureFlag.active?(:participant_outcomes_feature)
+
+      participant_profile&.npq? &&
+        declaration_type == "completed" &&
+        valid_course_identifier_for_participant_outcome?
+    end
+
+    def valid_course_identifier_for_participant_outcome?
```
cwrw, line 36: since this is repeated logic similar to record declaration, maybe we can move it into the create outcome service?

app/services/npq/void_participant_outcome.rb
mooktakim: @cwrw 
My initial implementation was to use `CreateParticipantOutcome`, but that has extra validation we don't need for void. Thought it would be better to create a simpler service.

cwrw (approved):
🙌

PR #2885, opened 2023-01-06
CPDLP-1924 Change default cohort and schedule selection on NPQ EHCO applications

cwrw:
### Context

NPQ EHCO schedules currently differ in cohorts:
- Cohort 2021: we had only one schedule: `npq-ehco-june` which all NPQ applications with the EHCO course were set to.
- Cohort 2022: we can have multiple schedules: [`npq-ehco-november`, `npq-ehco-december`, `npq-ehco-march`, `npq-ehco-june`]. 
NPQ applications are set the schedule relevant to the date range when an application is to be created. 

Since we need to support both 2021 and 2022 cohorts, we need to maintain the previous logic and support setting the new schedules. Also since we are still receiving applications for 2021 NPQ EHCO course, and have not started receiving EHCO applications for 2022.

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1924

- Maintain previous logic schedule for 2021 cohort
- Add new date range logic to set schedule for 2022 cohort
- Set a default per cohort for NPQ finance calculations, as well as import scripts

### Guidance to review
Did I miss anything?

### DIFF
--- a/app/models/finance/schedule/npq_ehco.rb
+++ b/app/models/finance/schedule/npq_ehco.rb
@@ -9,8 +9,8 @@ module Finance
 
       PERMITTED_COURSE_IDENTIFIERS = IDENTIFIERS
 
-      def self.default
-        find_by(cohort: Cohort.find_by!(start_year: 2022), schedule_identifier: "npq-ehco-december")
+      def self.default_for(cohort: Cohort.current)
+        find_by(cohort:, schedule_identifier: "npq-ehco-november")
       end
 
       def self.schedule_for(cohort: Cohort.current)
--- a/app/models/finance/schedule/npq_ehco.rb
+++ b/app/models/finance/schedule/npq_ehco.rb
@@ -12,6 +12,27 @@ module Finance
       def self.default
         find_by(cohort: Cohort.find_by!(start_year: 2022), schedule_identifier: "npq-ehco-december")
       end
+
+      def self.schedule_for(cohort: Cohort.current)
+        return Finance::Schedule::NPQEhco.find_by!(cohort:) unless cohort_with_multiple_schedules?(cohort)
+
+        case Date.current
+        when Date.new(cohort.start_year, 9, 1)..Date.new(cohort.start_year, 11, -1)
+          Finance::Schedule::NPQEhco.find_by!(cohort:, schedule_identifier: "npq-ehco-november")
+        when Date.new(cohort.start_year, 12, 1)..Date.new(cohort.start_year + 1, 2, -1)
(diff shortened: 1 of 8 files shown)


app/models/finance/schedule/npq_ehco.rb
```diff
@@ -9,8 +9,29 @@ class NPQEhco < NPQ
...
-        find_by(cohort: Cohort.find_by!(start_year: 2022), schedule_identifier: "npq-ehco-december")
+      def self.default_for(cohort: Cohort.current)
+        find_by!(cohort:, schedule_identifier: "npq-ehco-june")
+      end
+
+      def self.schedule_for(cohort: Cohort.current)
+        return Finance::Schedule::NPQEhco.find_by!(cohort:) unless cohort_with_multiple_schedules?(cohort)
+
+        case Date.current
+        when Date.new(cohort.start_year, 9, 1)..Date.new(cohort.start_year, 11, -1)
```
peteryates, line 20: Interesting, I didn't know you could get the last day of a month like this.

app/models/finance/schedule/npq_ehco.rb
```diff
@@ -9,8 +9,29 @@ class NPQEhco < NPQ
...
+          Finance::Schedule::NPQEhco.find_by!(cohort:, schedule_identifier: "npq-ehco-november")
+        when Date.new(cohort.start_year, 12, 1)..Date.new(cohort.start_year + 1, 2, -1)
+          Finance::Schedule::NPQEhco.find_by!(cohort:, schedule_identifier: "npq-ehco-december")
+        when Date.new(cohort.start_year + 1, 3, 1)..Date.new(cohort.start_year + 1, 5, -1)
```
peteryates, line 24: Just a thought but could this be easier to read if the dates were set as variables and the ranges were like:

```ruby
when first_day_of_march_next_year..last_day_of_may_next_year
  ...
```

app/models/npq_course.rb
```diff
@@ -7,7 +7,7 @@ def self.identifiers
     pluck(:identifier)
   end
 
-  def self.schedule_for(npq_course:, cohort: Cohort.find_by!(start_year: 2021))
+  def self.schedule_for(npq_course:, cohort: Cohort.current)
```
peteryates, line 10: This makes much more sense.

app/models/finance/schedule/npq_ehco.rb
cwrw: good idea, done in [e11661e](https://github.com/DFE-Digital/early-careers-framework/pull/2885/commits/e11661ed216509fa0eae9696fbe0c2a383bb8086) although needed to pass cohort in, reads much better 👍

PR #2886, opened 2023-01-09
Ect support user api logic

mattonfoot:
### Context

- Ticket: Participants who have at any point been registered for FIP cannot see CIP materials [CST-1082](https://dfedigital.atlassian.net/browse/CST-1082?atlOrigin=eyJpIjoiNDZkY2IwMTY3ODk2NGJmMDg5OWNhZjg2MWM4MDg3OWUiLCJwIjoiaiJ9)

Replace the query service and serializer in the `/ecf-users` api endpoint with the new query service and serialiser that uses inudction records and preferred identities. These services have been tested against the original services to ensure that any changes to the data is mostly as expected.

[CST-1082]: https://dfedigital.atlassian.net/browse/CST-1082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


peteryates:
So I'm happy with making the switch from a code point of view, the IR route definitely makes more sense. Definitely needs eyes from the API team though.

> to ensure that any changes to the data is mostly as expected

🤣

tonyheadford (approved):

peteryates (approved):
Looks good to me so long as API team are happy.

app/controllers/api/v1/ecf_users_controller.rb
```diff
@@ -48,7 +48,7 @@ def email
       end
 
       def users
-        Api::V1::ECF::UsersQuery.new(updated_since:, email:).all
```
cwrw, line 51: could we also remove the rake task that diffs those out as those classes no longer exist?

cwrw (commented):
looks good, just the one comment. 

I had a look at the new query/serialiser, and without much background was wondering why we were not returning attribute values for some fields/returning other from the serialiser if the profile was consistently withdrawn?

app/controllers/api/v1/ecf_users_controller.rb
mattonfoot: I've removed the rake task, thanks for the spot.

On the query about "other" records. I was being consistent with the original implementation, however if you look at the actual code for ECT Support it just ignores these records.

There is more work that needs to be done to ECT Support and this API, but I wanted to do it in small steps rather than everything in one massive bang especially as there is no clear direction on how we want to handle things like multi role access and access for SITs or others. I think that this is a really poorly thought through API to be honest, its neither completely agnostic of usage or completely controlling of function so the actual logic for who should have access is shared across both applications.

If the POs want me to progress with sorting these service problems out I will, but I am not sure they want me to without proper research and design support.

cwrw (approved):
thanks @mattonfoot

PR #2893, opened 2023-01-11
Add NPQ Course seeds

cwrw:
### Context

We need NPQ courses when integrating with the NPQ registration application to create new applications and sync users successfully. We also need to add NPQ courses to NPQ applications when setting up scenarios and declarations as well.

- Ticket: n/a

- Add new seeds to set up the previous ids and NPQ courses that sync up with the NPQ registration application in dec
- Add an identifier to the NPQ course factory as that's required for declarations and validity across the app
- Add an option to pass in an NPQ course when creating NPQ applications and declarations

### Guidance to review
Does this follow the new seed format?

### DIFF
--- a/db/new_seeds/base/add_npq_courses.rb
+++ b/db/new_seeds/base/add_npq_courses.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+[
+  {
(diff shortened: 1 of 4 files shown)


peteryates (approved):
Looks great

PR #2894, opened 2023-01-12
Support participant profile id in admin search

peteryates:
### Context

Sometimes we have a participant's profile ID and need to find their record. The search didn't support doing this so admins had to make a round trip to the participant drill down (finance) to get the other details and then back to the admin console.


cwrw (approved):
Maybe we can add the external ID as well, as found myself doing that as well?

peteryates:
Good idea, will do!

cwrw (approved):
🙌

PR #2897, opened 2023-01-12
[CPDLP-1965] Add validation to has_passed field for npq completed declarations

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1965

* Added validation for declarations endpoint to validate `has_passed` when unsupported values are provided.

### DIFF
--- a/spec/services/record_declaration_spec.rb
+++ b/spec/services/record_declaration_spec.rb
@@ -345,6 +345,15 @@ RSpec.describe RecordDeclaration, :with_default_schedules do
         end
       end
 
+      context "has_passed is invalid text" do
+        let(:has_passed) { "no_supported" }
+
+        it "returns error" do
+          expect(service).to be_invalid
+          expect(service.errors.messages_for(:has_passed)).to eq(["The property '#/has_passed' must be true or false"])
+        end
+      end
+
       context "has_passed is true" do
         let(:has_passed) { true }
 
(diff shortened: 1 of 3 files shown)


spec/services/record_declaration_spec.rb
```diff
@@ -345,6 +345,15 @@
         end
       end
 
+      context "has_passed is invalid text" do
+        let(:has_passed) { "no_supported" }
+
+        it "returns error" do
+          expect(service).to be_invalid
+          expect(service.errors.messages_for(:has_passed)).to eq(["The property '#/has_passed' must be true or false"])
```
peteryates, line 353: Are we worried about case here? In the ticket it says `TRUE` / `FALSE`.

peteryates (commented):
Looks good over all, one comment on checking the case of requests.

spec/services/record_declaration_spec.rb
peteryates: Additionally if we change this spec group to `type: :model` we can check the validations with:

```
expect(service).to allow_value("true", "false").for(:has_passed).with_message("The property...")
```

which might be a bit more concise.

mooktakim: @peteryates 
Added fix for capitalisation.

mooktakim: Speaking with @cwrw, I've changed back to only accepting lowercase to stay consistent with the rest of the API. We'll update the guidance to reflect this.

cwrw (approved):
👍 , might need to close/reopen not sure whats going on with the review app 🤔

PR #2906, opened 2023-01-17
Switch around the header and hide the section

peteryates:
### Context

This is closer to the original design while keeping the distinct section name available for screen reader users.

| Header | HTML |
|------|-------|
|  image |  image |


Varsagod:
It looks shiny

cwrw (approved):

PR #2913, opened 2023-01-18
CST-1207: rectify mentor pools when a participant is rectified

ltello:
### Context

   When a school gets closed their participants are moved to its successor but the mentor pools are left unchanged.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-1207)

   Rectify the mentor pools of the school and its successor so the entries are moved from one to the other.

### DIFF
--- a/app/services/data_stage/process_school_changes.rb
+++ b/app/services/data_stage/process_school_changes.rb
@@ -98,7 +98,8 @@ module DataStage
         school_cohort.update!(school: successor)
         school_cohort.ecf_participant_profiles.each do |profile|
           RectifyParticipantSchool.call(participant_profile: profile,
-                                        school: successor,
+                                        from_school: school,
+                                        to_school: successor,
                                         transfer_pupil_premium_and_sparsity: false)
         end
       end
--- a/app/services/rectify_participant_school.rb
+++ b/app/services/rectify_participant_school.rb
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 
 class RectifyParticipantSchool < BaseService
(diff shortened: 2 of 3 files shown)


app/models/school_mentor.rb
```diff
@@ -5,5 +5,6 @@ class SchoolMentor < ApplicationRecord
   belongs_to :participant_profile, class_name: "ParticipantProfile::Mentor"
   belongs_to :preferred_identity, class_name: "ParticipantIdentity"
 
+  scope :by, ->(participant_profile) { where(participant_profile:) }
```
tonyheadford, line 8: Not sure of the value of this as it is only a simple `where(participant_profile:)` and is a bit ambiguous, what if I want want to look up `by` school?  I get it's adding a bit of sugar but it then also becomes something else to remember.

tonyheadford (commented):
Looks good, just one comment about the `by` scope.

app/models/school_mentor.rb
ltello: I agree it is a simple scope just filtering by participant. My idea was not to have to take the `where(participant_profile:)` out of the SchoolMentor model. I think this is something internal to the class.

Should renaming it to something more meaningful do the job?. When I wrote it I was not 100% happy with any name:
`by(participant_profile)`
`by_participant_profile(participant_profile)`
`by_mentor(participant_profile)`
`by_participant(participant_profile)`
 
Any suggestion?

ltello: Ok, code rewritten without the scope.

tonyheadford (approved):
Nice!

PR #2915, opened 2023-01-19
Add some validation data to the seeds

peteryates:
We're taking an extra step here and making the seeded validation data closely resemble the user by adding an initial to their name via the transient user attribute.

### DIFF
--- a/db/new_seeds/scenarios/participants/mentors/mentoring_multiple_ects_with_same_provider.rb
+++ b/db/new_seeds/scenarios/participants/mentors/mentoring_multiple_ects_with_same_provider.rb
@@ -5,7 +5,14 @@ module NewSeeds
     module Participants
       module Mentors
         class MentoringMultipleEctsWithSameProvider
-          Person = Struct.new(:user, :teacher_profile, :participant_profile, :participant_identity, keyword_init: true)
+          Person = Struct.new(
+            :user,
+            :teacher_profile,
+            :participant_profile,
+            :participant_identity,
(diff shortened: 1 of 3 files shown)


spec/seeds/seed_factories/ecf_participant_validation_data_fractory_spec.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
```
cwrw, line 1: ✂️ fractory|factory

cwrw (approved):

spec/seeds/seed_factories/ecf_participant_validation_data_fractory_spec.rb
```diff
@@ -0,0 +1,10 @@
...
+
+require_relative "./shared_factory_examples"
+
+RSpec.describe("seed_ecf_participant_validation_data") do
+  it_behaves_like("a seed factory") do
+    let(:factory_name) { :seed_ecf_participant_validation_data }
+    let(:factory_class) { ECFParticipantValidationData }
```
cwrw, line 8: would also be good to add `ECFParticipantEligibility` eventually as well

spec/seeds/seed_factories/ecf_participant_validation_data_fractory_spec.rb
peteryates: D'oh, thanks!

PR #2917, opened 2023-01-19
[CPDLP-1913] Restrict participants endpoint fields

tvararu:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1913

This endpoint is currently timing out in production for certain providers. One possible way to speed it up is by reducing the amount of columns returned to only the necessary ones.

Restrict the fields being returned by the query underlying the `/api/v1/participants/ecf` endpoint.

### Guidance to review

Before we merge this, it would be best to test it if possible to see if it has the expected effect of speeding up the endpoint when dealing with a prod-like data load.

This can be done locally, if we can get a prod-like database running with a similar amount of data, or it can be done on a prod-like env.

It's lacking a test, partly because I want to try it out first to see if it improves performance before I invest further, but also because testing for the removal of fields always feels weird to me. Checking if a field isn't found in a payload can easily be a false positive test. If someone has a better idea for a good way to test this, please shout and we can update this.

### DIFF
--- a/app/services/api/v1/ecf/participants_query.rb
+++ b/app/services/api/v1/ecf/participants_query.rb
@@ -24,7 +24,19 @@ module Api
                      },
                    )
 
+          necessary_fields = %i[
+            induction_programme_id
+            induction_status
+            mentor_profile_id
+            participant_profile_id
+            preferred_identity_id
+            schedule_id
+            training_status
+            updated_at
+          ]
+
           scope = InductionRecord
+                    .select(*necessary_fields)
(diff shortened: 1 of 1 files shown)


tvararu:
Tested this on a prod env using a benchmarking script, about a 2-3x improvement in this query which is great. This should be good to ship.

peteryates (approved):
Looks great!

tvararu:
Can be merged once https://github.com/DFE-Digital/early-careers-framework/pull/2921 fixes the review app build :crossed_fingers:

PR #2920, opened 2023-01-20
[CPDLP-1989] Fix big data participants query to fetch the training status from latest induction record

leandroalemao:
### Context

There's a participant that has transferred where the API shows the correct value for training_status but the check data service shows something wrong.

This is the participant in question: 55485a55-c6ec-49c5-b9b6-ccb68d05cea0 https://manage-training-for-early-career-teachers.education.gov.uk/admin/participants/b8fdaffa-da1c-418b-9ad8-ae31a32120fd/induction_recordsThe lead provider (Ambition) sees active while the DP (3 rivers) sees the equivalent of withdrawn. It appears to be a 'botched transfer' and the participant has changed cohort too

- Ticket: [CPDLP-1989](https://dfedigital.atlassian.net/browse/CPDLP-1989)

The `training_status` on the `participants.sql` query is being fetched from `participant_profiles` (profile level) and not from the latest induction record as we’re doing in the API. So we're making a change in the sql query to fetch the `training_status` from the latest induction record instead of doing it from participant profile, with that the delivery partner view will become the same as what a provider sees in the API.

[CPDLP-1989]: https://dfedigital.atlassian.net/browse/CPDLP-1989?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/analytics_queries/participants.sql
+++ b/analytics_queries/participants.sql
@@ -13,7 +13,7 @@
             WHERE pp2.id = pp.mentor_profile_id)                  as mentor_id,
           c.start_year                                            as cohort,
           pp.status                                               as status,
-          latest_induction_record.training_status                 as training_status,
+          latest_induction_records.training_status                as training_status,
           pps.reason                                              as training_status_reason,
           s.name                                                  as schedule,
           s.schedule_identifier                                   as schedule_identifier,
@@ -36,17 +36,23 @@
             JOIN participant_identities pi on pp.participant_identity_id = pi.id
             LEFT OUTER JOIN ecf_participant_validation_data epvd on pp.id = epvd.participant_profile_id
             LEFT OUTER JOIN ecf_participant_eligibilities epe on pp.id = epe.participant_profile_id
-            JOIN (
-                 SELECT
-                    DISTINCT ON (participant_profile_id) participant_profile_id,
-                    training_status
-                 FROM induction_records
-                 JOIN induction_programmes ON induction_programmes.id = induction_records.induction_programme_id
-                 JOIN partnerships ON partnerships.id = induction_programmes.partnership_id
-                 ORDER BY participant_profile_id, induction_records.created_at DESC
-            ) AS latest_induction_record ON latest_induction_record.participant_profile_id = pp.id
+            LEFT OUTER JOIN (
+              SELECT
+                induction_records.*,
+                partnerships.lead_provider_id,
+                schools.id AS school_id,
+                schools.name AS school_name,
+                lead_providers.name AS lead_provider_name,
+                induction_programmes.training_programme AS training_programme,
+                ROW_NUMBER() OVER (PARTITION BY participant_profile_id ORDER BY induction_records.created_at DESC) AS induction_record_sort_order
+              FROM induction_records
+              JOIN induction_programmes ON induction_programmes.id = induction_records.induction_programme_id
+              LEFT OUTER JOIN partnerships         ON partnerships.id         = induction_programmes.partnership_id
+              LEFT OUTER JOIN lead_providers       ON lead_providers.id       = partnerships.lead_provider_id
+              LEFT OUTER JOIN schools              ON schools.id              = partnerships.school_id
+            ) AS latest_induction_records ON latest_induction_records.participant_profile_id = pp.id AND induction_record_sort_order = 1
             LEFT OUTER JOIN participant_profile_states pps on pp.id = pps.participant_profile_id AND pps.id = (
-              SELECT id from participant_profile_states _pps WHERE _pps.participant_profile_id = pp.id AND _pps.state = latest_induction_record.training_status ORDER BY created_at desc LIMIT 1
+              SELECT id from participant_profile_states _pps WHERE _pps.participant_profile_id = pp.id AND _pps.state = latest_induction_records.training_status ORDER BY created_at desc LIMIT 1
             )
     WHERE pp.type IN ('ParticipantProfile::ECT', 'ParticipantProfile::Mentor')
       AND c.start_year > 2020
--- a/analytics_queries/participants.sql
+++ b/analytics_queries/participants.sql
@@ -1,5 +1,5 @@
 \copy (
-    SELECT u.id                                                    as participant_id,
+    SELECT u.id                                                   as participant_id,
           pi.external_identifier                                  as external_id,
           pp.created_at                                           as added_at,
           sch.urn                                                 as school_urn,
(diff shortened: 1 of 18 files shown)


analytics_queries/participants.sql
```diff
@@ -36,8 +36,17 @@
             JOIN participant_identities pi on pp.participant_identity_id = pi.id
             LEFT OUTER JOIN ecf_participant_validation_data epvd on pp.id = epvd.participant_profile_id
             LEFT OUTER JOIN ecf_participant_eligibilities epe on pp.id = epe.participant_profile_id
+            JOIN (
+                 SELECT
+                    DISTINCT ON (participant_profile_id) participant_profile_id,
+                    training_status
+                 FROM induction_records
+                 JOIN induction_programmes ON induction_programmes.id = induction_records.induction_programme_id
```
cwrw, line 44: Just want to confirm we just want the latest induction record, not scoped to a provider?

analytics_queries/participants.sql
```diff
@@ -36,8 +36,17 @@
...
+                 JOIN partnerships ON partnerships.id = induction_programmes.partnership_id
+                 ORDER BY participant_profile_id, induction_records.created_at DESC
+            ) AS latest_induction_record ON latest_induction_record.participant_profile_id = pp.id
```
cwrw, line 47: I think we might need to define that is is one record here, there are examples of this in the `participant_query` and the `ecf_duplicates_query`

cwrw: `ROW_NUMBER() OVER (PARTITION BY participant_profile_id ORDER BY induction_records.created_at DESC) AS induction_record_sort_order` then adding `AND induction_record_sort_order = 1`

cwrw (commented):
out of interest when this is merged how does resyncing work? cc @peteryates

analytics_queries/participants.sql
leandroalemao: Hi @cwrw I thought that.. but we don't have the results or joins broken by lead provider anywhere in this query.. so I reckon we want the latest induction record regardless the lead provider..

analytics_queries/participants.sql
leandroalemao: Hi @cwrw I've just made some changes to keep it consistent with the latest induction records selection from the examples. Tksss

leandroalemao:
> out of interest when this is merged how does resyncing work? cc @peteryates

Looks like there's a `analytics_queries/Makefile` file which runs the queries and exports the results onto csv files but not sure if that runs automatically or not. And then big data was a process to read from those files (???).  It's written on `analytics_queries/README` that Nathan Easey is the point of contact for questions around this.

analytics_queries/participants.sql
cwrw: chatted in person but for posterity: We agreed that getting the latest induction will not always yield the latest provider/training status in some edge cases, however if we would like to scope the latest induction record to a provider there might be some changes required on the BQ side. We will need discuss and coordinate that at a later time with the BQ team

leandroalemao:
Created https://github.com/DFE-Digital/early-careers-framework/pull/2928 to fix the failing test.

app/services/analytics/ecf_validation_service.rb
```diff
@@ -54,6 +54,10 @@ def eligible_for_funding?(participant_profile)
       def trn_verified?(participant_profile)
         participant_profile.ecf_participant_eligibility.present?
       end
+
+      def training_status_for(participant_profile)
+        participant_profile&.latest_induction_record&.training_status
```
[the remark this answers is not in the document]
leandroalemao: @cwrw All done.. tks 👍 😄

cwrw (approved):

peteryates:
@cwrw @leandroalemao as I said on Slack I'm afraid the `analytics_queries` are entirely separate and run in an non-automated fashion, it's the `app/models/analytics` objects we need to focus on. I sync them by looping through objects and creating sync jobs, then letting it churn away for a few minutes.

peteryates (approved):
Looks good to me! Will run everything past Nathan at the next export

spec/models/participant_profile/ecf_spec.rb
```diff
@@ -334,4 +334,47 @@
       expect(subject.record_to_serialize_for(lead_provider:)).to eq(subject.induction_records.latest.reload)
     end
   end
+
+  describe "#relevant_induction_record_for" do
+    let(:cpd_lead_provider) { create(:cpd_lead_provider, :with_lead_provider) }
+    let(:lead_provider) { cpd_lead_provider.lead_provider }
+    let(:cohort) { Cohort.next || create(:cohort, :next) }
```
cwrw, line 341: I think we can use current as now it's set to 2022?

app/controllers/delivery_partners/participants_controller.rb
```diff
@@ -62,5 +62,9 @@ def to_csv(hash)
         end
       end
     end
+
+    def filter_params
+      params.permit(:query, :role, :academic_year, :status).merge(delivery_partner:)
```
cwrw, line 67: I think we can permit the `delivery_partner_id` instead as it's already available? to avoid merging in an object to params?

app/services/participant_profile_status.rb
```diff
@@ -109,4 +110,8 @@ def nqt_plus_one?
   def participant_has_no_qts?
     ecf_participant_eligibility&.no_qts_reason?
   end
+
+  def training_status_withdrawn?
```
cwrw, line 114: could we add a test for this?

cwrw (commented):
looking good very minor comments

app/controllers/delivery_partners/participants_controller.rb
cwrw: nvm will destroy the serialiser performance otherwise

cwrw (approved):

app/services/participant_profile_status.rb
leandroalemao: Hi @cwrw Addressed in https://github.com/DFE-Digital/early-careers-framework/pull/2920/commits/653fbb7cf679581af150c659d0e675c0e7f27203 tks 👍

spec/models/participant_profile/ecf_spec.rb
leandroalemao: Hi @cwrw Addressed in https://github.com/DFE-Digital/early-careers-framework/pull/2920/commits/653fbb7cf679581af150c659d0e675c0e7f27203 tks 👍

PR #2929, opened 2023-01-23
[CPDLP-1904] Fix induction records issue when start & end dates are off

leandroalemao:
### Context

Some induction records have start_date > end_date.

- Ticket: [CPDLP-1904](https://dfedigital.atlassian.net/browse/CPDLP-1904)

The idea is to add a validation on InductionRecord model to prevent the dates being off and also 

Found out that `Induction::ChangeInductionRecord` is copying over the end_date to the newly created induction record.

example

before:

old induction record has start_date = 01/09/2022 & end_date = 11/09/2022
new induction record will have start_date = 01/10/2022 & end_date = 11/09/2022

proposed changes:

do not copy end date over to the new induction record

after:
old induction record has start_date = 01/09/2022 & end_date = 11/09/2022
new induction record will have start_date = 01/10/2022 & end_date = nil

[CPDLP-1904]: https://dfedigital.atlassian.net/browse/CPDLP-1904?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/induction_record.rb
+++ b/app/models/induction_record.rb
@@ -31,6 +31,7 @@ class InductionRecord < ApplicationRecord
   belongs_to :preferred_identity, class_name: "ParticipantIdentity", optional: true
 
   validates :start_date, presence: true
+  validate :validate_sensible_dates
 
   enum induction_status: {
     active: "active",
@@ -137,5 +138,11 @@ private
 
   def update_analytics
     Analytics::UpsertECFInductionJob.perform_later(induction_record: self) if saved_changes?
+  end
+
+  def validate_sensible_dates
+    return if start_date.blank? || end_date.blank?
+
+    errors.add(:end_date, :before_start_date) if start_date > end_date
   end
 end
--- a/app/services/induction/change_induction_record.rb
+++ b/app/services/induction/change_induction_record.rb
@@ -21,6 +21,7 @@ class Induction::ChangeInductionRecord < BaseService
 
         induction_record.changing!(time_now)
         new_record.assign_attributes(default_attrs.merge(changes))
+        new_record.start_date = new_record.end_date if new_record.start_date && new_record.end_date && new_record.start_date > new_record.end_date
         new_record.save!
       end
     end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -218,6 +218,10 @@ en:
               cannot_change_from_accepted: Once accepted an application cannot change state
               cannot_change_from_rejected: Once rejected an application cannot change state
               declarations_exist: There are already declarations for this participant on this course, please ask provider to void and/or clawback any declarations they have made before attempting to reset the application.
+        induction_record:
+          attributes:
+            end_date:
+              before_start_date: "End date must be after start date"
 
   activemodel:
     errors:


app/services/induction/change_induction_record.rb
```diff
@@ -12,7 +12,7 @@ def call
           start_date: time_now,
           school_transfer: false,
         }
-        new_record = induction_record.dup
+        new_record = InductionRecord.new(induction_record.attributes.except("id", "end_date"))
```
cwrw, line 15: I was wondering how amending induction records for transferred participants will look now, will we risk setting their induction records to no end date although they already transferred and are active with another provider? (If we withdrew them for instance from the old provider)

tonyheadford, line 15: The intention was to copy _everything_ and apply the changes you supply to the call.  If you want to set/enforce `end_date` to `nil` then you can pass it in as part of the changes. I think this will break some assumptions made elsewhere if we don't do this.

app/models/induction_record.rb
```diff
@@ -138,4 +139,10 @@ def transferred?
   def update_analytics
     Analytics::UpsertECFInductionJob.perform_later(induction_record: self) if saved_changes?
   end
+
+  def validate_sensible_dates
```
tonyheadford, line 143: It would be really good to understand why this pattern happens and address that scenario (not in this PR).

app/services/induction/change_programme.rb
```diff
@@ -3,7 +3,13 @@
...
+      Induction::ChangeInductionRecord.call(
+        induction_record: current_induction_record,
+        changes: {
+          induction_status: :changed,
+          end_date:,
+        },
+      )
 
       Induction::Enrol.call(participant_profile:,
                             induction_programme: new_induction_programme,
```
tonyheadford, line 15: This will duplicate the current IR and give the new record a changed status, so 2 IRs with changed status and then add a new IR during the Enrol which doesn't seem right to me.

app/services/induction/change_preferred_email.rb
```diff
@@ -3,7 +3,7 @@
 class Induction::ChangePreferredEmail < BaseService
   def call
     Induction::ChangeInductionRecord.call(induction_record:,
-                                          changes: { preferred_identity: })
+                                          changes: { preferred_identity:, end_date: nil })
```
tonyheadford, line 6: What if this IR had a future `end_date` e.g. the participant was transferring, this would remove their `end_date`?  Same issue with the `ChangeMentor`

app/services/induction/change_induction_record.rb
```diff
@@ -21,6 +21,7 @@ def call
 
         induction_record.changing!(time_now)
         new_record.assign_attributes(default_attrs.merge(changes))
+        new_record.start_date = new_record.end_date if new_record.start_date && new_record.end_date && new_record.start_date > new_record.end_date
```
cwrw, line 24: Do we risk setting induction records out of order if we are setting the previous induction record `end_date` to `induction_record.changing!(time_now)`, while the new induction record `end_date` and `start_date` might be in the past?

peteryates: There's quite a lot going on in this line too, worth setting a variable?

cwrw:
Do we want to close this PR until the induction record investigation is complete?

leandroalemao:
Closing this for now until https://dfedigital.atlassian.net/browse/CPDLP-2017 is completed.

PR #2936, opened 2023-01-25
Add lead_mentor and itt_provider to api and app table

shjohnson:
### Context
We now have two new fields in the NPQ reg, these need to be stored for the lead providers

- Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-834

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -642,7 +642,7 @@ ActiveRecord::Schema.define(version: 2023_01_25_182955) do
     t.boolean "sparsity_uplift"
     t.boolean "pupil_premium_uplift"
     t.uuid "delivery_partner_id"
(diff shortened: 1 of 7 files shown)


db/schema.rb
```diff
@@ -640,7 +642,7 @@
     t.boolean "sparsity_uplift"
     t.boolean "pupil_premium_uplift"
     t.uuid "delivery_partner_id"
-    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY ((ARRAY['submitted'::character varying, 'eligible'::character varying, 'payable'::character varying, 'paid'::character varying, 'clawed_back'::character varying])::text[]))"
+    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY (ARRAY[('submitted'::character varying)::text, ('eligible'::character varying)::text, ('payable'::character varying)::text, ('paid'::character varying)::text, ('clawed_back'::character varying)::text]))"
```
Lockyy, line 645: These ones are what I mean, I didn't dig into it too much but this ended up breaking the specs for my PR for 713 so worth reverting these index related changes un-related to the migration.

shjohnson: Missed this! Thanks 🍏

Lockyy (approved):

PR #2937, opened 2023-01-26
Add seeding step for SIT nominations

peteryates:
### Context

This change should make it possible for people to test out the SIT journey from the beginning. We do it by seeding some blank schools and generating a notification email. This is usually where it gets complex because we need to pick that email up from Notify.

~~We need an easier way for non-devs to see paths/tokens they can use to start the SIT journey from the beginning, writing them to `public/seed_log.txt` seems like a good solution, but in the longer term being able to sign into the admin panel and click a button would probably be better.~~

The logging functionality was removed as @tonyheadford has a better solution on the way!

### DIFF
--- a/db/new_seeds/base/add_sit_nominations.rb
+++ b/db/new_seeds/base/add_sit_nominations.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+FactoryBot
+  .create_list(:seed_school, 15)
+  .map { |school| NewSeeds::Scenarios::School::InviteSchool.new(school:).invite! }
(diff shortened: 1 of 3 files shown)


db/new_seeds/base/add_sit_nominations.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+FactoryBot
+  .create_list(:school, 15)
```
tonyheadford, line 4: `:seed_school` ?

tonyheadford (approved):
Looks good just the one question

PR #2951, opened 2023-01-31
Add induction record created_at to analytics data

peteryates:
### Context

The creation timestamp for induction records is useful for the analytics team. The existing `created_at` timestamp on the analytics table is the creation date of the analytics record, not the induction record itself.


tonyheadford (approved):
LGTM

PR #2960, opened 2023-02-02
[CST-1240] Standardise retrieving 'relevant' induction records

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-1240)
Tech debt ticket: attempt to standardise on fetching the "relevant" induction record.

This PR is a discussion point / spike into this.

Please add comments, suggest more use-cases etc.

Adds a service class `Induction::FindBy` to encapsulate finding the most relevant induction record in the situation.


app/services/induction/find_relevant_to.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    query = query.current if current_not_latest_record
+    query.latest
+  end
+
+private
+
+  attr_reader :participant_profile, :lead_provider, :delivery_partner, :schedule, :school, :date_range, :current_not_latest_record
+
+  def initialize(participant_profile:, lead_provider: nil, delivery_partner: nil, schedule: nil, school: nil, date_range: nil, current_not_latest_record: false)
```
peteryates, line 20: This is great, knowing you at least need a `participant_profile` but can hone the results further by adding more criteria makes total sense.

app/validators/induction_record_validator.rb
```diff
@@ -21,16 +21,19 @@ def has_an_induction_record_for_given_provider(record)
...
+    #   .joins(:participant_profile, induction_programme: { school_cohort: [:cohort], partnership: [:lead_provider] })
+    #   .where(participant_profile: record.participant_profile)
+    #   .where(
+    #     induction_programme: {
+    #       partnerships: { lead_provider: record.cpd_lead_provider.lead_provider },
+    #     },
+    #     schedule: record.schedule,
+    #   )
+    #   .order(start_date: :desc)
+    #   .exists?
```
peteryates, line 37: So much cleaner  👌🏽

peteryates (approved):
Huge improvement and written in a really easy to understand way. Can't see a single downside to this approach.

Not 100% convinced by the name `Induction::FindRelevantTo` though 😄 

Could `Induction::FindBy` work?

tonyheadford:
> Huge improvement and written in a really easy to understand way. Can't see a single downside to this approach.
> 
> Not 100% convinced by the name `Induction::FindRelevantTo` though 😄
> 
> Could `Induction::FindBy` work?

Was originally going to go with `Induction::Find` but either would be better I agree 😃

app/services/induction/find_relevant_to.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class Induction::FindRelevantTo < BaseService
+  def call
+    query = participant_profile.induction_records
+
+    query = add_provider_to(query:) if lead_provider.present? || delivery_partner.present?
```
cwrw, line 7: I wonder if we could prioritise `end_date_null`/`active` records regardless of the `current_not_latest_record` generally?

tonyheadford: Could you explain a bit more how you'd like this to work?

cwrw: chatted in person but for posterity: when prioritising records, we could also add `end_date: nil` to the list. So first we look for any records where the `end_date` is `nil`, then ones with the latest start date, and then created at etc.

db/new_seeds/scenarios/schools/fip_school.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module NewSeeds
+  module Scenarios
+    module Schools
+      class FipSchool
```
peteryates, line 6: As we're creating the school, partnership, cohort and programme here should the name be a bit more descriptive?

spec/models/induction_record_spec.rb
```diff
@@ -268,13 +268,15 @@
     describe "ordering" do
       describe ".oldest_first" do
         it "orders by created_at asc" do
-          expect(described_class.oldest_first.to_sql).to match(%(ORDER BY "induction_records"."created_at" ASC))
+          # expect(described_class.oldest_first.to_sql).to match(%(ORDER BY "induction_records"."created_at" ASC))
```
peteryates, line 271: ✂️

spec/models/induction_record_spec.rb
```diff
@@ -268,13 +268,15 @@
...
+          expect(described_class.oldest_first.to_sql).to match(%(ORDER BY CASE WHEN induction_records.end_date IS NULL THEN 1 ELSE 0 END, "induction_records"."start_date" ASC, "induction_records"."created_at" ASC))
         end
       end
 
       describe ".newest_first" do
         it "orders by created_at desc" do
-          expect(described_class.newest_first.to_sql).to match(%(ORDER BY "induction_records"."created_at" DESC))
+          # expect(described_class.newest_first.to_sql).to match(%(ORDER BY "induction_records"."created_at" DESC))
```
peteryates, line 278: ✂️

db/new_seeds/scenarios/schools/fip_school.rb
tonyheadford: Yes, struggling to think of a good name, it's a valid/populated/complete FIP school but without participants (or a SIT actually, perhaps that should be added).... anything spring to mind? `FipSchoolWithProvider` or `FipSchoolWithProgramme`?  I'm not sure they'd make sense to say it was a FIP School without those things though... 🤔

tonyheadford: or maybe the other bits could be added by chain methods? e.g. `FipSchool.with_partnership.with_programme.with_sit` ?

peteryates: This sounds good to me, so long as the additions aren't interdependent (I don't think they are) and can be added in any order/combination.

tonyheadford: Yes working on generalising this into just `School` with chainable options

peteryates (approved):
Looks great, keen to see this go in!

db/new_seeds/scenarios/schools/school.rb
```diff
@@ -0,0 +1,117 @@
...
+          school_attributes = { name:, urn: }.compact
+          @school = FactoryBot.create(:seed_school, **school_attributes)
+
+          self
+        end
+
+        ###
+        # adds a user with an induction coordinator profile for the school
+        ###
+        def with_an_induction_tutor(full_name: nil, email: nil)
```
peteryates, line 26: Making this modular is definitely going to be a win, there are so many places this could come in handy, in seeds or tests.

app/services/change_schedule.rb
```diff
@@ -85,7 +85,7 @@ def cohort
   def relevant_induction_record
     return if user.blank? || participant_profile.blank?
 
-    @relevant_induction_record ||= participant_profile.latest_induction_record_for(cpd_lead_provider:)
```
cwrw, line 88: chatted in person: I was wondering if instead of changing the interface to call this new method, we can instead amend the implementation of `latest_induction_record_for`, and keep the interface everywhere the same? Also maybe we can consolidate `relevant_induction_record` and `latest_induction_record_for` if that's possible (not sure it is though)

tonyheadford: @cwrw  I've added these in, also when retrieving the schedule too.  I left a couple of `FIXME` comments in there too around confirming what should really happen if `cpd_lead_provider` is nil in those methods, if you know whether that's legit or not as it came up in the specs.

app/models/participant_profile/ecf.rb
```diff
@@ -75,10 +75,8 @@ def ecf?
     end
 
     def latest_induction_record_for(cpd_lead_provider:)
-      induction_records
-        .joins(induction_programme: { partnership: { lead_provider: [:cpd_lead_provider] } })
-        .where(induction_programmes: { partnerships: { lead_providers: { cpd_lead_provider: } } })
-        .latest
+      # FIXME: should this method return nil if cpd_lead_provider.nil? or call FindBy regardless (FindBy won't consider a nil lead_provider)
```
cwrw, line 78: Good question I think we should return `nil` to preserve the previous behaviour  and not call `FindBy` regardless

app/models/participant_profile/ecf.rb
```diff
@@ -98,36 +96,18 @@ def policy_class
     end
 
     def relevant_induction_record(lead_provider:)
-      induction_records
-        .joins(induction_programme: { school_cohort: [:cohort], partnership: [:lead_provider] })
-        .where(induction_programme: { partnerships: { lead_provider: } })
-        .where(induction_programme: { school_cohorts: { cohort: Cohort.where(start_year: 2021..) } })
-        .order(start_date: :desc)
-        .first
+      # FIXME: should this method return nil if lead_provider.nil? or call FindBy regardless (FindBy won't consider a nil lead_provider)
```
cwrw, line 99: same here as above

app/models/participant_profile/ecf.rb
cwrw: Could we also reinstate everywhere we swapped out `latest_induction_record_for` for `FindBy` to make sure we're  using this guard/wrapper?

app/models/participant_profile/ecf.rb
```diff
@@ -98,36 +96,18 @@ def policy_class
     end
 
     def relevant_induction_record(lead_provider:)
```
cwrw, line 98: could we consolidate `relevant_induction_record` and `latest_induction_record_for`, I think the only difference is the cohort, but does it matter 🤔

cwrw: checking prod we have no induction records in cohort 2020, so maybe we can always use the cohort filter?

app/models/participant_profile/ecf.rb
tonyheadford: Thanks for confirming, I didn't change the interface so all that should be needed is a guard in these methods

app/models/participant_profile/ecf.rb
tonyheadford: Yes nothing relevant in 2020, I looked before starting this. I've changed the internal calls in `ecf.rb` to use `relevant_induction_record` instead of ` latest_induction_record_for` but kept that interface as it looks like finance stuff uses it.

cwrw (approved):
thanks for addressing all comments @tonyheadford, looks great. I tested the main new methods and all good 👍

PR #2974, opened 2023-02-08
Add TRN and Cohort to admin participant page header

laurentdfe:
### Context

TRN is Teacher Reference Number. It's a unique number given to all qualified teachers

Cohort is the year the person began their ECF.

We already show the data in the admin pages.

This PR surfaces them so they're visible from everywhere.


peteryates (approved):
Looks great 👌🏽

PR #2977, opened 2023-02-09
Remove the dependency on gem open_api-rswag-specs

leoapost:
### Context
The `open_api-rswag-specs` gem appears to be no longer maintained. 

It also prevents us from upgrading to the next major version of Rails.

- Ticket: CST-1299

- Replace `open_api-rswag-specs` with parent fork `rswag/rswag`. 

- Monkey patch the the rswag-specs's `schema` method, as the gem doesn't seem to have a way to set different content type in multiple responses.

- Remove the `request_body` blocks. `rswag-specs` uses the body parameters to build the `requestBody`.


swagger/v3/api_spec.json
```diff
@@ -211,1181 +211,14 @@
             }
           }
         }
-      },
```
cwrw, line 214: I think because API v3 has not been built fully these specs are removed as the endpoints don't exist. We however need those here to generate the draft API spec we are sharing with providers. Do you mind adding those back please?

swagger/v1/api_spec.json
```diff
@@ -1881,285 +1894,8 @@
           }
         }
       },
-      "ECFParticipantDeclarationAttributes": {
```
cwrw, line 1884: The declaration API spec is being removed as we combine those in our API endpoint, however in the documentation we have separated those to indicate what is required in ECF vs NPQ, do you mind adding those back please in v1/v2?

leoapost: @cwrw sure, I will put them back for now, but these were added manually directly in the `api_spec.json` files and they will be removed again the next time we run the `swaggerize` rake command.

If they need to stick around, then will need [spec/docs](https://github.com/DFE-Digital/early-careers-framework/tree/develop/spec/docs).

cwrw (approved):
👍 thanks @leoapost

leoapost:
I rebased my branch with the new default branch, main, to ensure the Github actions and workflows work as expected.

PR #2979, opened 2023-02-10
[CPDLP-1955] Create participant outcome api request model

leandroalemao:
### Context

Some new required models/fields are required for TRA outcome integration.

- Ticket: [CPDLP-1955](https://dfedigital.atlassian.net/browse/CPDLP-1955)

- Add the new fields qualifications_api_status and sent_to_qualifications_api_at  to the outcomes table; (think of better names that indicate which API). This will help us re-queue jobs and capture the timestamp of a successful request to TRA. 

- Add the following new model: ParticipantOutcomeAPIRequest;

[CPDLP-1955]: https://dfedigital.atlassian.net/browse/CPDLP-1955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/db/migrate/20230210115830_add_qualified_teachers_api_status_and_sent_to_qualified_teachers_api_at_to_participant_outcomes.rb
+++ b/db/migrate/20230210115830_add_qualified_teachers_api_status_and_sent_to_qualified_teachers_api_at_to_participant_outcomes.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddQualifiedTeachersApiStatusAndSentToQualifiedTeachersApiAtToParticipantOutcomes < ActiveRecord::Migration[6.1]
+  def change
+    add_column :participant_outcomes, :qualified_teachers_api_request_successful, :boolean
+    add_column :participant_outcomes, :sent_to_qualified_teachers_api_at, :datetime
+  end
+end
--- a/app/models/participant_outcome/npq.rb
+++ b/app/models/participant_outcome/npq.rb
@@ -8,6 +8,8 @@ class ParticipantOutcome::NPQ < ApplicationRecord
   private_constant :VALID_STATES
 
   belongs_to :participant_declaration, class_name: "ParticipantDeclaration::NPQ"
+  has_many :participant_outcome_api_requests, foreign_key: :participant_outcome_id
+
   enum state: VALID_STATES.index_with(&:to_s)
 
   validates :state, presence: true
--- a/app/models/participant_outcome_api_request.rb
+++ b/app/models/participant_outcome_api_request.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class ParticipantOutcomeApiRequest < ApplicationRecord
+  belongs_to :participant_outcome, class_name: "ParticipantOutcome::NPQ"
+end
--- a/db/migrate/20230210120533_create_participant_outcome_api_requests.rb
+++ b/db/migrate/20230210120533_create_participant_outcome_api_requests.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class CreateParticipantOutcomeApiRequests < ActiveRecord::Migration[6.1]
+  def change
+    create_table :participant_outcome_api_requests do |t|
+      t.string :request_path
+      t.integer :status_code
+      t.jsonb :request_headers
+      t.jsonb :request_body
+      t.jsonb :response_body
+      t.jsonb :response_headers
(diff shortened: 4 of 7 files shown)


db/migrate/20230210115830_add_qualified_teachers_api_status_and_sent_to_qualified_teachers_api_at_to_participant_outcomes.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddQualifiedTeachersApiStatusAndSentToQualifiedTeachersApiAtToParticipantOutcomes < ActiveRecord::Migration[6.1]
+  def change
+    add_column :participant_outcomes, :qualified_teachers_api_status, :boolean, null: false, default: false
+    add_column :participant_outcomes, :sent_to_qualified_teachers_api_at, :boolean, default: false
```
cwrw, line 6: could this column be set to a timestamp?

leandroalemao: all done 🤦‍♂️ tks @cwrw

db/migrate/20230210115830_add_qualified_teachers_api_status_and_sent_to_qualified_teachers_api_at_to_participant_outcomes.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddQualifiedTeachersApiStatusAndSentToQualifiedTeachersApiAtToParticipantOutcomes < ActiveRecord::Migration[6.1]
+  def change
+    add_column :participant_outcomes, :qualified_teachers_api_status, :boolean, null: false, default: false
```
cwrw, line 5: I wonder if we can find a better name for this, it took me a while to remember what this value was capturing, which was if a participant outcome request to the API was successful and didn't result in an error 🤔 
maybe `qualified_teachers_api_successful`? Let me know your thoughts

peteryates: or `qualified_teachers_api_request_succeeded_at` with a timestamp (which may or may not be useful to know)?

cwrw: I think we were hoping to avoid a timestamp on success because a failure would result in `nil` values, so we couldn't capture if we needed to re-retry this job or not. We were hoping to avoid re-queueing genuine API failures due to for instance the TRN not being available on the TRA side.

The other timestamp column hopefully captures a queue attempt, at the job level. 

So the main data we're trying to capture is:
- Queued to send to TRA
- Response received from TRA successfully or unsuccessfully

Response details/timestamps are captured in the other model hopefully for auditing
Let me know your thoughts

mooktakim: It might be good to keep "tra" in the name as that is what we use to refer to this API when we speak generally.

leandroalemao: Tks guys made the suggested changes... 🔝 😄 👍

db/migrate/20230210115830_add_qualified_teachers_api_status_and_sent_to_qualified_teachers_api_at_to_participant_outcomes.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddQualifiedTeachersApiStatusAndSentToQualifiedTeachersApiAtToParticipantOutcomes < ActiveRecord::Migration[6.1]
+  def change
+    add_column :participant_outcomes, :qualified_teachers_api_request_successful, :boolean, null: false, default: false
```
cwrw, line 5: could we remove the `null:false` and default, as we would like to distinguish between the following values:

true: response successful
false: error in response
`nil`: has not been sent to TRA

leandroalemao: All sorted tks @cwrw 👍

cwrw (approved):
🙌 thanks @leandroalemao!

PR #2980, opened 2023-02-10
Create job to handle notify callbacks offline

peteryates:
### Context

We don't need to deal with the notify callbacks immediately, instead create jobs and let them be picked up in due course.


leoapost (approved):
Looks good to me

PR #2981, opened 2023-02-13
Critical stats task

mattonfoot:
### Context

It should be possible to see the historical events or changes to a CPD record over time in order to help determine why a record is in the state is is and where we might go to correct or undo any actions that misreported information.

Rake task creates a lists of CPD record change events that can be used to build up metrics of frequency and type of changes made over time

a rake task that can be run in small batches to process user profiles into event history lists

### Guidance to review

login as an admin user and proceed to the participants dashboard, from here you can bring up any participants screen and replace the last item in the path with `/event_history`.

bring up the rails console in your chosen environment and run a query to find a `User` or `ParticipantProfile` object. Call `ParticipantEventHistory.from_user` or `ParticipantEventHistory.from_participant_profile` and access the `events` attribute to see a full list of events created for a record.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -642,7 +642,7 @@ ActiveRecord::Schema.define(version: 2023_02_10_120533) do
     t.boolean "sparsity_uplift"
     t.boolean "pupil_premium_uplift"
     t.uuid "delivery_partner_id"
-    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY ((ARRAY['submitted'::character varying, 'eligible'::character varying, 'payable'::character varying, 'paid'::character varying, 'clawed_back'::character varying])::text[]))"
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -642,7 +642,7 @@
     t.boolean "sparsity_uplift"
     t.boolean "pupil_premium_uplift"
     t.uuid "delivery_partner_id"
-    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY ((ARRAY['submitted'::character varying, 'eligible'::character varying, 'payable'::character varying, 'paid'::character varying, 'clawed_back'::character varying])::text[]))"
+    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY (ARRAY[('submitted'::character varying)::text, ('eligible'::character varying)::text, ('payable'::character varying)::text, ('paid'::character varying)::text, ('clawed_back'::character varying)::text]))"
```
peteryates, line 645: I don't think this file should be included in the PR, I think it's caused by running migrations against a slightly different version of PostgreSQL than others are using.

peteryates (commented):
Looks good overall, works fine for me in the review app!

I think we just need to lose the updtes to `schema.rb` and we're good to go.

db/schema.rb
mattonfoot: bugger missed this, thanks for the spot

peteryates (approved):
Looks good. If this page was in the nav I'd probably be more worried about performance but as it's hidden and probably used sparingly I'm happy for it to go in.

PR #2986, opened 2023-02-13
Fix Selenium deprecation warning

slorek:
Fixes the following warning when running feature specs:

`WARN Selenium [DEPRECATION] [:capabilities] The :capabilities parameter for Selenium::WebDriver::Chrome::Driver is deprecated. Use :options argument with an instance of Selenium::WebDriver::Chrome::Driver instead.`

### DIFF
--- a/spec/features_helper.rb
+++ b/spec/features_helper.rb
@@ -8,13 +8,14 @@ require "site_prism"
 require "site_prism/all_there" # Optional but needed to perform more complex matching
 
 Capybara.register_driver :chrome_headless do |app|
-  options = ::Selenium::WebDriver::Chrome::Options.new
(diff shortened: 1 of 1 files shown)


cwrw (approved):
nice one 🙌 for the deploy failure, I suggest closing the PR to tear down the review app and reopening. If this doesn't work might need to reopen under a new branch (this is an issue we also need to look into)

cwrw:
> Thanks, I don't think the review app is a blocker to merging in this particular case. I should be able to clean up the PaaS resources once I get access, however.

I agree, it shouldn't be, however I would prefer we get a green build regardless as sometimes there are subtle dependencies which might affect the deploy. I might take you up on the PaaS resource cleanup as there are a few apps still hanging around we need to delete so keen on seeing how to do that safely when you get the chance.

PR #2987, opened 2023-02-13
Improve performance of School Participants request

ltello:
### Context

  The analysis of the incident that happened last week when the service crashed after heavy traffic due to reminders sent for SITs to make sure any of their ECTs have a mentor associated, unveiled the performance issues of the participants dashboard for a specific cohort (`GET schools/xxxx/cohort/xxxx/participants requests`).

This view displays the ECF participants of a school on a specific cohort grouped by different categories.
To compute it the controller action calls `CocSetParticipantCategories` service class.
We have identified a couple of issues that impact the performance of the related requests:

-  `CocSetParticipantCategories` retrieves the groups of participants by triggering ~10 heavy queries to the database with complex joined tables.
-  The rendering of the page triggers extra N+1 queries to the database due to some missing eager_loaded associations on the data structure of participants returned.

- Refactor `CocSetParticipantCategories` so that it triggers only one complex join query to the database. The categorisation of the records retrieved (1..75) is done in memory. That will save ~20 heavy hits to the db.
- Eager load missing associations on the retrieval so that the page can be rendered very efficiently with no extra hits to the db.

### Guidance to review
- `CocSetParticipantCategories`: refactored to trigger a single query  memoized in `induction_records` to retrieve all the participants induction records no matter their type (ECT or Mentor). 
This class computes  both types of participants at once.
This means the controller action would have to call this class only once to get both types of participants categorized.

- Schools::ParticipantsController index action and view adjusted to use the new instance variable (`@categories`) with all the participants data.


app/services/coc_set_participant_categories3.rb
```diff
@@ -0,0 +1,186 @@
...
+
+  def current_induction_records
+    @current_induction_records ||= InductionRecordPolicy::Scope.new(
+      user,
+      school_cohort
+        .induction_records
+        .current_or_transferring_in_or_transferred
+        .eager_load(induction_programme: %i[core_induction_programme lead_provider delivery_partner],
+                    participant_profile: %i[user ecf_participant_eligibility ecf_participant_validation_data])
+        .order("users.full_name"),
```
peteryates, line 65: 🔥

app/controllers/schools/participants_controller.rb
```diff
@@ -12,8 +12,7 @@ class Schools::ParticipantsController < Schools::BaseController
 
   def index
     if FeatureFlag.active?(:change_of_circumstances)
-      @mentor_categories = CocSetParticipantCategories.call(@school_cohort, current_user, ParticipantProfile::Mentor)
-      @ect_categories = CocSetParticipantCategories.call(@school_cohort, current_user, ParticipantProfile::ECT)
+      @mentor_categories, @ect_categories = CocSetParticipantCategories3.call(@school_cohort, current_user)
       @transferring_in = @ect_categories.transferring_in + @mentor_categories.transferring_in
       @transferring_out = @ect_categories.transferring_out + @mentor_categories.transferring_out
       @transferred = @ect_categories.transferred + @mentor_categories.transferred
```
peteryates, line 18: Here it feels like it could be made a bit easier to follow if we did something like:

```ruby
  all_categories = CocSetParticipantCategories3.call(@school_cohort, current_user)

  @mentor_categories = all_categories.mentor_categories
  @ect_categories = all_categories.ect_categories
  @transferring_in = all_categories.transferring_in
end
```

Or perhaps those assignments don't even need to happen in the controller, but can be called from `@all_categories` in the view?

peteryates (commented):
Love it, definite speed improvements.

Comparison would be a touch easier if the calls to versions 1 and 2 were commented out in the controller.

I think we need a few pairs of eyes to make sure no bits of criteria that weren't covered by tests weren't lost, but having had a decent look through it looks fine to me.

Great stuff!

app/controllers/schools/participants_controller.rb
ltello: Good idea. Done.

app/services/coc_set_participant_categories2.rb
```diff
@@ -0,0 +1,121 @@
...
+      induction_record.participant_profile.contacted_for_info?
+    end
+  end
+
+  def current_induction_records
+    @current_induction_records ||= InductionRecordPolicy::Scope.new(
+      user,
+      school_cohort
+        .induction_records
+        .current_or_transferring_in_or_transferred
```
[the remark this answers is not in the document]
ltello: You are right. 
Renamed to  just `induction_records`. 
A comment above the method explains the type of record retrieved for each participant in the school cohort.

app/controllers/schools/participants_controller.rb
ltello: I wonder if we can remove the feature flag :change_of_circumstances. 
It seems it will stay active forever.

app/services/coc_set_participant_categories3.rb
ltello: Exactly! haha

app/controllers/schools/participants_controller.rb
peteryates: I'll add that as a tech debt ticket, I'd also like to see the back of it.

app/models/induction_record.rb
```diff
@@ -122,7 +141,7 @@ def matches_school_appropriate_body?
   end
 
   def transferring_in?
-    active_induction_status? && start_date > Time.zone.now
+    active_induction_status? && start_date > Time.zone.now && school_transfer
```
tonyheadford, line 144: I'm wondering if this didn't originally have the `school_transfer` bit so that it included newly qualified teachers starting at the school not just transfers from another school, but I'm not sure (or whether it is used anywhere)?

ltello: I see this is only used in this request code.

I have added the `& school_transfer` bit to match the method exactly to the existing scope: 
`  scope :transferring_in, -> { active_induction_status.merge(start_date_in_future.and(school_transfer)) }
`

peteryates (approved):
This looks great @ltello, more efficient and way tidier 👌🏽 

I think it needs a second set of eyes but I certainly can't see any lost functionality

app/models/induction_record.rb
tonyheadford: Thanks for checking 👍

tonyheadford (approved):
Looks great!

PR #2988, opened 2023-02-13
NPQ funding eligibility checks

peteryates:
### Context

This PR allows policy to change the funding status and add notes on an applicants application. 

- Create multiple controllers for listing NPQ application edge cases, editing and updating the funding status, eligible for funding and adding notes 
- Add manual review trait so we can seed examples
- Add applicant details to the show template
- Allow eligibility status to be updated
- Add two new statuses for EligibilityStatus

| Listing applicants | Viewing an applicant | Editing an applicant's status | Editing an applicant's notes |
| ----- | ----- | ----- | ----- |
| Screenshot 2023-03-16 at 09 50 51 | Screenshot 2023-03-16 at 09 51 24 | Screenshot 2023-03-16 at 09 52 30 | Screenshot 2023-03-16 at 09 53 09 |


app/views/admin/npq/applications/edge_cases/edit.html.erb
```diff
@@ -1,6 +1,6 @@
 <h1><%= @npq_application.user.full_name %></h1>
 
-<%= form_for(@npq_application, url: admin_npq_applications_non_school_registrant_path(@npq_application), method: "PATCH") do |form| %>
+<%= form_for(@npq_application, url: admin_npq_applications_edge_case_path(@npq_application), method: "PATCH") do |form| %>
```
peteryates, line 3: Yeah `edge_case` definitely makes more sense than `non_school` 👍🏽

app/views/admin/npq/applications/edge_cases/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+
+    table.body do |body|
+      @npq_applications.each do |application|
+        body.row do |row|
+          row.cell(text: govuk_link_to(application.user.full_name, admin_npq_applications_edge_case_path(application)))
+          row.cell(text: boolean_red_green_tag(application.eligible_for_funding))
+          row.cell(text: application.funding_eligiblity_status_code.humanize.downcase)
+          row.cell(text: application.employment_type.try(:humanize) || '-')
+          row.cell(text: application.employer_name || '-')
+          row.cell(text: application.employment_role.try(:humanize).downcase || '-')
```
Lockyy, line 27: This fails if employment_role is nil despite the try(:humanize) because you can't downcase nil.

app/views/admin/npq/applications/edge_cases/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+          row.cell(text: application.created_at)
```
Lockyy, line 28: This needs to be humanized with govuk_short, it outputs strings like `2023-01-30 15:30:18 UTC` at the moment.

app/views/admin/npq/applications/edge_cases/show.html.erb
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    sl.row do |row|
+      row.key(text: "Created at")
+      row.value(text: @npq_application.created_at)
+    end
+
+    sl.row do |row|
+      row.key(text: "Updated at")
+      row.value(text: @npq_application.updated_at)
```
Lockyy, line 76: Same as previous date output, needs humanizing.

app/controllers/admin/npq/applications/notes_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+        def update
+          npq_application = NPQApplication.find(params[:id])
+          npq_application.assign_attributes(note_params)
+
+          if npq_application.save
+            name = npq_application.participant_identity.user.full_name
+
+            flash[:success] = {
+              title: "#{name} updated",
+              content: "#{name} has a new note: '#{npq_application.notes.humanize.downcase}' ",
```
Lockyy, line 23: For particularly long notes this can get weird:

Screenshot 2023-03-21 at 14 24 49

We may want to just say "#{name} has a new note, see below for details." rather than inserting the entire note into the flash.

Lockyy (changes requested):
See comments, all failure paths through the update actions lead to 500 errors and the specs don't cover all the changes.

app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+        def update
+          npq_application = NPQApplication.find(params[:id])
+          npq_application.assign_attributes(eligible_for_funding_params)
+
+          case eligible_for_funding_params["eligible_for_funding"]
+          when "true"
+            npq_application.funding_eligiblity_status_code = :marked_funded_by_policy
+          when "false"
+            npq_application.funding_eligiblity_status_code = :marked_ineligible_by_policy
+          end
```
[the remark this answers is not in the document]
shjohnson: These are options already in the DB and show that they have had manual intervention.

app/models/npq_application.rb
```diff
@@ -40,6 +40,17 @@ class NPQApplication < ApplicationRecord
   }
 
   scope :with_targeted_delivery_funding_eligibility, -> { where(targeted_delivery_funding_eligibility: true) }
+  scope :does_not_work_in_school, -> { where(works_in_school: false) }
+  scope :does_not_work_in_childcare, -> { where(works_in_childcare: false) }
+  scope :not_eligible_for_funding, -> { where(eligible_for_funding: false) }
+  scope :no_institution, -> { where(funding_eligiblity_status_code: "no_institution") }
```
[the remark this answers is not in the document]
shjohnson: This was an old scope that needs removing.

app/views/layouts/admin.html.erb
```diff
@@ -12,7 +12,7 @@
     <%= component.nav_item(path: "/admin/administrators") do %>
       Admin users
     <% end %>
-    <%= component.nav_item(path: admin_npq_applications_exports_path, selected: on_admin_npq_application_page?) do %>
+    <%= component.nav_item(path: admin_npq_applications_edge_cases_path, selected: on_admin_npq_application_page?) do %>
```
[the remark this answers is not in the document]
shjohnson: This was requested by Chesca on Monday. I thought Applications but edge cases was requested.

app/controllers/admin/npq/applications/eligibility_status_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+        StatusOption = Struct.new(:value, :label)
+
+        skip_after_action :verify_authorized
+        skip_after_action :verify_policy_scoped
+
+        def edit
+          @status_options = [
+            StatusOption.new(:marked_ineligible_by_policy, "Not eligible for funding"),
+            StatusOption.new(:awaiting_more_information, "Awaiting more information"),
+            StatusOption.new(:re_register, "Needs to re-register"),
```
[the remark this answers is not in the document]
shjohnson: Didnt quite get this as felt made more sense keeping in the controller and displaying the object via the view how we do in other instances. The flash messages where added later and read more humanised i think now?

app/controllers/admin/npq/applications/eligibility_status_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+        def update
+          npq_application = NPQApplication.find(params[:id])
+          npq_application.assign_attributes(eligiblity_status_params)
+
+          if npq_application.save
+            name = npq_application.participant_identity.user.full_name
+
+            flash[:success] = {
+              title: "#{name} updated",
+              content: "#{name} has been marked '#{npq_application.funding_eligiblity_status_code.humanize.downcase}'",
```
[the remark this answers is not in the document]
shjohnson: Not sure what you meant exactly as if i move into the transaltion file i will still have to create the translations in two places right? ie the edit and update unless i select the translation higher up so instead of `....npq.applications.eligibility_status.edit.x` it would be `....npq.applications.eligibility_status.x` I think what it says above makes sense?

app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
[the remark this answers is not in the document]
Lockyy: What I mean is why aren't we using these in the form then, rather than true/false and then needing this translation layer. If the form just used marked_funded_by_policy and marked_ineligible_by_policy as the values on the radios then it could be fed directly through.

app/controllers/admin/npq/applications/eligibility_status_controller.rb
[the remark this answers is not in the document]
Lockyy: Yeah but we still shouldn't have localisation in the controller, that's not where localisation goes, which is what this is. If we move that to where it should go (en.yml) then we just have an array of option values passed into `govuk_collection_radio_buttons`, which yes is how we do it elsewhere but that happens via a Form Object, I haven't seen anywhere else we use Structs in controllers to build up radio options.

app/controllers/admin/npq/applications/eligibility_status_controller.rb
```diff
@@ -6,43 +6,47 @@ module Applications
...
+            StatusOption.new(:re_register, t(".re_register")),
           ]
-
-          @npq_application = NPQApplication.find(params[:id])
         end
 
         def update
-          npq_application = NPQApplication.find(params[:id])
-          npq_application.assign_attributes(eligiblity_status_params)
+          @npq_application = NPQApplication.find(params[:id])
```
Lockyy, line 22: If we've set up assign_npq_application then we can drop this duplicate loading right?

app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
```diff
@@ -6,48 +6,53 @@ module Applications
...
+            StatusOption.new(true, t(".true")),
+            StatusOption.new(false, t(".false")),
           ]
-          @npq_application = NPQApplication.find(params[:id])
         end
 
         def update
-          npq_application = NPQApplication.find(params[:id])
-          npq_application.assign_attributes(eligible_for_funding_params)
+          @npq_application = NPQApplication.find(params[:id])
```
Lockyy, line 21: Same as previous, we can drop this with assign_npq_application in place right?

app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
```diff
@@ -6,48 +6,53 @@ module Applications
...
-            redirect_to admin_npq_applications_edge_case_path(npq_application)
           else
             flash[:alert] = {
               title: "#{name} not updated",
               content: "#{name} failed to update",
             }
 
-            render(:edit)
           end
+          redirect_to admin_npq_applications_edge_case_path(@npq_application)
```
Lockyy, line 45: Same question as above, should we be kicking the user off the form on a failure?

app/controllers/admin/npq/applications/notes_controller.rb
```diff
@@ -4,36 +4,40 @@ module Admin
...
-          @npq_application = NPQApplication.find(params[:id])
-        end
+        before_action :assign_npq_application
+
+        def edit; end
 
         def update
-          npq_application = NPQApplication.find(params[:id])
-          npq_application.assign_attributes(note_params)
+          @npq_application = NPQApplication.find(params[:id])
```
Lockyy, line 14: Same as previous, do we need this duplicate loading with assign_npq_application in place?

app/controllers/admin/npq/applications/notes_controller.rb
```diff
@@ -4,36 +4,40 @@ module Admin
...
             }
-            redirect_to admin_npq_applications_edge_case_path(npq_application)
           else
             flash[:alert] = {
               title: "#{name} not updated",
               content: "#{name} failed to update",
             }
-            render(:edit)
           end
+          redirect_to admin_npq_applications_edge_case_path(@npq_application)
```
Lockyy, line 30: Pushing the user here on a failure would mean that the form contents got wiped out, if we render(:edit) instead then the user wouldn't lose everything they've written out if the record fails to save, which might be really important information!

spec/requests/admin/npq/applications/notes_controller_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Admin::NPQ::Applications::Notesapp/controllers/admin/npq/applications/notes_controller.rbController", :with_default_schedules, type: :request do
```
Lockyy, line 5: Copy and paste error in the spec title?

app/controllers/admin/npq/applications/eligibility_status_controller.rb
```diff
@@ -6,43 +6,47 @@ module Applications
...
-            redirect_to admin_npq_applications_edge_case_path(npq_application)
           else
             flash[:alert] = {
               title: "#{name} not updated",
               content: "#{name} failed to update",
             }
 
-            render(:edit)
           end
+          redirect_to admin_npq_applications_edge_case_path(@npq_application)
```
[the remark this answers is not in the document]
shjohnson: Well i guess they could be kept on that page allowed to try again? although if for some reason it keeps failing then could be a bit annoying? i dont see it as issue sending them back to the edit page and then they can either try again or do something else.

app/controllers/admin/npq/applications/eligibility_status_controller.rb
shjohnson: mistake yes will remove

app/controllers/admin/npq/applications/eligibility_status_controller.rb
```diff
@@ -6,43 +6,47 @@ module Applications
...
 
-          if npq_application.save
-            name = npq_application.participant_identity.user.full_name
+          if @npq_application.save
+            name = @npq_application.participant_identity.user.full_name
 
             flash[:success] = {
               title: "#{name} updated",
-              content: "#{name} has been marked '#{npq_application.funding_eligiblity_status_code.humanize.downcase}'",
+              content: "#{name} has been marked '#{@npq_application.funding_eligiblity_status_code.humanize.downcase}'",
```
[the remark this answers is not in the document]
shjohnson: I mean i was following the prototype to be honest. Regarding the localisation could be further decision.

app/controllers/admin/npq/applications/notes_controller.rb
shjohnson: thats a fair comment.

shjohnson:
@Lockyy I think ive corrected all your comments as we discussed, please let me know if any issues and ive tried to make it as user friendly to the user for the error messages and tested that the note does stay on the page and your not redirected like the other options with addtional tests. Also the strings added to the localisation file. I had to keep the case statement because we are changing another field if a certain option is changed on the eligible for funding controller which ive tided into a method to keep more obvious. I didnt change to the marked as funded as we havea validation on the elgible for funding field which seems silly to change and then dynamically change to a boolean.

spec/requests/api/v1/npq_profiles_spec.rb
```diff
@@ -194,52 +194,51 @@ def owner
       let(:json) { json_hash.to_json }
 
       it "creates the npq validation data" do
-        Timecop.freeze(Date.new(2023, 3, 20)) do
```
Lockyy: Why was this removed? I think it was just added as a part of the cohorts changes.

Lockyy (approved):

Lockyy (approved):

cwrw (approved):

PR #2992, opened 2023-02-14
Ensure preferred identities are created in the initial seed

peteryates:
### Context

Local dev environments and review apps sometimes crash when viewing a participant from a SIT's point of view because the `preferred_identity` hasn't been set on an induction record.

This PR

* ensures seeded records have a preferred identity
* displays no summary list row (instead of crashing) when there's no email address present

- Don't blow up when preferred_identity is missing
- Ensure we're setting preferred_identity in seeds
- Add induction record 'preferred_identity' trait


peteryates:
Closing as this has been superseded.

PR #2996, opened 2023-02-14
Change target branch across all workflows

peteryates:
### Context

Conventionally we use `main` for the default branch. When working on multiple projects having to remember ECF uses `develop` is a bit of a hassle.

### Changes

- switch `develop` for `main` in the workflows that are triggered by activity on a branch or clone the repo


peteryates:
Not sure why the deploy and smoke-test outcomes aren't being reported 🤔

PR #2999, opened 2023-02-15
Implement new attributes for ITT lead mentors in API responses

slorek:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2010
- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2033

* Implements the `itt_provider` and `lead_mentor` attributes for NPQ applications in API responses
* Adds specifications for two new properties
* Added a link to the reference list referred to following discussion in the ticket
* Updated release notes

### DIFF
--- a/docs/source/release-notes.html.erb
+++ b/docs/source/release-notes.html.erb
@@ -9,7 +9,7 @@ weight: 8
   If you have any questions or comments about these notes, please contact DfE via Slack or email.
 </p>
 
-<h2 class="govuk-heading-l" id="22-2-2023">22nd February 2023</h2>
+<h2 class="govuk-heading-l" id="23-2-2023">23rd February 2023</h2>
 <p class="govuk-body-m">
   Initial teacher training (ITT) lead mentors who are not employed by schools are now eligible for DfE funding.
   </br>


cwrw (commented):
looks good, could you add this change to both v2 and v3 as we will be supporting those for all versions going forward?

An issue not introduced in this PR, but it seems that the multiple response for NPQ applications is pointing to the wrong example see https://ecf-review-pr-2999.london.cloudapps.digital/api-reference/reference-v1.html#api-v1-npq-applications-get-responses-examples

Do you mind amending this so it points to NPQ application rather than participant? thanks

cwrw (approved):
thanks @slorek!

cwrw (approved):

app/serializers/api/v1/npq_application_csv_serializer.rb
```diff
@@ -82,6 +84,8 @@ def to_row(record)
           teacher_catchment(record),
           record.teacher_catchment_country,
           record.teacher_catchment_iso_country_code,
+          record.itt_provider,
```
cwrw, line 87: could you add a test for those in the `NPQApplicationCsvSerializer` spec for v1/v2? There seems to be a lot of duplication in the API specs, which hopefully we'll get to tidy up soon

cwrw (approved):

PR #3004, opened 2023-02-15
Rails 7

peteryates:
### Context

A bit of a spike to see how smooth the upgrade to Rails 7 might be.

### Changes proposed in this pull request

- Upgrade to Rails 7.0.4.2
- Upgrade sidekiq and rspec-rails

PR #3008, opened 2023-02-16
Remove references to ParticipantProfile::ECF

peteryates:
### Context

`ParticipantProfile` is inherited by a (almost) abstract model, `ParticipantProfile::ECF`. No `ParticipantProfile::ECF` records can exist in prod but we did use them in a few places for testing.

When trying to upgrade to Rails 7 #3004 this caused problems as Rails complains about the [STI structure](https://api.rubyonrails.org/classes/ActiveRecord/Inheritance.html). Removing the references to `ParticipantProfile::ECF` should allow us to restructure the STI and make it work with newer versions of Active Record

The longer term goal is to flatten the STI structure but to do that we first need to remove references to `ParticipantProfile::ECF`.

| What we have now | What we're aiming for (not in this PR!) |
| --------------------- | ---------------------|
| Screenshot from 2023-02-15 18-51-08 | Screenshot from 2023-02-15 18-58-21 |

### DIFF
--- a/app/models/ecf_participant_validation_data.rb
+++ b/app/models/ecf_participant_validation_data.rb
@@ -1,9 +1,19 @@
 # frozen_string_literal: true
 
 class ECFParticipantValidationData < ApplicationRecord
-  belongs_to :participant_profile, class_name: "ParticipantProfile::ECF", touch: true
+  belongs_to :participant_profile, class_name: "ParticipantProfile", touch: true
+  validate :belongs_to_ecf_participant
 
   def can_validate_participant?
     date_of_birth.present? && (trn.present? || nino.present?)
+  end
+
+private
+
+  def belongs_to_ecf_participant
+    return if participant_profile.blank?
+    return if participant_profile.type.in?(["ParticipantProfile::ECT", "ParticipantProfile::Mentor"])
+
+    errors.add(:participant_profile_id, "participant is not an ECT or Mentor")
   end
 end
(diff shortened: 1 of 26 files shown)


peteryates:
This came to a halt because #3017 got in the way, should be resumeable once that's merged.

app/models/ecf_participant_validation_data.rb
```diff
@@ -1,9 +1,19 @@
 # frozen_string_literal: true
 
 class ECFParticipantValidationData < ApplicationRecord
-  belongs_to :participant_profile, class_name: "ParticipantProfile::ECF", touch: true
+  belongs_to :participant_profile, class_name: "ParticipantProfile", touch: true
+  validate :belongs_to_ecf_participant
```
peteryates, line 5: Do we lose anything here by making the relationship generic and validating rather than prescribing types? 🤔

ltello: I can't think of anything right now but we can fix it later if something arises.

ltello (approved):
Cool!

I think NPQ profiles are so different (no induction records, no schedules, etc, etc) to ECTs or Mentors that I would even take them out of ParticipantProfile hierarchy as part of the aiming design.

peteryates:
Thanks for the review @ltello. I agree, I'd like to explore removing NPQ from the hierarchy. Let's take a closer look at approaches.

PR #3009, opened 2023-02-16
GIT Workflow action version updates

mattonfoot:
### Context

Updates versions of actions to help reduce Node12 dependency warnings


mattonfoot:
@slorek wondering if we are both doing the same thing. except I started removing redundent steps at the same time

mattonfoot:
Closing and reopening this pull request to see if the expected checks are just stuck

PR #3010, opened 2023-02-16
Select TRA participant outcomes and enqueue to send

slorek:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1956

* Implements the logic to select which outcomes to send to the qualified teachers API
* Enqueues the job (which will be implemented separately) which sends the data to the qualified teachers API
* Implements rate limiting to ensure we don't exceed the qualified teachers API rate limit
* Checks the queue is empty before enqueueing another batch to prevent a backlog forming (doesn't requeue automatically if this fails - as I understand the job will be monitored manually initially).

### Guidance to review

* Currently the job will fail silently if there are pending jobs (to ensure we do not hit the rate limit) but we could consider raising an exception - one to discuss
* I've used conservative rate limits to ensure we don't hit the ceiling. These can be tweaked if we feel it is necessary.
* We've agreed to rework the acceptance criteria to be more forgiving in order to make the database queries more efficient, and eliminate n+1 queries. Now the job is completed using only three database queries.

### DIFF
--- a/app/models/participant_outcome/npq.rb
+++ b/app/models/participant_outcome/npq.rb
@@ -17,35 +17,29 @@ class ParticipantOutcome::NPQ < ApplicationRecord
 
   after_commit :push_outcome_to_big_query
 
-  def self.latest
-    order(created_at: :desc).first
-  end
-
-  def self.to_send_to_qualified_teachers_api
-    eligible_outcomes = for_completed_declarations
-      .not_sent_to_qualified_teachers_api
-      .where(id: latest_per_declaration.map(&:id))
-
-    eligible_outcomes.passed
-      .or(
-        eligible_outcomes
-          .not_passed
-          .where(participant_declaration_id: declarations_where_previous_outcome_passed_and_sent),
-      )
-  end
-
   class << self
+    def latest
+      order(created_at: :desc).first
+    end
+
+    def to_send_to_qualified_teachers_api
+      eligible_outcomes = not_sent_to_qualified_teachers_api
+        .where(id: latest_per_declaration.map(&:id))
+
+      eligible_outcomes.passed
+        .or(
+          eligible_outcomes
+            .not_passed
+            .where(participant_declaration_id: declarations_where_outcome_passed_and_sent),
+        )
+    end
+
     def latest_per_declaration
       select("DISTINCT ON(participant_declaration_id) *")
         .order(:participant_declaration_id, created_at: :desc)
     end
 
-    def for_completed_declarations
-      joins(:participant_declaration)
-        .where("participant_declarations.declaration_type='completed'")
-    end
-
-    def declarations_where_previous_outcome_passed_and_sent
+    def declarations_where_outcome_passed_and_sent
       latest_per_declaration
         .passed
         .sent_to_qualified_teachers_api
--- a/spec/models/participant_outcome/npq_spec.rb
+++ b/spec/models/participant_outcome/npq_spec.rb
@@ -81,60 +81,118 @@ RSpec.describe ParticipantOutcome::NPQ, :with_default_schedules, type: :model do
   end
 
   describe ".to_send_to_qualified_teachers_api" do
-    subject(:result) { described_class.to_send_to_qualified_teachers_api }
+    subject(:result) { described_class.to_send_to_qualified_teachers_api.map(&:id) }
 
-    context "when a declaration is not closed" do
-      let!(:declaration) { create(:npq_participant_declaration, declaration_type: "started") }
-      let!(:outcome_1) { create(:participant_outcome, :passed, :not_sent_to_qualified_teachers_api, participant_declaration: declaration) }
+    let!(:declaration) { create(:npq_participant_declaration, declaration_type: "completed") }
+
+    context "when the latest outcome for a declaration has been sent to the qualified teachers API" do
+      let!(:outcome_1) { create(:participant_outcome, :passed, :sent_to_qualified_teachers_api, participant_declaration: declaration) }
+      let!(:outcome_2) { create(:participant_outcome, :failed, :sent_to_qualified_teachers_api, participant_declaration: declaration) }
 
       it { is_expected.to be_empty }
     end
 
-    context "when a declaration is closed" do
-      let!(:declaration) { create(:npq_participant_declaration, declaration_type: "completed") }
+    context "when the latest outcome for a declaration has not been sent to the qualified teachers API but a previous outcome has been sent" do
+      let!(:outcome_1) { create(:participant_outcome, :passed, :sent_to_qualified_teachers_api, participant_declaration: declaration) }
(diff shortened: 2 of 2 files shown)


app/jobs/participant_outcomes/batch_send_latest_outcomes_job.rb
```diff
@@ -0,0 +1,40 @@
...
+        .includes(:outcomes)
+        .with_outcomes_not_sent_to_qualified_teachers_api
+        .map { |declaration| declaration.outcomes.to_send_to_qualified_teachers_api }
+        .flatten
+    end
+
+    def can_enqueue_jobs?
+      ActiveJob::Base.queue_adapter.enqueued_jobs
+        .select { |job| job[:job] == SendToQualifiedTeachersApiJob }
+        .empty?
```
mooktakim, line 33: This would be better with `detect`, there's no need to go through the whole list if one is found.

app/jobs/participant_outcomes/batch_send_latest_outcomes_job.rb
```diff
@@ -0,0 +1,40 @@
...
+      @batch_size = batch_size
+
+      outcomes.first(@batch_size).each { |outcome| SendToQualifiedTeachersApiJob.perform_later(outcome.id) }
+
+      self.class.set(wait: REQUEUE_DELAY).perform_later if should_enqueue_again?
+    end
+
+  private
+
+    def outcomes
```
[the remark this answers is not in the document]
cwrw: hmmm let me have a think about how we can improve efficiency and get back to you on that

cwrw (commented):
looking great @slorek, a few minor comments and going to take away the query to think about and will get back to you shortly

app/jobs/participant_outcomes/batch_send_latest_outcomes_job.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    def can_enqueue_jobs?
+      ActiveJob::Base.queue_adapter.enqueued_jobs
+        .detect { |job| job[:job] == SendToQualifiedTeachersApiJob }
+        .nil?
+    end
+
+    def should_enqueue_again?
+      outcomes.count > @batch_size
```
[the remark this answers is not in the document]
cwrw: sorry maybe I misunderstood, but the check above is comparing it to the batch size (which is currently set to 200), so if the outcome size is 10 its not more than the batch size (200) so won't enqueue again, unless I missed something?

app/jobs/participant_outcomes/batch_send_latest_outcomes_job.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module ParticipantOutcomes
+  class BatchSendLatestOutcomesJob < ApplicationJob
+    queue_as :participant_outcomes
+
+    DEFAULT_BATCH_SIZE = 200
```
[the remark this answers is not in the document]
cwrw: Our estimate for how many will be required to send by March is around 14k+, so it will take substantially longer. However Im happy to wait for when we test on pre-prod to ensure how much we can bump up though, hopefully to at least change the delay to 1 minute

app/jobs/participant_outcomes/batch_send_latest_outcomes_job.rb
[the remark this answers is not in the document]
cwrw: ah oops I missed that, gotcha thanks 👍

cwrw (commented):
chatted in person around a few things we can tweak, but leaving previous comments for posterity

app/models/participant_outcome/npq.rb
```diff
@@ -21,6 +21,54 @@ def self.latest
...
+
+  class << self
+    def latest_per_declaration
+      select("DISTINCT ON(participant_declaration_id) *")
+        .order(:participant_declaration_id, created_at: :desc)
+    end
+
+    def for_completed_declarations
+      joins(:participant_declaration)
+        .where("participant_declarations.declaration_type='completed'")
```
cwrw, line 45: suggestion: maybe we can use the `completed` scope on declarations? Although I think we might not need this anymore since we are performing the queries from the outcomes side now

app/models/participant_outcome/npq.rb
```diff
@@ -21,6 +21,54 @@ def self.latest
     order(created_at: :desc).first
   end
 
+  def self.to_send_to_qualified_teachers_api
```
cwrw, line 24: could we move all class methods into `class << self` below?

spec/models/participant_outcome/npq_spec.rb
```diff
@@ -80,6 +80,60 @@
...
+    subject(:result) { described_class.to_send_to_qualified_teachers_api }
+
+    context "when a declaration is not closed" do
+      let!(:declaration) { create(:npq_participant_declaration, declaration_type: "started") }
+      let!(:outcome_1) { create(:participant_outcome, :passed, :not_sent_to_qualified_teachers_api, participant_declaration: declaration) }
+
+      it { is_expected.to be_empty }
+    end
+
+    context "when a declaration is closed" do
```
cwrw, line 93:
```suggestion
    context "when a declaration is completed" do
```

spec/models/participant_outcome/npq_spec.rb
```diff
@@ -80,6 +80,60 @@
     end
   end
 
+  describe ".to_send_to_qualified_teachers_api" do
+    subject(:result) { described_class.to_send_to_qualified_teachers_api }
+
+    context "when a declaration is not closed" do
```
cwrw, line 86: worth noting outcomes only ever exist on completed NPQ declarations. NPQ completed declarations will never exist without an outcome. I propose we amend/remove this spec to avoid confusion

db/migrate/20230221191651_add_missing_indexes.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class AddMissingIndexes < ActiveRecord::Migration[6.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :participant_declarations, :declaration_type, algorithm: :concurrently
```
cwrw, line 7: 💯

cwrw (approved):
perfectly done! Thanks for addressing all comments. I've left some very minor suggestions but otherwise 👍

spec/models/participant_outcome/npq_spec.rb
cwrw: discuss in person: yea no longer needed

app/models/participant_outcome/npq.rb
cwrw: discussed in person: no longer needed 👍

cwrw (approved):

peteryates (approved):
Looks great!

PR #3012, opened 2023-02-17
Bump node version to resolve Docker build errors

slorek:
Resolves Docker build errors, example from https://github.com/DFE-Digital/early-careers-framework/actions/runs/4203388895/jobs/7292751839

`#11 83.15 error early-careers-framework@: The engine "node" is incompatible with this module. Expected version "16.17.x". Got "16.19.1"`

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 yarn 1.22.0
 ruby 3.1.2
-nodejs 16.19.0
+nodejs 16.19.1


.tool-versions
```diff
@@ -1,3 +1,3 @@
 yarn 1.22.0
 ruby 3.1.2
-nodejs 16.17.0
+nodejs 16.19.0
```
cwrw, line 3: `16.19.1`?

cwrw: unless it pulls the latest then ignore me

cwrw (approved):

PR #3016, opened 2023-02-17
Replace shakapacker with jssbundling-rails and cssbundling-rails

slorek:
This branch replaces the [Shakapacker](https://github.com/shakacode/shakapacker) gem with [jsbundling-rails](https://github.com/rails/jsbundling-rails) and [cssbundling-rails](https://github.com/rails/cssbundling-rails), which is the new conventional aproach in Rails 7.

Sprockets is used for images and fonts only. When [upgrading to Rails 7,](https://github.com/DFE-Digital/early-careers-framework/pull/3004) we will then be able to replace `sprockets` with [propshaft](https://github.com/rails/propshaft).

Currently JS is still compiled using webpack, but jsbundling-rails also supports [esbuild](https://esbuild.github.io/), so this can be swapped out later if desired.

The compile JS: `yarn build`. To compile CSS: `yarn build:css`.

Developers now have a new binstub which combines all processes needed for local development: `bin/dev`. This uses Foreman to run the processes defined in `Procfile.dev`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -135,7 +135,7 @@ GEM
       railties (>= 6.0.0)
     cypress-on-rails (1.13.1)
       rack
-    database_cleaner-active_record (2.0.1)
+    database_cleaner-active_record (2.1.0)
       activerecord (>= 5.a)
       database_cleaner-core (~> 2.0.0)
     database_cleaner-core (2.0.1)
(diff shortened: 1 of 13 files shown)


peteryates (approved):
Nice, works perfectly. Glad to see the back of `bin/webpacker-dev-server` too 🎉

cwrw (approved):
nicely done!

PR #3034, opened 2023-02-22
CST-1297: add validation data to participant profiles seeded

ltello:
### Context

- [Ticket](https://dfedigital.atlassian.net/browse/CST-1297):  add validation data to ECTs/Mentors seeded so they can be displayed properly in the participants dashboard.

### DIFF
--- a/db/new_seeds/base/add_mentor_scenarios.rb
+++ b/db/new_seeds/base/add_mentor_scenarios.rb
@@ -84,7 +84,7 @@ ucl = CoreInductionProgramme.find_by!(name: "UCL Institute of Education")
     .build(
       schedule: mentor_params.schedule,
     )
-    .chain_add_validation_data
-    .chain_add_eligibility
+    .with_validation_data
+    .with_eligibility
     .add_induction_record(induction_programme: mentor_params.induction_programme)
 end
--- a/db/new_seeds/base/add_mentor_scenarios.rb
+++ b/db/new_seeds/base/add_mentor_scenarios.rb
@@ -80,10 +80,11 @@ ucl = CoreInductionProgramme.find_by!(name: "UCL Institute of Education")
   ),
 ].each do |mentor_params|
   NewSeeds::Scenarios::Participants::Mentors::MentorWithNoEcts
-    .new(full_name: mentor_params.full_name)
+    .new(school_cohort: mentor_params.school_cohort, full_name: mentor_params.full_name)
     .build(
-      school_cohort: mentor_params.school_cohort,
       schedule: mentor_params.schedule,
     )
+    .chain_add_validation_data
+    .chain_add_eligibility
     .add_induction_record(induction_programme: mentor_params.induction_programme)
 end
--- a/db/new_seeds/base/add_schools_and_local_authorities.rb
+++ b/db/new_seeds/base/add_schools_and_local_authorities.rb
@@ -16,18 +16,14 @@ def add_school_to_local_authority(school:, local_authority:, lead_providers:, co
 
       scenarios = Random.rand(1..4).times.map do
         NewSeeds::Scenarios::Participants::Mentors::MentoringMultipleEctsWithSameProvider
-          .new(
-            school:,
-            lead_provider:,
-            number: Random.rand(1..4), # number of mentees
-          )
+          .new(school:, lead_provider:)
           .build
+          .add_mentees(Random.rand(1..4), with_eligibility: false, with_validation_data: true)
       end
 
       scenarios.flat_map(&:mentees).map(&:participant_profile).each do |participant_profile|
         Rails.logger.debug("seeding eligibility for #{participant_profile.user.full_name}")
 
-        participant_profile.ecf_participant_eligibility&.destroy!
         FactoryBot.create(:seed_ecf_participant_eligibility, random_weighted_eligibility_trait, participant_profile:)
       end
     end
--- a/db/new_seeds/base/add_schools_and_local_authorities.rb
+++ b/db/new_seeds/base/add_schools_and_local_authorities.rb
@@ -27,7 +27,8 @@ def add_school_to_local_authority(school:, local_authority:, lead_providers:, co
       scenarios.flat_map(&:mentees).map(&:participant_profile).each do |participant_profile|
         Rails.logger.debug("seeding eligibility for #{participant_profile.user.full_name}")
 
-        FactoryBot.create(:seed_ecf_participant_eligibilty, random_weighted_eligibility_trait, participant_profile:)
+        participant_profile.ecf_participant_eligibility&.destroy!
+        FactoryBot.create(:seed_ecf_participant_eligibility, random_weighted_eligibility_trait, participant_profile:)
       end
     end
 
--- a/db/new_seeds/base/add_schools_and_local_authorities.rb
+++ b/db/new_seeds/base/add_schools_and_local_authorities.rb
@@ -12,16 +12,13 @@ def add_school_to_local_authority(school:, local_authority:, lead_providers:, co
     end
 
     lead_providers.sample.tap do |lead_provider|
-      FactoryBot.create(:seed_partnership, :with_delivery_partner, school:, lead_provider:, cohort: cohorts.sample)
-
       scenarios = Random.rand(1..4).times.map do
         NewSeeds::Scenarios::Participants::Mentors::MentoringMultipleEctsWithSameProvider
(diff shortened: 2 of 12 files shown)


db/new_seeds/base/add_schools_and_local_authorities.rb
```diff
@@ -27,7 +27,8 @@ def add_school_to_local_authority(school:, local_authority:, lead_providers:, co
       scenarios.flat_map(&:mentees).map(&:participant_profile).each do |participant_profile|
         Rails.logger.debug("seeding eligibility for #{participant_profile.user.full_name}")
 
-        FactoryBot.create(:seed_ecf_participant_eligibilty, random_weighted_eligibility_trait, participant_profile:)
+        participant_profile.ecf_participant_eligibility&.destroy!
+        FactoryBot.create(:seed_ecf_participant_eligibility, random_weighted_eligibility_trait, participant_profile:)
```
peteryates, line 31: How come we're destroying the eligibility first? I thought unless it had been explicitly created it shouldn't be there? 🤔

db/new_seeds/base/add_schools_and_local_authorities.rb
```diff
@@ -27,7 +27,8 @@ def add_school_to_local_authority(school:, local_authority:, lead_providers:, co
       scenarios.flat_map(&:mentees).map(&:participant_profile).each do |participant_profile|
         Rails.logger.debug("seeding eligibility for #{participant_profile.user.full_name}")
 
-        FactoryBot.create(:seed_ecf_participant_eligibilty, random_weighted_eligibility_trait, participant_profile:)
```
peteryates, line 30: Ah thanks for catching this typo! 🙈

db/new_seeds/base/add_schools_and_local_authorities.rb
ltello: Yes, not elegant, I know. The thing is that now those mentees have eligibility when they are created because the scenario adds it. Otherwise ECTs and mentors not get properly displayed and categorised in the dashboard.

I thought this dirty approach of deleting it and then recreate it with a random status would be better than passing parameters to the scenario to say "Do not create eligibility for these ECTs". That would also make the scenario code more complicated. 

Happy to rethink this if we are not happy even though this is just seeding code.

db/new_seeds/base/add_schools_and_local_authorities.rb
```diff
@@ -16,18 +16,15 @@ def add_school_to_local_authority(school:, local_authority:, lead_providers:, co
 
       scenarios = Random.rand(1..4).times.map do
         NewSeeds::Scenarios::Participants::Mentors::MentoringMultipleEctsWithSameProvider
-          .new(
-            school:,
-            lead_provider:,
-            number: Random.rand(1..4), # number of mentees
-          )
+          .new(school:, lead_provider:)
```
tonyheadford, line 19: Feels like this was better before as if you don't call `add_mentees` the scenario "name" doesn't make sense?

db/new_seeds/base/add_mentor_scenarios.rb
```diff
@@ -80,10 +80,11 @@
   ),
 ].each do |mentor_params|
   NewSeeds::Scenarios::Participants::Mentors::MentorWithNoEcts
-    .new(full_name: mentor_params.full_name)
+    .new(school_cohort: mentor_params.school_cohort, full_name: mentor_params.full_name)
     .build(
-      school_cohort: mentor_params.school_cohort,
       schedule: mentor_params.schedule,
     )
+    .chain_add_validation_data
```
tonyheadford, line 87: How do you feel about renaming these to `.with_validation_data` and `.with_eligibility`?  Think it would flow nicer when reading the code

db/new_seeds/scenarios/participants/ects/ect.rb
```diff
@@ -5,7 +5,14 @@ module Scenarios
...
         class Ect
-          attr_reader :user, :new_user_attributes, :participant_profile, :participant_identity, :teacher_profile, :school_cohort
+          attr_reader :new_user_attributes,
+                      :participant_identity,
+                      :participant_profile,
+                      :school_cohort,
+                      :teacher_profile,
+                      :user
+
+          delegate :ecf_participant_eligibility, :ecf_participant_validation_data, to: :participant_profile
```
tonyheadford, line 15: Wondering if this is necessary?  I'm not sure whether it's a good idea to treat these scenarios as the actual entities or whether we should consider renaming them with a "Builder" suffix e.g. `EctBuilder` or something like that to make it clear. Would be good to get everyones thoughts on that.

db/new_seeds/scenarios/participants/transfers/fip_to_fip.rb
```diff
@@ -38,6 +35,13 @@ def initialize(from_school: nil, to_school: nil)
             @school_to = to_school
           end
 
+          delegate :ecf_participant_eligibility,
+                   :ecf_participant_validation_data,
+                   :participant_identity,
+                   :teacher_profile,
+                   :user,
+                   to: :participant_profile
+
```
tonyheadford, line 44: Same comment as above

db/new_seeds/base/add_mentor_scenarios.rb
ltello: I think it is a good idea

db/new_seeds/scenarios/participants/ects/ect.rb
ltello: Agreed. I have removed  all these method and expose only the participant profile created.

db/new_seeds/base/add_schools_and_local_authorities.rb
ltello: Yes, I have made a refactor and simplified a few methods.

db/new_seeds/scenarios/participants/transfers/fip_to_fip.rb
ltello: Removed

db/schema.rb
```diff
@@ -645,7 +645,7 @@
     t.boolean "sparsity_uplift"
     t.boolean "pupil_premium_uplift"
     t.uuid "delivery_partner_id"
-    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY (ARRAY[('submitted'::character varying)::text, ('eligible'::character varying)::text, ('payable'::character varying)::text, ('paid'::character varying)::text, ('clawed_back'::character varying)::text]))"
+    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY ((ARRAY['submitted'::character varying, 'eligible'::character varying, 'payable'::character varying, 'paid'::character varying, 'clawed_back'::character varying])::text[]))"
```
cwrw, line 648: could we avoid committing these changes please?

ltello: @cwrw why am I getting these changes after running rails db:migrate if I have the standard setup?

ltello: Anyway I have reverted the change. Not sure if my version of postgres might be the cause.

cwrw: I think its our different PG versions (I get this too) as you mentioned, but others on the team don't

tonyheadford: Yes I get this too.

tonyheadford (approved):
Thanks for making those changes, looks good to me

db/new_seeds/scenarios/participants/ects/ect.rb
```diff
@@ -14,25 +14,25 @@ def initialize(school_cohort:, full_name: nil, email: nil)
...
           end
 
-          def chain_add_induction_record(**induction_args)
+          def with_induction_record(**induction_args)
             add_induction_record(**induction_args)
             self
           end
 
-          def add_induction_record(induction_programme:, start_date:, end_date:, induction_status: "active", training_status: "active")
+          def add_induction_record(induction_programme:, mentor_profile: nil, start_date: 6.months.ago, end_date: nil, induction_status: "active", training_status: "active")
```
peteryates, line 31: I think the convetion of `add_` for non-chainable and `with_` for chainable makes sense, let's continue that.

peteryates (approved):
Love it, this looks super clean.

PR #3040, opened 2023-02-23
Remove extra parens from schema.rb

peteryates:
### Context

We had a odd situation where running `rails db:migrate` on `main` caused the `db/schema.rb` to be updated. The updates were innocuous and just added extra parenthesis but then mean other devs need to be concious not to re-commit their version, causing the file to flip flop unnecessarily.

I've set myself as codeowner just to keep an eye on it.

- Re-run db:migrate to reset schema.rb
- Set myself as codeowner of schema.rb

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-
+db/schema.rb @peteryates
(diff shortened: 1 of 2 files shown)


Lockyy:
Nice to have this fixed, I had an issue a while ago where the flip flopping caused what seemed to be genuine changes in how the schema was loaded which lead to these odd failures: https://github.com/DFE-Digital/early-careers-framework/actions/runs/3988006839/jobs/6853824483

PR #3041, opened 2023-02-23
[CPDNPQ-883] Add api/v1/npq/previous_funding#show

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-883

This API is built as a replacement for api/v1/npq-funding/:trn.

The primary goal of building this replacement is to introduce an API that doesn’t depend on the existence of a TRN for it to function.

The existing API embedded the TRN being queried against into the URL itself. The new API doesn’t include any identifiers in the URL, instead receiving the user identifier data in the same way as course identifier information via query parameters.

This allows us to break out from a model where you have to have a TRN to where you can use multiple different kind of identifiers.

Old API formatting example:
`/api/v1/npq-funding/1234567?npq_course_identifier=npq-senior-leadership`

New API formatting example:
`/api/v1/npq/previous-funding?npq_course_identifier=npq-senior-leadership&trn=1234567`

The response body for the API remains identical so that this should function as a drop in replacement. Though people swapping should be aware that errors are now more descriptive:

#### Success response

Status: **200**
Body: 
```
{
  "previously_funded" => false,
  "previously_received_targeted_funding_support" => false,
}
```

#### No npq_course_identifier error

Status: **400**
Body: 
```
{
  "error" => "No npq_course_identifier provided",
}
```

#### No identifier error

Status: **400**
Body: 
```
{
  "error" => "No identifier provided. Valid identifier params: trn or get_an_identity_id",
}
```

### DIFF
--- a/app/controllers/api/v1/npq/previous_fundings_controller.rb
+++ b/app/controllers/api/v1/npq/previous_fundings_controller.rb
@@ -4,7 +4,7 @@ module Api
   module V1
     module NPQ
       class PreviousFundingsController < ApiController
-        before_action :ensure_identifier_present, only: :show
+        before_action :ensure_user_identifier_present, only: :show
         before_action :ensure_npq_course_present, only: :show
 
         def show
@@ -25,10 +25,10 @@ module Api
           params[:npq_course_identifier]
         end
 
-        def ensure_identifier_present
+        def ensure_user_identifier_present
           return if trn.present? || get_an_identity_id.present?
 
-          error_body = { error: "No identifier provided. Valid identifier params: trn or get_an_identity_id" }
+          error_body = { error: "No user identifier provided. Valid identifier params: trn or get_an_identity_id" }
           render json: error_body, status: :bad_request
         end
 
--- a/app/controllers/api/v1/npq/previous_fundings_controller.rb
+++ b/app/controllers/api/v1/npq/previous_fundings_controller.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module NPQ
+      class PreviousFundingsController < ApiController
+        before_action :ensure_identifier_present, only: :show
+        before_action :ensure_npq_course_present, only: :show
+
+        def show
+          render json: service.call
+        end
+
+      private
+
+        def trn
+          params[:trn]
+        end
+
+        def get_an_identity_id
+          params[:get_an_identity_id]
+        end
+
+        def npq_course
+          params[:npq_course_identifier]
+        end
+
+        def ensure_identifier_present
+          return if trn.present? || get_an_identity_id.present?
+
+          error_body = { error: "No identifier provided. Valid identifier params: trn or get_an_identity_id" }
+          render json: error_body, status: :bad_request
+        end
+
+        def ensure_npq_course_present
+          return if npq_course.present?
+
+          error_body = { error: "No npq_course_identifier provided" }
+          render json: error_body, status: :bad_request
+        end
+
+        def service
+          @service ||= ::NPQ::FundingEligibility.new(
+            trn:,
+            get_an_identity_id:,
+            npq_course_identifier: npq_course,
+          )
+        end
+      end
+    end
+  end
+end
--- a/app/services/npq/funding_eligibility.rb
+++ b/app/services/npq/funding_eligibility.rb
@@ -2,10 +2,13 @@
 
 module NPQ
   class FundingEligibility
-    attr_reader :trn, :npq_course_identifier
+    attr_reader :npq_course_identifier,
+                :trn,
+                :get_an_identity_id
 
-    def initialize(trn:, npq_course_identifier:)
+    def initialize(npq_course_identifier:, trn: nil, get_an_identity_id: nil)
       @trn = trn
+      @get_an_identity_id = get_an_identity_id
       @npq_course_identifier = npq_course_identifier
     end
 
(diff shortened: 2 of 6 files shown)


app/services/npq/funding_eligibility.rb
```diff
@@ -2,10 +2,13 @@
 
 module NPQ
   class FundingEligibility
-    attr_reader :trn, :npq_course_identifier
+    attr_reader :npq_course_identifier,
+                :trn,
+                :get_an_identity_id
 
-    def initialize(trn:, npq_course_identifier:)
+    def initialize(npq_course_identifier:, trn: nil, get_an_identity_id: nil)
```
Lockyy, line 9: This addition of a default value for trn does introduce a change where using one isn't strictly necessary. At first glance this may seem like it would impact the existing API to make it no longer require a TRN. Howver, the TRN in that API is in the URL, if you don't send one you'll get a routing error.

spec/requests/api/v1/npq/previous_funding/show_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+  let(:base_url) { "/api/v1/npq/previous_funding" }
+
+  let(:npq_course_identifier) { "npq-leading-literacy" }
+  let(:trn) { "1234567" }
+  let(:get_an_identity_id) { "87654321" }
+
+  def stub_funding_eligibility_response(trn: nil, get_an_identity_id: nil, npq_course_identifier: nil)
+    stubbed_response = {
+      "previously_funded" => SecureRandom.uuid,
+      "previously_received_targeted_funding_support" => SecureRandom.uuid,
```
shjohnson, line 19: Minor but any reason this is not: `previously_received_targeted_support_funding` in line with other definitions across our system?

spec/requests/api/v1/npq/previous_funding/show_spec.rb
```diff
@@ -0,0 +1,113 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "NPQ Funding API", type: :request do
```
shjohnson, line 5: Do we want to add a test when all params are sent?

spec/services/npq/funding_eligibility_spec.rb
```diff
@@ -3,167 +3,417 @@
...
 
-      before do
-        NPQ::Application::Accept.new(npq_application: application).call
+        context "with a get_an_identity_id" do
+          let(:get_an_identity_id) { SecureRandom.uuid }
+          let(:user) { create(:user, get_an_identity_id:) }
+
+          context "that is not previously funded" do
+            before do
+              create(
```
shjohnson, line 53: When creating the application here is it linked to the previous one when your checking? basically i wondered if this is having no affect on the call to if its previously funded?

spec/services/npq/funding_eligibility_spec.rb
```diff
@@ -3,167 +3,417 @@
...
+        let(:application) do
+          create(
+            :npq_application,
+            eligible_for_funding: true,
+            teacher_reference_number_verified: true,
+          )
+        end
+        let(:npq_application_course) { application.npq_course }
+        # Making sure they are completely separate courses
+        let(:npq_course) { create(:npq_course, identifier: npq_application_course.identifier.reverse) }
```
shjohnson, line 128: Id only be worried here about doing a random string rather than a course that exists if we have some weird checking against certain course in future and this test doesnt catch it if makes sense? might be worth just selecting two tests we know exist and comparing them against each other?

spec/requests/api/v1/npq/previous_funding/show_spec.rb
Lockyy: Good call

spec/requests/api/v1/npq/previous_funding/show_spec.rb
Lockyy: No particular reason for this API to use it in particular other than to replicate the existing API's response body. Why it was phrased like this when I added it to that originally is likely down to a mistake.

app/controllers/api/v1/npq/previous_fundings_controller.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module NPQ
+      class PreviousFundingsController < ApiController
+        before_action :ensure_identifier_present, only: :show
```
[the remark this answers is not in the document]
Lockyy: Definitely a good point

spec/services/npq/funding_eligibility_spec.rb
Lockyy: Good point, I've updated the spec to make the funded/unfunded specs link much more closely together so that the only thing that differs between the two is the funding status of the application. That way, it has to be that that's making the decision.

spec/services/npq/funding_eligibility_spec.rb
Lockyy: Updated

spec/requests/api/v1/npq/previous_funding/show_spec.rb
```diff
@@ -61,6 +61,20 @@ def stub_funding_eligibility_response(trn: nil, get_an_identity_id: nil, npq_cou
           expect(parsed_response).to eq(stubbed_response)
         end
 
+        context "with get_an_identity_id" do
```
shjohnson, line 64: `..get_an_identity_id and trn`

shjohnson (approved):
LGTM

spec/requests/api/v1/npq/previous_funding/show_spec.rb
Lockyy: This is nested inside "with trn". So the contexts build up the full test name into:
`GET /api/v1/npq/previous_funding when authorized with trn with get_an_identity_id returns correct response`

PR #3043, opened 2023-02-24
Refactor Admin::Schools::InductionCoordinatorsController

lazydevorg:
### Context

- Ticket: CST-1293

* It's using separate hardcoded pages (via render_email_used_page and render_name_different_page) for no reason. It should re-render the form with an [error summary](https://govuk-form-builder.netlify.app/introduction/error-handling).
* The #create action us using CreateInductionTutor but there is no corresponding UpdateInductionTutor.


app/services/update_induction_tutor.rb
```diff
@@ -0,0 +1,18 @@
...
+    @school = school
+    @email = email
+    @full_name = full_name
+  end
+
+  def call
+    ActiveRecord::Base.transaction do
+      @school.induction_tutor.update!(full_name: @full_name,
+                                      email: @email)
+    end
```
tonyheadford, line 16: Doesn't need to be in a transaction for a single update, however i think this needs to check that the person that is being updated does not have participant profiles (ECF or NPQ) and validation data etc and we're changing the name on their user record to a completely different person.

app/forms/nominate_induction_tutor_form.rb
```diff
@@ -44,13 +37,38 @@ def existing_name
     existing_user.full_name
   end
 
+  def name_different?
+    existing_name.present? && existing_name != full_name
+  end
+
+  def email_already_taken?
+    ParticipantProfile.active_record.ects.joins(:user).where(user: { email: }).any?
```
tonyheadford, line 45: Not sure if we should go further here because a SIT won't have a participant profile unless they are also a Mentor?

tonyheadford (changes requested):
Think we could improve the checks we're performing in here, as it looks like it could could issues (I think that's the existing journey not your changes) but we should use this as an opportunity to fix I think

app/forms/nominate_induction_tutor_form.rb
```diff
@@ -44,13 +37,38 @@ def existing_name
...
+  end
```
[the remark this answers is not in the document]
tonyheadford: Think that should be `Identity.find_user_by(email:)` we virtually never want to use `User.find_by` when trying to find people.

app/forms/nominate_induction_tutor_form.rb
lazydevorg: @tonyheadford if I change this line with `Identity.find_user_by(email:).present?` then most of the tests in `spec/requests/admin/schools/induction_coordinators_spec.rb` fail.
I'm not sure it's safe to change. Any idea?

tonyheadford: Whenever we check for the existence of an email we have to check the participant identities as well as the User.email, all of which `Identity.find_user_by(email:)` does correctly.  I'm not sure which specs are failing but I would suspect that there are errors somewhere in those spec's setup.

tonyheadford (approved):

PR #3044, opened 2023-02-24
Rake task to enqueue participant outcomes to send to the qualified teachers API

slorek:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1959

- Adds a Rake task to enqueue `ParticipantOutcomes::BatchSendLatestOutcomesJob`
- Custom batch size can optionally be specified by CLI argument, e.g. `bin/rake "participant_outcomes:send_to_qualified_teachers_api[500]"`

### Guidance to review

Regrettably I had to monkey patch the Sidekiq queue adapter to provide a common interface with the test adapter to inspect the queue. This avoids having to conditionally call the `Sidekiq` interface directly depending on which environment the job is running. No methods are overwritten - by default it seems to try to delegate this method to the `TestAdapter` class even when this is not specified as the adapter, but it results in a stack level too deep error. Alternatives to monkey patching could include inline conditional logic, or removing the guard from `ParticipantOutcomes::BatchSendLatestOutcomesJob` completely.


lib/active_job/queue_adapters/sidekiq_adapter_patch.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module ActiveJob
+  module QueueAdapters
+    # Adds enqueued_jobs method consistent with the TestAdapter
+    #
+    class SidekiqAdapter
+      def enqueued_jobs
```
cwrw, line 8: Ah I wasn't aware this method wasn't available. Instead of monkey patching this, can we use available `Sidekiq` methods, I think this might work:
```
Sidekiq::Queue.new("participant_outcomes").detect { |job| job.display_class == "ParticipantOutcomes::SendToQualifiedTeachersApiJob" }
```

lib/tasks/participant_outcomes.rake
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+namespace :participant_outcomes do
+  desc "Send participant outcomes to the qualified teachers API"
+  task :send_to_qualified_teachers_api, %i[batch_size] => :environment do |_task, args|
+    job_klass = ParticipantOutcomes::BatchSendLatestOutcomesJob
+    batch_size = args[:batch_size]&.to_i || job_klass::DEFAULT_BATCH_SIZE
```
cwrw, line 7: could we also allow the `REQUEUE_DELAY` to be passed in as an argument as well, to give us more flexibility when running?

app/jobs/participant_outcomes/batch_send_latest_outcomes_job.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require Rails.root.join "lib/active_job/queue_adapters/sidekiq_adapter_patch"
+
 module ParticipantOutcomes
   class BatchSendLatestOutcomesJob < ApplicationJob
     queue_as :participant_outcomes
```
cwrw, line 7: not introduced in this PR but I realised that the interface for `SendToQualifiedTeachersApiJob` is expecting a keyword parameter, could we change it to:
```
SendToQualifiedTeachersApiJob.perform_later(participant_outcome_id: outcome.id)
```

lib/active_job/queue_adapters/sidekiq_adapter_patch.rb
cwrw: That's fair, however I think I'd like to avoid mapping all jobs in Sidekiq (as there might be quiet a few in other queues) to only find the jobs in the `participant_outcomes` queue, and even though it's not ideal in the tests to stub this at least we ensure we are very explicit on exactly what we are checking

app/jobs/participant_outcomes/batch_send_latest_outcomes_job.rb
```diff
@@ -1,25 +1,27 @@
...
         return
       end
 
       @batch_size = batch_size
 
-      outcomes.first(@batch_size).each { |outcome| SendToQualifiedTeachersApiJob.perform_later(outcome.id) }
+      outcomes.first(@batch_size).each { |outcome| SendToQualifiedTeachersApiJob.perform_later(participant_outcome_id: outcome.id) }
 
-      self.class.set(wait: REQUEUE_DELAY).perform_later if should_enqueue_again?
+      self.class.set(wait: delay).perform_later if should_enqueue_again?
```
cwrw, line 24: could we pass in the delay/batch size to when we're rescheduling this job, as otherwise it will fallback to the old settings?

cwrw (approved):
thanks @slorek 👍

PR #3047, opened 2023-02-24
[CPDLP-1973] Update 'Search participant data' page to show outcomes

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1973

* Added component to list participant outcomes on search participant data page

### Guidance to review

* Login as `finance@example.com`
* Go to 'Search participant data' page
* Enter an NPQ user, see outcomes table.


app/components/finance/npq/participant_outcomes/table.html.erb
```diff
@@ -0,0 +1,23 @@
+<h4>Declaration Outcomes</h4>
+
+<%=
+  govuk_table do |table|
+    table.with_head do |head|
```
peteryates, line 5: This could potentially be more accessible with a caption rather than heading:

```suggestion
<%=
  govuk_table do |table|
    table.with_caption(text: "Declaration outcomes")
    table.with_head do |head|
```

peteryates (approved):
Looks great overall. Left a minor suggestion on using table captions.

Also, and not one for this PR, but the heading hierarchy is difficult to follow here as the table headers are the same size as the section headings:

image

Maybe we should use something [card-like to add some separation](https://design-system.service.gov.uk/components/summary-list/#adding-card-actions)?

app/components/finance/npq/participant_outcomes/table.rb
```diff
@@ -0,0 +1,37 @@
...
+        def completion_date(outcome)
+          if outcome.voided?
+            outcome.updated_at.to_s(:govuk)
+          else
+            outcome.completion_date.to_s(:govuk)
+          end
+        end
+
+        def sent_to_tra(outcome)
+          if outcome.failed?
```
cwrw, line 28: I wonder if we should only use `sent_to_qualified_teachers_api_at` as an indication, as failed outcomes can be sent if they are correcting a previous error 🤔

app/components/finance/npq/participant_outcomes/table.rb
```diff
@@ -0,0 +1,37 @@
...
+          @participant_declaration = participant_declaration
+        end
+
+        def render?
+          participant_declaration.participant_profile.npq?
+        end
+
+        def completion_date(outcome)
+          if outcome.voided?
+            outcome.updated_at.to_s(:govuk)
```
cwrw, line 21: voided outcomes still have a `completion_date`, maybe we can surface it as well?

app/components/finance/npq/participant_outcomes/table.rb
mooktakim: @cwrw 
I've based on this around the acceptance criteria on the ticket. Other than that, I agree with you.

app/components/finance/npq/participant_outcomes/table.rb
mooktakim: @cwrw 
Again this one is also specified in the ticket.

Maybe it makes sense to show the "created_at" as a new column, that way its clear when the outcome was "changed".

app/components/finance/npq/participant_outcomes/table.rb
cwrw: hmm I mean as long as they are aware that this is the case and want it this way then that's fine

app/components/finance/npq/participant_outcomes/table.rb
cwrw: same here, a bit confusing but if they understand this and prefer this set up then we can go with it

app/views/finance/participants/_declaration.html.erb
```diff
@@ -45,31 +45,28 @@
...
+          row.with_cell(text: declaration_state.state)
+          row.with_cell(text: declaration_state.created_at.to_s(:govuk))
+          row.with_cell(text: declaration_state.state_reason || 'N/A', numeric: true)
+        end
+      end
+    end
+  end
+%>
+
+<%= render Finance::NPQ::ParticipantOutcomes::Table.new(participant_declaration: declaration) %>
```
cwrw, line 72: could we only show outcomes if there are any, otherwise we'll see the table empty?

cwrw (approved):
one comment otherwise 👍

app/views/finance/participants/_declaration.html.erb
cwrw: I don't think we will ever not have declaration states for a declaration, however outcomes only exist for NPQ declarations and only completed

mooktakim: Ok cool, will hide if empty

cwrw (approved):
thanks mook 🙌

PR #3048, opened 2023-02-27
Replace SQL scope with ActiveRecord one

peteryates:
### Context

Minor change to remove some hardcoded SQL.

- Add between years scope for cohorts
- Replace SQL scope with active record conditions


leandroalemao (approved):
LGTM 👍 😄  tks

PR #3049, opened 2023-02-27
Amend qualified API teacher logic to send empty completion date on failure/voided

cwrw:
### Context
When an outcome is voided or failed to unset the certificate we need to send a request to the API with no completion date

- Ticket: n/a

Amend logic to not send completion date if an outcome is not passed

### Guidance to review

Did I miss anything?


leandroalemao (approved):
LGTM 👍 😄 tks

PR #3053, opened 2023-02-27
Run rubocop with progress formatter too

peteryates:
It's handy to be able to see the results of rubocop runs in the GitHub actions log as well as the JSON artefact. Rubocop [allows multiple `--format` params to be passed in](https://docs.rubocop.org/rubocop/formatters.html) so we can do both.

image


leandroalemao (approved):
LGTM 👍 😄 tks

PR #3055, opened 2023-02-27
Add partnership_id to analytics ecf_inductions table

peteryates:
### Context

Nathan requested that the `partnership_id` field be added to `ecf_inductions` in the anaytics database as it would make downstream querying much easier.

- Add `partnership_id` column to analytics ecf_inductions
- Set `partnership_id` for induction records analysis


tonyheadford (approved):
LGTM

PR #3057, opened 2023-02-28
Stop returning a Sentry object when calculating missing qualification type

cwrw:
### Context

When we can't find an NPQ Outcome qualification type, we are logging this to Sentry. Since this is the final step in the method, the return value from the capture method is set to the qualification type and is attempted to be sent to the qualification API.

- Ticket: n/a

Send an empty qualification type to log the correct error from the API instead.

### Guidance to review
Did I miss anything?


leandroalemao (approved):
LGTM tks 👍😀

PR #3058, opened 2023-02-28
CST-1301: add default induction programme to school cohorts in seeds

ltello:
### Context

- [Ticket](https://dfedigital.atlassian.net/browse/CST-1301): Ensure `SchoolCohort.default_induction_programme` is set when we're seeding schools/induction programmes

- Added new scenarios to create flexible school_cohorts and induction programmes that also set default programmes

### DIFF
--- a/db/new_seeds/scenarios/induction_programmes/cip.rb
+++ b/db/new_seeds/scenarios/induction_programmes/cip.rb
@@ -10,18 +10,22 @@ module NewSeeds
       class Cip
         attr_accessor :induction_programme
 
-        def initialize(school_cohort:, core_induction_programme: nil)
+        def initialize(school_cohort:)
           @school_cohort = school_cohort
-          @core_induction_programme = core_induction_programme
         end
 
-        def build(default_induction_programme: true, default_core_induction_programme: true)
+        def build(default_induction_programme: true)
           tap do
-            @induction_programme = FactoryBot.create(:seed_induction_programme,
-                                                     :cip,
-                                                     school_cohort:,
-                                                     core_induction_programme:)
-            set_defaults!(default_induction_programme:, default_core_induction_programme:)
+            @induction_programme = FactoryBot.create(:seed_induction_programme, :cip, school_cohort:)
+            set_default_induction_programme! if default_induction_programme
+          end
+        end
+
+        def with_core_induction_programme(core_induction_programme: nil, default_core_induction_programme: true)
+          tap do
+            core_induction_programme ||= FactoryBot.create(:seed_core_induction_programme)
+            induction_programme.update!(core_induction_programme:)
+            set_default_core_induction_programme! if default_core_induction_programme
           end
         end
 
@@ -29,20 +33,14 @@ module NewSeeds
 
         attr_reader :school_cohort
 
-        def core_induction_programme
-          return if with_no_core_induction_programme?
+        delegate :core_induction_programme, to: :induction_programme
 
-          @core_induction_programme ||= FactoryBot.create(:seed_core_induction_programme)
+        def set_default_induction_programme!
+          school_cohort.update!(default_induction_programme: induction_programme)
         end
 
-        def set_defaults!(default_induction_programme:, default_core_induction_programme:)
-          defaults = default_core_induction_programme ? { core_induction_programme: } : {}
-          defaults.merge!(default_induction_programme: induction_programme) if default_induction_programme
-          school_cohort.update!(**defaults) if defaults.present?
-        end
-
-        def with_no_core_induction_programme?
-          @with_no_core_induction_programme ||= @core_induction_programme == :none
+        def set_default_core_induction_programme!
+          school_cohort.update!(core_induction_programme:)
         end
       end
     end
--- a/db/new_seeds/scenarios/induction_programmes/cip.rb
+++ b/db/new_seeds/scenarios/induction_programmes/cip.rb
@@ -15,18 +15,18 @@ module NewSeeds
         end
 
         def build(default_induction_programme: true)
-          tap do
-            @induction_programme = FactoryBot.create(:seed_induction_programme, :cip, school_cohort:)
-            set_default_induction_programme! if default_induction_programme
-          end
+          @induction_programme = FactoryBot.create(:seed_induction_programme, :cip, school_cohort:)
+          set_default_induction_programme! if default_induction_programme
+
+          self
         end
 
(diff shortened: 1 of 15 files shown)


db/new_seeds/scenarios/induction_programmes/cip.rb
```diff
@@ -0,0 +1,50 @@
...
+        end
+
+        def set_defaults!(default_induction_programme:, default_core_induction_programme:)
+          defaults = default_core_induction_programme ? { core_induction_programme: } : {}
+          defaults.merge!(default_induction_programme: induction_programme) if default_induction_programme
+          school_cohort.update!(**defaults) if defaults.present?
+        end
+
+        def with_no_core_induction_programme?
+          @with_no_core_induction_programme ||= @core_induction_programme == :none
```
tonyheadford, line 45:
```suggestion
          @core_induction_programme.blank?
```
Don't think this is worth memoizing?

db/new_seeds/scenarios/induction_programmes/fip.rb
```diff
@@ -0,0 +1,55 @@
...
+
+          @partnership ||= FactoryBot.create(:seed_partnership, cohort:, school:, delivery_partner:, lead_provider:)
+        end
+
+        def set_default_induction_programme!
+          school_cohort.update!(default_induction_programme: induction_programme)
+        end
+
+        def with_no_partnership?
+          @with_no_partnership ||= @partnership == :none
```
tonyheadford, line 50:
```suggestion
          @partnership.blank?
```
memoizing these doesn't seem worth it to me?

db/new_seeds/scenarios/participants/transfers/fip_to_fip_changing_training_provider.rb
```diff
@@ -17,9 +15,11 @@ def build
...
             Rails.logger.info("seeded transfer of #{participant_profile.full_name} from #{school_from.name} to #{school_to.name} using the new school's training provider")
 
-            FactoryBot.create(:seed_induction_record,
-                              participant_profile:,
-                              induction_programme: induction_programme_to)
+            Induction::TransferToSchoolsProgramme.call(participant_profile:,
+                                                       induction_programme: induction_programme_to,
+                                                       start_date:,
+                                                       email:,
+                                                       mentor_profile:)
```
tonyheadford, line 22: I thought we were trying to avoid using services in scenarios/seeds?

db/new_seeds/scenarios/participants/transfers/fip_to_fip_keeping_original_training_provider.rb
```diff
@@ -17,29 +13,15 @@ def initialize
...
-            @relationship_induction_record = FactoryBot.create(:seed_induction_record,
-                                                               participant_profile:,
-                                                               induction_programme: relationship_induction_programme)
+            @induction_programme_to = Induction::TransferAndContinueExistingFip
+                                        .call(school_cohort: school_cohort_to,
+                                              participant_profile:,
+                                              email:,
+                                              start_date:,
+                                              mentor_profile:)
+                                        .induction_programme
```
tonyheadford, line 24: Same question about using services here, would be good to get your thoughts.

tonyheadford (commented):
A couple of suggestions and comments but looks great.

db/new_seeds/scenarios/induction_programmes/cip.rb
ltello: This is to check the value given to that parameter is the special one `:none` meaning: create no core_induction_programme

db/new_seeds/scenarios/induction_programmes/fip.rb
ltello: Maybe I can remove the memoization. The idea is not to search for that symbol every time this method is called.

db/new_seeds/scenarios/participants/transfers/fip_to_fip_changing_training_provider.rb
ltello: Yes but if I don't call the service I would need to fully replicate its code here. 
I think that would make this very dirty.
And if we change the code to transfer a participant, we would need to changed this duplicated code?

db/new_seeds/scenarios/induction_programmes/cip.rb
ltello: Ok, refactored and created  the chainable method `with_core_induction_programme` .
That will make creating the core induction programme really optional.

db/new_seeds/scenarios/induction_programmes/fip.rb
ltello: Ok, refactored and created  the chainable method `with_partnership`.
That will make creating the partnership really optional.

db/new_seeds/scenarios/participants/transfers/fip_to_fip_keeping_original_training_provider.rb
ltello: I can't think of any good reason to not use service objects in the seeding process. 
In any case, I have refactored it all to create the final records "by hand"

tonyheadford (approved):
Looks great. One observation though - unless I'm missing something doesn't the `tap` blocks  in the methods instead of just returning `self` just add an extra line of code and perhaps make it slightly less obvious what's going on e.g. I've had to stop and wonder why?

db/schema.rb
```diff
@@ -645,7 +645,7 @@
     t.boolean "sparsity_uplift"
     t.boolean "pupil_premium_uplift"
     t.uuid "delivery_partner_id"
-    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY (ARRAY[('submitted'::character varying)::text, ('eligible'::character varying)::text, ('payable'::character varying)::text, ('paid'::character varying)::text, ('clawed_back'::character varying)::text]))"
+    t.index ["cpd_lead_provider_id", "participant_profile_id", "declaration_type", "course_identifier", "state"], name: "unique_declaration_index", unique: true, where: "((state)::text = ANY ((ARRAY['submitted'::character varying, 'eligible'::character varying, 'payable'::character varying, 'paid'::character varying, 'clawed_back'::character varying])::text[]))"
```
peteryates, line 648: Ah these changes snook in again. Out of interest, which version of PostgeSQL are you running @ltello?

ltello: PostgreSQL 13 @peteryates

PR #3060, opened 2023-03-01
Add all expected default schedules for NPQ EHCO

cwrw:
### Context
When selecting default schedules we are assuming all those are in the setup, as they are selected according to time ranges. 

- Ticket: N/A

Ensure all those schedules are available in our test set up, to avoid any future failures

### Guidance to review
Did I miss anything?

### DIFF
--- a/spec/support/with_default_schedules.rb
+++ b/spec/support/with_default_schedules.rb
@@ -15,7 +15,7 @@ RSpec.shared_context "with default schedules", shared_context: :metadata do
       npq_specialist_schedule: %w[npq-specialist-spring npq-specialist-autumn],
       npq_leadership_schedule: %w[npq-leadership-spring npq-leadership-autumn],
       npq_aso_schedule:        %w[npq-aso-december],
-      npq_ehco_schedule:       %w[npq-ehco-december],
+      npq_ehco_schedule:       %w[npq-ehco-november npq-ehco-december npq-ehco-march npq-ehco-june],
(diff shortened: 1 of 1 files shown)


spec/support/with_default_schedules.rb
```diff
@@ -15,7 +15,7 @@
       npq_specialist_schedule: %w[npq-specialist-spring npq-specialist-autumn],
       npq_leadership_schedule: %w[npq-leadership-spring npq-leadership-autumn],
       npq_aso_schedule:        %w[npq-aso-december],
```
ltello, line 17: What about these 3 above?

cwrw: good question, `npq_aso_schedule` we don't pick the schedule according to time, so should be ok, the other two include all the schedules we pick according to the time so hopefully also will be ok

ltello: Ah ok! I think those are all the variants. Forget about my comment.

ltello (approved):

spec/support/with_default_schedules.rb
cwrw: > Wouldn't it be better to make it Date.current dependent?
To unblock main, if this passes I'm hoping we can merge, and I'll spend more time on your suggestion to avoid having to set up all schedules and rely on the defaults at the time, as that will require more time and I don't want to block the pipeline. Does that sound ok?

PR #3063, opened 2023-03-01
Integration test for participant outcomes and separate BigQuery jobs into own queue

slorek:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2054

- Adds an integration spec of the happy path when a provider submits a participant outcome. The spec verifies that the outcome is stored correctly and sent to the qualified teachers API, and the API's response is stored.

### Guidance to review

- I created a new `integration` specs folder as it didn't really fit in the mold of a `request` or `feature` spec. Ideally, we would add more integration specs when appropriate. If so, there will be opportunities to extract some helpers from this spec to avoid code duplication and make it more readable.
- I created a `big_query` queue because there was a job configured to run on the `participant_outcomes` queue which was out of scope of the spec (for now) and so I didn't want it to run, and it made sense to group these together, since they can potentially be deprioritised as timely delivery to BigQuery is not as critical as some of the other jobs. On other services we have seen a large backlog of BigQuery sync jobs which caused long delays to mail delivery, preventing users logging in.


cwrw (commented):
I'm wondering about the longevity of the integration folder: 
Due to the task we've created being a one off and manual I agree testing it end to end does not fit our current testing structure. However when this moves to an automated format soon I believe the test to confirm its sent to the TRA API when an outcome is created via the API will then live in the request specs, which we use to test anything from creating the records, any side effects that follow, which I think this will then fit into nicely.

Maybe as a half way point for now, we can move your test to the [request specs](https://github.com/DFE-Digital/early-careers-framework/blob/main/spec/requests/api/v1/participant_declarations_spec.rb#L576) with the comments to amend when this is automated, as you've done most of the work already?

peteryates (commented):
Code wise I think it's great, definitely doing the right thing and the spec is easy to follow.

I agree with @cwrw that having a lone integration spec feels a touch odd - with it being of type `:request`, would `spec/requests/integration` make more sense?

cwrw (approved):
just the suggestion around where this should live otherwise all good, 👍 for big query queue

PR #3065, opened 2023-03-02
Update Handle notify callback job to log missing emails.

cwrw:
### Context

Jobs are failing due to emails missing. We'd like more information on which templates are failing, log this as a warning to allow us to investigate further, and avoid the jobs failing and retrying

- Ticket: n/a

If email doesn't exist log it as a warning with the template

### Guidance to review
Did I miss anything?


tonyheadford (approved):
Great!

PR #3066, opened 2023-03-03
CST-1331: partnerships when setting up a school cohort might exist al…

ltello:
…ready. Reuse it

### Context

- [Ticket](https://dfedigital.atlassian.net/browse/CST-1331): 

  - Reuse partnership when a school amends a cohort setup and the partnership already exists for any reason.

### DIFF
--- a/app/models/partnership.rb
+++ b/app/models/partnership.rb
@@ -54,8 +54,7 @@ class Partnership < ApplicationRecord
     challenged_at.nil? && challenge_reason.nil? && pending == false
   end
 
-  def unchallenge!(**attrs)
-    assign_attributes(attrs)
+  def unchallenge!
     update!(challenged_at: nil, challenge_reason: nil, challenge_deadline: cohort_challenge_deadline)
   end
 
--- a/app/models/partnership.rb
+++ b/app/models/partnership.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Partnership < ApplicationRecord
+  CHALLENGE_PERIOD_SINCE_ACADEMIC_YEAR_START = 2.months
+
   enum challenge_reason: {
     another_provider: "another_provider",
     not_confirmed: "not_confirmed",
@@ -52,7 +54,16 @@ class Partnership < ApplicationRecord
     challenged_at.nil? && challenge_reason.nil? && pending == false
   end
 
+  def unchallenge!(**attrs)
+    assign_attributes(attrs)
+    update!(challenged_at: nil, challenge_reason: nil, challenge_deadline: cohort_challenge_deadline)
+  end
+
 private
+
+  def cohort_challenge_deadline
+    cohort.academic_year_start_date + CHALLENGE_PERIOD_SINCE_ACADEMIC_YEAR_START
+  end
 
   def update_analytics
     Analytics::UpsertECFPartnershipJob.perform_later(partnership: self) if saved_changes?
--- a/app/controllers/schools/setup_school_cohort_controller.rb
+++ b/app/controllers/schools/setup_school_cohort_controller.rb
@@ -180,15 +180,11 @@ module Schools
     end
 
     def use_the_same_training_programme!
-      # Copy the previous active partnership for the new cohort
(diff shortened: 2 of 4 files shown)


app/models/partnership.rb
```diff
@@ -52,8 +54,17 @@ def active?
     challenged_at.nil? && challenge_reason.nil? && pending == false
   end
 
+  def unchallenge!(**attrs)
+    assign_attributes(attrs)
+    update!(challenged_at: nil, challenge_reason: nil, challenge_deadline: cohort_challenge_deadline)
+  end
+
```
tonyheadford, line 61: we moved `challenge!` out of the model to reduce logic in there, this kind of feels like the same.  I think this logic could be incorporated into the `Partnerships::Report` service or as a separate `Partnerships::Unchallenge` service? What do you think?

ltello: Yes, I considered it but I think there is a difference with the `challenge` service object:
- `challenge` not only affects the Partnership instance (setting `challenged_at`, `challenge_reason`, etc) but also
the `school_cohort` instance (setting `:no_early_career_teachers` choice, nullifying `default_induction_programme`...) and even sending email notifications.

- this `unchallenge` however only needs to affect the Partnership instance itself (cleaning challenge related fields) so like an internal action inside a partnership.

ltello: I also think `Partnership` should have its own `challenge!` method extracted from the service object.

ltello: So, IMO the `Partnerships::Challenge` service object makes perfect sense given that its job is not only to challenge a partnership object itself but also, adjust their associated school cohort, notify users, etc.

However, internal `challenge!` and `unchallenge!` methods inside Partnership makes sense as well because those are task involving only a partnership object and therefore internal to the object itself, exposing only an interface for the outside world to invoke them.

tonyheadford: Funny enough IIRC it used to be like that but one of the reasons that was pulled out was because people were using it not realising all the side affects or tasks that may also need to be done in the process (but there is always nothing to stop someone manually changing things).
I also don't think that passing in a set of unspecified args to `unchallenge` is a good thing, if there was to be a simple `unchallenge!` method then I don't think it should take args like this as this is no longer a single concern, you're updating other attributes potentially as well as doing the unchallenge operation don't you think?

ltello: > Funny enough IIRC it used to be like that but one of the reasons that was pulled out was because people were using it not realising all the side affects or tasks that may also need to be done in the process (but there is always nothing to stop someone manually changing things).

I get it, but the issue then was the lack of a service object to serve the whole feature of a school rejecting the  partnership created by a LP for an specific cohort, not that the subtask inside partnership was wrong.

I think the good practice is that each responsability must be put in their right place instead of developing defensively to prevent people from doing wrong things. IMO code reviews are for that purpose instead.


> I also don't think that passing in a set of unspecified args to `unchallenge` is a good thing, if there was to be a simple `unchallenge!` method then I don't think it should take args like this as this is no longer a single concern, you're updating other attributes potentially as well as doing the unchallenge operation don't you think?

I agree with you but (you won't believe it!) this is the result of removing `tap` from the initial implementation: 
```
      new_partnership = @school.active_partnerships
                               .find_by(cohort: previous_cohort, relationship: false)
                               .dup
                               .tap do |partnership|
        partnershp.cohort_id = cohort.id
      end
      lead_provider = new_partnership.lead_provider
      delivery_partner = new_partnership.delivery_partner
      existing_partnership = school.partnerships.find_by(cohort:, lead_provider:, delivery_partner:)
      existing_partnership ? existing_partnership.unchallenge! : new_partnership.unchallenge!
```
More than happy to revert it and have `unchallenge!` signature with no parameters!!

tonyheadford: That sounds good. I think overall we (as a team) had decided to prefer functional/operational code in services and out of the models and given we already had the "challenge" functionality in a service it seemed fairly logical to have unchallenge as a service too (regardless of what's involved/side affects etc.), but I see your point, it's a blurry line.

PR #3067, opened 2023-03-04
Move CSV uploads from ActiveStorage to an array of URNs

peteryates:
### Context

We will soon be migrating the ECF app from GOV.UK PaaS to Azure. The application's data is currently spread across four places:

- Production database (PostgreSQL)
- analytics database (PostgreSQL)
- Sidekiq database (Redis)
- File storage (S3 via PaaS)

The first three are easy to migrate and will involve a restore of the database in the new environment.

The file storage move is difficult because S3 and Azure Blob Storage aren't identical. We'd need to use ActiveStorage mirrors and rely on a syncing process before switching the 'master' copy.

The uploaded CSVs only contain a single column of URNs. We can easily use PostgreSQL's array functionality to save the provided list of URNs. This will make the migration considerably easier and remove a unnecessary infrastrucutre component in the process.

- Add new column for uploaded partnership URNs
- Add PartnershipCsvUpload#sync_uploaded_urns
- Write uploaded urns to the uploaded_urns field
- Add a job that performs the CSV to array sync

### After merge

This is designed to be run using something like:

```
PartnershipCsvUpload.find_in_batches do |batch|
  batch.each do |partnership_csv_upload|
    SyncPartnershipCsvUploadJob.perform_now(partnership_csv_upload_id: partnership_csv_upload.id)
  end
end
```

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2023_02_22_142649) do
+ActiveRecord::Schema.define(version: 2023_03_04_184644) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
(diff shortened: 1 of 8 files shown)


db/schema.rb
```diff
@@ -753,6 +753,7 @@
     t.datetime "updated_at", precision: 6, null: false
     t.uuid "delivery_partner_id", null: false
     t.uuid "cohort_id"
+    t.integer "uploaded_urns", array: true
```
tonyheadford, line 756: Should this be a `string` array to match what a URN currently is/prevent issues with typos trailing zeroes ?

tonyheadford (approved):
Nice idea, although left a comment about the array type

db/schema.rb
peteryates: Good call, will switch to string. I came at it purely from a GIAS point of view.

PR #3073, opened 2023-03-06
[CPDLP-2063] Delete participant schedules when moving an npq application to pending

cwrw:
### Context
When attempting to move an NPQ applicaton back to pending we are getting an error as we aren't destroying all associated 
records to the profile

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2063

When changing a schedule we create a `ParticipantProfileSchedule`. Ensure we delete those records when deleting the profile.

### Guidance to review
Did I miss anything?

### DIFF
--- a/app/services/npq/change_to_pending.rb
+++ b/app/services/npq/change_to_pending.rb
@@ -28,6 +28,7 @@ module NPQ
       ActiveRecord::Base.transaction do
         if profile
           profile.participant_profile_states.destroy_all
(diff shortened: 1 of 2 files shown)


leandroalemao (commented):
@cwrw Sorry for my delay on that.. LGTM.. tks 👍 😄

PR #3075, opened 2023-03-06
[CPDLP-2037] Change API endpoints and lookups to surface User ID

leandroalemao:
### Context

- Ticket: [CPDLP-2037](https://dfedigital.atlassian.net/browse/CPDLP-2037)

After an investigation, we found that we surface user IDs and external identifiers from participant identity external identifier interchangeably in the API. For some requests we return user IDs, and in some the external identifiers. These values are not equivalent and have caused issues for providers when attempting to map participants in NPQ applications to NPQ participants, and when recording declarations.

We would like to change the IDs we surface to providers to only be User IDs. We’re informing providers of the changed IDs so they can update their integrations accordingly.

[CPDLP-2037]: https://dfedigital.atlassian.net/browse/CPDLP-2037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/mailers/participant_transfer_mailer.rb
+++ b/app/mailers/participant_transfer_mailer.rb
@@ -90,7 +90,7 @@ class ParticipantTransferMailer < ApplicationMailer
       rails_mail_template: action_name,
       personalisation: {
         joining_date: induction_record.start_date.to_date.to_s(:govuk),
-        external_participant_id: induction_record.participant_profile.participant_identity.new_external_identifier,
+        external_participant_id: induction_record.participant_profile.participant_identity.user_id_or_external_identifier,
       },
     ).tag(:provider_transfer_out_notification).associate_with(lead_provider_profile, as: :lead_provider_profile)
   end
@@ -106,7 +106,7 @@ class ParticipantTransferMailer < ApplicationMailer
       rails_mail_template: action_name,
       personalisation: {
         joining_date: induction_record.start_date.to_date.to_s(:govuk),
-        external_participant_id: induction_record.participant_profile.participant_identity.new_external_identifier,
+        external_participant_id: induction_record.participant_profile.participant_identity.user_id_or_external_identifier,
       },
     ).tag(:provider_transfer_in_notification).associate_with(lead_provider_profile, as: :lead_provider_profile)
   end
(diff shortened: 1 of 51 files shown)


cwrw (commented):
Discussed in person but for posterity: we'd like to only change this value for providers, internally we are ok to use external identifier. Also we discussed revisiting lookups to ensure we are using user and then narrowing down using the lead provider and course identifier. Also we discussed naming of new external identifier as well as thinking about fallbacks vs feature flags

app/services/transfers/npq_participants.rb
```diff
@@ -2,31 +2,36 @@
 
 module Transfers
   class NPQParticipants
```
cwrw, line 4: could we avoid changes to this service as it's a tool we use and not hooked up to anything?

app/mailers/participant_transfer_mailer.rb
```diff
@@ -90,7 +90,7 @@ def provider_transfer_out_notification(induction_record:, lead_provider_profile:
       rails_mail_template: action_name,
       personalisation: {
         joining_date: induction_record.start_date.to_date.to_s(:govuk),
-        external_participant_id: induction_record.participant_profile.participant_identity.external_identifier,
+        external_participant_id: induction_record.participant_profile.participant_identity.user_id_or_external_identifier,
```
cwrw, line 93: excellent spot ⭐

app/services/finance/ecf/assurance_report/query.rb
```diff
@@ -88,6 +88,12 @@ def sql
         def where_values
           ParticipantDeclaration::ECF.sanitize_sql_for_conditions(["clp.id = ? AND s.id = ?", statement.cpd_lead_provider_id, statement.id])
         end
+
+        def user_id_or_external_identifier
```
cwrw, line 92: ⭐ excellent spot here as well!

cwrw (commented):
nicely done, excellent spots and attention to detail! I added a few minor comments around using the feature flag in some places. Also please could you add tests for all the serializers with the feature flag vs not, thanks!

leandroalemao:
Hi @cwrw .. think I've addressed all comments.. tks for your review..

Some serialisers refer to `mentor_id` like below.. should we replace the external id by the user id for the mentors as well?

```
attribute :mentor_id do |induction_record|
  if induction_record.participant_profile.ect?
    induction_record.mentor_profile&.participant_identity&.external_identifier
  end
end
```

cwrw:
Thanks @leandroalemao for addressing all the comments, looking good.
Ah good eye and catch! Yes we'll need to do the same for `mentor_id`, i.e. feature flag in serialisers and return both user and external id in the participant query. I think after that change PR should be ready :)

cwrw (approved):
thanks for addressing all comments and spotting all the places we needed to swap, well done 🙌

PR #3092, opened 2023-03-10
Fix flakey NPQ application request and search specs

cwrw:
### Context
https://github.com/DFE-Digital/early-careers-framework/actions/runs/4378534706/jobs/7663398663

To avoid flakey tests with `updated_at` field, freeze time to avoid different times. This is because we now have multiple values to choose from when calculating the NPQ application `updated_at` field.

https://github.com/DFE-Digital/early-careers-framework/actions/runs/4374663161/jobs/7654373465

Explicitly set user name when setting up applications, to avoid flakey tests when users have similar names

- Ticket: N/A

- Freeze time when setting up NPQ applications to avoid different times
- Set exact names on users to avoid accidentally looking up multiple records

### Guidance to review
Did I miss anything?


mooktakim (approved):
👍

PR #3094, opened 2023-03-10
Cohortless Dashboard

ltello:
### Context

- Ticket:


app/views/schools/participants/index.html.erb
```diff
@@ -5,50 +5,91 @@
...
+          <br>
         <% end %>
+      <% end %>
+
+      <% @participants.mentors.keys.sort_by(&:participant_full_name).each do |mentor| %>
+        <div class="group-card">
+          <%= govuk_summary_list(actions: false, borders: nil) do |list| %>
+            <% list.row do |row| %>
+              <% row.key(text: "Mentor") %>
+            <% end %>
```
peteryates, line 39: I don't think this works from a markup point of view. It will output a `<dt>Mentor</dt>` and multiple can be used, but, from [the docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dt):

> It is usually followed by a \<dd\> element; however, multiple \<dt\> elements in a row indicate several terms that are all defined by the immediate next \<dd\> element.

This means if we have more than one mentor the 'heading' will only apply to the first.

I think this should be a heading, which will give the page some hierarchy:

```suggestion
          <h2 class="govuk-heading-m">Mentor</h2>
          <%= govuk_summary_list(actions: false, borders: nil) do |list| %>
```

To a screen reader, when there's a big list of mentors, the page hierarchy will sound like:

> Manage mentors and ECTs, Mentor, Mentor, Mentor, Mentor

We probably want some visually hidden text in there too so it's a bit more informative.

```suggestion
          <h2 class="govuk-heading-m">Mentor <span class="govuk-visually-hidden>- <%= mentor.full_name %></span></h2>
          <%= govuk_summary_list(actions: false, borders: nil) do |list| %>
```

app/views/schools/participants/index.html.erb
```diff
@@ -5,50 +5,91 @@
...
+              <% row.value(text: render(ParticipantStatusTagComponent.new(profile: mentor.participant_profile,
+                                                                          induction_record: mentor,
+                                                                          has_mentees: true))) %>
+            <% end %>
 
-        <% if participants_active?(@categories.mentor_categories) %>
-          <% component.tab(label: 'Mentors') do %>
-            <%= render partial: 'mentors', locals: { mentor_profiles: @categories.mentor_categories } %>
+            <% list.row do |row| %>
+              <% row.key(text: "ECTs") %>
```
peteryates, line 53: Same here, probably should be a heading.

app/views/schools/participants/index.html.erb
```diff
@@ -5,50 +5,91 @@
...
+      <% end %>
 
-        <% if @categories.transferring_in.any? || @categories.transferring_out.any? %>
-          <% component.tab(label: 'Moving school') do %>
-            <%= render partial: 'moving_school', locals: { transferring_in: @categories.transferring_in,
-                                                            transferring_out: @categories.transferring_out } %>
+      <% if @participants.orphan_mentors.any? %>
+        <%= govuk_summary_list(actions: false, borders: nil) do |list| %>
+          <% list.row do |row| %>
+            <% row.key(text: @participants.orphan_mentors.many? ? "Mentors" : "Mentor") %>
```
peteryates, line 76: And here. I think there should probably be some (many visually hidden) text here giving a clue as to why they're listed separately. "Mentors without ECTs" perhaps?

app/views/schools/participants/index.html.erb
```diff
@@ -5,50 +5,91 @@
...
-    <h1 class="govuk-heading-xl">Your ECTs and mentors</h1>
+    <h1 class="govuk-heading-xl">Manage mentors and ECTs</h1>
 
-    <p class="govuk-body">We’ll check your early career teacher (ECTs) and mentors details with the Teaching Regulation Agency to make sure they’re eligible for funding for the programme.</p>
-    <p class="govuk-body">In the meantime ECTs can begin training with their training provider.</p>
+    <div class="wrapper-dashboard-v2 manage-mentors-ects">
+      <%= govuk_button_link_to(
+            "Add ECT or mentor",
+            who_schools_add_participants_path(cohort_id: @school.school_cohorts.dashboard_cohorts.first.start_year),
+            class: "govuk-link--no-visited-state",
```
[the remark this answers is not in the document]
ltello: You are right. I just converted a button to a link styled as button a forgot to remove the visited class.

app/views/schools/participants/index.html.erb
ltello: Done

app/views/schools/participants/index.html.erb
ltello: Done

app/views/schools/participants/index.html.erb
ltello: Done

app/components/participant_status_tag_component.rb
```diff
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
 class ParticipantStatusTagComponent < BaseComponent
-  def initialize(profile:, induction_record: nil)
+  def initialize(profile:, induction_record: nil, has_mentees: false)
```
leoapost, line 4: How about use the ParticipantProfile::Mentor `has_mentees` relationship instead?
i.e. `profile.has_mentees.any?`

app/components/participant_status_tag_component.rb
```diff
@@ -16,26 +17,38 @@ def call
...
+      return { text: "Eligible to start", colour: "green" } if ineligible? && mentor_was_in_early_rollout?
+      return { text: "Not eligible", colour: "red" } if ineligible?
+      return { text: "Contacted for information", colour: "grey" } if latest_email&.delivered?
+      return { text: "Check email address", colour: "grey" } if latest_email&.failed?
+
+      { text: "Contacting for information", colour: "grey" }
+    end
   end
 
+  # Remove this code when we remove FeatureFlag.active?(:cohortless_dashboard) - start #
```
leoapost, line 51: I think we can remove this comment.

app/components/participant_status_tag_component.rb
```diff
@@ -77,4 +90,9 @@ def nqt_plus_one?
   def participant_has_no_qts?
     profile.no_qts?
   end
+  # Remove this code when we remove FeatureFlag.active?(:cohortless_dashboard) - start
```
leoapost, line 93: I think we can remove this comment

app/components/participant_status_tag_component.rb
ltello: This are comments I have put to facilitate the job to be done in the future when we definitely go for cohortless.
At that moment, the feature flag :cohortless_dashboard won't make sense anymore and we should remove it from the service.
If we do, the methods inside those comments will become unreachable, never called from the rest of the service, and would be good to remove them.
These comments are to facilitate the quick detection of those methods and wipe them!

app/components/participant_status_tag_component.rb
ltello: Please, see my comment above

spec/features/schools/participants/add_participants/sit_adding_self_as_mentor/happy_path_sit_adds_self_as_mentor_spec.rb
```diff
@@ -17,7 +17,7 @@
     set_sit_dqt_validation_result
   end
 
-  scenario "Induction tutor can add themselves as a mentor and validates" do
+  xscenario "Induction tutor can add themselves as a mentor and validates" do
```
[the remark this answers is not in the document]
ltello: This test was skipped because in the new dashboard the SITs still has no way to add themselves as mentors.
This is something I need to speak with Nicola.

app/components/participant_status_tag_component.rb
leoapost: makes sense

PR #3097, opened 2023-03-10
Codespaces

peteryates:
This change enables Codespaces for the ECF app. This will help less technical colleagues make direct changes.

The only change that has any risk of affecting the app is the one to the initializer that lets us easily specify the `REDIS_URI` using a environment variable instead of letting it default to `localhost`.

Note, in its current form this PR adds `entr` as a dependency for running `bin/dev` - not ideal but I can't find a way of making Middleman continuously monitor for changes (like `middleman serve`) but actually output built files (like middleman build`). @DFE-Digital/teacher-cpd - any ideas?

| Preview |
| ------- |
| image |

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -23,6 +23,16 @@ if ENV.key?("VCAP_SERVICES")
   end
 end
 
+if ENV.key?("REDIS_URI")
+  Sidekiq.configure_server do |config|
+    config.redis = { url: ENV.fetch("REDIS_URI") }
+  end
+
+  Sidekiq.configure_client do |config|
+    config.redis = { url: ENV.fetch("REDIS_URI") }
+  end
+end
+
 # Sidekiq Cron
 if Sidekiq.server?
   Rails.application.config.after_initialize do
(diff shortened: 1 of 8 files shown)


cwrw (approved):
🙌

config/initializers/sidekiq.rb
```diff
@@ -23,6 +23,16 @@
   end
 end
 
+if ENV.key?("REDIS_URI")
```
peteryates, line 26: `REDIS_URI` is set in the GitHub Codespaces environment variable settings. As per the the docker compose it's `redis://redis`.

peteryates:
So it turns out building the guide using Foreman in a Codespace that uses rvm doesn't play nicely.

I've removed the `api_docs` line from the Procfile and added a section to the readme detailing how to run it separately. I left `entr` in because it should work fine on a mac and is easily installable with brew/macports - but if it turns out to be problematic for any reason happy to switch to another.

cwrw (approved):
tested new approach this morning and works nicely, thanks!

peteryates:
@slorek - mind doing a rereview? Requests have been addressed. Ta.

PR #3102, opened 2023-03-13
Send NPQ application cohort to BigQuery

slorek:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2066

Sends the Cohort start year to the `npq_registration.enrollments_{env}` tables in BigQuery.

### Guidance to review

The `npq_application` factory and specs had some issues which are fixed:

* The record was being saved in the `after_initialize` block. This shouldn't happen as it's unexpected behaviour, and unnecessarily slows down the specs.
* Not all attributes were being set in the `after_initialize` block, resulting in FactoryBot saving the record twice. This resulted in the job being enqueued twice by the model callback.
* The existing specs weren't checking which job was enqueued


spec/factories/npq_application.rb
cwrw, line 1: Could we keep this under the current directory `factories/services/npq/npq_application`? This is because this is a "special" factory directory which is calling services we use in our app. Ideally we shouldn't be doing this in our test setup, but those have been grouped under the `services` directory so they are easier to find. Hopefully we'll move off them when we move to the new seeds

cwrw (approved):

PR #3115, opened 2023-03-15
Add API V3 partnership endpoint

slorek:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2081

Adds the GET /api/v3/partnerships/ecf/:id endpoint to API V3 only, as described in the [API documentation](https://manage-training-for-early-career-teachers.education.gov.uk/api-reference/reference-v3.html#api-v3-partnerships-ecf-id-get).

### DIFF
--- a/app/services/api/v3/ecf/partnerships_query.rb
+++ b/app/services/api/v3/ecf/partnerships_query.rb
@@ -12,7 +12,7 @@ module Api
         end
 
         def partnerships
-          scope = lead_provider.partnerships.includes(:school, :cohort, :delivery_partner)
+          scope = lead_provider.partnerships.includes(:cohort, :delivery_partner, school: :induction_coordinators)
           scope = scope.where(partnerships: { cohort:  with_cohorts }) if filter[:cohort].present?
           scope = scope.where("partnerships.updated_at > ?", updated_since) if updated_since.present?
           scope = scope.where(partnerships: { delivery_partner: [delivery_partner_id_filter] }) if delivery_partner_id_filter.present?
@@ -21,7 +21,7 @@ module Api
         end
 
         def partnership
-          lead_provider.partnerships.includes(:school, :cohort, :delivery_partner).find(params[:id])
+          lead_provider.partnerships.includes(:cohort, :delivery_partner, school: :induction_coordinators).find(params[:id])
         end
 
       private


cwrw (commented):
could you also enable with a feature flag the doc specs `spec/docs/v3/ecf_partnerships_spec.rb`? We can prob just enable both listing and single endpoints and wrap the rest with `api_v3: true`

app/controllers/api/v3/ecf/partnerships_controller.rb
```diff
@@ -18,6 +18,10 @@ def index
           render json: serializer_class.new(paginate(ecf_partnerships)).serializable_hash.to_json
         end
 
+        def show
```
leandroalemao, line 21: @slorek I'd be nice to add a comment like we did in the index action. ie:

```
# Returns a specific ECF partnership given its ID
# Providers can see a specific ECF partnership and which cohorts it applies to via this endpoint
#
# GET /api/v1/partnerships/ecf/:id
#
```

app/services/api/v3/ecf/partnerships_query.rb
```diff
@@ -20,6 +20,10 @@ def partnerships
           scope
         end
 
+        def partnership
+          lead_provider.partnerships.includes(:school, :cohort, :delivery_partner).find(params[:id])
```
cwrw, line 24: actually could we use `:cohort, :delivery_partner, school: :induction_coordinators` here and on [line](https://github.com/DFE-Digital/early-careers-framework/pull/3115/files#diff-7ac983a5a95593f7fca49240aef5ba88fe07d3f1c322980f69c21d20cd2c55baL15), it should do the trick, otherwise all good :)

mooktakim: I'm not sure if there's any benefit to adding `includes` here as we're only retrieving one record.

mooktakim (approved):

app/services/api/v3/ecf/partnerships_query.rb
mooktakim: When we retrieve one record n=1, therefore 1+1.
I don't think it'll make it worse though.

cwrw (approved):
thanks @slorek 🙌

PR #3121, opened 2023-03-17
[CPDLP-2069] Remove the dependency of a cohort being sent in the params for a NPQ registration application

leandroalemao:
### Context

- Ticket: [CPDLP-2069](https://dfedigital.atlassian.net/browse/CPDLP-2069)

As part of the ‘dynamic’ cohort piece of work and also opening up for registration, we will stop accepting the cohort param from the NPQ registration app. This has already been agreed upon with the NPQ reg team and they will then remove it from applications when we stop reading from it.

We receive an application via the API from NPQ registration team, and accept a cohort as an attribute. We should no longer accept this value from the NPQ registration team, and set this value to Cohort.active_npq_registration_cohort. This should default to 2022 as there is no data in the database for this value.

[CPDLP-2069]: https://dfedigital.atlassian.net/browse/CPDLP-2069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/api/v1/npq_profiles_controller.rb
+++ b/app/controllers/api/v1/npq_profiles_controller.rb
@@ -95,7 +95,6 @@ module Api
             :employment_role,
             :employment_type,
             :targeted_delivery_funding_eligibility,
-            :cohort,
             :works_in_nursery,
             :works_in_childcare,
             :kind_of_nursery,
--- a/app/services/npq/build_application.rb
+++ b/app/services/npq/build_application.rb
@@ -75,23 +75,23 @@ module NPQ
     end
 
     def cohort
-      @cohort ||= Cohort.find_by(start_year: npq_application_params[:cohort])
+      @cohort ||= Cohort.active_npq_registration_cohort
     end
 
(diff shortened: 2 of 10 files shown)


cwrw (commented):
looking good! could you amend the request specs for the internal endpoint to not include cohort in the params?

leandroalemao:
> Do we need to update the API spec as well?

Hi @slorek I believe we don't need to update the api spec as the affected endpoints are "internal" endpoints, used by NPQ reg team.. we just need to let them know about the changes..

cwrw:
@leandroalemao thanks for addressing comments, I was having a think around the fact that unfortunately our factories use the build application service, and the repercussions of the change, let's have a chat.

cwrw (approved):
🙌 nicely done

Lockyy (approved):
Nice, good to see this progressing 🙂

PR #3125, opened 2023-03-17
[CPDLP-2071] Add or amend scripts to develop Schedules for 2023 NPQ Cohort

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2071

* Script to create new ECF/NPQ schedule
* CSV file format:
  * schedule-identifier
  * schedule-name
  * schedule-cohort-year
  * milestone-name
  * milestone-declaration-type
  * milestone-start-date
  * milestone-date
  * milestone-payment-date

### Guidance to review

* Backend script only


app/services/cohorts/create_schedule.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module Cohorts
+  class CreateSchedule
+    attr_reader :path_to_csv, :cohort, :klass
+
+    def initialize(path_to_csv:, cohort:, klass:)
+      @path_to_csv = path_to_csv
```
cwrw, line 10: maybe we can fail early if there is no cohort/csv passed to the class?

cwrw (commented):
I wonder if we can reuse this for also seeding the schedules, as it's similar except for creating the cohort (in the seeds). I think we can probably amend the `SeedSchedules` and the wrapping scripts to pass in cohorts easily so we don't have duplication, let me know your thoughts

cwrw (approved):
thanks @mooktakim!

PR #3126, opened 2023-03-18
Filter PII attributes from models

cwrw:
### Context

To avoid PII bubbling up in exception errors us the `filter_attributes` on select models

- Ticket: n/a

- `filter_attributes` filters selected columns when calling #inspect, add to models with PII

### Guidance to review
Did I miss anything?


peteryates (approved):

cwrw (commented):
@slorek added TRNs, thanks

PR #3127, opened 2023-03-19
Attempt 2: Ensure user emails are different to avoid flakey tests in application search spec

cwrw:
### Context

When emails are similar in test setup, spec will fail as it will find the second application.

- Ticket: n/a

Explicitly set different emails to avoid flakey spec

### Guidance to review
Hopefully I got it this time :)


cwrw:
@slorek yes that is true they will never be the same. However in the context of this [flakey test](https://github.com/DFE-Digital/early-careers-framework/actions/runs/4427534175/jobs/7765259589) we are selecting part of a string and passing it through for search , which then is getting to [email matches](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/models/participant_identity.rb#L21). Here we're using `like` to match an email to that string. So if there are is any similar text between both emails although they are unique the test will fail as the records are returned.

I was hoping to set up unique emails which have no similarities to avoid any accidental matching

PR #3138, opened 2023-03-20
Implement API V3 schools endpoint

slorek:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2082

- Implements the `/api/v3/schools/ecf` endpoint described in the [API documentation](https://manage-training-for-early-career-teachers.education.gov.uk/api-reference/reference-v3.html#api-v3-schools-ecf-get).

### Guidance to review

* Error message copy is subject to change pending agreement/approval.
* There are n+1 queries caused by the `School#partnered?` method. I don't think this can be wholly avoided as it would require subqueries to replicate within the parent query, which would therefore not reduce the overall number of queries, and this would duplicate the business logic in a non-obvious place. On balance I think it is better to keep as-is.
* For simplicity I didn't create an unnecessary query class as there is no complicated logic; it's just a one-line query.
* We need to query the `SchoolCohort` model because the `has_many` relationship prevents us using `School`. This meant I needed to tweak the `ApiOrderable` concern to support sorting params on an associated model (`School` in this case).
* I've modified the `ApiFilter` concern to check for required filter params.
* I've updated the API spec to reflect the correct values.

### DIFF
--- a/app/controllers/concerns/api_filter.rb
+++ b/app/controllers/concerns/api_filter.rb
@@ -12,7 +12,7 @@ private
   def validate_filter_param
     errors = []
     errors << "Filter must be a hash" unless filter.as_json.is_a?(Hash)
-    missing_filter_params.each { |param| errors << "#{param} filter must be supplied" }
+    missing_filter_params.each { |param| errors << "The filter '#/#{param}' must be included in your request" }
 
     if errors.any?
       error_factory = Api::ParamErrorFactory.new(params: errors, error: I18n.t(:bad_parameter))
--- a/spec/requests/api/v3/ecf/schools_spec.rb
+++ b/spec/requests/api/v3/ecf/schools_spec.rb
@@ -32,7 +32,7 @@ RSpec.describe "API ECF schools", :with_default_schedules, type: :request, with_
         end
 
         it "returns an error message" do
-          expect(parsed_response["errors"][0]["detail"]).to eq("cohort filter must be supplied")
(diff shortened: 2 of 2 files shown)


cwrw:
> There are n+1 queries caused by the School#partnered? method. I don't think this can be wholly avoided as it would require subqueries to replicate within the parent query, and this would duplicate the business logic in a non-obvious place. On balance I think it is better to keep as-is.

I think that is a very fair point, however I think the repercussions of us adding more database load to an already overloaded service would not be wise as we are expecting all providers to sync their data to pull in new v3 fields.

Instead of duplicating the business logic, maybe we can move it to a shared "in partnership" service, which allows the query to be extended/nested in the schools API query and reused in the model to avoid the n+1s?

cwrw:
> The number of database queries will be the same, as I think we'll need to use a subselect to compute this boolean value (or a count). This might even create more overhead on the database as it has to combine and compute the results, whereas at the moment it's done on the application server. I'd be keen to see how you envisage it working, however.

Ah yea you're right!
The other option would be to do the `partnered?(cohort)` in memory (I know not great):

```
partnerships.detect { |partnership| partnership.challenged_at == nil && partnership.challenge_reason == nil && partnership.cohort_id == cohort.id }.present?
```
If we include `:partnerships` seems to perform much better. I checked the number of partnerships for schools and most have 1/2 and a few outliers up to 6 🤔 

Otherwise all looks great, will confirm error message tomorrow on ticket :)

cwrw (approved):
@slorek looks great approving the rest of the code, if you don't mind having a look at the latest commit I pushed, thanks

PR #3139, opened 2023-03-20
Check induction record schedule before raising validation error when changing schedule

cwrw:
### Context

There are some cases where the profile schedule will be the same as the new schedule we are attempting to change to, however the induction record has a different schedule.

This can happen in transfers, when a profile is amended after a participant has transferred, leaving the induction record with an incorrect schedule, but the profile with the correct one.

Allow the change to happen to align all schedules without raising a validation error

- Ticket: n/a

Add extra check for ECF profiles on induction record schedules to ensure they aren't equal to allow schedules to change

### Guidance to review
Did I miss anything?


mooktakim (approved):
👍

PR #3142, opened 2023-03-21
[CPDLP-2087] Add script to create cohort data

leandroalemao:
### Context

- Ticket: [CPDLP-2087](https://dfedigital.atlassian.net/browse/CPDLP-2087)

To support cohort 2023 for NPQ registration, records relating to the cohort need to be populated. This includes NPQ contracts, schedules and statements.

Create a new script which will consume production data CSVs and creates a new cohort.

[CPDLP-2087]: https://dfedigital.atlassian.net/browse/CPDLP-2087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/importers/create_cohort.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+module Importers
```
cwrw, line 3: I think we need to add `require "csv"` for the review app to work as it's falling over at the seeding stage

leandroalemao: Good catch tks @cwrw

db/legacy_seeds/initial_seed.rb
```diff
@@ -1,21 +1,9 @@
...
-                          registration_start_date: Date.new(2020, registration_month, registration_day),
-                          academic_year_start_date: Date.new(2020, academic_year_start_month, academic_year_start_day))
-
-# Create cohorts since 2021 until Cohort.next
-next_cohort_start_year = Date.current.year + (Date.current.month < academic_year_start_month ? 0 : 1)
-cohorts = (2021..next_cohort_start_year).to_a.map do |start_year|
-  Cohort.find_or_create_by!(start_year:,
-                            registration_start_date: Date.new(start_year, registration_month, registration_day),
-                            academic_year_start_date: Date.new(start_year, academic_year_start_month, academic_year_start_day))
+cohorts = (2021..Cohort.next.start_year).to_a.map do |start_year|
```
cwrw, line 5: I like the idea of using the new seed data here, however we will lose the dynamic aspect, where when cohort 2024 starts `Cohort.next.start_year` will then fail and we'll need to manually add 2024 to the csv

cwrw (commented):
script is perfect! Just a few comments around waiting for Mook's pr to go in, and the dynamic cohorts otherwise 👍

app/services/importers/seed_schedule.rb
```diff
@@ -17,10 +17,7 @@ def call
       next unless row["schedule-identifier"]
 
       year = row["schedule-cohort-year"].to_i
-      cohort = Cohort.find_or_create_by!(start_year: year) do |c|
-        c.registration_start_date = Date.new(year, 5, 10)
-        c.academic_year_start_date = Date.new(year, 9, 1)
-      end
+      cohort = Cohort.find_by!(start_year: year)
```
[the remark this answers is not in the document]
mooktakim: @leandroalemao my pr is merged in now, you can rebase main now.

app/services/importers/create_cohort.rb
```diff
@@ -0,0 +1,85 @@
...
+      Cohort.find_or_create_by!(start_year:) do |c|
+        c.registration_start_date = safe_parse(row["registration-start-date"])
+        c.academic_year_start_date = safe_parse(row["academic-year-start-date"])
+        c.npq_registration_start_date = safe_parse(row["npq-registration-start-date"])
+      end
+
+      logger.info "CreateCohort: Cohort for starting year #{start_year} successfully created"
+    end
+
+    def create_next_cohort
```
cwrw, line 44: I wonder if this could be dangerous in production, because we are only expecting to create cohorts according to the CSV only, where this might create the next cohort accidentally in production. Preferably we can avoid mixing dynamic seed data logic with the service to create Cohorts according to a given CSV. 

Maybe the best middle ground is to leave the seeds where we use dynamic calculations as is, as in the logic for dynamic cohorts remains the same and the service class remains as it was before, and only reads from the CSV?
Or we return early in this method if the environment is prod?

cwrw (approved):
thanks @leandroalemao :)

PR #3147, opened 2023-03-21
[CPDLP-2072] Add or amend scripts to develop NPQ Contract information for 2023 Cohort

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2072

* Updated seed script to use for production
* Change `course_name` to `course_identifier` in csv file.
* CSV columns:
  * provider_name
  * cohort_year
  * course_identifier
  * recruitment_target
  * per_participant
  * service_fee_installments

### DIFF
--- a/app/services/importers/npq_contracts.rb
+++ b/app/services/importers/npq_contracts.rb
@@ -1,14 +1,15 @@
 # frozen_string_literal: true
 
+require "csv"
+
 module Importers
   # Given a CSV with correct data
   # Find or create NPQContract
   class NPQContracts
-    attr_reader :path_to_csv, :errors
+    attr_reader :path_to_csv
 
     def initialize(path_to_csv:)
       @path_to_csv = path_to_csv
-      @errors = []
     end
 
     def call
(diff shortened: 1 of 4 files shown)


app/services/importers/npq_contracts.rb
```diff
@@ -18,7 +19,7 @@ def call
         cohort = Cohort.find_by!(start_year: row["cohort_year"])
         cpd_lead_provider = CpdLeadProvider.find_by!(name: row["provider_name"])
         npq_lead_provider = cpd_lead_provider.npq_lead_provider
-        course = NPQCourse.where("name ilike '%#{row['course_name']}%'").first
+        course = NPQCourse.find_by!(identifier: row["course_identifier"])
 
         contract = NPQContract.find_or_initialize_by(
           cohort:,
```
cwrw, line 25: I think we need to be careful with versions, and either fail or bump up the version if there is an existing contract rather than attempt to update it 🤔 ?

mooktakim: @cwrw 
I don't think we use versions with NPQContract, versions are used for ECF

cwrw (approved):
🙌

PR #3152, opened 2023-03-22
CST-1281: search for existing participant in TeacherProfile rather th…

ltello:
…an validation data

### Context

  To detect wether a participant already exists at registration time we actually check the trn stored in ECFParticipantValidationData. This is preventing the registration of teachers whose details have been used by the SIT to register a different person.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-1281)

 Search for existing participants in TeacherProfile instead.


tonyheadford (approved):
Nice!

PR #3155, opened 2023-03-22
NPQ Application and Test data flakey test fixes

cwrw:
### Context

https://github.com/DFE-Digital/early-careers-framework/actions/runs/4492527810/jobs/7902536598
https://github.com/DFE-Digital/early-careers-framework/actions/runs/4482598839/jobs/7880839932

- Ticket: n/a

- `spec/requests/admin/test_data_spec.rb:55`: set explicit names to avoid names such as "O'Really" which are rendered with different characters on the page and will fail when attempting to match in the spec
- `spec/models/npq_application_spec.rb:39`: set explicit year on cohort which won't happen to match the current year to avoid unique index failures

### DIFF
--- a/spec/models/npq_application_spec.rb
+++ b/spec/models/npq_application_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe NPQApplication, type: :model do
     end
 
     context "when cohort_id is modified" do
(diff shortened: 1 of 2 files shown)


leandroalemao (approved):
spot on tks 👍 😄

PR #3159, opened 2023-03-23
[CST-1286] Merge add and transfer journeys for participants

tonyheadford:
### Context

**There's still some spec additions needed for this, however I am starting work on a ticket that will be based off of this PR to make these journeys cohort-less so there will be more changes and have agreed that we will add the missing specs in during that work**

- Ticket: [ECTs ticket](https://dfedigital.atlassian.net/browse/CST-1286) & [Mentors ticket](https://dfedigital.atlassian.net/browse/CST-1325)

Combine and update the existing add and transfer journeys for ECTs and Mentors into one, where we work out what needs to happen to simplify the decisions for the SITs.
Also incorporate a new "wizard" approach with much less magic, that should be easier to change and maintain, based upon the wizards used in NPQ reg.

Quick wizard overview:
There is a routing scope to simplify implementing this approach in the routes.
For each wizard there needs to be a controller, a form and one or more WizardSteps.
Each step is a page in the journey and has an associated view (which view to render can be overridden in the step).
Each step validates the input from any form fields in it's view.
Each step is responsible for providing which step should be next and previous in the sequence.

Merger of the Add and Transfer participant journeys
Reducing the happy path questions
New wizard implementation
Refactor `ParticipantValidationService` to extract the DQT record retrieval and initial checks into a new service `DqtRecordCheck` which exposes the data matches and record to the caller that enable us to know that we found a record but the name doesn't match for example.
Adds a `FeatureFlag` to temporarily prevent 2023/24 registrations (in lieu of a better approach)

### Guidance to review

Starting point is the same as a SIT from the schools participants dashboard, except now add and transfer are the same journey.

I've added an extra magic date to test the "are they known by another name" loop (using `2/1/1900` as the date of birth will return as if the TRN matched but the name did not) - however you will be in this loop forever unless you use the back-link at some point and change the DoB to `1/1/1900`.
There is also a third magic date `3/1/1900` that returns the "not found" response which can be used to reach the new "Nino" question part of the journey, but again you're in a loop until the DoB is changed again.

Alternatively use a DQT test user and change the data entered to navigate the journey as you like.


app/forms/wizard_step.rb
```diff
@@ -0,0 +1,49 @@
...
+    false
+  end
+
+  def before_render; end
+
+  def view_name
+    # remove any module scope and 'Step' suffix from the class name as default view name for the step
+    # e.g. for Module::Group::AmazingThingStep we get "amazing_thing" back
+    # This can be overridden in steps that need to support multiple views.
+    self.class.name.demodulize[...-4].underscore
```
peteryates, line 36: Could this work here instead?

```suggestion
    self.class.name.delete_suffix("Step").underscore
```

app/views/schools/add_participants/_ect_or_mentor_answers.html.erb
```diff
@@ -0,0 +1,94 @@
...
+  <% if form.date_of_birth %>
+    <% summary_list.row do |row| %>
+      <% row.key { "Date of birth" } %>
+      <% row.value { form.date_of_birth.to_date.to_s(:govuk) } %>
+      <% unless form.found_participant_in_dqt? %>
+        <% row.action(text: "Change",
+                      visually_hidden_text: "date of birth",
+                      href: form.change_path_for(step: :date_of_birth)) %>
+      <% else %>
+        <% row.action() %>
```
peteryates, line 37: More recent versions of the summary list don't need a blank action, should be safe to drop (and if not we're best upgrading `govuk-frontend`)

app/views/schools/add_participants/add_ect/_eligibility_confirmation.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <%= render partial: 'schools/add_participants/add_ect/eligibility_confirmation/manual_check_needed', locals: { profile: profile } %>
+<% elsif previous_induction?(profile) %>
+  <%= render partial: 'schools/add_participants/add_ect/eligibility_confirmation/previous_induction', locals: { profile: profile } %>
+<% elsif ineligible_mentor_in_early_rollout?(profile) %>
+  <%= render partial: 'schools/add_participants/add_ect/eligibility_confirmation/eligible_ero', locals: { profile: profile } %>
+<% elsif profile.no_qts? %>
+  <%= render partial: 'schools/add_participants/add_ect/eligibility_confirmation/no_qts_reason', local: { profile: profile } %>
+<% elsif ineligible?(profile)%>
+  <%= render partial: 'schools/add_participants/add_ect/eligibility_confirmation/ineligible', local: { profile: profile } %>
+<% end %>
```
peteryates, line 13: Perhaps this would fit better in a helper?

app/views/schools/add_participants/add_ect/complete.html.erb
```diff
@@ -0,0 +1,33 @@
...
+        <p class="govuk-body">
+          We’ll let this person know you’ve registered them for ECF-based training at your school.
+          They do not need to provide us with any more information.
+        </p>
+
+        <h2 class="govuk-heading-m">Eligibility for funding</h2>
+        <%= render partial: 'eligibility_confirmation', locals: { profile: @profile } %>
+      <% else %>
+          <%= render partial: 'no_eligibility_record' %>
+      <% end %>
```
peteryates, line 27:
```suggestion
      <% else %>
        <%= render partial: 'no_eligibility_record' %>
      <% end %>
```

app/views/schools/add_participants/add_ect/complete.html.erb
```diff
@@ -0,0 +1,33 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <% if @profile.user == current_user %>
+      <%= govuk_panel title_text: "You’ve been added as a mentor", text: nil, classes: "govuk-!-margin-bottom-7" %>
+    <% elsif exempt_from_induction?(@profile) %>
+      <%= render partial: 'schools/add_participants/eligibility_confirmation/exempt_from_induction', locals: { profile: @profile } %>
+    <% else %>
+      <% if @profile.ect? %>
+        <%= govuk_panel title_text: "#{@profile.user.full_name} has been added as an ECT", text: nil, classes: "govuk-!-margin-bottom-7" %>
+      <% else %>
+          <%= govuk_panel title_text: "#{@profile.user.full_name} has been added as a mentor", text: nil, classes: "govuk-!-margin-bottom-7" %>
```
peteryates, line 13:
```suggestion
        <%= govuk_panel title_text: "#{@profile.user.full_name} has been added as a mentor", text: nil, classes: "govuk-!-margin-bottom-7" %>
```

app/services/dqt_record_check.rb
```diff
@@ -0,0 +1,126 @@
...
+
+    if matches < 3 && (trn_matches && trn != "1")
+      # If a participant mistypes their TRN and enters someone else's, we should search by NINO instead
+      # The API first matches by (mandatory) TRN, then by NINO if it finds no results. This works around that.
+      @trn = "0000001"
+      return check_record
+    end
+
+    CheckResult.new(dqt_record, trn_matches, name_matches, dob_matches, nino_matches, matches)
+  end
```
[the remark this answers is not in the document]
tonyheadford: That was really the point of breaking this out of the `ParticipantValidationService` - grab the record and do the matching checks - but also allow us to return the DQT record and the check results to the caller.  Do you think this needs breaking down further? I'm not sure what value it would have outside of this?

app/views/schools/add_participants/add_ect/_eligibility_confirmation.html.erb
tonyheadford: Yes agree, this file was just lifted directly from the existing code

app/forms/wizard_step.rb
tonyheadford: Nice, TIL `delete_suffix`

app/services/dqt_record_check.rb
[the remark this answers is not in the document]
tonyheadford: I've simplified this a bit now by adding a present for the returned dqt record

leoapost:
@tonyheadford while checking the transfer journey, I've noticed that mentors remain in the mentor pool of the previous schools despite choosing they won't be mentoring in other schools.

Here is an example of a [mentor](https://ecf-review-pr-3159.london.cloudapps.digital/admin/participants/30824a97-0f06-44f2-846e-9835abe0fd08/details) I transferred twice.You can see the links to the previous school mentor pools in the `school` tab.

tonyheadford:
> @tonyheadford while checking the transfer journey, I've noticed that mentors remain in the mentor pool of the previous schools despite choosing they won't be mentoring in other schools.
> 
> Here is an example of a [mentor](https://ecf-review-pr-3159.london.cloudapps.digital/admin/participants/30824a97-0f06-44f2-846e-9835abe0fd08/details) I transferred twice.You can see the links to the previous school mentor pools in the `school` tab.

Ah great spot @leoapost, I thought someone was adding a fix to the service for this, perhaps that hasn't been done yet then!

leoapost:
> > @tonyheadford while checking the transfer journey, I've noticed that mentors remain in the mentor pool of the previous schools despite choosing they won't be mentoring in other schools.
> > Here is an example of a [mentor](https://ecf-review-pr-3159.london.cloudapps.digital/admin/participants/30824a97-0f06-44f2-846e-9835abe0fd08/details) I transferred twice.You can see the links to the previous school mentor pools in the `school` tab.
> 
> Ah great spot @leoapost, I thought someone was adding a fix to the service for this, perhaps that hasn't been done yet then! -- edit -- yes, this has been done, it does defer removing the mentor if the leaving date is in the future I noticed, I wonder if that's what you saw?

Right, my bad, I didn't notice that part. It makes sense :+1:

leoapost (commented):
Tremendous work, thanks @tonyheadford! 

I left a couple of questions about some commented code I found, but other than that it looks great.

spec/forms/schools/add_participants/add_wizard_spec.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-RSpec.describe Schools::AddParticipantForm, type: :model do
+# TODO: needs refactoring for new wizard form
```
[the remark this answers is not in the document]
tonyheadford: Well if we had the time yes, but no, I added that in as a reminder that we need to do this. This is about to have some changes to support making this cohort-less so we agreed with Nicola to finish these up as part of that work.

leoapost (approved):

app/views/schools/add_participants/nino.html.erb
```diff
@@ -1,26 +1,23 @@
...
-
-    </div>
+    <%= form_with model: @form, url: url_for(action: :update), scope: @wizard.form_scope, method: :put do |f| %>
+      <%= f.govuk_error_summary %>
+      <%= f.govuk_text_field :nino,
+                             width: "two-thirds",
+                             spellcheck: false,
+                             label: { text: title, size: 'xl', tag: 'h1', class: 'govuk-heading-xl' },
+                             hint: -> { tag.div("It’s on their National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’.",
+                                                class: "govuk-hint") } %>
```
peteryates, line 19:
```suggestion
                             hint: { text: "It’s on their National Insurance card, benefit letter, payslip or P60. For example, ‘QQ 12 34 56 C’." } %>
```

app/views/schools/add_participants/still_cannot_find_their_details.html.erb
```diff
@@ -13,29 +12,37 @@
...
-      <li>date of birth</li>
-    </ul>
-
-    <p>
-      <%= govuk_link_to "View your ECTs and mentors",
-                        schools_participants_path,
-                        class: "govuk-link govuk-link--no-visited-state" %>
+      Contact us for help to register this <%= @wizard.ect_participant? ? "ECT" : "mentor" %> at your school: <%= render MailToSupportComponent.new %>
     </p>
+    <%= govuk_link_to "Return to your ECTs and mentors", schools_participants_path, class: "govuk-link govuk-link--no-visited-state" %>
```
peteryates, line 46: We can [lean on the link helper](https://govuk-components.netlify.app/helpers/link/#other-link-styles) for these classes:

```suggestion
    <%= govuk_link_to "Return to your ECTs and mentors", schools_participants_path, no_visited: true %>
```

peteryates (approved):
Looks great, left a few super-minor comments that can be done in the follow up.

app/views/schools/add_participants/choose_mentor.html.erb
```diff
@@ -1,28 +1,26 @@
...
-
-            <%= f.govuk_radio_divider %>
-                <%= f.govuk_radio_button :mentor_id, "later", label: { text: 'Assign mentor later' } %>
-            <% end %>
-
-            <%= f.govuk_submit "Continue" %>
+    <%= form_with model: @wizard.form, url: url_for(action: :update), scope: @wizard.form_scope, method: :put do |f| %>
+      <%= f.govuk_error_summary %>
+      <%= f.hidden_field :mentor_id, value: '' %>
+      <%= f.govuk_radio_buttons_fieldset :mentor_id, legend: { text: title, tag: 'h1', size: 'xl' } do %>
```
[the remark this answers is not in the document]
tonyheadford: Great spot, thanks!

PR #3160, opened 2023-03-23
Ensure school cohort combinations are unique 2

ltello:
### Context
This PR is an update of [this one](https://github.com/DFE-Digital/early-careers-framework/pull/2606) created by @peteryates a few months ago. All credits must go to him.

We had a case recently where a school had two 2021 cohort records. This should never happen. This change enforces it both at the database and model level.

* add a unique index to `school_cohorts` on `school_id` and `cohort_id`
* add a unique constraint to the `SchoolCohort` model

### Before merging

**Don't merge this until we have no more duplicates.**

This query should return no results in prod.

```
with dupes as (
  select school_id, cohort_id
  from school_cohorts
  group by school_id, cohort_id
  having count(*) > 1
)
select urn, name, postcode
from schools
join dupes on schools.id = dupes.school_id;
```


peteryates (approved):
Looks great!

PR #3161, opened 2023-03-23
Pass in `user_id` instead of `external_id` to change participant status

cwrw:
### Context
If a profile user is different than the participant identity changing training status is failing due to us switching services to use `user_id`

- Ticket: n/a

Switch forms to use `user_id` instead of `external_id`

### DIFF
--- a/app/forms/finance/ecf/change_training_status_form.rb
+++ b/app/forms/finance/ecf/change_training_status_form.rb
@@ -38,7 +38,7 @@ module Finance
         params = {
           cpd_lead_provider:,
           course_identifier:,
-          participant_id: participant_profile.participant_identity.external_identifier,
+          participant_id: participant_profile.participant_identity.user_id,
         }
 
         case training_status
--- a/app/forms/finance/npq/change_training_status_form.rb
+++ b/app/forms/finance/npq/change_training_status_form.rb
@@ -37,7 +37,7 @@ module Finance
         params = {
           cpd_lead_provider: participant_profile.npq_application.npq_lead_provider.cpd_lead_provider,
           course_identifier: participant_profile.npq_application.npq_course.identifier,
-          participant_id: participant_profile.participant_identity.external_identifier,
+          participant_id: participant_profile.participant_identity.user_id,
         }
 
         case training_status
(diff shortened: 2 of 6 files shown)


leandroalemao (commented):
Hi @cwrw Should we do similar in these as well?

[change_lead_provider_form.rb#L54](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/forms/finance/npq/change_lead_provider_form.rb#L54)
[ecf_induction_service.rb#L12](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/analytics/ecf_induction_service.rb#L12)
[participants.sql#L3](https://github.com/DFE-Digital/early-careers-framework/blob/main/analytics_queries/participants.sql#L3)
[induction_tutors.sql#L9](https://github.com/DFE-Digital/early-careers-framework/blob/main/analytics_queries/induction_tutors.sql#L9)

In another matter, I believe we don't need to fetch those anymore [participants_query.rb#L153](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/api/v1/ecf/participants_query.rb#L153) and [participants_query.rb#L157](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/api/v1/ecf/participants_query.rb#L157)

cwrw:
hello @leandroalemao, thanks for the review:
- [change_lead_provider_form.rb#L54](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/forms/finance/npq/change_lead_provider_form.rb#L54): needs an external id for the `Transfers::NPQParticipants`
- [ecf_induction_service.rb#L12](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/analytics/ecf_induction_service.rb#L12) and https://github.com/DFE-Digital/early-careers-framework/blob/main/analytics_queries/participants.sql#L3: we are sending both user id and external id and I think should continue to do so
- https://github.com/DFE-Digital/early-careers-framework/blob/main/analytics_queries/induction_tutors.sql#L9: I'll ask Pete tomorrow but I don't think we need to amend analytics queries as those are used outside the service 

I'll remove the external ids from the query

leandroalemao (approved):
lgtm 👍 😄

PR #3179, opened 2023-03-28
New primary eligible fields to npq application

shjohnson:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-860

This change is to include changes regarding the new tsf_primary_eligibility and tsf_primary_plus_eligibility outcomes for the application.


Lockyy (changes requested):
You'll need to add these fields to the allowed params for the API in app/controllers/api/v1/npq_profiles_controller.rb

Lockyy (approved):

peteryates (approved):

PR #3180, opened 2023-03-28
Fix Invalid pagination error

cwrw:
### Context

We accept pagination params in the API: `page[page]=1&page[per_page]100`
If an empty value is raised or an invalid value (a string "pageNum" as seen in this [error](https://dfe-teacher-services.sentry.io/issues/4010440797/?referrer=slack)) this is set to zero using `to_i` [here](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/controllers/concerns/api_pagination.rb#L31) and then passed on to set pagination which raises an error: 

```
expected :page >= 1; got 0
```

- Ticket: n/a

To resolve this I was in two minds, either we return page 1 if anyone sets the wrong values using: `[params.dig(:page, :page).to_i, 1].max` or catch this  error and return 422  with an appropriate message

Return bad request when wrong pagination param errors raised, error message has been approved by Hazel. This matches behaviour for updated_since param validation

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -13,6 +13,7 @@ module Api
     rescue_from ActiveRecord::RecordInvalid, with: :invalid_transition
     rescue_from Api::Errors::InvalidTransitionError, with: :invalid_transition
     rescue_from Api::Errors::InvalidDatetimeError, with: :invalid_updated_since_response
+    rescue_from Pagy::VariableError, with: :invalid_pagination_response
 
     def append_info_to_payload(payload)
       super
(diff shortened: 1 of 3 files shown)


app/controllers/api/api_controller.rb
```diff
@@ -49,5 +50,9 @@ def bad_request_response(exception)
     def invalid_transition(exception)
       render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:invalid_transition), params: exception).call }, status: :unprocessable_entity
     end
+
+    def invalid_pagination_response(_exception)
```
cwrw, line 54: this is also raised for `per_page` adapt message

PR #3181, opened 2023-03-29
Use relevant instead of latest induction record in admin interface

peteryates:
### Context

The participant's school tab in the admin interface was using the latest induction record rather than the relevant (active) one. This caused some inconsistencies and unexpected data being shown in the admin console.


leandroalemao:
@peteryates there're other mentions to `@latest_induction_record` .. not sure but should it be in the scope of this PR as well? tks

Screenshot 2023-03-29 at 12 25 17

peteryates:
Ah, well spotted @leandroalemao. I'll fix these too.

app/controllers/admin/participants/change_cohort_controller.rb
```diff
@@ -4,14 +4,14 @@ module Admin::Participants
   class ChangeCohortController < Admin::BaseController
     def edit
       @participant_profile = retrieve_participant_profile
-      @latest_induction_record = @participant_profile.latest_induction_record
+      @relevant_induction_record = Induction::FindBy.new(participant_profile: @participant_profile).call
```
tonyheadford, line 7: Fyi these were designed to be invoked without `.new` like `Induction::FindBy.call(participant_profile: @participant_profile)` but it will work either way

peteryates: Ah yes, fixed and force pushed. Thanks 👍🏽

PR #3185, opened 2023-03-29
[CPDLP-2102] Add  GET /api/v3/schools/ecf/:id endpoint

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2102

* Created new v3 endpoint for a single school.


cwrw (commented):
could you also enable/fix https://github.com/DFE-Digital/early-careers-framework/blob/main/spec/docs/v3/ecf_schools_spec.rb#L60?

PR #3189, opened 2023-03-29
Add a meaningful api updated_since filter parsing error message

leandroalemao:
### Context

- Ticket: N/A

Surface a meaningful error message when filter[:updated_since] has a unexpected format.

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -52,7 +52,7 @@ module Api
     end
 
     def invalid_updated_since_response(exception)
-      render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:invalid_updated_since_filter), params: exception).call }, status: :bad_request
+      render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:bad_request), params: exception.message).call }, status: :bad_request
     end
   end
 end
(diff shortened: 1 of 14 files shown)


app/controllers/api/api_controller.rb
```diff
@@ -49,5 +50,9 @@ def bad_request_response(exception)
     def invalid_transition(exception)
       render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:invalid_transition), params: exception).call }, status: :unprocessable_entity
     end
+
+    def invalid_updated_since_response(exception)
+      render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:invalid_updated_since_filter), params: exception).call }, status: :bad_request
```
cwrw, line 55: Maybe the title could remain `bad_request`, and the error message could be `invalid_updated_since_filter` instead, to be more consistent with the above errors?

spec/controllers/concerns/api_filter_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    it "returns formatted datetime" do
+      params = { filter: { updated_since: now.iso8601 } }
+      get("index", params:)
+      get_response = JSON.parse(response.body)
+
+      expect(get_response["updated_since_param"]).to eq("2023-03-29T10:10:00.000Z")
+    end
+
+    it "returns formatted datetime" do
```
cwrw, line 35: could we amend the description for each one as each test is different?

cwrw (commented):
looks great, thanks for this @leandroalemao, just two minor nits otherwise 👍

cwrw (approved):
thanks @leandroalemao for fixing this! also fixing the flakey test ⭐

PR #3191, opened 2023-03-29
[CPDLP-2120] Add filter by URN to GET schools api v3 endpoint

leandroalemao:
### Context

- Ticket: [CPDLP-2120](https://dfedigital.atlassian.net/browse/CPDLP-2120)

A user would like to filter the schools endpoint by a valid school urn.. so the API will return a response for the given URN (for the given cohort).

Add filter by URN to GET schools api v3 endpoint


swagger/v3/component_schemas/ECFSchoolsFilter.yml
```diff
@@ -2,6 +2,10 @@ description: "Filter schools to return more specific results"
 type: object
 properties:
   cohort:
-    description: "Return schools within the specified cohort"
+    description: "Return schools within the specified cohort(s). This is a comma delimited string of years."
```
cwrw, line 5: Ah we shouldn't allow multiple cohorts for schools, as each school will have different data per cohort (the in partnership field) so it only can be one

cwrw (commented):
a few comments but looking good otherwise

cwrw (approved):
thanks @leandroalemao 👍

PR #3192, opened 2023-03-30
Send impersonator back to starting point when done impersonating

laurentdfe:
### Current issue

When impersonating an induction tutor, clicking "stop impersonating" takes the impersonator back to the original home page rather than to the admin view of the impersonated user's school.

## Fix

When the admin user begins impersonating, we store the impersonation start location. When ending the impersonation, the admin user is sent back to the stored impersonation start location.

- Ticket: https://trello.com/c/r7QYqHjS/9-sit-impersonation


peteryates (approved):
Nice!

PR #3197, opened 2023-03-30
CreateInductionTutor => InductionTutors::Create

lazydevorg:
### Context

- Ticket: CST- 1294


leoapost:
The CreateInductionTutor class was deleted in [681a5dd](https://github.com/DFE-Digital/early-careers-framework/commit/681a5dd).

PR #3208, opened 2023-03-31
Show school's appropriate body in admin interface

peteryates:
### Context

The AB is now displayed per cohort on the cohorts index page.

image


hazelyule:
Looks good to me. Only comment is that where there is no AB, there should be copy to indicate one is not assigned. I'd recommend 'Not assigned'.

hazelyule (approved):

ltello (approved):

PR #3209, opened 2023-03-31
[CPDLP-2106] Improve NPQ participant seeds

leandroalemao:
### Context

- Ticket: [CPDLP-2106](https://dfedigital.atlassian.net/browse/CPDLP-2106)

We currently have two sets of seeds:

Legacy seeds: used in tests and used to populate review apps before being deprecated

New seeds: populate review apps

The new seeds do not have as much data as the old seeds, where we are missing some key data.

In order for Tom/Shahad to test single endpoints and then whole journeys, also for us to later prep some seed data for providers on sandbox, we need to ensure that our seed data is adequate.

"New seeds" checked for NPQ participants (npq application and profiles) and if they are adequate for testing our endpoints and journeys.

[CPDLP-2106]: https://dfedigital.atlassian.net/browse/CPDLP-2106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/db/new_seeds/base/add_npq_registrations.rb
+++ b/db/new_seeds/base/add_npq_registrations.rb
@@ -1,143 +1,107 @@
 # frozen_string_literal: true
 
 npq_lead_providers = NPQLeadProvider.all
-cohort_2022 = Cohort.find_by(start_year: 2022)
-cohort_2023 = Cohort.find_by(start_year: 2023)
+[Cohort.previous, Cohort.current, Cohort.next].map do |cohort|
+  # Create pending NPQ applications
+  seed_quantity(:npq_applications_pending).times do
+    NewSeeds::Scenarios::NPQ
+      .new(
+        lead_provider: npq_lead_providers.sample,
+        cohort:,
+      )
+      .build
+  end
 
-# Create pending NPQ applications
-seed_quantity(:npq_applications_pending).times do
-  NewSeeds::Scenarios::NPQ
-    .new(lead_provider: npq_lead_providers.sample)
-    .build
-end
+  # Create accepted NPQ applications with participant profiles
+  # and a declaration
+  seed_quantity(:npq_application_with_declarations).times do
+    NewSeeds::Scenarios::NPQ
+      .new(
+        lead_provider: npq_lead_providers.sample,
+        cohort:,
+      )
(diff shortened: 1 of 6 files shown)


db/new_seeds/base/add_npq_registrations.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 npq_lead_providers = NPQLeadProvider.all
+cohort_2022 = Cohort.find_by(start_year: 2022)
```
cwrw, line 4: maybe we can loop over current and next cohort (to avoid hardcoding years) and create the same npq applications for both?

db/new_seeds/base/add_npq_registrations.rb
```diff
@@ -1,56 +1,114 @@
...
+        cohort:,
+      )
+      .build
+      .accept_application
+      .add_declaration
+      .add_statement_line_items
+  end
+
+  # Create NPQEhco applications with participant profile and declaration
+  seed_quantity(:npq_applications_eligible_for_transfer).times do
```
cwrw, line 91: do we need this as we have already set up EHCO applications

db/new_seeds/base/add_npq_registrations.rb
```diff
@@ -1,56 +1,114 @@
...
+    NewSeeds::Scenarios::NPQ
+      .new(
+        lead_provider: npq_lead_providers.sample,
+        npq_course: NPQCourse.find_by(identifier: Finance::Schedule::NPQLeadership::IDENTIFIERS.sample),
+        cohort:,
+      )
+      .build
+      .accept_application
+      .add_declaration
```
cwrw, line 72: Out of curiosity, what is the specific scenario required for having an application with a declaration and statement line item? As I was hoping that new applications that come in, we could make declarations to etc.

cwrw (commented):
heya thanks for addressing round 1 of comments :D I had a few more comments, but might be worth us jumping on a quick call to discuss a few things

cwrw (approved):
🙌

PR #3211, opened 2023-04-03
Use default NPQ schedule helper in specs

cwrw:
### Context
There are failures on main due to the switch in default NPQ schedules. The failures are because we are using default schedule lookups rather than the new `schedule_for` method. Also we are creating all schedules across multiple cohorts

- Ticket: 

- With default schedules creates schedules across cohorts
- Add helper in tests to avoid creating schedules in the setup
- Remove use of default schedule for schedule for to avoid future failures

### Guidance to review
I will spend more time with this after to try to ensure we cleanup all tests so this doesn't happen again

### DIFF
--- a/spec/features/finance/change_lead_provider_approval_status_spec.rb
+++ b/spec/features/finance/change_lead_provider_approval_status_spec.rb
@@ -2,9 +2,7 @@
 
 require "rails_helper"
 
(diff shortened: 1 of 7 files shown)


peteryates (approved):
Nice.

PR #3212, opened 2023-04-03
[WIP ] Spike Part 1 - Schools record

cwrw:
### Context

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2123
slack thread: https://ukgovernmentdfe.slack.com/archives/C01J1J4FM50/p1679580199266459

We'd like to investigate adding new concepts to participant profiles to reduce the conflation of data in induction records.
We discussed adding a school "tenureship", as well as investigate having transfer records.

In this PR I've created school records, which hold data involving a participant's time in a school, with joining and leaving dates.

- add new school record
- suggest fields and associations
- example of how this can work with enrolments

### Guidance to review
This is an investigation, so please feel free to break down all the solutions proposed

This solution feels like it has similar problems to induction records. This will require some work for us to integrate into our API as well.

There is also the overhead of having to backfill any old data, and what the new induction record world will look like

### DIFF
--- a/app/models/school_record.rb
+++ b/app/models/school_record.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class SchoolRecord < ApplicationRecord
+  has_paper_trail
+
+  # Associations
+  belongs_to :school
+  belongs_to :participant_profile, class_name: "ParticipantProfile::ECF", touch: true
+  belongs_to :joining_induction_record, class_name: "InductionRecord"
+  belongs_to :leaving_induction_record, class_name: "InductionRecord", optional: true
+
+  # Validations
+  validates :joining_date, presence: true
+end
--- a/app/services/induction/transfer_and_continue_existing_fip.rb
+++ b/app/services/induction/transfer_and_continue_existing_fip.rb
@@ -25,6 +25,13 @@ class Induction::TransferAndContinueExistingFip < BaseService
                                                preferred_email: email,
                                                mentor_profile:,
                                                school_transfer: true)
+      ## Update old School record and create new one for new school:
+      ## SchoolRecord::Enrol.new(
+      ##  school: school_cohort.school,
+      ##  participant_profile:,
+      ##  joining_date: start_date,
+      ##  leaving_date: end_date,
+      # #).call
 
       if participant_profile.mentor?
         Mentors::ChangeSchool.call(mentor_profile: participant_profile,
--- a/app/services/school_record/enrol.rb
+++ b/app/services/school_record/enrol.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+class SchoolRecord::Enrol < BaseService
+  def call
+    ActiveRecord::Base.transaction do
+      if leaving_date
+        # update previous school record. Might need to pass school from/to
+        previous_school_record.update(leaving_date:)
+      end
+
+      participant_profile.school_records.create!(
+        school:,
+        joining_date:,
+      )
+    end
+  end
+
+private
+
+  attr_reader :participant_profile, :school, :joining_date, :leaving_date
+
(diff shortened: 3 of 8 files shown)


app/models/school_record.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class SchoolRecord < ApplicationRecord
+  has_paper_trail
+
+  # Associations
+  belongs_to :school
+  belongs_to :participant_profile, class_name: "ParticipantProfile::ECF", touch: true
+  belongs_to :joining_induction_record, class_name: "InductionRecord"
```
cwrw, line 9: Using induction records might be overkill and not required. As it adds extra overhead around linking and maintaining those

app/services/induction/transfer_and_continue_existing_fip.rb
```diff
@@ -25,6 +25,13 @@ def call
                                                preferred_email: email,
                                                mentor_profile:,
                                                school_transfer: true)
+      ## Update old School record and create new one for new school:
+      ## SchoolRecord::Enrol.new(
+      ##  school: school_cohort.school,
+      ##  participant_profile:,
+      ##  joining_date: start_date,
```
cwrw, line 32: Maybe we can stop sending start/end dates to induction records and allow them to be set sequentially according to current date/previous end date

app/services/school_record/enrol.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize(participant_profile:, school:, joining_date:, leaving_date: nil)
+    @participant_profile = participant_profile
+    @school = school
+    @joining_date = joining_date
+    @leaving_date = leaving_date
+  end
+
+  def previous_school_record
+    # participant_profile.school_records.latest
```
cwrw, line 30: we might need more logic around determining latest school record: 
no leaving date, or pass in previous school etc. as well as look at timestamp

db/migrate/20230402191059_create_school_record.rb
```diff
@@ -0,0 +1,16 @@
...
+  def change
+    create_table :school_records do |t|
+      t.datetime :joining_date, null: false
+      t.datetime :leaving_date, null: true
+
+      t.references :school, null: false, foreign_key: true, type: :uuid
+      t.references :participant_profile, null: false, foreign_key: true, type: :uuid
+      t.references :joining_induction_record, null: false, foreign_key: { to_table: :induction_records }, type: :uuid
+      t.references :leaving_induction_record, null: true, foreign_key: { to_table: :induction_records }, type: :uuid
+      t.timestamps
```
cwrw, line 13: do we also want to add cohort?

app/services/school_record/enrol.rb
cwrw: from @peteryates: For mentors, this is more complex as a participant can belong to many schools at once, how do we find the correct school record?

app/models/school_record.rb
mattonfoot: The opportunity here is that it can include a status specific to a school. If a participant moves through several schools you will know the status of the participant for each school that was responsible for their induction and when

app/services/school_record/enrol.rb
mattonfoot: We should not mix up mentors training and mentors mentoring. They should be two separate things.

leoapost: > from @peteryates: For mentors, this is more complex as a participant can belong to many schools at once, how do we find the correct school record?

Mentors receive training only from one school, so we should be fine.

They can also mentor ECTs in multiple schools via the school_mentor association, but like Matt mentioned above, it's a separate thing.

peteryates: I didn't even realise that's what the `schools_mentors` table meant! 🤦🏽

PR #3219, opened 2023-04-03
[WIP ] Spike Part 2 - Schools Transfer

cwrw:
### Context

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2123
slack thread: https://ukgovernmentdfe.slack.com/archives/C01J1J4FM50/p1679580199266459

We'd like to investigate adding new concepts to participant profiles to reduce the conflation of data in induction records.
We discussed adding a school "tenureship", as well as investigate having transfer records.

In this PR I've created school transfers, which only holds a transfer's data. PR for school tenureship can be found [here](https://github.com/DFE-Digital/early-careers-framework/pull/3212)

- add new school transfer record
- suggest fields and associations
- example of how this can work with transfers

### Guidance to review
This is an investigation, so please feel free to break down all the solutions proposed

Im not sure how this solution works with solving induction record problems, but it does resolve our API issues with transfers

There is also the overhead of having to backfill any old data, and what the new induction record world will look like

### DIFF
--- a/app/models/participant_profile.rb
+++ b/app/models/participant_profile.rb
@@ -48,6 +48,7 @@ class ParticipantProfile < ApplicationRecord
     merge(ParticipantProfileState.most_recent)
   }, class_name: "ParticipantProfileState"
   has_one :user, through: :teacher_profile
+  has_many :school_transfers
 
   # Scopes
   scope :ecf, -> { where(type: [ECT.name, Mentor.name]) }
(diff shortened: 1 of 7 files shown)


app/models/participant_profile.rb
mattonfoot, line 1: It feels like school transfers are not a thing really. There are multiple timelines at two schools which overlap which can be interpreted as a transfer.

cwrw: yes, its mostly a concept we will be surfacing in the API, and this would really help us with our calculations. Or we could infer this from induction records, which will be quite hard to work around. From our chat in slack is this what you had in mind or something different?

PR #3223, opened 2023-04-04
Use request.headers['HTTP_REFERER'] instead of request.referer

laurentdfe:
### Context

- Ticket: https://trello.com/b/va81LwGU/cpd-admin-interface

From my research, I was unable to find `request.referer` anywhere in the Rails documentation, so this PR replaces it with the surer and [documented](https://api.rubyonrails.org/classes/ActionDispatch/Request.html#method-i-headers) `request.headers['HTTP_REFERER']` instead of `request.referer`.

### DIFF
--- a/app/controllers/admin/impersonates_controller.rb
+++ b/app/controllers/admin/impersonates_controller.rb
@@ -11,6 +11,16 @@ module Admin
 
     def create
       impersonate_user(@user)
+
+      Sentry.with_scope do |scope|
+        scope.set_tags(
+          request_dot_referer: request.referer,
+          request_headers_http_referer: request.headers["HTTP_REFERER"],
+          request_headers_referer: request.headers["Referer"],
+        )
+        Sentry.capture_message("User was impersonated")
+      end
+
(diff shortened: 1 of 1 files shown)


app/controllers/admin/impersonates_controller.rb
```diff
@@ -29,19 +39,23 @@ def load_user
...
+        redirect_to URI(referer).path
       end
     end
 
     def pundit_user
       true_user
     end
+
+    def referer
+      request.headers["HTTP_REFERER"]
```
peteryates, line 58: I'm happy to give this a go but [it looks like it's doing the same as `request.referer`](https://github.com/rack/rack/blob/f5666bc8cb13b8d731ea0222fbd3ada670f2cd55/lib/rack/request.rb#L204).

```ruby
      def referer;         get_header('HTTP_REFERER')                     end
```

laurentdfe: yeah, agreed, I'll close this PR

PR #3226, opened 2023-04-05
Allow bin/dev to run without jobs

peteryates:
### Context

Running bin/dev is really noisy, mainly with failed email sending and background data sync. I don't usually want it.

With this change, passing in a `-j` argument to bin/dev (`bin/dev -j`) now only runs the web, css and js portions of Procfile.dev, which allows for a noise free experience when jobs aren't needed.

Tested on Linux. I assume this will work on macOS, definitely needs testing before merge!


peteryates:
Perhaps this should be the other way round and off by default, enabled via `-j`?

leandroalemao (approved):
Seems to be working on mac as well.. 👍 😄

leoapost (approved):

PR #3230, opened 2023-04-06
[CPDLP-2105] Add PUT /api/v3/partnerships/ecf/{partnership_id}

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2105

* New update partnership endpoint
* New service class `Partnerships::Update`

### DIFF
--- a/app/services/partnerships/create.rb
+++ b/app/services/partnerships/create.rb
@@ -69,8 +69,10 @@ module Partnerships
       return if delivery_partner_id.blank?
 
       if delivery_partner
-        unless delivery_partner_has_provider_relationship?
+        if !delivery_partner_has_provider_relationship?
           errors.add(:delivery_partner_id, :no_relationship)
+        elsif errors.empty? && duplicate_delivery_partner_request?
+          errors.add(:delivery_partner_id, :duplicate_delivery_partner)
         end
       else
         errors.add(:delivery_partner_id, :invalid)
@@ -99,6 +101,10 @@ module Partnerships
       return false if previous_year_lead_provider.blank?
 
       previous_year_lead_provider != lead_provider && school.school_cohorts.find_by(cohort: cohort_record).blank?
+    end
+
+    def duplicate_delivery_partner_request?
+      Partnership.where(school:, cohort: cohort_record, lead_provider:, delivery_partner:).exists?
     end
   end
 end
--- a/app/services/partnerships/update.rb
+++ b/app/services/partnerships/update.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+module Partnerships
+  class Update
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :partnership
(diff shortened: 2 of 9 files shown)


cwrw (commented):
looking good, just a few comments

app/services/partnerships/update.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def validate_delivery_partner
+      if delivery_partner
+        if !delivery_partner_has_provider_relationship?
+          errors.add(:delivery_partner_id, :no_relationship)
+        elsif partnership.challenged?
+          errors.add(:delivery_partner_id, :partnership_challenged)
+        end
+      else
+        errors.add(:delivery_partner_id, :invalid)
```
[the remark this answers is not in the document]
mooktakim: Updated 👍

app/services/partnerships/update.rb
```diff
@@ -0,0 +1,50 @@
...
+    attribute :partnership
+    attribute :delivery_partner_id
+
+    validates :partnership, :delivery_partner_id, presence: true
+    validate :validate_delivery_partner
+
+    def call
+      return false if invalid?
+
+      ::Partnerships::Report.call(
```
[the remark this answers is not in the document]
mooktakim: Updated 👍

app/services/partnerships/update.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    attribute :partnership
+    attribute :delivery_partner_id
+
+    validates :partnership, :delivery_partner_id, presence: true
+    validate :validate_delivery_partner
+
+    def call
+      return false if invalid?
+      return true if partnership.delivery_partner == delivery_partner
```
cwrw, line 16: I think we'll need to return the partnership here otherwise it breaks, could you add a test as well?

mooktakim: Added test

mooktakim: Added test

cwrw (approved):
🙌

app/services/partnerships/create.rb
```diff
@@ -100,5 +102,9 @@ def has_previous_cohort_and_matching_lead_provider?
 
       previous_year_lead_provider != lead_provider && school.school_cohorts.find_by(cohort: cohort_record).blank?
     end
+
+    def duplicate_delivery_partner_request?
```
cwrw, line 106: thanks for adding this! only comment is we will get duplicate errors with `recruited_by_other_provider` if there is another existing partnership, maybe we can only return this if the other one fails?

cwrw (approved):
thanks for this, and apologies for the multiple iterations, just one final comment otherwise 👍

app/services/partnerships/create.rb
mooktakim: Sure, added a check to see if we already have an error.

PR #3233, opened 2023-04-07
Super users

peteryates:
### Context

It's likely that in the upcoming registration window we'll need to add some _dangerous_ operations that only a subset of administrators can access. This access will be controlled using the `super_user` flag on `AdminProfile`.

- Add super_user flag to admin_profiles
- Add the superuser link to the nav
- Add super users policy and basic controller

### DIFF
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -63,5 +63,10 @@ private
     !!user&.admin?
   end
 
+  def super_user_only
+    admin_only && user.admin_profile.super_user?
+  end
+
   alias_method :admin?, :admin_only
+  alias_method :super_user?, :super_user_only
 end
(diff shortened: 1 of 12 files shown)


app/policies/application_policy.rb
```diff
@@ -63,5 +63,10 @@ def admin_only
     !!user&.admin?
   end
 
+  def super_user_only
+    admin_only && user.admin_profile.super_user?
```
leoapost, line 67: I thought you had a bug here because the `admin_only` suggests the user should have only one profile, but instead it's checking if the user is also an admin.

I think it could be simplified to
```suggestion
    !!user.admin_profile&.super_user?
```

leoapost (commented):
LGTM. I've left a small suggestion, but it's nothing huge.

leoapost (approved):

app/policies/application_policy.rb
peteryates: I think I did it the way I did because it makes it clear it's like the admin role **plus** another level (super user).

cwrw (approved):

PR #3254, opened 2023-04-12
Update partnership API spec

cwrw:
### Context

@hazelyule has pointed out some incorrect API responses for POST and PUT partnership endpoints

- Ticket: n/a

- Add manual examples to the v3 API spec for POST and PUT partnership endpoints which are not challenged to avoid confusion 
- Amend partnership response type to match API spec

### Guidance to review
Did I miss anything?


hazelyule:
Check PUT and POST response bodies and both appear active. So looks good!

mooktakim (approved):

PR #3259, opened 2023-04-13
[CPDLP-2132] Add uplift clawbacks in adjustments totals on financial statements page

leandroalemao:
### Context

- Ticket: [CPDLP-2132](https://dfedigital.atlassian.net/browse/CPDLP-2132)

Assurance made Digital aware that the adjustment section in the statements was not adding up the figures between uplift clawback and clawback adjustments in the adjustments table for ECF financial statements.

As this information is processed and passed to providers, we must ensure that our data is accurate.

When the statement has both uplift clawback and clawback from a declaration.. they should be aggregated to the total value in the adjustment table and in the summary table under the adjustments line item.

[CPDLP-2132]: https://dfedigital.atlassian.net/browse/CPDLP-2132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/finance/ecf/statement_calculator.rb
+++ b/app/services/finance/ecf/statement_calculator.rb
@@ -131,8 +131,12 @@ module Finance
         [available, delta_uplift_amount].min
       end
 
+      def uplift_clawback_deductions
+        uplift_deductions_count * -uplift_fee_per_declaration
+      end
+
       def adjustments_total
-        -clawback_deductions
+        -clawback_deductions + uplift_clawback_deductions
       end
 
       def clawback_deductions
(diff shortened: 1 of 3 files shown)


app/services/finance/ecf/statement_calculator.rb
```diff
@@ -131,8 +131,12 @@ def total_for_uplift
         [available, delta_uplift_amount].min
       end
 
+      def uplift_clawback_deductions
```
cwrw, line 134: suggestion: maybe we can use this method also here: https://github.com/DFE-Digital/early-careers-framework/blob/main/app/views/finance/ecf/statements/show.html.erb#L200

cwrw (approved):

app/services/finance/ecf/statement_calculator.rb
leandroalemao: Hi @cwrw spot on.. tks 👍 😄

cwrw (approved):
🙌

PR #3264, opened 2023-04-17
Reorganise the api docs

hazelyule:
### Context

Complete content rewrite of ECF api guidance. This PR is review content in review app. 

### To do

* [x] Add links 
* [x] Review content and design elements
* [x] Add additional inset text 
* [ ] Add accordion html
* [ ] Sign off schedule dates


peteryates:
Some notes for whoever looks into making this work with [govuk_markdown](/dfe-digital/govuk_markdown).

govuk_markdown is a gem that wraps RedCarpet and injects govuk specific styling and lets us use some simple text formatting for inset text, details boxes, etc.

In `middleman-core` there's a set of extension registrations like this:

```ruby
# Markdown Support
Middleman::Extensions.register :markdown_renderer, auto_activate: :before_configuration do
  require 'middleman-core/renderers/markdown'
  Middleman::Renderers::Markdown
end
```

I think we'd need to register a new one for `:govuk_markdown_renderer` and pass to it our own renderer class (see `lib/middleman-core/renderers/markdown.rb`), and hand that off to Tilt. There's a chance we'd need a custom Tilt template too.

We're probably best just getting the content in to start with and then worrying about formatting/display later. Do we even need Middleman here? Rails can render Markdown just as easily and it will significantly simplify our setup.

hazelyule:
@peteryates Just saved a commit with the release note. Once checks are complete, will try to merge!

PR #3265, opened 2023-04-17
[CST-1368] Change ECTs induction start date and cohort when DQT record updated

lazydevorg:
### Context

- Ticket: CST-1368

* Service `Participants::SyncDqtInductionStartDate` added and called from `Participants::ParticipantValidationForm`
* Model SyncDqtInductionStartDateError used to take track of participant who can't be processed. The error messages are saved in the DB for each participant.


db/migrate/20230424093735_create_sync_dqt_induction_start_date_error.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class CreateSyncDqtInductionStartDateError < ActiveRecord::Migration[6.1]
+  def change
+    create_table :sync_dqt_induction_start_date_errors, id: false do |t|
+      t.references :participant_profile, null: false, primary_key: true, foreign_key: true, type: :uuid, index: { name: "dqt_sync_participant_profile_id" }
+      t.text :error_message
+      t.timestamps
+    end
```
ltello, line 9: I would prefer that a participant profile can have_many :sync_dqt_induction_start_date_errors: 
each induction record of the participant can have a different issue when amending their cohort (i.e target cohort not setup in our service, target school_cohort not setup by the school, target school_cohort with no default induction programme...). 

If we have each of these errors in their own record, then we can later on query the db for specific ones (give me all errors where school_cohort has not been setup by the school, for instance) and do a comms campaign to all the affected schools.

Your implementation joins all the errors of a participant in one huge string which would make it more difficult to query for the purposes above.

app/services/participants/sync_dqt_induction_start_date.rb
```diff
@@ -0,0 +1,60 @@
...
+      return false unless FeatureFlag.active?(:cohortless_dashboard)
+      return false if @dqt_induction_start_date.nil?
+      return false if @participant_profile.induction_start_date.present?
+
+      current_induction_record = @participant_profile.induction_records.latest
+      participant_cohort = current_induction_record.cohort
+      dqt_cohort = Cohort.containing_date(@dqt_induction_start_date)
+      return false if dqt_cohort.nil?
+
+      if dqt_cohort == participant_cohort
```
ltello, line 20: Same as above. We should still call amend_cohort in this situation because even if the latest induction record is in the right cohort, some of the old induction records might still be in the wrong cohort and need to get it amended.

app/services/participants/sync_dqt_induction_start_date.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    def call
+      return false unless FeatureFlag.active?(:cohortless_dashboard)
+      return false if @dqt_induction_start_date.nil?
+      return false if @participant_profile.induction_start_date.present?
+
+      current_induction_record = @participant_profile.induction_records.latest
+      participant_cohort = current_induction_record.cohort
+      dqt_cohort = Cohort.containing_date(@dqt_induction_start_date)
+      return false if dqt_cohort.nil?
```
ltello, line 18: I think that in this case we should also save an error for the participant to track that the target cohort for this participant has not been setup yet in our service. 

This might help to warn us that we need to setup missing cohorts.

app/services/participants/sync_dqt_induction_start_date.rb
```diff
@@ -0,0 +1,60 @@
...
+          end
+        end
+      end
+    end
+
+  private
+
+    def save_error_message(amend_cohort)
+      error = SyncDqtInductionStartDateError.find_or_create_by(participant_profile: @participant_profile) # rubocop:disable Rails/SaveBang
+      error.error_message = amend_cohort.errors.full_messages.join(", ")
```
ltello, line 42: Please, see my comment below about each error recorded in its own entry.

ltello (changes requested):

db/migrate/20230424093735_create_sync_dqt_induction_start_date_error.rb
lazydevorg: @ltello it seems a bit overkilled as we want to save the errors from the latest run of the job.
I think the point is just to have a list of participants for which the job failed and that needs attention.
I am not sure we need an error record for each error message.

app/services/participants/sync_dqt_induction_start_date.rb
lazydevorg: if we do this and, as per [your comment above](https://github.com/DFE-Digital/early-careers-framework/pull/3265/files/44742161d9e37a4d0448ee836dfd61578b4df075#r1176695938), then this job will run all the time for each participant.
I think the ask was to do it just for the participants who did not have the induction start date set.
I'll double check whether this is still the case.

ltello: Yes, exactly, because:
- we already have participants with wrong cohorts in their historical induction records and this is the tool to fix them, no matter their existing induction start date. 
- their existing induction date might not match the one in their DQT record and better to have the right one coming from DQT.
- we want to run this tool for ALL participants only one time to update their induction start date

ltello: Once this one-off task is done, then your tool will only run for each new participant and for re-validations in support

db/migrate/20230424093735_create_sync_dqt_induction_start_date_error.rb
ltello: The idea is to be able to query the table to tackle all the participants/schools with issue A, all the participants/schools with issue B, etc... and do a different action onto each group

ltello (approved):

peteryates:
@lazydevorg - this looks good to me, will approve once the conflict is resolved 👌🏽

PR #3268, opened 2023-04-17
allow name and email change through the npq application tab

shjohnson:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-789

Allow name and email change to non-GAI users through the application tab

### DIFF
--- a/app/controllers/admin/npq/applications/change_name_controller.rb
+++ b/app/controllers/admin/npq/applications/change_name_controller.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Admin::NPQ::Applications
+  class ChangeNameController < Admin::BaseController
+    before_action :load_npq_application
+    before_action :check_gai_status
+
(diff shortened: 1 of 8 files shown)


Lockyy (approved):
Looks good. Do we not want this backfilling into NPQ? Or frontfilling there and propagating down into ECF like how GAI does its updates?

app/controllers/admin/npq/applications/change_name_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+  private
+
+    def load_npq_application
+      authorize NPQApplication
+
+      @npq_application = NPQApplication
+        .eager_load(:profile, participant_identity: :user)
+        .find(params[:id])
```
Lockyy, line 28: We should probably verify the user doesn't have a GAI ID here

Lockyy (approved):

PR #3271, opened 2023-04-18
Add new script to create ECF cohort

cwrw:
The script receives all csvs for other scripts and creates a standalone ECF cohort

### Context

To support cohort 2023 for ECF, records relating to the cohort need to be populated. This includes ECF contracts, schedules, statements and cohort.

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2139

 - Add new script that consumes CSVs and runs all required services to create an ECF cohort
 - Update Cohort script to create or update existing cohorts, as we need to update fields on existing cohort 2023 in prod

### Guidance to review
Did I miss anything?

### DIFF
--- a/app/services/importers/create_new_ecf_cohort.rb
+++ b/app/services/importers/create_new_ecf_cohort.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module Importers
+  class CreateNewECFCohort
+    def call
+      logger.info "CreateNewECFCohort: Started!"
+      check_csvs!
+
+      run_ecf_cohort_scripts
+      logger.info "CreateNewECFCohort: Finished!"
+    end
+
+  private
+
+    attr_reader :cohort_csv, :cohort_lead_provider_csv, :contract_csv, :schedule_csv, :statement_csv, :logger
+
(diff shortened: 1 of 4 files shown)


app/services/importers/create_new_ecf_cohort.rb
```diff
@@ -0,0 +1,49 @@
...
+      @logger = logger
+    end
+
+    def check_csvs!
+      return if cohort_csv.present? && cohort_lead_provider_csv.present? && contract_csv.present? && schedule_csv.present? && statement_csv.present?
+
+      raise "All scripts need to be present to create a new ECF cohort"
+    end
+
+    def run_ecf_cohort_scripts
```
leandroalemao, line 32: @cwrw Do you think we need some sort of rollback if any of the steps fails mid-process for some reason?

mooktakim (approved):

app/services/importers/create_new_ecf_cohort.rb
cwrw: @leandroalemao good question, each script we call is wrapped within a transaction block, so anything that isn't valid will rollback. We will also test the data before running in prod so there are no surprises

PR #3278, opened 2023-04-19
Remove clawed back from uniq declaration index

cwrw:
### Context
Whilst attempting to mark statements as paid, I got the following error:
```
ERROR:  duplicate key value violates unique constraint "unique_declaration_index" (ActiveRecord::RecordNotUnique)
DETAIL:  Key (cpd_lead_provider_id, participant_profile_id, declaration_type, course_identifier, state)=(xxxx, xxxx, started, ecf-mentor, clawed_back) already exists.
```
A unique index was added to prevent double posting, but we should allow the a declaration belonging to the same profile/lead provider/state be clawed back more than once if necessary

- Ticket: n/a

Remove clawed back from the unique index but keep the rest the same

### Guidance to review
Did I miss anything?


leandroalemao (approved):
lgtm 👍 😄

peteryates (approved):

PR #3282, opened 2023-04-19
Add reminder for updating induction tutor

peteryates:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CST-1422

We want to send reminders to GIAS contacts at schools so they can update their induction tutor if required. This change adds a service class that allows us to perform that using the console.

Suggested way to use it is:

```ruby
schools = School.where(urn: SOME_LIST_OF_URNS_FROM_SPREADSHEET)

schools.each { |s| UpdateInductionTutorReminder.new(s).send! }
```

- Add method to trigger induction tutor update email
- Add service class for sending update reminders


tonyheadford (approved):
LGTM!

PR #3285, opened 2023-04-20
Limit providers from adding partnerships unless cohort registration open

cwrw:
### Context

If the next cohort registration isn't open do not surface ability to create partnerships

- Ticket: n/a

Only show cohorts up to the Cohort open for registration

### DIFF
--- a/app/views/dashboard/show.html.erb
+++ b/app/views/dashboard/show.html.erb
@@ -3,7 +3,7 @@
     <% if current_user.lead_provider? %>
       <%= render partial: "navigation" %>
       <h1 class="govuk-heading-xl"><%= @lead_provider.name %></h1>
-      <% @lead_provider.cohorts.each do |cohort| %>
+      <% @lead_provider.cohorts.where(start_year: ..Cohort.active_registration_cohort.start_year).each do |cohort| %>
         <h2 class="govuk-heading-m">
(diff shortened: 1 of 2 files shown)


app/views/dashboard/show.html.erb
```diff
@@ -3,7 +3,7 @@
     <% if current_user.lead_provider? %>
       <%= render partial: "navigation" %>
       <h1 class="govuk-heading-xl"><%= @lead_provider.name %></h1>
-      <% @lead_provider.cohorts.each do |cohort| %>
+      <% @lead_provider.cohorts.where(start_year: ..Cohort.active_registration_cohort.start_year).each do |cohort| %>
```
leandroalemao, line 6: @cwrw Should we add a validation in the BE to prevent a partnership to be created for a cohort not yet opened for registrations?

leandroalemao (approved):
just a small comment.. more like a question really.. lgtm 👍  tkss 😄

app/views/dashboard/show.html.erb
cwrw: chatted in person: excellent question, agreed that because v3 isn't live the risk of anyone adding anything is low, we'll tackle later esp when v3 is live and we have clarity on the rules

PR #3313, opened 2023-04-25
Display the profile identity in the Admin pages

leoapost:
### Context

- Ticket: N/A

The identity linked to the participant's profile is important as we expose participants to Lead Providers base on that. Problem is that we don't display it anywhere in the Admin pages, making it hard to troubleshoot.

- Display the profile's identity to super users only, to avoid confusing the admins with too many emails
- Rename the `Email address` label to `User email address` to be clearer where it's coming from

### Guidance to review

| Before | After |
|--------|--------|
| image| image|

### DIFF
--- a/app/views/admin/participants/details/_show_ecf.html.erb
+++ b/app/views/admin/participants/details/_show_ecf.html.erb
@@ -21,6 +21,13 @@
     end
   end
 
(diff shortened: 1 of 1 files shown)


tonyheadford (approved):
Think there might still be scope for ambiguity but I don't think it's a blocker

PR #3315, opened 2023-04-25
amend cohort also in historical induction records

ltello:
### Context

The current service `Induction::AmendParticipantCohort` to move a participant to one cohort to another on admin only changes the `current_induction_record` of the participant. 

- Ticket: 

- Amend not only the latest induction record but all historical records of the participant, so all of them must be consistently in the same right cohort.


peteryates (approved):
Seems sensible.

PR #3316, opened 2023-04-26
Revert "Add Zendesk help widget"

peteryates:
Reverts DFE-Digital/early-careers-framework#3309

We don't need this any more as the original support email address has been restored.


cwrw (approved):
😢

PR #3320, opened 2023-04-26
Change preferred identity

leoapost:
### Context

- Ticket: N/A

This PR adds a new Admin page for super users only, to change the preferred identity of induction records.

- Display a link in each Induction Record to change the preferred email
- Add a page to select the new email. All the emails listed are taken from the user's participant identities

### Guidance to review
1. Visit a ECF participant as a super user
2. Go to Induction Records tab
3. Click the `Change` link in the preferred email row
4. Confirm all the email in the dropdown emails are from the User's participant identities
5. Pick a new email and click `Submit`
6. Confirm the Induction Record's new preferred email

| Screenshots  |
|--------|
| image |
| image |

### DIFF
--- a/app/controllers/admin/participants/induction_records_controller.rb
+++ b/app/controllers/admin/participants/induction_records_controller.rb
@@ -4,6 +4,8 @@ module Admin::Participants
   class InductionRecordsController < Admin::BaseController
     include RetrieveProfile
 
+    before_action :load_induction_record, only: %i[edit_preferred_email update_preferred_email]
+
     def show
       @participant_presenter = Admin::ParticipantPresenter.new(@participant_profile)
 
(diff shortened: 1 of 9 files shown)


ltello:
Nice one @leoapost. What if the user wants a brand new email?

ltello (approved):

app/controllers/admin/participants/induction_records_controller.rb
```diff
@@ -13,10 +16,35 @@ def show
...
+    def edit_preferred_email
+      authorize @induction_record
+
+      @participant_identities = @participant_profile.user.participant_identities
+    end
+
+    def update_preferred_email
+      authorize @induction_record
+
+      if @induction_record.update(induction_params)
```
cwrw, line 28: could we instead use the `Induction::ChangeInductionRecord` service to ensure this change creates a new induction record and the previous one is marked as changed??

leoapost: @cwrw any particular reason we would need that?

leoapost:
> Nice one @leoapost. What if the user wants a brand new email?

@ltello at the moment we add new emails manually. However, we could add an `New identity` action in the Identities tab in the Participant's admin page.

The scope of this PR is only to update the Induction records's preferred email.

app/controllers/admin/participants/induction_records_controller.rb
cwrw: I think usually induction records are audits to show the timeline of a profile, and if a status/school/provider/mentor id/schedule changes we create a new record and mark the previous as "changed". This also ensures the latest induction record is amended and not previous ones which might be out of date. Let me know your thoughts as to why not to do this

leoapost: @cwrw oh right. I don't mind creating a new IR, but we already amend IRs directly (i.e. when we defer/withdraw a training) or when we fix IRs created with the wrong preferred email or when we assign mentors to participants. 

So, unless the IR's preferred email should be protected due to business rules I'm unaware, then I think it's ok to amend the IR.

PR #3322, opened 2023-04-27
[CPDLP-2127] Add transfers endpoint

cwrw:
### Context

Add transfers endpoint to API v3

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2127

- Add transfers query, this looks for either leaving induction record for the provider OR school transfer for provider, and then ensures that both include an agnostic leaving induction record to ensure we don't have school transfers without a leaving induction record.
- Add new seeds to help with query transfer tests
- Add new transfers serialiser and build transfers service. The transfer service traverses the leaving induction records, and then picks the next joining induction inline, otherwise the transfer is an incomplete transfer. Complexity is high but I couldn't find a better way to do this
- Add endpoint and tests

### Guidance to review
There are some odd shapes in prod which will return empty transfers. This is the best shape I could get this in. I am making a list of those to see if we can fix them before release.
This includes school transfers without actually changing schools, and incorrect start/end dates


mooktakim (approved):

app/controllers/api/v3/ecf/transfers_controller.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+module Api
+  module V3
+    module ECF
+      class TransfersController < Api::ApiController
+        include ApiTokenAuthenticatable
+        include ApiPagination
+        include ApiFilter
+        include ApiOrderable
```
cwrw, line 10: remove not used

app/services/api/v3/ecf/build_transfers.rb
```diff
@@ -0,0 +1,86 @@
...
+
+          build_transfers
+        end
+
+      private
+
+        attr_reader :cpd_lead_provider, :induction_records
+
+        def build_transfers
+          sorted_induction_records.sort_by(&:created_at).each do |induction_record|
```
cwrw, line 28:
```suggestion
          sorted_induction_records.each do |induction_record|
```

app/services/api/v3/ecf/transfers_query.rb
```diff
@@ -0,0 +1,109 @@
...
+            .joins("JOIN induction_records ON participant_profiles.profile_id = induction_records.participant_profile_id")
+            .where(
+              participant_profiles: {
+                induction_records: {
+                  induction_status: "leaving",
+                },
+              },
+            )
+
+          if updated_since.present?
```
cwrw, line 29: double check updated at values

mooktakim (approved):

mooktakim (approved):

PR #3326, opened 2023-04-27
Participant state queries

mattonfoot:
### Context

Queryable Training Record States

Replace ruby service with a Materialised View Model


db/views/training_record_states_v01.sql
```diff
@@ -0,0 +1,252 @@
...
+
+         (SELECT MAX(x)
+          FROM (VALUES
+                    ("induction_records"."start_date"),
+                    ("participant_profiles"."updated_at"),
+                    ("ecf_participant_eligibilities"."updated_at"),
+                    ("ecf_participant_validation_data"."updated_at"),
+                    ("teacher_profiles"."updated_at"),
+                    ("emails"."updated_at")
+               ) AS v (x))                           as "changed_at",
```
peteryates, line 46:
```suggestion
         GREATEST(
               "induction_records"."start_date",
               "participant_profiles"."updated_at",
               "ecf_participant_eligibilities"."updated_at",
               "ecf_participant_validation_data"."updated_at",
               "teacher_profiles"."updated_at",
               "emails"."updated_at"
        )                                            as changed_at
```

app/components/status_tags/school_participant_status_tag.rb
```diff
@@ -31,49 +28,11 @@ def colour
...
-        registered_for_mentor_training
-        registered_for_mentor_training_ero
-        registered_for_mentor_training_primary
-        registered_for_mentor_training_primary_ero
-        registered_for_mentor_training_secondary
-        registered_for_mentor_training_secondary_ero
-        registered_for_mentor_training_duplicate
-        registered_for_mentor_training_duplicate_ero
-      ].include?(record_state)
+      @record_state ||= DetermineTrainingRecordState.call(participant_profile:, induction_record:, school:)&.record_state || :no_longer_involved
```
peteryates, line 35: I wonder if the `record_state` enum should have a fallback/default instead of the `|| :no_longer_involved`. They can be added with something like ` enum status: [:draft, :published], _default: :draft`.

peteryates:
Overall the changes are looking good and the query is clear and easy to follow.

The one concern @cwrw and I had is that when we run the materialized view refresh on the backup instance it takes approximately 25 seconds to refresh:

```
refresh materialized view training_record_states ;
REFRESH MATERIALIZED VIEW
Time: 25368.242 ms (00:25.368)
```

That's on a machine with no load. Perhaps the minutely refresh is too frequent?

Could we split this PR and introduce the materialized view on its own first, so we can refresh it manually a few times and see how long it takes in prod, and also gauge how heavy the hit is on CPU/memory?

If that's all fine we can fine tune the refresh interval and deploy the rest.

Code wise though, I think it's fine overall and I'll be happy to see it merged.

spec/services/determine_training_record_state_spec.rb
```diff
@@ -2,1504 +2,1644 @@
...
       expect(determined_state.training_state).to eq training_state
       expect(determined_state.record_state).to eq record_state
-      expect(I18n.t("status_tags.admin_participant_status").keys).to include record_state
-      expect(I18n.t("status_tags.appropriate_body_participant_status").keys).to include record_state
-      expect(I18n.t("status_tags.delivery_partner_participant_status").keys).to include record_state
-      expect(I18n.t("status_tags.school_participant_status").keys).to include record_state
-      expect(I18n.t("status_tags.school_participant_status_detailed").keys).to include record_state
+    end
+
+    it "Each StatusTag has a language file entry for the record_state of \"#{record_state}\"" do
```
cwrw, line 26: Could we move those expectations back into one `it` block as this will cause the tests to take substantially longer to run?

peteryates (approved):
I think this looks ok now and moving to a single source of truth will solve many problems.

Keeping the materialized view up to date might be problematic so we'll need to keep an eye on it and perhaps introduce some timings/reporting.

PR #3327, opened 2023-04-27
Add an example delivery partner and school

peteryates:
This change is intended to allow delivery partners to get a feel for how the school interface works. Using it in the dev environment will ensure the data is reseeded frequently so we can tweak and adjust what's seeded without needing to worry about manually applying changes in a more static env.

```
irb(main):018:0> school.name
=> "Ambition Institute Test School"
irb(main):019:0> school.partnerships.map(&:delivery_partner).map(&:name)
=> ["Ambition Institute Delivery Partner"]
irb(main):020:0> school.induction_coordinators.first.full_name
=> "Ambition Institute Test SIT"
```


leandroalemao (approved):
lgtm 👍

PR #3334, opened 2023-05-02
Switch super_user? check over to policy

peteryates:
### Context

In response to [this Sentry report](https://dfe-teacher-services.sentry.io/issues/4135058711/?project=5748989&referrer=slack), use policy instead of chaining `#super_user?`


Varsagod (approved):
This looks great

PR #3335, opened 2023-05-02
Make DQT name comparison more robust when one is nil

peteryates:
### Context

We saw some crashes in production when the `#full_name` attribute is nil/blank. When it is, we don't want to try comparing it to the value that comes back via the API, so skip entirely.

- Rename name param to dqt_name
- Don't compare names if either is blank


tonyheadford (approved):
Nice

PR #3307, opened 2023-04-24
CST-1372: cohortless dashboard + participant state determination + no ff

ltello:
### Context
The concept of cohorts is confusing for schools. 
This PR is the first step to remove the concept to the views SITs have access on our service.
We have fully redesigned the school dashboard of participants to make them cohortless.
SITs will manage a single list of all participants at their school.

- [Ticket CST-1372](https://dfedigital.atlassian.net/browse/CST-1372)

### Changes proposed in this pull request
- Landing school dashboard view redesigned: replace cohorts with academic years, single summary of participants
- Participants dashboard unique for all the participants no matter their "cohort" / academic year.
- Remove the option to mentor-unassign ECTs to prevent mentor-orphan ECTs.
- New confirmation screen after a ECT-mentor assignment and journey routing redefined.
- Added list of assigned ECTs to the mentor details view.
- Enable option to make mentor-ECT assignments either from the ECT details page or the mentor's page.
- Removed `cohorts/:cohort_id` segment from the route of some journeys starting from the dashboard (transfering out a participant...)
- Participant statuses has been fully redefined.
- Only schools included in the pilot will get the new academic year (2023/24) displayed.

### Guidance to review
- This is a big iteration in the service affecting many features and internals.
- Pilot schools are under feature flag :cohortless_dashboard.
- We should test not only the dashboard stuff works properly but also all the rest of the service!
- The review app turns this feature on by default but it would be great to test the behaviour with the FF turned off so to make sure the old behaviour has not been affected at all.

### New views
Screenshot 2023-04-14 at 14 45 24
Screenshot 2023-04-14 at 14 47 54
Screenshot 2023-04-14 at 14 47 54

PR #3337, opened 2023-05-02
Cohortless Pilot

ltello:
### Context

This PR merges together 4 pieces of work cohortless-related:
- Modified journey for SITs to be able to roll-over their training programme to next academic year
- Redefinition of statuses of participants 
- Permanent cohort setup for ECTs based on induction start date from their DQT record.
- School dashboard redesign to be fully cohortless for SITs.

See individual pull requests:

  - [roll-over LP/DP partnership for next academic year](https://github.com/DFE-Digital/early-careers-framework/pull/3336)
  - [permanent cohort setup for ECTs](https://github.com/DFE-Digital/early-careers-framework/pull/3265)
  - [cohortless dashboard](https://github.com/DFE-Digital/early-careers-framework/pull/3307)


app/serializers/appropriate_bodies/induction_records_serializer.rb
peteryates, line 1: This will probably need a quick look by a @DFE-Digital/lead-provider-cpd - the status derived differently.

app/serializers/appropriate_bodies/induction_records_serializer.rb
```diff
@@ -45,11 +45,7 @@ class InductionRecordsSerializer
     attribute :training_status, &:training_status
 
     attribute :status do |induction_record|
-      status_name = ParticipantProfileStatus.new(
-        participant_profile: induction_record.participant_profile,
-        induction_record:,
-      ).status_name
-      I18n.t("participant_profile_status.status.#{status_name}.title") if status_name
+      StatusTags::AppropriateBodyParticipantStatusTag.new(participant_profile: induction_record.participant_profile, induction_record:).label
```
peteryates, line 48: Status derived differently here too cc @DFE-Digital/lead-provider-cpd

app/serializers/delivery_partners/participants_serializer.rb
```diff
@@ -14,10 +14,10 @@ def induction_record(participant_profile, delivery_partner)
       end
 
       def status_name(participant_profile, delivery_partner)
-        ParticipantProfileStatus.new(
+        StatusTags::DeliveryPartnerParticipantStatusTag.new(
```
peteryates, line 17: And here cc @DFE-Digital/lead-provider-cpd

app/serializers/appropriate_bodies/induction_records_serializer.rb
cwrw: For this and below, we've chatted to Matt around those changes and those serialisers are not used in the API but the AB/DP UI for csvs, so should be ok. We also discussed the status, and there shouldn't be any changes either

app/serializers/appropriate_bodies/induction_records_serializer.rb
cwrw: see other comment

app/serializers/delivery_partners/participants_serializer.rb
cwrw: see other comment

peteryates (commented):
I've been through this in a decent amount of detail but not line-by-line. I'm happy with the overall direction and happy there's not really much overlap with the API work. I'll take another look once it's green and let's get it merged.

app/controllers/schools/cohort/setup_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Schools
+  module Cohort
```
tonyheadford, line 4: Hmm not sure this is a great module name. I'm guessing that this caused issues in the code (hence the addition of `::` to `Cohort` when referring to the model?

app/forms/schools/add_participants/base_wizard.rb
```diff
@@ -356,21 +360,19 @@ def cohort_to_place_participant
         if transfer?
           existing_participant_cohort || existing_participant_profile&.schedule&.cohort
         elsif ect_participant? && induction_start_date.present?
-          Cohort.containing_date(induction_start_date)
+          ::Cohort.containing_date(induction_start_date)
```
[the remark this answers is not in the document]
ltello: Reverted after rename aboved

app/helpers/application_helper.rb
```diff
@@ -39,12 +39,23 @@ def induction_coordinator_mentor_path(user)
     induction_coordinator_dashboard_path(user)
   end
 
+  def latest_manageable_cohort(school)
```
leoapost, line 42: I see this method is only used in controllers, so I perhaps it should be a service class and add some specs?

app/helpers/schools/dashboard_helper.rb
```diff
@@ -2,8 +2,24 @@
...
 module Schools
   module DashboardHelper
-    def manage_ects_and_mentors?(school_cohort)
-      school_cohort.full_induction_programme? || school_cohort.core_induction_programme?
+    def actions?(participants)
+      participants.orphan_ects.any?
+    end
+
+    def ect_count(school_cohorts)
+      school_cohorts.sum { |sc| sc.current_induction_records.ects.count }
```
leoapost, line 10:
```suggestion
      school_cohorts.sum { |sc| sc.current_induction_records.count }
```

The induction records have one participant, so I think counting the `current_induction_records` should be enough.

Also, Not sure if displaying the ects count is hugely important, so I wonder if it worth to add some specs to ensure the method is counting the `current_induction_records`. Someone could easily be confused and change the method to sum the `induction_records` or `active_induction_records`, which would count more induction records than intended.

app/services/form_data/data_store.rb
```diff
@@ -32,7 +33,20 @@ def get_date(key)
...
 
+  def clean
+    @session[@form_key] = {}
+  end
+
   def destroy
     @session.delete(@form_key)
   end
+
+  def to_s
```
leoapost, line 44: I can see any references to this method. Is it needed?

app/helpers/application_helper.rb
ltello: Yes, you are right. I just placed it there to speed up bug fixes. 
I have now extracted it on to its own service.
Thanks @leoapost

app/helpers/schools/dashboard_helper.rb
ltello: This scope (.ects) is to pick only current induction records belonging to an ECT.

app/services/form_data/data_store.rb
ltello: Not sure where it is used.
Maybe Tony needed it for debugging purposes?

ltello: Or what is me?
Such a long time that I hardly remember.
Let's remove it. We'll find out if it is needed somewhere.

spec/features/schools/training_dashboard/manage_no_ect_training_spec.rb
```diff
@@ -3,7 +3,7 @@
 require "rails_helper"
 require_relative "./manage_training_steps"
 
-RSpec.describe "Manage No ECT training", js: true do
+RSpec.describe "Manage No ECT training", :with_, js: true do
```
leoapost, line 6: Is this a typo?

app/controllers/schools/participants_controller.rb
```diff
@@ -3,21 +3,27 @@
...
+                                                latest_year: Dashboard::LatestManageableCohort.call(@chool).start_year)
   end
 
   def show
     @induction_record = @profile.induction_records.for_school(@school).latest
     @first_induction_record = @profile.induction_records.oldest
     @mentor_profile = @induction_record.mentor_profile
+    @ects = Dashboard::Participants.new(school: @school,
+                                        user: current_user,
+                                        latest_year: Dashboard::LatestManageableCohort.call(@chool).start_year)
```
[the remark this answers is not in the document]
ltello: Haha, You are reviewing quicker than the test runner!!

leoapost (approved):

app/services/dashboard/latest_manageable_cohort.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def initialize(school)
+      @school = school
+    end
+
+    def call
+      [
+        Cohort.latest,
+        pilot? ? Cohort.active_registration_cohort : Cohort.current,
+      ].compact.min_by(&:start_year)
```
tonyheadford, line 15: I might be missing something but I don't know if this could not be summed up as:
```ruby
pilot?  ? Cohort.active_registration_cohort : Cohort.current
```
Won't that always be the case?  The `Cohort.latest` worries me a bit for what it will do in test scenarios or if things are setup ahead of time in the future, it could get misused and it's only needed for this method?

app/views/schools/cohort_setup/_training_confirmation_school_funded_fip.html.erb
```diff
@@ -2,9 +2,9 @@
 <p class="govuk-body">
   You need to <%= govuk_link_to "make your own arrangements with a training provider (opens in a new tab)",
                                 "https://www.gov.uk/guidance/funding-and-eligibility-for-ecf-based-training",
-                                class: "govuk-link govuk-link--no-visited-state",
                                 rel: "noreferrer noopener",
-                                target: "_blank" %>, approved by the DfE.
+                                target: "_blank",
+                                no_visited_sted: true %>, approved by the DfE.
```
tonyheadford, line 7:
```suggestion
                                no_visited_state: true %>, approved by the DfE.
```

app/services/dashboard/latest_manageable_cohort.rb
ltello: Bear in mind that this is computing a min. 
The `Cohort.latest` thing is just useful in cases (specs) where the current cohort has not been created yet, therefore, returning the most recent cohort created rather than returning nil and breaking the test.
In prod, we will always have the Cohort.current created. If not, it would return the previous.

tonyheadford: `Cohort.current` should only change when a new cohort is added with a `registration_start_date` and an `academic_year_start_date` that takes over from the previous one?

ltello: `Cohort.current` is only based on the current date.

tonyheadford: Thanks for clarifying @ltello, I still don't think this should be here especially if it's just a "hack" to fix specs.  My main concern is `Cohort.latest` as I don't think that should exist because I think it could too easily be misunderstood and we have the necessary methods we need in `current`,`previous` and `next` on `Cohort`.  I don't want to hold up this PR though but can you add a tech debt ticket to get rid of this and correct the specs that might be missing a Cohort creation and/or `travel_to` blocks to work correctly with the cohorts used in those specs?

tonyheadford (approved):
Great job @ltello! Thanks for sticking with it

PR #3355, opened 2023-05-05
Fix failing mailers

cwrw:
### Context
To resolve errors in sentry, use correct way to call mailers and fix tests

- Ticket: n/a

- Use with instead of mailer name, and move name to be called before delivering later
- Access attributes using params in mailer methods themselves
- Fix tests with use `with` or call enqueued with.

### Guidance to review
I think in some cases we are now missing exact mailer methods to test, esp in feature tests where it's hard to expect a subject to enqueue the mailer. Open to suggestions

### DIFF
--- a/spec/services/induction/send_transfer_notification_emails_spec.rb
+++ b/spec/services/induction/send_transfer_notification_emails_spec.rb
@@ -15,7 +15,15 @@ RSpec.describe(Induction::SendTransferNotificationEmails) do
 
   shared_examples "notifying the participant" do
     it "sends a 'particicpant transfter in notification' email to the participant" do
-      expect(ParticipantTransferMailer).to(have_received(:participant_transfer_in_notification).with(induction_record:))
+      expect {
+        subject.call
+      }.to have_enqueued_mail(ParticipantTransferMailer, :participant_transfer_in_notification)
+        .with(
+          params: {
+            induction_record:,
+          },
+          args: [],
+        ).once
     end
   end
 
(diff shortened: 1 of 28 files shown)


spec/services/update_induction_tutor_reminder_spec.rb
```diff
@@ -16,18 +16,10 @@
...
-      subject.send!
-
-      expect(SchoolMailer).to have_received(method_name).with(
-        school:,
-        sit_name: school.induction_tutor.full_name,
-        nomination_link: subject.instance_variable_get(:@nomination_link),
-      )
+      expect {
+        subject.send!
+      }.to have_enqueued_mail(SchoolMailer, method_name)
```
cwrw, line 22: removed those as couldn't use `instance_variable_get` open to suggestions on how to get this

spec/services/induction/send_transfer_notification_emails_spec.rb
```diff
@@ -63,10 +69,17 @@
...
+        expect {
+          subject.call
+        }.to have_enqueued_mail(ParticipantTransferMailer, template)
+          .with(
+            params: {
+              induction_record:,
+              lead_provider_profile:,
+            },
+            args: [],
+          ).once
```
peteryates, line 81: This makes way more sense

spec/services/update_induction_tutor_reminder_spec.rb
peteryates: I've had a bit of a try, I expected [hash_including](https://www.rubydoc.info/gems/rspec-mocks/RSpec%2FMocks%2FArgumentMatchers:hash_including) to work but it wouldn't :( 

Think what you came up with is the best for the time being.

peteryates (approved):
Look great ❤️

PR #3381, opened 2023-05-12
[CPLD-2165] Amend schools endpoint

cwrw:
### Context

The behaviour of the schools endpoint has changed as we don't want to wait for school cohort to be available to show schools.

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2165

Return all schools per cohort which are
- Eligible
- Not CIP only
- OR have a partnership, as the eligible check is at the higher level and not the cohort level, meaning some schools might have been eligible and had relationships but are no longer eligible. Ensure those partnerships are not challenged or relationship partnerships

### Guidance to review
Did I miss anything?


leandroalemao (approved):
lgtm 👍 😄

mooktakim (approved):

PR #3385, opened 2023-05-12
Add versioning to multiple models

leoapost:
### Context

- Ticket: N/A

Add versioning to models that we need to track changes.

Add has_paper_trail to the following models
- AppropriateBody
- Cohort
- CoreInductionProgramme
- CpdLeadProvider
- ECFParticipantValidationData
- ParticipantIdentity
- School

### DIFF
--- a/app/models/ecf_participant_validation_data.rb
+++ b/app/models/ecf_participant_validation_data.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class ECFParticipantValidationData < ApplicationRecord
-  has_paper_trail
-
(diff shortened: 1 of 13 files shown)


app/models/ecf_participant_validation_data.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class ECFParticipantValidationData < ApplicationRecord
+  has_paper_trail
```
tonyheadford, line 4: Don't think we should have this one as it's PII and in theory we want to get rid if this as soon as we identify the person

tonyheadford (approved):

PR #3390, opened 2023-05-15
[CPDNPQ 1008] Add data filters to edge case page

javier-npq:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1008

- added data filters for fundingEligibilityStatusCode
- added data filters for employmentType
- added specs for new data filters

### DIFF
--- a/app/controllers/admin/npq/applications/edge_cases_controller.rb
+++ b/app/controllers/admin/npq/applications/edge_cases_controller.rb
@@ -15,6 +15,8 @@ module Admin
           query_string = params[:query]
           funding_eligiblity_status_code = params[:FundingEligiblityStatusCode]
           employment_type = params[:EmploymentType]
+          # Currently the component used is set up to build date parameters in the following manner
+          # date(3i) = day, date(2i) = month, date(1i) = year
           start_date = convert_date(params["start_date(3i)"], params["start_date(2i)"], params["start_date(1i)"])
           end_date = convert_date(params["end_date(3i)"], params["end_date(2i)"], params["end_date(1i)"])
 
--- a/app/controllers/admin/npq/applications/edge_cases_controller.rb
+++ b/app/controllers/admin/npq/applications/edge_cases_controller.rb
@@ -13,9 +13,13 @@ module Admin
 
         def index
           query_string = params[:query]
+          funding_eligiblity_status_code = params[:FundingEligiblityStatusCode]
+          employment_type = params[:EmploymentType]
+          start_date = convert_date(params["start_date(3i)"], params["start_date(2i)"], params["start_date(1i)"])
+          end_date = convert_date(params["end_date(3i)"], params["end_date(2i)"], params["end_date(1i)"])
 
           results = Admin::NPQApplications::EdgeCaseSearch
-            .new(policy_scope(NPQApplication), query_string:).call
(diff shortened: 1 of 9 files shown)


app/controllers/admin/npq/applications/edge_cases_controller.rb
```diff
@@ -13,9 +13,13 @@ class EdgeCasesController < Admin::BaseController
 
         def index
           query_string = params[:query]
+          funding_eligiblity_status_code = params[:FundingEligiblityStatusCode]
+          employment_type = params[:EmploymentType]
+          start_date = convert_date(params["start_date(3i)"], params["start_date(2i)"], params["start_date(1i)"])
```
shjohnson, line 18: is there anyway to give more human readable param names? something like `start_date(day)`, `start_date(month)` etc

app/models/npq_application.rb
```diff
@@ -50,6 +50,7 @@ class NPQApplication < ApplicationRecord
                                                                         marked_ineligible_by_policy
                                                                         marked_funded_by_policy])
                              }
+  scope :created_at_range, ->(start_date, end_date) { where(created_at: start_date..end_date) }
```
shjohnson, line 53: 👏

app/services/admin/npq_applications/edge_case_search.rb
```diff
@@ -37,6 +45,30 @@ def search_conditions
     end
   end
 
+  def funding_eligiblity_status_code_conditions
```
shjohnson, line 48: Was going to say all these methods are very similar and could be dried up but i think this is fine for just 3.

spec/services/admin/npq_applications/edge_case_search_spec.rb
```diff
@@ -102,5 +102,48 @@
         expect(subject.call).not_to include(application_2)
       end
     end
+
```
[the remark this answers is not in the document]
shjohnson: and not the sub query of the different statuses or dates?

shjohnson:
Can wee add some randomised seed data with different dates for the edge cases mainly for the review app when it gets built. possibly as well other bits of data.

app/controllers/admin/npq/applications/edge_cases_controller.rb
javier-npq: Yeah that's fiddly. Left a comment for now, Went through the [documentation for this component here ](https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder/Builder#govuk_date_field-instance_method)and couldn't find anything out of the box that we can use.

app/services/admin/npq_applications/edge_case_search.rb
javier-npq: I did try to do it like this in the beginning which was a much cleaner way image
But didn't get it running due to the way the flow is set up. Kept on getting the following error.
image
I can move them to a separate helper/concern class and call it from there but it's just three methods so i left them here.

spec/services/admin/npq_applications/edge_case_search_spec.rb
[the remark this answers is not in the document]
javier-npq: Yeah, they are now.

PR #3393, opened 2023-05-15
Change the Notify API key to a test one

peteryates:
### Context

We have been trying to send emails from our deployed development environments (review apps) to an email account that no longer exists (since the migration from Google to Microsoft). As a result the log in Notify is full of errors which add unnecessary noise.

This change switches the old key for a new one that doesn't actually send emails but does log that they've been pushed to Notify in the log.


tonyheadford (approved):
Awesome!

PR #3397, opened 2023-05-16
Fix failing version table creation in review app

cwrw:
### Context
Adding papertrail to cohorts caused failures in migrations where versions table was expected but didn't exist
- Ticket: n/a

Ignore papertrail in the particular migration to avoid failures when creating review apps


peteryates (approved):
Makes sense in this case

PR #3402, opened 2023-05-16
[Bug-fix] Prevent providers from claiming FIP schools that haven't setup their cohort

tonyheadford:
### Context

The rules in place currently are not 100% correct. When a provider claims a school, if that school ran FIP in the previous cohort with a partner and has not set up their next cohort, no provider should be able to claim the school.

Fixes to the CSV uploader for reporting schools
Fixes to the Create service for partnerships used by the API.


cwrw (approved):
thanks @tonyheadford 🙌

PR #3403, opened 2023-05-16
Support all Lead provider versions in Open API spec

cwrw:
### Context

We only support returning the Open API spec for v1 in the lead provider API. 
We need to support all versions
- Ticket: n/a

- Support all API versions in open API spec
- Return latest version as default
- Only allow existing version spec to be rendered, otherwise return 404
- Move lead provider specification to app/lib to avoid loading problems

### Guidance to review
Did I miss anything?


mooktakim (approved):

fused4 (approved):
lgtm 👍 😄

leandroalemao (approved):

PR #3416, opened 2023-05-19
Add the SIT email address to the SIT nomination Notify template

leoapost:
### Context

- Ticket: CST-1504

We want to include the SIT's email address in the body of the SIT nomination email template.

1. pass the email address to the template with a personalisation var
2. update the template content in Notify when this commit is released


leoapost:
There's no such need anymore.

PR #3417, opened 2023-05-19
Mentors from another school do not have induction record in the dashb…

ltello:
…oarded school

### Context

The dasboard breaks when one of the mentors do not belong to the shool of the dashboard.

- Ticket: 

The dashboard is made based on the induction records of all their participants.
Mentors from another school teaching ECTs in the school of the dashboard do not have induction record in it.

- This PR will take this into account and check the mentor pool instead for mentors.
- Also, make sure mentors get added to school mentor pool thoughout the code when they are created/moved to a school.

### DIFF
--- a/app/services/dashboard/mentor.rb
+++ b/app/services/dashboard/mentor.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Dashboard
+  class Mentor
+    attr_reader :induction_record, :participant_profile
+
+    def initialize(induction_record:, participant_profile: induction_record&.participant_profile)
+      @induction_record = induction_record
+      @participant_profile = participant_profile
+    end
+
+    def name
+      induction_record&.participant_full_name || participant_profile.full_name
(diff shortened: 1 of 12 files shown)


app/services/dashboard/mentor.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Dashboard
+  class Mentor
+    attr_reader :induction_record, :participant_profile
+
+    def initialize(induction_record:, participant_profile: induction_record&.participant_profile)
```
tonyheadford, line 7: Curious to know when these could not belong together? i.e. the induction_record.participant_profile is not equal to participant_profile ?  If this is not the case then could we remove participant_profile as a param?

ltello: Oh this is legacy from an initial implementation.
I can remove the default value because now this method is meant to be called with either the induction record or the participant profile.

PR #3437, opened 2023-05-24
[CPDLP-2204] Fix ordering of transfers endpoint

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2204

* Default ordering is now `induction_records.created_at ASC`

### DIFF
--- a/app/services/api/v3/ecf/transfers_query.rb
+++ b/app/services/api/v3/ecf/transfers_query.rb
@@ -10,14 +10,13 @@ module Api
         end
 
         def users
-          scope = participants_with_school_transfer.or(leaving_participants)
+          pp_scope = participants_with_school_transfer.or(leaving_participants)
 
-          scope = User
+          sub_query = User
             .includes(user_includes)
-            .select("users.*", "induction_records.created_at")
-            .distinct
+            .select("users.*", "induction_records.created_at AS induction_records_created_at")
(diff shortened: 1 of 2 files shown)


app/services/api/v3/ecf/transfers_query.rb
```diff
@@ -14,6 +14,7 @@ def users
 
           scope = User
             .includes(user_includes)
+            .select("users.*", "induction_records.created_at")
```
cwrw, line 17: discussed in person: we need to ensure distinct runs on ids and not on created at

spec/services/api/v3/ecf/transfers_query_spec.rb
```diff
@@ -172,6 +172,33 @@
...
+
+      let!(:another_transfer_induction_record) do
+        NewSeeds::Scenarios::Participants::Transfers::FipToFipKeepingOriginalTrainingProvider
+          .new(lead_provider_from: lead_provider)
+          .build
+      end
+      let!(:another_user) { another_transfer_induction_record.preferred_identity.user }
+
+      before do
+        teacher_profile = TeacherProfile.find_by!(user_id: another_user.id)
```
cwrw, line 192: discussed in person: we can update the induction record directly

PR #3453, opened 2023-05-26
[CPDLP-2212] Improve performance on participant declarations in v3

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2212

* `mentor_id` returned correct in a non-N+1 way

### DIFF
--- a/app/services/api/v3/participant_declarations_query.rb
+++ b/app/services/api/v3/participant_declarations_query.rb
@@ -34,6 +34,11 @@ module Api
           :participant_profile,
           :cpd_lead_provider,
         )
+        .joins(participant_profile: :induction_records)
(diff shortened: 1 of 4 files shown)


app/services/api/v3/participant_declarations_query.rb
```diff
@@ -34,6 +34,11 @@ def participant_declarations
           :participant_profile,
           :cpd_lead_provider,
         )
+        .joins(participant_profile: :induction_records)
```
cwrw, line 37: could we add the selection of the mentor profile id to the statement please, also a test :)?

cwrw (approved):

PR #3474, opened 2023-06-01
Small content amends in guidance

hazelyule:
Small content updates 
* corrected typos 
* added application id endpoint 
* updated application endpoint response bodies 

NB these updates had been in Brian's acccordion branch but pulling out here so as to release independently.


leandroalemao (approved):
lgtm 👍 😄 tks

PR #3485, opened 2023-06-03
Revert "Revert "Participant state queries""

peteryates:
Reverts DFE-Digital/early-careers-framework#3484


app/models/participant_profile.rb
```diff
@@ -56,6 +56,21 @@ class ParticipantProfile < ApplicationRecord
   scope :mentors, -> { where(type: Mentor.name) }
   scope :npqs, -> { where(type: NPQ.name) }
 
+  # search by record state
```
cwrw, line 59: I think we need to remove all this?

app/services/determine_training_record_state.rb
```diff
@@ -2,26 +2,156 @@
 
 # noinspection RubyTooManyMethodsInspection, RubyTooManyInstanceVariablesInspection, RubyInstanceMethodNamingConvention
 class DetermineTrainingRecordState < BaseService
+  attr_reader :participant_profile_id, :school_id, :appropriate_body_id, :delivery_partner_id, :induction_record_id
+
+  # enum validation_state: {
```
cwrw, line 7: did we need to convert the rest of the  enums or happy to remove completely?

peteryates: I think having them is going to be useful even if it's just in a documentation capacity; it's not going to be immediately clear what the possible values for each state are. I'll reformat it into a comment with a bit of an explanation

app/models/participant_profile.rb
```diff
@@ -56,6 +56,9 @@ class ParticipantProfile < ApplicationRecord
   scope :mentors, -> { where(type: Mentor.name) }
   scope :npqs, -> { where(type: NPQ.name) }
 
+  # search by validation state
+  scope :validation_state_valid, -> { joins(:training_record_states).where(training_record_states: { validation_state: "valid" }) }
```
cwrw, line 60: ✂️

spec/services/determine_training_record_state_spec.rb
```diff
@@ -2,1504 +2,1642 @@
...
       expect(determined_state.training_state).to eq training_state
       expect(determined_state.record_state).to eq record_state
-      expect(I18n.t("status_tags.admin_participant_status").keys).to include record_state
-      expect(I18n.t("status_tags.appropriate_body_participant_status").keys).to include record_state
-      expect(I18n.t("status_tags.delivery_partner_participant_status").keys).to include record_state
-      expect(I18n.t("status_tags.school_participant_status").keys).to include record_state
-      expect(I18n.t("status_tags.school_participant_status_detailed").keys).to include record_state
+    end
+
+    it "Each StatusTag has a language file entry for the record_state of \"#{record_state}\"" do
```
cwrw, line 24: could we keep this in the same it block to avoid long tests?

cwrw (approved):
minor comments but otherwise 👍

peteryates:
Thank you  @cwrw,  think I've addressed those points.

cwrw (approved):

PR #3486, opened 2023-06-04
Improve Participant declaration queries

cwrw:
### Context
Participant declaration queries are causing db load, improve them in all versions of the API

- Ticket: n/a

- Use OR instead of union
- Only select IDs in v3 to improve performance
- Eager load NPQ associations to avoid n+1s in v3

### Guidance to review
Tested before and after numbers for v1/v2/v3 and they match up
also tested filters for all endpoints

### DIFF
--- a/app/serializers/api/v3/participant_declaration_serializer.rb
+++ b/app/serializers/api/v3/participant_declaration_serializer.rb
@@ -89,7 +89,7 @@ module Api
 
       attribute :has_passed do |declaration|
         if declaration.npq?
-          declaration.outcomes.latest&.has_passed?
+          declaration.outcomes.sort_by(&:created_at).reverse!&.first&.has_passed?
         end
       end
     end
--- a/app/services/api/v3/participant_declarations_query.rb
+++ b/app/services/api/v3/participant_declarations_query.rb
@@ -25,7 +25,10 @@ module Api
           scope = scope.where(delivery_partner_id: delivery_partner_ids)
         end
 
-        scope.select(:id).distinct
+        scope
+          .select(:id, :created_at)
+          .order(:created_at)
+          .distinct
       end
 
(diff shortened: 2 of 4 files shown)


app/serializers/api/v3/participant_declaration_serializer.rb
```diff
@@ -89,7 +89,7 @@ class ParticipantDeclarationSerializer
 
       attribute :has_passed do |declaration|
         if declaration.npq?
-          declaration.outcomes.latest&.has_passed?
+          declaration.outcomes.sort_by(&:created_at).reverse!&.first&.has_passed?
```
cwrw, line 92: to avoid n+1s as latest runs a new query to order

app/services/api/v3/participant_declarations_query.rb
```diff
@@ -51,9 +48,14 @@ def participant_declarations_from(paginated_join)
               :cpd_lead_provider,
             )
             .from("(#{sub_query.to_sql}) as participant_declarations")
+            .order(:created_at)
             .distinct
 
-        scope.order(:created_at)
+        preloader = ActiveRecord::Associations::Preloader.new
+        preloader.preload(scope.select { |p| p.type == "ParticipantDeclaration::NPQ" }, :outcomes)
+        preloader.preload(scope.select { |p| p.type == "ParticipantDeclaration::NPQ" }, participant_profile: :npq_application)
```
peteryates, line 56: I'm sure there's room for improvement here but as it's working let's go with it for now

peteryates (approved):

peteryates:
Looks good, happy with this if you are

cwrw:
@peteryates could you have a look at the latest commit please

peteryates (approved):

peteryates:
Makes sense 👍🏽

PR #3488, opened 2023-06-05
[CST-1536] Set partnership challenge deadline for 2023 cohort

lazydevorg:
### Context

- Ticket: CST-1536


app/models/partnership.rb
```diff
@@ -67,4 +68,10 @@ def cohort_challenge_deadline
   def update_analytics
     Analytics::UpsertECFPartnershipJob.perform_later(partnership: self) if saved_changes?
   end
+
+  def set_challenge_deadline_by_cohort
+    if cohort.start_year == 2023 && created_at < Date.new(2023, 10, 17)
```
ltello, line 73: I think what we need here is something like:

```
CHALLENGE_PERIOD_AFTER_CREATION = 2.weeks

def set_challenge_deadline
  [
     cohort.academic_year_start_date + CHALLENGE_PERIOD_SINCE_ACADEMIC_YEAR_START,
     Time.current + CHALLENGE_PERIOD_AFTER_CREATION
  ].max
end
```
so that the deadline will be 31 Oct each academic year (1 Sept + 2.months) for partnerships reported before Oct, 17th
or 2.weeks after the creation for partnerships reported after Oct, 17th

ltello (changes requested):

app/models/partnership.rb
lazydevorg: is two weeks the default already?
Furthermore the ask in the ticket is to explicitly do this for 2023 only.
Saying so, do you think we still need the above?

tonyheadford: Yes agree with @ltello's suggestion

spec/factories/partnership.rb
```diff
@@ -16,6 +16,12 @@
     challenge_reason { "mistake" }
   end
 
+  trait :not_challenged do
+    challenged_at { nil }
+    challenge_reason { nil }
+    challenge_deadline { nil }
```
tonyheadford, line 22: This shouldn't have `challenge_deadline` in for "not challenged"

spec/models/partnership_spec.rb
```diff
@@ -75,7 +75,8 @@
     end
 
     it "returns false when the challenge_deadline is blank" do
-      partnership = create(:partnership, challenge_deadline: nil)
+      partnership = create(:partnership)
+      partnership.update!(challenge_deadline: nil)
```
tonyheadford, line 79: Why does this need to change?
```suggestion
      partnership = create(:partnership, challenge_deadline: nil)
```

spec/services/partnerships/report_spec.rb
```diff
@@ -150,4 +150,36 @@
...
+  context "challenge deadline to 31st October from 2023" do
+    let(:cohort) { create(:cohort, start_year: 2023) }
+
+    it "sets the challenge deadline to 31st October when the partnership is created before the 17th October" do
+      travel_to Date.new(2023, 5, 1)
+
+      expect(result.challenge_deadline).to eq(Date.new(2023, 10, 31))
+
+      travel_back
+    end
```
tonyheadford, line 163:
```suggestion
    it "sets the challenge deadline to 31st October when the partnership is created before the 17th October", travel_to: Date.new(2023, 5, 1) do

      expect(result.challenge_deadline).to eq(Date.new(2023, 10, 31))
    end
```

tonyheadford (commented):
I've made some suggestions, think it's important to use time travel at a block level to prevent leakage errors occurring

spec/factories/partnership.rb
lazydevorg: The challenge_deadline is assigned randomly by the factory. We need to explicitly set it to nil.

tonyheadford (approved):

ltello (approved):

app/services/partnerships/report.rb
```diff
@@ -2,7 +2,8 @@
 
 module Partnerships
   class Report < BaseService
-    CHALLENGE_WINDOW = 14.days.freeze
+    INITIAL_CHALLENGE_WINDOW = (2.months - 1.day).freeze
```
tonyheadford, line 5: Can we use better naming here if you want to do this, suggest something like:
DEFAULT_CHALLENGE_WINDOW (2 weeks from now)
EARLY_YEAR_CHALLENGE_WINDOW (the October one)?

tonyheadford (changes requested):

tonyheadford (approved):

PR #3495, opened 2023-06-05
Fix training record state ordering

peteryates:
### Context

Add an `ORDER BY` to the state query so the results are always consistent and there's no chance of us displaying different statuses for the same participant.

- Ensure training states are ordered by changed_at
- Remove grouping from emails

### DIFF
--- a/app/services/determine_training_record_state.rb
+++ b/app/services/determine_training_record_state.rb
@@ -500,7 +500,7 @@ private
           "individual_training_record_states"."lead_provider_id",
           "individual_training_record_states"."delivery_partner_id",
           "individual_training_record_states"."appropriate_body_id",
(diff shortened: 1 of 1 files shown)


app/services/determine_training_record_state.rb
```diff
@@ -555,8 +566,9 @@ def final_grouping
 
       WHERE
           "individual_training_record_states"."participant_profile_id" = '#{participant_profile_id}'
+
       ORDER BY
-        "individual_training_record_states"."changed_at"
+        "individual_training_record_states"."changed_at" DESC
```
peteryates, line 571: 🤦🏽‍♂️

cwrw (approved):

PR #3496, opened 2023-06-05
Add TrainingRecordState table and model

peteryates:
### Context

This change adds a new table called `training_record_states` that will eventually hold the various statuses for each participant profile.

Unlike the previous materialized view this one isn't related directly to an induction record because they are appended to the table so the relationship is likely to be outdated.

### DIFF
--- a/app/services/determine_training_record_state.rb
+++ b/app/services/determine_training_record_state.rb
@@ -148,14 +148,14 @@ class DetermineTrainingRecordState < BaseService
     not_registered_for_training
   ].each_with_object({}) { |v, h| h[v] = v }.freeze
 
-  # for now, if there is a TrainingRecordState matching our conditions return it, otherwise
-  # perform a sync
   def call
-    TrainingRecordState.latest_for(**params) || sync
-  end
+    if (training_record_state = TrainingRecordState.find_by(**params))
+      training_record_state.assign_attributes(**execute_query.first)
 
-  def sync
-    TrainingRecordState.create(**execute_query.first)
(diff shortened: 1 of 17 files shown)


db/migrate/20230605110212_readd_training_record_states_as_table.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class ReaddTrainingRecordStatesAsTable < ActiveRecord::Migration[6.1]
```
mattonfoot, line 3: Spelling mistake... ReaddTrainingRecordStatesAsTable > ReadTrainingRecordStatesAsTable

peteryates: That was my attempt at spelling re-add 😅

app/services/determine_training_record_state.rb
```diff
@@ -148,39 +148,23 @@ class DetermineTrainingRecordState < BaseService
...
-    result = ActiveRecord::Base.connection.execute(query)
+    if (training_record_state = TrainingRecordState.find_by(**params))
+      training_record_state.assign_attributes(**execute_query.first)
 
-    Record.new(**result.first)
+      training_record_state.save! if training_record_state.changed?
+      training_record_state
+    else
+      TrainingRecordState.create!(**execute_query.first)
+    end
```
peteryates, line 159: This is the slightly complicated bit I need to revisit and add plenty of tests around.

peteryates:
Closing for now, the view is doing fine. Will revisit when time allows.

PR #3498, opened 2023-06-06
Handle deleted record entities in participant change log

mattonfoot:
### Context

- Ticket: fix change log page broken due to missing user entities

use find_by when looking for entities from history rather than find which raises exceptions


peteryates (approved):
Looks good to me. If this was doing something more fundamental than building history I'd suggest rescuing from `RecordNotFound` but here it's fine.

peteryates (approved):
Looks good, works well

PR #3500, opened 2023-06-06
Use new cohort setup wizard in place of choose programme wizard

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-1600)

Fixing issues spotted whilst testing a different feature. A new cohort setup journey was recently added but it did not fully replace the existing cohort setup journey so we have 2 journey's doing basically the same thing. This PR replaces the use of the old journey with the new one.
It is noted that they are not 100% the same journey and the UX may not be optimal in all scenarios with the new journey currently.

Use new cohort setup journey in place of old one.
Fix issue with schools that had chosen FIP but not partnered causing an error.


ltello:
There are several other scenarios in the service where the SIT is redirected to the old setup-cohort journey `school_choose_programme_path`. I think we should review them all and decide wether to replace them with the new one `schools_cohort_setup_start_path` or not.

tonyheadford:
> There are several other scenarios in the service where the SIT is redirected to the old setup-cohort journey `school_choose_programme_path`. I think we should review them all and decide wether to replace them with the new one `schools_cohort_setup_start_path` or not.

Yes, I just pushed this early so you could see what I was doing

PR #3501, opened 2023-06-06
Order V3 Participant query pagination by user

cwrw:
### Context
If we don't order on the initial pagination query, we will get any random first 100 users, and then order them. This avoids getting random users first and then sorting

- Ticket: n/a

Add ordering on the pagination query using users

### Guidance to review
Did I miss anything?


leandroalemao (approved):
lgtm 👍 😄

mooktakim (approved):

PR #3503, opened 2023-06-07
Patch training state for induction records

mattonfoot:
### Context

Still seeing some induction statuses being ordered badly because touching overrides all start_dates

- only use participant_profile.updated_at if induction_record.start_date is not present
- group by changed_at and select the earliest as may see duplicate records when partnerships, eligibility or validation changes
- also ensure Mentee counts only uses active and leaving records to calculate current mentee counts rather than historical mentee counts


peteryates (approved):
Makes sense to me

PR #3505, opened 2023-06-07
Allow filtering of NPQ applications on participant_id

ethax-ross:
[Jira-CPDLP-2209](https://dfedigital.atlassian.net/browse/CPDLP-2209)

### Context

The API v3 schema surfaces the ability to filter NPQ applications by `participant_id`, however we have not yet implemented the functionality.

We also want to always default the sorting of NPQ applications in the v1 API as the `sort` parameter is not supported.

- Always sort v1 applications by created_at
- Allow filtering of applications by participant_id

### Guidance to review

Test added to v1 request specs to cover default sorting.

Migrates to a query model for consistency with other v3 endpoints.

Request test added for `participant_id` filtering, but we may want to clean up these controller tests at a later date so we're not over-testing the query models. It looks like there's also scope to DRY up the query models with a concern similar to `ApiFilter` and/or additional scopes on the core models.

Example request using the `ambition-token` auth token:

https://ecf-review-pr-3505.london.cloudapps.digital/api/v3/npq-applications?filter[participant_id]=abb6bd5f-dd1a-4352-93d0-6f3aa0a6c302

### DIFF
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -57,7 +57,7 @@ module Api
                   .includes(:cohort, :npq_course, :profile, participant_identity: [:user])
                   .where(cohort: with_cohorts)
         scope = scope.where("updated_at > ?", updated_since) if updated_since.present?
-        params[:sort].blank? ? scope.order("npq_applications.created_at ASC") : scope
+        scope.order("npq_applications.created_at ASC")
       end
 
       def access_scope
--- a/app/controllers/api/v3/npq_applications_controller.rb
+++ b/app/controllers/api/v3/npq_applications_controller.rb
@@ -8,13 +8,30 @@ module Api
       # Returns a list of NPQ applications
       # Providers can see their NPQ applications which cohorts they apply to via this endpoint
       #
-      # GET /api/v3/npq-applications?filter[cohort]=2022&sort=name,-updated_at
+      # GET /api/v3/npq-applications?filter[cohort]=2022&filter[updated_since]=2023-05-10%2015%3A00%3A00.000&filter[participant_id]=be483a5a-0bb3-4336-82fa-9bd0e4c58761&sort=name,-updated_at
       #
       def index
-        render json: json_serializer_class.new(paginate(query_scope.order(sort_params(params)))).serializable_hash.to_json
+        render json: json_serializer_class.new(paginate(npq_applications)).serializable_hash.to_json
       end
 
     private
+
+      def npq_applications
+        @npq_applications ||= npq_applications_query.applications.order(sort_params(params))
+      end
+
+      def npq_applications_query
+        Api::V3::NPQApplicationsQuery.new(
+          npq_lead_provider:,
+          params: npq_application_params,
+        )
+      end
+
+      def npq_application_params
+        params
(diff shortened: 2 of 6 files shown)


app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -57,7 +57,7 @@ def query_scope
                   .includes(:cohort, :npq_course, :profile, participant_identity: [:user])
                   .where(cohort: with_cohorts)
         scope = scope.where("updated_at > ?", updated_since) if updated_since.present?
-        params[:sort].blank? ? scope.order("npq_applications.created_at ASC") : scope
```
cwrw, line 60: sorry that wasn't clear in the ticket, could we keep `scope.order("npq_applications.created_at ASC")` but remove the `params[:sort].blank?` as that param only applies to v3?

app/controllers/api/v3/npq_applications_controller.rb
```diff
@@ -19,6 +19,17 @@ def index
       def json_serializer_class
         Api::V3::NPQApplicationSerializer
       end
+
+      def query_scope
+        scope = super
```
cwrw, line 24: We introduced new patterns in v3 where we moved queries out of controllers into query services, maybe we can do something similar here and move this into it's own query service with specs? please find an example of another query service [here](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/api/v3/npq_participants_query.rb).

cwrw (commented):
looking good, a few suggestions

app/controllers/api/v3/npq_applications_controller.rb
```diff
@@ -8,14 +8,31 @@ class NPQApplicationsController < V1::NPQApplicationsController
       # Returns a list of NPQ applications
       # Providers can see their NPQ applications which cohorts they apply to via this endpoint
       #
-      # GET /api/v3/npq-applications?filter[cohort]=2022&sort=name,-updated_at
+      # GET /api/v3/npq-applications?filter[cohort]=2022&updated_since=2023-05-10%2015%3A00%3A00.000&participant_id=be483a5a-0bb3-4336-82fa-9bd0e4c58761&sort=name,-updated_at
```
cwrw, line 11:
```suggestion
      # GET /api/v3/npq-applications?filter[cohort]=2022&filter[updated_since]=2023-05-10%2015%3A00%3A00.000&filter[participant_id]=be483a5a-0bb3-4336-82fa-9bd0e4c58761&sort=name,-updated_at
```

app/services/api/v3/npq_applications_query.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+module Api
+  module V3
+    class NPQApplicationsQuery
+      INITIAL_COHORT_YEAR = 2021
```
cwrw, line 6: suggestion: maybe we can use 2020 and more than similar to other queries, so when we get to clean this and maybe add a scope it's easier to find in the code

cwrw (approved):
🙌

app/services/api/v3/npq_applications_query.rb
ethax-ross: Yeah that's a sensible idea I'll update it to be consistent 👍

PR #3513, opened 2023-06-08
Default cohort filtering to exclude pre-2021 cohorts

ethax-ross:
### Context

If a `cohort` filter is not supplied we should default to filtering out cohorts earlier than 2021. Due to a bug in these query models if `cohort` was not specified we would return all models irrespective of cohort.

Default cohort filtering to exclude pre-2021 cohorts and add test coverage.


cwrw (approved):
thanks for fixing this :)

PR #3525, opened 2023-06-13
[CST-1622] Fix moving mentors with Induction::ChangeProgramme

leoapost:
### Context

- Ticket: CST-1622

There are a few bugs when using the `Induction::ChangeProgramme` with mentors:
1. When mentors are moving schools, it will remove them from the new school's mentor pool
2. When the mentors are not moving schools, it will remove them from the school's mentor pool and will also remove them from their mentees by updating their induction records.

1. Store the participant_profile.current_induction_record before reuse it, so it always returns the correct induction record
2. Create an additional check to determine if the mentor is actually moving school

### DIFF
--- a/app/services/induction/change_programme.rb
+++ b/app/services/induction/change_programme.rb
@@ -22,10 +22,11 @@ class Induction::ChangeProgramme < BaseService
 
 private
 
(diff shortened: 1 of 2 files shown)


tonyheadford (approved):
LGTM, thanks

PR #3526, opened 2023-06-13
Populate mentor_user_id for new ParticipantDeclarations

ethax-ross:

### Context

The `ParticipantDeclaration` has a new foreign key `mentor_user_id` that is currently unpopulated. We want to
populate this with the induction record relative to the `declaration_date` at the time of creation.

- Populate mentor_user_id for new declarations
- Clarify induction record date edge cases

### Guidance to review

The field is not yet referenced/read anywhere. We will backfill the existing declarations in a separate PR.

We have induction records in the database that are in an invalid/unexpected state. Until we are able to clean up the
data and ensure consistency we need to make sure the latest induction record logic is expected/documented.

Update `InductionRecord` scopes to prioritise records based on the latest `end_date` if they have the same `start_date`.

Add test coverage to the `FindBy` class to cover the following edge cases:

- Induction records with the same start date where one has a `nil` end date
- Induction records with the same start date where one has a later end date
- The expected scenario of the latest induction record having a `nil` end date

### DIFF
--- a/app/services/record_declaration.rb
+++ b/app/services/record_declaration.rb
@@ -156,8 +156,21 @@ private
       declaration_type:,
       cpd_lead_provider:,
       delivery_partner:,
+      mentor_user_id:,
       user: participant_identity.user,
     }
+  end
+
+  def mentor_user_id
+    return nil unless participant_profile.ect?
+
+    induction_record = Induction::FindBy.call(
+      participant_profile:,
+      lead_provider: cpd_lead_provider.lead_provider,
+      date_range: ..declaration_date,
+    )
+
+    induction_record&.mentor_profile&.participant_identity&.user_id
   end
 
   def delivery_partner
--- a/app/models/induction_record.rb
+++ b/app/models/induction_record.rb
@@ -70,8 +70,8 @@ class InductionRecord < ApplicationRecord
 
   scope :for_school, ->(school) { joins(:school).where(school: { id: school.id }) }
 
-  scope :oldest_first, -> { order(Arel.sql("CASE WHEN induction_records.end_date IS NULL THEN 1 ELSE 0 END"), start_date: :asc, created_at: :asc) }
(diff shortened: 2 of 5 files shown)


app/services/record_declaration.rb
```diff
@@ -156,10 +156,23 @@ def declaration_parameters
...
     }
   end
 
+  def mentor
+    return nil unless participant_profile.ect?
+
+    induction_record = Induction::FindBy.call(
+      participant_profile:,
+      lead_provider: cpd_lead_provider.lead_provider,
+      date_range: (Date.new(0)..declaration_date),
```
peteryates, line 170: not that it matters here but would `..declaration_date` make sense here instead of the zeroed date?

ethax-ross:
The review app database is in a funny state due to changing the migrations under its feet; I'm going to close the PR to destroy the review apps and re-open it to create them fresh.

app/services/record_declaration.rb
```diff
@@ -156,10 +156,23 @@ def declaration_parameters
       declaration_type:,
       cpd_lead_provider:,
       delivery_partner:,
+      mentor_user_id: mentor&.id,
       user: participant_identity.user,
     }
   end
 
+  def mentor
```
cwrw, line 164: suggestion: could we name this `mentor_user` to avoid confusing it with a profile?

app/services/record_declaration.rb
```diff
@@ -156,10 +156,23 @@ def declaration_parameters
...
+    return nil unless participant_profile.ect?
+
+    induction_record = Induction::FindBy.call(
```
cwrw, line 167: having a deeper think about this, we probably need the time range between the start date AND end date of an induction record, if the end date is empty it means that one is the latest (in some cases 😅). Let's chat tomorrow

app/services/record_declaration.rb
ethax-ross: I'll rename to `mentor_user_id` as it looks like I can drop the `.user` access and just grab `user_id` of the participant identity

cwrw (approved):

PR #3534, opened 2023-06-14
Backfill mentor_user_id on ParticipantDeclaration

ethax-ross:

### Context

We have a new `mentor_user_id` attribute on `ParticipantDeclaration` that is currently only being populated for new declarations. We want to backfill all the ECF declarations so that we can start referencing the `mentor_user_id` field.

- Add rake task to backfill mentor_user_id on ParticipantDeclaration

### Guidance to review

I thought about doing a rake task that spawns jobs for each of the update operations, but assuming there aren't _loads_ of declarations and updating them is a relatively short operation then doing it from the dev console should be fine/less complex.

Example usage:

Dry run: `rake backfill:mentor_user_id`
Run (updates records): `rake 'backfill:mentor_user_id[false]'`

### DIFF
--- a/lib/backfill_mentor_user_id.rb
+++ b/lib/backfill_mentor_user_id.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+class BackfillMentorUserId
+  attr_reader :dry_run
+
+  def initialize(dry_run: true)
+    @dry_run = dry_run
+  end
+
+  def run
+    log_start
+    process_declarations
+    log_finish
+  end
+
+private
+
+  def log_start
+    logger.info("~~DRY RUN~~") if dry_run
+    logger.info("Backfilling #{total_declarations} declarations")
+  end
+
+  def process_declarations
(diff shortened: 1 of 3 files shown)


lib/backfill_mentor_user_id.rb
```diff
@@ -0,0 +1,75 @@
...
+  def find_induction_record(declaration)
+    Induction::FindBy.call(
+      participant_profile: declaration.participant_profile,
+      lead_provider: declaration.cpd_lead_provider.lead_provider,
+      date_range: (..declaration.declaration_date),
+    )
+  end
+
+  def extract_mentor_user_id(induction_record)
+    induction_record&.mentor_profile&.participant_identity&.user&.id
```
cwrw, line 62:
```suggestion
    induction_record&.mentor_profile&.participant_identity&.user_id
```

cwrw (approved):
nicely done, ran this locally with backup restore and checked a few and all good! Only thing is that this will take a while to run as discussed

PR #3542, opened 2023-06-14
Fix rolling into next cohort year failures

cwrw:
### Context
When the next cohort rolls in, we get failures in the specs. Fix those not to rely on the current cohort start date
- Ticket: n/a

- Amend existing cohort factory to be in the past and set specific specs to that date
- Amend feature setup to always point at specific cohort to avoid failures, as the cohort year is set up in the query param
- Amend training state scenarios to look into the future rather than the cohort start to avoid failures

### Guidance to review
Need to merge before tomorrow when the cohort year rolls again 😅


peteryates (approved):
Looks good

PR #3545, opened 2023-06-15
Try moving back towards transactional specs

peteryates:
### Context

With the current setup nearly every spec test include `:with_default_schedules` which sets up every cohort from 2020 and adds finance schedules to every cohort - despite that not being reality.

`:with_default_schedules` takes ~1 second, running it ones at the start of the suite and wrapping every test in a database transaction that can be rolled back reduces the total spec suite runtime by 35 minutes!

- Switch DataBaseCleaner to use the transaction strategy
- change with_default_schedules to run in `before(:suite)`
- fix tests that rely on cohorts and schedules not being setup

### Guidance to review

- early on  I have tried to fix other clashes with real world data then decided to stop as there was too much use of real Lead Provider names but in unrealistic scenarios so I stopped
- I could not easily get cypress to run transactionally but as there is still a desire to replace the 14 features with capybara features or just remove them I felt it unnecessary to fix that
- It highlight a lot of spec tests that test what the system does when the years is either 2020, 2021 or 2022 as this scenarios will never happen again another review should take place but understand when the use of those cohort years are actually examples of future scenarios.


spec/features/schools/participants/update_participants_details_spec.rb
```diff
@@ -7,6 +7,7 @@
   include ManageTrainingSteps
 
   before do
+    travel_to Date.new(2022, 9, 1)
```
tonyheadford, line 10: Could use the `travel_to: Date.new(2022, 9, 1)` on the describe or scenario blocks so that it never leaks?

ltello (approved):
Massive performance boot. 
Lovely!!

spec/models/finance/schedule/npq_ehco_spec.rb
```diff
@@ -87,10 +88,10 @@
     context "when selected cohort is before multiple schedules existed for EHCO" do
       let(:cohort) { Cohort.find_by!(start_year: 2021) }
 
-      it "returns NPQ EHCO June schedule" do
-        expected_schedule = described_class.find_by(cohort:, schedule_identifier: "npq-ehco-june")
+      it "returns NPQ EHCO November schedule" do
```
cwrw, line 91: not sure why this needed to change the default should be June?

spec/models/finance/schedule/npq_specialist_spec.rb
```diff
@@ -10,7 +10,7 @@
   it "seeds from csv" do
     schedule = described_class.find_by(schedule_identifier: "npq-specialist-spring")
     expect(schedule).to be_present
-    expect(schedule.milestones.count).to eql(3)
+    expect(schedule.milestones.count).to eql(4)
```
cwrw, line 13: Checking the csv, this value should be 3 milestones rather than 4, worth finding out how we're getting the fourth?

spec/services/importers/create_cohort_spec.rb
```diff
@@ -6,21 +6,20 @@
...
         csv.write "\n"
-        csv.write "2020,2020/05/10,2020/09/01,"
-        csv.write "\n"
-        csv.write "2021,2021/05/10,2021/09/01,"
-        csv.write "\n"
-        csv.write "2022,2022/05/10,2022/09/01,"
-        csv.write "\n"
-        csv.write "2023,2023/05/10,2023/09/01,"
-        csv.write "\n"
+        4.times.each do |n|
```
cwrw, line 18: suggestion: I prefer the explicitness of the old test, where exact cohorts are specified and tested against. Having relative cohorts makes it harder to understand the test and debug.

cwrw: update I guess we can't work the explicit ones in the new world

spec/features/lead_providers/dashboard_page_spec.rb
```diff
@@ -31,6 +31,7 @@
   end
 
   scenario "Confirming schools" do
+    travel_to Date.new(2021, 9, 1)
```
cwrw, line 34: maybe we can make this relative the a registration start date?

spec/models/finance/schedule/npq_specialist_spec.rb
mattonfoot: Pretty sure this is because the CSV is adding to the already set up cohort

spec/services/importers/create_cohort_spec.rb
mattonfoot: Moved to explicitly using 3031-4 and 4041 as the years in the tests

peteryates:
Looks great to me, thanks @mattonfoot 🚀

tonyheadford (approved):
Awesome

PR #3548, opened 2023-06-15
Optimize ParticipantDeclaration queries using mentor_user_id attribute

ethax-ross:
> ⚠️ WIP

### Context

The `ParticipantDeclaration` model now contains a `mentor_user_id` that has been backfilled and is populated
for new declarations. This means we no longer have to query the latest induction record in order to retrieve the
`mentor_id`. 

- Optimize serializer with mentor_user_id
- Optimize ParticipantDeclarationQuery with mentor_user_id


app/services/api/v3/participant_declarations_query.rb
```diff
@@ -36,13 +36,8 @@ def participant_declarations_for_pagination
 
       def participant_declarations_from(paginated_join)
         sub_query = ParticipantDeclaration
-          .select("participant_declarations.*", "COALESCE(jsonb_agg(DISTINCT participant_identities_mentor_profiles.user_id) FILTER (WHERE participant_identities_mentor_profiles.user_id IS NOT NULL), '[]') AS transient_mentor_user_id")
+          .select("participant_declarations.*")
           .joins("INNER JOIN (#{paginated_join.to_sql}) as tmp on tmp.id = participant_declarations.id")
```
cwrw, line 40: not sure we need the subquery anymore as we no longer need to group by participant declaration id, maybe we can move the inner join into one query with the one below?

app/services/api/v3/participant_declarations_query.rb
```diff
@@ -35,25 +35,14 @@ def participant_declarations_for_pagination
       end
 
       def participant_declarations_from(paginated_join)
-        sub_query = ParticipantDeclaration
-          .select("participant_declarations.*", "COALESCE(jsonb_agg(DISTINCT participant_identities_mentor_profiles.user_id) FILTER (WHERE participant_identities_mentor_profiles.user_id IS NOT NULL), '[]') AS transient_mentor_user_id")
-          .joins("INNER JOIN (#{paginated_join.to_sql}) as tmp on tmp.id = participant_declarations.id")
-          .joins(left_outer_join_participant_profiles)
```
cwrw, line 41: could you please remove the extra left join methods as they are no longer used?

cwrw (approved):

PR #3551, opened 2023-06-15
[CST-1624] Ensure safe induction record navigation

Lockyy:
### Context

- Ticket:  https://dfedigital.atlassian.net/browse/CST-1624

In several places there is the assumption of at least one induction record for a participant profile and even when safe navigation is used it is assumed that a nil value is acceptable.

Removes the assumption that one-to-many induction_record relations will always return at least one value and return an empty array if the result would be nil.

Sends fallback text to `govuk_tag` when induction records are missing.

#### Replication steps

1. Visit /admin/participants/:participant_id/school for a participant without an induction record
2. The above page will produce a 500 error

1. Visit /admin/participants/:participant_id/statuses for a participant without an induction record
2. The above page will produce a 500 error

#### Post fix behavior

The above pages render correctly for these participant profiles

### DIFF
--- a/app/controllers/admin/participants_controller.rb
+++ b/app/controllers/admin/participants_controller.rb
@@ -48,7 +48,7 @@ module Admin
     end
 
     def historical_induction_records
-      @historical_induction_records ||= induction_records[1..]
+      @historical_induction_records ||= induction_records[1..].presence || []
     end
 
     def latest_induction_record
--- a/app/controllers/concerns/admin/participants/find_induction_records.rb
+++ b/app/controllers/concerns/admin/participants/find_induction_records.rb
@@ -10,7 +10,7 @@ module Admin
       end
 
       def historical_induction_records
-        induction_records[1..]
+        induction_records[1..].presence || []
       end
 
       def all_induction_records
--- a/app/presenters/admin/participant_presenter.rb
+++ b/app/presenters/admin/participant_presenter.rb
@@ -70,7 +70,7 @@ class Admin::ParticipantPresenter
   end
 
   def historical_induction_records
-    induction_records[1..]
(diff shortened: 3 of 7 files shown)


ltello (approved):
Good!! 
I just left a few optional suggestions.

app/controllers/concerns/admin/participants/find_induction_records.rb
```diff
@@ -10,7 +10,7 @@ def relevant_induction_record
       end
 
       def historical_induction_records
-        induction_records[1..]
+        induction_records[1..].presence || []
```
ltello, line 13:
```suggestion
        induction_records[1..].to_a
```

app/presenters/admin/participant_presenter.rb
```diff
@@ -70,7 +70,7 @@ def appropriate_body_name
   end
 
   def historical_induction_records
-    induction_records[1..]
+    induction_records[1..].presence || []
```
ltello, line 73:
```suggestion
    induction_records[1..].to_a
```

app/controllers/admin/participants_controller.rb
```diff
@@ -48,7 +48,7 @@ def induction_records
     end
 
     def historical_induction_records
-      @historical_induction_records ||= induction_records[1..]
+      @historical_induction_records ||= induction_records[1..].presence || []
```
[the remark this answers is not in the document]
Lockyy: That's a good suggestion, I'm gonna stick with the `.presence || []` pattern here as it's more explicit about why the check is in place. I could see .to_a being accidentally removed as it wouldn't be *as* clear that it's guarding against nil values.

PR #3555, opened 2023-06-16
[CPDLP-2262] Add job to send outcomes to TRA

leandroalemao:
### Context

- Ticket: [CPDLP-2262](https://dfedigital.atlassian.net/browse/CPDLP-2262)

We’d like to also monitor any errors that could stem from TRNs being missing or errors returned from the TRA API.

Create a new sidekiq job to send outcomes to TRA.

[CPDLP-2262]: https://dfedigital.atlassian.net/browse/CPDLP-2262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/jobs/participant_outcomes/batch_send_latest_outcomes_job_spec.rb
```diff
@@ -44,17 +29,17 @@
...
       it "only enqueues the send job for the first records up to the batch_size" do
         described_class.perform_now(batch_size)
 
         expect(ParticipantOutcomes::SendToQualifiedTeachersApiJob).to have_been_enqueued.exactly(:once).with(participant_outcome_id: 1)
         expect(ParticipantOutcomes::SendToQualifiedTeachersApiJob).not_to have_been_enqueued.with(participant_outcome_id: 2)
       end
     end
+
+    it "does not requeue itself" do
+      expect { described_class.perform_now(default_batch_size) }.not_to have_enqueued_job(described_class)
```
ethax-ross, line 41: I don't think this is testing the retry behaviour; it's calling `perform` and checking it doesn't enqueue itself - which it doesn't - but it doesn't check if Sidekiq requeues it on failure?

As the retry behaviour is an implementation detail of Sidekiq we probably don't need to test it 🤔 although asserting the `sidekiq_options` are correct may be worthwhile, but I'm not sure how straight-forward that would be. There is [a gem](https://github.com/philostler/rspec-sidekiq#be_retryable) that can do it, but it may break our other job-related tests.

spec/jobs/participant_outcomes/batch_send_latest_outcomes_job_spec.rb
```diff
@@ -16,23 +15,9 @@
...
-
     context "when there are no more than batch_size records" do
       let(:batch_size) { 2 }
 
-      it "does not requeue itself" do
-        expect { described_class.perform_now(batch_size) }.not_to have_enqueued_job(described_class)
-      end
-
       it "enqueues the send job for each record" do
         described_class.perform_now
```
ethax-ross, line 22: Should this be `described_class.perform_now(batch_size)`?

ethax-ross (commented):
Looking good, just a couple of comments/questions 👍

config/sidekiq_cron_schedule.yml
```diff
@@ -29,3 +29,7 @@ remove_school_mentor_job:
   cron: "45 4 * * *"
   class: "RemoveSchoolMentorJob"
   queue: default
+send_outcomes_to_qualified_teachers_api:
+  cron: "*/10 * * * *"
+  class: "ParticipantOutcomes::BatchSendLatestOutcomesJob"
+  queue: default
```
cwrw, line 35: maybe this should be on `participant_outcomes`?

spec/jobs/participant_outcomes/batch_send_latest_outcomes_job_spec.rb
cwrw: Agree around not testing retry behaviour from sidekiq, also since the implementation of manually requeuing in the job itself has been removed I don't think we need to test that it doesn't either

cwrw (approved):
👍 past the minor comments

cwrw (approved):

PR #3561, opened 2023-06-19
CPDNPQ-1188 - Create endpoint in ecf to synchronize data with npq

javier-npq:
### Context

Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1172

### NPQ counterpart

PR: https://github.com/DFE-Digital/npq-registration/pull/819

- Added a new controller under `npq` namespace to send the required data
- Added the required scopes in respective modals to query the data
- Added a serializer under the `npq` namespace


app/models/participant_declaration/npq.rb
```diff
@@ -70,4 +70,8 @@ def qualification_type
 
     nil
   end
+
+  def latest_outcome_of_declaration
+    outcomes.order(completion_date: :desc)&.first
```
slawosz, line 75: How about `outcomes.order(completion_date: :desc).limit(1)`?

javier-npq: Yea that would also work as the same in this case

slawosz: I would still prefer `limit(1)` as it lower risk, but in this case we wouldn't have many outcomes, for `first` is fine.

app/serializers/api/v1/npq/application_synchronization_serializer.rb
```diff
@@ -0,0 +1,19 @@
...
+module Api
+  module V1
+    module NPQ
+      class ApplicationSynchronizationSerializer
+        include JSONAPI::Serializer
+        include JSONAPI::Serializer::Instrumentation
+
+        attributes :lead_provider_approval_status, :id, :participant_outcome_state
+        attribute(:participant_outcome_state) do |object|
+          object.latest_declaration_of_user&.latest_outcome_of_declaration&.state if object.latest_declaration_of_user.present?
```
slawosz, line 14: Here we have potential of many `N+1` queries.

slawosz (approved):

app/models/participant_declaration/npq.rb
javier-npq: Switched to limit and resolving this comment

app/serializers/api/v1/npq/application_synchronization_serializer.rb
javier-npq: Yes I know. We don't have any direct association to fetch the outcomes. Have to chain a lot of methods to get it. I have eliminated the serializer and now doing all of the operation in a single query/iteration in hope for improving performance. Also added a comment in the code. It's still not the most efficient way but better than this.

PR #3573, opened 2023-06-20
Optimize participant declarations query

ethax-ross:
[Jira-2238](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2238)

### Context

The participant declarations query is slow and, if the endpoint is under load, it can cause the database to bog down. We want to optimise the query so that we can better handle high loads.

- Add activerecord-explain-analyze gem

ActiveRecord can `explain` a query, but its a lot more informative if you can run `EXPLAIN ANALYZE`. This gem adds the functionality to ActiveRecord:

```
query.explain(analyze: true, format: :json)
```

- Optimise participant declarations query

I looked into a number of ways to optimise this query; the main bottleneck is the Postgres sort operation in the query plan, which uses a whopping 46mb of space and results in a disk write as our working memory is only ~4mb.

The high sort memory is due to a `UNION` over a lot of columns. We can reduce the sort overhead by performing the `UNION` over the `id` columns and then re-select the records from the result. This was more efficient, but less efficient than performing a `UNION` over a subset of columns (that are used in subsequent filtering) and using the resulting records directly instead of re-selecting them. Our `ActiveRecordUnion` model was doing both; `UNION` over all columns _and_ re-selecting the results for some reason. I opted to swap out our `ActiveRecordUnion` service as it wasn't tested and there is a gem that performs an optimal `UNION` available. The sort now uses ~12mb of space.

After the sort operation the next slowest operation is a nested loop due to the amount of join statements. To reduce the impact of this I've optimised the joins (hopefully without breaking anything) and only applied them when they're actually needed (such as when filtering by cohort). We can't make up much performance-wise here though.

The final optimisation was to remove the unnecessary `distinct` at the end of the query; we do a `UNION` which internally performs distinct so this isn't necessary. I tried `UNION ALL` (which doesn't do a distinct) and keeping a manual call to `distinct` at the end but this was less efficient.

All of this results in almost ~70% performance improvement; from ~2.3s to ~700ms.

### Guidance to review

I looked into improving the v1/2 queries; they will benefit from using the gem for performing the `UNION`, but we can't make the same optimisations around selecting a subset of columns as we directly serialize the result of the query..

[Original execution plan](http://tatiyants.com/pev/#/plans/plan_1687259214446)

[New execution plan 1](http://tatiyants.com/pev/#/plans/plan_1687341959666)
[New execution plan 2](https://explain.depesz.com/s/n5Al)


app/services/api/v3/participant_declarations_query.rb
```diff
@@ -69,7 +69,7 @@ def declarations_scope
 
       def previous_declarations_scope
         scope = with_joins(ParticipantDeclaration)
-          .where(participant_profile: { induction_records: { induction_programme: { partnerships: { lead_provider: } } } })
+          .where(cpd_lead_provider_id: lead_provider.id)
```
cwrw, line 72: Could we keep the old where clause as this is required to surface declarations that don't belong to the current provider, but an application that has transferred now to them (thus the induction record/lead provider lookup)?

cwrw (approved):
Thank you, this is excellent! All the optimisations for v3 are great, and hopefully we'll move to a better DB tier where it will be even better.

I've tested v1 - v3 with the old implementation and new implementation and they match in terms of numbers of records, so good to move to the gem as there don't seem to be any issues (from the article attached to the deleted implementation)

mooktakim (approved):
Great work 👍

leandroalemao (approved):
well done 👍 😄 👏

PR #3574, opened 2023-06-20
Remove API v3 feature flag

cwrw:
### Context

API v3 has been released, we no longer need the API v3 flag

- Ticket: n/a

- Remove feature flag
- Remove tag to disable api v3 specs
- Remove feature flag from specs

### Guidance to review
🎉

### DIFF
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -17,7 +17,6 @@ class FeatureFlag
   # Short-lived feature flags
   TEMPORARY_FEATURE_FLAGS = %i[
     eligibility_notifications
-    api_v3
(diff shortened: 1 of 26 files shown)


ethax-ross (approved):

leandroalemao (approved):
@cwrw Lgtm 👍 should we also remove the `api_v3` record from `features` table?

cwrw:
good question @leandroalemao, we don't usually purge the flag from the table when removing feature flags historically however your right it might cause issues if someone uses the same feature flag name in the future. I'll make sure to remove those from the db after merged thanks for flagging, as a migration might be overkill

PR #3582, opened 2023-06-22
Add request access link to the start page

frankieroberto:
Conversations with the support team have revealed that users are getting in touch with them to request access, without realising that they can get access by requesting an link be sent to their school email address.

This is possibly because the link to request access was only available on the Sign in page, and wasn't signposted from the service start page.

This PR adds the link to the service start page.

A missing back link is also added to the Sign in page, and the link text made consistent with the new link on the start page.

## Screenshots

| Before | After|
|---|---|
| before | after |
| sign-in-before | sign-in-after |

## Alternatives

Alternatively, we could re-label the "Sign in" button on the first page as "Start", and then keep the link where it is. However on balance I think it's probably clear to highlight both options on the first page.

### DIFF
--- a/app/views/start/index.html.erb
+++ b/app/views/start/index.html.erb
@@ -19,7 +19,7 @@
     <div class="govuk-button-group">
       <%= govuk_button_link_to "Sign in", new_user_session_path %>
       <span class="govuk-!-margin-right-2">or</span>
-      <%= govuk_link_to "request a link to use the service", resend_email_request_nomination_invite_path, class: " govuk-link--no-visited-state" %>
+      <%= govuk_link_to "request access to the service", resend_email_request_nomination_invite_path, class: " govuk-link--no-visited-state" %>
     </div>
   </div>
 
--- a/app/views/start/index.html.erb
+++ b/app/views/start/index.html.erb
@@ -2,9 +2,11 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl">Manage training for early career teachers</h1>
 
-    <p>Use this service to set up ECF-based training for your early career teachers (ECTs) or tell us about a change at your school.</p>
+    <p class="govuk-body">Use this service to set up ECF-based training for your early career teachers (ECTs) or tell us about a change at your school.</p>
 
-    <p>You need to tell us:</p>
+    <h2 class="govuk-heading-m">Set up and manage your training</h2>
+
+    <p class="govuk-body">Tell us:</p>
 
     <ul class="govuk-list govuk-list--bullet">
       <li>how you want to run your training</li>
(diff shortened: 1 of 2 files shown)


frankieroberto:
@Lloyd-K what do you think?

Lloyd-K:
Hi, I definitely agree with adding the 'request a link' guidance to the start page. I think it's sensible to have it on the sign in page too, to cover both start points. I think the content on the start and sign in pages needs a fuller review to make sure it's reflective of the tasks most schools are being asked to complete at this stage. Thanks

app/views/start/index.html.erb
```diff
@@ -2,25 +2,29 @@
...
       <li>which ECTs and mentors will take part</li>
     </ul>
 
-    <p>Your school must complete these steps before your ECTs can start their statutory induction programme.</p>
+    <p class="govuk-body">Your school must complete these steps before your ECTs can start their statutory induction programme.</p>
 
     <h2 class="govuk-heading-m">Tell us about changes at your school</h2>
 
-    <p>Tell us about any changes, for example, nominating a new Induction tutor.</p>
+    <p class="govuk-body">Tell us about any changes, for example, nominating a new Induction tutor.</p>
```
frankieroberto, line 21:
```suggestion
    <p class="govuk-body">Tell us about any changes, for example, nominating a new induction tutor.</p>
```

Lloyd-K (approved):

app/views/users/sessions/new.html.erb
```diff
@@ -14,7 +14,7 @@
             <%= f.govuk_submit "Sign in" %>
         <% end %>
 
-        <p class="govuk-body"><%= govuk_link_to "Request a link to sign in or change the induction tutor", resend_email_request_nomination_invite_path, class: "govuk-link govuk-link--no-visited-state" %></p>
-
+        <p class="govuk-body">Or <%= govuk_link_to "request access to the service", resend_email_request_nomination_invite_path, class: " govuk-link--no-visited-state" %>
```
tonyheadford, line 17:
```suggestion
        <p class="govuk-body">Or <%= govuk_link_to "request access to the service", resend_email_request_nomination_invite_path, no_visited_state: true %>
```

app/views/start/index.html.erb
```diff
@@ -2,25 +2,29 @@
...
     <h2 class="govuk-heading-m">Tell us about changes at your school</h2>
 
-    <p>Tell us about any changes, for example, nominating a new Induction tutor.</p>
+    <p class="govuk-body">Tell us about any changes, for example, nominating a new induction tutor.</p>
 
-    <%= govuk_start_button text: "Sign in", href: new_user_session_path %>
+    <div class="govuk-button-group govuk-!-margin-top-7">
+      <%= govuk_button_link_to "Sign in", new_user_session_path %>
+      <span class="govuk-!-margin-right-2">or</span>
+      <%= govuk_link_to "request access to the service", resend_email_request_nomination_invite_path, class: " govuk-link--no-visited-state" %>
```
tonyheadford, line 26:
```suggestion
      <%= govuk_link_to "request access to the service", resend_email_request_nomination_invite_path, no_visited_state: true %>
```

tonyheadford (commented):
A couple of suggestions, we've been updating these to leverage the component more as we come across them.

tonyheadford (approved):

PR #3583, opened 2023-06-22
Optimize NPQApplicationsQuery

ethax-ross:
[Jira-2238](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2238)

### Context

The query to retrieve `NPQApplications` is slow; we want to optimize the query so that the endpoint is more responsive.

- Optimize NPQApplicationsQuery

### Guidance to review

The query to retrieve `NPQApplications` is slow for two reasons:

1. When calculating `previously_funded?` for each record we retrieve all previously funded applications, exclude the current application and then check if there are any.
 * Speeding up this is straight-forward; we can bring the exclusion logic into the query and call `exists?` so we're only returning a boolean and not all the previous records. 
2. For each record we need to query the previously funded applications.
 * Fixing this is not so straightforward. To speed this query up we need to find the previously funded applications as part of the query to retrieve all applications. Most of the sub-query to do this is straight-forward; the only difficulty is filtering by 'rebranded alternative' courses. Each course has a set of these, as determined by `NPQCourse#rebranded_alternative_courses`. To be able to dynamically filter on these across all records we first build a JSON map of course ID -> alternative course IDs. We can then use this in a subquery to check each application and determine if it has been previously funded. The result is stored on a `transient_previously_funded` attribute of the model, which can then be used to optimise the `previously_funded?` method of the application during serialisation.

These optimisations reduce the endpoint response time from ~3.8s to ~650ms in my (fairly limited) testing.

The approach is similar to how we previosuly optimised the `mentor_id` on participant declarations. The downside is we're duplicating the `previously_funded?` logic in SQL; I can't see a way around this and I've added comments for clarity.


spec/models/npq_application_spec.rb
```diff
@@ -80,6 +80,34 @@
       end
     end
 
+    context "when transient_previously_funded is declared on the model" do
```
mooktakim, line 83: We should have one test for `#eligible_for_dfe_funding` and `#ineligible_for_funding_reason` each where by default it does not respond_to? `transient_previously_funded`

mooktakim (changes requested):

spec/models/npq_application_spec.rb
ethax-ross: That case is already covered by the existing tests?

cwrw (approved):
⭐ nicely done, tested locally and query massively improves and values are exactly the same across both versions

mooktakim (approved):

PR #3585, opened 2023-06-22
Add Redis in AKS

peteryates:
Add Redis in the AKS infrastructure

### DIFF
--- a/terraform/aks/databases.tf
+++ b/terraform/aks/databases.tf
@@ -1,0 +1,15 @@
+module "redis" {
+  source = "git::https://github.com/DFE-Digital/terraform-modules.git//aks/redis?ref=testing"
+
+  namespace             = var.namespace
+  environment           = local.environment
+  azure_resource_prefix = var.azure_resource_prefix
+  service_name          = local.service_name
+  service_short         = var.service_short
+  config_short          = var.config_short
+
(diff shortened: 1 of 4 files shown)


terraform/aks/databases.tf
```diff
@@ -0,0 +1,15 @@
+module "redis" {
+  source = "git::https://github.com/DFE-Digital/terraform-modules.git//aks/redis?ref=testing"
+
+  namespace             = var.namespace
```
cwrw, line 4: do we need to define this in the `variable.aks`?

terraform/aks/databases.tf
```diff
@@ -0,0 +1,15 @@
...
+  environment           = local.environment
+  azure_resource_prefix = var.azure_resource_prefix
+  service_name          = local.service_name
```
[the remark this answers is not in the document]
peteryates: I think so, it looks like they're used by the redis module we're using

https://github.com/DFE-Digital/terraform-modules/blob/main/aks/redis/resources.tf

terraform/aks/databases.tf
peteryates: Yeah, that's later in the chain 🤦🏽‍♂️ I'm all for just dropping these PRs in favour of #3587

cwrw (approved):

PR #3586, opened 2023-06-22
Add PostgreSQL in AKS

peteryates:
Follows on from #3585

### DIFF
--- a/terraform/aks/databases.tf
+++ b/terraform/aks/databases.tf
@@ -1,0 +1,15 @@
+module "redis" {
+  source = "git::https://github.com/DFE-Digital/terraform-modules.git//aks/redis?ref=testing"
+
+  namespace             = var.namespace
+  environment           = local.environment
+  azure_resource_prefix = var.azure_resource_prefix
+  service_name          = local.service_name
+  service_short         = var.service_short
+  config_short          = var.config_short
+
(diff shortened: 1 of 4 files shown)


terraform/aks/databases.tf
```diff
@@ -0,0 +1,32 @@
...
+  azure_resource_prefix = var.azure_resource_prefix
+  service_name          = local.service_name
+  service_short         = var.service_short
+  config_short          = var.config_short
+
+  cluster_configuration_map = module.cluster_data.configuration_map
+
+  use_azure               = var.deploy_azure_backing_services
+  azure_enable_monitoring = var.enable_monitoring
+  azure_extensions        = ["citext", "pg_stat_statements", "pgcrypto", "plpgsql"]
```
cwrw, line 31: we have `fuzzystrmatch` in our PaaS config not sure if we need it though 🤔

cwrw (approved):

terraform/aks/databases.tf
peteryates: I matched this to [what's in schema.rb](https://github.com/DFE-Digital/early-careers-framework/blob/main/db/schema.rb#L16) - perhaps it's something we used in the past?

cwrw: yea prob even a copy paste 👍

PR #3386, opened 2023-05-12
Rails 7.0.4.2

peteryates:
Another attempt at Rails 7, following on from #3132 when the review app started to go a bit funny.

PR #3590, opened 2023-06-23
Rails 7.0.5.1

leoapost:
### Context

Hopefully the last rails 7 PR, following on from https://github.com/DFE-Digital/early-careers-framework/pull/3386.

- Rails upgrade to 7.0.5.1
- Tests fixes
- Refactor preloading associations in v3 api
- Update ransackable models to improve security
- Replace the deprecated `to_s(:format)` with the `#to_s(:format)`
- Remove Spring
- Turn `config.cache_classes` to `true` on test env

### Review
Run the test suite on your local


spec/cypress/integration/admin/ParticipantManagment.feature
```diff
@@ -13,7 +13,7 @@ Feature: Admin user managing participants
 
     When I type "example" into "search box"
     And I press enter in "search box"
-    Then the table should have 5 rows
+    Then the table should have 0 rows
```
peteryates, line 16: This concerns me a bit, do we know why it changed?

leoapost: @peteryates because there are 0 results for the "example" keyword. There is a similar test right beneath it with a keyword that actually returns a results.

peteryates: Ah, well deduced! Thank you @leoapost

peteryates (approved):

cwrw (approved):
Nice work all! I have one question around unnesting ECF/NPQ models, and if we need to do it for schedules as it's the only one remaining that's using the old patterns?

PR #3592, opened 2023-06-23
Add overview admin page

frankieroberto:
This adds an Overview page to the admin navigation, which contains some basic stats about the numbers of schools who have registered for the upcoming academic year.

## Screenshot

Screenshot 2023-06-23 at 16 37 32

### DIFF
--- a/app/components/primary_nav_component.rb
+++ b/app/components/primary_nav_component.rb
@@ -13,13 +13,14 @@ class PrimaryNavComponent < ViewComponent::Base
   class NavItemComponent < ViewComponent::Base
     attr_reader :path
 
-    def initialize(path:, selected: false)
+    def initialize(path:, selected: nil)
       @path = path
       @selected = selected
     end
 
     def current_section?(path)
-      @selected || request.path.include?(path)
+      return request.path.include?(path) if @selected.nil?
+      @selected
     end
   end
 end
--- a/app/components/primary_nav_component.rb
+++ b/app/components/primary_nav_component.rb
@@ -2,8 +2,12 @@
 
 class PrimaryNavComponent < ViewComponent::Base
   include ViewComponent::SlotableV2
-
+  attr_reader :wide
   renders_many :nav_items, "NavItemComponent"
+
+  def initialize(wide: false)
+    @wide = wide
+  end
 
   class NavItemComponent < ViewComponent::Base
     attr_reader :path
--- a/app/components/primary_nav_component.rb
+++ b/app/components/primary_nav_component.rb
@@ -3,6 +3,7 @@
 class PrimaryNavComponent < ViewComponent::Base
   include ViewComponent::SlotableV2
   attr_reader :wide
+
   renders_many :nav_items, "NavItemComponent"
 
   def initialize(wide: false)
--- a/app/controllers/admin/home_controller.rb
+++ b/app/controllers/admin/home_controller.rb
@@ -5,7 +5,45 @@ module Admin
     skip_after_action :verify_authorized, only: :show
     skip_after_action :verify_policy_scoped, only: :show
 
-    # this isn't used yet
-    def show; end
+    # show 2023 pilot stats
+    def show
+      @pilot_stats = get_pilot_stats
+    end
+
+  private
+
+    def get_pilot_stats
+      choices = programme_choices
+
+      OpenStruct.new({
+        cip_total: choices.fetch("core_induction_programme", 0),
+        diy_total: choices.fetch("design_our_own", 0),
+        fip_total: choices.fetch("full_induction_programme", 0),
+        no_ects_total: choices.fetch("no_early_career_teachers", 0),
+        total: choices.values.sum,
+        partnership_totals:,
+      })
+    end
+
+    def programme_choices
+      SchoolCohort.where(cohort:).group(:induction_programme_choice).count
+        .reject { |choice| choice.in? %w[school_funded_fip not_yet_known] }
+    end
+
+    def partnership_totals
+      totals = Partnership.unchallenged.where(cohort:).group(:lead_provider_id).count
+
+      providers.map do |provider|
+        { name: provider.name, total: totals.fetch(provider.id, 0) }
+      end
+    end
+
+    def providers
+      LeadProvider.where(id: ProviderRelationship.where(cohort:).select(:lead_provider_id)).order(:name)
+    end
+
+    def cohort
+      @cohort ||= Cohort.find_by(start_year: 2023)
+    end
   end
 end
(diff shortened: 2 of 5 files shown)


tonyheadford (approved):
LGTM but should really have someone else approve

app/controllers/admin/home_controller.rb
```diff
@@ -5,7 +5,45 @@ class HomeController < Admin::BaseController
...
+      totals = Partnership.unchallenged.where(cohort:).group(:lead_provider_id).count
+
+      providers.map do |provider|
+        { name: provider.name, total: totals.fetch(provider.id, 0) }
+      end
+    end
+
+    def providers
+      LeadProvider.where(id: ProviderRelationship.where(cohort:).select(:lead_provider_id)).order(:name)
+    end
```
ltello, line 43:
```suggestion
    def get_pilot_stats
      OpenStruct.new(
        cip_total: program_choices.fetch("core_induction_programme", 0),
        diy_total: program_choices.fetch("design_our_own", 0),
        fip_total: program_choices.fetch("full_induction_programme", 0),
        no_ects_total: program_choices.fetch("no_early_career_teachers", 0),
        total: program_choices.values.sum,
        partnership_totals:,
      )
    end
    
    def programme_choices
      @programme_choices ||= SchoolCohort.where(cohort:)
                                         .group(:induction_programme_choice)
                                         .count
                                         .except("school_funded_fip", "not_yet_known")
    end
    
    def partnership_totals
      Partnership
        .unchallenged
        .joins(:lead_provider)
        .where(cohort:)
        .order("lead_providers.name")
        .group("lead_providers.name")
        .count
    end
```
    def get_pilot_stats
      choices = programme_choices

      OpenStruct.new({
        cip_total: choices.fetch("core_induction_programme", 0),
        diy_total: choices.fetch("design_our_own", 0),
        fip_total: choices.fetch("full_induction_programme", 0),
        no_ects_total: choices.fetch("no_early_career_teachers", 0),
        total: choices.values.sum,
        partnership_totals:,
      })
    end

    def programme_choices
      SchoolCohort.where(cohort:).group(:induction_programme_choice).count
        .reject { |choice| choice.in? %w[school_funded_fip not_yet_known] }
    end

    def partnership_totals
      totals = Partnership.unchallenged.where(cohort:).group(:lead_provider_id).count

      providers.map do |provider|
        { name: provider.name, total: totals.fetch(provider.id, 0) }
      end
    end

    def providers
      LeadProvider.where(id: ProviderRelationship.where(cohort:).select(:lead_provider_id)).order(:name)
    end
```

Comment (ltello) [reply app/controllers/admin/home_controller.rb:43]: Do not commit if you don't like it.

File: app/views/admin/home/show.html.erb [hunk 2]
```diff
@@ -1,3 +1,41 @@
...
+    </tr>
+    <tr class="govuk-table__row">
+      <th scope="row" class="govuk-table__header govuk-!-font-weight-regular">Designing their own programme</th>
+      <td class="govuk-table__cell govuk-table__cell--numeric"><%= @pilot_stats.diy_total %></td>
+    </tr>
+    <tr class="govuk-table__row">
+      <th scope="row" class="govuk-table__header govuk-!-font-weight-regular">Using a training provider (full induction programme)</th>
+      <td class="govuk-table__cell govuk-table__cell--numeric govuk-!-font-weight-regular"><%= @pilot_stats.fip_total %></td>
+    </tr>
+    <% @pilot_stats.partnership_totals.each do |provider| %>
```
ltello, line 24: If you liked my suggestion above, this would need to be changed as well to something like this:
`<% @pilot_stats.partnership_totals.each do |(provider_name, provider_total)| %>`

ltello:
> @tonyheadford @frankieroberto I have reservations about this overview page.
> 
> Here's some concerns:
> 
> 1. It looks like it needs some more thought from a product point of view. Merging it as would just feel awkward and without a real purpose
> 2. Doesn't look like to me that such a page would add much value to the target audience of the Admin pages
> 3. Down the road, it will open a can of worms in terms of complexity and performance
> 4. I would prefer to use specialised tools for analysing data and displaying metrics. Especially for metrics with short term relevancy like the number of the registered schools in the upcoming academic year

I like to have these basic stats Leo.
At least until Sept.
If it creates a performance issue that's a different story.

ltello:
BTW Number of participants registered for 2023 cohort?

app/controllers/admin/home_controller.rb
tonyheadford: Thanks, not a massive fan of this in this case, I deliberately kept this simple and this is only a temporary page to enable visibility of the stats that I pull out each day.

frankieroberto:
> > BTW Number of participants registered for 2023 cohort? @tonyheadford usually includes that number in his daily stats report on Slack
> 
> Yes this is a good shout 👍, what do you think @frankieroberto ?

Yep! Let’s do that as a follow-up PR though.

app/controllers/admin/home_controller.rb
ltello: I'm not sure I get your comment @tonyheadford 
I think my solution is simpler than the existing one.

tonyheadford: Oh sorry @ltello I was rushing on Friday and I missed what you actually changed and thought you were just reformatting the whole file. By simple I meant that the code was very simple and, in my head at least, easy to understand, it's just lifted from the console queries I run. I didn't communicate that this was a quick and dirty temporary page when I asked for a sanity check, so my bad, but it's good that you picked up on I could improve the queries, thanks.

tonyheadford: Also my original version was designed to allow us to show which providers _haven't_ partnered yet which I think is missed in your version.

app/components/primary_nav_component.rb
```diff
@@ -2,19 +2,26 @@
...
 
+  def initialize(wide: false)
+    @wide = wide
+  end
+
   class NavItemComponent < ViewComponent::Base
     attr_reader :path
 
-    def initialize(path:, selected: false)
+    def initialize(path:, selected: nil)
```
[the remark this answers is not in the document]
ltello: Ok, I get what you are trying to do here. Makes sense. Thanks

app/controllers/admin/home_controller.rb
ltello: Well, if you think it adds value to display the rest of lead providers with 0 count then we can leave everything as it was.

tonyheadford: Yes, definitely, I know Kate is interested in why there's one hasn't entered any and I think given in prod there are very few providers (unlike the loads in the dev env) it makes it clear to see all.  Plus it was in the mockup @frankieroberto built.

frankieroberto: Yep, I’d definitely like to see all providers listed, even if they have 0 schools signed up. 👍

frankieroberto: @tonyheadford @leoapost are we happy to resolve this comment?

ltello (approved):

PR #3593, opened 2023-06-26
Cst 1604 update transfer out content 3

lazydevorg:
### Context

- Ticket: CST-1604

Update the transfer-out journey with new content

### DIFF
--- a/app/views/schools/transfer_out/check_transfer.html.erb
+++ b/app/views/schools/transfer_out/check_transfer.html.erb
@@ -11,12 +11,9 @@
 
     <p class="govuk-body">Confirm if this participant is moving to another school to continue their ECF-based training.</p>
 
-    <div class="govuk-inset-text">
-      <p class="govuk-body">
-        If they are leaving your school for any other reason, do not confirm.<br/>
-        Contact your training provider instead.
-      </p>
-    </div>
+    <p class="govuk-body">
+      If they are leaving your school for any other reason, contact your training provider instead.
+    </p>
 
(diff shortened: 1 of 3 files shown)


app/views/schools/transfer_out/check_transfer.html.erb
```diff
@@ -9,16 +9,15 @@
...
-      continue their induction, confirm below.</p>
+    <p class="govuk-body">Confirm if this participant is moving to another school to continue their ECF-based training.</p>
 
-    <%= govuk_button_link_to "Continue",  teacher_end_date_school_participant_transfer_out_path %>
+    <div class="govuk-inset-text">
+      <p class="govuk-body">
+        If they are leaving your school for any other reason, do not confirm.<br/>
+        Contact your training provider instead.
+      </p>
+    </div>
```
frankieroberto, line 19: I don’t think we should use inset text here, as this is just a continuation of the guidance. It might seem surprising, but in all the usability tests I've observed, inset text often gets ignored more than regular text. 🤦 

We should also avoid single line breaks too.

```suggestion
    <p class="govuk-body">
      If they are leaving your school for any other reason, contact your training provider instead.
    </p>
```

lazydevorg: @frankieroberto yes I agree with your suggestion however, this is required by the design.

frankieroberto (approved):

frankieroberto:
@lazydevorg let's get this merged? 🚀

PR #3600, opened 2023-06-28
CST-1681: restrict changes to induction records of participants not y…

ltello:
…et/anymore in a school

### Context
When a participant is no longer at a school, the SIT might still see them as “no longer training…” but should not be allowed to change their details: name, school email, mentor or AB.

Currently the service allow this to happen.
Not only it feels wrong but also, it is one of the causes we have a category of oddly-shaped induction records.

[Ticket](https://dfedigital.atlassian.net/browse/CST-1681)

Add authorisation restrictions to the links to change data from a participant that have left the school or  not yet in it (transferring in).

### Guidance to review
In the participant view of the dashboard check that the SIT:
- can only see the "Change" link next to the name of the participant if the ECT/Mentor is already/still in the school.
- can only see the "Change" link next to the email of the participant if the ECT/Mentor is in the school or is being transferred in.
- can only see the "Change" link next to the mentor of the participant if the ECT/Mentor is in the school or is being transferred in.
- can only see the "Change" link next to the appropriate body of the participant if the ECT/Mentor is in the school or is being transferred in.

### DIFF
--- a/app/controllers/schools/participants_controller.rb
+++ b/app/controllers/schools/participants_controller.rb
@@ -7,6 +7,14 @@ class Schools::ParticipantsController < Schools::BaseController
   before_action :set_participant, except: %i[index email_used]
   before_action :build_mentor_form, only: :edit_mentor
   before_action :set_mentors_added, only: %i[index show]
+  before_action :authorize_induction_record, only: %i[edit_name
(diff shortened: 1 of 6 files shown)


tonyheadford (approved):
Nice!

PR #3604, opened 2023-06-29
Prevent cohort change for NPQ providers

ethax-ross:
[Jira-2095](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2095)

### Context

We currently allow NPQ providers to change cohort. Going forward we want to restrict this action so that they can only change cohort if:

- They are changing to an earlier cohort year
- There are no submitted/eigible/payable/paid declarations

- Prevent cohort change for NPQ providers

### DIFF
--- a/app/services/change_schedule.rb
+++ b/app/services/change_schedule.rb
@@ -20,7 +20,8 @@ class ChangeSchedule
   validate :validate_new_schedule_valid_with_existing_declarations
   validate :change_with_a_different_schedule
   validate :validate_permitted_schedule_for_course
-  validate :validate_cannot_change_cohort
+  validate :validate_cannot_change_cohort_ecf
+  validate :validate_cannot_change_cohort_npq
 
   def call
     return if invalid?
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -42,6 +42,7 @@ en:
   invalid_transition: "Invalid action"
   invalid_page_parameters: "The '#/page[page]' and '#/page[per_page]' parameter values must be a valid number (equal and more than 1)"
   cannot_change_cohort: "The property '#/cohort' cannot be changed"
+  cannot_change_to_future_cohort: "The property '#/cohort' cannot be changed to a later cohort"
```
mooktakim, line 45: I think this would be better if we change "later cohort" to "future cohort"

app/services/change_schedule.rb
```diff
@@ -133,6 +132,12 @@ def validate_new_schedule_valid_with_existing_declarations
     end
   end
 
+  def applicable_declarations
+    @applicable_declarations ||= participant_profile.participant_declarations.select do |declaration|
+      declaration.state.in?(%w[submitted eligible payable paid])
+    end
```
mooktakim, line 138: We should do a `participant_declarations.where(state: %w[submitted eligible payable paid])` here to avoid loading all the declarations unnecessarily.

mooktakim (changes requested):
A couple of changes, but otherwise is looks good 👍

mooktakim (approved):
Looks good 👍

PR #3605, opened 2023-06-29
Remove '<' from Manage your schools breadcrumb

peteryates:
### Context

The '>' is added by the component and not needed in the markup.


frankieroberto:
Found the original design, which actually suggested using a Back link (but renamed) with the left arrow:

image

However I think the breadcrumb is more appropriate here.

frankieroberto (approved):

PR #3613, opened 2023-06-30
DRY up v3 query classes

ethax-ross:
[Jira-2302](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2302)

### Context

The v3 query classes duplicate some common logic around filtering. We want to remove the duplication to make the code easier to maintain.

- DRY up v3 query classes

### DIFF
--- a/app/services/api/v3/concerns/filter_cohorts.rb
+++ b/app/services/api/v3/concerns/filter_cohorts.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Api::V3::Concerns::FilterCohorts
+  extend ActiveSupport::Concern
+
+protected
+
+  def filter
+    params[:filter] ||= {}
+  end
+
+  def cohort_filter
+    filter[:cohort].to_s
+  end
+
+  def cohorts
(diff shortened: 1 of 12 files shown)


cwrw (commented):
looking good, thanks for doing this. I was wondering as a different approach, if we could use concerns instead, and separate the types of filters, and add them as needed to the services (one for cohorts and one for `updated_since`), to add only the filter that's required to the service. Let me know you're thoughts

ethax-ross:
> looking good, thanks for doing this. I was wondering as a different approach, if we could use concerns instead, and separate the types of filters, and add them as needed to the services (one for cohorts and one for `updated_since`), to add only the filter that's required to the service. Let me know you're thoughts

I actually started out implementing them as concerns but I switched to a base class because I was finding I would have to duplicate parts across the concerns (such as the parameter accessing and time formatting for date filtering). I think you're right though concerns do make more sense here over inheritance; I'll have another crack at doing it that way.

cwrw (approved):
🙌 nicely done, great to see this consolidated! Just the one comment

PR #3618, opened 2023-07-03
Fix call to logger.warning, should be logger.warn

peteryates:


peteryates:
Fixes already in #3616

PR #3624, opened 2023-07-03
Check all users for email clashes when updating SIT email addresses

peteryates:
### Context

We've seen [an error in production](https://dfe-teacher-services.sentry.io/issues/4157545020/?project=5748989&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h&stream_index=1) where updating a induction coordinator's email address fails if the new address has been used by someone who has done an NPQ.

The error isn't detected by `NominateInductionTutorForm` because it just looks at emails used by ECF participants. We store them all in the same table so this means we get clashes regardless.

Now we check if any user has the new email address and show an error message (rather than a 422) if they do. This doesn't really address the underlying problem but it does at least inform us what the problem is.


tonyheadford (approved):

frankieroberto:
I think ideally we'd have a more helpful error message - but that could be a follow-up PR, given this is currently throwing errors in production.

tonyheadford (approved):

PR #3628, opened 2023-07-04
Deduplicate single user

slawosz:
### Context

- Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDNPQ-1181

Part 1 of the dedup proposal - only for single user, but it would allow us to deduplicate currently affected users. The part 2 will be rake task to deduplicate user a nightly basis.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1051,6 +1051,8 @@ ActiveRecord::Schema.define(version: 2023_07_05_152210) do
     t.datetime "updated_at", precision: 6, null: false
     t.datetime "discarded_at"
     t.string "get_an_identity_id"
+    t.string "archived_email"
+    t.datetime "archived_at"
     t.index ["discarded_at"], name: "index_users_on_discarded_at"
     t.index ["email"], name: "index_users_on_email", unique: true
     t.index ["get_an_identity_id"], name: "index_users_on_get_an_identity_id", unique: true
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -1051,14 +1051,15 @@
...
+    t.string "archived_email"
+    t.datetime "archived_at"
     t.index ["discarded_at"], name: "index_users_on_discarded_at"
     t.index ["email"], name: "index_users_on_email", unique: true
     t.index ["get_an_identity_id"], name: "index_users_on_get_an_identity_id", unique: true
   end
 
   create_table "versions", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
-    t.string "item_type"
-    t.string "{:null=>false}"
```
peteryates, line 1061: 😬  not sure how that crept in

slawosz: I guess I am using different postgres version. Will fix it.

tonyheadford (approved):
Agree with @peteryates points but this is great, thanks.

db/schema.rb
peteryates: I've put in a PR to fix it [here](https://github.com/DFE-Digital/early-careers-framework/pull/3639). Thanks for highlighting it!

PR #3634, opened 2023-07-05
[CPDLP-1928] Ensure NPQ users cannot defer an NPQ participant that has no declarations

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-1928

* Added validation to `DeferParticipant` service.
* Added validation to `Finance::NPQ::ChangeTrainingStatusForm`
* Updated release notes.

### DIFF
--- a/app/services/defer_participant.rb
+++ b/app/services/defer_participant.rb
@@ -25,6 +25,7 @@ class DeferParticipant
             }
   validate :not_already_deferred
   validate :not_already_withdrawn
+  validate :do_not_defer_if_without_declarations
 
   def call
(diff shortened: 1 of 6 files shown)


app/services/defer_participant.rb
```diff
@@ -60,6 +61,15 @@ def schedule
 
 private
 
+  def do_not_defer_if_without_declarations
```
cwrw, line 64: maybe we can add this to a shared validator to avoid duplication?

cwrw (approved):

PR #3644, opened 2023-07-05
Check if accidental column exists before removing

cwrw:
### Context
This migration is failing in envs the column does not exist in. Add exists true to ensure migration passes if column doesn't exist

- Ticket: n/a

Check if column exists before running remove column migration


leandroalemao (approved):
lgtm 👍😀

PR #3657, opened 2023-07-06
[CST-1731] - Prevent error in Notify callback when association object is now nil

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-1731)

Prevent error being raised when the association object is `nil`. This could happen if a school changed their SIT between the mail being queued and receiving the Notify callback and the Email was associated with the `SchoolInductionCoordinatorProfile` for example.


cwrw (approved):
🙌

PR #3666, opened 2023-07-10
Change the GIAS nomination link in the launch comms

leoapost:
The GIAS contact should is now sent to the `Choose how to continue` page, where they can opt out, without having to nominate an SIT.

### DIFF
--- a/app/mailers/school_mailer.rb
+++ b/app/mailers/school_mailer.rb
@@ -607,7 +607,7 @@ class SchoolMailer < ApplicationMailer
   def pilot_chase_gias_contact_to_report_school_training_details
     school = params[:school]
     gias_contact_email = params[:gias_contact_email]
-    nomination_link = params[:nomination_link]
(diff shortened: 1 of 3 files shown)


ltello (commented):
Should we change the names (method, placeholder in the email, etc) to something closer to "do your school expects ECTs...?" ? 

Otherwise, looks confusing because the GIAS person might not nominate a SIT?

leoapost:
> Should we change the names (method, placeholder in the email, etc) to something closer to "do your school expects ECTs...?" ?
> 
> Otherwise, looks confusing because the GIAS person might not nominate a SIT? And actually the link is different to the one pointed in the email to SITs

Done. I've also updated the Notify templates.

ltello:
So the g.csv and .deliver_now is because you want to have it fully ready to call the rake task in prod?

ltello (approved):

PR #3669, opened 2023-07-10
[CST-1206] Treat school based roles as one role to avoid seeing options at sign in

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-1206)

School users (SITs) can be confused by the options presented to them if they have more than 1 type of profile e.g. SIT/mentor or SIT who's also doing a NPQ

This PR tries to ensure school users ie. SITs, are never presented with these options in general.

Treat "school" roles more logically:
- A SIT with a mentor profile will be treated as a single role SIT
- A SIT with a `TeacherProfile` will be treated as a single role SIT

### DIFF
--- a/app/forms/choose_role_form.rb
+++ b/app/forms/choose_role_form.rb
@@ -83,9 +83,9 @@ private
   end
 
   def teacher_role?
-    # in order to show the correct page we need to use the presence of teacher_profile
-    # because a withdrawn participant will not have any roles
-    user.teacher_profile.present?
+    # in order to show the correct page we need to participant profiles
+    # because a withdrawn participant will not have any roles in user_roles
+    user.participant_profiles.ecf.any?
   end
 
   def role_values
--- a/app/forms/choose_role_form.rb
+++ b/app/forms/choose_role_form.rb
@@ -10,7 +10,6 @@ class ChooseRoleForm
     "admin" => "DfE admin",
     "finance" => "DfE Finance",
     "induction_coordinator" => "Induction tutor",
-    "induction_coordinator_and_mentor" => "Induction tutor and mentor",
     "teacher" => "Teacher",
   }.freeze
 
@@ -20,7 +19,7 @@ class ChooseRoleForm
   validates :role, inclusion: { in: :role_values }
 
   def role_options
-    USER_ROLES.slice(*user.user_roles)
+    USER_ROLES.slice(*sanitized_user_roles)
   end
 
   def redirect_path(helpers:)
@@ -35,8 +34,6 @@ class ChooseRoleForm
       helpers.delivery_partners_path
     when "appropriate_body"
       helpers.appropriate_bodies_path
-    when "induction_coordinator_and_mentor"
-      helpers.induction_coordinator_mentor_path(user)
     when "induction_coordinator"
       helpers.induction_coordinator_dashboard_path(user)
     when "teacher"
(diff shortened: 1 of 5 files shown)


frankieroberto:
@tonyheadford thanks! I'm not sure how to test this in the review app (or if it even needs testing)?

Does this mean that if someone is both an induction tutor and a mentor, that they skip the "What role do you want to view?" screen?

tonyheadford:
> @tonyheadford thanks! I'm not sure how to test this in the review app (or if it even needs testing)?
> 
> Does this mean that if someone is both an induction tutor and a mentor, that they skip the "What role do you want to view?" screen?

Yes @frankieroberto , if a school user has multiple "school" profiles ECT/mentor/NPQ/SIT they should no longer see the menu

app/forms/choose_role_form.rb
```diff
@@ -47,21 +44,44 @@ def redirect_path(helpers:)
...
+  def sanitized_user_roles
+    roles = user_roles.reject { |role| role.in? %w[induction_coordinator induction_coordinator_and_mentor teacher] }
+
+    if sit_role?
+      roles << "induction_coordinator"
+    elsif teacher_role?
+      roles << "teacher"
+    end
+    roles
+  end
```
mooktakim, line 71: It might be better to move this to `User.user_roles`
And since we no longer have `induction_coordinator_and_mentor` role, should we look to remove the `induction_coordinator_mentor_path(user)` section? no one would be able to visit those pages.

mooktakim (changes requested):

app/forms/choose_role_form.rb
tonyheadford: So I think those are probably legit roles just that we don't want to treat them separately here.  I don't know if anyone else uses `user_roles` for anything?  Really the `user_roles` is a duplicate/adaption of `user_description` could be completely handled here instead?  What do you think?

tonyheadford: Good spot about the paths will look at those

mooktakim: The method `User.user_roles` needs to be updated in how we calculated the roles for a user.
It might be better to split up `mentor` into its own role. That way we can still assign it to a user that requires it.

`user_description` is not related to `user_roles`, we don't really use it anywhere I don't think other than `AnalyticsDataLayer`.

mooktakim: The roles are used to allow users to visit different sections of the app and are used by Pundit for authorisation.

We assign the roles by creating the profile associations, eg: `AdminProfile`, `FinanceProfile`, `InductionCoordinatorProfile` etc

app/forms/choose_role_form.rb
```diff
@@ -47,21 +44,50 @@ def redirect_path(helpers:)
...
+  def sanitized_user_roles
+    roles = user_roles.reject { |role| rejected_roles.include?(role) }
+
+    if sit_role?
+      roles << "induction_coordinator"
+    elsif teacher_role?
+      roles << "teacher"
+    end
+    roles
+  end
```
mooktakim, line 75: Can we move this to `User.user_roles`? that way it can be the source of truth.
We won't need to sanitize in this form as `USER_ROLES` already filters out roles we don't care about.

mooktakim (changes requested):

app/forms/choose_role_form.rb
tonyheadford: But _we_ care about them?  If we remove them at source then there is no point in adding them in the first place. A bit confused I thought we'd gone over this already?

mooktakim (approved):
Looks good 👍

cwrw (approved):

PR #3672, opened 2023-07-11
Add rake task to change milestone dates

ethax-ross:
[Jira-1478](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1478)

### Context

We need to amend the milestone `start_date` and `milestone_date` of various milestones that currently have incorrect dates in our database.

If the milestones do not have any associated participant declarations then we can do this with ease. If, however, there are declarations associated with the milestone we need to ensure that the date changes will not cause any of those declarations to become invalid (as the `declaration_date` should remain between the milestone `start_date` and `milestone_date`).

- Add rake task to change milestone dates

Add library to support safely changing milestone dates; validation is performed on existing declarations to ensure they will remain valid.

Add rake task to allow milestone date changes to be issued from the console:

```
bundle exec rake 'change_milestone_date:<validate|execute>[schedule_identifier,cohort_start_year,milestone_number,new_start_date,new_milestone_date]
```

### Guidance to review

Example usage:

```
bundle exec rake 'change_milestone_date:validate[ecf-standard-january,2021,6,,30-9-2023]
bundle exec rake 'change_milestone_date:execute[ecf-standard-january,2021,6,1-8-2023,30-9-2023]'
```

### DIFF
--- a/lib/milestone_date_change.rb
+++ b/lib/milestone_date_change.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+class MilestoneDateChange
+  class DateCannotBeChangedError < RuntimeError; end
+
+  DECLARATION_STATES_TO_IGNORE = %i[voided awaiting_clawback clawed_back].freeze
+
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Validations
+
+  attribute :schedule_identifier, :string
+  attribute :start_year, :integer
+  attribute :milestone_number, :integer
+  attribute :new_start_date, :date
+  attribute :new_milestone_date, :date
+
+  validates :new_milestone_date, presence: { message: I18n.t(:milestone_date_required) }, if: -> { new_start_date.blank? }
+  validates :new_start_date, presence: { message: I18n.t(:milestone_date_required) }, if: -> { new_milestone_date.blank? }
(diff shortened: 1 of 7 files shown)


lib/milestone_date_change.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+  end
+
+  def milestone_declarations
+    return [] unless milestone
+
+    ParticipantDeclaration::ECF
+      .includes(:participant_profile)
+      .where(participant_profile: { schedule_id: schedule.id })
+      .where(declaration_date: existing_date_range)
```
cwrw, line 43: could you append `.where.not(state: ["voided", "awaiting_clawback", "clawed_back"])` to the query to ensure we don't get any voided declarations which we won't want to include in our validations?

lib/milestone_date_change.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+class MilestoneDateChange
```
cwrw, line 3: suggestion: maybe we can move this class into the app/services folder instead so we can reuse and possibly then automate?

cwrw (approved):
looks great, thanks for this. Two minor comments otherwise 👍

cwrw (approved):

PR #3673, opened 2023-07-11
Update schemas to match Rails 7 format

cwrw:
### Context

Since Rails 7.0 the datetime column precision is 6 by default, the dumper will generate the precision as 0 for columns without precision and omit it when the precision is 6.

- Ticket: n/a

Update schemas with latest format for Rails 7

### DIFF
--- a/db/analytics_schema.rb
+++ b/db/analytics_schema.rb
@@ -10,8 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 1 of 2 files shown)


ltello (approved):

peteryates (approved):
Looks great, thanks!

leandroalemao (commented):
lgtm 👍 😄  tks

PR #3680, opened 2023-07-12
[CPDLP-2312] Remove duplication for api latest induction record queries

leandroalemao:
### Context

- Ticket: [CPDLP-2312](https://dfedigital.atlassian.net/browse/CPDLP-2312)

In API queries, we are calculating the latest induction records to extract relevant data. However this logic is duplicated across most queries in different versions.

Consolidate where possible the induction record latest queries in the API.

[CPDLP-2312]: https://dfedigital.atlassian.net/browse/CPDLP-2312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/api/v1/ecf/participants_query.rb
+++ b/app/services/api/v1/ecf/participants_query.rb
@@ -4,9 +4,9 @@ module Api
   module V1
     module ECF
       class ParticipantsQuery
-        include Api::V3::Concerns::FilterCohorts
-        include Api::V3::Concerns::FilterUpdatedSince
-        include Api::V3::Concerns::FetchLatestInductionRecords
+        include Concerns::FilterCohorts
+        include Concerns::FilterUpdatedSince
+        include Concerns::FetchLatestInductionRecords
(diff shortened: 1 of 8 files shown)


ethax-ross (approved):
Nice 💯

app/services/api/v1/ecf/participants_query.rb
```diff
@@ -4,26 +4,16 @@ module Api
   module V1
     module ECF
       class ParticipantsQuery
+        include Api::V3::Concerns::FilterCohorts
```
cwrw, line 7: could we keep v3 concerns out of v1 to avoid unintentionally changing v1 when we amend v3 thinking it's only used there? unless we move the concerns out of versioning, let me know your thoughts

cwrw (approved):
🙌

PR #3683, opened 2023-07-13
Add contidional validation to npq application eligibility status

slawosz:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1224

To prevent financial mismatch, disable ability for admin to turn `eligible` application to `ineligible` when there are declarations that are payble against application.

### DIFF
--- a/app/models/npq_application.rb
+++ b/app/models/npq_application.rb
@@ -1,6 +1,21 @@
 # frozen_string_literal: true
 
 class NPQApplication < ApplicationRecord
+  VALID_FUNDING_ELIGIBILITY_STATUS_CODES = %w[
+    funded
+    no_institution
+    ineligible_establishment_type
+    ineligible_institution_type
+    previously_funded
+    not_new_headteacher_requesting_ehco
+    school_outside_catchment
+    early_years_outside_catchment
+    not_on_early_years_register
+    early_years_invalid_npq
+    marked_funded_by_policy
+    marked_ineligible_by_policy
+  ].freeze
+
   has_paper_trail only: %i[user_id npq_lead_provider_id npq_course_id created_at updated_at lead_provider_approval_status]
 
   self.ignored_columns = %w[user_id]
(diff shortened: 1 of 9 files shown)


app/models/npq_application.rb
```diff
@@ -1,6 +1,21 @@
 # frozen_string_literal: true
 
 class NPQApplication < ApplicationRecord
+  VALID_FUNDING_ELIGIBILITY_STATUS_CODES = %w[
```
slawosz, line 4: I moved this constant from service as its better suited in the model, already discussed with @cwrw

app/models/npq_application.rb
```diff
@@ -115,4 +135,10 @@ def push_enrollment_to_big_query
       NPQ::StreamBigQueryEnrollmentJob.perform_later(npq_application_id: id)
     end
   end
+
+  def validate_funding_eligiblity_status_code_change
+    if declared_as_payable? && eligible_for_funding == false
+      errors.add(:funding_eligiblity_status_code)
```
cwrw, line 141: is there an error message we'd like to add here, also should we add this to base as it involves two fields?

app/models/npq_application.rb
```diff
@@ -94,6 +110,10 @@ def self.participant_declaration_finder(participant_identity_id)
     ParticipantDeclaration::NPQ&.find_by_participant_profile_id(ParticipantProfile&.find_by_participant_identity_id(participant_identity_id)&.id)
   end
 
+  def declared_as_payable?
+    profile.participant_declarations.paid_payable_or_eligible.count.positive?
```
cwrw, line 114: could we use the `billable` scope here? also maybe we can rename the method to `with_billable_declarations?` as it might be more descriptive?

cwrw (commented):
looking good, a few suggestions otherwise 👍

app/models/npq_application.rb
slawosz: For now the way it works is that we don't allow to navigate to the page where this validation might be triggered, its more as a failsafe. But its a good idea and I will implement as suggested.

app/models/npq_application.rb
slawosz: Sure, Just out of curiosity, why we have `billable` and `paid_payable_or_eligible` scopes?

cwrw (approved):
🙌 thanks for addressing all comment!

PR #3691, opened 2023-07-14
Clean up filtering/sorting logic

ethax-ross:
[Jira-2313](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2313)

### Context

The filtering and sorting logic is a bit fragmented at the moment; we have a concern for validating filter attributes that also pulls in a bunch of filter parsing to the v3 controllers that isn't needed. We also perform sorting in both the v3 controllers (user specified sort orders) and v3 query models (default sort ordering).

We want to clean this up so that only the filter validation is included in the v3 controllers and all the sorting logic is encapsulated in the v3 query models.

-  Decouple v3 APIs from ApiFilter concern

Decouple v3 API controllers from `ApiFilter` with a new `ApiFilterValidation` concern that simply validates the filter params.

Unfortunately we can't scope the `ApiFilter` to only v1 as its referenced by a v2 controller so I'm leaving this in the global scope for now.

- Clean up ApiOrderable concern

Tweaks the concern to make it a bit easier to follow and adds some missing tests.

- Encapsulate sorting logic in v3 query models

Ensures sorting is only applied in the v3 query models and not the controllers.

- Bring ApiOrderable concern into query service

The `ApiOrderable` concern would make more sense applied at the query model level, which is consistent with how we do the filtering.

### Guidance to review

Best reviewed by commit.


cwrw (approved):
Beautifully done, thanks @ethax-ross!

mooktakim (approved):

leandroalemao (approved):

PR #3693, opened 2023-07-14
Add chaser emails and reorg other mailer services

tonyheadford:
### Context

- Related to ongoing launch comm bulk emails, add the chaser emails to a service class and move other bulk mailer services to a `BulkMailers` namespace

Add the current batch of chaser comms
Refactor other bulk mailers into a new namespace

### DIFF
--- a/app/services/mailers/add_participants_reminder.rb
+++ b/app/services/mailers/add_participants_reminder.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class AddParticipantsReminder
+class Mailers::AddParticipantsReminder
   def initialize(cohort:)
     @cohort = cohort
   end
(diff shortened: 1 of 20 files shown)


app/services/mailers/launch_reminder_comms.rb
```diff
@@ -0,0 +1,128 @@
+# frozen_string_literal: true
+
+class Mailers::LaunchReminderComms
```
ltello, line 3: Maybe `class Comms::LaunchReminder`?

ltello (approved):

app/services/mailers/add_participants_reminder.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class AddParticipantsReminder
+class Mailers::AddParticipantsReminder
```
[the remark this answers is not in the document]
tonyheadford: Yes, this is a good call.

app/services/mailers/launch_reminder_comms.rb
tonyheadford: Thanks for the suggestion, I decided to use `BulkMailers` for the module name 👍

PR #3697, opened 2023-07-17
Add a review environment for the review apps

peteryates:
This will ultimately replace deployed_development in a post-PaaS world.


config/environments/review.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require Rails.root.join("config/environments/production")
+
+Rails.application.configure do
+  config.log_level = :debug
```
peteryates, line 6: Change this to `INFO` before merging 😱

cwrw (commented):
looks good, we need to add `review` to other parts of the app (like bypassing sign in etc.) but can do in subsequent pr

cwrw (approved):

PR #3708, opened 2023-07-19
[CST-1641] Existing ECT as new mentor

lazydevorg:
### Context

- Ticket: CST-1641

The add participant wizard wasn't allowing adding a mentor participant if it was already an ECT.
The wizard now checks whether the participant to add is a mentor and if it's already an ECT it allows to proceed.


tonyheadford (approved):
I think it looks good, just one suggestion. I think it needs a PO check to ensure the journey is working how they want it but it looked fine to me in the review app.

PR #3711, opened 2023-07-20
Add deploy env to AKS step

peteryates:
This change adds automatic review app deployment to AKS. There is a generic action, `deploy-environment-to-aks` that will deploy any docker image to any environment, and a new workflow, `aks_deploy_review` that builds a docker image for the current PR and calls the action.

We will be able to deploy other environments using the same process.

### DIFF
--- a/.github/workflows/aks_deploy_review.yml
+++ b/.github/workflows/aks_deploy_review.yml
@@ -42,6 +42,7 @@ jobs:
   deploy_review:
     name: Deploy to review environment
     concurrency: deploy_review_${{ github.event.pull_request.number }}
+    if: contains(github.event.pull_request.labels.*.name, 'review-app')
     needs: [docker]
     runs-on: ubuntu-latest
     environment:
--- a/.github/workflows/aks_deploy_review.yml
+++ b/.github/workflows/aks_deploy_review.yml
@@ -1,0 +1,58 @@
+name: "Deploy review"
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Deploy environment"
+        required: true
+        default: review_aks
+        type: environment
+        options:
+          - review_aks
+  push:
+    branches:
+      - main
+
+  pull_request:
+    branches:
+      - main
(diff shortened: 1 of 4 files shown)


.github/workflows/aks_deploy_review.yml
```diff
@@ -0,0 +1,58 @@
...
+          - review_aks
+  push:
+    branches:
+      - main
+
+  pull_request:
+    branches:
+      - main
+    types:
+      - labeled
```
cwrw, line 20: wondering if this will always trigger a deploy vs this is triggered via adding a label. I think maybe we start with a label until we add the destroy workflow?

.github/workflows/aks_deploy_review.yml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: DFE-Digital/github-actions/build-docker-image@master
+        id: build-docker-image
+        with:
+          docker-repository: ghcr.io/dfe-digital/early-careers-framework
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+  deploy_review:
```
cwrw, line 42: could we add smoke tests to the steps? And maybe the step which adds a comment to the pr?

peteryates: Was going to do those as a follow up as we'll be deploying in parallel for a while anyway.

.github/workflows/aks_deploy_review.yml
peteryates: Temporarily removed the `if` statement so we always build them, think we should add it back as a nice to have once we've seen this run a few times and are confident it's fully working

peteryates:
cc @cwrw 

| No label | Label |
| --------| --------- |
| Screenshot from 2023-07-24 10-18-27| Screenshot from 2023-07-24 10-22-49 |

cwrw (approved):
🙌

PR #3714, opened 2023-07-20
CPDNPQ-1253: Create migration for entries and updated them accordingly

javier-npq:
### Context
Create migration for entries and updated them accordingly
### Ticket
https://dfedigital.atlassian.net/browse/CPDNPQ-1253
Created migration for add new entries name `updated_by` and `eligible_for_funding_updated_at` and updated them accordingly.

### DIFF
--- a/app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
+++ b/app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
@@ -22,6 +22,7 @@ module Admin
 
           set_eligibility_status(eligible_for_funding_params["eligible_for_funding"])
           name = @npq_application.participant_identity.user.full_name
+          @npq_application.current_user = current_user
 
           if @npq_application.save(context: :admin)
             flash[:success] = {
--- a/app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
+++ b/app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
@@ -24,7 +24,9 @@ module Admin
           name = @npq_application.participant_identity.user.full_name
 
           if @npq_application.save(context: :admin)
-
+            if @npq_application.saved_change_to_eligible_for_funding?
+              @npq_application.update_eligibility_information(current_user)
+            end
             flash[:success] = {
               title: "#{name} updated",
               content: "#{name} has been marked '#{@npq_application.funding_eligiblity_status_code.humanize.downcase}'",
--- a/app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
+++ b/app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
@@ -24,9 +24,6 @@ module Admin
           name = @npq_application.participant_identity.user.full_name
 
           if @npq_application.save(context: :admin)
-            if @npq_application.saved_change_to_eligible_for_funding?
-              @npq_application.update_eligibility_information(current_user)
-            end
             flash[:success] = {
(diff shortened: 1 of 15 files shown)


app/models/npq_application.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class NPQApplication < ApplicationRecord
+  attr_accessor :current_user
```
slawosz, line 4: I would name this field differently, maybe `admin_user`? Or different approach - field that store `bool` value that application can be modified. Approving anyway.

slawosz (approved):

app/models/npq_application.rb
javier-npq: Actually, `current_user` instance does not exist within the scope of this modal. That's why I used this approach to pass it in this way.

db/migrate/20230719140256_add_updated_by_and_eligible_for_funding_updated_at_to_npq_application.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddUpdatedByAndEligibleForFundingUpdatedAtToNPQApplication < ActiveRecord::Migration[7.0]
+  def change
+    add_column :npq_applications, :updated_by, :string
```
cwrw, line 5: Just because I can't see the ticket, do we intend to surface those values anywhere or is it explicitly for auditing purposes? If it's only for auditing purposes I wonder if we can use papertrail and add the `eligibility_for_funding` to the tracked fields?

If not and we intend to surface somewhere I suggest renaming to `eligible_for_funding_updated_by` as we're tracking the setting of that field only

app/controllers/admin/npq/applications/eligible_for_funding_controller.rb
```diff
@@ -22,9 +22,9 @@ def update
 
           set_eligibility_status(eligible_for_funding_params["eligible_for_funding"])
           name = @npq_application.participant_identity.user.full_name
+          @npq_application.current_user = current_user
```
cwrw, line 25: I would suggest moving the logic of updating the eligibility for funding into a service where we can add all the current logic and avoid the callbacks.

If not maybe we can do this here:
```
@npq_application.updated_by = current_user.email
@npq_application.eligible_for_funding_updated_at = Time.zone.now
```
to avoid adding a callback?

app/models/npq_application.rb
```diff
@@ -141,4 +144,12 @@ def validate_funding_eligiblity_status_code_change
       errors.add(:base, :billable_declaration_exists)
     end
   end
+
+  def update_eligibility_information
+    return unless current_user&.admin?
+
+    self.updated_by = current_user.email
```
cwrw, line 151: maybe we should be storing the user itself as emails can change?

cwrw (commented):
a few suggestions, happy to chat if its easier

db/migrate/20230719140256_add_updated_by_and_eligible_for_funding_updated_at_to_npq_application.rb
javier-npq: Yes, we do intend to surface these values. We want to show if an application was updated by an admin along with the timestamp. The answer to `where do we want to show it?` is currently not defined and we are working on grooming that part. 

I'll update the rest of the code according to the feedback.

PR #3719, opened 2023-07-21
Set explicit challenge deadline on partnership request spec

cwrw:
### Context
The spec `spec/requests/schools/partnerships_spec.rb:85` is currently failing due to us approaching the next cohort in the factories.
The test is checking that partnership cannot be challenged after it passes the challenge deadline.

- Ticket: n/a

To avoid failures when approaching the next cohort, set the challenge deadline to current academic year with extra days, and travel to after the deadline explicitly.

### Guidance to review
Did I miss anything?


peteryates (approved):
Looks good, definitely more robust.

PR #3738, opened 2023-07-27
Build the AKS pipeline

peteryates:
### Context

- Rename aks_deploy_review to aks_deploy
- Build docker image for PR pushes and main
- Run RSpec as part of the deployment pipeline


.github/workflows/aks_deploy.yml
```diff
@@ -26,7 +26,7 @@ jobs:
   docker:
     name: Build and push Docker image
     runs-on: ubuntu-latest
-    if: contains(github.event.pull_request.labels.*.name, 'review-app')
+    if: contains(github.event.pull_request.labels.*.name, 'review-app') || github.event_name != 'pull_request'
```
peteryates, line 29:
```suggestion
    if: contains(github.event.pull_request.labels.*.name, 'review-app') || github.ref == 'refs/heads/main'
```

Would this work? 🤔

.github/workflows/rspec.yml
```diff
@@ -0,0 +1,133 @@
...
+          cache: "yarn"
+
+      - name: Install node.js dependencies
+        run: yarn install --production=false --frozen-lockfile --ignore-engines --prefer-offline
+
+      - name: Set up test database
+        run: bin/rails db:create db:schema:load
+
+      - name: Prepare assets
+        run: bin/rails test:prepare
```
peteryates, line 130: Is there a nice way of reducing the duplication between here and just above?

.github/workflows/aks_deploy.yml
RMcVelia: I think you will always want the docker step to run i.e. on pr update (with or without review-app), and merge to main. So the if statement would be dropped.

RMcVelia: Also deploy-review if should be on pr only, so line 46 `if: ${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'review-app') }}`

.github/workflows/aks_deploy.yml
```diff
@@ -66,3 +66,43 @@ jobs:
...
             Review app deployed to ${{ steps.deploy.outputs.url }}
+
+  test:
+    name: Run the test suite
+    uses: ./.github/workflows/rspec.yml
+
+  deploy_staging:
+    needs: [docker, test]
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
```
RMcVelia, line 77: requires environment setting as you are using secrets in that format, so
```
environment:
    name: staging
```
Same for the sandbox and production deploys

.github/workflows/rspec.yml
RMcVelia: can you add it as a composite action? and then just call it in each job

peteryates: Ah, we have `prepare-application-environment` which should do the trick with a few amendments 👌🏽  Good call @RMcVelia

peteryates: 82ae947a3686d7e9c40dd1ca010977331e1f59a4 🤞🏽

RMcVelia (approved):

.github/workflows/aks_deploy.yml
```diff
@@ -26,7 +26,6 @@ jobs:
   docker:
     name: Build and push Docker image
     runs-on: ubuntu-latest
-    if: contains(github.event.pull_request.labels.*.name, 'review-app')
```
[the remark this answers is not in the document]
peteryates: Ah yeah, that's why we added it isn't it 🤦🏽  Will add it back for now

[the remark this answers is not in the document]
peteryates: Done, repushed

PR #3741, opened 2023-07-27
Add back requires to allow data locally to be loaded as expected

cwrw:
### Context
Without those requires development environments do not load the different ECF/NPQ records causing confusion. 

- Ticket: n/a
those were removed from https://github.com/DFE-Digital/early-careers-framework/pull/3590/commits/3e760ceb74c4fe794742e791fbc68263680f744b

Add back requires to statements and schedules to allow records to be loaded in development 

### Guidance to review
Did I miss anyhting?

### DIFF
--- a/app/models/finance/schedule.rb
+++ b/app/models/finance/schedule.rb
@@ -10,3 +10,6 @@ module Finance
     validates :schedule_identifier, presence: true
   end
 end
+
+require "finance/schedule/ecf"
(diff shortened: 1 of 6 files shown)


leandroalemao (approved):
👍 lgtm 👍 
related to https://github.com/DFE-Digital/early-careers-framework/pull/3692

PR #3744, opened 2023-07-28
Add Brakeman vulnerability scanner to the pipeline

peteryates:
WIP


cwrw (approved):

config/brakeman.ignore
```diff
@@ -0,0 +1,155 @@
+{
```
cwrw, line 1: discussed in person but for posterity: we might not need this as Github adds comments inline with the PR

peteryates: Done, I didn't even realise GitHub had this level of integration 💡

PR #3748, opened 2023-07-28
Remove old pages that appear to contain CIP stuff

peteryates:
### Context

This was moved to the [engage-and-learn app](https://support-for-early-career-teachers.education.gov.uk/) in 2021, I don't think there's any benefit to us hanging on to it in the ECF app.


cwrw (commented):
looks good, should we also remove those from routes as well?

peteryates:
Turns out it is necessary but it needs to be moved within the app before we migrate.

PR #3766, opened 2023-08-01
[CST-1831] [CST-1832] Add maintenance mode functionality

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CST-1831
- Ticket: https://dfedigital.atlassian.net/browse/CST-1832

- Adds easily toggleable maintenance mode notification banner.
- Adds maintenance mode static site that can be toggled on/off via makefile.

image

### Guidance to review

The banner is currently off, so you'll need to pull it down locally and set the constants to see it working.

The maintenance mode itself is usable in all environments and [the documentation on the maintenance mode](https://github.com/DFE-Digital/early-careers-framework/blob/f170885e165756811f86539279be37cd535d922d/docs/source/ecf/maintenance_mode.html.md) has details on how to use it on a review app, which would let you test it on this PR's review app.


Makefile
```diff
@@ -128,3 +128,25 @@ terraform-plan: terraform-init
...
 	terraform -chdir=terraform/aks destroy -var-file workspace_variables/${DEPLOY_ENV}.tfvars.json ${AUTO_APPROVE}
+
+enable-maintenance:
+	$(if ${DEPLOY_ENV} == "review_aks", $(eval VARIABLE_FILE_NAME=review), $(eval VARIABLE_FILE_NAME=${DEPLOY_ENV}))
+	$(eval include terraform/workspace-variables/${VARIABLE_FILE_NAME}.tfvars)
+	$(eval SPACE=${paas_space_name})
+
+	cf target -s ${SPACE}
+	cd service_unavailable_page && cf push
+	cf map-route ecf-unavailable london.cloudapps.digital --hostname ecf-review-pr-3766
```
ltello, line 139: should this be

`cf map-route ecf-unavailable london.cloudapps.digital --hostname ${APP_NAME}`

instead?

tonyheadford (commented):
Could we have a screenshot or two of the banner and the maintenance mode app in the PR description so we could see what it looks like without needing to make changes etc.?

ltello (approved):
Brilliant work!!

peteryates:
Looks great, a rebase and force push should fix the review app deploy failure.

app/components/banners/maintenance_component.rb
```diff
@@ -0,0 +1,70 @@
...
+    DATE       = nil # Date.new(2021, 8, 4)
+    START_TIME = nil # Time.new(2021, 8, 4, 9, 0, 0)
+    END_TIME   = nil # Time.new(2023, 8, 4, 17, 0, 0)
+
+    def initialize(wide_container_view: false)
+      @wide_container_view = wide_container_view
+    end
+
+    def render?
+      date.present?
```
cwrw, line 21: I wonder if we should allow this to be toggled via a feature flag instead? So we don't have to deploy to reset date to `nil`?

tonyheadford: Agree, I know we need a date but switching on/off via a `FeatureFlag` would be great

Lockyy: That's a fair point, I only just saw this and have just pushed another change which is that the banner goes away automatically once the date it is set to passes.

ltello (approved):

PR #3768, opened 2023-08-02
Fix validation in the admin school transfer form

leoapost:
### Context

- Ticket: CST-1856

The validation is retrieving the participant's current school and checks if it's not the same with the one moving into. Problem is that it's not retrieving the participant's school from their induction record.

- Update the validation to use the participant's latest induction record to retrieve their current school


tonyheadford (approved):
LGTM

PR #3769, opened 2023-08-02
Add linting step to the pipeline

peteryates:
### Context

This will run first as it should always be quick and will highlight problems sooner rather than later. Might move it later if it gets in the way.


cwrw (approved):
🙌

PR #3773, opened 2023-08-02
[CST-1827] Access DQT V3 API

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-1827)

First part of enabling us to retrieve the induction completion dates is to access the Teacher Qualifications V3 API.
This enables us to look up a participants induction details from their TRN.

Refactor currently used API client into a V1 namespace and to make it more suitable as a base for the V3 Client to inherit from.
Add a V3 Client that leverages much of the connectivity handling of the V1 Client.


lib/full_dqt/v3/client.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+# DO NOT USE THIS CLIENT FOR VALIDATION
+#
+# (in theory at some point in the future we could use this for validation,
+# but using the V1 API keeps the id matching logic on the API side whereas
+# using the V3 API we would need to match and validate everything our side)
+#
+module FullDQT
+  module V3
+    class Client < V1::Client
```
ltello, line 16: Is the same api key valid for both versions of the API?

ltello: Do we have a sandbox V3 environment of this API like those V1 we use in our review and dev environments?

tonyheadford: Yes and yes, it works against the same test data. I run queries in prod and dev and it all works fine with our current keys

ltello: Great! Thanks

ltello (approved):

PR #3781, opened 2023-08-04
Add sandbox and production environments

peteryates:

- Remove duplicated setting of `DEPLOY_ENV`
- Add sandbox and production global configs
- Add deployment steps for sanbdox and production

### Tasks

* [x] Get production credentials sorted, will need help from @RMcVelia or @saliceti 
* [ ] Actually test it 🙈 
* [ ] Can we populate `$GITHUB_OUTPUT` with some URLs?

### DIFF
--- a/.github/workflows/aks_deploy.yml
+++ b/.github/workflows/aks_deploy.yml
@@ -106,23 +106,33 @@ jobs:
   deploy_sandbox:
     needs: [deploy_staging]
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/main'
     environment:
       name: sandbox
 
     steps:
       - uses: actions/checkout@v3
-      - run: |
-          echo "deploying sandbox"
+
+      - uses: ./.github/actions/deploy-environment-to-aks
+        id: deploy
+        with:
+          environment: sandbox
+          docker-image: ${{ needs.docker.outputs.image }}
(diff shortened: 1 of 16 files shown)


.github/workflows/aks_deploy.yml
```diff
@@ -102,23 +102,33 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - run: |
-          echo "deploying sandbox"
+
+      - uses: ./.github/actions/deploy-environment-to-aks
+        id: deploy
+        with:
+          environment: sandbox
+          docker-image: ${{ needs.docker.outputs.image }}
```
saliceti, line 115: This will probably not work since the deploy_sandbox job doesn't "need" the docker job

peteryates: Ah - well spotted. So looking [at the docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idoutputs) this could simply be this (I think):

```suggestion
          docker-image: ${{ docker.outputs.image }}
```

peteryates: My assumption was incorrect, repushed in a way that passes the `docker-image` along

peteryates:
Closing, will reopen

PR #3783, opened 2023-08-07
Add library to deduplicate ECF profiles

ethax-ross:
[Jira-1728](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-1728)

### Context

We have a number of duplicate profiles in the database and we want to reconcile these records. At the moment its a manual process; instead, we want to write a library that can handle the deduplication logic. Ideally we want to be able to deduplicate a pair of profiles straight from the existing duplicates UI. To begin with the library will have limited functionality; the idea being we will expand it as and when we meet new shapes of duplicated data.

- Add library to deduplicate ECF profiles

Add library to reconcile a duplicate profile with the primary record. Currently handles school changes, validation/eligibility merging, declaration reconciliation and serialization then deletion of the duplicate record.

- Add additional criteria/checks to deduplication lib

Add additional checks to ensure we are only automating the deduplication of records that we know the library can handle.

Improves the logging/changelog so that we can more easily output it to the caller.

- Support plain text highlighting

We are going to use this to show the output of the deduplication script.

Removes class on `pre` tag; it was misspelled and we don't appear to declare it anywhere anyway.

- Support safe HTML in flash messages

We are going to output multi-line text from the deduplication script in a flash message so that its clear to see whats changed after the script has ran. Allow basic/safe HTML in the flash message.

- Update duplicates UI with deduplication button

Adds a button to perform a 'dry run' where nothing changes but it outputs what _will_ change. If the dry run is successful and the deduplication can be automated the 'deduplicate' button will be enabled and the user can click this
to perform the deduplication.

### Guidance to review

Rake task usage:

```
bundle exec rake 'participant_profile_deduplicator:dedup_dry_run[<primary-profile-uuid>,<duplicate-profile-uuid>]'
```

| Error      | Dry Run | Success |
| ----------- | ----------- | --------- |
| screencapture-0-0-0-0-3000-finance-ecf-duplicates-012ad231-9ce6-4550-a64c-475816810f55-compare-0fbf3b12-186d-4bbb-a27b-32daec97c484-deduplicate-2023-08-10-14_57_37 |  screencapture-0-0-0-0-3000-finance-ecf-duplicates-cd860222-ddfb-471e-999e-e47c619e9e18-compare-2557b527-f1f8-4ed0-accc-9e7020ed54b1-deduplicate-2023-08-10-15_02_38       | screencapture-0-0-0-0-3000-finance-ecf-duplicates-2023-08-10-15_02_50       |.


lib/participant_profile_deduplicator.rb
```diff
@@ -0,0 +1,150 @@
...
+      induction_status: :leaving,
+      end_date:,
+      preferred_identity: preferred_identity(duplicate_induction_record),
+    )
+
+    log_info("Preferred identity updated on duplicate induction record") if duplicate_induction_record.saved_change_to_preferred_identity_id?
+
+    log_info("Duplicate profile induction record transferred. End date: #{end_date}")
+
+    primary_profile.latest_induction_record.update!(school_transfer: true)
```
cwrw, line 57: I think we need to set only the oldest induction record coming after the transfer to `school_transfer` true

lib/participant_profile_deduplicator.rb
```diff
@@ -0,0 +1,150 @@
...
+  def primary_profile
+    @primary_profile ||= ParticipantProfile::ECF.find(primary_profile_id)
+  end
+
+  def duplicate_profile
+    @duplicate_profile ||= ParticipantProfile::ECF.find(duplicate_profile_id)
+  end
+
+  def preferred_identity(duplicate_induction_record)
+    if duplicate_induction_record.preferred_identity.user_id == primary_profile.latest_induction_record.preferred_identity.user_id
```
cwrw, line 140: maybe we should go with the oldest as this will precede that 🤔

cwrw (approved):
💯 fantastic!

PR #3789, opened 2023-08-08
[CST-1851] Set induction completion date and induction status

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-1851)

Set the `induction_completion_date` and add a new induction record with the `completed` induction status.

Adds a service to mark a participant as completed and set the completion date.
The actual induction completion date will be retrieved in a subsequent PR

### DIFF
--- a/app/models/induction_record.rb
+++ b/app/models/induction_record.rb
@@ -103,6 +103,10 @@ class InductionRecord < ApplicationRecord
     update!(induction_status: :changed, end_date: date_of_change)
   end
 
+  def completing!(date_of_change = Time.zone.now)
+    update!(induction_status: :completed, end_date: date_of_change)
+  end
+
   def claimed_by_another_school?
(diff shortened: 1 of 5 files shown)


cwrw (commented):
looking great, thanks @tonyheadford I've tested locally and everything works as expected in the API. I was wondering if its worth adding any completed seeds for review apps, maybe a couple of completed profiles for examples for later?

tonyheadford:
> looking great, thanks @tonyheadford I've tested locally and everything works as expected in the API. I was wondering if its worth adding any seeds for review apps, maybe a couple of completed profiles for examples for later?

Thanks. That's a good idea although there is another ticket regarding ensuring that completed participants are displayed correctly in the UI which might be a better place for the scenarios but seeds could be here 👍

PR #3796, opened 2023-08-09
GDPR request scenario tests

mattonfoot:
### Context

- Ticket: https://ukgovernmentdfe.slack.com/archives/C03RM65L928/p1689670001677279

- add a new seed that adds a participant scenario matching the structure of the record the GDPR request is targeting so that we can run simulations on changes to allow us to hide the users name and email address from the ECF provider but not the NPQ provider

### Guidance for review

- find the school called "GDPR requests school" in either the console or the admin UI. A single participant is there and it is worth examining their attributes in both the Admin UI and the Finance UI.


app/models/gdpr_request.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class GDPRRequest < ApplicationRecord
+  has_paper_trail
+
+  belongs_to :cpd_lead_provider
+  belongs_to :teacher_profile
```
cwrw, line 7: could the GDPR request be attached to the user instead?

app/models/gdpr_request.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class GDPRRequest < ApplicationRecord
```
cwrw, line 3: Could we name the model something around hiding data, and the `request_type` is GDPR request?

app/models/gdpr_request.rb
```diff
@@ -0,0 +1,12 @@
...
+  has_paper_trail
+
+  belongs_to :cpd_lead_provider
+  belongs_to :teacher_profile
+
+  enum type: {
```
cwrw, line 9:
```suggestion
  enum reason: {
```

app/services/api/v1/ecf/participants_query.rb
```diff
@@ -17,6 +17,7 @@ def induction_records
           scope = InductionRecord
                     .select(*necessary_fields)
                     .eager_load(:schedule)
+                    .joins(induction_programme: { partnership: { lead_provider: %i[cpd_lead_provider] } })
```
cwrw, line 20: I'd like to have a think about this as we already join on the latest induction record so need to join on this again as we have access to the lead provider already, will wait till we agree on the structure of the table first

app/serializers/api/v3/ecf/participant_serializer.rb
```diff
@@ -30,6 +30,11 @@ def eligible_for_funding?(record)
             return false if ecf_participant_eligibility_status == "ineligible"
           end
 
+          def can_view_personal_details?(user, cpd_lead_provider)
+            gdpr_request = user.teacher_profile.gdpr_requests.where(cpd_lead_provider:).first
```
cwrw, line 34: this will add n+1 queries, and needs to be added to the main query, happy to amend the queries after discussion

cwrw (commented):
A few comments to discuss around what to return in obfuscated fields, which entity the new model should link to (and amending the queries to accommodate that avoiding n+1) and API release note and guidance. 
Also I think we need to roll this out to NPQ endpoints as well, another thing to discuss

PR #3799, opened 2023-08-10
Remove the old test workflow

peteryates:
### Context

Now we run these as part of the main deploy flow and at the moment they're run twice in parallel, which isn't necessary.

There's a 'Permit merge' task that we can use to simplify the branch protections

| Current branch protections | New step |
| ----------------------------| ------------ |
|image | image|


Lockyy (approved):
This looks good and everything is passing but the merge requirements be met till we merge and swap them over.

peteryates:
New required steps:

image

PR #3800, opened 2023-08-10
[CST-1845] Clean up mailers

leoapost:
### Context

- Ticket: CST-1845

We have lots of mailers and Notify templates that we no longer use or no longer necessary to send.

- Remove mailers asking participants to validate
- Remove various unused mailers
- Remove code that's using the mailers that are being removed
- Annotate inactive mailers to be removed in follow up PRs


app/services/mentors/create.rb
```diff
@@ -31,25 +31,18 @@ def call
         Mentors::AddToSchool.call(school: school_cohort.school, mentor_profile:)
       end
 
-      unless sit_validation
-        ParticipantMailer.with(participant_profile: mentor_profile).participant_added.deliver_later
-        mentor_profile.update_column(:request_for_details_sent_at, Time.zone.now)
-        ParticipantDetailsReminderJob.schedule(mentor_profile)
```
tonyheadford, line 37: There's a ton of these jobs scheduled in prod we should probably remove at the same time

tonyheadford (approved):
Thanks, it will be great to tidy this all up

PR #3810, opened 2023-08-11
[CPDNPQ-1226] Track npq eligibility changes via Papertail

pmanrubia:
### Context

- Ticket: [CPDNPQ-1226](https://dfedigital.atlassian.net/browse/CPDNPQ-1226)

We aim to track eligibility changes to NQP applications in the Edge cases section of the Admin panel. For this first iteration we are only considering two attributes:

- `eligible_for_funding`
- `funding_eligiblity_status_code`

I am extending the use of `PaperTrail` to track changes to these attributes as it seems cohesive with the existing patterns of the application. I have only included eligibility attributes, but we could extend to others should we need to.

image

image

### Technical notes

1. [I have noticed that we are listing [`created_at` and `updated_at`](https://github.com/DFE-Digital/early-careers-framework/blob/7c167e8d28f591ca844b51a7bbc25258d78a4b6e/app/models/npq_application.rb#L19) as attributes where changes would create a new version in `Papertrail`. This was a bit confusing to me because it seems that it defeats the purpose of having a list of attributes to track changes through time.
2. I tried to add `Breadcrums` but when adding new paths [they all seem linked to `Schools`](https://github.com/DFE-Digital/early-careers-framework/blob/80456f3d99c66ac7c0095441c48f4101dd910f90/app/controllers/admin/base_controller.rb#L20-L23). It is probably worth removing this dependency in a separate PR. 
3. I have added a method to `NPQApplication` to return the `#change_logs` with only the eligibility attributes. This does not look ideal, as `change_log` is a global concept that would apply to most attributes. I will be clarifying this matter with the BA. 
4. This PR is related to #3714. Initially we thought that we only wanted to track the last user that made a change to the eligibility attributes, so two new attributes were added to `NPQApplication` model class: `eligible_for_funding_updated_by_id` and `eligible_for_funding_updated_at`. We can probably remove these new fields now because we requirements have changed.

### Guidance to review

https://ecf-review-pr-3810.london.cloudapps.digital/

[CPDNPQ-1226]: https://dfedigital.atlassian.net/browse/CPDNPQ-1226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -107,6 +107,20 @@ RSpec.describe ApplicationHelper, type: :helper do
     end
   end
 
+  describe "#boolean_to_yes_no" do
+    it 'returns "Yes" when value is true' do
+      expect(helper.boolean_to_yes_no(true)).to eq("Yes")
+    end
+
+    it 'returns "No" when value is false' do
+      expect(helper.boolean_to_yes_no(false)).to eq("No")
+    end
+
+    it 'returns "No" when value is nil' do
+      expect(helper.boolean_to_yes_no(nil)).to eq("No")
+    end
+  end
+
   describe "#service_name" do
     it "displays the default service name" do
       helper.request.path = "/"


javier-npq (approved):
Looks good to me

app/views/admin/npq/applications/change_logs/show.html.erb
```diff
@@ -0,0 +1,50 @@
...
+<% @versions.each_with_index do |version| %>
+
+  <p class="govuk-heading-m govuk-!-margin-bottom-0 govuk-!-padding-top-4"><%= version.created_at.strftime('%d %B %Y %H:%M:%S') %></p>
+  <p class="govuk-caption-m govuk-!-margin-top-0 govuk-!-margin-bottom-2">Updated
+    by <%= User.find(version.whodunnit).email %></p>
+
+  <%= govuk_table(classes: 'govuk-!-margin-top-2') do |table|
+    table.with_head do |head|
+      head.with_row do |row|
+        row.with_cell(header: true, text: 'Field changed', classes: 'govuk-!-width-one-third')
```
peteryates, line 18: The table component supports the `width` argument here:

```suggestion
        row.with_cell(header: true, text: 'Field changed', width: 'one-third')
```

app/models/npq_application.rb
```diff
@@ -119,6 +119,16 @@ def has_submitted_declaration?
...
   end
 
+  def change_logs
+    v1 = versions.where_attribute_changes("eligible_for_funding")
+    v2 = versions.where_attribute_changes("funding_eligiblity_status_code")
+
+    (v1 + v2)
+      .uniq
+      .sort { |a, b| b.created_at <=> a.created_at }
+      .compact
```
peteryates, line 129: Not sure how it would be possible to end up with a `nil` here 🤔

peteryates (approved):
Looks great!

Even though it's _super simple_, I think there should be a basic spec covering `#boolean_to_yes_no`

PR #3815, opened 2023-08-14
Add concurrency group and enable cancel-in-progress

peteryates:
Add [concurrency](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) so that we can automatically cancel runs when one part fails.

We've tested this by force pushing and it appears to work, the first run was correctly cancelled.

image

### DIFF
--- a/.github/workflows/aks_deploy.yml
+++ b/.github/workflows/aks_deploy.yml
@@ -1,4 +1,9 @@
 name: "Deploy"
+
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
 on:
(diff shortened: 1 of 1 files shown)


terraform/aks/terraform.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = "1.5.4"
+  required_version = "1.5.5"
```
cwrw, line 2: I think if we were to bump this to 1.5.5, we have to do it everywhere now (there are quite a few other places)

cwrw (approved):

PR #3818, opened 2023-08-14
[550] Custom domains

saliceti:
### Context
Custom domains on AKS
- Ticket: https://trello.com/c/FjC2l5Mx

- Create custom domains for staging and sandbox
- Update terraform versions

### Guidance to review
- https://sb.manage-training-for-early-career-teachers.education.gov.uk/sandbox
- https://st.manage-training-for-early-career-teachers.education.gov.uk/sandbox


terraform/custom_domains/environment_domains/terraform.tf
```diff
@@ -1,10 +1,10 @@
 terraform {
 
-  required_version = "= 1.3.5"
+  required_version = "= 1.5.4"
```
cwrw, line 3: is there a way we can lock this to only the latest minor version instead?

saliceti: From experience we found it's safer be as explicit as possible so everyone uses the same exact version and updates the same state file accordingly. And version updates are done explicitly via PR.
I've added an entry for [asdf](https://technical-guidance.education.gov.uk/infrastructure/dev-tools/#asdf) in this PR to help moving between version seemlessly.

saliceti:
[ECF Custom domains](https://trello.com/c/FjC2l5Mx/550-ecf-custom-domains)

cwrw (approved):

PR #3831, opened 2023-08-15
Add job to update the induction start dates from DQT for 2021/22 ECTs

tonyheadford:
### Context

- Ticket: Follow on from #3819 to actually call the services and update the participants that are missing those dates.
- There are approx 58,000 ECTs in total from 2021 and 2022 when we didn't store the date.
- The job will run every 2 minutes from midnight to 8am.  This should do most if not all of these and keep within our 300 per minute quota.
- This job can then probably be removed.

### DIFF
--- a/app/jobs/set_participant_start_date_job.rb
+++ b/app/jobs/set_participant_start_date_job.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+# This is a one off job to update all of the induction start dates from DQT for
+# all of the 2021/2022 ECTs that do not currently have them.
+# It runs in batches of 200 as we have a 300 lookup per minute limit on the API
+# There are approx 57600 participants in this state currently
+class SetParticipantStartDateJob < ApplicationJob
+  def perform
+    ParticipantProfile::ECT
+      .joins(schedule: :cohort)
+      .eligible_status
+      .where(cohort: { start_year: [2021, 2022] })
+      .where(induction_start_date: nil)
+      .order(:created_at)
+      .limit(200)
+      .each do |participant_profile|
+        ActiveRecord::Base.no_touching do
+          Participants::SetStartDateFromDQT.call(participant_profile:)
+        end
+      end
+  rescue StandardError => e
+    Rails.logger.error("SetParticipantStartDateJob: #{e.message}")
+  end
+end
--- a/config/sidekiq_cron_schedule.yml
+++ b/config/sidekiq_cron_schedule.yml
@@ -33,3 +33,7 @@ send_outcomes_to_qualified_teachers_api:
   cron: "*/10 * * * *"
   class: "ParticipantOutcomes::BatchSendLatestOutcomesJob"
   queue: participant_outcomes
+set_participant_start_date_job:
+  cron: "*/2 0-8 * * *"
+  class: "SetParticipantStartDateJob"
+  queue: default


app/jobs/set_participant_start_date_job.rb
```diff
@@ -0,0 +1,24 @@
...
+# This is a one off job to update all of the induction start dates from DQT for
+# all of the 2021/2022 ECTs that do not currently have them.
+# It runs in batches of 200 as we have a 300 lookup per minute limit on the API
+# There are approx 57600 participants in this state currently
+class SetParticipantStartDateJob < ApplicationJob
+  def perform
+    ParticipantProfile::ECT
+      .joins(schedule: :cohort)
+      .eligible_status
+      .where(cohort: { start_year: [2021, 2022] })
```
ltello, line 12: I don't think we should trust the cohort pointed by the ParticipantProfile but their latest induction record instead?

app/jobs/set_participant_start_date_job.rb
```diff
@@ -0,0 +1,24 @@
...
+      .limit(200)
+      .each do |participant_profile|
+        ActiveRecord::Base.no_touching do
+          Participants::SetStartDateFromDQT.call(participant_profile:)
+        end
+      end
+  rescue StandardError => e
+    Rails.logger.error("SetParticipantStartDateJob: #{e.message}")
+  end
+end
```
ltello, line 24: I think this job is bypassing the "setup permanent cohort" even for participants registered since June?!

peteryates:
I think the approach is sound and will work well, agree with @ltello's concern about which start year to pay attention to though.

app/jobs/set_participant_start_date_job.rb
tonyheadford: Thought the `schedule.cohort` was about as authoritative as it got (shouldn't those match?)

app/jobs/set_participant_start_date_job.rb
tonyheadford: Yes that's a good point I hadn't consider that they might've said they were late 2022 cohort ones but actually turned out to be 2023, I'll add that as a condition 👍

ltello: @peteryates The so called `Participants::SyncDqtInductionStartDate` actually calls `Induction::AmendParticipantCohort`  in case the participant needs their cohort amended once we get their induction start date from the AB via DQT

app/jobs/set_participant_start_date_job.rb
ltello: `Participants::SyncDqtInductionStartDate` probably fits better for this purpose. 
It actually takes that into account and only swap cohorts for the agreed participants.
That means it will actually only update the induction start date to pre_2023 participants (exactly what you need).

tonyheadford: @ltello yes you might be right, I thought that service did more than just update the induction start date and I definitely don't want to update anything other than that as this is supposed to be a non-intrusive back-fill task. In addition because the query is only including `eligible` ECTs they should all have induction start dates that we've checked so no shifting should be necessary.

ltello: @tonyheadford what I am not getting is why we need a new service to sync the induction start dates of some participants from DQT when we carefully analised and made decisions on how to do it a few weeks ago and the result was the existing service `Participants::SyncDqtInductionStartDate`. 
Why not use it?
The participants you are trying to update now were already considered in that analysis and the existing service is doing exactly what your new code will do.

tonyheadford: @ltello as mentioned above I (wrongly) thought the `SyncDQTInductionStartDate` contained more logic than it does and would not work for this but I am planning to switch to it instead when I update this PR.

ltello: @tonyheadford sorry, I misread your message. I (mis)understood you didn't want to use the existing service because it does more than you want to do (which is true but not for the participants you are dealing with in this PR).

app/jobs/set_participant_start_date_job.rb
```diff
@@ -7,15 +7,27 @@
 class SetParticipantStartDateJob < ApplicationJob
   def perform
     ParticipantProfile::ECT
-      .joins(schedule: :cohort)
       .eligible_status
-      .where(cohort: { start_year: [2021, 2022] })
+      .joins(:teacher_profile)
+      .includes(:teacher_profile)
```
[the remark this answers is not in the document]
tonyheadford: I'm eager loading teacher profile or have I misunderstood?

app/jobs/set_participant_start_date_job.rb
```diff
@@ -7,15 +7,27 @@
...
       .where(induction_start_date: nil)
-      .order(:created_at)
+      .where(created_at: ...Cohort.find_by(start_year: 2023).registration_start_date)
+      .order(:updated_at)
       .limit(200)
       .each do |participant_profile|
-        ActiveRecord::Base.no_touching do
-          Participants::SetStartDateFromDQT.call(participant_profile:)
+        induction = DQT::GetInductionRecord.call(trn: participant_profile.teacher_profile.trn)
+        next if induction.blank?
```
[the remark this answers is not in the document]
tonyheadford: Yes good spot. I don't think there should be any without the whole section but it won't hurt to do that.

app/jobs/set_participant_start_date_job.rb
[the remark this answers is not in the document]
ltello: Well, in that case we can get rid of the `joins` clause instead because Rails will trigger a `joins` query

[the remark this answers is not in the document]
tonyheadford: No not really as `.includes` without `.join` generates a `LEFT OUTER JOIN` and `.join` generates an `INNER JOIN`. As it happens it probably won't make much difference but I like to use `.join` when the intention is an `INNER JOIN`.

ltello (approved):

PR #3832, opened 2023-08-15
Increase postgres db size in sandbox and production AKS

cwrw:
### Context
Out current db size is too small in PaaS, which causes CPU usage spikes with just a few API queries. We'd like to scale up our database and work memory in postgres in Azure

I will set the `work_mem` later when we copy over the data, with the following formula:
```
 Total RAM * 0.25 / max_connections
```
we can probably go to 40mb

- Ticket: n/a

- Bump db size in Azure in sandbox and prod (see details in azure [here](https://azure.microsoft.com/en-gb/pricing/details/postgresql/flexible-server/))

### Guidance to review
Did I miss anything?


terraform/aks/workspace_variables/production_aks.tfvars.json
```diff
@@ -1,4 +1,5 @@
 {
+  "postgres_flexible_server_sku": "B_Standard_B4ms",
```
peteryates, line 2: 🔥

peteryates (approved):
I like it, the sizes seem sensible and give us some extra capacity should we need it.

saliceti (approved):

PR #3833, opened 2023-08-15
Update seed name for mathematics NPQ course

cwrw:
### Context
We're keeping the content the same on statements but for consistency amend old and new seed data names
- Ticket: n/a

Update seed data names for NPQ mathematics course


cwrw:
seems like there's a mismatch in seeds, will close for now

PR #3835, opened 2023-08-16
Upgrade to latest version of GOV.UK frontend

peteryates:
### Context

We're _a long way_ behind the latest versions of GOV.UK Frontend, and the form builder and components libraries that are tied to it. This is a quick attempt at doing the upgrade.

- Upgrade to GOV.UK Frontend 4.7.0
- Upgrade components, form builder and pagy gems
- Update tables to use new 'with' syntax
- Update summary lists to use new 'with' syntax
- Update header to use new 'with' syntax
- Update accordions to use new 'with' syntax
- Update tabs to use new 'with' syntax
- Remove deprecated SlotableV2 inclusion
- Update subnav component to use 'with' syntax
- Update banner to use the new 'with' syntax
- Fix visually hidden spacing

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -181,7 +181,6 @@ group :test do
   gem "shoulda-matchers", "~> 5.3"
   gem "simplecov"
   gem "site_prism", "~> 3.7"
-  gem "webdrivers", "~> 5.3"
   gem "webmock", "~> 3.18"
   gem "with_model"
 end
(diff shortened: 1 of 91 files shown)


frankieroberto:
@peteryates nice! I did some manual checking of the review app, and also looked through the release notes for the versions of govuk-frontend since 4.4.1.  Couldn't spot anything which would need fixing.

IE8 styles and mixing are being deprecated, but it looks like we’re already not building the IE8 stylesheet?

frankieroberto (approved):
Checked the schools interface in the review app and couldn't spot any issues. 👍

ltello (approved):

cwrw (approved):
looks good, some misalignments of search in the drilldown but otherwise looks good

PR #3844, opened 2023-08-17
Add new service classes in importers for create new course

javier-npq:
### Context

Add new service classes in importers for create new course

### Ticket: 

https://dfedigital.atlassian.net/browse/CPDNPQ-1257

Introduced new service classes in importers for create new courses through csv and update `create_npq_contract` service for update or create new contract with specified version. The version is **Finance::Statement** latest version.

### Guidance to review

Looked into create `new_npq_course`, `create_npq_course` as they are inspire from old code and reviewed the private method named `get_latest_version` in **app/services/importers/create_npq_contract.rb** because this method use to get latest version.

### DIFF
--- a/app/services/importers/create_npq_contract.rb
+++ b/app/services/importers/create_npq_contract.rb
@@ -13,7 +13,7 @@ module Importers
 
     def call
       check_headers
-      version = get_latest_version if new_course_flag == true
+      version = Finance::Statement::NPQ.latest.contract_version if new_course_flag == true && Finance::Statement::NPQ.count.positive?
 
       ActiveRecord::Base.transaction do
         rows.each do |row|
@@ -90,10 +90,6 @@ module Importers
 
     def rows
       @rows ||= CSV.read(path_to_csv, headers: true, skip_blanks: true)
-    end
-
-    def get_latest_version
-      Finance::Statement::NPQ.all.order(contract_version: :asc).pluck(:contract_version).last
     end
   end
 end
--- a/app/services/importers/create_npq_contract.rb
+++ b/app/services/importers/create_npq_contract.rb
@@ -4,14 +4,16 @@ require "csv"
 
 module Importers
   class CreateNPQContract
-    attr_reader :path_to_csv
+    attr_reader :path_to_csv, :new_course_flag
 
-    def initialize(path_to_csv:)
+    def initialize(path_to_csv:, new_course_flag: false)
       @path_to_csv = path_to_csv
+      @new_course_flag = new_course_flag
     end
 
     def call
       check_headers
+      version = get_latest_version if new_course_flag == true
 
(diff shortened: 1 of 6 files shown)


app/services/importers/create_npq_contract.rb
```diff
@@ -4,14 +4,16 @@
...
 
-    def initialize(path_to_csv:)
+    def initialize(path_to_csv:, new_course_flag: false)
       @path_to_csv = path_to_csv
+      @new_course_flag = new_course_flag
     end
 
     def call
       check_headers
+      version = Finance::Statement::NPQ.latest.contract_version if new_course_flag == true && Finance::Statement::NPQ.count.positive?
```
cwrw, line 16: I think to get the latest version for a contract we'll prob need the following logic:
```
Finance::Statement::NPQ.where(cohort: , cpd_lead_provider:).pluck(:contract_version).sort.last
```
so might be best moving this into the row block as might be different between providers.
Also do you mind adding a spec for this as well?

cwrw (commented):
looking great, a few comments but otherwise 👍

app/services/importers/create_npq_contract.rb
```diff
@@ -4,14 +4,16 @@
 
 module Importers
   class CreateNPQContract
-    attr_reader :path_to_csv
+    attr_reader :path_to_csv, :new_course_flag
 
-    def initialize(path_to_csv:)
+    def initialize(path_to_csv:, new_course_flag: false)
```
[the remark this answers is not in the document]
javier-npq: Yea, that sounds a lot better

spec/services/importers/create_npq_contract_spec.rb
```diff
@@ -55,6 +56,36 @@
       end
     end
 
+    context "when new version" do
+      let(:set_to_latest_version) { true }
```
cwrw, line 60: suggestion: Maybe if we add 
```
let(:contract_version) { "0.0.3" }
let!(:statement) { create(:npq_statement, cpd_lead_provider:, cohort:, contract_version: contract_version) }
```
and remove line https://github.com/DFE-Digital/early-careers-framework/pull/3844/files#diff-dd0f018c7180c0386edef9776579c09f54dff0c176a6b58f2d9475705459bfe9R74
 The test will be more explicitly testing for the newer version as we currently aren't setting up any statements in the spec

cwrw (approved):
Nicely done ⭐ one small suggestion/nit but otherwise tested the script and all works perfectly, thanks!

spec/services/importers/create_npq_contract_spec.rb
javier-npq: That would be better, let me make add this change to be more specific to newer versions.

PR #3848, opened 2023-08-18
Add steps for restoring PaaS databases on Azure

peteryates:
Add database syncing capability between PaaS and Azure AKS.

The job currently needs to be kicked off manually but the plan is to set it so it runs nightly via cron.

## Tested on

* [x] staging
* [x] sandbox
* [x] production (backup)


cwrw (approved):

Makefile
```diff
@@ -164,3 +164,7 @@ install-konduit: ## Install the konduit script, for accessing backend services
 		&& curl -s https://raw.githubusercontent.com/DFE-Digital/teacher-services-cloud/master/scripts/konduit.sh -o bin/konduit.sh \
 		&& chmod +x bin/konduit.sh \
 		|| true
+
+.PHONY: terraform-refresh
```
cwrw, line 168: out of curiosity why do we need this step?

.github/workflows/backup-paas-database.yml
```diff
@@ -10,31 +10,110 @@ on:
...
         run: |
-          cf conduit ecf-postgres-staging -- pg_dump -E utf8 --clean --compress=1 --if-exists --no-owner --no-privileges --verbose -f backup.sql.gz
+          cf conduit ecf-postgres-${{ github.event.inputs.environment }} -- pg_dump -E utf8 --clean --compress=1 --if-exists --no-owner --no-privileges --verbose -f backup.sql.gz
+
+      # this is temporary while we test restoring production using the backup, re-merge it with
+      # the step above when we're ready to test for real
+      - name: Backup database (production)
+        if: github.event.inputs.environment == 'production'
+        run: |
+          cf conduit ecf-postgres-production-backup-restore -- pg_dump -E utf8 --clean --compress=1 --if-exists --no-owner --no-privileges --verbose -f backup.sql.gz
```
cwrw, line 63: maybe we should use different filenames, as in `sandbox-backup` etc. for each env?

Makefile
peteryates: Not entirely sure, I think we added it while on the call with Colin. Will make a note to check it and we'll remove if unnecessary/unuseful.

Perhaps it was related to the extra steps caused by needing to request PIM approval then log out/in?

.github/workflows/backup-paas-database.yml
cwrw: discussed in person: not necessary as artifact scoped to workflow

peteryates: Leaving for now as the artefacts are scoped by workflow. Will change them if our test run fails.

PR #3855, opened 2023-08-20
CST-1852: add filters to participants dashboard

ltello:
### Context

This year participants are starting to complete their ECF training. 
To accommodate this and allow SITs to differentiate and manage their participants in the various status of induction, we need to make some changes to their school participants dashboard. 

Through this PR we will:
- create a filter in the participants dashboard so SITs can classify and display ECTs in 3 categories: currently training, completion induction and no longer training.
- sort training participants by mentor name or induction start date
- display induction start dates in the listings
- display induction completion dates for participants that have completed ECF training.

### Screenshots

Screenshot 2023-08-22 at 09 43 54Screenshot 2023-08-22 at 09 47 01

Screenshot 2023-08-22 at 09 45 10

Screenshot 2023-08-22 at 09 45 26

Screenshot 2023-08-22 at 09 45 37


peteryates:
The change of layout/design feels a little weird when changing the sorting direction. I'd expect it to show the same information in a different order. I know that's not really possible because we're grouping by mentor - perhaps the 'Induction start date'/'Mentor' should be called something other than 'Sort by'.

An example of this is in GitHub projects where [the same data is displayed in different ways](https://github.com/orgs/alphagov/projects/53/views/30) using tabs (not suggesting we use tabs, just using this as an example):

image

frankieroberto:
@ltello woah you did a lot in this PR! The "sort by" feature wasn't meant to be in scope, that was an idea I added after (sorry for the confusion! 🤦 ).

@peteryates and yeah, it's definitely a bit odd/unexpected for the display to change between the sort types - but it seems like grouping by mentor is useful but there might also been the need to see "newer" ECTs at the top and older ones at the bottom (until they get marked as completed).  Maybe there's a way to combine them, but it could get confusing if a mentor has 2 ECTs with different induction start dates...

We could "hide" the sort by feature for now until we're more confident (maybe do some quick user testing on it) – or we just go for it and then iterate again if we have any further improvements...?

@leoapost I'm not sure what you mean by resetting the filters? Do you mean resetting back to the first radio option ("Currently training")? If so, I'm not sure that's necessary as it's easy to do already. If we end up adding more filters, then a reset option might be worth considering.

app/assets/stylesheets/schools/dashboard.scss
```diff
@@ -7,6 +7,23 @@
...
+    .govuk-summary-list__missing {
+      border-left: 10px solid #1d70b8;
+      padding-left: 15px;
+    }
+    .govuk-summary-list__missing-heading {
+      margin-bottom: 10px;
+    }
+  }
+
+  .filter {
```
frankieroberto, line 19: Could we start using the `app-` prefix for our own components? That’s [recommended by GDS](https://design-system.service.gov.uk/get-started/extending-and-modifying-components/) to avoid naming conflicts. I don't always remember but I'll try to!

```suggestion
  .app-filter {
```

app/assets/stylesheets/schools/dashboard.scss
```diff
@@ -7,6 +7,23 @@
...
+    border: #f3f2f1 4px solid;
```
frankieroberto, line 20:
```suggestion
    border: govuk-colour("light-grey") 4px solid;
```

app/views/schools/participants/_currently_training_sorted_by_induction_start_date.html.erb
```diff
@@ -0,0 +1,43 @@
...
+  <b class="govuk-!-margin-left-2">Induction start date</b>
+</p>
+
+<% participants.ects.each do |ect| %>
+  <%= govuk_summary_card(title: link_to_participant(ect.participant_profile, school)) do |card|
+    card.with_summary_list(
+        rows: [
+            { key: {
+                text: "Status",
+                classes: ["xgovuk-!-font-weight-regular"],
```
frankieroberto, line 17: My bad - this class isn't needed (and doesn't do anything). sorry!
```suggestion
```

app/views/schools/participants/_currently_training_sorted_by_induction_start_date.html.erb
```diff
@@ -0,0 +1,43 @@
...
+                  text: render(StatusTags::SchoolParticipantStatusTag.new(
+                      participant_profile: ect.participant_profile,
+                      induction_record: ect.induction_record,
+                      display_description: false,
+                      display_induction_start_date: true)),
+              },
+            },
+            { key: {
+                text: "Mentor",
+                classes: ["xgovuk-!-font-weight-regular"],
```
frankieroberto, line 29: This isn't needed.
```suggestion
```

app/views/schools/participants/_currently_training_sorted_by_induction_start_date.html.erb
```diff
@@ -0,0 +1,43 @@
+<p class="govuk-body">
+  Sort by:
+  <%= govuk_link_to("Mentor (A-Z)",
+                    school_participants_path(school, sorted_by: :mentor),
+                    no_visited_state: true,
+                    classes: ["govuk-!-margin-left-2"]) %>
```
frankieroberto, line 6: The `classes` param isn’t working here - think this might have been a recent change to the components gem (@peteryates?). Looks like it should be `class` instead.

```suggestion
  <%= govuk_link_to("Mentor (A-Z)",
                    school_participants_path(school, sorted_by: :mentor),
                    no_visited_state: true,
                    class: "govuk-!-margin-left-2") %>
```

app/views/schools/participants/_currently_training_sorted_by_mentor.html.erb
```diff
@@ -0,0 +1,69 @@
+<p class="govuk-body">Sort by:
+  <b class="govuk-!-margin-left-2">Mentor (A-Z)</b>
+  <%= govuk_link_to("Induction start date",
+                    school_participants_path(school, sorted_by: :induction_start_date),
+                    no_visited_state: true,
+                    class: ["govuk-!-margin-left-2"]) %>
+</p>
```
frankieroberto, line 7: Rather than duplicating this across both partials (which does work!) would it be better to extract it into its own mini component, and then tell that component which sort state is currently active? Just an idea!

app/views/schools/participants/index.html.erb
```diff
@@ -5,107 +5,55 @@
...
-                       display_description: false)),
-                     classes: ["govuk-\!-padding-bottom-static-0"]) %>
-              <% end %>
-            <% end %>
-          <% end %>
-        </div>
-        <br>
+<div class="govuk-grid-row wrapper-dashboard-v2">
+  <div class="govuk-grid-column-one-third">
+    <div class="filter">
```
frankieroberto, line 29:
```suggestion
    <div class="app-filter">
```

app/views/schools/participants/index.html.erb
```diff
@@ -5,107 +5,55 @@
...
+      <%= form_for @filter, as: '', url: school_participants_path(@school), method: :get do |f| %>
+        <%= f.govuk_error_summary %>
```
frankieroberto, line 31: I don't think we need this error summary - as it should be impossible to get the form in an error state (as the first radio option is always selected by default)?

```suggestion
```

app/views/schools/participants/index.html.erb
```diff
@@ -5,107 +5,55 @@
...
+        <%= f.govuk_collection_radio_buttons(
+                :filtered_by,
+                @filter.filter_options,
+                :id,
+                :label,
+                legend: { text: 'Show', tag: 'h2', size: 's' },
+                classes: ["govuk-\!-padding-right-1"],
```
frankieroberto, line 38: Don't think this class is needed.

```suggestion
```

app/views/schools/participants/_currently_training_sorted_by_mentor.html.erb
ltello: Good idea! That's what components are for!.

leoapost:
>@leoapost I'm not sure what you mean by resetting the filters? Do you mean resetting back to the first radio option ("Currently training")? If so, I'm not sure that's necessary as it's easy to do already. If we end up adding more filters, then a reset option might be worth considering.

@frankieroberto In my mind, filters are optional. You see all the data and then you filter them down and then you reset/remove the filters to see all the data again.

I presume the new design allows one filter at a time for simplicity reasons, so the "reset" could be a `Show all` option to display all the results.

If I was an SIT with a few participants, I would want to see them all on the same page. But, please ignore me if that's not the intention of the new design.

frankieroberto:
> > @leoapost I'm not sure what you mean by resetting the filters? Do you mean resetting back to the first radio option ("Currently training")? If so, I'm not sure that's necessary as it's easy to do already. If we end up adding more filters, then a reset option might be worth considering.
> 
> @frankieroberto In my mind, filters are optional. You see all the data and then you filter them down and then you reset/remove the filters to see all the data again.
> 
> I presume the new design allows one filter at a time for simplicity reasons, so the "reset" could be a `Show all` option to display all the results.
> 
> If I was an SIT with a few participants, I would want to see them all on the same page. But, please ignore me if that's not the intention of the new design.

Ah! So we’ve actually removed the ability to "show all".  We did start this design by having checkboxes not radios, so you could see everyone by ticking all 3 checkboxes, however in research everyone only wanted to see one 'view' at a time, so I switched it to radios to make this easier (saves having to uncheck one box and check another box). 

Showing everything at once is also slightly complicated by having a different design for each view (one is grouped by mentor and 'no longer training' uses a list instead of cards) and different sort options.

We should keep this under review though, and if we see a strong need for schools to be able to see all participants at once we can reconsider!

app/assets/stylesheets/schools/dashboard.scss
```diff
@@ -7,6 +7,23 @@
...
+
+  .app-filter {
+    border: govuk-colour("light-grey") 4px solid;
+    padding: 20px;
+    margin-bottom: 20px;
+
+    .govuk-radios__label {
+      padding-right: 5px;
+    }
   }
```
frankieroberto, line 27: This is super picky, but could this CSS get moved to something like `assets/stylesheets/filter.scss` (or `assets/stylesheets/schools/filter.scss`), even though it's only a few lines?  That might help us find it again in future, and avoid any issues if we add any other filters.

app/assets/stylesheets/schools/dashboard.scss
```diff
@@ -7,6 +7,23 @@
     .govuk-summary-list__actions {
       width: 13%;
     }
+    .govuk-summary-list__missing {
+      border-left: 10px solid #1d70b8;
+      padding-left: 15px;
+    }
+    .govuk-summary-list__missing-heading {
+      margin-bottom: 10px;
+    }
```
frankieroberto, line 16: Move to `assets/stylesheets/summary-list.scss`? I can see us re-using these styles on other pages than the dashboard.

Also I think we should change the prefix from `govuk-` to `app-` to make it clearer that it's app-specific styles rather than from upstream `govuk-frontend` package.

```suggestion
    .app-summary-list__missing {
      border-left: 10px solid #1d70b8;
      padding-left: 15px;
    }
    .app-summary-list__missing-heading {
      margin-bottom: 10px;
    }
```

app/helpers/schools/dashboard_helper.rb
```diff
@@ -25,5 +29,14 @@ def mentor_count(school_cohorts)
...
       school_cohorts.sum { |sc| sc.current_induction_records.count }
     end
+
+    def missing_mentor_html(participant_profile)
+      tag.div(class: "govuk-summary-list__missing") do
+        concat tag.div("No mentor assigned", class: "govuk-summary-list__missing-heading")
+        concat govuk_link_to("Assign a mentor",
+                             school_participant_edit_mentor_path(participant_id: participant_profile.id),
+                             no_visited_state: true)
+      end
```
frankieroberto, line 39: My bad - we should use the `app-` prefix here for our custom styles.

```suggestion
    def missing_mentor_html(participant_profile)
      tag.div(class: "app-summary-list__missing") do
        concat tag.div("No mentor assigned", class: "app-summary-list__missing-heading")
        concat govuk_link_to("Assign a mentor",
                             school_participant_edit_mentor_path(participant_id: participant_profile.id),
                             no_visited_state: true)
      end
```

frankieroberto (approved):
Tested in the review app and it works great!

Thanks for removing the filter options where there are 0 participants, and for removing the filter altogether where there’s only 1 option that applies.

Left a couple of final minor comments about CSS naming conventions (which are my fault - sorry!).

app/assets/stylesheets/schools/dashboard.scss
ltello: done

frankieroberto: Thank you!

leoapost (approved):

app/helpers/schools/dashboard_helper.rb
```diff
@@ -14,6 +10,14 @@ def ect_with_no_mentor_count(school_cohorts)
       school_cohorts.sum { |sc| sc.current_induction_records.ects.where(mentor_profile: nil).count }
     end
 
+    def link_to_participant(participant_profile, school)
+      govuk_link_to(participant_profile.full_name,
+                    school_participant_path(id: participant_profile.id,
+                                            school_id: school.slug),
+                    no_visited_state: true,
+                    class: "govuk-!-font-weight-regular")
```
frankieroberto, line 18: @ltello we’ve decided to keep this link as bold, to make it a bit clearer that this is the header for the card.

```suggestion
                    no_visited_state: true
```

app/views/schools/participants/_currently_training_sorted_by_induction_start_date.html.erb
```diff
@@ -0,0 +1,33 @@
...
+                      display_induction_start_date: true)),
+              },
+            },
+            { key: {
+                text: "Mentor",
+            },
+              value: {
+                  text: (
+                      if ect.mentored?
+                        link_to_participant(ect.mentor, school)
```
frankieroberto, line 24: @ltello can we change this to just show the mentor’s full name please, rather than being a link? Thanks!

frankieroberto (changes requested):
2 design changes following Nicola’s helpful usability feedback that it could be confusing clicking the mentor name link expecting it to take you to the ECT page:

* make the ECT name bold in the card header
* remove the link to the mentor page and just show their name instead

Thanks!

frankieroberto (approved):

PR #3858, opened 2023-08-21
Remove induction start date job from the cron schedule

tonyheadford:
### Context

Remove the `SetParticipantStartDateJob` from the schedule as it has now updated all that it can in the given criteria.  Not removing the actual job now as that may be still required (perhaps with a tweak to the query) in the near future.

Remove the cron job from the schedule


leoapost (commented):
Many times I'm inclined to remove code that's nowhere used.

Should we disable it instead with a comment to provide some context?

tonyheadford:
> Many times I'm inclined to remove code that's nowhere used.
> 
> Should we disable it instead with a comment to provide some context?

Yes, good idea

leoapost (approved):

PR #3866, opened 2023-08-22
Make destroy review app action run for all PRs

peteryates:
### Context

We loosened the criteria for review apps to be generated from having to have the 'review-app' label to simply not being generated by Dependabot to prevent problems where the app won't boot.

This change sets the same critera in the destory action.

Refs #3857

### DIFF
--- a/.github/workflows/aks_destroy_review.yml
+++ b/.github/workflows/aks_destroy_review.yml
@@ -9,7 +9,7 @@ jobs:
   delete-review-app:
     name: Delete Review App ${{ github.event.pull_request.number }}
     concurrency: deploy_review_${{ github.event.pull_request.number }}
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'review-app') }}
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'dependencies') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/aks_destroy_review.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   delete-review-app:
     name: Delete Review App ${{ github.event.pull_request.number }}
     concurrency: deploy_review_${{ github.event.pull_request.number }}
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'review-app') }}
+    if: github.actor != 'dependabot[bot]' && github.event_name == 'pull_request'
```
cwrw, line 12: I think we could probably use `${{ !contains(github.event.pull_request.labels.*.name, 'dependencies') }}` (like the old workflows) as we're merging the prs so the actor will not be dependabot

peteryates: Have updated accordingly. Feels a bit weird not detecting the originating user directly but there's no obvious clean way of doing that.

`github.event.pull_request.user.login` seemed like it might.

cwrw (approved):

PR #3871, opened 2023-08-23
Fix the scheduled database backup/restore task

peteryates:
### Context

The run failed previously because this flow calls other flows and the repo hadn't been checked out.

```
Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/early-careers-framework/early-careers-framework/.github/workflows/backup-paas-database.yml'. Did you forget to run actions/checkout beforerunning your local action?
```

- Ensure code checked out before calling other flows
- Allow scheduled backup to be despatched manually


.github/workflows/backup-paas-schedule.yml
```diff
@@ -2,6 +2,7 @@ name: Run scheduled database backups
 on:
   schedule:
     - cron: "30 4 * * *"
+  workflow_dispatch:
```
cwrw, line 5: do we need inputs here as well?

cwrw: ok as looks strange without anything but let's try

PR #3875, opened 2023-08-23
[CST-1835] Allow ECTs withdrawn for the current school in a past cohort to be reinstated

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CST-1835

- Update the AddParticipants form to check if the pre-existing participant profile is for the current school and withdrawn, if it is, reactivate it instead of raising an error.

### DIFF
--- a/app/forms/schools/add_participants/base_wizard.rb
+++ b/app/forms/schools/add_participants/base_wizard.rb
@@ -190,17 +190,23 @@ module Schools
         using_sit_email? && participant_type == "ect"
       end
 
+      def email_owner_participant_profile
+        @email_owner.teacher_profile&.ecf_profiles&.first
+      end
+
+      def email_owner_induction_record_for_current_school
+        return if email_owner_participant_profile.blank?
+
+        Induction::FindBy.call(
+          participant_profile: email_owner_participant_profile,
+          school:,
+        )
+      end
+
       def email_used_in_the_same_school?
-        if email_in_use?
-          participant_profile = @email_owner.teacher_profile&.ecf_profiles&.first
-          if participant_profile
-            Induction::FindBy.call(participant_profile:)&.school == school
-          else
-            false
-          end
-        else
-          false
-        end
+        return false unless email_in_use?
+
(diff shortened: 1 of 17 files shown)


app/forms/schools/add_participants/base_wizard.rb
```diff
@@ -160,22 +160,36 @@ def email_in_use?
...
+        return false if email_owner_participant_profile_withdrawn_from_current_school?
+        return false if email_owner_induction_record_for_current_school&.training_status_withdrawn?
+
         return true unless transfer?
 
         @email_owner != existing_user
       end
 
+      def email_owner_participant_profile
+        @email_owner.teacher_profile&.ecf_profiles&.first
```
tonyheadford, line 172: `.ecf_profiles` doesn't return withdrawn ones so perhaps `@email_owner.teacher_profile.participant_profiles.ecf` would give you their profiles (they _could_ have more than one though - e.g. ECT and Mentor or active and withdrawn)

tonyheadford (commented):
Quite a lot of complexity in this one that I think you've sorted out well.  Just wanted to check whether this was this only intended for reinstating ECTs not mentors and had you catered for a person that might have multiple profiles where there may be an active and a withdrawn profile of the same type (left over legacy stuff but still out there sadly)?
I also left a comment about navigating to

app/forms/schools/add_participants/base_wizard.rb
Lockyy: Ah yeah, that may be better. Here at least I've just replicated exactly what was being used in the now refactored `email_used_in_the_same_school?` method. I didn't want to make any changes at all to any other logic so just broke it up so that this was available to use in `email_owner_participant_profile_withdrawn_from_current_school?`.

tonyheadford: Think that wasn't (probably incorrectly) considering withdrawn profiles though

leoapost (approved):
LGTM

tonyheadford (approved):
LGTM well done for persisting with it

PR #3878, opened 2023-08-24
Display the AB in participant's profile even when schools don't have any

leoapost:
This helps with transferred participants that are associated to an Appropriate  Body and want to keep it.

### Context

- Ticket: CST-1803

Remove the school cohort AB check from the `can_appropriate_body_be_changed?` helper method

### Guidance to review
- Find a school with participants in a cohort without an Appropriate Body
- Visit the page of one of those participants
- The Appropriate Body row should be displayed

### DIFF
--- a/app/controllers/schools/participants_controller.rb
+++ b/app/controllers/schools/participants_controller.rb
@@ -158,6 +158,6 @@ private
   end
 
   def can_appropriate_body_be_changed?
-    @induction_record.school_cohort.appropriate_body.present? && @profile.ect? && !@induction_record.training_status_withdrawn?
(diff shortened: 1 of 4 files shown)


frankieroberto (changes requested):
Tested in the review app and the row appears. 👍 

Can I ask for a couple of quick amends:

1. Add the placeholder text "Not assigned" to the value column, rather than having an empty space.
2. Add visually-hidden text " appropriate body" within the "Add" link, so that the link will be read as "Add appropriate body" for screen reader users. The "Change" links should also have this, so it reads as "Change appropriate body".

Thanks!

frankieroberto (approved):

PR #3887, opened 2023-08-25
Update deduplicator to handle additional scenarios

ethax-ross:
### Context

We can now handle duplicates with a different lead provider but the same school - providing there are no declarations on the duplicate - by destroying the duplicate profile _without_ reconciling it with the primary.

After working through a change in training programme with Nathan on a call we've decided we can probably handle these but they need a warning to ensure we double check the ordering and scrutinise them more closely.

- Update deduplicator to handle additional scenarios

### DIFF
--- a/lib/participant_profile_deduplicator.rb
+++ b/lib/participant_profile_deduplicator.rb
@@ -16,21 +16,24 @@ class ParticipantProfileDeduplicator
   def dedup!
     @changes = []
 
+    prevent_same_school_different_lead_provider_with_declarations
+
     log_overview_info
 
     warning_ect_mentor_duplicate
     warning_only_void_declarations_on_primary
-
-    prevent_same_school_different_lead_provider
-    ensure_training_programmes_match
+    warning_training_programmes_different
 
     ActiveRecord::Base.transaction do
-      reconcile_schedules!
-      reconcile_declarations!
-      transfer_validation_data!
-      transfer_participant_eligibility!
-      handle_school_change!
-      reconcile_remaining_induction_records!
+      if reconcile?
+        reconcile_schedules!
+        reconcile_declarations!
+        transfer_validation_data!
+        transfer_participant_eligibility!
+        handle_school_change!
+        reconcile_remaining_induction_records!
+      end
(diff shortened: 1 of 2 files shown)


lib/participant_profile_deduplicator.rb
```diff
@@ -48,17 +51,29 @@ def log_overview_info
...
   end
 
-  def prevent_same_school_different_lead_provider
-    return unless primary_profile.lead_provider&.id != duplicate_profile.lead_provider&.id && primary_profile.school == duplicate_profile.school
+  def reconcile?
+    !same_school_different_lead_provider? || duplicate_has_declarations?
+  end
+
+  def same_school_different_lead_provider?
+    primary_profile.lead_provider&.id != duplicate_profile.lead_provider&.id && school(primary_profile) == school(duplicate_profile)
```
cwrw, line 59: I think we might need to do the same here around looking at the latest IR to find the lead provider, similar to schools?

lib/participant_profile_deduplicator.rb
```diff
@@ -48,17 +51,29 @@ def log_overview_info
     log_info("Duplicate profile: #{duplicate_profile_id}")
   end
 
-  def prevent_same_school_different_lead_provider
-    return unless primary_profile.lead_provider&.id != duplicate_profile.lead_provider&.id && primary_profile.school == duplicate_profile.school
+  def reconcile?
+    !same_school_different_lead_provider? || duplicate_has_declarations?
```
cwrw, line 55: Im wondering around the second check for declarations, so let's say the duplicate/primary have the same lead provider, but different schools, and duplicate has no declarations, I think we'll probably want to run through some steps, unless I misunderstood the check 🤔

ethax-ross: I thought if we had the same school but different lead provider and no declarations on the duplicate then we wanted to merge without reconciling any data? and if we have the same school but different lead providers and the duplicate has declarations it will raise an error. Have I misunderstood?

cwrw (approved):

PR #3898, opened 2023-08-30
Remove staging/development environment workflows

peteryates:
Staging has been recreated in Azure and Development is no longer needed.


cwrw (commented):
Thanks for this, we will need to maintain `deploy_to_dev` as well as the tag creation in staging. I'll move them to relevant named files to avoid confusion

cwrw (approved):

peteryates (commented):
Additional commits look good, happy to see this go in.

PR #3900, opened 2023-08-30
Remove the `deployed_development` environment

peteryates:
### Context

- Remove all traces of the deployed_development env
- Update the README with the new environments


cwrw:
@peteryates thanks for this, added two new commits if you can have a look. I think we could probably merge https://github.com/DFE-Digital/early-careers-framework/pull/3898 before this one as some envs reference deployed_dev there as well

cwrw (approved):

peteryates:
New commits look good, thanks for tidying those up. Gotta catch 'em all! 😄

PR #3846, opened 2023-08-17
Rewrite the cohort components

peteryates:
### Context

The admin cohort components were a mess and full of bugs. This replaces them with a version that uses ViewComponent's slots and adds specs.

### Changes proposed in this pull request

- Remove old cohort info components
- Switch over the rendered cohort component
- Add cohort, partnership and relationship components
- Add specs for new cohort components

### Preview

| Before | After |
| ------- | -------- |
| image |image |

When logged in as a super user the 'Challenge relationship' link will be visible

Screenshot from 2023-08-21 09-48-13

PR #3908, opened 2023-08-30
Adjust the cohort component behaviour

peteryates:
### Context

In my previous attempt to rewrite these components (#3846) I misunderstood the role of partnerships/relationships in CIP/FIP. Only FIP schools have partnerships and relationships; CIP schools instead have materials provided by their core induction programme.

As a result I've had to move some of the functionality previosuly in `Schools::PartnershipComponent` into `Schools::CohortComponent`.

Hopefully this makes more sense 😅

image

### DIFF
--- a/app/components/admin/schools/cohort_component.rb
+++ b/app/components/admin/schools/cohort_component.rb
@@ -19,12 +19,28 @@ module Admin
         build_relationships
       end
 
+      def training_programme
+        {
+          "core_induction_programme" => "Using DfE-accredited materials",
+          "design_our_own"           => "Designing their own training",
+          "no_early_career_teachers" => "No ECTs this year",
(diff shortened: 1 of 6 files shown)


app/components/admin/schools/cohort_component.rb
```diff
@@ -38,6 +56,34 @@ def build_relationships
           with_relationship_component(school:, school_cohort:, relationship:, superuser:)
         end
       end
+
+      def cip?
+        induction_programme_choice == "core_induction_programme"
```
tonyheadford, line 61: These could be `school_cohort.core_induction_programme?` and `school_cohort.full_induction_programme?` instead?

tonyheadford (approved):
LGTM, one possible change to use the enum helpers and reduce a bit of code if you prefer it

PR #3918, opened 2023-08-31
Do not merge: using rspec helpers for govuk components

frankieroberto:


app/helpers/view_helper.rb
```diff
@@ -0,0 +1,7 @@
+module ViewHelper
+
+  def title_with_error_prefix(title, error:)
+    "#{"Error: " if error}#{title}"
+  end
```
peteryates, line 5: Definitely room for this in the govuk-components helpers. Every service will need something like it.

frankieroberto: Yeah, I wasn't sure whether it'd be worth making as component or not, as it's such a small amount of code. Maybe?

PR #3939, opened 2023-09-06
[CST-1924] Send access info when unregistered users are trying to login

leoapost:
### Context

- Ticket: CST-1924

Unregistered users try to sign into the service and they get confused. We want to send them an email about what they should do to gain access.

- Add a new user mailer
- Update the passwordless strategy to send the email when the user is not found

### Guidance to review
- Try to sign-in with a random email address
- Verify that the email is sent by checking on `Notify -> API integration`
- Confirm the email body is correct (mainly the URLs passed to the Notify template)

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -2,6 +2,7 @@
 
 class UserMailer < ApplicationMailer
   SIGN_IN_EMAIL_TEMPLATE = "7ab8db5b-9842-4bc3-8dbb-f590a3198d9e"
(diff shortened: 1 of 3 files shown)


Lloyd-K:
That works for me. @leoapost I can update the Notify template now if you like? Thanks

leoapost:
@frankieroberto I've updated the template and pushed another commit to pass the 3rd link to the template as well.

frankieroberto (approved):

tonyheadford (approved):

PR #3941, opened 2023-09-06
[CST-1828] Add job and schedule to pull in completion dates from DQT

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-1828)

Pull in the induction completion dates for pre-2023 ECTs as a one time job.  The job is scheduled to run every 2 minutes from midnight to 9am and pull in 200 participants at a time so that we don't have/cause issues with the DQT API 300 requests per minute limit.

Add a job to read the induction records from DQT and use the `Induction::Complete` service to update the participants records.

### DIFF
--- a/app/jobs/set_participant_completion_date_job.rb
+++ b/app/jobs/set_participant_completion_date_job.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# This is a one off job to update all of the induction completion dates from DQT for
+# all of the 2021/2022 ECTs that do not currently have them.
+# It runs in batches of 200 as we have a 300 lookup per minute limit on the API
+# This could get adapted later to be a regular running process with a bit more logic
+class SetParticipantCompletionDateJob < ApplicationJob
+  def perform
+    ParticipantProfile::ECT
+      .eligible_status
+      .joins(:teacher_profile)
+      .includes(:teacher_profile)
+      .where.not(teacher_profile: { trn: nil })
+      .where.not(induction_start_date: nil)
+      .where(created_at: ...Cohort.find_by(start_year: 2023).registration_start_date)
+      .order(:updated_at)
+      .limit(200)
+      .each do |participant_profile|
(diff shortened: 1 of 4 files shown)


app/jobs/set_participant_completion_date_job.rb
```diff
@@ -0,0 +1,39 @@
...
+# This could get adapted later to be a regular running process with a bit more logic
+class SetParticipantCompletionDateJob < ApplicationJob
+  def perform
+    ParticipantProfile::ECT
+      .eligible_status
+      .joins(:teacher_profile)
+      .includes(:teacher_profile)
+      .where.not(teacher_profile: { trn: nil })
+      .where.not(induction_start_date: nil)
+      .where(created_at: ...Cohort.find_by(start_year: 2023).registration_start_date)
```
ltello, line 15: So this is only for participants registered in 2023?

ltello, line 15: why exclude ects registered this year? It will become invalid next year or the next one

app/jobs/set_participant_completion_date_job.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# This is a one off job to update all of the induction completion dates from DQT for
+# all of the 2021/2022 ECTs that do not currently have them.
+# It runs in batches of 200 as we have a 300 lookup per minute limit on the API
+# This could get adapted later to be a regular running process with a bit more logic
```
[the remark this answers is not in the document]
tonyheadford: One off as in it needs to run small batches multiple times to avoid rate limiting but will be removed/refactored/disabled once it's had chance to complete (maybe one or two nights as it was for the induction start dates). There will need to be a different process with maybe different query logic to run something periodically to update ECTs later.

app/jobs/set_participant_completion_date_job.rb
tonyheadford: This is only for backfilling for now.

ltello (approved):

PR #3956, opened 2023-09-07
Make it clearer you're adding ECT or mentor

frankieroberto:
🗂️ [Jira card](https://dfedigital.atlassian.net/browse/CST-1966)

We had some people contact support because they'd accidentally added an ECT as a mentor or vice versa.

We think this is partly because the page after selector "ECT" or "Mentor" is generic to both.  By updating this page to be more specific (in the page title and the note about privacy), we hope to make it more obvious to users when they’ve mis-clicked, so that they can quickly go back and select the right option.

### Before

Screenshot 2023-09-11 at 17 58 58

### After

| ECTs | Mentors |
|------|---------|
| Screenshot 2023-09-11 at 17 59 35 | Screenshot 2023-09-11 at 17 59 51 |


peteryates (approved):
Definite improvement

PR #3966, opened 2023-09-11
[CST-1924] Fix the unregistered users mailer

leoapost:
### Context

- Ticket: CST-1924

This PR fixes the issue introduced in the #3939 and re-enables the email.

The issue:
Calling the `. associate_with (recipient) ` in the mailer with a String param raised an exception after Notify sent the email, which failed and re-queued the job. That created a loop that was sending the emails multiple times to the unregistered users who were trying to login.

- There's no real need to associate the email with the recipient's email, so I'm removing the`.associate_with(recipient)` call
- Re-enabling the mailer

### Guidance to review

- Try to sign-in with a random email address
- Verify on `Notify -> API integration` page that the email was sent only once
- Check the mailer job in not re-queued on sidekiq

### DIFF
--- a/lib/devise/strategies/passwordless_authenticatable.rb
+++ b/lib/devise/strategies/passwordless_authenticatable.rb
@@ -40,7 +40,7 @@ module Devise
             UserMailer.with(email: email.downcase, full_name: user.full_name, url:, token_expiry: token_expiry.localtime.to_fs(:govuk)).sign_in_email.deliver_later(queue: "priority_mailers")
             raise LoginIncompleteError
           else
-            # UserMailer.with(recipient: email.downcase).access_info_email.deliver_later(queue: "priority_mailers")
+            UserMailer.with(recipient: email.downcase).access_info_email.deliver_later(queue: "priority_mailers")
             raise EmailNotFoundError
           end
         end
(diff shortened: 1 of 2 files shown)


lib/devise/strategies/passwordless_authenticatable.rb
```diff
@@ -40,7 +40,7 @@ def authenticate!
             UserMailer.with(email: email.downcase, full_name: user.full_name, url:, token_expiry: token_expiry.localtime.to_fs(:govuk)).sign_in_email.deliver_later(queue: "priority_mailers")
             raise LoginIncompleteError
           else
-            # UserMailer.with(recipient: email.downcase).access_info_email.deliver_later(queue: "priority_mailers")
+            UserMailer.with(recipient: email.downcase).access_info_email.deliver_later(queue: "priority_mailers")
```
ltello, line 43: would it be worth passing the user instance as recipient so the mailer can do `.associate_with(recipient)` but we pass only the user.email to the template?

ltello: @leoapost fair enough!

ltello (approved):

PR #3987, opened 2023-09-13
Add statuscake monitoring to AKS

cwrw:
### Context
Terraform statuscake module: https://github.com/DFE-Digital/terraform-modules/tree/main/monitoring/statuscake

- Ticket: n/a
We have statuscake monitoring in PaaS, add it to Azure since we are migrating soon

- Add external hostname variable and set in all envs 
- Amend monitoring to true in production as this is used in both db setup and statuscake
- Add monitoring and set our contact group
- Add statuscake to providers and set token in action

### Guidance to review
Did I miss anything?


peteryates (approved):
Looks great

RMcVelia (approved):

PR #3988, opened 2023-09-13
Add back missing variables, these are required for sending emails

cwrw:
### Context
We are getting errors in sandbox when attempting to send emails:
```
`full_url_for': Missing host to link to! Please provide the :host parameter, set default_url_options[:host], or set :only_path to true (ArgumentError)
```
We set the domain in environment config, this is required for sending emails.

- Ticket: n/a

- Add send email to ENV to allow redirect in non prod/test envs (needs amending for the vcap attribute later)
- Add domain variable and set in application to allow emails to be sent

### Guidance to review

There is a bit of cleanup still required after the migration

### DIFF
--- a/terraform/aks/application.tf
+++ b/terraform/aks/application.tf
@@ -25,6 +25,10 @@ module "application_configuration" {
     GIAS_API_SCHEMA     = "https://ea-edubase-api-prod.azurewebsites.net/edubase/schema/service.wsdl"
     GIAS_EXTRACT_ID     = 13904
     GIAS_API_USER       = "ecftech"
+    DOMAIN              = var.domain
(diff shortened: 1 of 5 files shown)


terraform/aks/application.tf
```diff
@@ -1,6 +1,7 @@
 locals {
   environment  = "${var.app_environment}${var.app_suffix}"
```
peteryates, line 2: Not one fornow but I think it'd be clearer if we added the `-` to the suffix here instead of when doing `F_VAR_app_suffix=-$(PULL_REQUEST_NUMBER)`, it's easily lost in the assignment

peteryates (approved):

PR #3989, opened 2023-09-13
CST-1961: allow sits to be registered with an existing email address

ltello:
### Context

We’re getting a lot of support tickets around schools not being able to add / change a SIT as the email address is already in use. 

We suspect this may be because the new SIT is currently (or at some point has been) an NPQ participant, ECT or mentor. We currently have a journey for the active SIT to register themselves as a mentor. This could be an issue of the reverse happening i.e. a participant is trying to register themselves as a SIT.

The policy states that any of these participant types can be a SIT so we should not be blocking nomination of the SIT and they should be able to use the same email. 

[Ticket](https://dfedigital.atlassian.net/browse/CST-1961)

 - Refactor `NominateInductionTutorForm` so that:
    -  its sole purpose is to nominate/replace a school SIT
    - allows any other person (no matter their profiles) to be nominated
    - do not allow the nomination of a person already a SIT at the same school
    - requires the email and full_name of the new person to match
    
- Create `EditInductionTutorForm` so that:
   - its sole purpose is to edit the details of a SIT
   - do not allow the new email to belong to a different user
   - only allow names matching the first name of the current name
      
 - Adjust:
   - school controller to nominate a SIT (`Nominations::NominateInductionCoordinatorController`), 
   - school controller to replace a SIT (`Schools::ChangeSitController`),
   - admin controller to nominate/replace a SIT and edit their details (`Schools::InductionCoordinatorsController`) 
   so that they consume the service objects mentioned above.
 
- Remove legacy admin induction coordinators controller not referenced throughout the service.
- Extract `NameMatcher` from `DQTRecordCheck` class to compare two names.

### Caveats
This fix should cover most of the cases we currently have on Zendesk from schools not being able to nominate/replace SITs or edit their details. However, a few facts still can make this to keep happening:
  - The school dashboard currently only have a single feature to allow a SIT to replace himself as school SIT but they can't change their own details (which is one of the types of issues we find in Support). 
  - The lack of feature for a SIT to change their own email can have bad consecuencies if they try and do so using the current feature to replace themselves with a new person. Actually they would be creating a second user for themselves instead.
  - It would be very beneficial to add that missing feature to the school dashboard ASAP.
  - There are still users with 2 profiles, so even through the admin console might be impossible to update a SIT with an email from their other user.
    
### Guidance to review
 - Try and nominate a school SIT for the first time
 - Try and replace a school SIT from the school dashboard
 - Try and nominate a school SIT from admin console
 - Try and replace a school SIT from admin console
 - Try and change a school SIT's details from admin console
 
Play with different names and emails to check the service behaviour and error messages returned.


app/controllers/admin/schools/induction_coordinators_controller.rb
```diff
@@ -49,20 +45,11 @@ def set_school
...
-        {}
-      end
-    end
-
-    def school_using_this_email(email)
-      Identity.find_user_by(email:).schools.first
+      {
+        induction_tutor: @school.induction_tutor,
+        email: @school.induction_tutor&.email,
+        full_name: @school.induction_tutor&.full_name,
```
tonyheadford, line 51: Could these use the attr_reader for `school` now?

app/forms/nominate_induction_tutor_form.rb
```diff
@@ -5,70 +5,94 @@ class NominateInductionTutorForm
...
-    if email_already_taken?
-      other_school = school_using_this_email(email)
-      if other_school.present?
-        errors.add(:email, "The email address #{email} is already used by a user at #{other_school.name}")
-      else
-        errors.add(:email, "The email address #{email} is already in use")
-      end
+  def name_matches
+    if different_name?
+      errors.add(:full_name, "A user with a different name (#{user.full_name}) has already been registered with this email address. Please, change the name or email you have entered.")
```
tonyheadford, line 56: Perhaps run by @Lloyd-K (unless you have already) as I didn't think we're supposed to use "Please"

app/forms/nominate_induction_tutor_form.rb
```diff
@@ -5,70 +5,94 @@ class NominateInductionTutorForm
...
+    errors.add(:email, "The user with email #{user.email} is already an induction coordinator at #{school.name}") if sit_at_school?
+  end
+
+  def remove_current_sits
+    school.induction_coordinators.each do |sit|
+      next sit.induction_coordinator_profile.schools.delete(school) if sit.schools.count > 1
+      next sit.induction_coordinator_profile.destroy! if sit.teacher_profile.present? || sit.npq_registered?
+
+      sit.destroy!
     end
```
tonyheadford, line 70: Should we be removing all SITs in this instance?  Would that be the desired outcome of nominating a new SIT?

app/services/name_matcher.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class NameMatcher
```
tonyheadford, line 3: Nice! The class name might be a little ambiguous but I can't think of anything better

app/controllers/admin/schools/induction_coordinators_controller.rb
ltello: I can do that but for consistency I also should change the rest of ocurrencies in the controller.

app/forms/nominate_induction_tutor_form.rb
ltello: Yes, I was expecting Nicola/Benedict to complain about the error messages 😄

app/forms/nominate_induction_tutor_form.rb
ltello: Why only the first one?

tonyheadford: Exactly. I just wanted to ensure we knew why you were removing all, as I don't think we have much option here at the moment.

app/forms/nominate_induction_tutor_form.rb
Lloyd-K, line 56:
```suggestion
      errors.add(:full_name, "A user with a different name (#{user.full_name}) has already been registered with this email address. Change the name or email address you entered.")
```

Lloyd-K: Are we OK to play back the name to this user?

frankieroberto: I don't quite understand why we need people to enter a matching name as well as the email. Can't we just match on the email and ignore the name, and re-use the existing profile?

tonyheadford (approved):
Amazing job, thanks @ltello

app/controllers/admin/schools/induction_coordinators_controller.rb
ltello: Done

app/forms/nominate_induction_tutor_form.rb
ltello: @frankieroberto I would also remove it but that is a question for Nicola.

ltello: I have made my own commit for this suggestion because there were other replicas of this messages to update as well.

frankieroberto: Let's ask!

PR #3990, opened 2023-09-13
Remove ability to upload CSV files to AWS

cwrw:
### Context
We don't have access to AWS in azure, remove uploading csvs as we instead have a uploaded_urns column which has been backfilled fully in prod
- Ticket: n/a

- Remove tables associated with uploading csvs
- Remove gem
- Remove initialiser setup for AWS
- Remove old jobs to sync urns with uploaded urns as well as methods
- Cleanup tests to read directly from the column

### Guidance to review
Due to time constraints we can revisit to re-add the cleanup code for urns


config/environments/production.rb
```diff
@@ -39,7 +39,7 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
-  config.active_storage.service = :amazon
+  config.active_storage.service = :local
```
peteryates, line 42: I think these lines can go altogether, will do that in a followup PR

peteryates (approved):

PR #3993, opened 2023-09-14
[CST-1961] Add rake tasks to easily update/create SITs for schools

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CST-1961

There's a blocker on the form that prevents users from being updated. This allows you to do that via a rake task

### DIFF
--- a/app/services/support/school_induction_tutors/replace.rb
+++ b/app/services/support/school_induction_tutors/replace.rb
@@ -1,9 +1,11 @@
+# frozen_string_literal: true
+
 module Support
   module SchoolInductionTutors
     class Replace
       class << self
         def call(school_id:, email:, full_name:)
-          new(school_id: school_id, email: email, full_name: full_name).call
+          new(school_id:, email:, full_name:).call
         end
       end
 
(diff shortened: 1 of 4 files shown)


app/services/support/school_induction_tutors/replace.rb
```diff
@@ -0,0 +1,59 @@
...
+    class Replace
+      class << self
+        def call(school_id:, email:, full_name:)
+          new(school_id:, email:, full_name:).call
+        end
+      end
+
+      attr_reader :school, :email, :full_name
+
+      def initialize(school_id:, email:, full_name:)
```
leoapost, line 14: I wonder if it would be better to use the URN instead of the school_id, since that's what we usually use when mentioning schools?

app/services/support/school_induction_tutors/update.rb
```diff
@@ -0,0 +1,66 @@
...
+    class Update
+      class << self
+        def call(school_id:, email:, full_name:)
+          new(school_id:, email:, full_name:).call
+        end
+      end
+
+      attr_reader :school, :email, :full_name
+
+      def initialize(school_id:, email:, full_name:)
```
leoapost, line 14: Same here, does it worth using the school URN instead of the school_id?

leoapost (approved):
LGTM. 

I left a few comments about using the school URN instead of the school ID, but it's not a big deal.

app/services/support/school_induction_tutors/replace.rb
Lockyy: Good plan, swapped over

PR #4006, opened 2023-09-15
Switch `f.submit` over to `f.govuk_submit` for consistent HTML

peteryates:
The `govuk_submit` form helper ensures the relevant `data-module` attributes are set and double-click prevention is enabled too.

### DIFF
--- a/app/views/users/sessions/redirect_from_magic_link.html.erb
+++ b/app/views/users/sessions/redirect_from_magic_link.html.erb
@@ -2,11 +2,11 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl">You’re now signed in</h1>
-    <%= form_tag(users_sign_in_with_token_path) do %>
+    <%= form_with(url: users_sign_in_with_token_path) do |f| %>
       <div class="field">
         <%= hidden_field_tag :login_token, @login_token %>
       </div>
-      <%= submit_tag "Continue", class: "govuk-button" %>
+      <%= f.govuk_submit %>
     <% end %>
   </div>
 </div>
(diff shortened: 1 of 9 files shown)


frankieroberto (approved):

app/views/users/sessions/redirect_from_magic_link.html.erb
```diff
@@ -6,7 +6,7 @@
       <div class="field">
         <%= hidden_field_tag :login_token, @login_token %>
       </div>
-      <%= submit_tag "Continue", class: "govuk-button" %>
+      <%= tag.button("Continue", class: "govuk-button", data: { module: "button" }) %>
```
frankieroberto, line 9: Switch this to `f.govuk_submit` by changing `form_tag` to `form_for` (even if there’s no model object)?

peteryates: I was under the impression `form_for` _needed_ an object to bind to

peteryates: Actually maybe `form_with` is the answer here. Let me test it.

frankieroberto: > I was under the impression form_for needed an object to bind to

I think that was maybe true once but changed at some point.

frankieroberto (approved):
Some of the Cypress tests are failing as they’re overly specific. 😔

peteryates:
Ah, I fixed a few but couldn't find a 'run all' button in the client so lazily pushed. Will get it green in the morning.

frankieroberto:
> Ah, I fixed a few but couldn't find a 'run all' button in the client so lazily pushed. Will get it green in the morning.

👍  (longer-term maybe we could consider reducing the number of Cyprus / Chromium tests and rely more on capybara tests?)

peteryates:
Yeah, I've wanted to get rid of them for a long time.

I don't think we really use any custom JavaScript so I'm not sure what they're bringing to the table at the moment.

PR #4009, opened 2023-09-15
Add database populated checks to the healthcheck and smoke test

peteryates:
### Context

The healthcheck is used by the smoke tests to ensure that the environments are in good shape before continuing the continuous deployment process.

This check provides some extra safety by ensuring if we make changes to resource names in terraform that result in them being deleted (#3992, #4008) and recreated, the emptiness of the tables will cause a failure.

- Add a database populated check to the healthcheck
- Alter healthcheck status based on db connectivity
- Add connected/populated checks to the smoke test

### DIFF
--- a/app/controllers/healthcheck_controller.rb
+++ b/app/controllers/healthcheck_controller.rb
@@ -10,6 +10,7 @@ class HealthcheckController < ApplicationController
       environment: Rails.env,
       database: {
         connected: database_connected?,
+        populated: database_populated?,
         migration_version:,
       },
       sidekiq: {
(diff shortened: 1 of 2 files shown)


app/controllers/healthcheck_controller.rb
```diff
@@ -27,6 +28,14 @@ def check
 
 private
 
+  def status
+    if database_connected? && database_populated?
+      :ok
+    else
+      :internal_server_error
```
cwrw, line 35: since we will be returning this, should Sentry ignore this endpoint as well?

peteryates: I'm happy for it to be ignored if it becomes a problem but I think for now just let everything through.

cwrw (approved):

PR #4016, opened 2023-09-15
[CPDLP-2427] Implement token-based rate limiting

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2427

* API rate limiting now uses token instead of IP
* Added tests to make sure rate limiting is working

### DIFF
--- a/spec/requests/api/v3/rate_limiting_spec.rb
+++ b/spec/requests/api/v3/rate_limiting_spec.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "API Rate Limiting", type: :request do
+  let(:cpd_lead_provider) { create(:cpd_lead_provider, :with_lead_provider) }
+  let(:lead_provider) { cpd_lead_provider.lead_provider }
+  let(:cohort) { Cohort.current || create(:cohort, :current) }
+  let(:delivery_partner) { create(:delivery_partner, name: "First Delivery Partner") }
+  let!(:provider_relationship) { create(:provider_relationship, cohort:, delivery_partner:, lead_provider:) }
+
+  let(:token) { LeadProviderApiToken.create_with_random_token!(cpd_lead_provider:) }
+  let(:bearer_token) { "Bearer #{token}" }
+
+  let(:token2) { LeadProviderApiToken.create_with_random_token!(cpd_lead_provider:) }
+  let(:bearer_token2) { "Bearer #{token2}" }
+
+  let(:cache_store) { ActiveSupport::Cache.lookup_store(:file_store, Rails.root.join("tmp/cache/rate_limiting")) }
+
+  before do
+    Rack::Attack.cache.store = cache_store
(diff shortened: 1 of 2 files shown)


spec/requests/api/v3/rate_limiting_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        # Should return error
+        get "/api/v3/delivery-partners/#{delivery_partner.id}"
+        expect(response).to have_http_status(429)
+      end
+    end
+
+    context "Rate limiting is off" do
+      before do
+        Rack::Attack.enabled = false
```
cwrw, line 60: Im not sure we need to test this?

spec/requests/api/v3/rate_limiting_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    Rack::Attack.cache.store.clear
+    Rack::Attack.enabled = false
+  end
+
+  describe "Rate limiting /api/v3/delivery-partners/{id}" do
+    context "When going over limit" do
+      it "returns 429 when going over 5 requests within 5 min" do
+        expect(Rack::Attack.enabled).to be_truthy
+
+        get "/api/v3/delivery-partners/#{delivery_partner.id}"
```
cwrw, line 37: maybe we can loop through the first 5 requests? also maybe we bring it down to 3 instead just to avoid too many?

spec/requests/api/v3/rate_limiting_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      it "allows you to go over 5 requests" do
+        10.times do
+          get "/api/v3/delivery-partners/#{delivery_partner.id}"
+          expect(response).to have_http_status(:success)
+        end
+      end
+    end
+
+    context "Rate limiting with HTTP_AUTHORIZATION" do
```
cwrw, line 71: maybe we can avoid reaching into the internals of the cache and instead add a test for another provider key which isn't rate limited?

cwrw (approved):
looks good, a few spec suggestions inline

spec/requests/api/v3/rate_limiting_spec.rb
mooktakim: I was thinking it would be good to add a negative test.

spec/requests/api/v3/rate_limiting_spec.rb
mooktakim: I wanted to make sure we were using `HTTP_AUTHORIZATION` and not IP or something else.
Although it looks into the internals, as we're specifying the cache store (file store), it shouldn't cause issues. Its the best way I could find to check if we're using the correct rate-limiting key.

spec/requests/api/v3/rate_limiting_spec.rb
mooktakim: Will update to 3 requests

cwrw (approved):

PR #4021, opened 2023-09-18
Add induction dates to admin view of ECT

frankieroberto:


app/presenters/admin/participant_presenter.rb
```diff
@@ -87,6 +87,22 @@ def all_induction_records
     induction_records
   end
 
+  def induction_start_date
+    if participant_profile.induction_start_date
+      participant_profile.induction_start_date.to_formatted_s(:govuk)
+    else
+      "Not yet recorded"
+    end
```
peteryates, line 95: Are we definitely using this field now? I [added some related functionality](https://github.com/DFE-Digital/early-careers-framework/pull/2634) and it hadn't been plumbed in so it was retracted.

peteryates (approved):
Looks great, just worth checking the field is fully in use now as per my comment.

app/presenters/admin/participant_presenter.rb
frankieroberto: Yep, we’re now pulling the data from DQT and displaying it in the schools interface.

tonyheadford (approved):
Nice!

PR #4023, opened 2023-09-18
Fix admin cohort page when school cohort present and partnerships missing

peteryates:
### Context

This is an attempt to fix a bug in the admin console's cohorts tab where:

* a school is FIP
* the school has no partnerships/relationships
* the school does have a `school_cohort`

Previously this circumstance (which sounds invalid to me!) left a gap on the page. Now that gap is filed with a summary list that has a change link:

| Before | After |
|-----|-----|
| Screenshot from 2023-09-18 16-57-05 | Screenshot from 2023-09-18 16-57-18 |

Because we're now displaying FIP details in multiple context I've moved the FIP entries from `#training_programme` up into the `CohortComponent` and just pass the text into the `PartnershipComponent` .

Also added a few extra tests for better coverage of odd situations.

### DIFF
--- a/app/components/admin/schools/cohort_component.html.erb
+++ b/app/components/admin/schools/cohort_component.html.erb
@@ -29,7 +29,7 @@
             row.with_value(text: training_programme)
 
             if allow_change_programme?
-              row.with_action(text: "Change", visually_hidden_text: "induction programme",  href: change_programme_href)
+              row.with_action(text: "Change", visually_hidden_text: "training programme",  href: change_programme_href)
             end
           end
         end
--- a/app/components/admin/schools/cohort_component.html.erb
+++ b/app/components/admin/schools/cohort_component.html.erb
@@ -9,16 +9,31 @@
 
   <% elsif fip? %>
 
-    <% partnership_components.each do |pc| %>
-      <%= pc %>
-    <% end %>
-
-    <% if relationship_components.any? %>
-      <h3 class="govuk-heading-m">Additional relationships</h3>
-
-      <% relationship_components.each do |rc| %>
-        <%= rc %>
+    <% if has_partnerships_or_relationships? %>
+      <% partnership_components.each do |pc| %>
+        <%= pc %>
       <% end %>
+
+      <% if relationships.present? %>
+        <h3 class="govuk-heading-m">Additional relationships</h3>
+
+        <% relationship_components.each do |rc| %>
+          <%= rc %>
+        <% end %>
+      <% end %>
+    <% else %>
+      <%=
+        govuk_summary_list do |summary_list|
+          summary_list.with_row do |row|
+            row.with_key(text: "Training programme")
+            row.with_value(text: training_programme)
+
+            if allow_change_programme?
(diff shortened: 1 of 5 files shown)


app/components/admin/schools/cohort_component.html.erb
```diff
@@ -9,16 +9,31 @@
...
-      <%= pc %>
-    <% end %>
+    <% if has_partnerships_or_relationships? %>
+      <% partnership_components.each do |pc| %>
+        <%= pc %>
+      <% end %>
 
-    <% if relationship_components.any? %>
-      <h3 class="govuk-heading-m">Additional relationships</h3>
+      <% if relationships.present? %>
```
peteryates, line 17: This is cosmetic.

app/components/admin/schools/cohort_component.html.erb
peteryates, line 1: It's easier to review this if you disable whitespace in the diff.

app/components/admin/schools/cohort_component.html.erb
```diff
@@ -9,16 +9,31 @@
 
   <% elsif fip? %>
 
-    <% partnership_components.each do |pc| %>
-      <%= pc %>
-    <% end %>
+    <% if has_partnerships_or_relationships? %>
+      <% partnership_components.each do |pc| %>
+        <%= pc %>
+      <% end %>
```
frankieroberto, line 15: Would it be helpful to always show these Training provider and Delivery partner rows whenever the training programme is the full induction programme, even if the partnerships haven’t been set up?  (They could show "Not set up yet" or similar in the value column?)

app/components/admin/schools/cohort_component.html.erb
```diff
@@ -9,16 +9,31 @@
...
       <% end %>
+    <% else %>
+      <%=
+        govuk_summary_list do |summary_list|
+          summary_list.with_row do |row|
+            row.with_key(text: "Training programme")
+            row.with_value(text: training_programme)
+
+            if allow_change_programme?
+              row.with_action(text: "Change", visually_hidden_text: "induction programme",  href: change_programme_href)
```
frankieroberto, line 32: Visually-hidden text should match the key here, I think?
```suggestion
              row.with_action(text: "Change", visually_hidden_text: "training programme",  href: change_programme_href)
```

peteryates: Well caught, thanks!

app/components/admin/schools/cohort_component.html.erb
peteryates: I'm not sure what the actual logic needs to be here but each partnership component will be its own summary list (card).

I'm happy to tweak it but would rather see this update released first as it fixes a bug.

frankieroberto: I guess I mean that if there's no partnership set up for a cohort (but there should be, as they've picked the full induction programme), then it might be worth showing that explicitly?

But agree that this can get released first!

frankieroberto (approved):

PR #4025, opened 2023-09-19
Add primary key to HABTM join table

ethax-ross:
### Context

We have a `has_and_belongs_to_many` table `cohorts_lead_providers`. By default Rails doesn't add a primary key `id` column to this table (as it as a composite of the foreign keys). The `dfe-analytics` gem, however, requires an `id` column that is a primary key on all tables it sends to
BigQuery.

- Add primary key to HABTM table
- Blocklist active_record_storage_variants

- ### Guidance to review

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -178,6 +178,7 @@
   - induction_start_date
   - induction_completion_date
   :cohorts_lead_providers:
+  - id
   - lead_provider_id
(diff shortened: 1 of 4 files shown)


cwrw (approved):

ethax-ross:
Closing and re-opening in a new PR as the review app is stuck

PR #4031, opened 2023-09-19
Update mentees when deduplicating

ethax-ross:
[Jira-2455](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2455)

### Context

Currently when we deduplicate two mentor profiles we are neglecting to assign the `mentee_profiles` from the duplicate to the primary, so we end up with mentees that have no mentor.

- Update mentees when deduplicating

### DIFF
--- a/lib/participant_profile_deduplicator.rb
+++ b/lib/participant_profile_deduplicator.rb
@@ -105,10 +105,11 @@ private
   def update_mentored_ects!
     return unless [primary_profile, duplicate_profile].all?(&:mentor?)
 
-    duplicate_profile.mentee_profiles.each do |mentee|
-      mentee.update!(mentor_profile_id: primary_profile.id)
-      mentee.induction_records.each { |induction_record| induction_record.update!(mentor_profile_id: primary_profile.id) }
-    end
+    duplicate_profile.mentee_profiles.update_all(mentor_profile_id: primary_profile.id)
+
+    InductionRecord
+      .where(mentor_profile_id: duplicate_profile.id)
+      .update_all(mentor_profile_id: primary_profile.id)
   end
 
   def update_primary_profile_schedule(new_schedule)
--- a/lib/participant_profile_deduplicator.rb
+++ b/lib/participant_profile_deduplicator.rb
@@ -31,6 +31,7 @@ class ParticipantProfileDeduplicator
         transfer_validation_data!
         transfer_participant_eligibility!
         handle_school_change!
+        update_mentored_ects!
         reconcile_remaining_induction_records!
       end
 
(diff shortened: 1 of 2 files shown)


lib/participant_profile_deduplicator.rb
```diff
@@ -101,6 +102,12 @@ def reconcile_schedules!
     update_primary_profile_schedule(profile_with_correct_schedule.latest_induction_record.schedule)
   end
 
+  def update_mentored_ects!
+    return unless [primary_profile, duplicate_profile].all?(&:mentor?)
+
+    duplicate_profile.mentee_profiles.update_all(mentor_profile_id: primary_profile.id)
```
cwrw, line 108: I think we'll also need to update the induction records as well with the new profile? see `Induction::ChangeMentor.call(induction_record:, mentor_profile: new_mentor_profile)` as might help with this

lib/participant_profile_deduplicator.rb
```diff
@@ -101,6 +102,15 @@ def reconcile_schedules!
     update_primary_profile_schedule(profile_with_correct_schedule.latest_induction_record.schedule)
   end
 
+  def update_mentored_ects!
+    return unless [primary_profile, duplicate_profile].all?(&:mentor?)
+
+    duplicate_profile.mentee_profiles.each do |mentee|
+      mentee.update!(mentor_profile_id: primary_profile.id)
+      mentee.induction_records.each { |induction_record| induction_record.update!(mentor_profile_id: primary_profile.id) }
```
cwrw, line 110: bah sorry I realised something I missed again, I think we need to only update the IRs `mentor_profile_id` **IF** the `mentor_profile_id` equals the one we want to delete. As we might be accidentally be removing history here:  
- Someone has a `nil` `mentor_profile_id` in the beginning
- Someone has a different mentor profile and transferred to the current one later.

So maybe safest way is to just find the IRs where the `mentor_profile_id` equal the one changing and updating those only, and ensuring that the profile has the old one as well before updating, sorry again! Hopefully this catches all variations

lib/participant_profile_deduplicator.rb
ltello: Agreed, we need to update any induction record whose `mentor_profile_id` matches the `duplicate_profile.id`.
Actually I would only update the induction records and not the `duplicate_profile.mentee_profiles...` because that is a legacy association not used/updated anymore.

ltello:
```suggestion
    InductionRecord.where(mentor_profile_id: duplicate_profile.id).update_all(mentor_profile_id: primary_profile.id)
```

ltello (changes requested):

lib/participant_profile_deduplicator.rb
cwrw: I think it's safer to update both to avoid any invalid linking. I think @ethax-ross has the IR update below :)

cwrw (approved):
👍

lib/participant_profile_deduplicator.rb
ltello: It's fine. I think what we should do is to delete those legacy mentee associations once and for all.
And yes, sorry I didn't see the update of the induction records below!.

ltello (approved):

PR #4032, opened 2023-09-19
[CST-1981] Remove NTA from 2023 AB options list

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-1981)
Remove NTA from the list of options of National ABs from 2023

Make cohort setup form respect the `AppropriateBody.disable_from_year` attribute when providing the options for selection
Set the disable_from_year to 2023 in the AB importer for NTA.

### Guidance to review
As a SIT when setting up the 2023 cohort, when choosing an AB select "National" and then the only AB option presented should be IStip. NTA should not be an option.

### DIFF
--- a/app/forms/appropriate_body_selection_form.rb
+++ b/app/forms/appropriate_body_selection_form.rb
@@ -46,7 +46,7 @@ class AppropriateBodySelectionForm
 
   def self.body_type_choices_for_year(cohort_start_year)
     if cohort_start_year.present?
-      TYPES.select { |ab_type| ab_type.disable_from_year.nil? || ab_type.disable_from_year > cohort_start_year }
+      TYPES.select { |ab_type| ab_type.disable_from_year.nil? || cohort_start_year < ab_type.disable_from_year }
     else
       TYPES
     end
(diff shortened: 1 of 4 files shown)


frankieroberto:
Hmm, this screen doesn’t make much sense with just one option?!

Screenshot 2023-09-20 at 09 37 46

We could ship it for now, but maybe a follow-up iteration might be to remove that screen altogether, and move the single option to the previous page something like this:

Screenshot 2023-09-20 at 09 40 07

What do you think?

cc @Lloyd-K

Lloyd-K:
Hi, I agree with shipping as is and creating a follow-up task to redesign the flow as Frankie suggests. Thanks

tonyheadford:
> Hmm, this screen doesn’t make much sense with just one option?!
> 
> Screenshot 2023-09-20 at 09 37 46
> We could ship it for now, but maybe a follow-up iteration might be to remove that screen altogether, and move the single option to the previous page something like this:
> 
> Screenshot 2023-09-20 at 09 40 07
> What do you think?
> 
> cc @Lloyd-K

100% agree, thanks

frankieroberto (approved):

ltello (approved):

PR #4034, opened 2023-09-20
Update _no_ects_cohort_details.html.erb

Lloyd-K:
Mentors are now referenced too. This is because induction tutors need to set up training for 23/24 if they expect new mentors to begin training, even if they do not expect any new ECTs.

See https://dfedigital.atlassian.net/browse/CST-2004 for more information.


app/views/schools/dashboard/_no_ects_cohort_details.html.erb
```diff
@@ -1,2 +1,2 @@
-<p class="govuk-body">Your school has told us you do not expect any ECTs to join you in the <%= school_cohort.description %> academic year.</p>
+<p class="govuk-body">Your school has told us you do not expect any ECTs pr mentors to join you in the <%= school_cohort.description %> academic year.</p>
```
frankieroberto, line 1:
```suggestion
<p class="govuk-body">Your school has told us you do not expect any ECTs or mentors to join you in the <%= school_cohort.description %> academic year.</p>
```

frankieroberto (changes requested):

PR #4045, opened 2023-09-22
[CPDLP-2457] Ensure API spec contains all possible schedule-identifier values for ECF and NPQ

leandroalemao:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2457

We’re missing some schedule-identifier values from the API spec.

Ensure API spec contains all possible schedule-identifier values for ECF and NPQ.

### DIFF
--- a/swagger/v3/api_spec.json
+++ b/swagger/v3/api_spec.json
@@ -3187,6 +3187,20 @@
           "schedule_identifier": {
             "description": "The schedule of the ECF participant. For the possible values please refer to the <a href=\"/api-reference/ecf/schedules-and-milestone-dates.html#schedules-and-milestone-dates\" class=\"govuk-link\">ECF schedules and milestone dates guidance</a>.",
             "type": "string",
+            "enum": [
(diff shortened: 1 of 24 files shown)


swagger/v3/api_spec.json
```diff
@@ -5237,19 +5246,9 @@
             ]
           },
           "schedule_identifier": {
-            "description": "The schedule the participant is enrolled in",
+            "description": "The schedule the participant is enrolled in. For the possible values please refer to the <a href=\"/api-reference/npq/schedules-and-milestone-dates.html#schedules-and-milestone-dates\" class=\"govuk-link\">NPQ schedules and milestone dates guidance</a>.",
```
cwrw, line 5249: I think because this is an API spec, its safer to have the full enum list on the API spec even if we refer to the guidance, could we please add the full enum list here and everywhere else? (also ECF)

cwrw (approved):
👍

PR #4047, opened 2023-09-22
Convert training record state to Ruby

ethax-ross:
[Jira-2458](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2458)

### Context

The `DetermineTrainingRecordState` was updated to be an SQL query so that we could use materialized views (for performance). We found that this locked up our production database, so we ended up moving away from using the materialized views.

The remaining SQL is quite hard to follow and contains a lot of business logic. We would like to move back to a Ruby implementation for ease of understanding and potentially look at other ways to improve performance (such as supporting multiple profiles or caching the training states in their own model/table).

Revert implementation to Ruby and update it to be consistent with the latest SQL logic.

Add a new `DetermineTrainingRecordState`/`TrainingRecordState` service that calculates training states over an existing `ParticipantProfile`/`InductionRecord` data set (passed down from the controller) to improve performance.

- Support filtering on appropriate_body
- Convert DetermineTrainingRecordState to Ruby
- Support multiple participant profiles for `AppropriateBodyStatusTag`
- Update to use Ruby pattern matching

### Guidance to review

The only changes to the specs were to migrate from strings to symbols.

This PR migrates the `AppropriateBodyParticipantStatusTag` to use the new `DetermineTrainingRecordState` service. The other status tag view components will be updated in follow-up PRs.

> ⚠️ Before we ship this we will need to clean up 14 ECF profiles in production that have no induction records!
```
ParticipantProfile::ECF.left_joins(:induction_records).where(induction_records: { id: nil })

["90cfd1d8-a98a-436b-9d84-1b8d03aaf350",
 "dfd0356b-bc17-4c58-9075-26680dd98125",
 "fb01bb14-bada-4923-bed1-39bf13b073e6",
 "481bc064-a6ca-461b-b4de-7335dd2ae963",
 "76c2aba0-3ef8-4fac-a7c8-60249588684c",
 "883732d7-a5cf-408c-b6b8-96a8090cef02",
 "3a5000ac-ed88-4dac-9d55-6029415aff9d",
 "af3404c4-10e1-4aa5-8cb4-4cacbba8d319",
 "0f379295-c028-4fcc-9c6c-4067eb8658a4",
 "92233b7a-245e-4982-8ed8-dee9e9d7e9a7",
 "5f2765b1-5cbd-4135-aa03-b811594f8a60",
 "a9c49949-a286-456c-934d-4358407146ad",
 "f664c509-f307-4bba-bcae-9f4a2fc92fdb",
 "6dcf5335-6fc0-466c-b0a6-ef1fc0eda5f2"]
```


peteryates:
This is fantastic, far easier to read now.

peteryates:
Having had a closer look through this looks like a great first step. Like I said above, it's much easier to read and work out the logic behind each status.

I think this might be a good place for [pattern matching](https://docs.ruby-lang.org/en/master/syntax/pattern_matching_rdoc.html), I created [a little test branch with an example](https://github.com/DFE-Digital/early-careers-framework/commit/1b08e3f0e40f4bf375a13e52099da6c2fe043c83). That would allow some of the more complex boolean methods be given a single responsibility and then they're more composable.

peteryates (approved):
100% thumbs up from me. I think it's definitely worth having someone more closely involved give it a once over too though :)

app/controllers/appropriate_bodies/participants_controller.rb
```diff
@@ -5,9 +5,11 @@
 module AppropriateBodies
   class ParticipantsController < BaseController
     def index
-      collection = InductionRecordsQuery.new(appropriate_body:).induction_records
+      induction_records = InductionRecordsQuery.new(appropriate_body:).induction_records
+      participant_profiles = induction_records.map(&:participant_profile)
```
leandroalemao, line 9: @ethax-ross Seems like we have a N+1 query in this line..
We are fetching all participant profiles for all induction records before sending it to the filters and to pagy to only display 50 records per page in the end.
Can we somehow fetch participant profile and other data for only those 50 records in the first place?

ethax-ross: hmm yeah there is an N+1 here, I'll see if I can address that. We can't filter/apply pagination first unfortunately because the training states are used _in_ the filters 😬

cwrw (approved):
⭐ excellent work, thank you for being very very patient with this and taking on feedback which was spaced out! substantial performance improvements and fixing a few bugs with the filters as well.

A few things we discussed for next time:
- to look into the filter which loops through induction records and improvements (can be done as part of DP work)
- possibly looking to have includes on only paginated results (also later if needed)

mattonfoot (approved):
Thanks for this, looks good to me

tonyheadford (approved):
Nice, great improvement!

PR #4064, opened 2023-09-27
[CST-2025] Show CIP materials on CIP induction records

tonyheadford:
### Context

On the admin UI when looking at a participant's induction record the LP/DP is always present and the CIP materials never show even if the induction record is a CIP one.  I've added a condition to only show the LP/DP details when FIP and show CIP materials when CIP instead.

Show CIP materials on CIP induction records and FIP LP/DP on FIP induction records

### Example CIP IR
image
### Example FIP IR
image

### DIFF
--- a/app/views/admin/participants/induction_records/show.html.erb
+++ b/app/views/admin/participants/induction_records/show.html.erb
@@ -75,8 +75,8 @@
           end
         elsif induction_record.enrolled_in_cip?
           sl.with_row do |row|
-            row.with_key(text: "Core materials provider")
-            row.with_value(text: induction_record.core_induction_programme_name || "No materials provider")
+            row.with_key(text: "Materials supplier")
+            row.with_value(text: induction_record.core_induction_programme_name || "No materials supplier")
           end
         end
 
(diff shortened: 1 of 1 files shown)


app/views/admin/participants/induction_records/show.html.erb
```diff
@@ -56,20 +56,27 @@
...
+              row.with_key(text: "Delivery partner record state")
+              row.with_value(text: render(StatusTags::DeliveryPartnerParticipantStatusTag.new(participant_profile: induction_record.participant_profile, induction_record: induction_record)))
+            end
+          end
+        elsif induction_record.enrolled_in_cip?
           sl.with_row do |row|
-            row.with_key(text: "Delivery partner record state")
-            row.with_value(text: render(StatusTags::DeliveryPartnerParticipantStatusTag.new(participant_profile: induction_record.participant_profile, induction_record: induction_record)))
+            row.with_key(text: "Core materials provider")
+            row.with_value(text: induction_record.core_induction_programme_name || "No materials provider")
```
frankieroberto, line 79: We use "Materials supplier" on the schools interface, so I think it'd be good to be consistent with that.

```suggestion
            row.with_key(text: "Materials supplier")
            row.with_value(text: induction_record.core_induction_programme_name || "No materials supplier")
```

frankieroberto (changes requested):

PR #4070, opened 2023-09-29
Add Redis instance for Rails cache

ethax-ross:
[Jira-2434](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2434)

### Context

We want to switch to using Redis as our Rails cache store. Currently hosted environments use an in-memory cache store, which is causing issues with RackAttack as each instance will be throttling independently (so not applying the rates as expected).

- Add Redis cache to AKS
- Rename existing redis env var to REDIS_QUEUE
- Use Redis as Rails cache

### Guidance to review

The Redis instance for caching is configured as Standard/C/1 (inline with the Apply service). Our Redis instance for Sidekiq is high availability, but this shouldn't be necessary for our cache (session data is stored in Postgres).

I was going to rename the current `redis` resource to `redis-queue`, but that would result in it being destroyed and re-created in production and potentially losing jobs so I've left it as-is for now.

`rack-attack` is configured by default to use the Rails cache:

```
Rack::Attack.cache.store = Rails.cache 
```

Env vars checked in review app:

```
irb(main):005:0> ENV["REDIS_QUEUE_URL"]
=> "redis://cpd-ecf-review-4070-redis-queue:6379/0"
irb(main):006:0> ENV["REDIS_CACHE_URL"]
=> "redis://cpd-ecf-review-4070-redis-cache:6379/0"
```
I checked the Sidekiq -> Redis connection with `Sidekiq.redis(&:info)` and `Sidekiq.redis{ |conn| conn }._client.options[:url]`:

```
irb(main):014:0> Sidekiq.redis{ |conn| conn }._client.options[:url]
=> "redis://cpd-ecf-review-4070-redis:6379/0"
```

The Rails -> Redis connection can be checked with `Rails.cache`:

```
 #<ActiveSupport::Cache::RedisCacheStore options={:namespace=>nil, :compress=>true, :compress_threshold=>1024, :expires_in=>nil, :race_condition_ttl=>nil} redis={:url=>"redis://cpd-ecf-review-4070-redis-cache:6379/0", :pool_size=>5}>
```

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -50,7 +50,7 @@ Rails.application.configure do
                            read_timeout: 0.2, # Defaults to 1 second
                            write_timeout: 0.2, # Defaults to 1 second
                            reconnect_attempts: 1, # Defaults to 0
-                           error_handler: lambda { |_method:, _returning:, exception:|
+                           error_handler: lambda { |_method, _returning, exception|
                              Sentry.capture_exception(exception)
                            },
                        }
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -50,9 +50,6 @@ Rails.application.configure do
                            read_timeout: 0.2, # Defaults to 1 second
                            write_timeout: 0.2, # Defaults to 1 second
                            reconnect_attempts: 1, # Defaults to 0
-                           error_handler: lambda { |_method, _returning, exception|
-                             Sentry.capture_exception(exception)
-                           },
                        }
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -43,7 +43,7 @@ Rails.application.configure do
   config.session_store :cookie_store, key: "_early_career_framework_session", secure: true, expire_after: 2.weeks
 
   # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  config.cache_store = :redis_cache_store, { url: ENV["REDIS_CACHE_URL"], pool_size: ENV.fetch("RAILS_MAX_THREADS", 5) }
 
   # Use a real queuing backend for Active Job (and separate queues per environment).
   config.active_job.queue_adapter = :sidekiq
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -42,8 +42,18 @@ Rails.application.configure do
   config.force_ssl = true
   config.session_store :cookie_store, key: "_early_career_framework_session", secure: true, expire_after: 2.weeks
 
-  # Use a different cache store in production.
-  config.cache_store = :redis_cache_store, { url: ENV["REDIS_CACHE_URL"], pool_size: ENV.fetch("RAILS_MAX_THREADS", 5) }
+  config.cache_store = :redis_cache_store,
+                       {
+                         url: ENV["REDIS_CACHE_URL"],
+                           pool_size: ENV.fetch("RAILS_MAX_THREADS", 5),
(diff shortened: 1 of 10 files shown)


mooktakim (approved):
Looks good 👍

leandroalemao (approved):
LGTM 👍 
Just as a suggestion.. maybe we can tweak the default settings a bit and/or add an error handler?

https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore

```
ache_servers = %w(redis://cache-01:6379/0 redis://cache-02:6379/0)
config.cache_store = :redis_cache_store, { url: cache_servers,

  connect_timeout:    30,  # Defaults to 20 seconds
  read_timeout:       0.2, # Defaults to 1 second
  write_timeout:      0.2, # Defaults to 1 second
  reconnect_attempts: 1,   # Defaults to 0

  error_handler: -> (method:, returning:, exception:) {
    # Report errors to Sentry as warnings
    Raven.capture_exception exception, level: 'warning',
      tags: { method: method, returning: returning }
  }
}
```

terraform/aks/databases.tf
```diff
@@ -1,3 +1,23 @@
...
+  namespace                 = var.namespace
+  environment               = local.environment
+  azure_resource_prefix     = var.azure_resource_prefix
+  service_name              = local.service_name
+  service_short             = var.service_short
+  config_short              = var.config_short
+  azure_capacity            = var.redis_cache_capacity
+  azure_family              = var.redis_cache_family
+  azure_sku_name            = var.redis_cache_sku_name
+  name                      = "cache"
```
cwrw, line 13: could you add:
```
  patch_schedule {
    day_of_week    = "Sunday"
    start_hour_utc = 01
  }
 ```
 Just an extra setting to ensure if maintenance was to happen it happens outside hours?

File: config/environments/production.rb [hunk 2]
```diff
@@ -43,7 +43,7 @@
   config.session_store :cookie_store, key: "_early_career_framework_session", secure: true, expire_after: 2.weeks
 
   # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  config.cache_store = :redis_cache_store, { url: ENV["REDIS_CACHE_URL"], pool_size: ENV.fetch("RAILS_MAX_THREADS", 5) }
```
cwrw, line 46:
```suggestion
  config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_CACHE_URL"), pool_size: ENV.fetch("RAILS_MAX_THREADS", 5) }
```
just to ensure its not missing :)

cwrw (commented):
looking good, some minor suggestions.

After all 🍏 let's pair tomorrow on freezing/blocking the pipeline and running terraform plan/apply on each environment to be safe before merging, thanks!

config/environments/production.rb
ethax-ross: The issue with doing this is it causes the `assets:precompile` to fail (as the `REDIS_CACHE_URL` isn't set during the docker build, which makes sense). It feels a bit wrong to me to have something like `REDIS_CACHE_URL=fake bundle exec assets:precompile` to work around it 🤔 if the Redis cache is inaccessible Rails will treat it as a cache miss; I've updated with Leandro's suggestion of sending the Redis errors to Sentry as well so we should be notified there if its an issue

cwrw: ok sounds fair 👍

cwrw (approved):
nicely done! let me know when you get a bit of time so we can run terraform on the envs before merging

PR #4073, opened 2023-09-29
[CPDLP-2445] Fix bugs around DP/LP user access management

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2445

* Updated lead provider user creation to allow existing users
* Updated choose role page to include lead provider

### DIFF
--- a/app/forms/supplier_user_form.rb
+++ b/app/forms/supplier_user_form.rb
@@ -37,6 +37,11 @@ class SupplierUserForm
 private
 
   def email_not_taken
-    errors.add(:email, :unique, message: I18n.t("errors.supplier_email.taken")) if Identity.find_user_by(email:)
+    return if supplier.blank?
+    return unless (user = Identity.find_user_by(email:))
+
+    if LeadProviderProfile.where(user:, lead_provider: chosen_supplier).exists?
+      errors.add(:email, :unique, message: I18n.t("errors.supplier_email.taken"))
+    end
   end
 end
--- a/app/forms/choose_role_form.rb
+++ b/app/forms/choose_role_form.rb
@@ -7,6 +7,7 @@ class ChooseRoleForm
   USER_ROLES = {
     "appropriate_body" => "Appropriate body",
     "delivery_partner" => "Delivery partner",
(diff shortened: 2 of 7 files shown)


leandroalemao (commented):
Hi mook,
Shouldn't we remove this validation for DPs as well?

`2 A support user can add an existing DP user (i.e. same email) as an LP user w/ same details`

https://github.com/DFE-Digital/early-careers-framework/blob/main/app/forms/delivery_partners/create_user_form.rb#L13

Not in the scope of this card but we have the same validation for ABs as well:

https://github.com/DFE-Digital/early-careers-framework/blob/main/app/forms/appropriate_bodies/create_user_form.rb#L13

app/forms/supplier_user_form.rb
```diff
@@ -33,10 +32,4 @@ def save!
       ),
     )
   end
-
-private
-
-  def email_not_taken
```
cwrw, line 39: maybe instead of removing this all together we can amend the validation to be similar to delivery partners which looks like this:
```
def email_not_taken
        return if delivery_partner_id.blank?
        return unless (user = Identity.find_user_by(email:))

        if DeliveryPartnerProfile.where(user:, delivery_partner:).exists?
          errors.add(:email, :unique, message: I18n.t("errors.email.taken"))
        end
      end
```
So raise this validation if a lead provider profile already exists for the user?

cwrw (commented):
looks good and works 🎉 just a small suggestion re the validation

app/forms/supplier_user_form.rb
mooktakim: @cwrw @leandroalemao 
Updated the validation to match DP/AB forms 👍

docs/source/release-notes.html.md
```diff
@@ -7,6 +7,14 @@ weight: 8
 
 If you have any questions or comments about these notes, please contact DfE via Slack or email.
 
+## 4 October 2023
```
cwrw, line 10: wrong pr 😅

cwrw (approved):
just the release note comment otherwise 👍

PR #4077, opened 2023-10-02
[CPDLP-2032] Add csv export for eligible ECF participant declarations which have not moved from submitted state

leandroalemao:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2032

Investigate historical ECF declarations which have not moved from the submitted state (post duplicates)

Add csv export for eligible ECF participant declarations which have not moved from submitted state.

### DIFF
--- a/lib/tasks/exports.rake
+++ b/lib/tasks/exports.rake
@@ -75,4 +75,81 @@ namespace :exports do
       # scp -P 2222 -o StrictHostKeyChecking=no -o User=cf:${GUID}/0 ssh.london.cloud.service.gov.uk:/app/payable_declarations.csv .
     end
   end
+
+  desc "ECF declarations which have not moved from the submitted state (post duplicates) - [CPDLP-2032]"
+  task ecf_declarations_not_moved_from_submitted_extract: :environment do
+    headers = [
+      "Participant ID",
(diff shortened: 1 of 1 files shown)


lib/tasks/exports.rake
```diff
@@ -75,4 +75,81 @@ namespace :exports do
...
+    CSV.open(file_path, "w", headers: true) do |csv|
+      csv << headers
+      puts ">>> Initializing Report"
+
+      ParticipantDeclaration::ECF.submitted.find_in_batches do |pds|
+        pds.map do |pd|
+          pd.participant_profile
+          cpd_lead_provider = pd.cpd_lead_provider
+
+          relevant_induction_record = Induction::FindBy.call(
```
cwrw, line 114: I think might not need all this data, I think we just need the declaration id, and participant id as well as the numbers to check a few manually?

cwrw (commented):
left a few comments, is it possible that we get numbers first as an initial take before prs are merged just to get a sense of what's in there?

leandroalemao:
Closing this as we don’t actually have any participant declaration in this scenario.. if we need to re-check in the future we can get/run the same export from the PR.

PR #4078, opened 2023-10-02
[CST-2028] Adopt new way to process induction completion dates

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-2028)
Our original "one-time" backfill job ended up being run every night to check the DQT for induction completion dates.
The `touch` method used to move records to the end of the queue resulted in phantom updates being exposed to providers.

This PR adds a table to hold list of candidates for checking that can be processed and cleared down without causing any side-effects to other parts of the service.

A future piece of work is planned that will review the candidate selection algorithm and perhaps find a more intelligent and cohort-proof method. At the moment it is the same as the previous method that includes all ECTs registered prior to the 2023 academic year that have started an induction.

Add a new `completion_candidates` table and model
Add a new service to check for a completion date and update participant
Add a new service to build candidate list
Add new job to run candidate list builder

### DIFF
--- a/app/services/participants/build_completion_candidate_list.rb
+++ b/app/services/participants/build_completion_candidate_list.rb
@@ -33,10 +33,14 @@ module Participants
           AND c.start_year IN (2020,2021,2022)
           AND pp.induction_start_date IS NOT NULL
           AND pp.induction_completion_date IS NULL
-          AND pp.created_at < '2023-9-1'
+          AND pp.created_at < '#{registration_start_date}'
           AND tp.trn IS NOT NULL;
         SQL
       )
+    end
+
+    def registration_start_date
+      Cohort.find_by(start_year: 2023).registration_start_date.to_date.to_s
     end
   end
 end
--- a/app/services/participants/build_completion_candidate_list.rb
+++ b/app/services/participants/build_completion_candidate_list.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+# This service builds a list of candidates to check for an induction completion date
+# The list will be processed by the SetParticipantCompletionDateJob
+# This could be improved later to more intelligently select which records to include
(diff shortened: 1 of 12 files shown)


app/services/participants/build_completion_candidate_list.rb
```diff
@@ -0,0 +1,46 @@
...
+    def clean_candidate_list
+      CompletionCandidate.delete_all
+    end
+
+    def build_candidate_list
+      ActiveRecord::Base.connection.execute(
+        <<~SQL,
+          INSERT INTO completion_candidates (participant_profile_id)
+          SELECT pp.id
+          FROM participant_profiles pp
```
cwrw, line 28: Do we need to add the with an eligible status here as well as teacher profiles requiring trns as the old query above did?

cwrw (commented):
looks good, just the one comment.
From our conversation in person but for posterity:
- We will monitor the candidate list to see if it completes in the 8 hours
- Worth monitoring the load on the database incase as well

app/services/participants/build_completion_candidate_list.rb
tonyheadford: I don't think it matters, I added the eligible status before because I was trying to only do what was necessary but I don't think it makes any difference, if they have completed we should record that I think?  I will check with Nicola though, thanks.

tonyheadford: Confirmed that PO/BA would like to ignore eligibility for this

tonyheadford: I am excluding teacher profiles without TRNs, it's hidden under the SQL Injection warning

cwrw: ah gotcha, 👍

cwrw (approved):

PR #4085, opened 2023-10-03
Rate limit on remote_ip instead of ip

ethax-ross:
[Trello-2433](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2433)

### Context

Our rate limiting by IP address is not currently working as expected as the `ip` method we are using in Rack is returning our load balancer IP address and not that of the end client.

- Rate limit on `remote_ip` instead of `ip`

The IP address returned by Rails `ActionDispatch#remote_ip` is not the same as Rack's `ip` method. The `ip` method will return the client IP address, even if its a proxy. The `remote_ip` method will leverage the `RemoteIp` middleware, which is smarter and selects the last IP in the chain (via `X-Forwarded-For` headers).

In Azure/AKS we should get an `X-Real-IP` header populated; this is going to be the easiest/most reliable way of obtaining the remote's IP address, falling back to `ActionDispatch#remote_ip` if it is not set.

Update `RackAttack` throttling to use `ActionDispatch#remote_ip` via small monkey patch on `Rack::Request`/

Improve the testing around the rate limiting and centralise it in a single spec file.

### Guidance to review

It's hard to test this without it being in production as our other environments don't have load balancers. You can test it on the review app using a tool like `wrk`:

```
wrk -t5 -c5 -d5s https://cpd-ecf-review-4085-web.test.teacherservices.cloud
Running 5s test @ https://cpd-ecf-review-4085-web.test.teacherservices.cloud
  5 threads and 5 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    59.45ms   38.07ms 443.11ms   94.07%
    Req/Sec    18.04      8.59    40.00     74.04%
  438 requests in 5.03s, 5.12MB read
  Non-2xx or 3xx responses: 131
Requests/sec:     87.10
Transfer/sec:      1.02MB
```

^ If you run multiple threads/connections its possible to go slightly over the limits, presumably due to race conditions. If you run the test using a single thread/connection it always returns 429 after the limit has been exceeded.

The `x-real-ip` header as been verified as populated in production in a separate PR #4093.


config/environments/sandbox.rb
```diff
@@ -156,4 +156,6 @@
   # config.active_record.database_resolver_context = ActiveRecord::Middleware::DatabaseSelector::Resolver::Session
 
   config.ssl_options = { redirect: { exclude: ->(request) { request.path.include?("/check") } } }
+
+  config.middleware.insert_before ActionDispatch::RemoteIp, FixAzureXForwardedForMiddleware
```
cwrw, line 160: ✂️

cwrw (approved):

PR #4103, opened 2023-10-04
[CPDLP-2402] Add the schedule identifier as a parameter into the api accept NPQ application endpoints

leandroalemao:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2402

Providers want to pick which schedule when accepting an NPQ application, so they can have participants on the right schedule.

Add the schedule identifier as a parameter into the api accept NPQ application endpoints.


leandroalemao:
Closing this for now as it was part of a spike done

PR #4107, opened 2023-10-05
Skip DfE analytics events for /check endpoint

ethax-ross:
We call this endpoint frequently and it just creates noise in dfe-analytics/other monitoring platforms.


cwrw (approved):
🧠 !

PR #4111, opened 2023-10-06
6 october release-note.html.md

mark-avery-dfe:
Release note to say we've added a new field, induction_end_date, to the participant profile ECF endpoint.


leandroalemao (approved):
Lgtm 👍 tks mark

PR #4131, opened 2023-10-10
[CPDLP-2467] UCL Deduping:Track deduping participant users

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2467

* New model created `ParticipantIdChange`
* Updated `Identity::Transfer` to create participant id change
* Updated `Api::V3::ECF::ParticipantSerializer and `Api::V3::NPQParticipantSerializer` to include `participant_id_changes`


db/migrate/20231010153003_create_participant_id_changes.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateParticipantIdChanges < ActiveRecord::Migration[7.0]
+  def change
+    create_table :participant_id_changes do |t|
+      t.references :user, null: false, foreign_key: true
+
+      t.references :from_participant, null: false, type: :uuid
```
cwrw, line 8: maybe we can name those from user id/to user id 🤔 ?

db/migrate/20231010153003_create_participant_id_changes.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateParticipantIdChanges < ActiveRecord::Migration[7.0]
+  def change
+    create_table :participant_id_changes do |t|
+      t.references :user, null: false, foreign_key: true
```
cwrw, line 6: if someone was to delete this user, what would happen 🤔

cwrw (commented):
looking good just a suggestion/question

db/migrate/20231010153003_create_participant_id_changes.rb
mooktakim: @cwrw 
this should be fine as it will move to a new user, it should never link to a user that doesn't exist.

db/migrate/20231010153003_create_participant_id_changes.rb
mooktakim: @cwrw 
I went back and forth with this lol
I didn't want the API to be "participant_id_changes" and the table to be "user_id_changes", thought might be good to stay consistent.

cwrw (approved):

PR #4139, opened 2023-10-11
Reinstate DFE analytics to API controller

cwrw:
### Context

Reinstate API requests as `web_request` in DfE Analytics
Adding this the first time https://github.com/DFE-Digital/early-careers-framework/pull/4080 caused issues in the NPQ reg application when querying our API due to the following https://dfe-teacher-services.sentry.io/issues/4519374530/?alert_rule_id=7310304&alert_type=issue&notification_uuid=d4ae4668-3d2a-4566-9ac2-0fe446909ca7&project=5748989&referrer=slack

Where DfE analytics expects a record as the current user, however the owner in NPQ reg APIs is a string.
We now override this check in the config to only check the user IF the user is a record.
- Ticket: n/a

- Add dfe analytics to API controller to start tracking API requests in BQ events
- Add a custom user method in the config to not return anything if the user is not a record
- Add tests for both lead provider and NPQ API

### Guidance to review
We will test this with all APIs before merging

### DIFF
--- a/config/initializers/dfe_analytics.rb
+++ b/config/initializers/dfe_analytics.rb
@@ -53,5 +53,5 @@ DfE::Analytics.configure do |config|
   # return the identifier for the user. This is useful for systems with
   # users that don't use the id field.
   #
-  # config.user_identifier = proc { |user| user&.id }
+  config.user_identifier = proc { |user| user&.id if user.respond_to?(:id) }
 end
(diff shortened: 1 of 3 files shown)


config/initializers/dfe_analytics.rb
```diff
@@ -53,5 +53,5 @@
   # return the identifier for the user. This is useful for systems with
   # users that don't use the id field.
   #
-  # config.user_identifier = proc { |user| user&.id }
+  config.user_identifier = proc { |user| user&.id if user.respond_to?(:id) }
```
peteryates, line 56: 👍🏽

peteryates (approved):

ethax-ross (approved):
Looks good 👍

PR #4152, opened 2023-10-13
Upgrade to dfe-analytics 1.11.1

peteryates:
This release fixes a problem we encountered when using DfE-analytics with [single-table inheritance](https://guides.rubyonrails.org/association_basics.html#single-table-inheritance-sti).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -113,7 +113,7 @@ gem "rouge"
 
 gem "auto_strip_attributes", "~> 2.6"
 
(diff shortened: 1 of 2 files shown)


terraform/aks/.terraform.lock.hcl
```diff
@@ -0,0 +1,82 @@
+# This file is maintained automatically by "terraform init".
```
cwrw, line 1: ✂️

peteryates: Oof. Thanks. I'll add it to the gitignore too 🤦‍♂️

PR #4153, opened 2023-10-13
[CPDLP-2497] Add 'resend' service which allows users resubmit failed NPQ outcomes

leandroalemao:
### Context

- Ticket: [CPDLP-2497](https://dfedigital.atlassian.net/browse/CPDLP-2497)
Finance users want to resubmit NPQ outcomes that failed the initial attempt. So that participants can get their certification.

Add 'resend' service which allows users resubmit failed NPQ outcomes.


app/models/participant_outcome/npq.rb
```diff
@@ -61,6 +63,23 @@ def passed
     def not_passed
       where.not(state: :passed)
     end
+
+    def latest_per_declaration
```
cwrw, line 67: Is the intention of this method to check the outcome in the table is the latest? I wonder if we can just just `declaration.outcomes.latest` is equal to the current outcome?

db/schema.rb
```diff
@@ -119,7 +119,7 @@
     t.index ["start_year"], name: "index_cohorts_on_start_year", unique: true
   end
 
-  create_table "cohorts_lead_providers", force: :cascade do |t|
+  create_table "cohorts_lead_providers", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
cwrw, line 122: 🤔 I guess this was not committed before

spec/components/finance/npq/participant_outcomes/table_spec.rb
```diff
@@ -94,6 +99,7 @@
         is_expected.to have_table_text(participant_outcome.created_at.to_date.to_fs(:govuk), col: 3)
         is_expected.to have_table_text(tra_datetime.to_date.to_fs(:govuk), col: 4)
         is_expected.to have_table_text("NO. CONTACT THE DIGITAL SERVICE TEAM", col: 5)
+        is_expected.to have_table_text("Resend", col: 6)
```
cwrw, line 102: could we add a test here for resending and the new pending state?

cwrw (commented):
looking good 👍 , just some minor comments

cwrw (approved):
🙌

PR #4159, opened 2023-10-16
[CPDLP-2469] UCL Deduping: Add service class to create csvs for deduped participants

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2469

* New service class `ParticipantIdChangesCsv` to return all `participant_id_changes` in the last 30 days

### DIFF
--- a/app/services/participant_id_changes_csv.rb
+++ b/app/services/participant_id_changes_csv.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require "csv"
+
+class ParticipantIdChangesCsv < BaseService
+  include Api::Concerns::FetchLatestInductionRecords
+
+  def initialize(cpd_lead_provider:, from_date: 30.days.ago)
+    @cpd_lead_provider = cpd_lead_provider
+    @from_date = from_date.beginning_of_day
+  end
+
+  def call
+    CSV.generate do |csv|
+      csv << %w[
+        participant_id
+        from_participant_id
+        to_participant_id
+        changed_at
+      ]
+
+      ParticipantIdChange.where(created_at: from_date..).order(created_at: :asc).each do |c|
+        next unless cpd_lead_provider?(c.user)
+
+        csv << [
+          c.user_id,
+          c.from_participant_id,
+          c.to_participant_id,
+          c.created_at.rfc3339,
+        ]
+      end
+    end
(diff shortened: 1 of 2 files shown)


ethax-ross (approved):
Looks good 👍 only a minor suggestion

app/services/participant_id_changes_csv.rb
```diff
@@ -0,0 +1,53 @@
...
+  attr_reader :cpd_lead_provider, :from_date
+
+  delegate :lead_provider, :npq_lead_provider,
+           to: :cpd_lead_provider
+
+  def cpd_lead_provider?(user)
+    user.participant_profiles.each do |pp|
+      if pp.npq? && pp.npq_application.npq_lead_provider == npq_lead_provider
+        return true
+      elsif pp.ecf? && pp.lead_provider == lead_provider
```
cwrw, line 46: could we use induction records here instead?

cwrw, line 46: I think this method will only get the latest provider for a profile, we would like to surface any profile that ever belonged to the lead provider even after a transfer. Could we instead use: `latest_induction_record_for` method?

spec/services/participant_id_changes_csv_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    expect(parsed_result[2][1]).to eql(participant_id_change1.from_participant_id)
+    expect(parsed_result[2][2]).to eql(participant_id_change1.to_participant_id)
+    expect(parsed_result[2][3]).to eql(participant_id_change1.created_at.rfc3339)
+  end
+
+  it "should not return participant_id_changes older than 30 days" do
+    result = described_class.call(cpd_lead_provider: cpd_lead_provider2)
+    parsed_result = CSV.parse(result, headers: true).to_a
+
+    expect(parsed_result.size).to eql(1)
```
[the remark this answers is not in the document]
mooktakim: Ah yes, since i'm converting to array the headers params doesn't matter. Will include tests for the headers too.

app/services/participant_id_changes_csv.rb
mooktakim: Updated to use `latest_induction_record_for`

PR #4165, opened 2023-10-17
Fix flaky spec

tonyheadford:
### Context

Spec seems flaky, I think the order of the returned records is not explicitly defined so this changes the spec to use `match_array` instead of expecting the individual records in a particular order.
However if this not the desired behaviour and a particular order is expected then there is a bug somewhere.


cwrw (approved):
checked order tested in other service so all good

PR #4167, opened 2023-10-17
[CST-2026] Display the materials supplier in Admin and Schools pages

leoapost:
### Context

- Ticket: CST-2026

We want to display the materials supplier in the following pages for CIP participants:
- Admin Training page
- Admin Statuses page
- Schools participant page

- Update the views to display the Materials supplier or the LP/DP info based on their CIP/FIP enrollment

### Guidance to review
Check all three pages are displaying:
- the Materials supplier when participants are on CIP
- the LP/DP info when participants are on FIP

### DIFF
--- a/app/components/admin/schools/cohort_component.html.erb
+++ b/app/components/admin/schools/cohort_component.html.erb
@@ -51,9 +51,9 @@
 
         if cip?
           summary_list.with_row do |row|
-            row.with_key(text: "Materials")
+            row.with_key(text: "Materials supplier")
             row.with_value(text: materials)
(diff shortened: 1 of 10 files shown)


frankieroberto (approved):
Checked in review app and looks good to me.

I noticed that the Materials supplier is also list on the admin view for school cohorts, except that it says "Materials" rather than "Materials supplier" – would you mind making that consistent by adding " supplier" to the label?

leoapost:
> Checked in review app and looks good to me.
> 
> I noticed that the Materials supplier is also list on the admin view for school cohorts, except that it says "Materials" rather than "Materials supplier" – would you mind making that consistent by adding " supplier" to the label?

Done

PR #4169, opened 2023-10-18
Accurate cohort seeds

mattonfoot:
### Context

Seeded environments break when adding NPQ Registration seeds. The seeds expect there to be schedules available for `Cohort.next` but at certain times of the year the "schedules.csv" will not have details of schedules for the cohort to add. This will cause the seed scripts to fail. 

The last solution to solve this issue was to push the `academic-year-start-date` into the future but, this has had the knock on effect of meaning that seed environments fail to make 2023 the current cohort like the production system does and means no seed schools have chosen FIP or partnered with anyone for the 2023 academic year.

1. Use the "cohort.csv" to easily define the cohorts that are seeded - giving us the option to mimic production in a simple way
2. Ensure that Cohort seed defaults create a more realistic timeline
3. Ensure that the previous logic to ensure a `Cohort.next` runs with the default seed settings
4. Ensure that if a `Cohort.next` has been created that some default NPQ schedules and milestones are also created so that the NPQ registration seeds have something to reference

### DIFF
--- a/db/data/cohorts/cohorts.csv
+++ b/db/data/cohorts/cohorts.csv
@@ -1,5 +1,5 @@
-start-year,registration-start-date,academic-year-start-date,npq-registration-start-date
-2020,2020/05/10,2020/09/01,
-2021,2021/05/10,2021/09/01,
-2022,2022/05/10,2022/09/01,
-2023,2023/05/10,2023/09/01,2023/04/03
+start-year,registration-start-date,academic-year-start-date,npq-registration-start-date,automatic-assignment-period-end-date
+2020,2020/05/10,2020/09/01,,2021/03/31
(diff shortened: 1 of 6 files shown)


cwrw (commented):
Thanks for adding real data to the CSVs to make more accurate and inline with production. I was wondering since now we know the issue for 2023 not being present was because of having to push the date to 1/11 which Im hoping to tackle and fix by next week, if we need to fix the cohort factory dates, as it can cause a knock on effect with inconsistent seeds/data in specs where the other seeds are set to 1/11. 
Can we can keep the same dates for now until fixed next week?

Everything else past the failure and minor comment looks good

cwrw:
closing as commits have been cherry picked and merged in https://github.com/DFE-Digital/early-careers-framework/pull/4257

PR #4173, opened 2023-10-18
CST-2012: allow admin users to edit the training status of CIP participants

ltello:
### Context
When a FIP participant is no longer training, their provider sets training_status = withdrawn.

CIP participants, however, have no lead provider, so we have to find another way of indicating they’re no longer training.

We also use induction_status = withdrawn. However, at the moment this needs to be actioned by a dev, which is time-consuming.

We therefore want to add a facility for first line agents to change induction_status = withdrawn for CIPsters ONLY. Note that this will not affect their access to funding.

[Ticket](https://dfedigital.atlassian.net/browse/CST-2012)

276339403-cd90c39a-06b8-46f2-9fa1-782ef69d35b0

Screenshot 2023-10-26 at 12 12 07

### DIFF
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -106,8 +106,8 @@ module AdminHelper
 
   def training_statuses
     [
-      OpenStruct.new(id: "deferred", value: "deferred"),
-      OpenStruct.new(id: "withdrawn", value: "withdrawn"),
+      OpenStruct.new(id: "deferred", value: "Deferred"),
+      OpenStruct.new(id: "withdrawn", value: "Withdrawn"),
     ]
   end
 end
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -103,4 +103,11 @@ module AdminHelper
       "unknown"
     end
   end
+
+  def training_statuses
+    [
+      OpenStruct.new(id: "deferred", value: "deferred"),
+      OpenStruct.new(id: "withdrawn", value: "withdrawn"),
+    ]
+  end
 end
--- a/app/views/admin/participants/induction_records/edit_training_status.html.erb
+++ b/app/views/admin/participants/induction_records/edit_training_status.html.erb
@@ -7,7 +7,7 @@
               training_statuses,
               :id,
               :value,
-              legend: { text: "Training status for #{@participant_profile.name} at #{@induction_record.school_name}", tag: 'h1', size: 'l' },
+              legend: { text: "Training status for #{@participant_profile.full_name} at #{@induction_record.school_name}", tag: 'h1', size: 'l' },
               options: { selected: @induction_record.training_status }) %>
       <%= f.govuk_submit "Save" %>
     <% end %>
--- a/app/views/admin/participants/induction_records/edit_training_status.html.erb
+++ b/app/views/admin/participants/induction_records/edit_training_status.html.erb
@@ -1,16 +1,15 @@
 <% content_for :before_content, govuk_back_link(text: "Back", href: admin_participant_path(@participant_profile)) %>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l">Change the training status for this participant at <%= @induction_record.school_name %></h1>
     <%= form_for @induction_record, url: update_training_status_admin_participant_induction_records_path(@participant_profile, @induction_record), method: :put do |f| %>
       <%= f.govuk_collection_radio_buttons(
               :training_status,
               training_statuses,
               :id,
               :value,
-              legend: { text: "Select training status", tag: 'h1', size: 'l' },
+              legend: { text: "Training status for #{@participant_profile.name} at #{@induction_record.school_name}", tag: 'h1', size: 'l' },
               options: { selected: @induction_record.training_status }) %>
-      <%= f.govuk_submit "Submit" %>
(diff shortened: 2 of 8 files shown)


app/views/admin/participants/induction_records/edit_training_status.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">Change the training status for this participant at <%= @induction_record.school_name %></h1>
+    <%= form_for @induction_record, url: update_training_status_admin_participant_induction_records_path(@participant_profile, @induction_record), method: :put do |f| %>
+      <%= f.govuk_collection_select :training_status,
+                                    training_statuses,
+                                    :id,
+                                    :value,
+                                    label: { text: "Select training status" },
+                                    options: { selected: @induction_record.training_status }
+      %>
```
frankieroberto, line 12: Can we make this radio buttons? (Easier to use and less error-prine - especially when there's only 2 options!)

ltello: Sure, I will convert it

app/views/admin/participants/induction_records/edit_training_status.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    <%= form_for @induction_record, url: update_training_status_admin_participant_induction_records_path(@participant_profile, @induction_record), method: :put do |f| %>
+      <%= f.govuk_collection_radio_buttons(
+              :training_status,
+              training_statuses,
+              :id,
+              :value,
+              legend: { text: "Select training status", tag: 'h1', size: 'l' },
+              options: { selected: @induction_record.training_status }) %>
+      <%= f.govuk_submit "Submit" %>
+    <% end %>
```
frankieroberto, line 14: We’ve currently got 2 `<h1>` tags - we can combine these into a single tag which is also the legend. 

I also think it’s worth including the participant name as well as the school - just to make double-sure that people are updating the right person!

```suggestion
    <%= form_for @induction_record, url: update_training_status_admin_participant_induction_records_path(@participant_profile, @induction_record), method: :put do |f| %>
      <%= f.govuk_collection_radio_buttons(
              :training_status,
              training_statuses,
              :id,
              :value,
              legend: { text: "Training status for #{@participant_profile.name} at #{@induction_record.school_name}", tag: 'h1', size: 'l' },
              options: { selected: @induction_record.training_status }) %>
      <%= f.govuk_submit "Save" %>
    <% end %>
```

app/helpers/admin_helper.rb
```diff
@@ -103,4 +103,11 @@ def admin_participant_role_name(class_name)
       "unknown"
     end
   end
+
+  def training_statuses
```
frankieroberto, line 107: I think we should also list "active"? Otherwise once you’ve set them as deferred or withdrawn, there’s no way to undo and make them active again?

app/helpers/admin_helper.rb
```diff
@@ -103,4 +103,11 @@ def admin_participant_role_name(class_name)
...
+    [
+      OpenStruct.new(id: "deferred", value: "deferred"),
+      OpenStruct.new(id: "withdrawn", value: "withdrawn"),
```
frankieroberto, line 110: Can we capitalise these in the radio button labels?

```suggestion
      OpenStruct.new(id: "deferred", value: "Deferred"),
      OpenStruct.new(id: "withdrawn", value: "Withdrawn"),
```

frankieroberto (changes requested):

app/helpers/admin_helper.rb
ltello: That was considered by Benedict when we discussed the ticket and he discarded it

ltello: SITs can re-register a participant withdrawn or deferred and won't contact support for that, he says

frankieroberto: Hmm really? I’d have thought it’d be better for the same participant profile to get put back into active status after someone defers (or if they un-withdraw) rather than creating a new one?

ltello: Well, internally the service doesn't create a new one but modifies the existing record in the db

app/views/admin/participants/induction_records/edit_training_status.html.erb
```diff
@@ -0,0 +1,15 @@
+<% content_for :before_content, govuk_back_link(text: "Back", href: admin_participant_path(@participant_profile)) %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @induction_record, url: update_training_status_admin_participant_induction_records_path(@participant_profile, @induction_record), method: :put do |f| %>
+      <%= f.govuk_collection_radio_buttons(
+              :training_status,
+              training_statuses,
+              :id,
+              :value,
+              legend: { text: "Training status for #{@participant_profile.name} at #{@induction_record.school_name}", tag: 'h1', size: 'l' },
```
ltello, line 10:
```suggestion
              legend: { text: "Training status for #{@participant_profile.full_name} at #{@induction_record.school_name}", tag: 'h1', size: 'l' },
```

app/helpers/admin_helper.rb
frankieroberto: ah ok! Fair.

frankieroberto (approved):

PR #4174, opened 2023-10-18
Add Snapshot database Backup/Restore

cwrw:
### Context

Add a new snapshot database where production data can be restored and refreshed via Github workflow.

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2535

- Add new snapshot database that will only be in production. This will need to be run manually before pr is merged
- Add workflow which backs up the prod database on demand and uploads to Github artifact 
- Add a restore step which uses `konduit` and connects to another database using `-d` and where the keyvault secret holding the db URL will already be populated. This will restore the snapshot database from the uploaded artifact

### Guidance to review
This was tested on staging and seems to work with the `konduit` commands, which also allows us to connect to the snapshot locally


.github/actions/backup-database/action.yml
```diff
@@ -0,0 +1,44 @@
...
+      uses: azure/setup-kubectl@v3
+
+    - name: Install konduit
+      shell: bash
+      run: make install-konduit
+
+    - name: Backup database
+      shell: bash
+      run: |
+        bin/konduit.sh cpd-ecf-${{ inputs.environment }}-web -- pg_dump -E utf8 --compress=1 --clean --if-exists --no-privileges --no-owner --verbose -f backup-${{ inputs.environment }}.sql.gz
```
ethax-ross, line 38: Should we be concerned about storing production data outside of Azure/in GitHub artifacts, albeit temporarily?

ethax-ross (approved):
Looks great! 💯

.github/actions/backup-database/action.yml
saliceti: Could we combine backup and restore in the same job? So we don't need to store anything

saliceti: Or siimply store it in azure storage, such as: https://github.com/DFE-Digital/register-trainee-teachers/blob/main/.github/workflows/database-backup.yml

cwrw: amended now and combined under one job, if you can both have another look, thanks

ethax-ross: Looks good to me 👍

saliceti (approved):

PR #4177, opened 2023-10-19
[CPDLP-2523] Investigate if there is a data and/or logic bug for EDT GET declarations calls

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2523

* Updated controller and query service classes to use the collection when returning single participant declaration.

### DIFF
--- a/app/services/api/v3/participant_declarations_query.rb
+++ b/app/services/api/v3/participant_declarations_query.rb
@@ -57,6 +57,13 @@ module Api
         scope
       end
 
+      def participant_declaration(id)
+        ParticipantDeclaration.union(
+          declarations_scope,
+          previous_declarations_scope,
+        ).find(id)
+      end
+
     private
 
       delegate :lead_provider, to: :cpd_lead_provider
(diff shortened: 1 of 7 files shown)


app/services/api/v3/participant_declarations_query.rb
```diff
@@ -57,6 +57,13 @@ def participant_declarations_from(paginated_join)
         scope
       end
 
+      def find(id)
```
cwrw, line 60: suggestion: could we name it maybe `participant_declaration` similar to other services singular methods?

spec/requests/api/v1/participant_declarations_spec.rb
```diff
@@ -759,6 +759,63 @@
           expect(response.status).to eq 404
         end
       end
+
+      context "when querying a single old participant declaration owned by another provider" do
```
cwrw, line 763: could you add a spec for the voided endpoint (that it won't allow you to void a declaration from another provider) here and in other versions?

cwrw (commented):
looking good, some minor comments

app/services/api/v3/participant_declarations_query.rb
mooktakim: Updated

spec/requests/api/v1/participant_declarations_spec.rb
mooktakim: added tests

cwrw (approved):
looks good 👍

PR #4189, opened 2023-10-23
Yarn 4.0.0

peteryates:
[Yarn 4.0.0 was released](https://yarnpkg.com/blog/release/4.0). This PR is an experiment to try it out. It has several interface improvements and is much faster than old Yarn 1.22.

- Upgrade to Yarn 4.0.0
- Run yarn for yarn 4.0.0


frankieroberto:
@peteryates could also explore switching to using the npm native installer and lockfile? I think it’s pretty similar to Yarn nowadays?

peteryates:
Absolutely, just did this for experimentation reasons. I've used [esbuild](https://esbuild.github.io/) elsewhere and would love to try that here too.

peteryates:
@frankieroberto this now seems decent and it's definitely faster locally. I'm happy to switch to a different package manager but in a perfect world would like to get rid of webpack, which would definitely simplify things and let us clear out a load of dependencies from the `package.json` too.

frankieroberto:
@peteryates as I understand it, the package manager (yarn vs npm) is separate from the build process (webpacker, grunt, etc etc)?

I think Rails still defaults to Yarn over npm though, so happy to stick with that. (I just need to remember to run `yarn install` instead of `npm install`). Sounds like Yarn is still quicker than npm as is runs stuff in parallel.

peteryates:
@frankieroberto usually yeah, but I'm totally up for giving [bun](https://bun.sh/) a try!

.yarn/install-state.gz
leoapost, line 1: The Yarn docs [mention](https://yarnpkg.com/getting-started/qa#:~:text=yarn%2Finstall%2Dstate.,your%20workspaces%20all%20over%20again.) we shouldn't ever have to commit this file.

leoapost:
Do we need to update the `.gitignore`?

Here are some [suggestions](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored) from the Yarn docs.

.yarn/install-state.gz
peteryates: Ah good point, will fix in a sec

leoapost:
@peteryates do we need to update the [Dockerfile](https://github.com/DFE-Digital/early-careers-framework/blob/yarn-4.0.0/Dockerfile#L41)?

The docs mention that the `yarn install --production` is [renamed](https://yarnpkg.com/migration/guide#renamed-commands) to `yarn workspaces focus --all --production`.

leoapost (approved):

PR #4211, opened 2023-10-30
[CPDLP-2545] NPQ Maths - As an SRM I want to view NPQ Maths on the NPQ statement

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2545

* Updated NPQ statement page to show `special_course=false`
* Include a new section "Standalone payments" that show `special_course=true`

### DIFF
--- a/app/services/oneoffs/npq/set_special_course_for_npq_contracts.rb
+++ b/app/services/oneoffs/npq/set_special_course_for_npq_contracts.rb
@@ -10,28 +10,35 @@ module Oneoffs::NPQ
 
     def call
       ActiveRecord::Base.transaction do
+        old_contracts = []
+
         statements.each do |statement|
-          npq_lead_provider = statement.npq_lead_provider
-          version = statement.contract_version
+          NPQContract.where(
+            version: statement.contract_version,
+            npq_lead_provider: statement.npq_lead_provider,
+            cohort: statement.cohort,
+          ).each do |contract|
+            # Contract already set to true
+            next if contract.special_course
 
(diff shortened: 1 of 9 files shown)


cwrw (commented):
looks good, some minor comments.

However when testing locally I noticed something in this service: https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/oneoffs/npq/set_special_course_for_npq_contracts.rb#L25

- We should not fail if no contract is found as not all providers have contracts for the special course so we should just go to the next statement
- currently if run it will create 174 new contracts, due to the following check: `next if old_contract.special_course` as this will never be true as its the old contract and not the new one.
I think we need to amend this maybe with this pr before merging/running in other envs :)

app/services/oneoffs/npq/set_special_course_for_npq_contracts.rb
```diff
@@ -10,28 +10,35 @@ def initialize(cohort_year:, payment_date_range:, course_identifier:)
 
     def call
       ActiveRecord::Base.transaction do
-        statements.each do |statement|
-          npq_lead_provider = statement.npq_lead_provider
-          version = statement.contract_version
+        old_contracts = []
```
cwrw, line 13: thanks for fixing this, any chance we can update the specs for this service to ensure we don't miss anything? Also there are failures in the specs from I think the versioning in factories, happy to have a look together/revert if its too gnarly :)

cwrw (approved):
looks great and tested on the snaphot, thanks @mooktakim !

PR #4212, opened 2023-10-31
Avoid overpaying uplifts on statements

ethax-ross:
[Jira-2546](https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2546)

### Context

We currently restrict the declaration payments to the defined upper band limit, but we do not restrict payments on uplifts.

Going forward we want to also restrict the uplift payments in the same way, such that when the number of uplift payments reaches the upper band limit we will not make any more uplifts payable (until refundable uplifts brings the available amount back within the upper band limit).

- Avoid overpaying uplifts on statements

We do this by capping the `uplift_breakdown` values in the `OutputCalculator`.

For example:

> Upper band limit: 2

| Statement | Billable Amount | Refundable Amount | Uplift Breakdown                                                |
| --------- | --------------- | ----------------- | --------------------------------------------------------------- |
| A         | 3               | 0                 | available: 2, prev: 0, count: 2, additions: 2, subtractions: 0  |
| B         | 2               | 1                 | available: 1, prev: 2, count: 1, additions: 1, subtractions: 1  |
| C         | 0               | 2                 | available: 2, prev: 2, count: -2, additions: 0, subtractions: 2 |

### DIFF
--- a/app/services/finance/ecf/output_calculator.rb
+++ b/app/services/finance/ecf/output_calculator.rb
@@ -29,12 +29,10 @@ module Finance
       end
 
       def uplift_breakdown
-        available_uplifts = upper_band_limit - previous_fill_level_for_uplift + current_refundable_count_for_uplift
-
         @uplift_breakdown ||= {
           previous_count: previous_fill_level_for_uplift,
-          count: [available_uplifts, current_billable_count_for_uplift - current_refundable_count_for_uplift].min,
-          additions: [available_uplifts, current_billable_count_for_uplift].min,
+          count: current_billable_count_for_uplift - current_refundable_count_for_uplift,
+          additions: current_billable_count_for_uplift,
           subtractions: current_refundable_count_for_uplift,
         }
       end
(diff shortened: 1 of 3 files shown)


app/services/finance/ecf/output_calculator.rb
```diff
@@ -207,7 +213,7 @@ def previous_fill_level_for_uplift
           .where("participant_declarations.sparsity_uplift = true OR participant_declarations.pupil_premium_uplift = true")
           .count
 
-        billable - refundable
+        [upper_band_limit, billable - refundable].min
```
cwrw, line 216: I think we probably need to use the remaining pot size here rather than the band max, as we need to compare it with the total number of starts (regardless of uplifts) I think rather the band limit scoped to uplifts, happy to chat tomorrow

ethax-ross:
Closing this as we are re-thinking the task.

PR #4217, opened 2023-10-31
[CST-2058] Enable super user admins to activate the participants induction statuses

leoapost:
### Context

- Ticket: CST-2058

SITs withdraw participants or set them as leaving by mistake. 

We want to enable super user admins to set them back to active.
For the leaving ones, we also want the school transfer to be `true`. This will display to a provider as a transfer from school A, into school A.

- Add a new service class with the logic required to set the `ParticipantProfile@status` and `InductionRecord@induction_status` to `active`
- Add a new admin page to display the new statuses with a button to confirm the changes
- Display conditionally a `Change` link in the Participant->Statuses page that leads to the new admin page

### Guidance to review
1. Go to the Admin pages as a super user admin
2. Find an active or completed ECF participant
3. Confirm there is no "Change" link in the "induction status" row as a super user admin, but not as regular admin
4. Set the induction status in their induction record to either "withdraw" or "leaving"
5. Go back to their Profile->Statuses page and confirm the "Change" link is there
6. Click "Change" to take you to the induction status activation page
7. Click "Confirm" to set the statuses to active and redirect you back to the Statuses page
8. Confirm both the `induction status` in the IR section and the `status` in the Profile section are "active"

### DIFF
--- a/app/views/admin/participants/statuses/show.html.erb
+++ b/app/views/admin/participants/statuses/show.html.erb
@@ -129,7 +129,7 @@ end %>
   sl.with_row do |row|
     row.with_key(text: "induction status")
     row.with_value(text: govuk_tag(text: @participant_presenter.relevant_induction_record&.induction_status || "No Induction Record Found", colour: "grey"))
-    if policy(@participant_presenter.participant_profile).edit_induction_status? && ['withdrawn', 'leaving'].include?(@participant_presenter.relevant_induction_record&.induction_status)
+    if allowed_to_change_induction_status?(@participant_presenter)
       row.with_action(
         href: edit_admin_participant_change_induction_status_path(@participant_presenter.participant_profile),
         visually_hidden_text: "induction status"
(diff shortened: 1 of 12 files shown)


app/views/admin/participants/statuses/show.html.erb
```diff
@@ -129,6 +129,12 @@ end %>
   sl.with_row do |row|
     row.with_key(text: "induction status")
     row.with_value(text: govuk_tag(text: @participant_presenter.relevant_induction_record&.induction_status || "No Induction Record Found", colour: "grey"))
+    if policy(@participant_presenter.participant_profile).edit_induction_status? && ['withdrawn', 'leaving'].include?(@participant_presenter.relevant_induction_record&.induction_status)
```
lazydevorg, line 132: it seems quite a big piece of logic that can be put in the participant presenter (or another class).
It would be good to replace this line with something like `@participant_presenter.can_change_induction_status?` (or something like that).

lazydevorg (changes requested):

lazydevorg (approved):

PR #4219, opened 2023-10-31
[CPDLP-2525] Add schedule identifier as an optional field on the request body when accepting an npq application

leandroalemao:
### Context

- Ticket: [CPDLP-2525](https://dfedigital.atlassian.net/browse/CPDLP-2525)

This enables lead providers, if they wish to, to specify schedule when accepting an npq application.

[CPDLP-2525]: https://dfedigital.atlassian.net/browse/CPDLP-2525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -33,7 +33,7 @@ def reject
       end
 
       def accept
-        service = ::NPQ::Application::Accept.new(npq_application:)
+        service = ::NPQ::Application::Accept.new({ npq_application: }.merge(accept_npq_application_params["attributes"] || {}))
```
cwrw, line 36: Im not sure I remember what we agreed, is this only for v3, or for all versions? as the release note is stating it's only for v3 🤔

spec/requests/api/v1/npq_applications_spec.rb
```diff
@@ -377,6 +377,89 @@
         expect(parsed_response.dig("errors", 0, "detail")).to eql("Once rejected an application cannot change state")
       end
     end
+
+    context "when schedule identifier is in the params", with_feature_flags: { accept_npq_application_can_change_schedule: "active" } do
```
cwrw, line 381: pending which version this is for (v3 vs all versions) I think we might need to add this test to all request npq application versions

app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -68,6 +68,16 @@ def access_scope
       def npq_application
         @npq_application ||= npq_lead_provider.npq_applications.includes(:cohort, :npq_course, participant_identity: [:user]).find(params[:id])
       end
+
+      def accept_npq_application_params
+        return {} unless FeatureFlag.active?(:accept_npq_application_can_change_schedule)
+
+        params
+          .require(:data)
```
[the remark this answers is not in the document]
cwrw: we could then also add the usual error message we have for params (See other controllers)

cwrw (commented):
nicely done! a few qs and suggestions but otherwise 👍

cwrw (approved):
looks great, thanks Leandro, just one last minor comment otherwise all good 👍

PR #4220, opened 2023-11-01
Display authorised for payment date in local timezone

ethax-ross:
### Context

We currently display the authorised for payment date in the same timezone that we store it in the database (UTC). Instead, we want to format the date and display it in the London timezone.

- Display authorised for payment date in local timezone

This is a app-wide issue, however applying a 'fix all' solution of setting the time zone in the application config is risky (it would change it everywhere, including in our API serializers). Instead, we are going to fix this on a case-by-case basis by converting the dates into London timezone prior to formatting them.

### Guidance to review

If we find we are going to change this in lots of other places it may be be worth creating a date formatting service.


ethax-ross:
Closing to fix review app

ethax-ross:
Closing as review app is jammed

PR #4221, opened 2023-11-01
Display authorised for payment date in local timezone

ethax-ross:
### Context

We currently display the authorised for payment date in the same timezone that we store it in the database (UTC). Instead, we want to format the date and display it in the London timezone.

- Display authorised for payment date in local timezone

This is a app-wide issue, however applying a 'fix all' solution of setting the time zone in the application config is risky (it would change it everywhere, including in our API serializers). Instead, we are going to fix this on a case-by-case basis by converting the dates into London timezone prior to formatting them.

### Guidance to review

If we find we are going to change this in lots of other places it may be be worth creating a date formatting service.

To review, check the authorised time [here](https://manage-training-for-early-career-teachers.education.gov.uk/finance/ecf/payment_breakdowns/c3bc3cee-a636-42d6-8324-c033a6c38d31/statements/140c964b-3a29-433d-a0d6-059d10db4eeb) and compare it when running the PR locally against the snapshot DB; it should be an our ahead due to BST.

### DIFF
--- a/app/views/finance/ecf/statements/show.html.erb
+++ b/app/views/finance/ecf/statements/show.html.erb
@@ -118,7 +118,7 @@
               <% if authorise_for_payment_button_visible?(@statement) %>
                 <%= button_to t("finance.statements.payment_authorisations.button"), new_finance_statement_payment_authorisation_path(@statement), method: :get, class: "govuk-button govuk-button--primary" %>
               <% elsif @statement.marked_as_paid? %>
-                <%= govuk_tag(text: t("finance.statements.payment_authorisations.tag.content", statement_marked_as_paid_at: @statement.marked_as_paid_at.strftime("%-I:%M%P on %-e %b %Y"))) %>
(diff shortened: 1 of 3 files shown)


leandroalemao (approved):
lgtm 👍 😄 

As an improvement.. we could also add a new format in `config/initializers/date_formats.rb` and reuse it in everywhere we have to display this formatted marked as paid at date.

```
Time::DATE_FORMATS[:statement_marked_as_paid_at] = "%-I:%M%P on %-e %b %Y"
```

```
ie: statement.marked_as_paid_at.in_time_zone("London").to_fs(:statement_marked_as_paid_at)
```

ethax-ross:
> lgtm 👍 😄
> 
> As an improvement.. we could also add a new format in `config/initializers/date_formats.rb` and reuse it in everywhere we have to display this formatted marked as paid at date.
> 
> ```
> Time::DATE_FORMATS[:statement_marked_as_paid_at] = "%-I:%M%P on %-e %b %Y"
> ```
> 
> ```
> ie: statement.marked_as_paid_at.in_time_zone("London").to_fs(:statement_marked_as_paid_at)
> ```

Yeah standardising the date formatting is a good idea, I'd been keen to tackle that as part of a wider date clean up though so I'll leave this PR as is for now 👍

PR #4222, opened 2023-11-01
Remove access restriction to ECF providers only

cwrw:
### Context
NPQ only Providers are getting 403 errors when attempting to access declaration endpoints.
- Ticket: n/a

Declarations endpoint serves both ECF and NPQ, so both providers require access. This being there means that only ECF providers have access to declarations which is not always the case.

### Guidance to review
Test with NPQ provider hitting declarations endpoint v3


ethax-ross (approved):
Looks good to me 👍

PR #4224, opened 2023-11-01
Update participant count for schools to use IRs

ethax-ross:
### Context

When we find participants for schools we are not using the induction records and therefore we don't match all the participants (as we do in the API endpoints).

- Update participant count for schools to use IRs

Update queries to go via the induction records.

- Scope participants query to lead provider

We want to scope the participants to the signed in lead provider; to do this we find the latest induction record by lead provider and also cross-reference it with the cohort. If a participant has transferred school it should only appear in counts once per lead provider.

- ~~Cache mentor/ect profile queries~~

~~Cache the query results so we don't hit the database multiple times when rendering the mentor and ECT count.~~

^ this change wasn't enough of an improvement; see the following commit:

- Optimise ect/mentor counts

Previously we were calculating the counts per-school, however as we need to evaluate the latest induction records by lead provider as part of this it was too slow.

Instead, the query has been updated to support multiple partnerships and extracted into a `Participants::ActiveByPartnership` service. We calculate this once per request and pass it down into the views.

### Guidance to review

To test this I logged in as a supplier for Ambition and searched for 'Three Rivers' under the 2023 cohort. I then found a school where the counts are different in `main` from this branch (136348 has 3 ECTs vs 4 in this branch). Hitting the participants endpoint as Ambition filtered to the 2023 cohort I then manually checked how many ECTs were returned for 136348 and it was 4, which matches the amount calculated in this branch.


app/models/school.rb
```diff
@@ -205,4 +205,18 @@ def slug_candidates
   def cip_only_establishment_type?
     CIP_ONLY_TYPE_CODES.include?(school_type_code)
   end
+
+  def active_profiles_for(cohort:, type: ParticipantProfile)
+    school_cohort = school_cohorts.find_by(cohort:)
```
cwrw, line 210: discussed in person: do we want the latest school (latest Induction record) vs if they ever were with that school?

leandroalemao (approved):
lgtm 👍 👏

app/services/participants/active_count_by_partnership.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    def query_counts
+      @query_counts ||= ParticipantProfile::ECF
+        .joins(induction_records: :induction_programme)
+        .joins("JOIN (#{latest_induction_records_join.to_sql}) AS latest_induction_records
+            ON latest_induction_records.latest_id = induction_records.id")
+        .where(
+          induction_programmes: { partnership_id: partnerships },
+          status: :active,
```
cwrw, line 33: could we check the training status on IRs for this status?

cwrw (approved):
🙌 thanks Ross

PR #4227, opened 2023-11-02
Remove the banner around completed ECTs

frankieroberto:
### Context

We can remove this banner, now that we are pulling the completed induction status for ECTs from the DQT database and reflecting this in our interface.

Removing this banner improve the visibility of other information in the service, and avoids users from having to read information unnecessarily. 

🗂️  [Jira ticket](https://dfedigital.atlassian.net/browse/CST-1998)  [Second Jira ticket](https://dfedigital.atlassian.net/browse/CST-2100)

#### Before

Screenshot 2023-11-02 at 15 50 52

#### After

Screenshot 2023-11-02 at 15 51 02


leoapost (approved):
Looks good

PR #4232, opened 2023-11-04
CST-2075:  New mentors can be linked to the school's previous providers

ltello:
### Context
At the moment, SITs who are registering a new mentor have to register them with the training provider that’s partnered with the school for the 2023 cohort.

However, quite often SITs want a mentor to be training with a different provider. A common use case is that the school is partnered with provider A for 2022, and provider B for 2023. The mentor is mentoring a 2022 participant, so the school want them to be training with provider A.

We therefore want to allow schools to register a mentor either with this year’s partnership, or last year’s partnership if it’s available.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2075)

- Add new step at the end of the add_participant journey for SITs to choose the partnership where to sit the mentor in (school current default providers, school previous years default providers, other providers)
- Add new screen to send the SIT to support when they choose "other providers" for their mentor.

### DIFF
--- a/app/forms/schools/add_participants/add_wizard.rb
+++ b/app/forms/schools/add_participants/add_wizard.rb
@@ -11,13 +11,23 @@ module Schools
           start_term
           cannot_add_registration_not_yet_open
           cannot_add_yourself_as_ect
+          cannot_add_mentor_to_providers
           yourself
           need_training_setup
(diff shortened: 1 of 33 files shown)


tonyheadford (approved):
LGTM!

PR #4237, opened 2023-11-06
Performance testing

mattonfoot:
### Context

Adds basic API performance tests and HTML reports to CI on pull requests

- add scripts for K6 (used across DfE)
- add steps to CI
- add HTML report templates

### Guidance to review

- review scripts
- review artefacts of performance test run


performance/run_performance_tests_locally.sh
```diff
@@ -0,0 +1,20 @@
...
+mkdir ../reports
+
+docker-compose up -d web
+docker-compose up seed
+docker-compose up k6
+docker-compose cp k6:/home/k6/${PERF_REPORT_FILE} ../reports/${PERF_SCENARIO}-report.json
+
+node dfe-k6-reporter.js ../reports/${PERF_SCENARIO}-report.json ../reports/${PERF_SCENARIO}-report.html ../reports/${PERF_SCENARIO}-summary.md
+
+docker-compose down
```
peteryates, line 20: Nice, love how simple this is.

peteryates (approved):
Looks good

mattonfoot:
I need to clean up this PR it seems to have gotten mixed up with some recent changes to main branch

config/initializers/rack_attack.rb
```diff
@@ -32,21 +32,23 @@ def remote_ip
     request.path == "/check"
   end
 
-  throttle("Login attempts by ip", limit: 5, period: 20.seconds) do |request|
-    if request.path == "/users/sign_in" && request.post?
-      request.remote_ip
+  unless Rails.env.performance?
```
cwrw, line 35: Maybe we can try setting `Rack::Attack.enabled = false` in the environment config instead, to avoid wrapping the settings here?

config/environments/performance.rb
```diff
@@ -0,0 +1,24 @@
...
+  # Used to handle HTTP_X_WITH_SERVER_DATE header for server side datetime overwrite
+  config.middleware.use TimeTraveler
+
+  config.require_master_key = false
+
+  config.public_file_server.enabled = true
+  config.force_ssl = false
+  config.action_mailer.perform_deliveries = false
+  config.active_support.deprecation = :log
+  config.log_level = :debug
```
cwrw, line 17: do we need debug, as this might be noisy for logging and maybe we can lower (just to avoid getting pinged around too many logs from ops 😅 like we've gotten before)

.github/workflows/aks_deploy.yml
```diff
@@ -89,12 +89,37 @@ jobs:
           statuscake-api-token: ${{ secrets.STATUSCAKE_API_TOKEN }}
 
       - name: Post sticky pull request comment
-        if: github.event_name == 'pull_request'
```
cwrw, line 92: not sure why this got removed?

package-lock.json
```diff
@@ -0,0 +1,11906 @@
+{
+  "name": "early-careers-framework",
```
cwrw, line 2: I don't think we usually commit this?

performance/scenarios/api-load-test.js
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable import/extensions, import/no-unresolved */
+
+import http from 'k6/http';
+import { group, check, sleep } from 'k6';
+
+import { randomIntBetween, fromEnv } from '../common/utils.js';
+
+// eslint-disable-next-line import/prefer-default-export
+export const apiLoadTest = () => {
+  const path = '/participants/ecf';
```
cwrw, line 10: Just to confirm are we only testing this endpoint in the API?

cwrw (commented):
Thanks Matt, some good work here. I was wondering about how long this usually takes to run (I had a look and seems like around 18 minutes), will this slow down pr merges as the tests are quite fast now (after your improvements)? Also wondering does this run once or on every commit, and will the comments get added or updated to show results? Also maybe worth chatting in person around what would happen if things fail, as I didn't quite understand the results and what to watch out for. Lastly maybe we can allow these steps to be optional in the pipeline to pass incase anything fails to avoid blocking merging? Ah also we might want to avoid running those on dependency bumps?

performance/scenarios/api-load-test.js
mattonfoot: No, this is just a framework to allow us to test, we need to collaborate on what to test, what is a useful set of tests and what are appropriate thresholds.

This PR is just putting something in place to allow developers to see the effects of their code changes in case it affects performance or resilience in anyway. We need to continue and define what a performant and resilient system is.

package-lock.json
mattonfoot: Removed

config/environments/performance.rb
mattonfoot: changed to depreciation :silence and log_level :fatal, seeing as you can't look at them in CI anyway

config/initializers/rack_attack.rb
mattonfoot: used that instead as I did not know about it

mattonfoot:
> Thanks Matt, some good work here. I was wondering about how long this usually takes to run (I had a look and seems like around 18 minutes), will this slow down pr merges as the tests are quite fast now (after your improvements)? Also wondering does this run once or on every commit, and will the comments get added or updated to show results? Also maybe worth chatting in person around what would happen if things fail, as I didn't quite understand the results and what to watch out for. Lastly maybe we can allow these steps to be optional in the pipeline to pass incase anything fails to avoid blocking merging? Ah also we might want to avoid running those on dependency bumps?

This runs in parallel with the other processes but yes currently I have set it to run for about 5-10 mins and seeding 10K participants take about 30 mins to complete which is not great either from a time taken perspective or a useful test perspective.

If we are happy with the over all process, I suggest we consider a database restore from a backup with more like 1-200K participants. Additionally I think we should look to have a proper over night performance test running against a proper deployed performance environment with a restored database. But, I wanted to get something in reporting something to us before I moved on to that.

PR #4238, opened 2023-11-06
Remove old PaaS-specific config from database.yml

peteryates:
We no longer need this as PaaS is a thing of the past.


cwrw (approved):

peteryates:
Fixed in #4330

PR #4240, opened 2023-11-07
[Bug-fix] Fix method name used in view

tonyheadford:
### Context

Spotted an error occurring in Sentry related to a bad method name. Call was missing the `?` at the end of the method name


ltello (approved):

ltello:
Thanks!

PR #4244, opened 2023-11-08
Optimise training record state for delivery partners

ethax-ross:
[Jira-2485](https://dfedigital.atlassian.net/browse/CPDLP-2485)

### Context

We have a new/optimised way of calculating the training record states for participants. In order to use this new service we need to provide it with the latest induction records for the participants. We do this with appropriate bodies at the moment and want to replicate this in the delivery partner controller/views.

- Pass training_record_state into DeliveryPartnerParticipantStatusTag

Pass the training record states in as an argument as we will be pre-calculating this at the controller-level. Remove unused code from `AppropriateBodyParticipantStatusTag`. These view components are ending up very similar; it may be worth reconciling them in a future PR.

- Add InductionRecordsQuery for delivery partners

The training state service accepts induction records instead of participant profiles (which we currently query in the delivery partner controllers).

Add `InductionRecordsQuery` for delivery partners to bring it inline with how appropriate bodies is structured. Pull the training record state subqueries into a concern and use shared examples.

- Change delivery partner ParticipantsFilter to use induction records

As we will be querying over induction records instead of participant profiles in the controller it makes sense to update the filter to also work over induction records. We will be passing in the pre-calculated training record states here as well, which is inline with how appropriate bodies work.

Includes a minor tidy up in the appropriate bodies `ParticipantsFilter` and adds a spec file for the delivery partners `ParticipantsFilter`.

- Replace ParticipantsSerializer with InductionRecordsSerializer

As we are moving away from querying participant profiles to induction records we will need to replace this serializer. Also updates to use pre-calculated in training record states and adds test coverage.

- Update delivery partner controller/views for new training states

Use the new query, filter and training states services to populate the delivery partner views. This is more efficient and results in a ~3x performance improvement.

### Guidance to review

Best reviewed by-commit.

Tested agains the snapshot DB and it shows a reasonable improvement; particularly when filtering, which I gave up waiting for on `main` but this branch returns in ~5 seconds (still not quick, but _much_ quicker!).


spec/services/delivery_partners/induction_records_query_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    it_behaves_like "a query optimised for calculating training record states"
+
+    it "returns latest induction record for delivery partner" do
+      expect(subject.induction_records).to match_array([induction_record])
+    end
+
+    context "when there are more induction records for the same delivery partner" do
+      let!(:latest_induction_record) { create(:induction_record, participant_profile:, induction_programme:) }
+
+      it "returns latest induction record for appropriate body" do
```
cwrw, line 33:
```suggestion
      it "returns latest induction record for delivery partner" do
```

spec/services/delivery_partners/induction_records_query_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        expect(subject.induction_records).to match_array([latest_induction_record])
+      end
+    end
+
+    context "when there are newer induction records for a different appropriate body" do
```
cwrw, line 38:
```suggestion
    context "when there are newer induction records for a different delivery partner" do
```

cwrw (approved):
fantastic, nicely done! 🙌

app/services/delivery_partners/participants_filter.rb
```diff
@@ -36,27 +37,25 @@ def filter_query(scoped, query)
...
         user_teacher_profile_trn
-        induction_records_induction_programme_partnership_lead_provider_name
-        induction_records_school_cohort_school_name
-        induction_records_school_cohort_school_urn
+        induction_programme_partnership_lead_provider_name
+        school_cohort_school_name
+        school_cohort_school_urn
       ].join("_or_")
 
       scoped.includes(
```
cwrw, line 45: suggestion: whilst filtering by mentor/academic year there are a lot of n+1 coming up (some queries however are cahed and some aren't) not sure if we're missing an includes here?

ethax-ross: Good catch, I've fixed up a commit with additional includes. It might be worth considering a gem like goldiloader for our app, it would simplify the queries we have around filtering by handling the includes for us.

cwrw (approved):

PR #4245, opened 2023-11-08
[CST-1874] Decommission cohortless_dashboard feature flag - 02

lazydevorg:
### Context

- Ticket: CST-1874

### DIFF
--- a/app/forms/schools/cohorts/setup_wizard/success.rb
+++ b/app/forms/schools/cohorts/setup_wizard/success.rb
@@ -100,8 +100,7 @@ module Schools
         end
 
         def should_send_the_pilot_survey?
-          cohort.start_year == 2023 && FeatureFlag.active?(:cohortless_dashboard, for: school) &&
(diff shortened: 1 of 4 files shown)


app/forms/schools/cohorts/setup_wizard/success.rb
```diff
@@ -100,8 +100,7 @@ def what_changes_programme
         end
 
         def should_send_the_pilot_survey?
-          cohort.start_year == 2023 && FeatureFlag.active?(:cohortless_dashboard, for: school) &&
-            expect_any_ects? && current_user.induction_coordinator?
+          cohort.start_year == 2023 && expect_any_ects? && current_user.induction_coordinator?
```
tonyheadford, line 103: Can you check this is the right thing to do, looks like it will start ending the pilot survey to anyone setting up in this year, perhaps it needs to be removed instead?  Can you check with Kate and Nicola please.

tonyheadford (approved):

PR #4246, opened 2023-11-08
[CST-2088] Add form for changing Appropriate Body on Cohorts

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CST-2088

- Adds form for admins to modify appropriate bodies on cohorts.
- Adds link to admin cohorts page to the appropriate body change form.

### DIFF
--- a/app/components/admin/schools/cohort_component.html.erb
+++ b/app/components/admin/schools/cohort_component.html.erb
@@ -32,6 +32,13 @@
               row.with_action(text: "Change", visually_hidden_text: "training programme",  href: change_programme_href)
             end
           end
+
+          summary_list.with_row do |row|
+            row.with_key(text: "Appropriate body")
+            row.with_value(text: school_cohort&.appropriate_body&.name || "No appropriate body")
+
+            row.with_action(text: "Change", visually_hidden_text: "appropriate body for their #{school_cohort.cohort.start_year} programme", href: change_appropriate_body_href)
+          end
         end
       %>
     <% end %>
@@ -55,12 +62,12 @@
             row.with_value(text: materials)
             row.with_action(text: "Change", visually_hidden_text: "materials supplier", href: change_materials_href)
           end
+        end
 
-          summary_list.with_row do |row|
-            row.with_key(text: "Appropriate body")
-            row.with_value(text: school_cohort&.appropriate_body&.name || "No appropriate body")
-            row.with_action(text: "Change", visually_hidden_text: "appropriate body for their #{school_cohort.cohort.start_year} programme", href: change_appropriate_body_href)
-          end
+        summary_list.with_row do |row|
+          row.with_key(text: "Appropriate body")
+          row.with_value(text: school_cohort&.appropriate_body&.name || "No appropriate body")
+          row.with_action(text: "Change", visually_hidden_text: "appropriate body for their #{school_cohort.cohort.start_year} programme", href: change_appropriate_body_href)
         end
       end
     %>
--- a/spec/features/admin/schools/changing_appropriate_bodies_spec.rb
+++ b/spec/features/admin/schools/changing_appropriate_bodies_spec.rb
@@ -4,6 +4,7 @@ require "rails_helper"
 
 RSpec.feature "Admin managing appropriate bodies", js: true, rutabaga: false do
   scenario "Admin changes appropriate body for a 2021 cohort to IStip" do
+    given_there_is_a_cip_school_in(2021)
     i_should_be_able_to_navigate_to_the_edit_page(2021)
     and_there_should_be_a_radio_button_for "Independent Schools Teacher Induction Panel (IStip)"
     and_there_should_be_a_radio_button_for "National Teacher Accreditation (NTA)"
@@ -13,6 +14,7 @@ RSpec.feature "Admin managing appropriate bodies", js: true, rutabaga: false do
   end
 
   scenario "Admin changes appropriate body for a 2021 cohort to NTA" do
+    given_there_is_a_cip_school_in(2021)
     i_should_be_able_to_navigate_to_the_edit_page(2021)
     and_there_should_be_a_radio_button_for "Independent Schools Teacher Induction Panel (IStip)"
     and_there_should_be_a_radio_button_for "National Teacher Accreditation (NTA)"
@@ -22,6 +24,7 @@ RSpec.feature "Admin managing appropriate bodies", js: true, rutabaga: false do
   end
 
   scenario "Admin changes appropriate body for a 2021 cohort to a teaching school hub" do
+    given_there_is_a_cip_school_in(2021)
     i_should_be_able_to_navigate_to_the_edit_page(2021)
     and_there_should_be_a_radio_button_for "Independent Schools Teacher Induction Panel (IStip)"
     and_there_should_be_a_radio_button_for "National Teacher Accreditation (NTA)"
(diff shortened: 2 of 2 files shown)


app/components/admin/schools/cohort_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="govuk-!-margin-bottom-9">
+<div class="govuk-!-margin-bottom-9" id="cohort-<%= cohort.start_year %>">
```
Lockyy, line 1: I've added this in so I could use `within "#cohort-2023"` in the tests.

app/forms/admin/schools/cohorts/appropriate_bodies/update_form.rb
```diff
@@ -0,0 +1,72 @@
...
+
+            errors.add(:appropriate_body_id, "Please select an appropriate body")
+          end
+
+          def teaching_school_hub_selected?
+            appropriate_body_id == TEACHING_SCHOOL_HUB_ID
+          end
+
+          def appropriate_bodies
+            AppropriateBody.active_in_year(cohort_start_year)
```
Lockyy, line 54: This'll exclude NTA from 2023+ cohorts by checking against the disable_from_year column.

app/forms/admin/schools/cohorts/appropriate_bodies/update_form.rb
```diff
@@ -0,0 +1,72 @@
...
+          include ActiveModel::Model
+          include ActiveRecord::AttributeAssignment
+          include ActiveModel::Serialization
+
+          TEACHING_SCHOOL_HUB_ID = "teaching_school_hub"
+
+          attr_accessor :appropriate_body_id, :teaching_school_hub_id, :school_cohort
+
+          validates :appropriate_body_id, presence: { message: I18n.t("errors.appropriate_body.blank") }
+          validates :teaching_school_hub_id, presence: { message: I18n.t("errors.appropriate_body.blank") }, if: :teaching_school_hub_selected?
```
frankieroberto, line 17: Could we use a slightly more specific error message for this one – "Select teaching school hub"?

app/components/admin/schools/cohort_component.html.erb
```diff
@@ -59,6 +59,7 @@
           summary_list.with_row do |row|
             row.with_key(text: "Appropriate body")
             row.with_value(text: school_cohort&.appropriate_body&.name || "No appropriate body")
+            row.with_action(text: "Change", visually_hidden_text: "appropriate body", href: change_appropriate_body_href)
```
frankieroberto, line 62: This is my fault for not thinking about this before, but as there can be multiple of these links per page, they need to be disambiguated for screen reader users (using links mode) - so we should include the year within the visually-hidden text.

```suggestion
            row.with_action(text: "Change", visually_hidden_text: "appropriate body for their #{ cohort.start_year} programme", href: change_appropriate_body_href)
```

app/forms/admin/schools/cohorts/appropriate_bodies/update_form.rb
```diff
@@ -0,0 +1,72 @@
...
+          delegate :teaching_school_hubs, to: :appropriate_bodies
+
+        private
+
+          def validate_selected_appropriate_body
+            return if selected_appropriate_body_id.blank?
+            return if appropriate_bodies.pluck(:id).include?(selected_appropriate_body_id)
+
+            errors.add(:appropriate_body_id, "Please select an appropriate body")
+          end
```
frankieroberto, line 47: When does this error appear? I can't seem to trigger it at all.

app/views/admin/schools/cohorts/appropriate_bodies/edit.html.erb
```diff
@@ -0,0 +1,41 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for @form, url: admin_school_cohort_appropriate_body_path, method: :patch do |form| %>
+      <%=
+        form.govuk_radio_buttons_fieldset(
+          :appropriate_body_id,
+          legend: { text: @question, tag: "h1", size: "l" },
+          hint: { text: "This will change the default for any new ECTs registered, but will not affect any existing ECTs" },
+          caption: { text: school.name }
+        ) do
+      %>
```
frankieroberto, line 16: Minor one, but I think we should move this caption out of the h1 here.  The captions can either be in or out of the h1, which visually appears the same, but for screen reader users it’s better here if the caption it outside, I think, as the school name doesn’t naturally form part of the question.

```suggestion
      <span class="govuk-caption-l"><%= school.name %></span>
      <%=
        form.govuk_radio_buttons_fieldset(
          :appropriate_body_id,
          legend: { text: @question, tag: "h1", size: "l" },
          hint: { text: "This will change the default for any new ECTs registered, but will not affect any existing ECTs" }
        ) do
      %>
```

app/views/admin/schools/cohorts/appropriate_bodies/edit.html.erb
```diff
@@ -0,0 +1,41 @@
...
+
+          <%= form.govuk_radio_button(:appropriate_body_id, radio_option.id, label: { text: radio_option.name }) do %>
+            <% if radio_option.id == @form.class::TEACHING_SCHOOL_HUB_ID %>
+              <%=
+                form.govuk_collection_select(
+                  :teaching_school_hub_id,
+                  @form.teaching_school_hubs,
+                  :id,
+                  :name,
+                  label: { text: "Teaching School Hub", class: "bold-label" },
```
frankieroberto, line 28: "Teaching School Hub" should be lowercase. We can also use the 'size' option here for the css class.
```suggestion
                  label: { text: "Teaching school hub", size: "s" },
```

frankieroberto (changes requested):
Tested in the review app – works well!  👏 

The appropriate body row doesn't appear for DIY cohorts, even though SITs can set (and change) this – can we add it for this scenario too?

A few minor front end amends. Thanks!

app/forms/admin/schools/cohorts/appropriate_bodies/update_form.rb
Lockyy: This is a fallback validation that ensures a user can't modify page data to send an ID that they shouldn't be able to set.

app/forms/admin/schools/cohorts/appropriate_bodies/update_form.rb
Lockyy: I've updated to be more specific

app/components/admin/schools/cohort_component.html.erb
Lockyy: Updated

app/views/admin/schools/cohorts/appropriate_bodies/edit.html.erb
Lockyy: Updated

app/views/admin/schools/cohorts/appropriate_bodies/edit.html.erb
Lockyy: Updated

app/forms/admin/schools/cohorts/appropriate_bodies/update_form.rb
```diff
@@ -0,0 +1,72 @@
...
+          TEACHING_SCHOOL_HUB_ID = "teaching_school_hub"
+
+          attr_accessor :appropriate_body_id, :teaching_school_hub_id, :school_cohort
+
+          validates :appropriate_body_id, presence: { message: I18n.t("errors.appropriate_body.blank") }
+          validates :teaching_school_hub_id, presence: { message: I18n.t("errors.teaching_school_hub.blank") }, if: :teaching_school_hub_selected?
+          validate :validate_selected_appropriate_body
+
+          def save!
+            school_cohort.update!(appropriate_body_id: selected_appropriate_body_id)
```
ltello, line 21: We have this SO to set it: Induction::SetSchoolCohortAppropriateBody

Lockyy: I looked that one over and that'd update all of the existing induction records, the screenshots in the ticket indicated that this form shouldn't do that and should only apply for future participants.

frankieroberto (approved):
Thanks!

PR #4248, opened 2023-11-09
Export SIT-school associations, schedules and nomination emails to an…

ltello:
…alytics

### Context

- Ticket:


leoapost (approved):
LGTM

PR #4255, opened 2023-11-10
Remove Logit, Grafrana and Analytics info from README

peteryates:
### Context

These could potentially confuse a new joiner so get rid.

* Logging is now accessible within Azure
* DfE are working on a replacement for the old Grafana setup
* Analytics is now handled via dfe-analytics/BigQuery, the old analytics database will be removed soon


cwrw (approved):
the dream is for logit to comeback soon :)

PR #4258, opened 2023-11-13
[CST-2094] Enable SITs to access profiles of all their school mentors

leoapost:
### Context

- Ticket: CST-2094

Currently, SITs can't see profiles of mentors who are only mentoring at the school (a.k.a mentors in the school's mentor pool).

We want to enable SITs to see such profiles, but without altering them.

- Update the ECFPolicy to enable SITs to see mentor profiles that are mentoring in their schools
- Update the ParticipantsController to get the mentor's induction record

### Guidance to review
1. Find to a school with mentors in their school mentor pool
2. Impersonate the SIT of the school
3. Visit the school's "Manage mentors and ECTs" page
4. Click on the name of the mentor
5. You should be able to see their profile, but not change any info

### DIFF
--- a/app/policies/participant_profile/ecf_policy.rb
+++ b/app/policies/participant_profile/ecf_policy.rb
@@ -2,7 +2,7 @@
 
 class ParticipantProfile::ECFPolicy < ParticipantProfilePolicy
   def show?
-    admin? || (user.induction_coordinator? && same_school?)
+    admin? || (user.induction_coordinator? && (same_school? || mentoring_at_school?))
   end
 
   alias_method :edit_mentor?, :show?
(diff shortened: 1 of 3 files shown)


app/policies/participant_profile/ecf_policy.rb
```diff
@@ -47,8 +47,20 @@ def withdraw_record?
...
+    record.mentor? && sit_school_ids.intersect?(mentor_school_ids)
+  end
+
+  def sit_school_ids
+    user.induction_coordinator_profile.schools.select(:id).map(&:id)
+  end
+
+  def mentor_school_ids
+    record.school_mentors.select(:school_id).map(&:school_id)
+  end
```
tonyheadford, line 65: How does this work if the SIT is at multiple schools and the mentor is at a different school to the one being viewed?

leoapost: @tonyheadford the SIT will only see the mentor listed in the "Manage mentors and ECTs" page, when they are viewing School B.

tonyheadford (approved):

PR #4262, opened 2023-11-13
Add oneoff service to migrate declarations between statements

ethax-ross:
[Jira-2592](https://dfedigital.atlassian.net/browse/CPDLP-2592)

### Context

We occasionally need to migrate declarations between different finance statements.

When doing this, we expect both statements to be output statements and after migrating the from statement should be updated to `output_fee: false`.

We only want to allow migration of NPQ declarations between NPQ statements and only when they are of the same cohort. We need to ensure there is a to/from statement for each provider.

- Add oneoff service to migrate declarations between statements

### Guidance to review

It may be that in the future we want to allow an output fee of `false` for the to/from statements. I've hard-coded an output-fee of `true` for now and I figure we can expand the interface if we meet that use case in the future.

Dry run output of the declaration migration in the ticket of Dec 23 -> Feb 24 is in the Jira ticket.

### DIFF
--- a/app/services/oneoffs/npq/migrate_declarations_between_statements.rb
+++ b/app/services/oneoffs/npq/migrate_declarations_between_statements.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+require "has_recordable_information"
+
+module Oneoffs::NPQ
+  class MigrateDeclarationsBetweenStatements
+    class StatementMismatchError < StandardError; end
+
+    include HasRecordableInformation
+
+    def initialize(cohort:, from_statement_name:, to_statement_name:)
+      @cohort = cohort
+      @from_statement_name = from_statement_name
+      @to_statement_name = to_statement_name
+    end
+
+    def migrate(dry_run: true)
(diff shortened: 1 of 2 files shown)


app/services/oneoffs/npq/migrate_declarations_between_statements.rb
```diff
@@ -0,0 +1,92 @@
...
+    def migrate_declarations_between_statements!
+      each_statements_by_provider do |provider, from_statement, to_statement|
+        record_declarations_info(provider, from_statement.participant_declarations)
+        migrate_line_items!(from_statement, to_statement)
+        from_statement.update!(output_fee: false)
+      end
+    end
+
+    def migrate_line_items!(from_statement, to_statement)
+      from_statement.statement_line_items.update_all(statement_id: to_statement.id)
```
cwrw, line 44: could we use `update` instead to trigger callback?

cwrw (approved):
very nicely done! just the one comment otherwise 👍

cwrw (approved):

PR #4267, opened 2023-11-15
Update appropriate body status translations

ethax-ross:
[Jira-2613](https://dfedigital.atlassian.net/browse/CPDLP-2613)

### Context

We have a new set of copy for some of the record states in the appropriate body status tag.

- Update AB status tag localisations

Also removes unused code from view component and clean up the specs.

### Guidance to review

A side-effect of this is that the status filter labels will have changed.

I've updated the review app so the example AB has participants for testing.


leandroalemao (approved):
lgtm 👍 😄

cwrw (approved):

PR #4271, opened 2023-11-17
[CPDLP-2609] Ensure that updated_at value of unfunded mentor is touched when an unfunded mentor gets linked to an ECT

leandroalemao:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2609

Ensure that updated_at value of unfunded mentor is touched when an unfunded mentor gets linked to an ECT.

### DIFF
--- a/app/models/participant_profile/ecf.rb
+++ b/app/models/participant_profile/ecf.rb
@@ -146,7 +146,10 @@ private
   def sync_status_with_induction_record
     induction_record = induction_records.latest
     induction_record&.update!(induction_status: status) if saved_change_to_status?
-    induction_record&.update!(mentor_profile:) if saved_change_to_mentor_profile_id?
+    if saved_change_to_mentor_profile_id?
+      induction_record&.update!(mentor_profile:)
+      mentor_profile&.user&.touch
+    end
   end
 end
 
--- a/app/models/participant_profile/ecf.rb
+++ b/app/models/participant_profile/ecf.rb
@@ -146,10 +146,7 @@ private
   def sync_status_with_induction_record
     induction_record = induction_records.latest
     induction_record&.update!(induction_status: status) if saved_change_to_status?
-    if saved_change_to_mentor_profile_id?
-      induction_record&.update!(mentor_profile:)
-      mentor_profile&.user&.touch
-    end
+    induction_record&.update!(mentor_profile:) if saved_change_to_mentor_profile_id?
   end
 end
 
--- a/app/services/induction/change_mentor.rb
+++ b/app/services/induction/change_mentor.rb
@@ -7,6 +7,7 @@ class Induction::ChangeMentor < BaseService
                                             changes: { mentor_profile: })
 
       induction_record.participant_profile.update!(mentor_profile:)
+      mentor_profile&.user&.touch
     end
   end
 
--- a/docs/source/release-notes.html.md
+++ b/docs/source/release-notes.html.md
@@ -7,6 +7,14 @@ weight: 8
 
 If you have any questions or comments about these notes, please contact DfE via Slack or email.
 
+## 20 November 2023
+
+We've fixed a bug that meant some providers were having issues finding unfunded mentor IDs in GET participants.
+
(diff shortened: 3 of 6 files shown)


app/models/participant_profile/ecf.rb
```diff
@@ -146,7 +146,10 @@ def update_analytics
   def sync_status_with_induction_record
     induction_record = induction_records.latest
     induction_record&.update!(induction_status: status) if saved_change_to_status?
-    induction_record&.update!(mentor_profile:) if saved_change_to_mentor_profile_id?
+    if saved_change_to_mentor_profile_id?
```
ethax-ross, line 149: Would it be better to encapsulate these in the `InductionRecord`/`ParticipantProfile` models and touch the user through lifecycle callbacks to cover other use cases? doing it this way means if the `induction_record.mentor_profile` or `participant_profile.mentor_profile` is updated elsewhere it won't touch the user?

app/services/induction/change_mentor.rb
```diff
@@ -7,6 +7,7 @@ def call
                                             changes: { mentor_profile: })
 
       induction_record.participant_profile.update!(mentor_profile:)
+      mentor_profile&.user&.touch
```
ethax-ross, line 10: Would it be better to encapsulate this in the `InductionRecord` model as above and touch the user through lifecycle callbacks to cover other use cases? doing it this way means if the `induction_record.mentor_profile` is updated elsewhere it won't touch the user?

cwrw: We usually read the `mentor_profile_id` from induction records rather than the profile or always should be, although checking the repository seems like a mix due to old code/habits. So ideally we'll need to add callbacks to both IRs and profiles. Adding a callback on induction records will become too noisy as we always create new induction records with mentor profile id populated (see https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/induction/change_mentor.rb#L6) rather than update in place, which could accidentally catch any new induction record created for other reasons.

I think adding touching the user in explicit cases when changing the mentor as it was in [app/services/induction/change_mentor.rb](https://github.com/DFE-Digital/early-careers-framework/pull/4271/files#diff-1fb68b60719d16f750323df92b961b0d719f270fc50a15bba53297bd8f6abfee) would be preferable to avoid any unexpected behaviour. For instance we already have some strange syncs here which appear cyclical: https://github.com/DFE-Digital/early-careers-framework/blob/main/app/models/participant_profile/ecf.rb#L149 and will catch us off guard. Happy to chat after standup tomorrow to go through the touch model scenarios

cwrw:
We usually read the `mentor_profile_id` from induction records rather than the profile or always should be, although checking the repository seems like a mix due to old code/habits. So ideally we'll need to add callbacks to both IRs and profiles. Adding a callback on induction records will become too noisy as we always create new induction records with mentor profile id populated (see https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/induction/change_mentor.rb#L6) rather than update in place, which could accidentally catch any new induction record created for other reasons.

I think adding touching the user in explicit cases when changing the mentor as it was in [app/services/induction/change_mentor.rb](https://github.com/DFE-Digital/early-careers-framework/pull/4271/files#diff-1fb68b60719d16f750323df92b961b0d719f270fc50a15bba53297bd8f6abfee) would be preferable to avoid any unexpected behaviour. For instance we already have some strange syncs here which appear cyclical: https://github.com/DFE-Digital/early-careers-framework/blob/main/app/models/participant_profile/ecf.rb#L149 and will catch us off guard. Happy to chat after standup tomorrow to go through the touch model scenarios

cwrw (approved):
🙌 Thanks @leandroalemao 

@ethax-ross after Leandro chatted to the schools team, they either use the UI which uses this service OR the induction records for updates in the console, which is tough to add callbacks to as a new record is created rather than updated in place. We think this might be the safest bet, although there are caveats let's chat on Monday after standup with @leandroalemao and Tom to highlight incase

PR #4272, opened 2023-11-20
Send rate limited requests to BigQuery

ethax-ross:
[Jira-2596](https://dfedigital.atlassian.net/browse/CPDLP-2596)

### Context

We want to ensure web requests that are rate limited by Rack::Attack still get sent to BigQuery as a web request.

- Send rate limited requests to BigQuery

Web requests are sent to BigQuery via a concern in the ApplicationController. If Rack intercepts the request it won't reach the controller, so we need to manually send web requests that are rate limited to BigQuery.

### Guidance to review

You can check 429s come through correctly in BigQuery:

```
SELECT * FROM `ecf-bq.ecf_events_review.events` WHERE response_status = "429" AND TIMESTAMP_TRUNC(occurred_at, DAY) = TIMESTAMP("2023-11-22") LIMIT 1000
```

You can generate more using something like `wrk`:

```
wrk -t5 -c5 -d5s https://cpd-ecf-review-4272-web.test.teacherservices.cloud
```

I tested by generating a bunch of fake traffic and Sidekiq had no problem staying on top of the jobs, I don't think there would be any issues shipping this to prod.

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -52,8 +52,27 @@ class Rack::Attack
 end
 
 ActiveSupport::Notifications.subscribe("throttle.rack_attack") do |_name, _start, _finish, request_id, payload|
-  ip = payload[:request].remote_ip
-  path = payload[:request].fullpath
+  request = ActionDispatch::Request.new(payload[:request].env)
+  response = ActionDispatch::Response.new(429)
+
+  ip = request.remote_ip
+  path = request.fullpath
+  user = request.env["warden"]&.user
 
   Rails.logger.warn("[rack-attack] Throttled request #{request_id} from #{ip} to '#{path}'")
+
+  # Web requests are sent to BigQuery via a concern in the ApplicationController.
+  # If Rack intercepts the request it won't reach the controller, so we need
+  # to manually send web requests that are rate limited to BigQuery.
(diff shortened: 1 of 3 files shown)


config/initializers/rack_attack.rb
```diff
@@ -52,8 +52,25 @@ def remote_ip
...
+  # to manually send web requests that are rate limited to BigQuery.
+  rate_limit_event = DfE::Analytics::Event.new
+    .with_type(:web_request)
+    .with_request_details(request)
+    .with_response_details(response)
+    .with_request_uuid(request.uuid)
+
+  rate_limit_event.with_user(user) if user.respond_to?(:id)
+
+  DfE::Analytics::SendEvents.do([rate_limit_event])
```
cwrw, line 75: could we only send this if `DfE::Analytics.enabled?`

cwrw (approved):
looks good, past the minor comment, can we test this with API requests from both lead provider and NPQ to make sure all is well just incase?

ethax-ross:
Screenshot 2023-11-23 at 10 30 38

@cwrw I've beefed up the testing and manually checked it rate limits NPQ reg tokens

cwrw (approved):
Thanks @ethax-ross 🙌

PR #4277, opened 2023-11-21
Add link to save statements as PDF

ethax-ross:
[Jira-2499](https://dfedigital.atlassian.net/browse/CPDLP-2499)

### Context

We want to let lead providers download a PDF copy of their financial statements. To avoid the complexity of generating PDFs locally we are instead going to leverage the browser print -> save as PDF functionality.

- Add link to save statements as PDF

When a user clicks the 'Save as PDF' link we perform some minor page formatting via Javascript and update the page title as this is used as the filename when saving. We also apply print-specific CSS to hide certain elements (such as links). After the user has finished saving the PDF we restore the page to its previous state (although we don't remember the detail open/close state; we just assume they were closed initially).


app/views/finance/npq/statements/show.html.erb
```diff
@@ -147,6 +147,11 @@
               <% end %>
             </div>
             <div class="govuk-grid-column-one-half govuk-!-text-align-right">
+              <p class="govuk-body-s">
+                <%=
+                  print_link("Save as PDF", filename: "#{@npq_lead_provider.name} #{@statement.name} Statement (#{@statement.cohort.start_year} Cohort)")
```
cwrw, line 152: could we add the type here as well in the name i.e. "NPQ", same with ECF?

app/assets/stylesheets/print.scss
```diff
@@ -0,0 +1,24 @@
+@media print {
+  // header
```
cwrw, line 2: maybe we can avoid global rules here and only apply them in the statement pages, as worried someone else wants to print something on another page fully and we mess with that here? let me know your thoughts

cwrw (commented):
fantastic 🙌 just some minor comments

app/assets/stylesheets/print.scss
ethax-ross: I talked to Mark and Shahad about this and they agree that would be the ideal approach. Its a bit tricky as some of the CSS selectors are global (header/footer) so I've split it out into a new top-level CSS file that gets compiled and included manually by views that want to use it.

leandroalemao (approved):
👍 😄 lgtm

PR #4281, opened 2023-11-22
[CPDLP-2647] Add nightly production database backup

cwrw:
### Context
We would like to keep a daily backup of the production database in case of errors.

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2647

- Add new workflow which runs nightly at 3:00 am which backs up the production database, and uploads it to the azure storage account
- Add an alert if the job fails. A new secret has been added to the key vault which we can use in the action
- The values have been hardcoded as this is only for production
- The backup is a duplicate of an existing backup restore to the snapshot but to avoid having to upload artifacts to Github (the db dump) when splitting actions, I've gone with duplication

### Guidance to review
- The slack alert has been tested and reports correctly in cpd-dev-alerts
- The uploaded db backup can be found in our storage container database backup

### DIFF
--- a/.github/actions/backup-and-upload-database/action.yml
+++ b/.github/actions/backup-and-upload-database/action.yml
@@ -1,0 +1,70 @@
+name: Backup and Upload production DB
+description: Backup production DB and upload to azure
+
+inputs:
+  azure-credentials:
+    description: Azure credentials
+    required: true
+
+runs:
+  using: composite
+
+  steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Login
+      uses: azure/login@v1
+      with:
+        creds: ${{ inputs.azure-credentials }}
+
+    - name: Set AKS credentials
+      shell: bash
+      run: az aks get-credentials -g s189p01-tsc-pd-rg -n s189p01-tsc-production-aks
+
+    - name: Install kubectl
+      uses: azure/setup-kubectl@v3
+
+    - name: Install konduit
+      shell: bash
+      run: make install-konduit
+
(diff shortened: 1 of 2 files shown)


.github/actions/backup-and-upload-database/action.yml
```diff
@@ -0,0 +1,70 @@
...
+
+    - name: Upload backup
+      shell: bash
+      run: |
+        az config set extension.use_dynamic_install=yes_without_prompt
+        az config set core.only_show_errors=true
+        az storage azcopy blob upload \
+          --container database-backup \
+          --source backup-production.sql.gz \
+          --destination $(date +"%F-%H").sql.gz
```
ethax-ross, line 52: It looks like this will create a new file for each backup; do we need to worry about deleting them, or is there a retention period setup on the blob storage?

ethax-ross (approved):
Looks great 👍

leandroalemao (approved):
Hi @cwrw It'd be nice to see if we can also restore one of the backups to check if the backup is working in the case of we need to restore it one day.

.github/actions/backup-and-upload-database/action.yml
saliceti: Indeed backups are deleted after 7 days by default: https://github.com/DFE-Digital/terraform-modules/blob/main/aks/postgres/resources.tf#L133

saliceti (approved):

cwrw:
@leandroalemao good shout, tested restoring the new backup and all works as expected

PR #4286, opened 2023-11-22
[CPDLP-2589] Spike: What are options for solving/managing ECF participants with induction records in multiple cohorts (stop flip flopping)

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2589

* Solution returns full list of latest InductionRecord and then we filter out cohorts (as before we'd filter cohort while getting latest induction records). This should correctly return participants for the filter cohort.

### DIFF
--- a/app/services/api/v3/ecf/participants_query.rb
+++ b/app/services/api/v3/ecf/participants_query.rb
@@ -62,14 +62,9 @@ module Api
         end
 
         def latest_induction_records_join
-          # InductionRecord
-          #   .select("induction_records.id AS latest_id")
-          #   .joins("JOIN (#{super.to_sql}) AS latest_induction_records_join ON latest_induction_records_join.latest_id = induction_records.id")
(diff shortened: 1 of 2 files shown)


app/services/api/v3/ecf/participants_query.rb
```diff
@@ -62,7 +62,9 @@ def from_participant_id
         end
 
         def latest_induction_records_join
-          super
+          InductionRecord
+            .select("induction_records.id AS latest_id")
+            .joins("JOIN (#{super.to_sql}) AS latest_induction_records_join ON latest_induction_records_join.latest_id = induction_records.id")
             .joins(:schedule)
```
cwrw, line 68: Could this not be included in `participants_for_pagination` instead and we can leave `latest_induction_records_join` as it is without cohort, as we're doing the latest induction record join twice?

leandroalemao (approved):

PR #4287, opened 2023-11-23
Account for minimum band A being set to 0

cwrw:
### Context
In cohort 2023 all band A declarations are appearing one over the max value, see https://manage-training-for-early-career-teachers.education.gov.uk/finance/ecf/payment_breakdowns/c3bc3cee-a636-42d6-8324-c033a6c38d31/statements/084aa7ad-2545-48a9-9388-a2c963855755 as an example where it should be 2000 rather than 2001 in band A

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2654

In legacy bandings, the minimum value for band A was set to `nil`, which meant in the output calculator it defaulted to 1. Since we introduced services to accept csvs and create contracts/bandings we set the band A min to 0. This pushed started declarations in band A above the max value, for instance band A max is set to 2000, it would appear in the dashboard as 2001.

Ensure that we account for both `nil` and 0 values in calculations to avoid wrong banding for declarations by setting both 1 otherwise the band min value

### Guidance to review
Tested on snapshot

### DIFF
--- a/app/services/finance/ecf/output_calculator.rb
+++ b/app/services/finance/ecf/output_calculator.rb
@@ -147,7 +147,11 @@ module Finance
         pot_size = previous_fill_level_for_declaration_type(declaration_type)
 
         bands.zip(:a..:z).map do |band, letter|
-          band_capacity = band.max - (band.min || 1) + 1
+          # minimum band should always be 1 or more, otherwise band a will go over
+          # its max limit
+          band_min = band.min.to_i.zero? ? 1 : band.min
+
+          band_capacity = band.max - band_min + 1
 
           fill_level = [pot_size, band_capacity].min
 
@@ -157,7 +161,7 @@ module Finance
 
           {
             band: letter,
-            min: band.min || 1,
+            min: band_min,
             max: band.max,
             key_name => fill_level,
           }
--- a/spec/services/finance/ecf/output_calculator_spec.rb
+++ b/spec/services/finance/ecf/output_calculator_spec.rb
@@ -228,6 +228,54 @@ RSpec.describe Finance::ECF::OutputCalculator do
         ]
         expect(first_statement_calc.banding_breakdown.map { |e| e.slice(*relevant_started_keys) }).to eql(expected)
       end
+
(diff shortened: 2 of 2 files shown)


app/services/finance/ecf/output_calculator.rb
```diff
@@ -154,10 +154,13 @@ def previous_banding_for_declaration_type(declaration_type)
           pot_size -= fill_level
 
           key_name = "previous_#{declaration_type.underscore}_count".to_sym
+          # minimum band should always be 1 or more, otherwise band a will go over
+          # its max limit
+          band_min = band.min == 0 || band.min.nil? ? 1 : band.min
```
mooktakim, line 159: This is a complicated mental logic.
Maybe something like:
```
band_min = (band.min.to_i == 0 ? 1 : band.min)
```

spec/services/finance/ecf/output_calculator_spec.rb
```diff
@@ -228,6 +228,54 @@
         ]
         expect(first_statement_calc.banding_breakdown.map { |e| e.slice(*relevant_started_keys) }).to eql(expected)
       end
+
+      context "when band a min is set to zero" do
+        before do
```
mooktakim, line 233: Should we add one more with nil?

mooktakim (changes requested):

ethax-ross (approved):
Looks good to me 👍 agree with Mook's comment around being more explicit or maybe ensuring there is one there with `min: nil`, so could do something like this which should error out if one with `nil` is not found:

```
 contract.participant_bands.find_by(min: nil).update(min: 0)
```

leandroalemao (commented):
Should we do the same logic in this [line](https://github.com/DFE-Digital/early-careers-framework/blob/6875e7f6d47ede0e7b718b76382f616ba6ff41a9/app/services/finance/ecf/output_calculator.rb#L150) as well?

leandroalemao (approved):
lgtm 👍 😄

mooktakim (approved):

ethax-ross (approved):

PR #4289, opened 2023-11-23
[CST-2126]`/pages/core-materials-info` redirects to `https://support-for-early-…

lazydevorg:
…career-teachers.education.gov.uk/`

### Context

- Ticket: CST-2126


frankieroberto (approved):
Nice! Do we know if this page is linked to internally within the service at all? If so, can we update those to link direct and avoid the redirect?

PR #4291, opened 2023-11-23
Fix missing `call` in Associations preloader

cwrw:
### Context

When querying declaration endpoint locally I noticed that we had many n+1s, although we were preloading the associations.
 
- Ticket: n/a

Turns out we need to add call here as it was missed in the syntax migration

### Guidance to review
Test on snapshot with 3000 per page 0_0


app/services/api/v3/participant_declarations_query.rb
```diff
@@ -52,7 +52,7 @@ def participant_declarations_from(paginated_join)
         ActiveRecord::Associations::Preloader.new(
```
ethax-ross, line 52: Is it worth testing theses more specific eager loads? we don't for others, but we could with something like:

```
expect(npq_declarations.map { |d| d.association(:outcomes) }).to all(be_loaded)
```

ethax-ross (approved):
Nice catch 👍

app/services/api/v3/participant_declarations_query.rb
cwrw: added, TIL you could test that

ethax-ross (approved):

PR #4297, opened 2023-11-27
[CST-2152] Fix mentor mentoring status

tonyheadford:
### Context

When a mentor has a mentee that has left (`induction_status = leaving` and `end_date=a_date_in_the_past`) they should not have the "mentoring" status tag and should appear as not mentoring to the school.

Add a check for the `end_date` when the mentee induction_status is `leaving` in the SQL query

### DIFF
--- a/app/services/determine_training_record_state_legacy.rb
+++ b/app/services/determine_training_record_state_legacy.rb
@@ -243,7 +243,7 @@ private
       FROM "induction_records"
       WHERE
           "induction_records"."mentor_profile_id" = '#{participant_profile_id}'
-          AND ("induction_records"."induction_status" = 'active' OR "induction_records"."induction_status" = 'leaving')
+          AND ("induction_records"."induction_status" = 'active' OR ("induction_records"."induction_status" = 'leaving' AND "induction_records"."end_date" > CURRENT_TIMESTAMP))
       GROUP BY
           "induction_records"."mentor_profile_id",
           "induction_records"."participant_profile_id"
--- a/spec/services/determine_training_record_state_legacy_spec.rb
+++ b/spec/services/determine_training_record_state_legacy_spec.rb
@@ -1137,6 +1137,30 @@ RSpec.describe DetermineTrainingRecordStateLegacy do
                          "not_yet_mentoring"
       end
 
+      context "and their mentee is leaving" do
+        let!(:participant_profile) { scenarios.mentor_on_fip_leaving_mentee.participant_profile }
+
+        include_examples "determines states as",
+                         "valid",
+                         "eligible_for_mentor_training",
+                         "eligible_for_mentor_funding",
+                         "active_mentoring",
+                         "registered_for_fip_training",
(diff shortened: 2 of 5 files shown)


app/services/determine_training_record_state_legacy.rb
```diff
@@ -243,7 +243,7 @@ def mentee_counts
       FROM "induction_records"
       WHERE
           "induction_records"."mentor_profile_id" = '#{participant_profile_id}'
-          AND ("induction_records"."induction_status" = 'active' OR "induction_records"."induction_status" = 'leaving')
+          AND ("induction_records"."induction_status" = 'active' OR ("induction_records"."induction_status" = 'leaving' AND "induction_records"."end_date" > CURRENT_TIMESTAMP))
```
ethax-ross, line 246: Is it worth adding a test cases for this?

We also have this logic in a few other places now as this is the legacy record state calculator that is in the process of being replaced. 

The new one calculates the mentoring state in two of places; the first is a concern that calculates the `current_mentees_count` on the initial query in the controller (which is an optimisation):

https://github.com/DFE-Digital/early-careers-framework/blob/0b65b163904c75a28dd8b4cede6ed345df3da94d/app/services/api/concerns/training_record_state_optimizable.rb#L22

The second place is in the`TrainingRecordState` model which serves as a fallback in case the concern has not been used:

https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/training_record_state.rb#L554

ethax-ross (commented):
Nice catch! I think this is going to be an issue in other places of the new record state code; happy to help with those changes if that would be useful.

app/services/determine_training_record_state_legacy.rb
tonyheadford: Thanks @ethax-ross this came via a support ticket where a school is complaining that their mentor isn't mentoring but appears as if they are in their dashboard so I was working back from the view and wasn't aware this logic was elsewhere.

spec/services/determine_training_record_state_legacy_spec.rb
```diff
@@ -1137,6 +1137,30 @@
                          "not_yet_mentoring"
       end
 
+      context "and their mentee is leaving" do
```
ethax-ross, line 1140: Its a pain because we have the logic duplicated just now, but it may be worth duplicating these specs here as well:

https://github.com/DFE-Digital/early-careers-framework/blob/4ceecee858ba566983ea3405ca160454d8874c2c/spec/support/shared_examples/optimised_training_record_state_query_support.rb#L16

ethax-ross (approved):
Looks great, just a couple of minor suggestions

spec/services/determine_training_record_state_legacy_spec.rb
tonyheadford: Yes good idea, I've added those

ethax-ross (approved):
LGTM 👏

PR #4306, opened 2023-11-29
Fix DQT api url on staging

ltello:
### Context

Fix DQT api url to be `https://preprod.teacher-qualifications-api.education.gov.uk` instead of `https://preprod-teacher-qualifications-api.education.gov.uk`

- Ticket:


peteryates (approved):
Thanks for catching this!

PR #4307, opened 2023-11-30
Remove unused css loading packages

frankieroberto:
Since #3016 the compiling of CSS is now done with [cssbundling-rails](https://github.com/rails/cssbundling-rails), so all of these `npm` packages aren't used at all.

I think.


peteryates (approved):
Nice 🧹

PR #4310, opened 2023-11-30
[CST-2091] Add support query form system

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CST-2091

- Adds framework for support query forms and sending queries to Zendesk
- Adds "change lead provider for participant profile" support form

### DIFF
--- a/app/components/support_form_component.html.erb
+++ b/app/components/support_form_component.html.erb
@@ -66,7 +66,7 @@
             :message,
             rows: 5,
             width: "three-quarters",
-            label: { text: "Your Message", tag: "h1", size: "s" },
+            label: { text: "Your message", size: "s" },
           ) %>
 
       <%= f.govuk_submit "Confirm" %>
--- a/app/components/support_form_component.html.erb
+++ b/app/components/support_form_component.html.erb
@@ -69,7 +69,7 @@
             label: { text: "Your message", size: "s" },
           ) %>
 
-      <%= f.govuk_submit "Confirm" %>
+      <%= f.govuk_submit "Send message to support" %>
     <% end %>
   </div>
 </div>
--- a/app/components/support_form_component.html.erb
+++ b/app/components/support_form_component.html.erb
@@ -3,61 +3,64 @@
     <span class="govuk-caption-l"><%= school.name if school.present? %></span>
     <h1 class="govuk-heading-l">Contact support</h1>
 
-    <p class="govuk-body">
-      <%= t("support_form.introduction.#{subject}", participant_name: participant_profile_full_name) %>
-    </p>
-
-    <p class="govuk-body">We’ll automatically include these details for you: </p>
-
-    <dl class="govuk-summary-list">
-      <div class="govuk-summary-list__row">
-        <dt class="govuk-summary-list__key">
-          Your name
-        </dt>
-        <dd class="govuk-summary-list__value">
-          <%= current_user.full_name %>
-        </dd>
-      </div>
-
-      <div class="govuk-summary-list__row">
-        <dt class="govuk-summary-list__key">
-          Your email address
-        </dt>
-        <dd class="govuk-summary-list__value">
-          <%= current_user.email %>
-        </dd>
-      </div>
-
-      <% if school.present? %>
-        <div class="govuk-summary-list__row">
-          <dt class="govuk-summary-list__key">
-            School
-          </dt>
-          <dd class="govuk-summary-list__value">
-            <%= school.name %> <br>
-
-            <span class="govuk-caption-m">URN <%= school.urn %></span>
-          </dd>
-        </div>
-      <% end %>
-
-      <% if participant_profile.present? %>
-        <div class="govuk-summary-list__row">
-          <dt class="govuk-summary-list__key">
-            Teacher
-          </dt>
-          <dd class="govuk-summary-list__value">
-            <%= participant_profile_full_name %>
-          </dd>
-        </div>
-      <% end %>
-    </dl>
-
-    <p class="govuk-body">
-      <%= t("support_form.instructions.#{subject}_html", participant_name: participant_profile_full_name) %>
-    </p>
 
     <%= form_for @form, url: support_path, method: :post do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <p class="govuk-body">
+        <%= t("support_form.introduction.#{subject}", participant_name: participant_profile_full_name) %>
+      </p>
+
+      <p class="govuk-body">We’ll automatically include these details for you: </p>
+
+      <dl class="govuk-summary-list">
+        <div class="govuk-summary-list__row">
+          <dt class="govuk-summary-list__key">
+            Your name
+          </dt>
+          <dd class="govuk-summary-list__value">
+            <%= current_user.full_name %>
+          </dd>
+        </div>
+
+        <div class="govuk-summary-list__row">
+          <dt class="govuk-summary-list__key">
+            Your email address
+          </dt>
+          <dd class="govuk-summary-list__value">
+            <%= current_user.email %>
+          </dd>
+        </div>
+
+        <% if school.present? %>
+          <div class="govuk-summary-list__row">
+            <dt class="govuk-summary-list__key">
+              School
+            </dt>
+            <dd class="govuk-summary-list__value">
+              <%= school.name %> <br>
+
+              <span class="govuk-caption-m">URN <%= school.urn %></span>
+            </dd>
+          </div>
+        <% end %>
+
+        <% if participant_profile.present? %>
+          <div class="govuk-summary-list__row">
+            <dt class="govuk-summary-list__key">
+              Teacher
+            </dt>
+            <dd class="govuk-summary-list__value">
+              <%= participant_profile_full_name %>
+            </dd>
+          </div>
+        <% end %>
+      </dl>
+
+      <p class="govuk-body">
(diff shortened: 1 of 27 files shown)


config/environments/test.rb
```diff
@@ -56,6 +56,10 @@
   config.gias_api_user = "gias-user"
   config.gias_api_password = "gias-password"
 
+  config.zendesk_url = "https://www.example.com"
+  config.zendesk_username = "zendesk-username"
+  config.zendesk_token = "zendesk-token"
```
ltello, line 61: Wouldn't it be better to move these envars into the encrypted rails credentials files?
Similarly to what we do with the credentials to DQT api.

Lockyy: See chat in Slack

db/schema.rb
```diff
@@ -134,7 +134,7 @@
     t.index ["start_year"], name: "index_cohorts_on_start_year", unique: true
   end
 
-  create_table "cohorts_lead_providers", force: :cascade do |t|
+  create_table "cohorts_lead_providers", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
tonyheadford, line 137: It probably doesn't affect anything but we shouldn't have this change I think?

Lockyy: Oops, yes this shouldn't have been included.

config/environments/development.rb
```diff
@@ -31,6 +31,10 @@
   config.gias_api_user = ENV["GIAS_API_USER"]
   config.gias_api_password = ENV["GIAS_API_PASSWORD"]
 
+  config.zendesk_url = ENV.fetch("ZENDESK_URL", "https://teachercpdhelp.zendesk.com/api/v2")
```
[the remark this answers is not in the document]
Lockyy: There's no real firm reason why, just that it isn't actually a secret. I'll remove.

tonyheadford (approved):
Looks great, nice one

app/components/support_form_component.html.erb
```diff
@@ -0,0 +1,75 @@
...
+      <%= f.hidden_field :school_id %>
+
+      <%= f.govuk_text_area(
+            :message,
+            rows: 5,
+            width: "three-quarters",
+            label: { text: "Your Message", tag: "h1", size: "s" },
+          ) %>
+
+      <%= f.govuk_submit "Confirm" %>
```
frankieroberto, line 72: I think a more explicit label on the button here is needed.

```suggestion
      <%= f.govuk_submit "Send message to support" %>
```

app/components/support_form_component.html.erb
```diff
@@ -0,0 +1,75 @@
...
+    <span class="govuk-caption-l"><%= school.name if school.present? %></span>
+    <h1 class="govuk-heading-l">Contact support</h1>
+
+    <p class="govuk-body">
+      <%= t("support_form.introduction.#{subject}", participant_name: participant_profile_full_name) %>
+    </p>
+
+    <p class="govuk-body">We’ll automatically include these details for you: </p>
+
+    <dl class="govuk-summary-list">
```
frankieroberto, line 12: Any reason not to use the [`summary_list` ruby component](https://govuk-components.netlify.app/components/summary-list/) here?

app/controllers/support_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def new
+    @form = SupportForm.new(new_params)
+  end
+
+  def create
+    @form = SupportForm.new(create_params)
+
+    if @form.valid? && @form.save
+      flash[:success] = { title: "Success", content: "Your support request has been submitted, our support team will get back to you shortly." }
```
frankieroberto, line 14: Could we take them to a dedicated 'success' page, [like this one](https://manage-training-for-ects-5c9ed99fc0a8.herokuapp.com/early-career-teachers/NDY745/support-contacted), instead of redirecting back to the contact form with a flash message?

If that's too complicated, then could we redirect them to the ECT page they started from and show the flash message there, instead of on the contact form page?

app/forms/support_form.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+class SupportForm
+  include ActiveModel::Model
+
+  attr_accessor :message, :participant_profile_id, :school_id, :current_user
+  attr_reader :subject
+
+  validates :message, presence: { message: "Please provide details of your request" }
```
frankieroberto, line 9: We generally don’t use 'please' in error messages 😄 

```suggestion
  validates :message, presence: { message: "Enter your message" }
```

config/locales/en.yml
```diff
@@ -852,3 +852,22 @@ en:
...
           submit: "View"
+
+  support_form:
+    introduction:
+      unspecified: "If you have a question about the Early Career Framework (ECF) service, you can contact the DfE using this form."
+      change-participant-lead-provider: "Complete this form to request a change to %{participant_name}’s registered training provider."
+    instructions:
+      unspecified_html: "Enter your support query and any relevant information below"
+      change-participant-lead-provider_html: |
+        Tell us:
```
frankieroberto, line 863:
```suggestion
        <p class="govuk-body">Tell us:</p>
```

config/locales/en.yml
```diff
@@ -852,3 +852,22 @@ en:
       statements:
         ecf_statement_selector:
           submit: "View"
+
+  support_form:
+    introduction:
+      unspecified: "If you have a question about the Early Career Framework (ECF) service, you can contact the DfE using this form."
```
frankieroberto, line 858: Should we not use the name of the service here?

```suggestion
      unspecified: "If you have a question about the Manage training for early career teachers service, you can contact the DfE using this form."
```

config/locales/en.yml
```diff
@@ -852,3 +852,22 @@ en:
...
+      change-participant-lead-provider: "Complete this form to request a change to %{participant_name}’s registered training provider."
+    instructions:
+      unspecified_html: "Enter your support query and any relevant information below"
```
frankieroberto, line 861: Drop this line? Not sure it's adding anything...

```suggestion
```

app/components/support_form_component.html.erb
```diff
@@ -0,0 +1,75 @@
...
+          </dd>
+        </div>
+      <% end %>
+    </dl>
+
+    <p class="govuk-body">
+      <%= t("support_form.instructions.#{subject}_html", participant_name: participant_profile_full_name) %>
+    </p>
+
+    <%= form_for @form, url: support_path, method: :post do |f| %>
```
frankieroberto, line 60: The page is missing the `<%= f.govuk_error_summary %>` component which summarises the error at the top of the page in a big red box.

frankieroberto (changes requested):
Looking good! Tested in the review app. There's a few minor amends but overall it works well!

app/components/support_form_component.html.erb
```diff
@@ -0,0 +1,75 @@
...
+          <dd class="govuk-summary-list__value">
+            <%= participant_profile_full_name %>
+          </dd>
+        </div>
+      <% end %>
+    </dl>
+
+    <p class="govuk-body">
+      <%= t("support_form.instructions.#{subject}_html", participant_name: participant_profile_full_name) %>
+    </p>
```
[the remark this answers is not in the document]
Lockyy: They don't contain the paragraph wrapper though and moving that into the locales would mean replicating it every time un-necessarily.

config/locales/en.yml
Lockyy: Moving this inside here means duplicating it every time rather than just having it one in the template. Plus, that's templating rather than copy which doesn't really belong in the locale file.

app/components/support_form_component.html.erb
Lockyy: Not really, it works either way.

app/controllers/support_controller.rb
Lockyy: I can do this, though getting in the customisation related to what the ticket was for would add some complication. I'd suggest we do that in a separate ticket if necessary.

config/environments/development.rb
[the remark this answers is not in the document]
Lockyy: Oh yeah I forgot, it's for the specs.

app/views/support/new.html.erb
```diff
@@ -0,0 +1 @@
+<%= render SupportFormComponent.new(form:) %>
```
frankieroberto, line 1: Can we add a back link on this page? (It'll have to be dynamic based on which ECT, mentor other page you were looking at)

app/views/support/show.html.erb
```diff
@@ -0,0 +1,11 @@
+<div class="govuk-grid-row">
```
frankieroberto, line 1: Add page title.

```suggestion
<% content_for :title, "Your support request has been submitted" %>
<div class="govuk-grid-row">
```

app/views/support/new.html.erb
frankieroberto, line 1: Add page title

```suggestion
<% content_for :title, "Contact support" %>
<%= render SupportFormComponent.new(form:) %>
```

frankieroberto (changes requested):
Thanks for making the updates. Just spotted a couple of missing page titles and a back link, sorry!

frankieroberto (approved):
Thanks for addressing the suggestions! Looks good to me. 😄

app/views/support/new.html.erb
frankieroberto: We resolved to leave this for a follow-up PR.

PR #4311, opened 2023-12-01
Add rake task to update verified data for NPQ

pmanrubia:
Ticket: https://dfedigital.atlassian.net/browse/CPDNPQ-1516

### Description

This Rake task is designed for a single execution to update verified data for users and applications
based on data provided in a CSV file. The full file path of the CSV is passed as a parameter to the task.

The task reads each row from the CSV file, finds the corresponding application and its associated user,
and updates them with the verified data. The task is intended to be run once and can be safely removed
after its execution. The code is deliberately not DRY to ensure it is explicit and clear to the reader.

Usage:

```
$ rake oneoff:update_verified_data['/path/to/your/file.csv']
```

The task logs detailed information about the processing, including any warnings or errors encountered.
It ensures that failure to update one record does not halt the entire process, allowing all records
in the CSV file to be processed.

### DIFF
--- a/lib/tasks/oneoff/ecf_validations_01_12_2023.rake
+++ b/lib/tasks/oneoff/ecf_validations_01_12_2023.rake
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+# This Rake task is designed for a single execution to update verified data for users and applications
+# based on data provided in a CSV file. The full file path of the CSV is passed as a parameter to the task.
+#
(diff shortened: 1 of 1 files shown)


javier-npq (approved):
Script looks good code wise @pmanrubia 👍, I was wondering how do you plan to execute this? If we are running this for every application then the number of records in the csv are going to be huge as last i checked we had around 100k applications (could be more now).

PR #4312, opened 2023-12-01
Import NPQ reg models into ECF app

ethax-ross:
> ⚠️ **DO NOT MERGE**

Brings the NPQ reg models into the ECF app so that we can connect to/query both datasets from the same Rails console.

You can connect to both databases via Make commands/konduit, for example:

```
make install-konduit
make ci production konduit-snapshot-ecf
make ci production konduit-snapshot-npq
rails c

irb(main):003:0> NPQApplication.count
  NPQApplication Count (40.7ms)  SELECT COUNT(*) FROM "npq_applications"                                    
=> 128310                                             
irb(main):004:0> NPQRegistration::Application.count
  NPQRegistration::Application Count (48.0ms)  SELECT COUNT(*) FROM "applications"                          
=> 128934 
```

In order for the Make commands to work you need to tweak the default `bin/konduit.sh` to spit out the database URL after connecting (I'm looking into a better way of doing this):

```
run_psql() {
  echo "$DB_URL"
  ...
}
```

Adds service classes to reconcile NPQ reg users/applications with ECF and generate reports of 'orphaned' records (that exist in one service but not the other).

Things of note for when we inevitably have to update the NPQ reg models/schema:

- `pg-search` and `omniauth` references need to be removed from the NPQ reg models (as we don't want to pull these gems into ECF).
- Acronyms from NPQ reg are in uppercase and need to be made camel-case when bringing into ECF.
- Rails expects both databases to use `uuid` primary keys, even though we only pull `plpgsql` and `uuid-ossp` into the ECF schema. To get the NPQ reg schema to use serial IDs we have to use this workaround:

```
create_table "a_table", id: false, force: :cascade do |t|
  t.serial :id, primary_key: true
  ...
end
```


config/analytics_blocklist.yml
```diff
@@ -578,3 +578,177 @@
...
+  - ukprn
+  - address_1
+  - address_2
+  - address_3
+  - town
+  - county
+  - postcode
+  - postcode_without_spaces
+  - high_pupil_premium
+  :lead_providers:
```
ethax-ross, line 617: I think we will end up with some NPQ models in our BigQuery here; need to check if the model class is sent as part of the event so that we can differentiate as it doesn't look like dfe-analytics will handle two models with the same class name in different namespaces here (as its using the table name, which is the same).

config/database.yml
```diff
@@ -22,15 +22,23 @@ default: &default
     port: "<%= ENV['ANALYTICS_DB_PORT'] || Rails.application.credentials.ANALYTICS_DB_PORT %>"
     database: "<%= ENV['ANALYTICS_DB_DATABASE'] || Rails.application.credentials.ANALYTICS_DB_DATABASE %>"
     migrations_paths: db/analytics_migrate
-
+  npq_registration: &default_npq_registration
```
ethax-ross, line 25: Need to set this up for all environments before this can be merged.

ethax-ross:
Closing this PR as we're shifting it over to NPQ reg

PR #4314, opened 2023-12-01
Make teacher profile selection consistent

peteryates:
Inside the `#same_trn_user` we are selecting all teacher profiles that match a TRN but aren't specifying any explicit order, so there's no guarantee which `TeacherProfile` record is found and used.

This appears to have caused some problems where providers expect a user to have one ID but they have another, because the `#deduplicate_by_trn!` method picked the 'wrong' teacher profile, so the wrong user to attach the `ParticipantIdentity` to.

This change orders the TeacherProfile records by created_at, so it will consistently link to the oldest record.

### DIFF
--- a/app/models/teacher_profile.rb
+++ b/app/models/teacher_profile.rb
@@ -20,6 +20,8 @@ class TeacherProfile < ApplicationRecord
 
   self.filter_attributes += [:trn]
 
+  scope :oldest_first, -> { order(created_at: "asc") }
+
   def self.trn_matches(search_term)
     return none if search_term.blank?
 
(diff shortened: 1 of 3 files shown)


app/models/teacher_profile.rb
```diff
@@ -20,6 +20,8 @@ class TeacherProfile < ApplicationRecord
 
   self.filter_attributes += [:trn]
 
+  scope :order_by_created_at, -> { order(created_at: "asc") }
```
ethax-ross, line 23: I don't feel strongly about it, but it may be worth being explicit to the direction here; something like `oldest_first`?

ethax-ross (approved):
Good catch! makes sense, assuming using the oldest matching profile is what providers are expecting 👍

app/models/teacher_profile.rb
peteryates: Done, thanks :)

PR #4316, opened 2023-12-04
CST-2131: only migrate active induction records to a new programme

ltello:
### Context

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2131)

From the latest induction record of each participant, add a new induction record to the chain instead of from each induction record

### DIFF
--- a/app/services/induction/migrate_participants_to_new_programme.rb
+++ b/app/services/induction/migrate_participants_to_new_programme.rb
@@ -3,7 +3,7 @@
 class Induction::MigrateParticipantsToNewProgramme < BaseService
   def call
     ActiveRecord::Base.transaction do
-      from_programme.induction_records.each do |induction_record|
+      from_programme.active_induction_records.each do |induction_record|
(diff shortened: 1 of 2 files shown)


tonyheadford (commented):
Looks great, do you think we could add a spec to demonstrate that we're not moving non-active induction records?

tonyheadford:
> > Looks great, do you think we could add a spec to demonstrate that we're not moving non-active induction records?
> 
> I think the spec I have modified is already testing this:
> 
> * Before the change: From 2 records we have initially (none in changed status) we migrate them and we get double number of records (4) 2 of them in changed status.
> * After the change: From 3 records we have initially (only 1 in changed status) we migrate the participants and end up with 5 records (not double number of records) and only 3 of them in changed status instead of 4.

Yes, I just meant an explicit case something like "it does not move non-active induction records" just to make the behaviour clear?

tonyheadford (approved):

PR #4318, opened 2023-12-04
Cst2167 change provider guidance

Lloyd-K:
### Context

We plan to add a standalone guidance page that lets users know how to change training provider for one or more ECTs. This will initially sit on the service platform but may migrate to GOV.UK in the future.

The Jira ticket provides more information: https://dfedigital.atlassian.net/browse/CST-2167?atlOrigin=eyJpIjoiNTZmNTEwNzJiMzg5NGE2NWIxZWIxZjZjYzJiNGQyNTIiLCJwIjoiaiJ9


app/views/pages/change_ECF_training_provider.html.erb
```diff
@@ -0,0 +1,112 @@
+<% content_for :title, "Change ECF-based training provider for ECTs and their mentors" %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+    <h1 class="govuk-heading-l">Change ECF-based training provider for ECTs and their mentors</h1>
+
+<ul class="govuk-list">
+  <li>
+    [Change training provider for ECTs and mentors across 1 or 2 years](#change-training-provider-for-ects-and-mentors-across-1-or-2-years)
```
frankieroberto, line 10: Sadly we can't use markdown here, so these need to be converted to `<a href="" class="govuk-link"> ... </a>`.

app/views/pages/change_ECF_training_provider.html.erb
```diff
@@ -0,0 +1,112 @@
...
+
+<p>Next, your school’s registered induction tutor must report the change to DfE. We need this information to pay the correct funding to each training provider.</p>
+
+<p class="govuk-body">Email <%= govuk_link_to "continuing-professional-development@digital.education.gov.uk", "mailto:continuing-professional-development@digital.education.gov.uk" %> and provide:</p>
+<ul class="govuk-list govuk-list--bullet">
+  <li>your school’s name and unique reference number (URN)</li>
+  <li>the name of the new lead provider</li>
+  <li>the confirmed date the new lead provider will start training your ECTs (and their mentors if relevant)</li>
+  <li>the years you want to apply the change to (year 1, year 2, or both years)<li> 
+</ul>
```
frankieroberto, line 71: Given that we’ll (soon) have the support form to let schools contact us about this within the service, should we direct users there instead? 🤔

frankieroberto:
@Lloyd-K this can be reviewed in the review app here: https://cpd-ecf-review-4318-web.test.teacherservices.cloud/pages/change_ECF_training_provider

app/views/pages/change_ECF_training_provider.html.erb
Lloyd-K: I think we go live as is to support next week's comms, and then update once the forms are rolled out.

PR #4321, opened 2023-12-04
Create dedupe participant service class

leandroalemao:
### Context

- Ticket:


app/models/npq_application.rb
```diff
@@ -128,6 +128,12 @@ def change_logs
       .sort { |a, b| b.created_at <=> a.created_at }
   end
 
+  def save_and_dedupe_participant
+    result = save
+    NPQ::DedupeParticipant.new(npq_application: self, trn: teacher_reference_number).call
+    result
```
peteryates, line 134: Does this need to be wrapped within a transaction?

app/controllers/api/v1/npq_profiles_controller.rb
```diff
@@ -13,7 +13,7 @@ def show
       def create
         @npq_application = build_npq_application
 
-        if @npq_application.save
+        if @npq_application.save_and_dedupe_participant
```
peteryates, line 16:
```suggestion
        if @npq_application.save && @npq_application.dedupe_participant!
```

🤔

peteryates:
Looks like a good start, left some early comments that you're welcome to ignore

PR #4322, opened 2023-12-05
[CST-2110] Add code to remove dup completed records created by accident

leoapost:
### Context

- Ticket:  CST-2110

A bug created 300k dup induction records in September and we have been asked to delete them.

These records are dups of the "completed" record of participants with completed induction. 
Also, their attributes have the same values as the active completed record, except the timestamps and the induction_status which was changed from "completed" to "changed".

1. Add a delete service class that
    1. safely deletes an induction record. It will skip
        - the first induction record
        - the latest induction record
        - records that show significant changes in the participant's induction like a school transfer or a change in their training status or induction programme
    2. updates the `end_date` of the previous record to maintain the induction records sequence

2. Add a rake task that sets off a background job to find the potential dup records created by the bug and deletes them in batches

### DIFF
--- a/app/jobs/delete_completed_induction_records_job.rb
+++ b/app/jobs/delete_completed_induction_records_job.rb
@@ -5,8 +5,8 @@ class DeleteCompletedInductionRecordsJob < ApplicationJob
     bug_introduced_at = "2023-09-04"
     job_running_window = { from: "00:00:00", to: "11:00:00" }
 
-    deleted = []
-    failed = []
+    deleted_count = 0
+    failed_count = 0
 
     # Potential dup records created by the bug that was running from 4 Sept 2023, between midnight and 11am
     dup_induction_records = InductionRecord
@@ -24,13 +24,13 @@ class DeleteCompletedInductionRecordsJob < ApplicationJob
       Induction::DeleteDupRecord.new(induction_record:).call
 
       Rails.logger.info "### Removed #{induction_record.id} record from #{induction_record.participant_profile_id} participant"
-      deleted << induction_record.id
+      deleted_count += 1
     rescue StandardError
       Rails.logger.debug "### Failed to remove #{induction_record.id} record from #{induction_record.participant_profile_id} participant: #{e.inspect}"
-      failed << induction_record.id
+      failed_count += 1
     end
 
-    Rails.logger.info "### Records deleted: #{deleted.count}"
-    Rails.logger.info "### Records failed to delete: #{failed.count}"
+    Rails.logger.info "### Records deleted: #{deleted_count}"
+    Rails.logger.info "### Records failed to delete: #{failed_count}"
   end
 end
--- a/app/jobs/delete_completed_induction_records_job.rb
+++ b/app/jobs/delete_completed_induction_records_job.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+class DeleteCompletedInductionRecordsJob < ApplicationJob
+  def perform
+    bug_introduced_at = "2023-09-04"
+    job_running_window = { from: "00:00:00", to: "11:00:00" }
+
+    deleted = []
+    failed = []
+
+    # Potential dup records created by the bug that was running from 4 Sept 2023, between midnight and 11am
+    dup_induction_records = InductionRecord
+      .joins(:participant_profile)
+      .joins(:versions)
+      .where.not(participant_profile: { induction_completion_date: nil })
+      .where(induction_status: :changed, school_transfer: false)
+      .where("induction_records.created_at >= ?", bug_introduced_at)
+      .where("induction_records.created_at::time >= :from AND induction_records.created_at::time <= :to ", job_running_window)
+      .where("versions.object_changes ->> 'induction_status' ILIKE '[%completed%changed%]'")
+      .where("versions.whodunnit IS NULL")
+
+    dup_induction_records.in_batches(of: 500).each_record do |induction_record|
+      # Try to safely delete the record without breaking the induction records sequence
+      Induction::DeleteDupRecord.new(induction_record:).call
+
+      Rails.logger.info "### Removed #{induction_record.id} record from #{induction_record.participant_profile_id} participant"
+      deleted << induction_record.id
+    rescue StandardError
+      Rails.logger.debug "### Failed to remove #{induction_record.id} record from #{induction_record.participant_profile_id} participant: #{e.inspect}"
+      failed << induction_record.id
+    end
+
+    Rails.logger.info "### Records deleted: #{deleted.count}"
+    Rails.logger.info "### Records failed to delete: #{failed.count}"
+  end
+end
--- a/app/services/induction/delete_dup_record.rb
+++ b/app/services/induction/delete_dup_record.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
-class Induction::DeleteRecord < BaseService
+class Induction::DeleteDupRecord < BaseService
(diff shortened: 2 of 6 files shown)


app/services/induction/delete_record.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+  end
+
+private
+
+  attr_reader :induction_record
+
+  def initialize(induction_record:)
+    @induction_record = induction_record
+  end
```
peteryates, line 23: As discussed earlier it seems like entering at the participant profile level makes more sense to me, because `Induction::DeleteRecord` makes it sound like it's simply deleting this record but actually it's doing that **and** potentially amending the onnes around it.

peteryates: Should the purpose of the class be to 'fix' a _participant's_ induction record by manipulating their `InductionRecord` rows? Rather than going induction record by induction record and merging them together.

tonyheadford: @peteryates I think the intention was that something like you describe would be the tier above this. This service is like removing a record from a linked list (with a few extra checks thrown in) that could be reused in cases where some precision surgery was needed and would generally do the right thing.

app/services/induction/delete_record.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  def check_record_is_deletable!
+    raise DeleteInductionRecordRestrictionError, "Cannot delete record because it is active" if active?
+    raise DeleteInductionRecordRestrictionError, "Cannot delete record because it is not in the middle of the induction records history" unless middle_of_history?
+    raise DeleteInductionRecordRestrictionError, "Cannot delete record because the school transfer flag does not matches the previous record" if transfer_flag_changed?
+    raise DeleteInductionRecordRestrictionError, "Cannot delete record because the training status does not matches the previous record" if training_status_changed?
+    raise DeleteInductionRecordRestrictionError, "Cannot delete record because the mentor does not matches the previous record" if mentor_changed?
+  end
```
ltello, line 29: what about if the induction_programme changed?

ltello: or the schedule?

ltello: or the preferred_identity?

leoapost: @ltello the `check_record_is_deletable!` has just a few extra safeguards that we thought are the most important ones, but we don't expect any of the records we are going to delete to have any of those changes really.

The bug created dup of the completed records, and such records only set the `induction_status = completed` and `mentor_profile = nil`.

leoapost: You can see the query in the rake task is only picking records created within the time window the bug was running and were "complete" records historically that changed to "change".

```
bug_introduced_at = "2023-09-04"
job_running_window = { from: "00:00:00", to: "11:00:00" }

dup_induction_records = InductionRecord
      .joins(:participant_profile)
      .joins(:versions)
      .where.not(participant_profile: { induction_completion_date: nil })
      .where(induction_status: :changed, school_transfer: false)
      .where("induction_records.created_at >= ?", bug_introduced_at)
      .where("induction_records.created_at::time >= :from AND induction_records.created_at::time <= :to ", job_running_window)
      .where("versions.object_changes ->> 'induction_status' ILIKE '[%completed%changed%]'")
      .where("versions.whodunnit IS NULL")
```

ltello (approved):

app/services/induction/delete_record.rb
```diff
@@ -0,0 +1,85 @@
...
+  def previous_record
+    participant_induction_records[induction_record_index + 1] unless last_record?
+  end
+
+  def next_record
+    participant_induction_records[induction_record_index - 1] unless first_record?
+  end
+
+  def participant_induction_records
+    induction_record
```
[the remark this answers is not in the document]
leoapost: yeah, good catch

leoapost:
>yes, but this is a class on its own, used by the rake task below.
I would agree if those checks were inside the rake task but what if that class stays in the codebase and someone reuses it in the future?
I think it is risky if we are not deleting it after the clean up

@ltello I understand your concerns, so I've refactored the Delete class to include safeguards for the attributes you mentioned in your previous comments. It's now also called DeleteDupRecord which should make it clearer it's meant to be used to delete dup records only.

app/services/induction/delete_dup_record.rb
```diff
@@ -0,0 +1,82 @@
...
+private
+
+  attr_reader :induction_record
+
+  def initialize(induction_record:)
+    @induction_record = induction_record
+  end
+
+  def check_record_is_deletable!
+    raise DeleteInductionRecordRestrictionError, "Cannot delete record because it is active" if active?
```
ltello, line 26: I think exceptions are very slow. Can we try error messages instead to reduce that 1 day completion time?

app/jobs/delete_completed_induction_records_job.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      Rails.logger.info "### Removed #{induction_record.id} record from #{induction_record.participant_profile_id} participant"
+      deleted << induction_record.id
+    rescue StandardError
+      Rails.logger.debug "### Failed to remove #{induction_record.id} record from #{induction_record.participant_profile_id} participant: #{e.inspect}"
+      failed << induction_record.id
+    end
+
+    Rails.logger.info "### Records deleted: #{deleted.count}"
+    Rails.logger.info "### Records failed to delete: #{failed.count}"
```
ltello, line 34: Maybe deleted and failed arrays get too big and consume a lot of memory.
As we only use them to count the number of elements, maybe can it be converted to simple counters?

app/services/induction/delete_dup_record.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def participant_induction_records
+    @participant_induction_records ||= induction_record
+      .participant_profile
+      .induction_records
+      .order(start_date: :desc, created_at: :desc)
+  end
+
+  def induction_record_index
+    participant_induction_records.index(induction_record)
```
ltello, line 72: I guess this could also be memoized for performance purposes.

app/services/induction/delete_dup_record.rb
```diff
@@ -0,0 +1,82 @@
...
+  def previous_record
+    participant_induction_records[induction_record_index + 1] unless last_record?
+  end
+
+  def next_record
+    participant_induction_records[induction_record_index - 1] unless first_record?
+  end
+
+  def participant_induction_records
+    @participant_induction_records ||= induction_record
```
ltello, line 65: If for every single IR to delete we need to find their participant and all their IRs, this could explain why it is taking so long to finish execution.

Is it too late for a solution where you iterate in batches over the participant_profiles affected by this issue  rather than IR by IR? This would allow to delete at once all the IRs of a participant

app/services/induction/delete_dup_record.rb
leoapost: I ran the deletion in a 10k sample records and I got no exceptions raised. Maybe there will be some edge cases that will raise some of them, but it doesn't seem it will make any difference performance wise.

app/services/induction/delete_dup_record.rb
leoapost: I'm on the support rota, so not sure if I have time this week.

Wanna give it a try?

app/jobs/delete_completed_induction_records_job.rb
leoapost: Good catch. I had this to help me troubleshoot failed deletions. I've pushed a commit to change the counters to integers.

app/services/induction/delete_dup_record.rb
leoapost: By the way, we decided to delete the records one by one as it was the safest way to ensure the IR sequence will be maintained.

PR #4324, opened 2023-12-06
run database sanitisation script and upload to database backups

mattonfoot:
### Context

Seeding a performance database takes too long to get anywhere near production levels. if we sanitise the production data then we can restore the sanitised database to the performance environment for use in our performance tests as a base line data set.

after dumping the product database, restore it locally and run a sanitise.sql script, dump the new data to a santisised dump and upload alongside the actual data.

### DIFF
--- a/db/scripts/sanitise.sql
+++ b/db/scripts/sanitise.sql
@@ -1,0 +1,256 @@
+/* create temporary look up tables for foreign key identifiers (TRNs, URNs, UK PRNs) */
+
+SELECT
+    schools.urn as original_urn,
+    RIGHT(CONCAT('000000', CAST(rows.index AS VARCHAR(6))), 6) as urn,
+    RIGHT(CONCAT('00000000', CAST(rows.index AS VARCHAR(8))), 8) as uk_prn
+INTO schools_ref
+FROM schools
+    LEFT OUTER JOIN (
+        SELECT entry.id as id, row_number() over () as index
+        FROM schools as entry
+        ORDER BY entry.created_at
+    ) AS rows ON schools.id = rows.id;
+
+SELECT
+    teacher_profiles.trn as original_trn,
+    RIGHT(CONCAT('0000000', CAST(rows.index AS VARCHAR(7))), 7) as trn
+INTO teacher_profiles_ref
+FROM teacher_profiles
+    LEFT OUTER JOIN (
+        SELECT entry.id as id, row_number() over () as index
+        FROM teacher_profiles as entry
+        ORDER BY entry.created_at
+    ) AS rows ON teacher_profiles.id = rows.id;
+
+SELECT
+    local_authority_districts.code as original_code,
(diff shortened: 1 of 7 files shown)


db/scripts/sanitise.sql
peteryates, line 1: Let's hope this is never accidentally run on prod 🙈

db/scripts/sanitise.sql
```diff
@@ -0,0 +1,256 @@
...
+
+TRUNCATE TABLE npq_application_eligibility_imports;
+TRUNCATE TABLE npq_application_exports;
+
+/* need to rewrite these IDs */
+UPDATE npq_applications
+    SET school_ukprn=schools_ref.uk_prn,
+        school_urn=schools_ref.urn,
+        teacher_reference_number=teacher_profiles_ref.trn,
+        date_of_birth=NULL
```
peteryates, line 140: There's the `nino` in here we probably want to hide too.

peteryates (approved):
Looks good, other than the `nino` on `npq_applications` nothing else personal comes to mind.

performance/docker-compose.yml
```diff
@@ -45,6 +45,8 @@ services:
     image: postgres:11-alpine
```
saliceti, line 45: Can we upgrade to 14 as well?

mattonfoot: good spot, thanks. upgraded

db/scripts/sanitise.sql
mattonfoot: good spot, missed that one.

db/scripts/sanitise.sql
mattonfoot: hopefully enforcing "localhost" in the script will ensure that it never uses that for the sanitisation. But I made sure it backed up the prod db before it did that, and as no one would be able to login in anymore no additional changes would be made to the data. 

Hopefully that goes in part to mitigate towards it. Also I copied the pattern from teaching vacancies...

db/scripts/sanitise.sql
```diff
@@ -0,0 +1,257 @@
...
+        ukprn=schools_ref.uk_prn,
+        school_website=CONCAT('school-', id, '.school'),
+        secondary_contact_email=CONCAT('school-', id, '-primary@example.com'),
+        primary_contact_email=CONCAT('school-', id, '-secondary@example.com'),
+        administrative_district_name=local_authority_districts_ref.name,
+        administrative_district_code=local_authority_districts_ref.code,
+        slug=CONCAT(schools_ref.urn, '-', CONCAT('school-', id))
+    FROM schools_ref, local_authority_districts_ref
+    WHERE schools_ref.original_urn = schools.urn
+        AND local_authority_districts_ref.original_code = schools.administrative_district_code;
```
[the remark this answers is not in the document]
mattonfoot: yep, I could, I just copied Teaching Vacancies pattern, but that makes sense. I will also do it for the other GIAS entities like district.

PR #4327, opened 2023-12-07
Move analytics tables from blocklist to main config

peteryates:
We have a second database that contains old analytics data that's automatically updated when data is changed. This change should hook the analytics model save behaviour into dfe-analytics, so it's sent to BigQuery

In theory we can test this with something like:

```sql
SELECT
  entity_table_name,
  count(*)
FROM
  `ecf-bq.ecf_events_review.events`
WHERE
  entity_table_name in ('ecf_school_cohorts', 'ecf_appropriate_bodies', 'ecf_participants', 'ecf_partnerships', 'ecf_schools', 'ecf_school_cohorts')
GROUP BY
  entity_table_name;
```

PR #4330, opened 2023-12-08
Remove paas config and reorder the config file

peteryates:
We still had old `VCAP_APPLICATION` references in the database.yml file. These have been removed and the environments reordered so they match the flows more closely.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -31,31 +31,6 @@ development:
     <<: *default_analytics
     database: "early_careers_framework_analytics_development"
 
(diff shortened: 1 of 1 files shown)


ethax-ross (approved):
Looks good. Worth removing the other references to `VCAP_APPLICATION` and `VCAP_SERVICES` as part of this PR as well?

peteryates:
Will do it as a follow up, pushing this through to get #4327 in.

PR #4366, opened 2023-12-18
Re-add the analytics_schools table

peteryates:
I thought this was no longer used but it is, and it being missing has caused some errors in queues following the nightly data sync.


ltello (approved):

peteryates:
image

Problem solved 😬

PR #4373, opened 2023-12-20
Add NPQ registration DBs to ECF

ethax-ross:
> ⚠️ we may need to manually exclude the models from dfe-analytics

In order to migrate data from ECF to NPQ reg we need to connect the databases and run a migration process that:

- Reads ECF models
- Reads current (and new) NPQ reg models
- Writes fixed up/restructured data to current/new NPQ reg models

This PR adds an NPQ reg database to environments where there is no corresponding NPQ reg environment (so review, test , dev) and will run migrations. For other hosted environments (sandbox, staging, prod) we will not run any database tasks and instead connect to the corresponding DB in the NPQ reg container.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,9 @@ inherit_gem:
 
 AllCops:
   Exclude:
+    - app/models/migration/**/*
+    - db/npq_registration_schema.rb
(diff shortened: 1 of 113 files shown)


ethax-ross:
Closing as we're going to pull ECF models into NPQ reg instead

PR #4374, opened 2023-12-20
Add migration rails env

peteryates:
### Context

This environment will be used in the upcoming `migration` environment in Azure, and used to test the NPQ separation work.

- Add database config
- Add migration env config
- Add migration env to Gemfile, seeds and seed quantities

### DIFF
--- a/config/environments/migration.rb
+++ b/config/environments/migration.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+require Rails.root.join("config/environments/production")
+
+Rails.application.configure do
+  config.log_level = :debug
+  config.ssl_options = { redirect: { exclude: ->(request) { request.path.include?("/check") } } }
+end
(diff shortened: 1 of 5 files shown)


config/environments/migration.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+require Rails.root.join("config/environments/production")
+
+Rails.application.configure do
+  config.log_level = :debug
+  config.ssl_options = { redirect: { exclude: ->(request) { request.path.include?("/healthcheck") } } }
```
leoapost, line 7: Should the excluded path be `/check`? It seems that's the case in all the other envs.

peteryates: ah yeah, my bad. It's `/healthcheck` on NPQ! Thanks, will force push in a sec

leoapost (approved):

PR #4376, opened 2023-12-21
Add migration environment config for Azure

peteryates:
This environment, and the corresponding one in NPQ, will be used to repeatedly test the migration process.

It's a copy of `sandbox`.

The long name will be `migration` and the short name `mg`

The tfvars are copied from sandbox but we don't need domains/external hostnames. This is just for internal use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,6 @@ sandbox: production-cluster
 .PHONY: migration
 migration: production-cluster
 	$(eval include global_config/migration_aks.sh)
-	$(eval SPACE=early-careers-framework-migration)
 
 .PHONY: production
 production: production-cluster
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,11 @@ sandbox: production-cluster
 	$(eval include global_config/sandbox_aks.sh)
 	$(eval SPACE=early-careers-framework-sandbox)
 
+.PHONY: migration
+migration: production-cluster
+	$(eval include global_config/migration_aks.sh)
+	$(eval SPACE=early-careers-framework-migration)
+
 .PHONY: production
 production: production-cluster
 	$(eval include global_config/production_aks.sh)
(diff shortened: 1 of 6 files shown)


peteryates:
@saliceti this is our new env we're going to use to test our migration process. I haven't tried running it yet but would appreciate some eyes to spot any obvious errors before doing that. Thanks!

Makefile
```diff
@@ -27,6 +27,11 @@ sandbox: production-cluster
 	$(eval include global_config/sandbox_aks.sh)
 	$(eval SPACE=early-careers-framework-sandbox)
 
+.PHONY: migration
+migration: production-cluster
+	$(eval include global_config/migration_aks.sh)
+	$(eval SPACE=early-careers-framework-migration)
```
saliceti, line 33: Is this needed?

config/credentials/migration.yml.enc
```diff
@@ -0,0 +1 @@
+VheEW39+12tw/s5Md/RSstMteEt9mv+w+Ez7atRwY06woq8HiBSrQTt8gC0xziTwl12wEwUAUX9N6Ew1IZJIeI4GDoo=--PaiDlYsKLXjojRC4--sxJ5tDIDFDS5rrg1aW0VpQ==
```
saliceti, line 1: Can we use keyvault instead?

terraform/aks/workspace_variables/migration_aks.tfvars.json
```diff
@@ -0,0 +1,9 @@
+{
+  "postgres_flexible_server_sku": "GP_Standard_D2ds_v4",
```
saliceti, line 2: If it's not for production use, can we use a cheaper one? Like the default in https://github.com/DFE-Digital/terraform-modules/blob/main/aks/postgres/variables.tf#L85

config/credentials/migration.yml.enc
peteryates: We do for all new things, but some of the older variables in production are in the credentials file so the app won't boot unless it's present.

For this env it's there but empty.

terraform/aks/workspace_variables/migration_aks.tfvars.json
peteryates: Yes, will amend.

saliceti (approved):

Makefile
peteryates: Also see #4379

peteryates:
This env is now up at https://cpd-ecf-migration-web.teacherservices.cloud/ 🎉

PR #4397, opened 2024-01-02
[CST-2192] Fix the Appropriate Body radio buttons

leoapost:
### Context

- Ticket: CST-2192

ABs are absent from the radio buttons on the Admin -> School -> Cohorts -> Change Appropriate Body page. 

This is due to our approach of generating the radio buttons by retrieving them from the database using hardcoded names extracted from a web page beyond our control, which could be subject to changes at any time.

- Remove code that selects AB by using hardcoded names
- Add a flag in the AB table to dynamically control their inclusion or exclusion from the radio buttons
- Manually flag in the DB the ABs we need to list in that page. Applies in all envs (local, dev, staging, prod etc)

### Guidance to review
1. Flag a few ABs in the DB as listed
2. Confirm they are listed as options in the Admin -> School -> Cohorts -> Change Appropriate Body page

### DIFF
--- a/app/forms/admin/schools/cohorts/appropriate_bodies/update_form.rb
+++ b/app/forms/admin/schools/cohorts/appropriate_bodies/update_form.rb
@@ -23,10 +23,9 @@ module Admin
 
           def radio_options
             @radio_options ||= [
-              istip_appropriate_body,
-              nta_appropriate_body,
+              appropriate_bodies.where(listed: true),
               OpenStruct.new(id: TEACHING_SCHOOL_HUB_ID, name: "A teaching school hub"),
-            ].compact
+            ].flatten.compact
           end
 
           def selected_appropriate_body_id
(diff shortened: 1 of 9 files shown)


ltello (approved):

frankieroberto:
@leoapost Tried this on the review app but the national appropriate bodies aren't listed:

Screenshot 2024-01-12 at 12 29 55

Does it need a data migration to add the flag to the existing records?

leoapost:
> @leoapost Tried this on the review app but the national appropriate bodies aren't listed:
> 

> Does it need a data migration to add the flag to the existing records?

@frankieroberto NTA specifically is flagged to be disabled after year 2023, so it won't appear regardless if the `listed` flag is true. 

Any other BAs we need to display must be flagged as `listed` manually.

lazydevorg (approved):

PR #4412, opened 2024-01-08
[CST-2166] Add support form link to the footer

Lockyy:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CST-2166

- When logged in user will now see a link to the support form page in the footer instead of seeing the contact us email address.

### DIFF
--- a/app/views/layouts/_support_links.html.erb
+++ b/app/views/layouts/_support_links.html.erb
@@ -1,7 +1,11 @@
 <h2 class="govuk-heading-m">Support and guidance</h2>
 <p class="govuk-body-s">
-  If you have a question, or you’ve had a problem using this service, please contact us at
-  <%= mail_to Rails.application.config.support_email, Rails.application.config.support_email, class: "govuk-footer__link govuk-link--no-visited-state" %>
+  <% if current_user.present? %>
+    If you have a question, or you’ve had a problem using this service, please <%= govuk_link_to("click here", support_path) %> to contact us.
+  <% else %>
+    If you have a question, or you’ve had a problem using this service, please contact us at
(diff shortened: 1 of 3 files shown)


leoapost (approved):

app/views/layouts/_support_links.html.erb
```diff
@@ -1,7 +1,11 @@
 <h2 class="govuk-heading-m">Support and guidance</h2>
 <p class="govuk-body-s">
-  If you have a question, or you’ve had a problem using this service, please contact us at
-  <%= mail_to Rails.application.config.support_email, Rails.application.config.support_email, class: "govuk-footer__link govuk-link--no-visited-state" %>
+  <% if current_user.present? %>
+    If you have a question, or you’ve had a problem using this service, please <%= govuk_link_to("click here", support_path) %> to contact us.
```
frankieroberto, line 4:
```suggestion
    If you have a question, or you’ve had a problem using this service,  <%= govuk_link_to("contact support", support_path) %>.
```

app/views/layouts/_support_links.html.erb
```diff
@@ -1,7 +1,11 @@
...
+  <% else %>
+    If you have a question, or you’ve had a problem using this service, please contact us at
```
frankieroberto, line 6:
```suggestion
    If you have a question, or you’ve had a problem using this service, contact us at
```

frankieroberto (changes requested):
Minor amend to the link - we avoid using words like "click here" as if a screenreader user hears that phrase out-of-context (there's a mode where a list of links are read out) it won't make sense.

Think we should also drop "please", as we usually avoid such filler words.

frankieroberto (approved):

PR #4413, opened 2024-01-08
Cleanup participant declarations for NPQ users

slawosz:

Participant declarations have 2 different ways to determine its user (which in general is denormalization we don't need). 
User of `ParticipantDeclaration` might be accessed in 2 ways:
`participant_declaration.user` (via `user_id` field in participant declaration)
or via
`participant_declaration.participant_profile.participant_identity.user` - this is correct way.
During deduping some of them become incorrect. This code fixes any denormalization.

### DIFF
--- a/app/services/oneoffs/npq/cleanup_participant_declaration.rb
+++ b/app/services/oneoffs/npq/cleanup_participant_declaration.rb
@@ -5,7 +5,9 @@ module Oneoffs::NPQ
     def migrate
       ActiveRecord::Base.transaction do
         incorrect_participant_declarations.in_batches.each_record do |declaration|
-          declaration.update(user_id: declaration.participant_profile.user.id)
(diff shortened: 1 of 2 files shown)


app/services/oneoffs/npq/cleanup_participant_declaration.rb
```diff
@@ -0,0 +1,19 @@
...
+        incorrect_participant_declarations.in_batches.each_record do |declaration|
+          declaration.update(user_id: declaration.participant_profile.user.id)
+        end
+      end
+    end
+
+  private
+
+    def incorrect_participant_declarations
+      ParticipantDeclaration.joins(participant_profile: :participant_identity).where("participant_declarations.user_id != participant_identities.user_id")
```
cwrw, line 16: could we scope to NPQ here please?

cwrw (approved):

PR #4414, opened 2024-01-08
[CST-2110] Delete the dup completed induction records by participant

leoapost:
### Context

- Ticket: CST-2110

This changes the code a bit to process the records by participant and add a limit to run the deletion in a smaller scale.


tonyheadford (approved):
LGTM

PR #4415, opened 2024-01-08
[CPDLP-2567] Is there a bug around changing schedule where the participant is a non-partnership cohort via a relationship?

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2567

* Added changes to `ChangeSchedule` and tests

### DIFF
--- a/app/services/change_schedule.rb
+++ b/app/services/change_schedule.rb
@@ -95,7 +95,12 @@ private
   end
 
   def induction_programme
-    @induction_programme ||= target_school_cohort&.default_induction_programme
+    @induction_programme ||=
+      if school_partnership&.relationship
+        relevant_induction_record&.induction_programme
+      else
+        target_school_cohort&.default_induction_programme
+      end
   end
 
   def update_participant_profile_schedule_references!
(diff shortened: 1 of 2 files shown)


app/services/change_schedule.rb
```diff
@@ -221,16 +226,37 @@ def relevant_induction_record_has_different_schedule
...
   end
 
-  def school_default_partnership
+  def school_partnership
     return if school.blank?
 
-    school
-      .active_partnerships
-      .find_by(cohort:, relationship: false, lead_provider_id: cpd_lead_provider.lead_provider_id)
+    @school_partnership ||= (school_partnership_without_relationship || school_partnership_including_relationship)
```
cwrw, line 237:
```suggestion
    @school_partnership ||= school_partnership_without_relationship || school_partnership_including_relationship
```
merging a rubocop update that doesn't allow that

app/services/change_schedule.rb
```diff
@@ -221,16 +226,37 @@ def relevant_induction_record_has_different_schedule
...
+    # We are assuming there is only ever one partnership record
+    school.active_partnerships.find_by(
+      cohort:,
+      lead_provider_id: cpd_lead_provider.lead_provider_id,
+    )
+  end
+
+  def change_schedule_identifier_only?
+    cohort == schedule.cohort &&
+      schedule_identifier != schedule.schedule_identifier
```
cwrw, line 260: maybe we can instead use `new_schedule != schedule` as there is the `identifer_alias` query to also think about if we only rely on the identifier?

cwrw (commented):
looks great, some minor comments otherwise 👌

PR #4416, opened 2024-01-09
CST-2199: re-process eligible ects with an error in SyncDQTInductionS…

ltello:
…tartDateError

### Context

Participants that failed sitting them in their permanent cohort (Monday eligibility checks process) but became “Eligible” as part of that process, have an entry in the SyncDQTInductionStartDateError table and won’t be picked anymore in the next Monday eligibility checks.

Therefore, it would be good to put a periodic process in place to retry sitting them in their right cohort because things might change at any moment in time (schools setting a school cohort, a participant reactivation,…) perhaps making the cohort change succeed this time.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2199)

### DIFF
--- a/app/jobs/check_eligible_ects_that_failed_permanent_cohort_setup_job.rb
+++ b/app/jobs/check_eligible_ects_that_failed_permanent_cohort_setup_job.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class CheckEligibleEctsThatFailedPermanentCohortSetupJob < ApplicationJob
+  def perform
+    ect_errors.find_each do |error|
+      ect = error.participant_profile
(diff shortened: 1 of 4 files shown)


config/credentials/development.yml.enc
tonyheadford, line 1: Should this be here?

config/database.yml-cp
tonyheadford, line 1: Should this be here too?

tonyheadford (commented):
Think theres a couple of files that shouldn't be in the commit but fine otherwise

config/database.yml-cp
ltello: Opps!. Thanks!

tonyheadford (approved):

PR #4424, opened 2024-01-10
Make tags lowercase

frankieroberto:
With the update to GOV.UK Frontend version5, the [Tag component](https://design-system.service.gov.uk/components/tag/) is now recommended to use capitalised phrases rather than UPPERCASE.

### Before

Screenshot 2024-01-10 at 16 52 32

### After

Screenshot 2024-01-10 at 16 52 18


peteryates (approved):

tonyheadford (approved):
Great spot!

PR #4429, opened 2024-01-14
CST-2209: mentor-first ect-mentor journey

ltello:
### Context

Schools are required to pair early career teachers with a mentor. This makes the mentor eligible for mentor training, as well as helping to make sure that all early career teachers have a mentor.

Previously, this had to be done from the profile page for the early career teacher. However we have some evidence that some schools would try to do this the other way around, navigating to a mentor and then expecting to pick which teacher should be assigned to them.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2209)

We’ve designed a flow allowing schools to approach the pairing of early career teachers and mentors in either direction.

### Guidance to review

- choose a mentor mentoring no ECT yet. 
- go to their profile
- click the link at the bottom to add their first ECT. 
- follow the journey

Same for a mentor mentoring some ECTs already.


app/views/schools/participants/show.html.erb
```diff
@@ -178,21 +178,55 @@
...
-                            visually_hidden_text: "mentor",
-                            href: school_participant_edit_mentor_path(participant_id: ect.participant_profile_id,
-                                                                      from_mentor: @profile.id)) %>
+    <% if @profile.mentor? %>
+      <h3 class="govuk-heading-m">Mentoring</h3>
+
+      <%= govuk_summary_list(classes: ["govuk-!-margin-bottom-7"], actions: false) do |slist| %>
+        <% slist.with_row do |row| %>
+          <% row.with_key do %>
+            Currently<br>mentoring
```
TomBr0wn, line 187: Can we please remove the `<br>` so the text renders as a single line?
```suggestion
            Currently mentoring
```

app/views/schools/participants/show.html.erb
```diff
@@ -178,21 +178,55 @@
...
+          <% end %>
+          <% row.with_value do %>
+            <% if @ects.present? %>
+              <%= govuk_summary_list(classes: ["govuk-!-margin-bottom-7"], borders: false) do |list| %>
```
TomBr0wn, line 191: Can we take a look at the nested summary list here or was it built this way for a specific reason? In the prototype the content inside the main summary list is structured with a regular unordered list

It means that both lists have a margin-bottom-7 on, which throws out the spacing on the bottom of that row.

TomBr0wn (commented):
Thanks for sharing the changes. I've left a few comments.
If you have any questions, I'm happy to schedule a quick call, or ping a message in Slack.

app/views/schools/participants/ect_added_confirmation.html.erb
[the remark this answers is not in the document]
ltello: Same for the back link and button.

app/views/schools/participants/show.html.erb
ltello: So it is fine if I just remove the css class from the nested summary-list?

app/views/schools/participants/show.html.erb
ltello: I added it because that is what seems to be in the picture attached to the ticket by @frankieroberto. 
We can confirm with him we want to display this in a single line.

frankieroberto: @ltello hiya - let's remove it please. It was probably only split across two lines in the screenshot because the browser window was a bit narrower.

app/views/schools/participants/ect_added_confirmation.html.erb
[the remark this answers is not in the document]
frankieroberto: @ltello you're right, that odd confirmation page is used on the reverse flow (assigning a mentor to an ECT). Not sure how I've not spotted that before... 🤦 

I think we should remove the confirmation from both flows, particularly as the back link behaves in an unexpected way – it takes you back to a page where you can assign another ECT to that mentor, or sometimes to a page where there are no radio buttons left to select at all... 😬 

Is it a big job to remove the confirmation page from both flows as part of this ticket? In the case of adding an ECT to a mentor, it would redirect to the mentor profile page instead, and in the case of assigning a mentor to an ECT, it'd redirect to the ECT page.

Also happy for it to be a follow-up ticket, but in that case let's try and get it out straight after this one so we don't forget it?

app/views/schools/participants/show.html.erb
frankieroberto: That'd look ok visually, but accessibility-wise maybe it should be a `<ul>` rather than a nested `<dl>` here? 🤔 

In the prototype I used a `<ul>` but then a bunch of `<div>`s with grid column classes within each `<li>` to get the 'Change' link to appear right-aligned: https://github.com/DFE-Digital/manage-training-for-early-career-teachers-prototype/blob/main/app/views/mentors/show.html#L95-L114

Possibly either is fine though, I'm not an accessibility expert on this.

app/views/schools/participants/ect_added_confirmation.html.erb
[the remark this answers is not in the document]
ltello: Ok, I can try and do these changes as part of this ticket.

app/views/schools/participants/show.html.erb
ltello: From a dev perspective it is much easier and cleaner to use the nested summary-list, I would say.

I have removed the extra margin so if we can leave it like that it would be great, otherwise I would have to add explicitly that buch of divs and css classes you used in your prototype making it all kind of dirty and messy.

But if you guys strongly think we need to change it to an unsorted list instead, i will do it.

ltello: Nested summary list replaced with <ul>

TomBr0wn (commented):
Tested flow in review app and it all looks good. Extra screen removed and nested summary list removed. Thanks for the changes

TomBr0wn (approved):

PR #4430, opened 2024-01-15
[CPDLP-2684] Amend `ParticipantIdentityResolver` to fetch accepted NPQ applications only

cwrw:
### Context

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2684

We can have multiple NPQ applications to the same course with the same provider, but only one is accepted. If there are multiple identities linked to the user, the resolver will sometimes pick the wrong identity if there is another profile attached to that identity. 

To ensure we account for multiple identities and multiple NPQ applications to the same course, ensure the application is accepted.

### Guidance to review
I will test this on the snapshot with the failing id in the ticket to ensure it works for the provider

### DIFF
--- a/app/services/participant_identity_resolver.rb
+++ b/app/services/participant_identity_resolver.rb
@@ -31,7 +31,7 @@ class ParticipantIdentityResolver
         ParticipantIdentity
          .joins(:npq_participant_profiles, npq_applications: [:npq_course, { npq_lead_provider: :cpd_lead_provider }])
          .where(npq_courses: { identifier: course_identifier })
-         .where(npq_applications: { npq_lead_providers: { cpd_lead_provider: } })
+         .where(npq_applications: { lead_provider_approval_status: "accepted", npq_lead_providers: { cpd_lead_provider: } })
          .where(user_id: participant_id)
          .first
       end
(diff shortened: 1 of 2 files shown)


ethax-ross (approved):
Makes sense 👍

app/services/participant_identity_resolver.rb
```diff
@@ -31,7 +31,7 @@ def call
         ParticipantIdentity
          .joins(:npq_participant_profiles, npq_applications: [:npq_course, { npq_lead_provider: :cpd_lead_provider }])
          .where(npq_courses: { identifier: course_identifier })
-         .where(npq_applications: { npq_lead_providers: { cpd_lead_provider: } })
+         .where(npq_applications: { lead_provider_approval_status: "accepted", npq_lead_providers: { cpd_lead_provider: } })
```
ethax-ross, line 34: Just had an afterthought; would this cause the identity resolver to fail for users that don't have an accepted NPQ application? is that a valid state?

leandroalemao (commented):
lgtm 👍 😄

PR #4437, opened 2024-01-15
Swap participant id changes to show real behaviour

cwrw:
### Context

To avoid confusion swap from/to participant ids to be closer to reality and inline with the current participant ID

- Ticket: n/a

- Update NPQ/ECF guidance
- Update API spec v3


mooktakim (approved):
👍

PR #4446, opened 2024-01-17
[CST-2147] mentor completions spike

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-2147)

Store mentor completion date. Adds a service to take the completion date from the ERO mentor list or from a completed declaration if present.


app/services/record_declaration.rb
```diff
@@ -258,4 +260,14 @@ def create_participant_outcome!
       raise Api::Errors::InvalidParticipantOutcomeError, I18n.t(:cannot_create_completed_declaration)
     end
   end
+
+  def set_mentor_completion
+    Mentors::CompleteTraining.call(mentor_profile: participant_profile) if mentor_completion_event?
+  end
+
+  def mentor_completion_event?
+    participant_profile.mentor? &&
```
cwrw, line 269: I think we might need more validations in place here, especially ones mentioned around preventing providers from making declarations dated after a mentor has completed training. Happy to discuss

db/schema.rb
```diff
@@ -933,6 +933,8 @@
     t.string "notes"
     t.date "induction_start_date"
     t.date "induction_completion_date"
+    t.date "mentor_completion_date"
+    t.string "mentor_completion_reason"
```
cwrw, line 937: from previous conversations I was wondering if it would be preferable to store multiple reasons instead of one, as that might apply to participants?

app/services/mentors/check_training_completion.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def completed_declaration
+      @completed_declaration ||= find_valid_declaration
+    end
+
+    def find_valid_declaration
+      mentor_profile
+        .participant_declarations
+        .for_declaration("completed")
+        .where(state: %w[submitted eligible payable paid])
```
cwrw, line 46: maybe the scope `billable_or_changeable` encapsulates this?

app/services/mentors/check_training_completion.rb
```diff
@@ -0,0 +1,51 @@
...
+    def ero_mentor?
+      trn = mentor_profile.teacher_profile.trn
+      trn.present? && ECFIneligibleParticipant.exists?(trn:)
+    end
+
+    def completed_declaration
+      @completed_declaration ||= find_valid_declaration
+    end
+
+    def find_valid_declaration
```
cwrw, line 42: maybe we should raise a validation error if there are multiple completed declarations, incase due to bad data?

app/services/mentors/check_training_completion.rb
```diff
@@ -0,0 +1,51 @@
...
+  private
+
+    attr_reader :mentor_profile
+
+    def initialize(mentor_profile:)
+      @mentor_profile = mentor_profile
+    end
+
+    def set_completion_values!
+      completion_date, completion_reason = if ero_mentor?
```
cwrw, line 22: I think a risk here is we override reasons/dates if something changes (even though we are favouring ERO), maybe we should guard against that if the date/reason already exists on a profile to avoid changes? Also a question elsewhere, do we want to support multiple reasons from the get go?

app/services/void_participant_declaration.rb
```diff
@@ -14,6 +14,8 @@ def call
 
     NPQ::VoidParticipantOutcome.new(participant_declaration).call
```
cwrw, line 15: I think we need some form of transaction here?

app/services/void_participant_declaration.rb
```diff
@@ -45,4 +47,16 @@ def make_voided
...
+  def participant_profile
+    @participant_profile ||= participant_declaration.participant_profile
+  end
+
+  def check_mentor_completion
+    Mentors::CheckTrainingCompletion.call(mentor_profile: participant_profile) if mentor_completion_event?
+  end
+
+  def mentor_completion_event?
+    participant_profile.mentor? && participant_declaration.declaration_type == "completed"
```
cwrw, line 60: might be worth moving this logic to the service as it's repeated? we can return early there

cwrw (commented):
Thanks @tonyheadford, left a few questions/suggestions. Im still waiting on an answer regarding schedule validations, but hoping we don't need to add any. If we do there will be extra work in record declarations

db/schema.rb
tonyheadford: I don't believe we need to support multiple current reasons, NAthan has confirmed there will only ever be one and will let us know the order of precedent when we support the other reasons. I will confirm again whether there is any need to support historic reasons if there is a change but I am pretty sure that wasn't necessary.

app/services/mentors/check_training_completion.rb
tonyheadford: These are always going to work in a given order of precedence so should be fine. Only the ERO one is not triggered by an event and currently you should not be both ERO and declared, and if you were, ERO should trump the declaration.  I don't think this is a problem.  I want to ensure we re-evaluate the completion state when we receive a void/clawback because when the other reasons are introduced we should ensure we're not clearing the completion when some other reason might apply further down the precedence list perhaps.

app/services/mentors/check_training_completion.rb
tonyheadford: There aren't any duplicate completed declarations that I can see in production at this moment in time and that really should be handled outside of setting the mentor completion values as this is checked _after_ the declaration is added (I suspect this is already being checked?)

app/services/void_participant_declaration.rb
tonyheadford: Moved the check into the other transaction based actions

app/services/void_participant_declaration.rb
tonyheadford: Yes, I've moved this now into a service that performs the check and calls the service only when necessary.

db/schema.rb
tonyheadford: Confirmation from policy colleagues that papertrail is enough, we don't need to support any other history for the dates/reason

app/services/record_declaration.rb
tonyheadford: Did we resolve this one? I have a feeling we didn't think this was an issue for this?

app/services/mentors/check_training_completion.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def completed_declaration
+      @completed_declaration ||= find_valid_declaration
+    end
+
+    def find_valid_declaration
+      mentor_profile
+        .participant_declarations
+        .for_declaration("completed")
+        .billable_or_changeable
```
cwrw, line 46: Just double confirming we want to intentionally include submitted declarations (which are not fundable meaning eligible)?

app/services/void_participant_declaration.rb
```diff
@@ -24,6 +24,8 @@ def make_awaiting_clawback
     if clawback.errors.any?
       raise Api::Errors::InvalidTransitionError, clawback.errors.full_messages.join(", ")
     end
+
+    check_mentor_completion
```
cwrw, line 28: should we move this to the clawback declaration service to ensure it's within that transaction?

app/services/mentors/check_training_completion.rb
tonyheadford: Yes, it was explicitly mentioned by policy that "submitted" would be the trigger for new ones.

cwrw (approved):
thanks @tonyheadford!

PR #4458, opened 2024-01-19
[CST-2164] Admin console view of 'most frequent error messages'

lazydevorg:
### Context

- Ticket: CST-2164

The solution has been taken from Apply and ported to ECF.
The `ValidationError` model has been created and the method `track_validation_error` has been added in the `ApplicationController` to be used by all controllers when validating a form or model.

### DIFF
--- a/app/controllers/admin/performance/validation_errors_controller.rb
+++ b/app/controllers/admin/performance/validation_errors_controller.rb
@@ -6,13 +6,13 @@ module Admin::Performance
     skip_after_action :verify_policy_scoped
 
     def index
-      @validation_errors = ValidationError.list_of_distinct_errors_with_count
+      @pagy, @validation_errors = pagy_array(ValidationError.list_of_distinct_errors_with_count, items: 50)
     end
 
     def show
       @form_object = params[:form]
       @attribute = params[:attribute]
-      @validation_errors = ValidationError.search({ form_object: @form_object, attribute: @attribute })
+      @pagy, @validation_errors = pagy(ValidationError.search({ form_object: @form_object, attribute: @attribute }), items: 50)
     end
   end
 end
--- a/app/views/admin/performance/validation_errors/index.html.erb
+++ b/app/views/admin/performance/validation_errors/index.html.erb
@@ -25,9 +25,8 @@
           <td class="govuk-table__cell  govuk-table__cell--numeric"><%= count %></td>
         </tr>
       <% end %>
-
       </tbody>
     </table>
-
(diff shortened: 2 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -83,4 +83,15 @@ def set_sentry_user
 
     Sentry.set_user(id: current_user.id)
   end
+
+  def track_validation_error(form)
+    ValidationError.create!(
+      form_object: form.class.name,
+      request_path: request.path,
+      user: current_user,
+      details: form.errors.messages.to_h { |field, messages| [field, { messages:, value: (form.public_send(field) if form.respond_to?(field)) }] },
```
Lockyy, line 92: Probably not a huge consideration but is there any risk of this containing PII?

app/controllers/finance/banding_tracker/provider_choices_controller.rb
```diff
@@ -11,7 +11,11 @@ def new
       def create
         choose_provider_form.attributes = choose_provider_params
 
-        return redirect_to finance_banding_tracker_provider_path(id: choose_provider_form.id) if choose_provider_form.valid?
+        if choose_provider_form.valid?
+          return redirect_to finance_banding_tracker_provider_path(id: choose_provider_form.id)
```
Lockyy, line 15: It was like this already but it might be a bit neater to move the `@providers =... render :new` into the else so we don't have to return. It's match more closely with the other controller actions that way.

Low priority of course.

app/models/validation_error.rb
```diff
@@ -0,0 +1,30 @@
...
+        details["messages"].map do |message|
+          [e.form_object, attribute, message]
+        end
+      end
+    end
+
+    distinct_errors
+      .tally
+      .sort_by { |_a, b| b }
+      .reverse
```
Lockyy, line 19: This seems like it might get slow very quickly, with us producing quite a lot of errors across all schools. Do we have any benchmarking from Apply for if we need to look for a quicker alternative or cache the results?

Lockyy (approved):
All looks good to go in, minor concern on speed of the data collation and PII though nothing that would block.

app/controllers/application_controller.rb
lazydevorg: Yes @Lockyy. It is supposed to show emails or any other info put into forms.

app/models/validation_error.rb
lazydevorg: I don't know about their benchmark. We can see how it works, and if it gets slow, we can fix it.
Is that ok?

PR #4463, opened 2024-01-23
Newly created teacher profile record should have trn

slawosz:
https://dfedigital.atlassian.net/jira/software/projects/CPDNPQ/boards/137?selectedIssue=CPDNPQ-1575

### DIFF
--- a/app/services/identity/transfer.rb
+++ b/app/services/identity/transfer.rb
@@ -24,6 +24,9 @@ module Identity
 
     def transfer_identities!
       teacher_profile = TeacherProfile.find_or_create_by!(user: to_user)
+      unless teacher_profile.trn
+        teacher_profile.update!(trn: from_user.teacher_profile&.trn)
+      end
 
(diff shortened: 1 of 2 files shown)


app/services/identity/transfer.rb
```diff
@@ -24,6 +24,9 @@ def initialize(from_user:, to_user:)
 
     def transfer_identities!
       teacher_profile = TeacherProfile.find_or_create_by!(user: to_user)
+      unless teacher_profile.trn
```
cwrw, line 27: @tonyheadford @slawosz I think this is worth chatting about, to confirm teacher profiles are only populated with a trn if the trn is validated? We only dedupe users when there are two teacher profiles with the same trn example https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/npq/dedupe_participant.rb#L31 so we should never get to this point where we are merging users without a trn and need this?

PR #4465, opened 2024-01-23
Dont include already archived users in lookup

slawosz:


spec/services/deduplication_service_spec.rb
```diff
@@ -29,6 +31,10 @@
   describe ".select_users_to_archive" do
     let(:users_to_delete) { DeduplicationService.new.send(:select_users_to_archive) }
 
+    it "does not include admin user" do
```
cwrw, line 34:
```suggestion
    it "does not include archived user" do
```

app/services/deduplication_service.rb
```diff
@@ -8,7 +8,7 @@ def call
 private
 
   def select_users_to_archive
-    ids = User.left_joins(:participant_identities).where(participant_identities: { user_id: nil }).pluck(:id)
+    ids = User.left_joins(:participant_identities).where(participant_identities: { user_id: nil }).where("archived_at is null").pluck(:id)
```
cwrw, line 11: this got me thinking, is there a risk archived users will appear in the API, in an ideal world that should never happen as archived users don't have identities, so will never need to be hidden, but maybe by mistake someone attached an identity to an archived user (need to check how many there are) are we risking those appearing? do we need extra checks in out API queries?

cwrw (approved):

cwrw (approved):

PR #4475, opened 2024-01-25
Fix participant query filtering by cohort

ethax-ross:
### Context

Currently, if we filter by cohort the participants query will return a user if _any_ of their induction records have a schedule with that cohort.

Instead, we only want to return a user if their **latest** induction record schedule has the matching cohort.

- Fix participant query filtering by cohort

Remove `where` clause from `latest_induction_record_query` and apply it in the `participants_for_pagination` (after joining on the latest induction record). 

Remove cohort filtering from `participants_from` method as this always receives the `participants_for_pagination` result as its argument and so the filtering is already applied.

- Switch specs to use ChangeSchedule service

The setup for this is way more verbose, but its probably a better solution in case that service ever changes behaviour.

### Guidance to review

I compared this branch against `main` when querying the participants by cohort for the same lead provider on the snapshot DB and it seems to add in the region of 3-500ms to the response time (though the tests were far from scientific!).

Based on Mook's spike #4286

### DIFF
--- a/spec/services/api/v3/ecf/participants_query_spec.rb
+++ b/spec/services/api/v3/ecf/participants_query_spec.rb
@@ -49,15 +49,16 @@ RSpec.describe Api::V3::ECF::ParticipantsQuery do
         let(:new_cohort) { Cohort.next }
 
         before do
+          partnership = create(:partnership,
+                               school: induction_record.school,
+                               lead_provider:,
+                               cohort: new_cohort,
+                               relationship: false)
+          default_induction_programme = create(:induction_programme, :fip, partnership:)
           create(:school_cohort,
-                 default_induction_programme: induction_programme,
+                 default_induction_programme:,
                  school: induction_record.school,
                  cohort: new_cohort)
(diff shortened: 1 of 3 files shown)


spec/services/api/v3/ecf/participants_query_spec.rb
```diff
@@ -38,12 +38,47 @@
...
+      it "only returns users with induction records in the specified cohort" do
         expect(subject.participants_for_pagination).to match_array([user])
       end
+
+      context "when the user has induction records against multiple cohorts" do
+        let(:new_cohort) { Cohort.next }
+
+        before do
+          create(:school_cohort,
+                 default_induction_programme: induction_programme,
```
cwrw, line 53: I think we need to be careful here as we're linking this school cohort to the existing induction programme, which is linked to another partnership in another cohort. Is there a way we can restructure this setup so all the records are linked to the same cohort?

cwrw (approved):

cwrw (approved):

PR #4489, opened 2024-01-29
CPLD-2748 Update declarations when transferring identities

cwrw:
### Context

- Ticket: https://dfedigital.atlassian.net/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2748

To ensure participant declarations are consistent with profile users when transferring identities, update the user id on any declarations attached to transferred profiles.

For each profile attached to a transferred identity, update participant declaration user id as well

### DIFF
--- a/app/services/identity/transfer.rb
+++ b/app/services/identity/transfer.rb
@@ -27,8 +27,9 @@ module Identity
 
       from_user.participant_identities.each do |identity|
         identity.update!(user: to_user)
-        identity.participant_profiles.each do |profile|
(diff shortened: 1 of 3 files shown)


ethax-ross (approved):
Makes sense 👍 do we need to run a backfill for previous transfers?

cwrw:
> Makes sense 👍 do we need to run a backfill for previous transfers?
Yes, there is a separate ticket for this, as has to be done separately to manage comms to Providers as there are around a few thousands I believe 😬

PR #4499, opened 2024-01-31
CST-2299: keep providers when moving cohort on historical induction r…

ltello:
…ecords

### Context
When a historical induction record is moved to a different cohort, the current code might change providers because the induction record is attached to the default induction programme of the school for the new cohort. 
This has the negative impact of the original LP not being able to see the participant anymore in the API.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2299)

When the historical induction record involved is doing FIP and LP/DP are already assigned, we will keep them by attaching it to an existing induction programme with this providers in the new cohort or create a new one with a new relationship for the same providers.

### DIFF
--- a/app/services/induction/amend_participant_cohort.rb
+++ b/app/services/induction/amend_participant_cohort.rb
@@ -123,12 +123,23 @@ module Induction
       end
     end
 
+    def fip_historical_induction_programme_for(school_cohort, lead_provider_id, delivery_partner_id)
+      partnership = Partnership.create_with(relationship: true)
+                               .find_or_create_by!(school_id: school_cohort.school_id,
+                                                   cohort_id: school_cohort.cohort_id,
+                                                   lead_provider_id:,
+                                                   delivery_partner_id:)
+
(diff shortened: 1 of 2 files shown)


app/services/induction/amend_participant_cohort.rb
```diff
@@ -123,12 +123,23 @@ def current_induction_record_changed?
       end
     end
 
+    def fip_historical_induction_programme_for(school_cohort, lead_provider_id, delivery_partner_id)
+      partnership = Partnership.create_with(relationship: true)
```
cwrw, line 127: Could we add two tests to ensure we check the following:
- defaults to an existing partnership which **isn't a relationship** if it exists for a "transferred" participants between two providers in that cohort
- otherwise creates a new relationship partnership for that provider in that cohort
just to make sure we're explicit in testing this?

ltello: Done

tonyheadford (approved):
LGTM

PR #4514, opened 2024-02-06
[CST-2181] Build automated emails admin dashboard

leoapost:
### Context

- Ticket: CST-2181

This PR implements the [Email Schedule dashboard](https://github.com/DFE-Digital/manage-training-for-early-career-teachers-prototype/pull/53)

The `sent to` and `bounce` stats need more work design wise, so I've left them out of this PR until the design is worked out.

Add the EmailSchedule model to persist email schedules
Add basic CRUD functionality for the admins to manage the email schedules
Add a daily job to run every midday to process the daily email schedules

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -288,11 +288,7 @@ Rails.application.routes.draw do
   get "admin/performance", to: "admin/performance/overview#show"
   namespace :admin do
     namespace :performance do
-      resources :email_schedules do
-        member do
-          get "delete", action: :delete
-        end
-      end
+      resources :email_schedules
       resource :overview, only: :show, controller: :overview
       get "/errors", to: "validation_errors#index", as: "validation_errors"
       get "/errors/:form/:attribute", to: "validation_errors#show", as: "validation_error"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -288,6 +288,11 @@ Rails.application.routes.draw do
   get "admin/performance", to: "admin/performance/overview#show"
   namespace :admin do
     namespace :performance do
+      resources :email_schedules do
+        member do
+          get "delete", action: :delete
+        end
+      end
       resource :overview, only: :show, controller: :overview
       get "/errors", to: "validation_errors#index", as: "validation_errors"
       get "/errors/:form/:attribute", to: "validation_errors#show", as: "validation_error"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -288,7 +288,7 @@ Rails.application.routes.draw do
   get "admin/performance", to: "admin/performance/overview#show"
   namespace :admin do
     namespace :performance do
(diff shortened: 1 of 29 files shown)


app/views/admin/performance/_nav.html.erb.orig
```diff
@@ -0,0 +1,19 @@
...
+  <%= component.with_nav_item(path: admin_performance_validation_errors_path) do %>
+    Validation errors
+  <% end %>
+=======
+  
+  <%= component.with_nav_item(path: admin_performance_email_schedules_path) do %>
+    Email Schedule
+  <% end %>
+
+>>>>>>> bdf515180 (Create CRUD pages the EmailSchedule)
```
tonyheadford, line 15: got some rebase stuff here

config/routes.rb
```diff
@@ -288,6 +288,11 @@
   get "admin/performance", to: "admin/performance/overview#show"
   namespace :admin do
     namespace :performance do
+      resources :email_schedules do
+        member do
+          get "delete", action: :delete
```
tonyheadford, line 293: why not standard http DELETE?

spec/models/email_schedule_spec.rb
```diff
@@ -6,6 +6,25 @@
...
+    describe "#validate_future_schedule_date" do
+      let(:past_scheduled_email) { build(:seed_email_factory, scheduled_at: Date.yesterday) }
+      let(:future_scheduled_email) { build(:seed_email_factory, scheduled_at: Date.tomorrow) }
+
+      it "adds an error when scheduled_at is not in the future" do
+        past_scheduled_email.valid?
+
+        expect(past_scheduled_email.errors[:scheduled_at]).to include("The schedule date must be in the future")
+      end
+
```
tonyheadford, line 19: could we add a spec to show today? might need a travel_to to help with that?  We have to allow changes to happen on the day prior to the midday send so would be good to demonstrate it

tonyheadford (commented):
Looks good, I've made a couple of comments

config/routes.rb
leoapost: Good catch. I had a confirmation delete page initially but then I realised it's not needed. I will change it to http DELETE.

spec/models/email_schedule_spec.rb
leoapost: @tonyheadford the schedule date needs to be in the future according to the ACs.

We could change it to allow schedules on the date prior to the midday, but same day schedules prior to midday seem like edge cases to me. I would keep it simple for now and change it in the future if need to.

app/views/admin/performance/_nav.html.erb.orig
leoapost: Good catch. I've removed it.

PR #4515, opened 2024-02-06
[CPDLP-2753] Fix change schedule service to update all cohorts on all Induction records

leandroalemao:
### Context

- Ticket: [CPDLP-2753](https://dfedigital.atlassian.net/browse/CPDLP-2753)

The schools team flagged to us that their amend cohort service is inconsistent with our change schedule service. 
Their service amends all IRs to point to a certain cohort, and select the default induction programme for that cohort meaning providers might change (via partnership).

Our service amends only the latest IR, meaning that another cohort exists in older IRs.

We’ve decided to make our approaches consistent, and to amend our service to update all induction records.

[CPDLP-2753]: https://dfedigital.atlassian.net/browse/CPDLP-2753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/change_schedule.rb
+++ b/app/services/change_schedule.rb
@@ -117,13 +117,17 @@ private
   def update_induction_records!
     return unless relevant_induction_record
 
-    Induction::ChangeInductionRecord.call(
-      induction_record: relevant_induction_record,
-      changes: {
-        schedule: new_schedule,
-        induction_programme:,
-      },
-    )
+    if in_target_schedule?(relevant_induction_record) && !in_target_cohort?(relevant_induction_record)
+      relevant_induction_record.update!(induction_programme:, schedule: new_schedule)
+    else
+      Induction::ChangeInductionRecord.call(
+        induction_record: relevant_induction_record,
+        changes: {
+          schedule: new_schedule,
+          induction_programme:,
+        },
+      )
+    end
   end
 
   def update_ecf_records!
@@ -183,14 +187,21 @@ private
     end
   end
 
+  def historical_schedule(historical_record)
+    Finance::Schedule::ECF.find_by(cohort: new_schedule.cohort, schedule_identifier: historical_record.schedule_identifier)
+  end
+
   def update_historical_induction_records!
     historical_records.all? do |historical_record|
       next true if in_target?(historical_record)
+      next true if !in_target_schedule?(relevant_induction_record) && in_target_cohort?(relevant_induction_record)
 
       historical_induction_programme = historical_induction_programme(historical_record)
-      if historical_induction_programme.present?
-        historical_record.update!(induction_programme: historical_induction_programme,
-                                  schedule: new_schedule)
+      historical_schedule = historical_schedule(historical_record)
+
+      if historical_induction_programme && historical_schedule
+        historical_record.update!(induction_programme: historical_induction_programme(historical_record),
+                                  schedule: historical_schedule(historical_record))
       end
     end
   end
--- a/app/services/change_schedule.rb
+++ b/app/services/change_schedule.rb
@@ -131,6 +131,64 @@ private
 
     update_school_cohort_and_schedule!
     update_induction_records!
+    update_historical_induction_records!
+  end
+
+  def in_target?(induction_record)
+    in_target_cohort?(induction_record) && in_target_schedule?(induction_record)
+  end
+
+  def in_target_cohort?(induction_record)
+    induction_record.cohort_start_year == new_schedule.cohort_start_year
+  end
+
+  def in_target_schedule?(induction_record)
+    induction_record.schedule == new_schedule
+  end
+
+  def historical_target_school_cohort(school)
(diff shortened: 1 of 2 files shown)


app/services/change_schedule.rb
```diff
@@ -131,6 +131,71 @@ def update_ecf_records!
...
+        historical_record.update!(induction_programme: historical_induction_programme(historical_record),
+                                  schedule: new_schedule)
+      rescue ActiveRecord::RecordInvalid
+        false
+      end
+    end
+  end
+
+  def historical_records
+    return [] unless participant_profile
```
cwrw, line 196: should we return early if the profile is ECF? Also should we return early unless there is a cohort change required and not just schedule? Otherwise we will be changing schedule on all induction records unnecessarily

app/services/change_schedule.rb
```diff
@@ -131,6 +131,71 @@ def update_ecf_records!
...
+  end
+
+  def update_historical_induction_records!
+    historical_records.all? do |historical_record|
+      next true if in_target?(historical_record)
+
+      begin
+        historical_record.update!(induction_programme: historical_induction_programme(historical_record),
+                                  schedule: new_schedule)
+      rescue ActiveRecord::RecordInvalid
```
cwrw, line 189: I am wondering around why would the record be invalid, and how this would bubble up in the service?

app/services/change_schedule.rb
```diff
@@ -131,6 +131,71 @@ def update_ecf_records!
...
+  def in_target_cohort?(induction_record)
+    induction_record.cohort_start_year == new_schedule.cohort_start_year
+  end
+
+  def in_target_schedule?(induction_record)
+    induction_record.schedule == new_schedule
+  end
+
+  def historical_target_school_cohort(historical_record)
+    historical_record.school.school_cohorts.for_year(new_schedule.cohort_start_year).first.presence || SchoolCohort.create!(school: historical_record.school,
```
cwrw, line 150: is there a risk that we get a school cohort which is a CIP for a FIP cohort transfer?

cwrw: maybe we can check with Shahad/Nathan around creating school cohorts for schools in different cohorts based on this to be sure, as the amend cohort tool errors on this point?

app/services/change_schedule.rb
```diff
@@ -131,6 +131,71 @@ def update_ecf_records!
...
+
+  def historical_induction_programme(historical_record)
+    return historical_record.induction_programme if in_target_cohort?(historical_record)
+
+    historical_school_cohort = historical_target_school_cohort(historical_record)
+
+    set_default_programme = !historical_record.enrolled_in_fip? || !historical_record.lead_provider_id || !historical_record.delivery_partner_id
+
+    if set_default_programme
+      historical_school_cohort.default_induction_programme.presence || InductionProgramme.create_with(partnership: historical_record.partnership)
```
cwrw, line 171: I think we get to this point if the participant does not belong to a lead provider/ is in CIP, so we shouldn't be creating a partnership unless I misunderstood something?

app/services/change_schedule.rb
```diff
@@ -131,6 +131,71 @@ def update_ecf_records!
...
+
+  def historical_school_cohort_partnership(historical_record)
+    Partnership.create_with(relationship: true)
+    .find_or_create_by!(school: historical_record.school,
+                        cohort: new_schedule.cohort,
+                        lead_provider: historical_record.lead_provider,
+                        delivery_partner: historical_record.delivery_partner)
+  end
+
+  def historical_induction_programme(historical_record)
```
cwrw, line 163: is there a way we can break up this method, it's doing quite a bit here?

spec/services/change_schedule_spec.rb
```diff
@@ -199,6 +199,32 @@
                 service.call
               }.to change { participant_profile.reload.school_cohort }.to(new_school_cohort)
             end
+
```
cwrw, line 202: could we add more test scenarios please, to cover the different cases.

First if there is an existing partnership vs new partnership, CIP vs FIP, induction records with cohort change vs no cohort change, existing school cohort vs non existing etc.

cwrw (commented):
thanks @leandroalemao I've left a few comments and happy to chat through those. Also thinking further on this, I think we should also consider as discussed before extracting this logic to a stand alone service which can be reused by the amend cohort service as well. There are a few differences but I wonder if we can set those via flags passed into the service to distinguish? happy to chat through that as well

leandroalemao:
Hi @cwrw I had thought about extracting to a service but am thinking about getting the logic right first and then extract as much as we can to a stand alone service which can be shared.

cwrw:
We've decided to not implement historic induction record cohort changes due to the high complexity and issues they will cause, closing for now

PR #4521, opened 2024-02-08
[CPDLP-2700] Send provider id to BigQuery when api rate limited

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2700

* Updated rack attack initializer to include correct `user_id` (from API token)
* Added test

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -62,6 +62,14 @@ ActiveSupport::Notifications.subscribe("throttle.rack_attack") do |_name, _start
   path = request.fullpath
   user = request.env["warden"]&.user
 
+  unless user
+    bearer_token = request.authorization.to_s.split("Bearer ").last
+    if bearer_token.present?
+      current_api_token = ApiToken.find_by_unhashed_token(bearer_token)
+      user = current_api_token&.owner
+    end
+  end
+
   Rails.logger.warn("[rack-attack] Throttled request #{request_id} from #{ip} to '#{path}'")
 
   # Web requests are sent to BigQuery via a concern in the ApplicationController.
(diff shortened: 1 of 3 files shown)


spec/support/shared_examples/rate_limiting_support.rb
```diff
@@ -38,6 +38,14 @@
         before { FeatureFlag.activate(:dfe_analytics) }
 
         it { expect { perform_request }.to have_sent_analytics_event_types(:web_request) }
+
+        it "analytics event.user_id set to current_user.id" do
```
ethax-ross, line 42: ```it "sends the current user id to dfe_analytics"```

ethax-ross (approved):
Looks good, just a question (probably my lack of understanding on how NPQ reg clients authenticate) and a wording suggestion

spec/requests/rate_limiting_spec.rb
```diff
@@ -88,6 +104,10 @@ def perform_request
       def change_condition
         set_auth_token(bearer_token2)
       end
+
+      def current_user
+        nil
```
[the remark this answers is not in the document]
mooktakim: The tokens don't have an "owner", they just return text `npq_registration_application`

spec/support/shared_examples/rate_limiting_support.rb
mooktakim: Updated 👍

ethax-ross (approved):
LGTM 👍

config/initializers/rack_attack.rb
```diff
@@ -62,6 +62,12 @@ def remote_ip
   path = request.fullpath
   user = request.env["warden"]&.user
 
+  unless user
+    bearer_token = request.authorization.to_s.split("Bearer ").last
```
cwrw, line 66: we're adding an extra lookup to all throttled requests, I wonder if we can return early if there is no bearer token to avoid looking up the API token if it's empty anyway?

mooktakim: Updated code to check if token exists 👍

cwrw (approved):

PR #4522, opened 2024-02-09
Fix flip/flopping of users in Identity::Transfer

ethax-ross:
[Jira-2793](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2793)

### Context

There are two places where users can flip/flop via the `Identity::Transfer` service; `NPQ::DedupeParticipant` and `StoreValidationResult`. This happens because we exclude the 'current' user in each of these services from the query to find the primary user for the TRN. If the current user _is_ the primary user for the TRN, we will always select a newer `TeacherProfile` and transfer the primary user to that.

- Prevent flip/flop when deduping NPQ participants

Currently, if we call `NPQP::DedupeParticipant` with an `NPQApplication` that is associated with the oldest `TeacherProfile` it will always transfer this user onto a more recent `TeacherProfile` (as we exclude the `from_user` from the `to_user` lookup).

This results in users flip/flopping on `TeacherProfile` records. Instead, we should never exclude the oldest `TeacherProfile` and only transfer if the `NPQApplication` user is more recent.

- Prevent flip/flop on StoreValidationResult

When we call `StoreValidationResult` it will dedupe the participant profile by default. If it happens to be called with the primary user (that has the oldest `TeacherProfile`) then this can result in the user transferring to a newer, duplicate user.

Always find the oldest `TeacherProfile` with a matching `trn` to dedupe to, and only dedupe if that user is not the same as the one on the `participant_profile`.

- Make Identity::Transfer more robust

We would never want to transfer to/from the same user, but currently we don't prevent this and it results in a `ParticipantIdChange`.

Do nothing if the `to_user` and `from_user` are the same.

Protect against service being called with a `nil` to/from user.

- Extract primary user lookup to service/simplify

Pull the primary user lookup into a service so that we can DRY up the callers.

Simplify the logic in the `DedupeParticipant` service (as the `Transfer::Identity` service is now more robust we can cut a lot of the validation here).

### Guidance to review

I thought about modifying the `Transfer::Identity` to prevent a transfer if the `TeacherProfile` of the `to_user` is newer than that of the `from_user`, but I'm not sure that this service should be that restrictive (there may be cases where we want to intentionally do this?).

Best reviewed by-commit.

### DIFF
--- a/app/services/identity/primary_user.rb
+++ b/app/services/identity/primary_user.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Identity
+  class PrimaryUser < BaseService
+    class << self
+      def find_by(trn:)
+        TeacherProfile
+          .includes(:user)
+          .oldest_first
+          .where(trn:)
+          .first
+          &.user
+      end
+    end
+  end
+end
--- a/app/services/npq/dedupe_participant.rb
+++ b/app/services/npq/dedupe_participant.rb
@@ -8,37 +8,20 @@ module NPQ
     attribute :npq_application
     attribute :trn
 
-    validates :npq_application, presence: { message: I18n.t(:missing_npq_application) }
-    validates :trn, presence: { message: I18n.t(:missing_trn) }
-    validates :primary_user_for_trn, presence: { message: I18n.t(:missing_primary_user_for_trn) }
-    validates :application_user, presence: { message: I18n.t(:missing_application_user) }
+    validates :trn, presence: true
+    validates :npq_application, presence: true
     validate :trn_validated
-    validate :duplication_exists
 
     def call
       return if invalid?
 
-      Identity::Transfer.call(from_user: application_user, to_user: primary_user_for_trn)
+      Identity::Transfer.call(from_user: npq_application.user, to_user: primary_user_for_trn)
     end
 
   private
 
-    def application_user
-      @application_user ||= User.find_by(id: npq_application&.user_id)
-    end
-
     def primary_user_for_trn
-      @primary_user_for_trn ||= TeacherProfile
-        .joins(:user)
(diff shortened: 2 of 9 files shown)


app/services/npq/dedupe_participant.rb
```diff
@@ -8,47 +8,24 @@ class DedupeParticipant
...
 
     def trn_validated
-      return if npq_application&.teacher_reference_number_verified?
-
-      errors.add(:trn, I18n.t(:trn_not_validated))
-    end
-
-    def dedupe_already_taken
-      return if from_user&.participant_id_changes.blank?
-      return unless from_user.participant_id_changes.where(from_participant: [from_user, to_user], to_participant: [from_user, to_user]).any?
```
cwrw, line 49: I wonder if we should keep some form of validation regarding participant id changes existing for this change, just incase to avoid creating duplicates of those?

cwrw (commented):
looking good, just the one comment

app/services/npq/dedupe_participant.rb
ethax-ross: I don't think we need it as the `Identity::Transfer` service is robust against this already?

https://github.com/DFE-Digital/early-careers-framework/blob/main/spec/services/identity/transfer_spec.rb#L166-L174

app/services/identity/primary_user.rb
```diff
@@ -0,0 +1,16 @@
...
+        TeacherProfile
+          .includes(:user)
+          .oldest_first
+          .where(trn:)
+          .first
+          &.user
+      end
+    end
+  end
+end
```
mooktakim, line 16: Great idea 👍

PR #4529, opened 2024-02-12
do not include the latest record in the list of historical records

ltello:
### Context
Bug fix in Induction::AmendParticipantCohort: Now that historical records are set a LP/DP in a different way to the latest induction record, we can't include the latest induction record in the list of historical records.

### DIFF
--- a/app/services/induction/amend_participant_cohort.rb
+++ b/app/services/induction/amend_participant_cohort.rb
@@ -105,9 +105,7 @@ module Induction
     end
 
     def all_records_in_target?
-      historical_records.all? do |induction_record|
-        in_target?(induction_record)
-      end
+      induction_records.all? { |induction_record| in_target?(induction_record) }
     end
 
     def current_induction_record_changed?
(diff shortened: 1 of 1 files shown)


app/services/induction/amend_participant_cohort.rb
```diff
@@ -148,9 +146,7 @@ def historical_records_changed?
     end
 
     def historical_records
-      return [] unless participant_profile
-
-      @historical_records ||= participant_profile.induction_records.order(created_at: :desc)
+      @historical_records ||= induction_records - [induction_record]
```
peteryates, line 149: Is the `#order` not important here?

peteryates (approved):
Makes sense to me

app/services/induction/amend_participant_cohort.rb
ltello: We need no order because the processing of each historical induction record is unique to that record and not depends on anything else.

PR #4533, opened 2024-02-13
Don't clear TRN if teacher profile has declarations

ethax-ross:
[Jira-2802](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2802)

### Context

Currently, we check to ensure that there are no NPQ profiles on the `TeacherProfile` before we clear the `trn`, however we also need to check that the associated participant profiles do not have any declarations (as declarations should be tied back to a `TeacherProfile` with a `trn`).

- Don't clear TRN if teacher profile has declarations

Update the `ValidateDetails` and `EcfManualValidation` services to check for declarations prior to clearing a `trn`.

### DIFF
--- a/app/models/teacher_profile.rb
+++ b/app/models/teacher_profile.rb
@@ -31,4 +31,8 @@ class TeacherProfile < ApplicationRecord
   def self.ransackable_attributes(_auth_object = nil)
     %w[trn]
   end
+
+  def has_participant_profile_with_declarations?
+    participant_profiles.any? { |profile| profile.participant_declarations.any? }
+  end
 end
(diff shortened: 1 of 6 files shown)


leandroalemao (approved):
lgtm 👍 😄

cwrw (approved):

ethax-ross:
We're closing this PR after talking to Tony:

- This change will have a knock-on effect to support users that we hadn't considered and it will make it harder to get out and may result in more support tickets falling to the devs.
- We're going to fix the data up where we can by populating TRNs for profiles with declarations.
- We're also fixing up user data separately which will mitigate the risk (using TRN to check for declarations won't happen if we can match on user first).

PR #4534, opened 2024-02-13
[CPDLP-2796] Clean up flipflopping users

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2796

* One off script to find all flip flopped users and transfer them to primary user.
* Updated release notes

### DIFF
--- a/app/services/oneoffs/flipflop_users_cleanup.rb
+++ b/app/services/oneoffs/flipflop_users_cleanup.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module Oneoffs
+  class FlipflopUsersCleanup
+    def call
+      result = []
+
+      flip_flop_users.each do |from_user|
+        if from_user.archived?
+          result << [from_user.id, "User is archived"]
+          next
+        end
+
+        from_trn = from_user.teacher_profile.trn
+        if from_trn.blank?
+          result << [from_user.id, "TRN does not exist"]
+          next
+        end
+
+        to_user = primary_user(from_trn)
+
+        if from_user == to_user
+          result << [from_user.id, "already the primary user"]
+          next
+        end
+
+        if to_user.archived?
+          result << [from_user.id, "to_user [#{to_user.id}] is archived"]
+          next
+        end
+
+        Identity::Transfer.call(from_user:, to_user:)
+        result << [from_user.id, "transfer to user [#{to_user.id}]"]
+      end
+
+      result
+    end
+
+    def primary_user(trn)
+      TeacherProfile
+        .joins(:user)
+        .includes(:user)
+        .oldest_first
+        .where(trn:)
+        .first&.user
+    end
+
+    def flip_flop_users
+      users = []
+
+      ParticipantIdChange.find_each do |pic|
+        if ParticipantIdChange.where(to_participant_id: pic.from_participant_id, from_participant_id: pic.to_participant_id).exists?
+          users << pic.user
+        end
+      end
+
+      users.uniq
+    end
+  end
+end
--- a/app/services/oneoffs/flipflop_users_cleanup.rb
+++ b/app/services/oneoffs/flipflop_users_cleanup.rb
@@ -17,7 +17,7 @@ module Oneoffs
           next
         end
 
-        to_user = primary_user(from_trn)
+        to_user = Identity::PrimaryUser.find_by(trn: from_trn)
 
         if from_user == to_user
           result << [from_user.id, "already the primary user"]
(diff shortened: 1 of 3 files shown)


app/services/oneoffs/flipflop_users_cleanup.rb
```diff
@@ -0,0 +1,60 @@
...
+        end
+
+        Identity::Transfer.call(from_user:, to_user:)
+        result << [from_user.id, "transfer to user [#{to_user.id}]"]
+      end
+
+      result
+    end
+
+    def primary_user(trn)
```
ethax-ross, line 39: This is in prod now https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/identity/primary_user.rb#L6

app/services/oneoffs/flipflop_users_cleanup.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module Oneoffs
+  class FlipflopUsersCleanup
+    def call
+      result = []
+
+      flip_flop_users.each do |from_user|
+        if from_user.archived?
+          result << [from_user.id, "User is archived"]
```
ethax-ross, line 10: I don't feel all that strongly about it, but I started trying to standardise how we record information in these one off scripts - if you want to use it there's a concern here:

https://github.com/DFE-Digital/early-careers-framework/blob/main/lib/has_recordable_information.rb

app/services/oneoffs/flipflop_users_cleanup.rb
```diff
@@ -0,0 +1,60 @@
...
+    def primary_user(trn)
+      TeacherProfile
+        .joins(:user)
+        .includes(:user)
+        .oldest_first
+        .where(trn:)
+        .first&.user
+    end
+
+    def flip_flop_users
```
ethax-ross, line 48: I think we could do this with fewer queries by joining on the `ParticipantIdChange` table, something like:

```
    def flip_flop_users
      ids = ParticipantIdChange
        .joins("INNER JOIN participant_id_changes a ON participant_id_changes.from_participant_id = a.to_participant_id")
        .joins("INNER JOIN participant_id_changes b ON participant_id_changes.to_participant_id = b.from_participant_id")
        .pluck(:user_id)
      User.where(id: ids)
    end
```

mooktakim: Cool, will replace the method with the above

spec/services/oneoffs/flipflop_users_cleanup_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    it "runs flip flop user cleanup" do
+      result = subject.call
+      expect(result.size).to eql(5)
+      expect(result[0]).to eql([user1.id, "User is archived"])
+      expect(result[1]).to eql([user2.id, "TRN does not exist"])
+      expect(result[2]).to eql([user3.id, "already the primary user"])
+      expect(result[3]).to eql([user4.id, "transfer to user [#{user3.id}]"])
+      # binding.pry
```
ethax-ross, line 33: 🗑️

app/services/oneoffs/flipflop_users_cleanup.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Oneoffs
+  class FlipflopUsersCleanup
+    def call
+      result = []
+
+      flip_flop_users.each do |from_user|
+        if from_user.archived?
```
ethax-ross, line 9: So on re-reading this it looks like we take the `user_id` in the query, which is the `to_user` of the `ParticipantIdChange` - so `from_user` is actually the `to_user` here (but having it as `from_user` maybe makes sense as its the user we're potentially flipping away from?). 

If I've got that right, then surely we would want to continue and transfer away from them if they are archived? It's doesn't really matter either way because we're going to fix all those users in #4530 anyway, but I don't think the logic makes sense here?

mooktakim: From the flip flop pov, its actually both from and to users. I guess `flip_flop_users` returns all affected users.

At this point, we're looping them and treating each one as the "from user" to determine if there is a "to user" (if primary, that means we shouldn't be transferring etc).

spec/services/oneoffs/flipflop_users_cleanup_spec.rb
mooktakim: 🤭

app/services/oneoffs/flipflop_users_cleanup.rb
ethax-ross: I see that makes sense 👍

ethax-ross (approved):
LGTM 👍

PR #4539, opened 2024-02-14
Add service to populate missing TRNs

ethax-ross:
[Jira-2803](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2803)

### Context

We have a number of `TeacherProfile` records in the database that do not have a `trn`. In some cases we can infer their `trn` by looking at associated `NPQApplication` and `ECFParticipantValidationData`. We can also use the `ParticipantIdChanges` to find a `trn` from a user that has been transferred to/from.

- Add service to populate missing TRNs

Add one-off service to populate missing `trn` values. Replaces the existing `BackfillTRNToTeacherProfile` service as this was NPQ-specific.

### Guidance to review

> :warning: As it stands the script will fix as many teacher profiles as it can regardless of if they have declarations; I didn't think there was harm in this as its what the previous backfill script did. There are cases where we clear a `trn` explicitly though, so we maybe want to take that into account and exclude those records from this?


app/services/oneoffs/populate_missing_trns.rb
```diff
@@ -0,0 +1,80 @@
...
+      reset_recorded_info
+
+      record_info("~~~ DRY RUN ~~~") if dry_run
+
+      ActiveRecord::Base.transaction do
+        teacher_profiles_without_trns.find_each do |teacher_profile|
+          trns = lookup_trns(teacher_profile)
+          next if trns.empty?
+
+          teacher_profile.update!(trn: trns.first)
```
cwrw, line 19: Maybe we should avoid updating records if they have more than 1 ten?

app/services/oneoffs/populate_missing_trns.rb
```diff
@@ -0,0 +1,80 @@
...
+          .where(teacher_reference_number_verified: true)
+          .pluck(:teacher_reference_number)
+      }.flatten
+    end
+
+    def lookup_trns_from_ecf_validation(users)
+      participant_profiles = users.map(&:participant_profiles).flatten
+
+      participant_profiles.map { |participant_profile|
+        participant_profile.ecf_participant_validation_data&.trn
```
cwrw, line 58: Just to confirm this record ensures that the ten is verified?

cwrw (commented):
Looks good, only some minor questions. I think  if we confirm we're populating verified trns we are probably ok

app/services/oneoffs/populate_missing_trns.rb
ethax-ross: Good point, I was checking for NPQ applications but not the others, which seem to be inferred from the `ECFParticipantEligibility`. I've added that in now.

cwrw (approved):
🙌

app/services/oneoffs/populate_missing_trns.rb
```diff
@@ -0,0 +1,94 @@
...
+          .where(teacher_reference_number_verified: true)
+          .pluck(:teacher_reference_number)
+      }.flatten
+    end
+
+    def lookup_trns_from_ecf_validation(users)
+      participant_profiles = users
+        .map(&:participant_profiles)
+        .flatten
+        .select(&method(:trn_verified?))
```
peteryates, line 60: I can't even find this method 🙃

ethax-ross: it's further down, using `&method` is just shorthand, so these are the same:

```
.select { |participant_profile| trn_verified?(participant_profile) }
.select(&method(:trn_verified)
```

app/services/oneoffs/populate_missing_trns.rb
```diff
@@ -0,0 +1,60 @@
...
+        raise ActiveRecord::Rollback if dry_run
+      end
+
+      recorded_info
+    end
+
+  private
+
+    def lookup_trns(user)
+      trns_from_npq_applications = lookup_trns_from_npq_appliations(user)
```
cwrw, line 35:
```suggestion
      trns_from_npq_applications = lookup_trns_from_npq_applications(user)
```

cwrw (approved):

PR #4540, opened 2024-02-14
Fix change log so that CPD lead providers show up by name instead of ID

tonyheadford:
### Context

Little fix to look up to see if a `whodunnit` from a version record is a `CpdLeadProvider` if it is not a `User` for the changes log

Look up `CpdLeadProvider` if `User` look up fails


peteryates (approved):
Looks good

PR #4541, opened 2024-02-14
CST-2348: keep schedule when the participant is already in the target…

ltello:
… cohort

### Context

Currently, the amend participant cohort service sets the received schedule to the participant or the default one for the target cohort, otherwise.
However that is not correct when the participant is already in the target cohort (maybe not all their historical records).
In such a case the tool should keep the participant's current schedule and make sure the historical records moved to the target cohort get assigned that current schedule.

[Ticket](https://dfedigital.atlassian.net/browse/CST-2348)

### DIFF
--- a/spec/services/induction/amend_participant_cohort_spec.rb
+++ b/spec/services/induction/amend_participant_cohort_spec.rb
@@ -260,7 +260,13 @@ RSpec.describe Induction::AmendParticipantCohort do
           let(:historical_school_source_cohort) { create(:school_cohort, cohort: source_cohort) }
           let(:historical_school) { historical_school_source_cohort.school }
           let!(:induction_programme) { create(:induction_programme, :fip, school_cohort: historical_school_source_cohort) }
-          let!(:historical_record) { create(:induction_record, participant_profile:, induction_programme:) }
+          let!(:historical_record) do
+            create(:induction_record,
+                   participant_profile:,
+                   induction_programme:,
+                   induction_status: :changed,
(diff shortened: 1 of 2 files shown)


spec/services/induction/amend_participant_cohort_spec.rb
```diff
@@ -377,6 +383,44 @@
...
+                                             .and not_change { current_induction_record.reload.induction_programme }
+                                                    .and not_change { current_induction_record.reload.participant_profile.schedule }
+                                                           .and not_change { current_induction_record.reload.participant_profile.school_cohort }
+                                                                  .and change { historical_induction_record.reload.schedule }
+                                                                         .from(historical_induction_record.schedule)
+                                                                         .to(current_induction_record.schedule)
+                                                                         .and change { historical_induction_record.reload.cohort_start_year }
+                                                                                .from(historical_induction_record.cohort_start_year)
+                                                                                .to(current_induction_record.cohort_start_year)
```
peteryates, line 421: Normally I'd say there's a lot going on here and suggest it's split into separate expectations, but if it's urgent let's get it shipped and revisit later?

peteryates (approved):

spec/services/induction/amend_participant_cohort_spec.rb
ltello: Thanks Peter. You are right. I just wanted to not make the suite too slow but the test looks quite unreadable and not clear what's tested. 
Refactored!.

PR #4542, opened 2024-02-15
Allow restricting by provider on declaration migration

ethax-ross:
[Jira-2811](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-2811)

### Context

We have a use case where we only want to migrate declarations from one statement to another for a particular lead provider.

- Allow restricting by provider on declaration migration

### Guidance to review

Screenshots of migration ran on snapshot DB are in the Jira ticket.


cwrw (commented):
Could we also add restricted types of declarations, as we would like to avoid moving completed ones?

cwrw (approved):

cwrw (approved):

PR #4544, opened 2024-02-15
[CPDLP-2799] Fix declarations to point to correct user

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2799

* Created `Oneoffs::FixParticipantDeclarationUser` with tests
* There are total 7208 declarations
  * 81 declarations have a `declaration.participant_identities.size > 1`
    * 47 declarations have `declaration.user.id` that does not match `declaration.participant_profile.user.participant_identities.user_id`

### DIFF
--- a/app/services/oneoffs/fix_participant_declaration_user.rb
+++ b/app/services/oneoffs/fix_participant_declaration_user.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Oneoffs
+  class FixParticipantDeclarationUser
+    def call
+      ActiveRecord::Base.transaction do
+        incorrect_participant_declarations.in_batches.each_record do |declaration|
+          declaration.update!(user_id: declaration.participant_profile.participant_identity.user_id)
+        end
+      end
+    end
+
+    def incorrect_participant_declarations
+      ParticipantDeclaration
+        .joins(participant_profile: :participant_identity)
+        .where("participant_declarations.user_id != participant_identities.user_id")
+    end
+  end
+end
--- a/spec/services/oneoffs/fix_participant_declaration_user_spec.rb
+++ b/spec/services/oneoffs/fix_participant_declaration_user_spec.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+RSpec.describe Oneoffs::FixParticipantDeclarationUser do
+  let(:cpd_lead_provider) { create(:cpd_lead_provider, :with_lead_provider) }
+  let(:declaration1) { create(:ect_participant_declaration, cpd_lead_provider:) }
+  let(:user1) { declaration1.user }
+  let(:declaration2) { create(:ect_participant_declaration, cpd_lead_provider:) }
+  let(:user2) { declaration2.user }
+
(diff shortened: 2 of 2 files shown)


app/services/oneoffs/fix_participant_declaration_user.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def incorrect_participant_declarations
+      ParticipantDeclaration
+        .joins(participant_profile: :participant_identity)
+        .where("participant_declarations.user_id != participant_identities.user_id")
+    end
+
+    def declaration_has_transferred_user?(declaration)
+      # participant_identities have been moved to another user using Identity::Transfer
+      declaration.user.participant_identities.empty? &&
```
cwrw, line 23: Im just curious if we didn't include this, what are the numbers for declarations we need to move across? As thinking maybe we should do that regardless?

app/services/oneoffs/fix_participant_declaration_user.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Oneoffs
+  class FixParticipantDeclarationUser
+    def call
+      ActiveRecord::Base.transaction do
+        incorrect_participant_declarations.in_batches.each_record do |declaration|
+          if declaration_has_transferred_user?(declaration)
+            declaration.update!(user_id: declaration.participant_profile.user.id)
```
cwrw, line 9: is this true for ECF users as well? Or do we need to do any induction record lookups?

app/services/oneoffs/fix_participant_declaration_user.rb
mooktakim: There are 7208 declarations that have incorrect `user_id`.
81 of return `false` with `declaration_has_transferred_user ?`

app/services/oneoffs/fix_participant_declaration_user.rb
cwrw: Mook has confirmed we dont need IR lookups

app/services/oneoffs/fix_participant_declaration_user.rb
cwrw: thanks Mook, I think might be worth checking a few of those incase and if they are ok maybe we change all of them and don't limit to transferred users? Let me know your thoughts :)

app/services/oneoffs/fix_participant_declaration_user.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Oneoffs
+  class FixParticipantDeclarationUser
+    def call
+      ActiveRecord::Base.transaction do
+        incorrect_participant_declarations.in_batches.each_record do |declaration|
+          declaration.update!(user_id: declaration.participant_profile.user.id)
```
cwrw, line 8: any chance we can use the participant profile participant identity user instead? as that's usually our source of truth? (not teacher profile user?)

cwrw (approved):

PR #4547, opened 2024-02-16
Include deleted duplicate data when archiving unvalidated/undeclared profiles

tonyheadford:
### Context

As part of finishing off archiving unvalidated/undeclared participants, some of the remaining participants have deleted duplicate records. It was decided that we will include the serialised data from the deleted duplicate record in the archive for the "primary" profile and enable us to continue archiving and removing these users/profiles.

**Note**: I have refactored duplication of the destruction of the participant_profile and associated records into it's own service in the `Archive` module. This also allows it to tested in one place but it may not be desirable to be so accessible that the routine could be misused. I have added a warning message in the service code but would appreciate some comments/feedback/opinions on this.

- Include deleted duplicates in participant profile archive
- Remove associated `Finance::ECF::DeletedDuplicate` records during archival
- Adds `Archive::DestroyECFProfileData` service

### DIFF
--- a/app/serializers/archive/participant_profile_serializer.rb
+++ b/app/serializers/archive/participant_profile_serializer.rb
@@ -39,5 +39,10 @@ module Archive
     attribute :participant_profile_states do |participant_profile|
       ParticipantProfileStateSerializer.new(participant_profile.participant_profile_states).serializable_hash[:data]
     end
+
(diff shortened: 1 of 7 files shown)


leoapost (approved):
LGTM

cwrw (approved):

PR #4559, opened 2024-02-19
Upgrade to govuk-frontend 5.1.0

peteryates:
### Context

The headline change here is the new crown graphic.

| Before | After |
|-------|------------|
| Screenshot from 2024-02-19 17-00-30 | Screenshot from 2024-02-19 17-00-38 |

### DIFF
--- a/spec/cypress/support/step_definitions/common-interaction.js
+++ b/spec/cypress/support/step_definitions/common-interaction.js
@@ -68,7 +68,7 @@ const elements = {
 const get = (element) => cy.get(elements[element] || element);
 
 When("I set {string} to {string}", (element, value) => {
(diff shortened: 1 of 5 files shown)


spec/cypress/support/step_definitions/common-interaction.js
peteryates, line 1: These `.click` and `.check` calls all work every time for me locally but fail in CI without `force: true`.

We're just using standard GOV.UK frontend stuff, no JS etc, no idea why it's required.

cwrw:
🙌 closing https://github.com/DFE-Digital/early-careers-framework/pull/4513

cwrw (approved):

PR #4560, opened 2024-02-20
[Bug-Fix] Modify queries to exclude ineligible ppts from chaser email

tonyheadford:
### Context

In a recent chaser email to schools it was noticed in support that some participants did not have QTS. This fixes the queries so that they do not include participants that have other eligibility issues.

Check the `ECFParticipantEligibility` record for other flags and only include those that have no induction

### DIFF
--- a/app/queries/ects/with_an_appropriate_body_and_unregistered_query.rb
+++ b/app/queries/ects/with_an_appropriate_body_and_unregistered_query.rb
@@ -6,9 +6,8 @@ module Ects
       InductionRecord
         .active
         .training_status_active
-        .joins(participant_profile: :ecf_participant_eligibility)
-        .where(participant_profile: { induction_start_date: nil, type: "ParticipantProfile::ECT" })
-        .where(ecf_participant_eligibility: { qts: true, active_flags: false, different_trn: false, previous_induction: false, no_induction: true })
+        .joins(:particpant_profile)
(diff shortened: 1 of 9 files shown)


app/queries/ects/with_an_appropriate_body_and_unregistered_query.rb
```diff
@@ -6,8 +6,9 @@ def call
       InductionRecord
         .active
         .training_status_active
-        .joins(:participant_profile)
+        .joins(participant_profile: :ecf_participant_eligibility)
         .where(participant_profile: { induction_start_date: nil, type: "ParticipantProfile::ECT" })
+        .where(ecf_participant_eligibility: { qts: true, active_flags: false, different_trn: false, previous_induction: false, no_induction: true })
```
peteryates, line 11: Do these conditions apply anywhere else too? Worth combining them into a scope?

peteryates (approved):
Looks good, left a open ended question

app/queries/ects/with_an_appropriate_body_and_unregistered_query.rb
tonyheadford: Good question. It might make it clearer if these were in a scope/scopes.

peteryates (approved):
Nice!

PR #4580, opened 2024-02-22
Enable major version updates in Dependabot with exception for Rails

leoapost:
### Context

- Ticket: N/A

This PR updates our Dependabot configuration to allow major version upgrades for Ruby gems, recognising the benefits of keeping our codebase current. 

Making an exception for the Rails gem to maintain tighter control over its major updates.


Lockyy:
As long as we're doing proper checks this should hopefully be alright, but we'll need to be really careful and make sure they're properly checked. Major versions being explicitly breaking after all.

leoapost:
We need to further assess our approach to major version upgrades. Closing this for the time being

PR #4588, opened 2024-02-26
[CST-2360] Enhance AB selection based on School Type Codes

leoapost:
### Context

- Ticket: CST-2360

Given that not every Appropriate Body (AB) supports all school type codes, it's essential to tailor the AB options displayed to School Induction Tutors (SITs) or Admins during the AB selection process for a school's induction programme.

IStip is an example that only supports independent schools.

- Introduce a new attribute in the Appropriate Body model to specify the supported school type codes, with a default setting of an empty array.
- Ensure only ABs compatible with a school's type code are shown as options when an SIT or Admin selects an AB for an induction programme.


tonyheadford (approved):
LGTM 👍

PR #4593, opened 2024-02-28
CST-2217: allow multi-cohort participants

ltello:
### Context

It has been decided to allow multi-cohort participants.
When a participant needs a permanent cohort setup, we will add a new induction record to the chain linked to the new school cohort default providers and the term equivalent schedule.
Historical records will be left untouched. 

The Induction::AmendParticipantCohort used to change the cohort of a participant will now stick to the rules above.


cwrw (approved):
thanks @ltello 👍

PR #4609, opened 2024-03-06
[CST-2398] Upgrade Ruby on Rails to v7.1.3.2

leoapost:
### Context

- Ticket: CST-2398

This is an attempt to upgrade Rails to the latest v7.1

- Upgrade Rails to v7.1.3.2
- Upgrade `ransack` and `paper_trail` to Rails 7.1 compatible versions
- Keep using the configuration defaults from v7.0 until we are comfortable bringing the codebase up to date
- Add attributes to the models with non-integer enums to clearly define the column types
- Fix failing specs


app/models/participant_declaration.rb
```diff
@@ -26,6 +26,7 @@ class ParticipantDeclaration < ApplicationRecord
            through: :billable_statement_line_items,
            source: :statement
 
+  attribute :state, :string
```
peteryates, line 29: Interesting that this is needed. [Background reading](https://github.com/rails/rails/issues/45668) for anyone else who comes across it.

spec/controllers/api/api_controller_spec.rb
```diff
@@ -28,7 +29,7 @@ class DummyController < Api::ApiController
       let(:bearer_token) { "Bearer #{token}" }
 
       it "requests authentication via the http header" do
-        expect(response.status).to eq(200)
+        expect(response.status).to eq(204)
```
peteryates, line 32: Not entirely sure why this changed from `200` to `204` 🤔  Is it just the default for the empty `#index` above?

peteryates (approved):
This looks good to me, all seems sensible and thorough.

Just left one question about the change in response code.

PR #4613, opened 2024-03-08
CST-2331: extract email personalisation stuff from Mail::Redirector i…

ltello:
Extract email personalisation stuff from Mail::Redirector into its own separate module and remove email redirection feature:

#### `Mail::Personalisation`
    This module is meant to be registered as an email interceptor on Action Mailer for debugging/testing purposes.
    
    For non-production or test emails it adds a personalisation entry like this:
        subject_tags: "[local to:aaa@d1.md1,bbb@d2.md2]"

    For production or test environments, the entry added is empty:
        subject_tags: ""

    Also, for non production or test environments, emails with a :subject personalisation field are modified to
    preprend the app_name (from ENV['VCAP_APPLICATION'] || 'local') and the list of destination emails addresses
    like this:
       subject: "[app_name to:aaa@d1.md1,bbb@d2.md2] original_subject_content"

### Context

[Ticket](https://dfedigital.atlassian.net/browse/CST-2331)


peteryates (approved):
Excellent, so much better!

app/lib/mail/personalisation.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        # TODO: This enforces backward compatibility with existing emails and should be deleted as soon as all the emails
+        # are modified to use `subject_tags` instead
+        override_personalisation(mail, :subject) { |subject| new_subject(mail, subject) } if enabled?
+      end
+
+    private
+
+      def app_name
+        ENV["VCAP_APPLICATION"] ? JSON.parse(ENV["VCAP_APPLICATION"])["application_name"] : "local"
```
cwrw, line 32: could we drop the `ENV["VCAP_APPLICATION"]` as this was for only PaaS specific implementation and doesn't apply in the new world?

app/lib/mail/personalisation.rb
```diff
@@ -0,0 +1,62 @@
...
+#
+module Mail
+  module Personalisation
+    class << self
+      def delivering_email(mail)
+        set_personalisation(mail, :subject_tags, subject_tags(mail))
+
+        # TODO: This enforces backward compatibility with existing emails and should be deleted as soon as all the emails
+        # are modified to use `subject_tags` instead
+        override_personalisation(mail, :subject) { |subject| new_subject(mail, subject) } if enabled?
```
cwrw, line 26: are we leaving this in for queued emails? as this only applies in test envs, and not sure if we still need it 🤔

app/lib/mail/personalisation.rb
ltello: Sure. Is there a way to get the name of the app in the new world?

peteryates: `Rails.env` probably makes most sense.

ltello: Yes, that's was my alternative if we don't have anything better.

cwrw: yea I agree, if we want to go with this, I think we're missing the target email logic from the new class, if we want to drop it but keep the personalised tags even better

cwrw (approved):
thanks @ltello

PR #4620, opened 2024-03-13
[CST-2366] Add pre-term reminder comms

leoapost:
### Context

- Ticket: CST-2366

We want to send pre-term comms to SITs in eligible schools that are running FIP or CIP programmes in 2023/24 cohort, to remind them to report any changes.

- Add the mailer for the Notify template that will be sent
- Add a query class to return all the schools that need to be contacted
- Add a bulk mailer to trigger the emails


tonyheadford (approved):
LGTM!

PR #4621, opened 2024-03-14
[CPDLP-2849] Fix bug when accepting a NPQ application where another for the same course has already been accepted

leandroalemao:
### Context

- Ticket: [CPDLP-2849](https://dfedigital.atlassian.net/browse/CPDLP-2849)

TDT have flagged that they were not able to submit a start declaration for this participant. The user has two accepted applications for the same course with different lead providers. Our service should not allow this to happen.

Fix bug when accepting a NPQ application where another for the same course has already been accepted.

[CPDLP-2849]: https://dfedigital.atlassian.net/browse/CPDLP-2849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/npq/application/accept.rb
+++ b/app/services/npq/application/accept.rb
@@ -106,8 +106,14 @@ module NPQ
         @participant_profile ||= ParticipantProfile.find(npq_application.id)
       end
 
+      def trn
+        @trn ||= npq_application.teacher_reference_number_verified? ? npq_application.teacher_reference_number : user.teacher_profile&.trn
+      end
+
+      # Make sure existing or the new teacher profile has a valid TRN so we fetch same_trn_user & other_accepted_applications_with_same_course correctly
       def teacher_profile
-        @teacher_profile ||= user.teacher_profile || user.build_teacher_profile
+        @teacher_profile ||= (user.teacher_profile&.assign_attributes(trn:)
+                              user.teacher_profile) || user.build_teacher_profile(trn:)
       end
 
       def user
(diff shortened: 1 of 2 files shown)


app/services/npq/application/accept.rb
```diff
@@ -106,8 +106,14 @@ def participant_profile
...
       end
 
+      def trn
+        @trn ||= npq_application.teacher_reference_number_verified? ? npq_application.teacher_reference_number : user.teacher_profile&.trn
+      end
+
+      # Make sure existing or the new teacher profile has a valid TRN so we fetch same_trn_user & other_accepted_applications_with_same_course correctly
       def teacher_profile
-        @teacher_profile ||= user.teacher_profile || user.build_teacher_profile
+        @teacher_profile ||= (user.teacher_profile&.assign_attributes(trn:)
```
cwrw, line 115: I wonder if instead of assigning attributes here, we use the new trn method you've added as follows in the query, so it's directly looking up the trn instead:

```
 def same_trn_user
  return if trn.blank?

  @same_trn_user ||= TeacherProfile
                     .oldest_first
                     .where(trn: trn)
                     .where.not(id: teacher_profile.id)
                     .first
                     &.user
end
```
Let me know your thoughts :)

cwrw (approved):
🙌

PR #4658, opened 2024-03-26
Update import scripts for cohort/npq contract

ethax-ross:
### Context

We have a new field `automatic_assignment_period_end_date` on Cohort that we want to allow in the CSV and default to the 31st March of the following year (relative to the `start_year` of the Cohort).

The NPQ Contract also has a new `monthly_service_fee` field; this was already defaulting to `0.0` in the import script, but this will also allow it to be added as part of the CSV.

- Update import scripts for cohort/npq contract

### DIFF
--- a/app/services/importers/create_cohort.rb
+++ b/app/services/importers/create_cohort.rb
@@ -31,12 +31,15 @@ module Importers
       start_year = row["start-year"].to_i
       logger.info "CreateCohort: Creating cohort for starting year #{start_year}"
 
+      default_automatic_assignment_period_end_date = Date.new(start_year + 1, 3, 31)
+
       Cohort.upsert(
         {
           start_year:,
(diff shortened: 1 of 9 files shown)


cwrw (commented):
looks great, only comment would be to add those new columns to `cohorts.csv` and `npq_contract` csvs that are used for seeding, otherwise 👍

ethax-ross:
> looks great, only comment would be to add those new columns to `cohorts.csv` and `npq_contract` csvs that are used for seeding, otherwise 👍

I've added `monthly_service_fee` to the npq contract CSVs, but it looks like the `automatic-assignment-period-end-date` was already added to `cohorts.csv`; it was missing from a few other CSVs dotted about so I've hopefully updated all of them now

cwrw (approved):

PR #4690, opened 2024-04-05
Add `funded_place` to NPQApplications

pmanrubia:
### Ticket

https://dfedigital.atlassian.net/browse/CPDLP-2935

### Description

We need to capture if a participant has been allocated a funding place by providers. This field will be updated via the API in cohorts 2024 to true and false. Other cohorts will default to nil

### DIFF
--- a/spec/requests/api/v1/npq_profiles_spec.rb
+++ b/spec/requests/api/v1/npq_profiles_spec.rb
@@ -210,6 +210,7 @@ RSpec.describe "NPQ profiles api endpoint", type: :request do
             updated_at
             participant_identity_id
             targeted_support_funding_eligibility
+            funded_place
           ])
 
           expect(application_as_json).to match({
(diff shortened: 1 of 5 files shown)


cwrw (commented):
apart from fixing the failure in `npq_profiles_spec.rb:200` looks great 👍

spec/requests/api/v1/npq_profiles_spec.rb
```diff
@@ -232,6 +232,7 @@ def owner
             "employer_name" => "Local Authority",
             "employment_role" => "manager",
             "employment_type" => "local_authority_virtual_school",
+            "funded_place" => nil,
```
cwrw, line 235: could this be added to the `except` on line 207 instead?

cwrw (approved):
one comment otherwise feel free to merge after

PR #4692, opened 2024-04-05
[CPDLP-2921] Add dummy/test data for 2024 cohort

leandroalemao:
### Context

- Ticket: [CPDLP-2921](https://dfedigital.atlassian.net/browse/CPDLP-2921)

The schools team are looking to create a 2024 cohort and spin up some review apps so that we can start testing scenarios of opening reg. In order to do this they need a review app with the 2024 schedules, contracts and data. They can't do this without the contract data.

Add dummy/test data for 2024 cohort.

[CPDLP-2921]: https://dfedigital.atlassian.net/browse/CPDLP-2921?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


db/new_seeds/base/add_schools_for_cohortless_testing.rb
```diff
@@ -6,7 +6,9 @@ def padded_urn(num)
 
 cohort_22 = Cohort.find_by(start_year: 2022) || FactoryBot.create(:seed_cohort, start_year: 2022)
 cohort_23 = Cohort.find_by(start_year: 2023) || FactoryBot.create(:seed_cohort, start_year: 2023)
+cohort_24 = Cohort.find_by(start_year: 2024) || FactoryBot.create(:seed_cohort, start_year: 2024)
```
cwrw, line 9: I wonder if instead of adding cohorts here we can use previous and current cohort? and amend the naming to match that? cc @tonyheadford any thoughts on this?

Gemfile.lock
```diff
@@ -333,6 +333,7 @@ GEM
     memory_profiler (1.0.1)
     method_source (1.0.0)
     mini_mime (1.1.5)
+    mini_portile2 (2.8.5)
```
cwrw, line 336: could we not commit this as part of this pr please?

db/new_seeds/base/add_schools_for_cohortless_testing.rb
```diff
@@ -4,10 +4,10 @@ def padded_urn(num)
   num.to_s.rjust(6, "0")
 end
 
-cohort_22 = Cohort.find_by(start_year: 2022) || FactoryBot.create(:seed_cohort, start_year: 2022)
-cohort_23 = Cohort.find_by(start_year: 2023) || FactoryBot.create(:seed_cohort, start_year: 2023)
-FactoryBot.create(:ecf_schedule, cohort: cohort_23) if Finance::Schedule::ECF.default_for(cohort: cohort_23).blank?
-
+previous_cohort = Cohort.previous
+current_cohort = Cohort.current
+next_cohort = Cohort.next || FactoryBot.create(:seed_cohort, start_year: current_cohort.start_year + 1)
```
cwrw, line 9: ah I was thinking we avoid the next cohort altogether, and just keep previous and current cohorts and stick to the current setup? happy to chat through this tomorrow

cwrw (approved):
thanks @leandroalemao 👍

db/new_seeds/base/add_schools_for_cohortless_testing.rb
tonyheadford: Good idea

PR #4694, opened 2024-04-08
[Bug-fix] Issue with changes out of the induction record timeline

tonyheadford:
### Context

We have recognised that there is an issue with retrospective changes added to the induction records when the timeline has continued past the record that is being altered. Typically this happens when a provider tells us that the training is withdrawn for participant after the participant has already changed programme/school/provider.
This causes issues because that new record recording the withdrawal status is added to the __end of the timeline__ . The LPDOB team scope the view of the induction records to a provider so do not have visibility of the continuing timeline with a different provider.

Currently this is creating a lot of support issues with participants appearing as not training

I _think_ this can be solved by checking for the presence of an `end_date` on the record that is changing and if present and in the past, nudge the `end_date` back by `1.minute` to allow space to _insert_ the new induction record into the timeline and not affect any subsequent induction records.

Handle changes to induction records with and historic `end_date` by rewinding the `end_date` by a minute and inserting the new one into that space.

### Guidance to review

Try calling `Induction::ChangeInductionRecord` on a induction record in a timeline that already had an`end_date`.  You should see that `end_date` adjusted and used on the new induction record inserted after it.

Please also have a look and ponder if there are any circumstances where this approach might cause issues

### DIFF
--- a/app/services/induction/change_induction_record.rb
+++ b/app/services/induction/change_induction_record.rb
@@ -5,6 +5,11 @@ class Induction::ChangeInductionRecord < BaseService
     ActiveRecord::Base.transaction do
       time_now = Time.zone.now
 
+      default_attrs = {
+        start_date: time_now,
+        school_transfer: false,
+      }
+
       if induction_record.start_date > time_now
         induction_record.update!(changes)
       else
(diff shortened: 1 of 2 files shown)


cwrw (commented):
@tonyheadford Having a think about the repercussions of amending the time for existing IRs and what that means for the touch model/lookups. Let's chat tomorrow around the scenarios so I can get a better understanding, thanks :)

ltello (approved):

peteryates:
@cwrw @tonyheadford - did we get anywhere with this?

app/services/induction/change_induction_record.rb
```diff
@@ -14,12 +19,24 @@ def call
         }
         new_record = induction_record.dup
 
-        if induction_record.end_date.present? && induction_record.end_date > time_now
-          default_attrs[:school_transfer] = induction_record.school_transfer?
-          induction_record.update!(school_transfer: false)
+        if induction_record.end_date.present?
+          if induction_record.end_date < time_now
+            # We're going to possibly be inserting a new record into an existing chain so
```
cwrw, line 24: I've had a long think about this, with three examples, one where the start date is before the end date. one where it is possible that the start date is after the end date, and one where they are the same.

The main issue I see here for all examples is that we are rewriting history, also where the start date is not the real date the change happened which I think is key, and end date on the previous induction record is also not the real date the change happened which is the current time, I find this will affect our timeline of when things happened and will really mess with our audits as well.

Since we are using the indication that the end date is present on an induction record, I'd like to explore options where we query those induction records, and happy to chat through any other options we might have @tonyheadford

tonyheadford:
As discussed with @cwrw closing this as it is possible it may introduce some issues and we may already have a solution to show the correct thing to schools without fixing the underlying data.

PR #4701, opened 2024-04-11
[CPDLP-2936] Add funding_cap field to NPQ contracts

leandroalemao:
### Context

- Ticket: [CPDLP-2936](https://dfedigital.atlassian.net/browse/CPDLP-2936)

We need to capture if a specific cohort for NPQ supports capping. We also would like to capture limits on funded spaces for providers.

### Changes proposed in this pull request

Add `funding_cap` field to NPQ contract.

Documentation has been updated as well https://dfedigital.atlassian.net/wiki/x/BwCT6

[CPDLP-2936]: https://dfedigital.atlassian.net/browse/CPDLP-2936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4726, opened 2024-04-22
NPQ capping: add funded place to accept npq application endpoint

pmanrubia:
- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-2938

### Context

Updates `accept` end point for NPQ capping in all versions of the API. 
The code is behind the feature flag: `npq_capping`.

Related PRs: #4690, #4697, #4701

### DIFF
--- a/app/services/npq/application/accept.rb
+++ b/app/services/npq/application/accept.rb
@@ -160,6 +160,7 @@ module NPQ
       def eligible_for_funded_place
         return unless FeatureFlag.active?("npq_capping")
         return if errors.any?
+        return unless npq_contract.funding_cap.to_i.positive?
 
         if funded_place && !npq_application.eligible_for_funding
           errors.add(:npq_application, I18n.t("npq_application.not_eligible_for_funded_place"))


app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -68,6 +72,20 @@ def access_scope
       def npq_application
         @npq_application ||= npq_lead_provider.npq_applications.includes(:cohort, :npq_course, participant_identity: [:user]).find(params[:id])
       end
+
+      def funded_place
+        permitted_params.dig("attributes", "funded_place")
+      end
+
+      def permitted_params
+        params.require(:data).permit(:type, attributes: %i[schedule_identifier funded_place])
```
cwrw, line 81: could we remove `schedule_identifier` from here as it's only permitted in v3?

app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -68,6 +72,20 @@ def access_scope
       def npq_application
         @npq_application ||= npq_lead_provider.npq_applications.includes(:cohort, :npq_course, participant_identity: [:user]).find(params[:id])
       end
+
+      def funded_place
+        permitted_params.dig("attributes", "funded_place")
+      end
+
+      def permitted_params
```
cwrw, line 80: could we name this accepted permitted params to avoid confusion with other params for other actions?

app/models/npq_application.rb
```diff
@@ -79,6 +79,14 @@ class NPQApplication < ApplicationRecord
   delegate :id, :name, to: :npq_course, prefix: true
   delegate :id, :name, to: :npq_lead_provider, prefix: true
 
+  def current_contract
```
cwrw, line 82: contracts are provider specific, and tied to statements, we need to think about where we could query this, and if this should be on application or on the npq provider/statement instead, happy to chat through this

app/services/npq/application/accept.rb
```diff
@@ -145,6 +151,28 @@ def validate_permitted_schedule_for_course
...
+
+        unless npq_application.eligible_for_funding && funded_place
+          errors.add(:npq_application, I18n.t("npq_application.not_eligible_for_funded_place"))
+        end
+      end
+
+      def funding_cap_available
+        return unless FeatureFlag.active?("npq_capping")
+        return if errors.any?
+        return if funded_place.nil?
```
cwrw, line 168: I was wondering if it actually should fail here since we are expecting it to be populated and not `nil`?

app/services/npq/application/accept.rb
```diff
@@ -145,6 +151,28 @@ def validate_permitted_schedule_for_course
...
+
+        contract = npq_application.current_contract
+
+        return if funded_place && contract.funding_cap.present? && contract.funding_cap.positive?
+
+        errors.add(:npq_application, I18n.t("npq_application.no_funding_cap_available"))
```
cwrw, line 174: I think maybe the check here will probably be that this field is required since there is a cap on the contract 🤔 otherwise it's ok?

spec/requests/api/v1/npq_applications_spec.rb
```diff
@@ -335,6 +339,41 @@
...
+
+      before do
+        default_npq_application.cohort.npq_contracts << create(:npq_contract, npq_course:, version: "1.0", funding_cap: 10)
+        default_npq_application.update!(eligible_for_funding: true)
+      end
+
+      context "when feature flag `npq_capping` is disabled" do
+        before { FeatureFlag.deactivate(:npq_capping) }
+
+        it "updates funded place attribute" do
```
cwrw, line 353:
```suggestion
        it "does not update funded place attribute" do
```

app/controllers/api/v1/npq_applications_controller.rb
mooktakim: removed from v1+v2 👍

app/services/npq/application/accept.rb
```diff
@@ -24,6 +27,9 @@ def call
           npq_application.update!(lead_provider_approval_status: "accepted")
           other_applications_in_same_cohort.update(lead_provider_approval_status: "rejected") # rubocop:disable Rails/SaveBang
           deduplicate_by_trn!
+          if FeatureFlag.active?("npq_capping")
```
cwrw, line 30: can we add the check that if funded place is set as well, incase it's in previous cohorts?

app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -68,6 +68,22 @@ def access_scope
...
+
+      def funded_place
+        if FeatureFlag.active?(:npq_capping)
+          accept_permitted_params.dig("attributes", "funded_place")
+        end
+      end
+
+      def accept_permitted_params
+        params.require(:data).permit(:type, attributes: %i[funded_place])
+      rescue ActionController::ParameterMissing => e
```
cwrw, line 80: I was testing with this and was thinking we avoid this because of past years where we won't need to pass in this request vs 2024, otherwise providers will get an error. Happy to chat through

cwrw (commented):
Looks great! just a minor comment around erroring if the request is not passed in

cwrw (approved):
thanks @mooktakim and @pmanrubia !

PR #4759, opened 2024-05-01
Spike multiple cohorts for participants

ethax-ross:
> ⚠️ WIP, do not merge!

Code to change schedules for the seeds:

```
# Generate token
LeadProviderApiToken.create_with_random_token!(cpd_lead_provider: CpdLeadProvider.find_by_name("Ambition Institute"))

# ECT
cpd_lead_provider = CpdLeadProvider.find_by!(name: "Ambition Institute")

ect_participant_profile = ParticipantProfile::ECT.first
schedule = ect_participant_profile.latest_induction_record.schedule
user = ect_participant_profile.user

service = ChangeSchedule.new({
  cpd_lead_provider:,
  participant_id: user.id,
  course_identifier: "ecf-induction",
  schedule_identifier: schedule.schedule_identifier,
  migrate_declarations: false,
  cohort: 2024,
})

service.call


# Mentor
cpd_lead_provider = CpdLeadProvider.find_by!(name: "Ambition Institute")

mentor_participant_profile = ParticipantProfile::Mentor.first
schedule = mentor_participant_profile.latest_induction_record.schedule
user = mentor_participant_profile.user

service = ChangeSchedule.new({
  cpd_lead_provider:,
  participant_id: user.id,
  course_identifier: "ecf-mentor",
  schedule_identifier: schedule.schedule_identifier,
  migrate_declarations: false,
  cohort: 2024,
})

service.call
```

## How do schools choose an induction programme and set up a default partnership from last year

When the `registration_start_date` of the new academic year/cohort is in the past, the next time a school signs into the dashboard they will be prompted to setup their default provider/delivery partner/appropriate body for the new academic year. 

- When the `registration_start_date` of the new cohort/academic year is in the past, the next time a school signs in they are asked if they expect any ECTs this year
  - If they don't, the journey ends setup for the year (they can view their dashboard with previous years still)
  - If they do, then:
    - They are asked if they want to keep the same provider/delivery partner for new ECTs
    - If they don't, they are asked about how they want to train and this seems to end up with a support ticket
    - If they do, a new `SchoolCohort`, `Partnership` and `InductionProgramme` is created and they are asked about an appropriate body:
      - If they have not appointed one yet, they are taken to the dashboard
      - Otherwise they are asked about their appropriate body and given guidance for next steps

- Any new/transferred ECTs will go to the default partnership unless the school specifies otherwise
  - DQT is checked on adding new ECTs (to get induction start date/participant data)
    - If the participant is not found in DQT they can go forward with registration
    - If they are found in DQT additional validation/checks are made
  - If they want to specify another partnership, it usually goes to a support ticket (there is a small subset they can choose from automatically)  
  - If they want the ECT to continue training with a provider form a previous school
    - Then a relationship type partnership is created

- The school can change their default provider, delivery partner and appropriate body at any time

## What happens if the school doesn't set it up and wait for providers to select different partnerships with delivery partners

If a school does not setup the 2024 academic year and a provider attempts to create a partnership we give an error message:

```
{"errors":[{"title":"school_id","detail":"The school you have entered has not yet confirmed they will deliver DfE-funded training. Contact the school for more information."}]}
```

It looks like an induction programme can be created when a participant is transferred to a school where a partnership doesn't already exist for the school/lead provider.

## How we can move a set of fake participants (ECTs/Mentors) from 2021 to the 2024 cohort (manually with a flag or via certain criteria)

We can use the `ChangeSchedule` service and pass `migrate_declarations: false` to avoid moving 2021 declarations/relax the validation rules around moving schedules with declarations.

We could also automate this when the school completes setup of the 2024 academic year, by querying the incomplete 2021 participants and migrating them:

https://github.com/DFE-Digital/early-careers-framework/blob/main/app/forms/schools/cohorts/setup_wizard/success.rb#L7-L27

## How we can allow providers to declare against them or change their schedule normally
  

Once the participants have been migrated to the 2024 cohort they will behave like any other participant in that cohort. The `RecordDeclaration` service appears to be irrespective of cohort, so it will apply the declaration validation logic across all declarations for the participant (so if they have a paid declaration in the 2021 cohort, they will not be allowed to submit another in the 2024 cohort, for example). Any further schedule changes should also be possible as with a regular 2024 participant.

## How those participants will look like in the API, can we make recommendations on how we can surface them (maybe in both cohorts like we did previously, do declarations need a fix etc.)

On testing this locally for ECTs and Mentors with declarations, they will surface correctly in the participants endpoint for the provider and show as 2024 cohort (and only be returned in the 2024 cohort if filtered):

0.0.0.0:3000/api/v3/participants/ecf/?filter[cohort]=2024

=> returns participants

0.0.0.0:3000/api/v3/participants/ecf/?filter[cohort]=2021

=> returns empty

Calling the participant declarations endpoint results in the declarations being returned for the 2024 cohort as it currently stands due to a bug in the query that filters on the participant schedule:

http://0.0.0.0:3000/api/v3/participant-declarations?filter[cohort]=2024

=> returns the 2024 declarations

http://0.0.0.0:3000/api/v3/participant-declarations?filter[cohort]=2021

=> returns no declarations

If we fix the bug and filter on the induction records cohort, we get both the 2024 and 2021 declarations back when filter by 2021 OR 2024, which is not ideal.

Instead, to get the 2021 declarations when filtering by 2021 and the 2024 declarations when filtering by 2024 we can look for the induction record relevant to when the declaration was recorded (see the updated ParticipantDeclarationsQuery in this PR).


app/services/api/v3/participant_declarations_query.rb
```diff
@@ -118,15 +119,26 @@ def delivery_partner_ids
...
-          participant_profile: [
-            [schedule: :cohort],
-            { induction_records: [
-              :cohort,
-              { induction_programme: :partnership },
-            ] },
-          ],
-        )
+        # Get the latest induction record for each participant declaration (kind of - may need more thought)
+        subquery = InductionRecord
```
ethax-ross, line 123: @cwrw to have a think about this

cwrw: chatted in person: from @ethax-ross best thing is to stamp the cohort on the declaration on creation to avoid all this. We will need to think about backfilling submitted/ineligible declarations

ethax-ross:
Closing the spike.

PR #4761, opened 2024-05-01
[CST-2547] Fix query for schools that have not registered participants

tonyheadford:
### Context

- Ticket: [CST-2547](https://dfedigital.atlassian.net/browse/CST-2547)

Query that support the mailer was using `.missing` to find school cohorts that did not have induction records, however the hidden left joins underneath this with `induction_programmes` led to the inclusion of schools that had both empty and induction programmes with participants

Change the query to be more explicit in not selecting school cohorts that have induction programmes with induction records.

### Guidance to review

[CST-2547]: https://dfedigital.atlassian.net/browse/CST-2547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/queries/schools/that_have_not_added_participants_query.rb
+++ b/app/queries/schools/that_have_not_added_participants_query.rb
@@ -21,9 +21,12 @@ module Schools
     end
 
     def school_cohorts_that_have_not_added_participants
+      # need to ensure that there are no induction programmes in the school/cohort with participants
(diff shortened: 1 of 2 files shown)


steventux (approved):
Query makes sense to me. :rocket:

PR #4762, opened 2024-05-01
Defaults funded_place attribute when changing cohorts

pmanrubia:
### Ticket

https://dfedigital.atlassian.net/browse/CPDLP-2951

### Description

- We can move applications across cohorts in 2021/2022/2023 where funding place remains null
- We can applications from cohort 2021/2022/2023 to  2024 where funding place gets defaulted to true if eligible_for_funding is set to true
- We can applications from cohort 2021/2022/2023 to  2024 where funding place gets defaulted to false if eligible_for_funding is set to false
- If an application has declarations which are billable we cannot move an application across


app/services/change_schedule.rb
```diff
@@ -143,11 +143,20 @@ def update_npq_application_cohort!
     participant_profile.npq_application.update!(cohort: new_schedule.cohort)
   end
 
+  def update_funded_place!
+    return unless FeatureFlag.active?(:npq_capping)
+    return unless participant_profile.npq_application.cohort.start_year >= 2024
```
cwrw, line 148: could we drive this via the contracts (see comment from accept pr around how to query those) happy to have a chat about that

cwrw: maybe we can also check if they are moving cohorts so we don't unset an existing value for 2024 moving between schedules?

app/services/change_schedule.rb
```diff
@@ -143,11 +143,34 @@ def update_npq_application_cohort!
...
 
+  def update_funded_place!
+    return unless FeatureFlag.active?(:npq_capping)
+    return unless npq_contract.funding_cap&.positive?
+
+    application = participant_profile.npq_application
+    application.update!(funded_place: application.eligible_for_funding)
+  end
+
+  def npq_contract
```
cwrw, line 154: discussed in person: if we can move maybe to the npq lead provider model or a service

spec/services/change_schedule_spec.rb
```diff
@@ -595,6 +595,58 @@
...
+              end
+
+              it "updates the cohort on the npq application" do
+                service.call
+                expect(participant_profile.npq_application.cohort).to eq(new_cohort)
+              end
+
+              context "when funding from non funding cohort to funding cohort" do
+                before do
+                  npq_contract.update!(funding_cap: 1)
```
cwrw, line 617: I think this should remain `nil` if it's from a nun funded cohort?

app/models/npq_contract.rb
```diff
@@ -12,4 +12,15 @@ class NPQContract < ApplicationRecord
   validates :per_participant, numericality: { greater_than: 0 }
   validates :recruitment_target, numericality: { only_integer: true, greater_than: 0 }
   validates :funding_cap, numericality: { only_integer: true, greater_than_or_equal_to: 0, allow_nil: true }
+
+  def find_latest_by(npq_lead_provider:, npq_course:, cohort:)
```
cwrw, line 16:
```suggestion
  def self.find_latest_by(npq_lead_provider:, npq_course:, cohort:)
```

cwrw (commented):
having a think about this might need to add a few things:
- I'll confirm tomorrow if we show the funded place in teh participant serialiser
- if the value is already set we should not reset it, and only set it if its `nil` and coming from a non funded cap cohort

Let's chat tomorrow :)

cwrw (approved):
🌟

PR #4754, opened 2024-04-30
CST-2540: get induction start date from DQT induction periods

ltello:
### Context

DQT records provide 2 ways to extract the induction start date of a participant:

“startDate” field in the “induction” section of the record

minimum “startDate” of the “periods” section in the “induction” section of the record.

Our service always has gone for 1. It seems the most accurate date is provided by 2. 

This ticket is about changing our code from 1. to 2. whenever we check DQT for a participant induction start date throughout the whole codebase.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2540)

### Changes proposed in this pull request

- Update `Participants::CheckAndSetCompletionDate` to get the DQT induction start date of a participant from the earliest `startDate` of the `induction` `periods` 
- Temporary workround (until we migrate this SO to DQT::V3) on `DQTRecordPresenter` to call V3 to pick the induction start date of the participant

PR #4765, opened 2024-05-03
Revert "CST-2540: get induction start date from DQT induction periods"

peteryates:
Reverts DFE-Digital/early-careers-framework#4754

Deployment was failing in staging while trying to create the new unique index.

I think we need to manually fix the broken data on staging (and maybe sandbox) before re-reverting this.


cwrw (approved):

peteryates:
Oops sorry yeah this wasn't necessary in the end.

PR #4771, opened 2024-05-07
Amend healthcheck to not fail if database not populated

cwrw:
### Context 
To allow the migraiton environment database to be restored from production there will be a certain amount of time the database is empty. Keeping the healthcheck happy means we can keep the refresh going as otherwise the pipeline will be broken since we rely on the migration web container to be up and healthy.

This work is similar to NPQ thanks @ethax-ross 
https://github.com/DFE-Digital/npq-registration/pull/1283

- Ticket: n/a

- Amend healthcheck to ignore migration environment on db populated check
- Add specs for healthcheck

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,6 +38,7 @@ Rails/UnknownEnv:
     - production
     - review
     - sandbox
+    - migration
     - staging
     - test
     - performance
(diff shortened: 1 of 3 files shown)


leandroalemao (approved):

mooktakim (approved):
👍

PR #4785, opened 2024-05-13
Backfill cohort on ParticipantDeclaration

ethax-ross:
### Context

[Jira-3057](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3057)

We are stamping `cohort` on declaration when created, but for previously created declarations we need to go back and backfill the `cohort`. We do this by looking at the:

Declaration → statement → cohort if available 
Profile → latest induction record → cohort for ECF 
Profile → schedule → cohort for NPQ

- Backfill cohort on ParticipantDeclaration

### Guidance to review

This logic aims to favour a statement if available, otherwise it will default to how the cohort is filtered on the v3 `participant-declarations` endpoint.

If no cohort can be found or the cohort is not clear (for example if there are multiple statements) then the backfill will raise an error and rollback changes.


app/services/oneoffs/backfill_declaration_cohort.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    def infer_cohort(declaration)
+      statement_cohorts = declaration.statements.map(&:cohort).uniq
+
+      if statement_cohorts.many?
+        statement_cohorts = declaration.statement_line_items.order(created_at: :asc).map { |item| item.statement.cohort }.uniq
+        record_info("Declaration #{declaration.id} has multiple statements with different cohorts - selecting the first one")
+      end
+
+      return statement_cohorts.first if statement_cohorts.any?
```
cwrw, line 80: could we be more explicit here, as in only selecting the paid statement cohort? as the order isn't guaranteed?

app/services/oneoffs/backfill_declaration_cohort.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      if declaration.ecf? || declaration.npq?
+        declaration.participant_profile.schedule.cohort
+      else
+        raise UnexpectedDeclarationTypeError, "Unexpected declaration type: #{declaration.class}"
```
cwrw, line 85: is there a risk we'll get any other type of declaration 🤔

ethax-ross: I was planning for the future 😂

cwrw (approved):

PR #4786, opened 2024-05-13
[CST-2546] Use DQT API V3 Get Teacher endpoint

steventux:
### Context

We currently check a teachers record via the DQT API using the v1 endpoint.

This doesn’t give us reliable induction start and end dates.

V3 of the DQT API exposes an array of induction periods which each have a start and end date, we should derive induction start date and end date from the oldest/latest dates in this array.

- Ticket:  https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CST/boards/119?selectedIssue=CST-2546

- Amend DQT Record presentation to use DQT API V3 [GetTeacherResponse](https://preprod.teacher-qualifications-api.education.gov.uk/swagger/index.html?urls.primaryName=v3#/Teachers/GetTeacherByTrn) fields
- Fetch and attempt to match DQT records via the V3 API.
- Derive induction start and end dates from induction periods.
- Rework `DQTRecordCheck` with an additional check via the v1 API endpoint in the case of a TRN mismatch.

### DIFF
--- a/app/presenters/dqt_record_presenter.rb
+++ b/app/presenters/dqt_record_presenter.rb
@@ -44,11 +44,17 @@ class DQTRecordPresenter < SimpleDelegator
   end
 
   def induction_start_date
-    dqt_record.dig("induction", "startDate")
+    induction_periods.to_a
+                     .map { |period| period["startDate"] }
+                     .compact
+                     .min
   end
 
   def induction_completion_date
-    dqt_record.dig("induction", "endDate")
+    induction_periods.to_a
+                     .map { |period| period["endDate"] }
+                     .compact
+                     .max
   end
 
   def exempt?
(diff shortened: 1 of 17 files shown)


tonyheadford (commented):
There's a lot of good stuff in here @steventux . Wondering whether we should split this up and refactor into separate "identifying" and "induction info" services, where for now V1 is used in to confirm the identity and ensuring the TRN belongs to the correct person and then once we have a confirmed TRN we use that with the V3 API for all induction related stuff?  Also as discussed there's a timing issue around registration, but I would like us to have accurate start dates when they are present. What do you think?

steventux:
@tonyheadford thanks for the review - I like the suggestion of splitting this PR up, perhaps I should start with a simpler PR introducing the new V1 and V3 client code and presentation logic.
I agree that sticking with the V1 API client for the initial record check is a simpler solution than trying to combine both API version calls in the same place (it gave me a bit of a headache trying to make this work).
I'll close this PR and split the work up.

PR #4790, opened 2024-05-14
Restore funded_place if modified after acceptance

pmanrubia:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3071

### Description

We set a funded place when an application is accepted. If we revert the acceptance we also need to unset the capping change.

### DIFF
--- a/app/services/npq/change_to_pending.rb
+++ b/app/services/npq/change_to_pending.rb
@@ -34,7 +34,7 @@ module NPQ
         end
         npq_application.update!(lead_provider_approval_status: "pending")
         if FeatureFlag.active?(:npq_capping) && !npq_application.funded_place.nil?
-          npq_application.update!(funded_place: false)
+          npq_application.update!(funded_place: nil)
         end
       end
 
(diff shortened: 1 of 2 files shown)


app/services/npq/application/reject.rb
```diff
@@ -16,6 +16,9 @@ def call
         return self unless valid?
 
         npq_application.update!(lead_provider_approval_status: "rejected")
+        if FeatureFlag.active?(:npq_capping) && npq_application.funded_place?
```
cwrw, line 19: ah this is the wrong service, can we please only do this on the `ChangeToPending` service and set this to `nil` please

app/services/npq/change_to_pending.rb
```diff
@@ -33,6 +33,9 @@ def call
           profile.destroy!
         end
         npq_application.update!(lead_provider_approval_status: "pending")
+        if FeatureFlag.active?(:npq_capping) && !npq_application.funded_place.nil?
+          npq_application.update!(funded_place: false)
```
cwrw, line 37: can we please set this to `nil` not false?

cwrw (approved):

PR #4792, opened 2024-05-14
Enforce cohort_id on ParticipantDeclaration

ethax-ross:
> ⚠️ cannot be merged until backfill has been ran!

### Context

[Jira-3057](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3057)

Now that new declarations are stamped with `cohort_id` and all existing declarations have been backfilled, we can enforce `cohort_id` as a non-nullable field in the database and model.

-  Make cohort_id required on ParticipantDeclaration

### Guidance to review

There are two separate migrations to avoid blocking the table while making it not-null (and still validating it as part of the migration).


db/migrate/20240514104324_validate_check_constraint_and_make_cohort_id_not_null_in_participant_declarations.rb
```diff
@@ -0,0 +1,12 @@
+class ValidateCheckConstraintAndMakeCohortIdNotNullInParticipantDeclarations < ActiveRecord::Migration[7.1]
+  def up
+    validate_check_constraint :participant_declarations, name: "participant_declarations_cohort_id_null"
```
cwrw, line 3: just to confirm, have we done the backfill on all envs or just prod?

cwrw:
> I'm not sure if having the cohort in the find/create block could cause us issues; we set the cohort using:

> Declaration → statement → cohort if available
> Profile → schedule → cohort for NPQ and ECF

> However when we record the declaration we always look at the participant_profile → schedule → cohort. This could lead to > inconsistencies. It might be worth splitting this find/create out into a separate find and create method, where the find looks > up using the previous declaration parameters and the create includes the cohort

Excellent point, Im worried about a scenario where if we keep as is, we'd allow providers to create the same declaration in another cohort as it won't find the existing one.

I'd be up for the find not including cohorts, and the create including cohorts as you suggested :)

db/migrate/20240514104324_validate_check_constraint_and_make_cohort_id_not_null_in_participant_declarations.rb
ethax-ross: Just production so far; I'll do the rest now

app/services/record_declaration.rb
```diff
@@ -131,20 +131,22 @@ def validate_milestone_exists
...
       .submitted
       .or(
         participant_profile
           .participant_declarations
           .billable,
       )
-      .create_or_find_by!(declaration_parameters)
+      .find_by(declaration_parameters)
+
+    existing_declaration || participant_profile.participant_declarations.create!(declaration_parameters.merge(cohort:))
```
cwrw, line 149: do you mind adding a spec for this to check that it doesn't try to create a new declaration if there is an existing one in another cohort?

cwrw (approved):
👍

PR #4801, opened 2024-05-16
Add method/scope to identify participants eligible to change cohort and continue training

ethax-ross:
### Context

[Jira-3060](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3060)

As part of the work to migrate 2021 candidates to the 2024 cohort to continue training we want to add methods on `ParticipantProfile` so that we can identify them and a scope so that we can find them all.

-  Add method to identify if participant can migrate cohort

Add a method `can_change_cohort_and_continue_training?` to check if a participant is eligible for changing cohort in order to continue training in a later cohort, even though they have paid/payable/eligible declarations.

- Add scope to retrieve all participants eligible to migrate cohort

Update the existing method to use the scope for simplicity.

### Guidance to review

The numbers appear consistent with Nathan's word doc (and I've checked with Nathan as well and he's happy):

```
ParticipantProfile::ECT.eligible_to_change_cohort_and_continue_training(in_cohort_start_year: 2024).count
3,199 (3,500 in the word doc)

ParticipantProfile::Mentor.eligible_to_change_cohort_and_continue_training(in_cohort_start_year: 2024).count
14,867 (15,000 in the word doc)
```

I'm not super if `eligible_to_change_cohort_and_continue_training` is the best way to describe this action/pot of participants; open to suggestions.

### DIFF
--- a/db/new_seeds/scenarios/npq.rb
+++ b/db/new_seeds/scenarios/npq.rb
@@ -105,6 +105,7 @@ module NewSeeds
           participant_profile:,
           course_identifier: npq_course.identifier,
           cpd_lead_provider: npq_lead_provider.cpd_lead_provider,
+          cohort: participant_profile.schedule.cohort,
         )
 
         self


app/models/participant_profile.rb
```diff
@@ -73,6 +73,20 @@ def approved?
   # delivery_partner
   delegate :delivery_partner, to: :latest_induction_record, allow_nil: true
 
+  def self.eligible_to_change_cohort_and_continue_training(in_cohort_start_year:)
+    billable_states = %w[eligible payable paid].freeze
```
cwrw, line 77: not in the scope of this pr but we should have this in declarations as a constant similar to statement line items as outside the scope it's used everywhere and needs cleanup 😂

app/models/participant_profile.rb
```diff
@@ -73,6 +73,20 @@ def approved?
...
+
+    joins(:participant_declarations, schedule: :cohort)
+      .where(cohorts: { start_year: in_cohort_start_year - Cohort::OPEN_COHORTS_COUNT })
+      .where("participant_declarations.state IN (?) AND declaration_type != ?", billable_states, "completed")
+      .where.not(id: ParticipantDeclaration.billable.for_declaration(:completed).pluck(:participant_profile_id))
```
cwrw, line 82: Why do we need this as well as thought the previous query is excluding billable completed declaration profiles 🤔

cwrw (commented):
very cleanly done ⭐ have a few questions but otherwise 💯

app/models/participant_profile.rb
ethax-ross: Yeah definitely, it would be nice to tidy all that up

app/models/participant_profile.rb
ethax-ross: Yeah its a bit hard to follow, but essentially we want profiles with billable declarations that are not completed and also have 0 billable declarations that are completed

cwrw (approved):
⭐

PR #4810, opened 2024-05-20
CST-2575: update amend cohort tool to cope with participants moved af…

ltello:
…ter their cohort is frozen for payments

### Context

Accomodate Induction::AmendParticipantCohort service object to accept participants coming from payments_frozen cohorts.

This piece of code will eventually be called by all the features/journeys/events we are updating as part of the Closing 21 cohort implementation.

[Ticket](https://dfedigital.atlassian.net/browse/CST-2575)


db/migrate/20240517220008_add_moved_after_payments_frozen_at_cohort_to_participant_profile.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddMovedAfterPaymentsFrozenAtCohortToParticipantProfile < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    add_column :participant_profiles, :moved_after_payments_frozen_at_cohort, :integer, limit: 2
```
steventux, line 7: I think the naming here is hard to understand. 
This might be because we're attempting to describe the reason for moving from this cohort in the column name.
It might be more flexible to have `previous_cohort` as a foreign key or integer and `cohort_change_reason` as a text or enum description.

steventux (commented):
Do we need feature specs to cover this end-to-end or are unit tests enough?
One small consideration made about db column naming. Happy to go with concensus.

db/migrate/20240517220008_add_moved_after_payments_frozen_at_cohort_to_participant_profile.rb
ltello: The thing is that we only keep track of these specific cohort changes.
I mean, we are not interested in tracking the previous_cohort for other cohort changes we do but for this specific type one.

ltello: Yes, there are other reasons why we move cohort but we are not caring about the reason.
In any case, I'm happy to add a reason field if that clarifies things for devs.

tonyheadford: Throwing an idea out there... would adding a table for this be a better idea?  The table could join participant_profile, cohort and a reason?  It could be more generic and we could include _any_ time we move a participant to a different cohort?  It would be useful for being able to track how many people we've moved because of induction dates/temp cohort placement because we do get asked about numbers for that kind of thing too?

tonyheadford: I agree with @steventux that the naming is difficult to understand, appreciate it's difficult to name though, so that for me is another reason a separate table with the reason might be easier to track these. Much easier to give it a nice reason rather than awkward column naming?

tonyheadford (commented):
I think it's worth considering whether a separate table for these would be better to allow tracking of multiple cohort moves and prevent us from adding even more fields to the participant profile table

db/migrate/20240517220008_add_moved_after_payments_frozen_at_cohort_to_participant_profile.rb
ltello: We can do it but the issue is that if we do, again we are persisting several times the same fact (therefore, data inconsistencies is a matter of time).

Given that we already have induction records that track any change in the induction facts of a participant, I think the best place to flag this is there. That is, every time we change the cohort of a participant we are adding a new induction record to the chain pointing to the new record. So to me, just adding a flag field there stating it is because of previous cohort froze payments is more than enough and no duplication of data persisted.

However, the LPDOB team think that querying the induction records is a pain and adding more fields make it awful for them. So they suggested to put the field into the ParticipantProfile. 
However by pointing the ParticipantProfile to the previous cohort we have persisted twice the cohort change fact: 
   - The previous IR pointing to the previous cohort and the new IR pointing to the new cohort means cohort change.
   - The field in ParticipantProfile pointing to the previous cohort also means cohort change.

These kind of data duplication create inconsistencies sooner or later and I dont' like them at all.

Adding a table would put the changes in its own model but again we would be persisting twice the same facts and therefore very likely to add inconsistencies again.

I'm open to any solution to agree though.

steventux: Would a foreign key (instead of an arbitrary integer column) also help to solve the problem of data duplication and the risk of corruption as you mention @ltello ? This way we always point to a valid Cohort record.

steventux (approved):
This looks good to me but appreciate I have a lot less context than @ltello and @tonyheadford

db/migrate/20240517220008_add_moved_after_payments_frozen_at_cohort_to_participant_profile.rb
tonyheadford: My thought was the separate table would effectively be a history of changes made to a participants cohort and the reason why. Also assuming there was a `created_at` timestamp we'd see the order of them too.  I'm happy either way, I think the addition of the reason is a good call but of course we can only persist the latest cohort move on the participant profile.

tonyheadford (approved):
LGTM

app/models/participant_profile.rb
```diff
@@ -35,10 +35,21 @@ class ParticipantProfile < ApplicationRecord
...
+    provisional_previous_cohort: "provisional_previous_cohort",
+    registration_mistake: "registration_mistake",
+    unknown: "unknown",
+  }, _prefix: true
+
   # Associations
   belongs_to :participant_identity
   belongs_to :schedule, class_name: "Finance::Schedule"
   belongs_to :teacher_profile, touch: true
+  belongs_to :previous_cohort, class_name: "Cohort", optional: true
```
ethax-ross, line 52: Are we sure this will only ever be one cohort? and do we need to be mindful if they migrate 2021 -> 2024 and then back again (would this be `nil` or 2024 in that scenario?)

app/models/participant_profile.rb
```diff
@@ -35,10 +35,21 @@ class ParticipantProfile < ApplicationRecord
     withdrawn: "withdrawn",
   }, _prefix: "training_status"
 
+  attribute :reason_for_new_cohort, :string
+  enum reason_for_new_cohort: {
+    dqt_mismatch_induction_start_date: "dqt_mismatch_induction_start_date",
+    lead_provider_change: "lead_provider_change",
+    payments_frozen_at_previous_cohort: "payments_frozen_at_previous_cohort",
```
ethax-ross, line 42: I was only aware of this reason; what do the others mean?

app/models/cohort.rb
```diff
@@ -82,6 +82,10 @@ def display_name
     start_year.to_s
   end
 
+  def payments_frozen?
```
ethax-ross, line 85: I've been referencing candidates with `eligible_to_change_cohort_and_continue_training?` - we should probably get on the same page here; I'm happy to adopt something like `eligible_to_change_cohort_due_to_payment_freeze?` if that makes sense?

ltello: @ethax-ross I think method names and fields should be reduced only to the domain of the class where they sit. 
I mean, a cohort gets their payments frozen or not.

Then what we do with participants or any other cohort related entity due to a cohort having frozen payments is to me something outside the scope of the Cohort and therefore out of it.

Let's see what other people think about it

app/models/participant_profile.rb
ltello: - `dqt_mismatch_induction_start_date` means that our induction start date mismatch that in DQT and therefore we moved the participant to the cohort calculated from DQT's one.

- `lead_provider_change` means basically the api throught the `change_schedule` endpoint 

But currently these won't be of any use

app/models/participant_profile.rb
ltello: That's why I suggested to put this field in InductionRecord instead, when we can keep track of those back and forth cohort changes. You suggested that paper_trail could be enough if we put this in ParticipantProfile though.

ltello: Please see the conversation above with @steventux and @tonyheadford about suggestions where to place this.

app/models/cohort.rb
ethax-ross: I think it makes sense to be consistent with the cohort naming; its a different model but it directly relates to this method (I'll be updating my method once this goes in; at the moment I just assume cohorts > 3 years older than the current are closed/payments frozen).

db/migrate/20240517220008_add_moved_after_payments_frozen_at_cohort_to_participant_profile.rb
ltello: @ethax-ross what's your view on this?

app/models/participant_profile.rb
cwrw: Im not sure about this, the payments frozen is quite different than the others, the others mean they changed cohort, the payments frozen means they span across both, not sure we should be grouping them the same?

app/models/participant_profile.rb
cwrw: hmm I wonder again if we are trying to group two types of changes under one, I think flagging a new type of participant which spans multiple cohorts is different to a participant which genuinely changed. Happy to chat through this though with the wider group

app/services/induction/amend_participant_cohort.rb
```diff
@@ -100,12 +111,20 @@ def save
...
     end
 
     def current_induction_record_updated?
       ActiveRecord::Base.transaction do
-        Induction::ChangeInductionRecord.call(induction_record:, changes: { induction_programme:, schedule: })
-        participant_profile.update!(school_cohort: target_school_cohort, schedule:)
+        Induction::ChangeInductionRecord.call(induction_record:,
+                                              changes: { induction_programme:,
+                                                         schedule:,
+                                                         training_status: :active })
```
cwrw, line 122: Im not sure about this, I think this field is mostly for provider actions and we should maintain? same below around the `training_status`

app/validators/unfinished_training_participant_profile_validator.rb
```diff
@@ -0,0 +1,16 @@
...
+    record.errors.add(attribute, I18n.t("errors.participant_profile.not_a_participant_profile")) unless participant_profile?(value)
+    record.errors.add(attribute, I18n.t("errors.participant_profile.training_complete")) if completed_training?(value)
+  end
+
+  def participant_profile?(instance)
+    instance.is_a?(ParticipantProfile)
+  end
+
+  def completed_training?(instance)
+    instance&.completed_training?
```
cwrw, line 14: maybe this is something we can hook into @ethax-ross's work?

app/models/cohort.rb
steventux: For me the `eligible_to_change_cohort_` method prefix doesn't really make sense for an instance of a `Cohort` record as it's attempting to explain the state of a relationship between two cohort records. 
`Cohort#payments_frozen?` is intuitive and feels like the sort of naming convention we employ in other services.

app/validators/unfinished_training_participant_profile_validator.rb
ethax-ross: This is the relevant method FYI https://github.com/DFE-Digital/early-careers-framework/blob/main/app/models/participant_profile/ect.rb#L25C12-L25C59

PR #4815, opened 2024-05-20
Update govuk-frontend and associated libs to 5.4.0

peteryates:
Fixes #4809, #4811, #4813

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,9 +58,8 @@ gem "mail-notify", "~> 1.2"
 # do not rely on host's timezone data, which can be inconsistent
 gem "tzinfo-data"
 
-gem "govuk-components", "~> 5.3.2"
-gem "govuk_design_system_formbuilder", "~> 5.3.3"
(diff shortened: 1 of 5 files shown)


peteryates:
I've had a good click around the admin/finance sections, the CSS looks like it's in tact.

Gemfile
```diff
@@ -58,9 +58,8 @@ gem "mail-notify", "~> 1.2"
 # do not rely on host's timezone data, which can be inconsistent
 gem "tzinfo-data"
 
-gem "govuk-components", "~> 5.3.2"
-gem "govuk_design_system_formbuilder", "~> 5.3.3"
-gem "view_component", require: "view_component/engine"
```
cwrw, line 63: just wanted to confirm we're ok removing view component 🤔

peteryates: Yeah it's pulled in by govuk-components anyway so it's unnecessary for it to be in the Gemfile.

cwrw (approved):

PR #4817, opened 2024-05-20
[CST-2554] Add importer to set mentor completions for rule 3

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-2554)
- Import mentor profile IDs from a CSV file containing those mentors that fall under the "rule 3" category that have started but not finished their training but are considered sufficiently trained not to require further training.  The CSV is provided by policy colleagues and this is expected to be an annual task.
  
Adds a new importer to read the CSV file and update those mentors contained within.


app/services/importers/rule_3_mentor_completions.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module Importers
+  class Rule3MentorCompletions < ::BaseService
```
steventux, line 6: Do we document what Rule 3 is anywhere? (I don't see a Rule 1 or Rule 2 importer either) which makes me wonder whether there isn't a better way of explaining what this importer does. Also I don't see any business logic here which would explain a rule of any sort, it's just setting the completion date on the given participants and giving a reason.
How about `CompleteStartedNotCompleted` or something equally descriptive? I get the feeling Rule 3 might not retain its meaning for long.

steventux (approved):
LGTM, I mentioned the naming of the importer class out in comments as I'm not sure what Rule 3 is. Non-blocking if the name makes sense to everyone else.

app/services/importers/rule_3_mentor_completions.rb
tonyheadford: Good to call it out. I'll put some comments in around it, it's very much a thing for how mentor completions are managed but rule 3 doesn't mean much if you don't have the context. Rule 1 is catered for already by our existing data in the DB for early rollout mentors and rule 2 is triggered when we receive "completed" declarations from providers. I'll have a think, thanks

steventux (approved):
:+1: For importer class renaming LGTM

PR #4823, opened 2024-05-21
CST-2556: add payments_frozen_at field to cohort model

ltello:
### Context

Add new field to Cohort model to mark cohorts that freeze payments

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2556):

### DIFF
--- a/spec/models/cohort_spec.rb
+++ b/spec/models/cohort_spec.rb
@@ -191,7 +191,7 @@ RSpec.describe Cohort, type: :model do
       it { is_expected.to be_payments_frozen }
     end
 
-    context "when a date to frozen payments has not been set" do
+    context "when a date to freeze payments has not been set" do
       before do
         subject.payments_frozen_at = nil
       end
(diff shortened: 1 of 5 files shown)


spec/models/cohort_spec.rb
```diff
@@ -182,6 +182,24 @@
     end
   end
 
+  describe "#payments_frozen?" do
+    context "when a date to frozen payments has been set" do
```
ethax-ross, line 186:
```suggestion
    context "when a date to freeze payments has been set" do
```

spec/models/cohort_spec.rb
```diff
@@ -182,6 +182,24 @@
...
+      before do
+        subject.payments_frozen_at = Date.current
+      end
+
+      it { is_expected.to be_payments_frozen }
+    end
+
+    context "when a date to frozen payments has not been set" do
```
ethax-ross, line 194:
```suggestion
    context "when a date to freeze payments has not been set" do
```

ethax-ross (approved):

PR #4831, opened 2024-05-22
CST-2575: closing 2021 allow cohort changes for participants from payments-frozen cohorts

ltello:
### Context

Accomodate Induction::AmendParticipantCohort service object to accept participants coming from payments_frozen cohorts.

This piece of code will eventually be called by all the features/journeys/events we are updating as part of the Closing 21 cohort implementation.

[Ticket](https://dfedigital.atlassian.net/browse/CST-2575)

### Guidance to review


config/locales/en.yml
```diff
@@ -203,6 +203,8 @@ en:
     participant_profile:
       blank: "Not registered"
       not_active: "Not active"
+      not_a_participant_profile: "Not a ParticipantProfile instance"
```
steventux, line 206:
```suggestion
      not_a_participant_profile: "Not a participant profile record"
```

app/models/participant_profile/ect.rb
```diff
@@ -18,6 +18,10 @@ def self.archivable(for_cohort_start_year:, restrict_to_participant_ids: [])
       .where("induction_start_date IS NULL OR induction_start_date < ?", latest_induction_start_date)
   end
 
+  def completed_training?
```
steventux, line 21:
```suggestion
  alias_method :completed_training?, :completed_induction?
```

app/validators/active_participant_profile_validator.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class ActiveParticipantProfileValidator < ActiveModel::EachValidator
```
steventux, line 3: Is this validator used at all?

steventux (approved):
A couple of small suggestions, nothing blocking though.
I might be lacking a bit of understanding but there are a couple of validators added in this PR which don't seem to be called, I assume this is because we have multiple stories pending which will need them?

app/services/induction/amend_participant_cohort.rb
```diff
@@ -147,11 +166,20 @@ def in_target_schedule?(induction_record)
 
     def participant_declarations
       return false unless participant_profile
+      return @participant_declarations if instance_variable_defined?(:@participant_declarations)
 
-      @participant_declarations ||= participant_profile
+      @participant_declarations = if cohort_changed_after_payments_frozen
```
ethax-ross, line 171: I think we're duplicating some of the existing eligibility checks in here; we have methods on `ParticipantProfile::ECF` that let you check if an ECT/mentor is eligible for transferring out of a frozen cohort to continue training, see:

https://github.com/DFE-Digital/early-careers-framework/blob/df43515d992b0e7c02754ae477d209ac6aefa6e4/app/models/participant_profile/ecf.rb#L95-L97

https://github.com/DFE-Digital/early-careers-framework/blob/df43515d992b0e7c02754ae477d209ac6aefa6e4/app/models/participant_profile/ect.rb#L31-L33

https://github.com/DFE-Digital/early-careers-framework/blob/df43515d992b0e7c02754ae477d209ac6aefa6e4/app/models/participant_profile/mentor.rb#L52-L54

These methods implement the logic Nathan provided to determine the outer pot of participants that are eligible:

```
1. The participant is in a payments frozen cohort
2. The profile does not have any completed declarations in the state eligible/payable/paid
3. The profile has at least 1 declaration of any other type that is eligible/payable/paid
4. The `induction_end_date` is `nil` for ECTs
5. The `mentor_completion_date` is `nil` for mentors
```

ltello: We are already checking for 4/5 of that logic in different validations across the service.

I didn't want to call your method yet, because we need to check (as part of another ticket) the scheduled jobs we have running that check DQT for data updates and trigger cohort changes in participants. 
I suspect that for very rare cases those jobs might need to call this service to move 21 participants that meet all the 5 points above but the 3rd.

Let me try the other way around and call your method first and then undo the commit if later on those jobs need to do what I say.

app/services/induction/amend_participant_cohort.rb
```diff
@@ -177,6 +199,16 @@ def target_school_cohort
...
 
     # Validations
+    def eligible_to_be_moved_after_payments_frozen
+      unless participant_profile.eligible_to_change_cohort_and_continue_training?
+        errors.add(:participant_profile, :participant_not_eligible_to_be_moved)
+      end
+    end
+
+    def target_cohort_payments_active
+      errors.add(:target_cohort_start_year, :payments_frozen) if target_cohort&.payments_frozen?
```
ethax-ross, line 209: I think this needs to be more restrictive; after talking with Nathan it seems we only want to let 2021 participants migrate to 2024. This will check the cohort they are migrating to allows payments, but that may be error prone depending on how this service is called/by who. Similarly, once a participant has migrated to 2024 we may want to migrate them _back_ to 2021 (if it was done by mistake). We should be able to do this providing the participant didn't get any billable declarations while in 2024 (but if they have billable declarations in 2021 and they span multiple cohorts they should still be able to migrate back to 2021).

In the `ChangeSchedule` service I'm going to apply this logic:

- When transferring from a frozen cohort you can only transfer into a cohort that is 3 years ahead (so 2021 -> 2024).
- When transferring to an earlier cohort and the participant has `cohort_changed_after_payments_frozen = true`, the cohort must be 3 years earlier and have payments frozen (so 2024 -> 2021). The participant must also have no billable declarations in their current cohort (2024).

ltello: That validation is to prevent participants to be sitting in payment-frozen cohorts.
Given that Nathan allows 2021 participants to go to 2024 and back, I will accomodate it

ltello: The 3 years rule could be too restrictive IMO. What if a 2021 participant resumes training not this year but next year?
This service only allows cohort changes of 2021 participants to the cohort currently open for registration if their original cohort gets payments-frozen.

app/validators/participant_profile_active_validator.rb
```diff
@@ -1,11 +0,0 @@
-# frozen_string_literal: true
-
-class ParticipantProfileActiveValidator < ActiveModel::EachValidator
-  def validate_each(record, attribute, value)
-    record.errors.add(attribute, I18n.t("errors.participant_profile.not_active")) unless active?(value)
-  end
-
-  def active?(participant_profile)
-    participant_profile&.active_record? && participant_profile&.training_status_active?
```
cwrw, line 9: Im curious as to why this changed, and we no longer want training status active? I've checked our service and we prevent changing things if a participant is withdrawn

app/services/induction/amend_participant_cohort.rb
```diff
@@ -102,10 +111,22 @@ def initialize(*)
       @target_cohort_start_year = (@target_cohort_start_year || @schedule&.cohort_start_year).to_i
     end
 
+    def back_to_payments_frozen_cohort?
+      participant_profile&.cohort_changed_after_payments_frozen? && target_cohort&.payments_frozen?
```
cwrw, line 115: In @ethax-ross pr there are a few extra check for declarations existing in one and not the other, maybe worth duplicating the logic here as well?

app/services/induction/amend_participant_cohort.rb
```diff
@@ -176,7 +194,36 @@ def target_school_cohort
...
+    end
+
+    def transfer_from_payments_frozen_cohort
+      unless participant_profile.eligible_to_change_cohort_and_continue_training?
+        errors.add(:participant_profile, :not_eligible_to_be_transferred_from_current_cohort)
+      end
+    end
+
+    def transfer_to_payments_frozen_cohort
+      unless participant_profile.eligible_to_change_cohort_back_to_their_payments_frozen_original?(to_cohort: target_cohort)
```
cwrw, line 220: let's align with @ethax-ross on the keyword arguments as think they are currently `cohort`

app/services/induction/amend_participant_cohort.rb
```diff
@@ -176,7 +194,36 @@ def target_school_cohort
...
+    alias_method :cohort_changed_after_payments_frozen, :transfer_from_payments_frozen_cohort?
+
     # Validations
+
+    def target_cohort_payments_active
+      errors.add(:target_cohort_start_year, :payments_frozen) if target_cohort&.payments_frozen?
+    end
+
+    def transfer_from_payments_frozen_cohort
+      unless participant_profile.eligible_to_change_cohort_and_continue_training?
```
cwrw, line 214: same here around adding the cohort, and aligning with the other pr (once merged)

app/services/induction/amend_participant_cohort.rb
```diff
@@ -147,11 +167,9 @@ def in_target_schedule?(induction_record)
 
     def participant_declarations
       return false unless participant_profile
+      return @participant_declarations if instance_variable_defined?(:@participant_declarations)
```
cwrw, line 170: not sure I understand why this change was required?

cwrw (commented):
thanks @ltello very minor comments/questions but generally looks great , and after @ethax-ross pr is merged we can align the methods :)

app/services/induction/amend_participant_cohort.rb
ltello: You mean the format change below to put it all in a single line?
I don't mind to revert it back to multiline. We have changed this method so many times back and forth during the live of this PR that it has ended up like this.

app/services/induction/amend_participant_cohort.rb
ltello: Sure. Waiting for him to merge first.

app/services/induction/amend_participant_cohort.rb
ltello: Sure. Waiting for him to merge first.

app/validators/participant_profile_active_validator.rb
ltello: In the question we made to Nathan about setting training_status: :active when executing the cohort change via this tool we finally concluded to leave training_status as it was originally.
So why required the participants to be training_status: :active?

Also it feels to me that the original training status refers to that training period. Why would that block starting a new training period somewhere else?

In any case, happy to reinstate it if needed

app/services/induction/amend_participant_cohort.rb
ltello: The memoization doesn't work when the value is false

app/validators/participant_profile_active_validator.rb
cwrw: having a think maybe we can keep it as it was currently? as I think resetting the training status is quite different than the validation itself. Just erroring on the side of caution as it's already there, if we later find it an issue maybe we can remove it?

config/locales/en.yml
```diff
@@ -319,18 +321,22 @@ en:
...
               not_active: "Not active"
+              not_eligible_to_be_transferred_back: "Not eligible to be transferred back to their original cohort"
+              not_eligible_to_be_transferred_from_current_cohort: "Not eligible to be transferred from their current cohort"
             source_cohort_start_year:
               invalid: "Invalid value. Must be an integer between %{start} and %{end}"
             target_cohort_start_year:
               excluded: "Invalid value. Must be different to %{year}"
               incompatible_with_schedule: "Invalid value. Must match that of the schedule"
               invalid: "Invalid value. Must be an integer between %{start} and %{end}"
+              payments_frozen: "Not permitted. The cohort is frozen for payments"
```
cwrw, line 339: ✂️

config/locales/en.yml
```diff
@@ -203,6 +203,8 @@ en:
     participant_profile:
       blank: "Not registered"
       not_active: "Not active"
+      not_a_participant_profile: "Not a participant profile record"
+      training_complete: "The participant has completed their ECF training"
```
cwrw, line 207: ✂️

cwrw (approved):
thanks @ltello 🙌

config/locales/en.yml
ltello: Thank you. Remainings from previous versions of the PR.

config/locales/en.yml
ltello: Thank you. Remainings from previous versions of the PR.

ltello: `not_a_participant_profile: "Not a participant profile record"` is still used by the other validator

PR #4838, opened 2024-05-23
Relax validation on ChangeSchedule service for certain participants 

ethax-ross:
### Context

We want to allow ECTs and mentors to change schedule/cohort even if they have billable declarations under certain scenarios.

The participants eligible are identified by `eligible_to_change_cohort_and_continue_training`.

Participants that change cohort using this mechanism should be able to change back to their original cohort providing no billable declarations were created against the new cohort.

- Relax validation on ChangeSchedule service for certain participants 

By default the service will not allow eligible participants to transfer; it needs to be called explicitly with `attempt_to_transfer_leaving_billable_declarations` set to `true`. This is to prevent lead providers performing these changes for now.

### Guidance to review

A participant will be able to transfer if:

- `attempt_to_transfer_leaving_billable_declarations` is `true`
- their participant profile is `eligible_to_change_cohort_and_continue_training?`
- the cohort they are transferring to is the active registration cohort
- their current cohort has payments frozen

On transferring, the `cohort_changed_after_payments_frozen` will be `true`.

A participant will be able to transfer _back_ if:

- `attempt_to_transfer_leaving_billable_declarations` is `true`
- they have no billable declarations on the cohort they are leaving
- `cohort_changed_after_payments_frozen` is true on the profile
- they have billable declarations in the profile they are transferring back to
- the cohort they are transferring back to is `payments_frozen`

On transferring back, the `cohort_changed_after_payments_frozen` will be `false`.

### DIFF
--- a/app/models/participant_profile/ecf.rb
+++ b/app/models/participant_profile/ecf.rb
@@ -72,10 +72,10 @@ class ParticipantProfile::ECF < ParticipantProfile
     query
   end
 
-  def eligible_to_change_cohort_back_to_their_payments_frozen_original?(cohort:)
+  def eligible_to_change_cohort_back_to_their_payments_frozen_original?(cohort:, current_cohort:)
     return false unless cohort_changed_after_payments_frozen?
     return false unless cohort.payments_frozen?
-    return false if participant_declarations.billable_or_changeable.where(cohort: schedule.cohort).exists?
+    return false if participant_declarations.billable_or_changeable.where(cohort: current_cohort).exists?
 
     participant_declarations.billable.where(cohort:).exists?
   end
--- a/app/models/participant_profile/ecf.rb
+++ b/app/models/participant_profile/ecf.rb
@@ -3,6 +3,7 @@
 class ParticipantProfile::ECF < ParticipantProfile
   # self.ignored_columns = %i[school_id]
 
+  CHANGE_COHORT_AND_CONTINUE_TRAINING_DELTA = 3
   POST_TRANSITIONAL_INDUCTION_START_DATE_DEADLINE = ActiveSupport::TimeZone["London"].local(2021, 9, 1).freeze
   VALID_EVIDENCE_HELD = %w[training-event-attended self-study-material-completed other].freeze
   COURSE_IDENTIFIERS = %w[ecf-mentor ecf-induction].freeze
@@ -53,7 +54,7 @@ class ParticipantProfile::ECF < ParticipantProfile
     %w[cohort participant_identity school user teacher_profile induction_records]
   end
 
-  def self.eligible_to_change_cohort_and_continue_training(restrict_to_participant_ids: [])
+  def self.eligible_to_change_cohort_and_continue_training(in_cohort_start_year:, restrict_to_participant_ids: [])
     billable_states = %w[eligible payable paid].freeze
 
     completed_billable_declarations = ParticipantDeclaration.billable.for_declaration(:completed)
@@ -61,6 +62,7 @@ class ParticipantProfile::ECF < ParticipantProfile
 
     query = joins(:participant_declarations, schedule: :cohort)
       .where.not(cohorts: { payments_frozen_at: nil })
+      .where(cohorts: { start_year: in_cohort_start_year - CHANGE_COHORT_AND_CONTINUE_TRAINING_DELTA })
       .where("participant_declarations.state IN (?) AND declaration_type != ?", billable_states, "completed")
       .where.not(id: completed_billable_declarations.select(:participant_profile_id))
       .distinct
@@ -68,6 +70,17 @@ class ParticipantProfile::ECF < ParticipantProfile
     query = query.where(id: restrict_to_participant_ids) if restrict_to_participant_ids.any?
 
     query
+  end
+
+  def eligible_to_change_cohort_back_to_their_payments_frozen_original?(to_cohort_start_year:)
+    to_cohort = Cohort.find_by_start_year(to_cohort_start_year)
+
+    return false unless schedule.cohort.start_year - to_cohort_start_year == CHANGE_COHORT_AND_CONTINUE_TRAINING_DELTA
+    return false unless cohort_changed_after_payments_frozen?
+    return false unless to_cohort.payments_frozen?
+    return false if participant_declarations.billable_or_changeable.where(cohort: schedule.cohort).exists?
+
+    true
   end
 
   def self.archivable(restrict_to_participant_ids: [])
@@ -92,8 +105,8 @@ class ParticipantProfile::ECF < ParticipantProfile
   end
 
   # Instance Methods
-  def eligible_to_change_cohort_and_continue_training?
-    self.class.eligible_to_change_cohort_and_continue_training(restrict_to_participant_ids: [id]).exists?
+  def eligible_to_change_cohort_and_continue_training?(in_cohort_start_year:)
+    self.class.eligible_to_change_cohort_and_continue_training(in_cohort_start_year:, restrict_to_participant_ids: [id]).exists?
   end
 
   def archivable?
--- a/app/models/participant_profile/ecf.rb
+++ b/app/models/participant_profile/ecf.rb
@@ -53,10 +53,10 @@ class ParticipantProfile::ECF < ParticipantProfile
     %w[cohort participant_identity school user teacher_profile induction_records]
   end
 
-  def self.eligible_to_change_cohort_and_continue_training(in_cohort_start_year:, restrict_to_participant_ids: [])
+  def self.eligible_to_change_cohort_and_continue_training(cohort:, restrict_to_participant_ids: [])
     billable_states = %w[eligible payable paid].freeze
 
-    return none unless Cohort.find_by_start_year(in_cohort_start_year) == Cohort.active_registration_cohort
+    return none unless cohort == Cohort.active_registration_cohort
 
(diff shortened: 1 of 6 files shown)


app/services/change_schedule.rb
```diff
@@ -9,6 +9,7 @@ class ChangeSchedule
   attribute :course_identifier
   attribute :schedule_identifier
   attribute :cohort, :integer
+  attribute :attempt_to_change_cohort_leaving_billable_declarations, :boolean, default: false
```
ltello, line 12: Do we need this security attribute? 
Nothing will change until we set a cohort as payments-frozen in the database, no?

app/models/participant_profile/ecf.rb
```diff
@@ -83,7 +92,7 @@ def self.archivable(for_cohort_start_year:, restrict_to_participant_ids: [])
     query = left_joins(:participant_declarations, schedule: :cohort)
       # Exclude participants that have any induction records that are not FIP.
       .where.not(id: not_fip_induction_records.select(:participant_profile_id))
```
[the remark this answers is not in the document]
ethax-ross: The requirements for the archive logic are at the bottom of [Nathan's word doc](https://educationgovuk.sharepoint.com.mcas.ms/:w:/s/TeacherServices/EYgnylIh7ctAtknjsx6X9rIBz-P3i9isE4anE6GnWhl5Lw?e=EpC3uQ), he's also included this explanation for those undeclared participants that will not be archived:

> This means that there will be some undeclared participants who are not archived. A tiny number of these participants may require FIP/CIP training in the 2024 cohort (estimate 10s not 100s). It may be preferable for these participants to be transferred manually by DfE admin users if flagged by providers/helpdesk.

app/services/change_schedule.rb
ethax-ross: We need this because the `ChangeSchedule` service is exposed via the lead provider API and we don't necessarily want them to be able to use it to perform this kind of cohort change yet

app/models/participant_profile/mentor.rb
```diff
@@ -21,8 +21,8 @@ class ParticipantProfile::Mentor < ParticipantProfile::ECF
     started_not_completed: "started_not_completed",
   }
 
-  def self.archivable(for_cohort_start_year:, restrict_to_participant_ids: [])
-    super(for_cohort_start_year:, restrict_to_participant_ids:)
+  def self.archivable(restrict_to_participant_ids: [])
+    super(restrict_to_participant_ids:)
       .where(mentor_completion_date: nil)
       .where.not(id: InductionRecord.where.not(mentor_profile_id: nil).select(:mentor_profile_id).distinct)
```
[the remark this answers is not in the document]
ethax-ross: I believe so, here's the full spec:

```
- 2021 cohort (according to their schedule -> cohort) 
- Profile has no declarations where declaration state is one of eligible/payable/paid/awaiting_clawback/clawed_back 
- Participant has only FIP induction records 
- Mentor funding end date is NULL 
- No mentees in induction records
```

app/models/participant_profile/ect.rb
```diff
@@ -10,12 +10,10 @@ class ParticipantProfile::ECT < ParticipantProfile::ECF
...
 
-  def self.archivable(for_cohort_start_year:, restrict_to_participant_ids: [])
-    latest_induction_start_date = Date.new(for_cohort_start_year, 9, 1)
-
-    super(for_cohort_start_year:, restrict_to_participant_ids:)
+  def self.archivable(restrict_to_participant_ids: [])
+    super(restrict_to_participant_ids:)
       .where(induction_completion_date: nil)
-      .where("induction_start_date IS NULL OR induction_start_date < ?", latest_induction_start_date)
+      .where("induction_start_date IS NULL OR induction_start_date < make_date(cohorts.start_year, 9, 1)")
```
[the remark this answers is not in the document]
ethax-ross: This was a requirement from Nathan, I'm not sure of the reasoning behind it but I'll follow up with him to check:

> I think if they have an induction start date of 1/9/2021 or later, we should exclude them and not archive them. We should only archive those where there is no induction start date or it is earlier then 1/9/2021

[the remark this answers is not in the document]
ethax-ross: here's the reasoning:

> Archiving is primarily intended for participants who were registered in error. The two typical cases of this are a) where the ECT wasn't actually an ECT but was instead starting their initial teacher training and b) the ECT had already started induction before the current funding and framework kicked in.

> Where an ECT did actually start induction on or after 1 Sep 2021, the training needs are trumped by the school funding needs. The record will have been used to inform school payments, so we should not archive any such records

app/models/participant_profile/ecf.rb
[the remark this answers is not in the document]
ltello: Ok, if he wants to transferred them manually when flagged by providers/helpdesk, I am happy to leave the code as it is now.

app/models/participant_profile/mentor.rb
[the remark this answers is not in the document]
ltello: That 5th condition (No mentees in induction records) is not in Nathan's doc. (At least I can't find it).
In any case, can it mean the mentor is not CURRENTLY assigned to any mentee?

app/services/change_schedule.rb
ltello: That's fine but if we don't set 2021 as payments-frozen yet, even if they try that cohort change the code we have written won't  allow it (even if we remove `attempt_to_change_cohort_leaving_billable_declarations`)
In any case, fine by me if you want to keep it for extra security reasons.

app/models/participant_profile/mentor.rb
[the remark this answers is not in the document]
ethax-ross: Yeah it looks like that requirement hasn't made it into the Word doc for some reason, but this was the ask from Nathan:

> Could we add a criterion that the mentor has no links to any ECT in any of their induction period records?

app/services/change_schedule.rb
ethax-ross: Yeah I understand that it won't let it happen until the payments frozen is switched on, but we want this as an extra safety net as it looks like we may not want to allow providers to use the service for this purpose (but we want the option to trigger it via other means, so that's where the flag will come in)

app/models/participant_profile/mentor.rb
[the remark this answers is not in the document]
ltello: Ok, so in that case, it is clear that he wants no mentor-mentee at any time.

ltello (approved):

app/services/change_schedule.rb
```diff
@@ -80,6 +81,18 @@ def cohort
...
 
+  def can_change_cohort_leaving_billable_declarations?
+    return false unless participant_profile.ecf?
+    return false unless attempt_to_change_cohort_leaving_billable_declarations
+    return true if participant_profile.eligible_to_change_cohort_and_continue_training?(cohort:)
+
+    participant_profile.eligible_to_change_cohort_back_to_their_payments_frozen_original?(cohort:)
+  end
+
+  def cohort_start_years_delta
```
cwrw, line 92: is this required?

app/services/change_schedule.rb
```diff
@@ -111,6 +124,12 @@ def update_participant_profile_schedule_references!
   end
 
   def update_school_cohort_and_schedule!
+    changing_cohort = new_schedule.cohort != participant_profile.schedule.cohort
```
cwrw, line 127: In the current service (as opposed to the declaration service) we use `relevant_induction_record.schedule.cohort.start_year != cohort&.start_year` the induction record to determine the current cohort, I think we should probably do the same here to maintain consistency. For the record declaration the schedule on profile is used to avoid providers declaring against older cohorts. Let me know you're thoughts

app/services/change_schedule.rb
cwrw, line 12: qq around participants moving from 2024 to 2021 outside the `attempt_to_change_cohort_leaving_billable_declarations`, do we intend to keep that available, I remember we discussed it but not sure what we landed on?

app/services/change_schedule.rb
```diff
@@ -111,6 +124,12 @@ def update_participant_profile_schedule_references!
...
+
+    if changing_cohort && can_change_cohort_leaving_billable_declarations?
+      participant_profile.toggle!(:cohort_changed_after_payments_frozen)
```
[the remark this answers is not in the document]
ethax-ross: No, good catch I'll change this 👍

app/services/change_schedule.rb
ethax-ross: No, missed this, left over from previous implementation

app/services/change_schedule.rb
ethax-ross: I'm not sure on this yet; Nathan caught up with Mili on this at the end of last week and the upshot was that we may not let providers move participants from 2021 -> 2024 using this mechanism initially and instead it would be something we will explore after registration opens. I think this flag may change to only prevent 2021 -> 2024 and we end up allowing 2024 -> 2021 to be triggered by providers, but its still TBC.

app/services/change_schedule.rb
ethax-ross: Yeah that makes sense, I'll update 👍

cwrw (approved):
Thanks @ethax-ross 💯 this took a lot of 🧠 to get through but very cleanly done

PR #4843, opened 2024-05-24
NPQ capping update the npq csv serialiser to surface funded place

pmanrubia:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3128

### Description

The new funded place field has been added to the NPQ application serialiser but needs adding to the csv NPQ application serialiser


cwrw (commented):
could we add this value to NPQ enrolments csv as well please?

cwrw (approved):

PR #4844, opened 2024-05-24
Pin azurerm version to 3.104.2 as the latest 3.105.0 version made the build to fail

leandroalemao:
### Context

We're fetching the latest version of `azurerm 3.105.0` which is now defaulting `public_network_access_enabled` from false to true.

`public_network_access_enabled = false -> true`
https://github.com/DFE-Digital/early-careers-framework/actions/runs/9222398909/job/25374325095

We need to pin/fix the version to the previous version `3.104.2` for now.

Ideally we'd need to set the variable as false in module https://github.com/DFE-Digital/terraform-modules/blob/main/aks/postgres/resources.tf#L39 but that'd cause many projects to get upgraded.

### DIFF
--- a/terraform/application/terraform.tf
+++ b/terraform/application/terraform.tf
@@ -4,6 +4,10 @@ terraform {
   backend "azurerm" {}
 
   required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "3.104.2"
+    }
     statuscake = {
       source  = "StatusCakeDev/statuscake"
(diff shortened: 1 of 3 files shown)


ethax-ross:
Would setting the variable in our project here not work?

https://github.com/DFE-Digital/npq-registration/blob/main/terraform/application/database.tf#L28

```
azurerm_postgresql_flexible_server = false
```

ethax-ross (approved):

PR #4851, opened 2024-05-24
[CST-2564] Show participants from 2021 onward in the school dashboard

leoapost:
### Context

- Ticket: CST-2564

We need to allow sits to manage participants from 2021 cohort onward, but keep displaying up to the latest three cohorts that are not payment frozen in the school cohort tabs.

_\* This is part of the broader changes we are introducing to enable participants from closed cohorts to continue their training in later cohorts._

- Fetch all school cohorts from 2021 onward when visiting the school dashboard
- Display a maximum of three tabs for academic years without payment freeze


app/helpers/dashboard_helper.rb
```diff
@@ -23,4 +23,8 @@ def latest_email_failed_for(profile)
   def participants_active?(profiles)
     profiles.eligible.any? || profiles.contacted_for_info.any? || profiles.details_being_checked.any? || profiles.no_qts_participants.any?
   end
+
+  def get_three_school_cohorts(school_cohorts)
+    school_cohorts.take(3)
```
ltello, line 28: I wonder if this should be no earlier than Cohort.active_registration_cohort - 2 years?

I mean, if a school didn't set the cohort in any of the last 3 registration years we only show the ones setup and we don't pick school_cohorts earlier than 2022?

leoapost: @ltello It might make sense to display only the "running" cohorts, and it's worth bringing this to the product team's attention. 

For now, that would be outside the scope of this PR.

leoapost: The task specs were updated to display up to 3 tabs for academic years without payment freeze. I've pushed the changes.

ltello (approved):
LGTM

PR #4855, opened 2024-05-28
Handle frozen payments for 2021 Cohort

steventux:
### Context

We intend to freeze payments on the 2021 cohort which necessitates a bunch of changes to handle participants who are currently assigned to this cohort but will continue with their training.

As part of this feature we also automatically assign certain participants when the registration period has begun for the next cohort. This changes the wizard flow to remove the step which asks what the Start Term is for the participant being added.
This change applies when a participant has no induction start date from DQT, or is a mentor, and the current date is within the registration period.

- Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CST-2580

- Renames `needs_to_confirm_start_term?` method to reflect that specific participants will be automatically assigned to the next cohort.
- Removes `:start_term` step from add participant journey and merges in wizard logic which determines whether the registration period is open or not as this was previously calculated in the `:start_term` step.
-  Participants with DQT induction start date in a payments-frozen cohort are placed in the currently open for registration. 
- Adds a couple of features to cover the amended wizard journey


app/forms/schools/add_participants/base_wizard.rb
```diff
@@ -473,14 +467,11 @@ def cohort_to_place_participant
...
-        elsif start_term.in? %w[autumn spring]
-          # we're in the registration window prior to 1/9 and chose autumn or spring the following year
+        elsif induction_start_date.blank?
           Cohort.next
         else
-          # default to now - but should ask the start_term question if not already asked
+          # default to now
           Cohort.current
         end
       end
```
ltello, line 477: I would go for:
```ruby
      def cohort_to_place_participant
        if transfer?
          existing_participant_cohort || existing_participant_profile&.schedule&.cohort
        elsif ect_participant? && induction_start_date.present?
          Cohort.containing_date(induction_start_date).tap do |cohort|
            return Cohort.current if cohort.blank? || cohort.npq_plus_one_or_earlier?
          end
        elsif Cohort.within_automatic_assignment_period?
          Cohort.current
        else
          Cohort.next
        end
      end
```

app/forms/schools/add_participants/wizard_steps/email_step.rb
```diff
@@ -31,8 +31,6 @@ def next_step
             end
           elsif wizard.sit_adding_themself_as_mentor?
             :yourself
-          elsif wizard.needs_to_confirm_start_term?
-            :start_term
           elsif wizard.needs_to_choose_a_mentor?
```
ltello, line 34:
```suggestion
          elsif wizard.needs_to_confirm_start_term? && !Cohort.within_next_registration_period?
             :cannot_add_registration_not_yet_open
          elsif wizard.needs_to_confirm_start_term? && wizard.need_training_setup?
             :need_training_setup
          elsif wizard.needs_to_choose_a_mentor?
```

app/forms/schools/add_participants/wizard_steps/who_to_add_participant_checks.rb
```diff
@@ -58,11 +58,7 @@ def transfer_step_for_type
         end
 
         def cohort_checks
-          if wizard.needs_to_confirm_start_term?
-            # we're not sure at this point which cohort is needed
-            # so allow to proceed at this point
-            :none
-          elsif !wizard.registration_open_for_participant_cohort?
+          if !wizard.registration_open_for_participant_cohort?
```
ltello, line 61: I would not delete this

app/forms/schools/add_participants/wizard_steps/yourself_step.rb
```diff
@@ -15,8 +15,6 @@ def next_step
             else
               :cannot_add_manual_transfer
             end
-          elsif wizard.needs_to_confirm_start_term?
-            :start_term
           elsif wizard.needs_to_choose_a_mentor?
```
ltello, line 18:
```suggestion
          elsif wizard.needs_to_confirm_start_term? && !Cohort.within_next_registration_period?
            :cannot_add_registration_not_yet_open
          elsif wizard.needs_to_confirm_start_term? && wizard.need_training_setup?
            :need_training_setup
          elsif wizard.needs_to_choose_a_mentor?
```

app/forms/schools/add_participants/base_wizard.rb
```diff
@@ -327,12 +327,6 @@ def same_delivery_partner?
         false
       end
 
-      def needs_to_confirm_start_term?
-        # are we in the next registration period (or pre-registration period) and the participant does not have
-        # an induction start date
-        (mentor_participant? || induction_start_date.blank?) && !Cohort.within_automatic_assignment_period?
-      end
-
       ## appropriate bodies
```
ltello, line 330: I think we need to keep this method.
Maybe renaming it.
See below to uses

app/forms/schools/add_participants/wizard_steps/email_step.rb
ltello: This is to keep the logic of steps that would go after the one we have removed.

app/forms/schools/add_participants/wizard_steps/who_to_add_participant_checks.rb
ltello: I think we still need the logic behind `wizard.needs_to_confirm_start_term?`

app/forms/schools/add_participants/wizard_steps/yourself_step.rb
ltello: Same as above

app/forms/schools/add_participants/wizard_steps/email_step.rb
steventux: Ah I see, there's a case where we aren't in the registration period yet so we need to respond with the `:cannot_add_registration_not_yet_open` step, but this doesn't have anything to do with the `needs_to_confirm_start_term?` method does it? Given two of the conditional clauses in the `needs_to_confirm_...` method are redundant now ie. `(mentor_participant? || induction_start_date.blank?)` we can combine like so here:

```ruby
if !Cohort.within_next_registration_period? && !Cohort.within_automatic_assignment_period?
  :cannot_add_registration_not_yet_open
```

app/forms/schools/add_participants/base_wizard.rb
steventux: Are there any edge cases where changing the fall-through `else` clause to `Cohort.next` would produce the wrong outcome? I agree it's more succinct this way but I'm nervous that the else clause changes.

app/forms/schools/add_participants/wizard_steps/who_to_add_participant_checks.rb
steventux: The only remaining logic in this method is `!Cohort.within_automatic_assignment_period?` given we no longer apply the clauses `(mentor_participant? || induction_start_date.blank?)` so I've re-added this as 

```ruby
if !Cohort.within_automatic_assignment_period?
  :none
```

app/forms/schools/add_participants/wizard_steps/yourself_step.rb
steventux: I don't understand why we have added logic and 2 different steps here - could you explain please?

app/forms/schools/add_participants/base_wizard.rb
ltello: There is a lot of casuistic about participants and their status at the time of registering them for the first time or transferring them to a new school. 

This ticket is about (at the end of the period of one academic year) sitting participants "potentially compatible" with 2 cohorts (the one about to finish and the new one about to start) into the latest cohort (Cohort.next) so I think the logic I suggest is ok.
But you know, there is always the chance that it is not 100% perfect. 
In any case, if we haven't coped with any specific situation we will know in the testing period or maybe via a support ticket and we'll have to fix it later.

spec/features/schools/participants/add_participants/sit_add_ect_spec.rb
```diff
@@ -7,20 +7,18 @@
...
+    and_i_am_signed_in_as_an_induction_coordinator
+    and_i_click_on(Cohort.current.description)
+    then_i_am_taken_to_fip_induction_dashboard
+
+    set_dqt_validation_result
   end
 
   scenario "SIT aborts process as it can't add themself as ECT" do
-    outside_auto_assignment_window do
+    inside_registration_window do
```
steventux, line 21: @ltello I'm concerned that a handful of feature specs now need to be inside the registration window and have the next cohort set up otherwise they end up with the _registration not yet open_ step and fail. Does this seem right?

app/forms/schools/add_participants/base_wizard.rb
steventux: Much cleaner as suggested, I've amended with this.

spec/features/schools/participants/add_participants/sit_add_ect_spec.rb
ltello: why?

steventux: @ltello  For the most part, the failing specs dealt with adding mentors.

If the specs are left in the timeframe of `outside_auto_assignment_window` we end up at the _registration not yet open step_. So the expectations of the tests fail as the user effectively ends on a terminating step and doesn't proceed through the wizard journey. 
This is because we are not inside the registration period for the next cohort and in most cases we are dealing with a mentor participant. This is consistent with the changes made I think.

If the specs are put in the timeframe of `inside_auto_assignment_window` most of them pass as the time sensitive clause in `BaseWizard#automatically_assign_next_cohort?` doesn't apply.

If the specs are put in the timeframe of `inside_registration_window` and the setup is done for the next cohort, then they will end up in the next cohort.

ltello: No worries @steventux.
Cover the more cases you can and let's move forward. We need this done to continue with other stuff.
Also we'll let Matt to manually test as many scenarios as he can

app/forms/schools/add_participants/wizard_steps/email_step.rb
steventux: Have renamed this method and call as suggested to preserve routing to the terminal _registration not yet open_ step

app/forms/schools/add_participants/wizard_steps/yourself_step.rb
steventux: Thanks for clarifying, this has been amended to allow both outcomes to be visited, I think this is where we perhaps need additional test coverage.

spec/features/schools/participants/add_participants/sit_add_ect_spec.rb
steventux: I think there's a couple of scenarios I can add which would cover arriving at the _registration not yet open_ step and the _need training setup_ step. I'm not sure it's necessary for all the specs which have changed but it would be good to test that these steps are visited and not bypassed completely. Will check.

steventux: @ltello I've added https://github.com/DFE-Digital/early-careers-framework/pull/4855/commits/63bfb29e6b0eba0895ea6882f03e5069993544ae and https://github.com/DFE-Digital/early-careers-framework/pull/4855/commits/509fa05dd1f332933630d7375807823f1959fd25 in an attempt to cover routing to the two other steps. Not 100% sure if the spec setup for _training setup needed_ is actually representative.

ltello (approved):

steventux:
@ltello Should we turn this PR back into a draft while it is being actively developed on the back of manual testing? 
Also I think the initial approval is very outdated now that we've added so many other commits. I think we should re-review as a team given the amount of changes here.

app/services/induction/amend_cohort_after_eligibility_checks.rb
```diff
@@ -0,0 +1,32 @@
...
+                                            source_cohort_start_year: participant_profile.schedule&.cohort&.start_year,
+                                            target_cohort_start_year: Cohort.active_registration_cohort.start_year)
+                                       .save
+    end
+
+    def eligible_participant_in_frozen_cohort?
+      return false unless participant_profile.eligible?
+      return false if participant_profile.ecf_participant_eligibility&.reason != "none"
+
+      participant_profile.schedule&.cohort&.payments_frozen?
```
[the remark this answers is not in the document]
ltello: This just needs to check if we should try and change the cohort of the participant (has become eligible for funding after re-validating it and is sitting in a payments_frozen cohort).

Then the tool to try and amend the cohort needs to do all the validations before doing it. 
Those validations include the one you state + others.

app/models/cohort.rb
```diff
@@ -16,6 +16,7 @@ class Cohort < ApplicationRecord
   scope :between_years, ->(lower, upper) { where(start_year: lower..upper) }
   scope :between_2021_and, ->(upper) { between_years(2021, upper) }
   scope :ordered_by_start_year, -> { order(start_year: :asc) }
+  scope :payments_frozen, -> { where.not(payments_frozen_at: nil) }
```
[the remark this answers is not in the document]
ltello: We better remove this scope because it was going to be used in the first version of these changes but finally not.

app/models/participant_profile/mentor.rb
```diff
@@ -49,6 +49,12 @@ def role
     "Mentor"
   end
 
+  def with_completion_date_status_or_declaration?
```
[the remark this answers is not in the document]
ltello: I'll add specs for this.

app/helpers/schools/early_career_teachers_helper.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Schools
+  module EarlyCareerTeachersHelper
+    def completed_participants(participants)
+      participants.sort_by { |participant| participant.induction_completion_date || Date.new(2000, 1, 1) }
```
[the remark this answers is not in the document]
ltello: done

PR #4858, opened 2024-05-29
Flag multi cohort participants to providers

ethax-ross:
[Jira-3145](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3145)

### Context

When a participant is migrated from a payments frozen cohort, we want to expose this to lead providers in the participant response so that they can identify them.

- Expose previous, frozen cohort to lead providers

Add `previous_payments_frozen_cohort` field to participant serializers that will be `nil` if the participant has not migrated cohort due to payments being frozen and will be the cohort start year of the frozen cohort if they have been migrated.

- Update ParticipantPresenter to use previous_payments_frozen_cohort

This was added for the serializers and we may as well use it here as well. It also contains an extra check to make sure the previous cohort is frozen, in case the participant moved cohort using the normal schedule change and then migrated cohort due to payments being frozen.

- Add freeze_payments! convenience method to Cohort

Add a convenience method to freeze payments on a cohort and also add a trait to the cohorts factory for creating frozen cohorts.

- Wrap previous_payments_frozen_cohort in feature flag

We only want to display this in sandbox initially.

- Change cohort_changed_after_payments_frozen flag to boolean

Change the flag to indicate a participant migrated cohorts due to payments being frozen on their original cohort to a boolean value for better performance. Rename to `cohort_changed_after_payments_frozen`.

### Guidance to review

This will be enabled in the `migration` environment initially and without documentation/release notes.

### DIFF
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -99,6 +99,10 @@ class Cohort < ApplicationRecord
     self.class.find_by(start_year: start_year - 1)
   end
 
+  def freeze_payments!
+    update!(payments_frozen_at: Time.zone.now)
+  end
+
   # e.g. "2022"
   def to_param
     start_year.to_s
--- a/app/models/participant_profile/ecf.rb
+++ b/app/models/participant_profile/ecf.rb
@@ -110,6 +110,16 @@ class ParticipantProfile::ECF < ParticipantProfile
     self.class.archivable(restrict_to_participant_ids: [id]).exists?
   end
 
+  def previous_payments_frozen_cohort
+    return nil unless cohort_changed_after_payments_frozen?
+
+    participant_declarations
+      .includes(:cohort)
+      .where.not(cohort: { payments_frozen_at: nil })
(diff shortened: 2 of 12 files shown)


cwrw (commented):
Thanks @ethax-ross there are a few missing responses from withdraw/defer actions, could we add them there?

swagger/v3/api_spec.json
```diff
@@ -161,12 +161,12 @@
         }
       }
     },
-    "/api/v3/partnerships/ecf": {
+    "/api/v3/participants/ecf/transfers": {
```
cwrw, line 164: ah my bad we can't run swaggerise on v3 as it messes the order, we have to add manually sorry 😓 v1 and v2 we can run but not on v3

cwrw (approved):
🙌

cwrw (approved):

PR #4859, opened 2024-05-29
Write use of `method_missing` in page objects helper to file

steventux:
### Context

We get a lot of `puts` messages about calls to missing page object methods which write to stdout.

- Ticket: 

Instead of filling stdout with messages about these missing page object method calls, log them in a file per RSpec suite run.


peteryates (approved):
Much nicer.

PR #4861, opened 2024-05-29
CST-2558: closing 2021 - add ect journey - take into account 2021 closing in cohort determination logic

ltello:
### Context

As part of the journey that allows a SIT to add a new ECT to their school there is some logic behind the scenes to determine and sit the participant in a provisional/temporary cohort.

We need to extend that logic to sit in 2024/25 cohort a potential new participant whose DQT induction start date at the time of registration would sit the participant in 2021/22 cohort.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2558):


steventux (approved):
It would be good to add a feature spec scenario for payments frozen but given the simplicity of this change I think that can be done later as part of the wider test coverage.

PR #4862, opened 2024-05-30
CST-2559: closing 2021 - transfer ect journey - take into account 2021 closing in cohort determination logic

ltello:
### Context

As part of the journey that allows a SIT to transfer an (already registered at a different school) ECT to their school there is some logic behind the scenes that keeps the ECT in the cohort they are already sitting.

We need to extend that logic to sit in 2024/25 cohort a transferring ECT currently sitting in 2021/22 cohort and not having completed induction.

Those with completed induction will be transferred school (not cohort) but there status should not be set to active but completed.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2559):

### DIFF
--- a/app/forms/schools/add_participants/base_wizard.rb
+++ b/app/forms/schools/add_participants/base_wizard.rb
@@ -463,18 +463,24 @@ module Schools
 
       # NOTE: not preventing registration here just determining where to put the participant
       def cohort_to_place_participant
-        if transfer?
-          existing_participant_cohort || existing_participant_profile&.schedule&.cohort
-        elsif ect_participant? && induction_start_date.present?
-          Cohort.for_induction_start_date(induction_start_date).tap do |cohort|
-            return Cohort.current if cohort.blank? || cohort.npq_plus_one_or_earlier?
-            return Cohort.active_registration_cohort if cohort.payments_frozen?
-          end
-        elsif Cohort.within_automatic_assignment_period?
-          Cohort.current
-        else
-          Cohort.next
+        return cohort_for_transfer if transfer?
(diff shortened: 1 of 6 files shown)


app/models/participant_profile/ecf.rb
```diff
@@ -100,6 +100,12 @@ def archivable?
     self.class.archivable(restrict_to_participant_ids: [id]).exists?
   end
 
+  # def with_completion_date_or_declaration?
```
steventux, line 103: Not sure if this was included intentionally but it's not serving any purpose.

app/forms/schools/add_participants/base_wizard.rb
```diff
@@ -462,19 +463,26 @@ def participant_cohort
...
-            return Cohort.current if cohort.blank? || cohort.npq_plus_one_or_earlier?
-          end
-        elsif Cohort.within_automatic_assignment_period?
-          Cohort.current
-        else
-          Cohort.next
+        return cohort_for_transfer if transfer?
+        return cohort_for_ect_with_induction_start_date if ect_participant? && induction_start_date.present?
+
+        Cohort.within_automatic_assignment_period? ? Cohort.current : Cohort.next
```
steventux, line 469:
```suggestion
        return Cohort.current if Cohort.within_automatic_assignment_period?
        
        Cohort.next
```

steventux: If we are refactoring into short-circuit return statements (I prefer this for readability) perhaps taking the last ternary out is worthwhile for consistency.

steventux (approved):
A couple of minor comments, nothing blocking

steventux (approved):

PR #4863, opened 2024-05-30
Validate eligibility value for funded places

padv2010:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3078

### Description

Update the admin tool in the NPQ tab so it allows updating the eligible for funding value, where we should return an error if the funded place is true and we attempt to set the eligible for funding to false.

### DIFF
--- a/spec/models/npq_application_spec.rb
+++ b/spec/models/npq_application_spec.rb
@@ -353,10 +353,36 @@ RSpec.describe NPQApplication, type: :model do
   end
 
   describe "validations" do
-    context "when validating funding eligibility" do
-      let(:participant_declaration) { create(:npq_participant_declaration, state: participant_declaration_state) }
-      let(:npq_application) { participant_declaration.participant_profile.npq_application }
+    let(:participant_declaration) { create(:npq_participant_declaration, state: participant_declaration_state) }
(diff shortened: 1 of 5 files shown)


cwrw (approved):

spec/models/npq_application_spec.rb
```diff
@@ -353,10 +353,36 @@ def subject.transient_previously_funded
...
+        expect(npq_application.valid?).to eq(true)
+      end
+
+      it "is valid if `eligible_for_funding` is false and `funded_ place` is false" do
+        npq_application.eligible_for_funding = false
+        npq_application.funded_place = false
 
+        expect(npq_application.valid?(:admin)).to eq(true)
+      end
+      it "is valid if `eligible_for_funding` is false and `funded_ place` is nil" do
```
cwrw, line 377: could we add a test for true/true, and true/nil?

padv2010: Addressed in https://github.com/DFE-Digital/early-careers-framework/pull/4863/commits/f5745404f8659707312308a12abfe78ce453e83a

cwrw (approved):

PR #4871, opened 2024-05-31
[CPLD-3077] Update Eligibility logic to include funded place

cwrw:
### Context

We've introduced funded place to funding eligibility for NPQ from 2024 going forward. Previously we only calculated eligibility for declarations using `eligible_for_funding`. This looked into previously funded applications with the field set to true or the field itself.

We now have to account for `funded_place` as well. If it's not set we default to the `eligible_for_funding`. If it's set, then it is the deciding factor if the application is eligible for funding.

`eligible_for_dfe_funding` is used in the API and the record declaration logic, so we have to distinguish when we want to factor in the funded place logic as well. However previously funded is always included.

- Ticket: https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3077

- Amend the `eligible_for_dfe_funding` method to include funded place logic if the flag is set.
- Amend previously funded logic to include funded place logic always, across both the API and funding eligibility service which feeds back to the NPQ app
- Amend the `fundable` method to include funded place logic so declarations now rely on `funded_place` as well
- Amend the funding eligibility service to include previously funded logic as this feeds back to the NPQ app 

### Guidance to review

- commit by commit
- I avoided shared examples in most places there are subtle differences between every context for funded place `nil`/`false`/`true`

### DIFF
--- a/app/models/npq_application.rb
+++ b/app/models/npq_application.rb
@@ -149,6 +149,7 @@ private
       .where.not(id:)
       .where(npq_course: npq_course.rebranded_alternative_courses)
       .where(eligible_for_funding: true)
+      .where(funded_place: [nil, true])
       .accepted
       .exists?
   end
--- a/app/models/npq_application.rb
+++ b/app/models/npq_application.rb
@@ -86,11 +86,11 @@ class NPQApplication < ApplicationRecord
   # eligible_for_funding is solely based on what NPQ app knows
   # eg school, course etc
   # here we need to account for previous enrollments too
-  def eligible_for_dfe_funding
+  def eligible_for_dfe_funding(with_funded_place: false)
     if previously_funded?
       false
     else
-      eligible_for_funding && (funded_place.nil? || funded_place)
+      funding_eligibility(with_funded_place:)
     end
   end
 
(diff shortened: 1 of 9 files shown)


app/models/npq_application.rb
```diff
@@ -148,10 +148,17 @@ def previously_funded?
       .where.not(id:)
       .where(npq_course: npq_course.rebranded_alternative_courses)
       .where(eligible_for_funding: true)
+      .where(funded_place: [nil, true])
```
ethax-ross, line 151: Does this have to be a nullable attribute, can we not default it to `true`? it makes it harder to follow/less readable if we need to check for `nil` and `true` and I imagine its something that could trip us up in the future?

cwrw: discussed in person: good point, it would make life much easier if we did default to true, however old values will be surfaced in the API and will need to be empty. Also we require providers to supply this value so we cannot default it to true and need to distinguish values submitted by them vs defaulted.

padv2010 (approved):
Nice work :+1:

mooktakim (approved):

leandroalemao (approved):
LGTM... just a heads up... think we need to do the same sort of tweaks in NPQ reg app as well? Tks

PR #4872, opened 2024-05-31
CST-2565: move mentors out of frozen cohort when assigned an ect in u…

ltello:
…nfrozen cohort

### Context

If a SIT creates a new link Mentor-ECT we might need to move any of them (ECT or Mentor or both) to 2024. 3 scenarios here:

unfinished 2021 mentor associated to a 2021 ECT (nothing to change)

unfinished 2021 mentor associated to a non-2021 ECT (move the mentor to 2024)

non-2021 mentor associated to a 2021 ECT (we could move the ECT to 2024 but we have decided not to implement it. Other triggers will catch this ECT and move them to 2024)

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2565):

### DIFF
--- a/app/services/induction/change_mentor.rb
+++ b/app/services/induction/change_mentor.rb
@@ -8,7 +8,7 @@ class Induction::ChangeMentor < BaseService
         changes: { mentor_profile: },
       )
       induction_record.participant_profile.update!(mentor_profile:)
-      amend_mentor_cohort if sit_mentor_out_of_frozen_cohort?
+      amend_mentor_cohort if change_mentor_cohort?
       send_training_materials_if_needed
 
       mentor_profile&.user&.touch
@@ -56,7 +56,7 @@ private
      .deliver_later
   end
 
-  def sit_mentor_out_of_frozen_cohort?
+  def change_mentor_cohort?
     return false if ect_in_payments_frozen_cohort?
 
     mentor_profile.eligible_to_change_cohort_and_continue_training?(cohort: Cohort.active_registration_cohort)
--- a/app/services/induction/change_mentor.rb
+++ b/app/services/induction/change_mentor.rb
@@ -8,7 +8,7 @@ class Induction::ChangeMentor < BaseService
         changes: { mentor_profile: },
       )
       induction_record.participant_profile.update!(mentor_profile:)
-
+      amend_mentor_cohort if sit_mentor_out_of_frozen_cohort?
       send_training_materials_if_needed
 
       mentor_profile&.user&.touch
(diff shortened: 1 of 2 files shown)


app/services/induction/change_mentor.rb
```diff
@@ -49,4 +45,24 @@ def send_training_materials_if_needed
...
+    MentorMailer.with(
+      mentor_profile:,
+      ect_name: induction_record.participant_profile.user.full_name,
+      cip_materials_name: cip_materials.name,
+      sit_name:,
+    ).training_materials
+     .deliver_later
+  end
+
+  def sit_mentor_out_of_frozen_cohort?
```
steventux, line 59: The method name doesn't really reflect the implementation as I understand it:

We return `false` if the ECT is in a cohort with payments frozen, otherwise apply the normal "eligible to change" rules?

I think this method is to do with eligibility to amend the cohort, and the payments frozen logic is just one of the conditions.

```suggestion
  def mentor_can_change_cohort?
```

steventux (approved):
I think a method name change would help here but otherwise all good :+1:

app/services/induction/change_mentor.rb
ltello: Not sure I follow.
A mentor like this one would be in a frozen cohort, with billable declarations, not completed declarations... and all the rest of conditions to be called unfinished participant.
This method is to check if we must move them out of that frozen cohort after being assigned an ECT.

ltello: Happy to rename it to anything else though.

steventux: I think there's ambiguity in the prefix `sit_` does this mean _to place_ or a type of Induction Tutor?

The implementation is concerned with the rules of whether a mentor can or should change cohort, only one of the conditionals is concerned with _payments frozen_, there's a further check on `eligible_to_change_cohort_and_continue_training?` which has nothing to do with frozen cohorts, so the more general name I suggested was just an idea of how to make the code more readable and perhaps a more generalised wrapper name around the two internal conditionals.

app/services/induction/change_mentor.rb
```diff
@@ -24,22 +24,18 @@ def initialize(induction_record:, mentor_profile: nil)
...
-  def send_training_materials
-    MentorMailer.with(
-      mentor_profile:,
-      ect_name: induction_record.participant_profile.user.full_name,
-      cip_materials_name: cip_materials.name,
-      sit_name:,
-    ).training_materials
-     .deliver_later
+  def amend_mentor_cohort
+    Induction::AmendParticipantCohort.new(participant_profile: mentor_profile,
```
steventux, line 28: Should we be using this service to amend a mentor profile? If it's generic enough to amend any type of profile or certain types of profile perhaps we should rename it to reflect that?

app/services/induction/change_mentor.rb
ltello: Yes, we usually use the verb sit meaning placing the participant in a cohort.

The first conditional is about the ECT (not the mentor) being in a payments_frozen cohort. The method name is about the mentor (not the ECT).

In any case, shall we use  `change_mentor_cohort?` instead?
If so, feel free to commit my 2 suggestions.

steventux: @ltello I think we should avoid `sit_` as a prefix as it could easily be mistaken for induction tutor in this codebase.

I understand what the method does, my suggestion was that we mask the various conditionals, some of which have little to do with frozen payments from the method name. I'll commit my original suggestion.

PR #4880, opened 2024-06-03
Cst 2567 closing 2021 monday eligibility checks include 2021 closing logic to sit the participant in 2024 cohort

ltello:
### Context
Every Monday we pick participants currently ineligible for funding and check their record in DQT to see if something has changed that might make them eligible. As part of this re-validation of the participant we might determine the cohort where to sit them.

We need to review that logic to put in 2024/25 cohort participants that as part of the re-validation would currently be placed in 2021/22.

Also, as part of this ticket we need to do something similar to the DQT daily checks for induction start dates.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2567):


app/services/induction/review_cohort_after_eligibility_checks.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# Change the cohort of an ECF participant whose eligibility has been recalculated.
+# If the participant has become eligible and is currently in a payments-frozen cohort,
+# transfer them to the currently active registration cohort.
+module Induction
+  class ReviewCohortAfterEligibilityChecks
```
steventux, line 7: This service name doesn't really suggest to me that anything will be updated here. I think we should be more explicit about the fact records are likely to change in this service.
Something like `UpdateCohortAfterEligibilityChecks` or `AmendCohortAfterEligibilityChecks` ?

app/services/participants/check_and_set_completion_date.rb
```diff
@@ -3,10 +3,11 @@
 module Participants
   class CheckAndSetCompletionDate < BaseService
     def call
+      debugger
```
steventux, line 6: This needs to be removed.

steventux (changes requested):
A couple of small changes needed.

app/services/induction/review_cohort_after_eligibility_checks.rb
ltello: Yes, the thing is that the cohort might be or might not be changed. So that's why I didn't use Amend or Update.

ltello: Anyway. Renamed

steventux (approved):
:rocket:

PR #4883, opened 2024-06-03
Add the custom API related fields to log output

peteryates:
### Context

This may or may not be entirely correct, but I _think_ it's doing the right thing. Need help from an @lead-provider-cpd person to make sure.

### Extra tasks

* [x] ~~disable SQL logging in prod:~~ (it's disabled in prod by default already)
   ```ruby
    config.active_record.logger = nil # Don't log SQL in production
    ```

### DIFF
--- a/lib/semantic_logger/formatters/json_with_api_metadata.rb
+++ b/lib/semantic_logger/formatters/json_with_api_metadata.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module SemanticLogger
+  module Formatters
+    class JsonWithApiMetadata < Raw
+      def call(log, logger)
+        super(log, logger).merge(api_metrics).to_json
+      end
+
+    private
+
+      def api_metrics
+        return {} unless payload
+
+        {
+          params: payload[:params],
+          exception: payload[:exception],
+          current_user_class: payload[:current_user_class],
+          current_user_id: payload[:current_user_id],
+        }.compact
+      end
+    end
+  end
+end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "active_support/core_ext/integer/time"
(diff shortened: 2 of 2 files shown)


lib/semantic_logger/formatters/custom.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module SemanticLogger
+  module Formatters
+    class Custom < Raw
```
peteryates, line 5: Yeah, not the best naming here.

cwrw (approved):
thank you for this @peteryates tested locally with the API and works a charm

lib/semantic_logger/formatters/json_with_api_metadata.rb
```diff
@@ -0,0 +1,24 @@
...
+        super(log, logger).merge(api_metrics).to_json
+      end
+
+    private
+
+      def api_metrics
+        return {} unless payload
+
+        {
+          params: payload[:params],
```
cwrw, line 16: maybe we can add the same rule here around not including `exceptions = %w[controller action format id]` similar to the old code: `payload[:params].except(*exceptions)`

lib/semantic_logger/formatters/json_with_api_metadata.rb
```diff
@@ -0,0 +1,24 @@
...
+          exception: payload[:exception],
+          current_user_class: payload[:current_user_class],
```
cwrw, line 18: bah just realised we add this already in the API controller: https://github.com/DFE-Digital/early-careers-framework/blob/main/app/controllers/api/api_controller.rb#L24

As I am seeing it twice in the responses, maybe we drop the current user class/id as its a duplicate and keep the other two?

lib/semantic_logger/formatters/json_with_api_metadata.rb
peteryates: This would affect all the logs though, aren't those things important elsewhere? 

Maybe leave them in for now and if things become messy exclude them based on route later?

PR #4905, opened 2024-06-07
Add X-With-Server-Date header support to migration env

ethax-ross:
We need this to be able to test some of the declaration scenarios.


config/environments/migration.rb
```diff
@@ -5,4 +5,8 @@
 Rails.application.configure do
   config.log_level = :debug
   config.ssl_options = { redirect: { exclude: ->(request) { request.path.include?("/check") } } }
+
+  # Used to handle HTTP_X_WITH_SERVER_DATE header for server side datetime overwrite
+  config.middleware.use TimeTraveler
+  config.middleware.use ApiRequestMiddleware
```
cwrw, line 11: Not sure we need this one as it created `ApiRequest`?

cwrw (approved):

PR #4912, opened 2024-06-10
Fix branch test failures

steventux:
### Context

At some point the Closing 2021 feature branch test suite started failing.

- Ticket: 

Fix failing tests.

### DIFF
--- a/spec/features/schools/participants/add_participants/sit_add_ect_spec.rb
+++ b/spec/features/schools/participants/add_participants/sit_add_ect_spec.rb
@@ -3,7 +3,7 @@
 require "rails_helper"
 require_relative "../../training_dashboard/manage_training_steps"
 
-RSpec.describe "SIT adding themself as ECT", js: true do
+RSpec.describe "SIT adding themself as ECT", js: true, mid_cohort: true do
   include ManageTrainingSteps
 
   before do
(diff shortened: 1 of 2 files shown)


spec/forms/schools/add_participants/wizard_steps/email_step_spec.rb
```diff
@@ -42,6 +42,7 @@
     context "when registration is not yet open" do
       let(:automatically_assign_next_cohort) { true }
       let(:registration_open) { false }
+      before { allow(Cohort).to receive(:within_next_registration_period?).and_return(registration_open) }
```
ltello, line 45: why this solution instead of using `mid_cohort: true` as above?
Isn't this solution implementation dependant?

spec/features/schools/participants/add_participants/sit_add_ect_spec.rb
```diff
@@ -3,7 +3,7 @@
 require "rails_helper"
 require_relative "../../training_dashboard/manage_training_steps"
 
-RSpec.describe "SIT adding themself as ECT", js: true do
+RSpec.describe "SIT adding themself as ECT", js: true, mid_cohort: true do
```
ltello, line 6: The issue was the date we are currently, wasn't it?

spec/forms/schools/add_participants/wizard_steps/email_step_spec.rb
steventux: The convention in this test is to use stubs to create setup conditions, I have followed this.

spec/features/schools/participants/add_participants/sit_add_ect_spec.rb
steventux: I'm not sure I understand what you mean. I've used the same helper we've used in other date dependent fixes here.

ltello: I was just trying to understand and you to confirm that the reason why these test starting to fail was because we are at the beginning of June and the testing cohorts are setup to start registration.
That has affected the tests scenarios basically.

spec/forms/schools/add_participants/wizard_steps/email_step_spec.rb
ltello: Given that we have a proper way (via that helper that travels to the right date) to simulate we are not opening registration yet, I would prefer to use it instead

spec/features/schools/participants/add_participants/sit_add_ect_spec.rb
steventux: Yes I understood this when we added the `mid_cohort` helper to `main` time bomb tests are very common in codebases like this.

spec/forms/schools/add_participants/wizard_steps/email_step_spec.rb
steventux: This is a unit test. The convention in the test is quite rightly to stub conditions for the set up of the test subject which I have followed with this change.
Introducing a date and time distortion via the `mid_cohort: true` or any other time travel mechanism is not only unnecessary but something we should only do in features.

ltello: What is a real distorsion is stubbing a method that depends on the current time and leave the current time unchanged. 
That contradiction has been the cause of these and many other flacky tests during the life of this project.

ltello: Anyway, not the right monent for these sort of discussions.

ltello (approved):

PR #4917, opened 2024-06-11
New way to determine when a cohort has been frozen

ltello:
### Context

To avoid manually doing it the day we open a new registration year, we are going to determine if a Cohort has been frozen or not based on the date store in the database.

So when the time setup comes, the service will automatically detect the new status of the cohort and enable all the dependant features.

  Redefine the logic in `Cohort#payments_frozen?`


ethax-ross (approved):
LGTM

cwrw (approved):

PR #4918, opened 2024-06-11
closing-21-1: remove start term from add-participant journey within next registration window

ltello:
### Context

- Ticket:

### DIFF
--- a/app/forms/schools/add_participants/base_wizard.rb
+++ b/app/forms/schools/add_participants/base_wizard.rb
@@ -7,6 +7,7 @@ module Schools
       include Rails.application.routes.url_helpers
 
       class AlreadyInitialised < StandardError; end
+
       class InvalidStep < StandardError; end
 
       attr_reader :current_step, :submitted_params, :data_store, :current_user, :participant_profile, :school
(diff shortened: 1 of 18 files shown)


app/forms/schools/add_participants/base_wizard.rb
```diff
@@ -464,27 +463,27 @@ def participant_cohort
...
-          # we're in the registration window prior to 1/9 and chose autumn or spring the following year
-          Cohort.next
-        else
-          # default to now - but should ask the start_term question if not already asked
-          Cohort.current
+        return cohort_for_transfer if transfer?
+        return cohort_for_ect_with_induction_start_date if ect_participant? && induction_start_date.present?
+        return Cohort.current if Cohort.within_automatic_assignment_period?
+
+        Cohort.next
```
ethax-ross, line 470: It looks like the default here has changed from `Cohort.current` to `Cohort.next`; is it the case that if the `start_term` is not `autumn` or `spring` then `Cohort.within_automatic_assignment_period?` holds `true`?

ltello: if the `start_term` is not `autumn` or `spring` then `Cohort.within_automatic_assignment_period?` == false

ltello: Actually the logic change this year is to place the participant in the next cohort if they are registered during summer term.

PR #4920, opened 2024-06-11
Closing-21-3: revalidating eligibility

ltello:
### Context

- Ticket:

### DIFF
--- a/app/services/induction/amend_cohort_after_eligibility_checks.rb
+++ b/app/services/induction/amend_cohort_after_eligibility_checks.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+# Change the cohort of an ECF participant whose eligibility has been recalculated.
+# If the participant has become eligible and is currently in a payments-frozen cohort,
+# transfer them to the currently active registration cohort.
+module Induction
+  class AmendCohortAfterEligibilityChecks
+    include ActiveModel::Model
+
+    attr_accessor :participant_profile
+
+    def call
+      amend_participant_cohort if eligible_participant_in_frozen_cohort?
+    end
+
+  private
+
+    def amend_participant_cohort
+      Induction::AmendParticipantCohort.new(participant_profile:,
+                                            source_cohort_start_year: participant_profile.schedule&.cohort&.start_year,
+                                            target_cohort_start_year: Cohort.active_registration_cohort.start_year)
+                                       .save
+    end
+
+    def eligible_participant_in_frozen_cohort?
+      return false unless participant_profile.eligible?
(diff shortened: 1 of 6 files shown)


app/services/induction/amend_cohort_after_eligibility_checks.rb
```diff
@@ -0,0 +1,32 @@
...
+    attr_accessor :participant_profile
+
+    def call
+      amend_participant_cohort if eligible_participant_in_frozen_cohort?
+    end
+
+  private
+
+    def amend_participant_cohort
+      Induction::AmendParticipantCohort.new(participant_profile:,
```
ethax-ross, line 19: `save!` since we're not checking the return value?

app/services/induction/amend_cohort_after_eligibility_checks.rb
```diff
@@ -0,0 +1,32 @@
...
+                                            source_cohort_start_year: participant_profile.schedule&.cohort&.start_year,
+                                            target_cohort_start_year: Cohort.active_registration_cohort.start_year)
+                                       .save
+    end
+
+    def eligible_participant_in_frozen_cohort?
```
ethax-ross, line 25: I'm still not sure I follow why we wouldn't want to call `eligible_to_change_cohort_and_continue_training?(Cohort.active_registration_cohort)` here; could it not be the case that a participant becomes eligible but does not meet the other criteria we check for, such as having billable (but not completed) declarations?

app/services/induction/amend_cohort_after_eligibility_checks.rb
ltello: Well... the code calling this SO actually checks the value.

app/services/induction/amend_cohort_after_eligibility_checks.rb
ltello: So this SO is meant to deal with participants that have just been re-validated and their eligibility re-assessed.
That is the only scope of this class.

What we want to do here is just checking if such a participant is eligible to have their cohort amended because of any potential changes happening to their eligibility:
  - After revalidation and re-eligibility assessment has been left in a frozen cohort,
  - After revalidation and re-eligibility assessment has become eligible for funding.

If so, let's call the tool that amends the cohort.
That tool has its own responsability and have to go through all sort of validations (generic validations not specific to any context == validations only specific to the participant himself and their destination cohort) and do or reject the cohort change. Those validations include the one you mention (`eligible_to_change_cohort_and_continue_training?(Cohort.active_registration_cohort`).

If I check here `eligible_to_change_cohort_and_continue_training?(Cohort.active_registration_cohort`) then we would be checking it twice (because `Induction::AmendParticipantCohort` will check it again).

PR #4924, opened 2024-06-12
Support migrating declarations to paid/payable statements

ethax-ross:
[Jira-3169](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3169)

### Context

We want to migrate a number of declarations `awaiting_clawback` for SLN from one statement to another (payable) statement.

The current declaration migrator service does not support this as it:

- Is restricted to migrating only to future dated statements
- Will change the `output_fee` to `false` for all from statements
- Does not support migrating to paid/payable statements

We want to update the migration service to support the new scenario.

- Support filtering by states in MigrateDeclarationsBetweenStatements

We want the ability to only migrate a subset of declarations with a particular state or states.

- Relax restrictions on MigrateDeclarationsBetweenStatements

Up until now we have restricted the usage to only statements that are future dated. This is too restrictive for a new use case, so we are relaxing it to migrate those that are not future dated (albeit with a warning).

We also want to remove the requirement of having to specify an `output_fee` for the from statement after the migration has completed. If we find we want to modify the from statement again in the future we could re-introduce in a more flexible way, such as with a `from_statement_updates` hash (similar to the `to_statement_updates` already supported).

-  Support migrating declarations to payable/paid statements

When we migrate declarations and their statement line items to a new statement and that statement is paid/payable, we need to ensure that we run the declarations through the appropriate service to get them inline with the new statement.

- Support generic updates to from statement

Similar to how we allow generic attribute updates for the to statements, we want to allow generic updates to the from statements. This can be useful for ensuring the statement declarations are migrated from ends up being `output_fee: false` after the migration, for example.

### DIFF
--- a/app/services/oneoffs/npq/migrate_declarations_between_statements.rb
+++ b/app/services/oneoffs/npq/migrate_declarations_between_statements.rb
@@ -65,23 +65,12 @@ module Oneoffs::NPQ
     end
 
     def migrate_line_items!(provider, from_statement, to_statement)
-      statement_line_items = from_statement.statement_line_items
+      statement_line_items = filter_statement_line_items(from_statement.statement_line_items)
 
-      if restrict_to_declaration_types
-        statement_line_items = statement_line_items
-          .includes(:participant_declaration)
-          .where(participant_declaration: { declaration_type: restrict_to_declaration_types })
-      end
-
-      if restrict_to_declaration_states
-        statement_line_items = statement_line_items
-          .includes(:participant_declaration)
-          .where(participant_declaration: { state: restrict_to_declaration_states })
-      end
-
-      record_line_items_info(provider, statement_line_items)
-
+      record_info("Migrating #{statement_line_items.size} declarations for #{provider.name}")
       statement_line_items.update!(statement_id: to_statement.id)
+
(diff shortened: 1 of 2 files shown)


app/services/oneoffs/npq/migrate_declarations_between_statements.rb
```diff
@@ -76,17 +92,37 @@ def each_statements_by_provider
       end
     end
 
+    def change_declaration_states_for_to_statement(to_statement, statement_line_items)
+      declarations = statement_line_items.map(&:participant_declaration).uniq
+
+      service = if to_statement.payable?
+                  ParticipantDeclarations::MarkAsPayable.new(to_statement)
+                elsif to_statement.paid?
```
cwrw, line 100: I think the paid is more tricky, first it has to go from payable to paid as there are checks in the service to only mark paid if payable. Can we avoid this all together as we shouldn't be moving declarations around if the previous statement has been paid?

cwrw (approved):

PR #4934, opened 2024-06-12
Amend sorting in NPQ participants query

cwrw:
### Context

- Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3182

Currently the default sorting looks into the nested `npq_profiles` created at timestamp. Without pagination this works ok, however with pagination, pagy includes the timestamp on the profiles in the limit/offset distinct query, which means that if a user has multiple profiles with different timestamps we return duplicate ids instead of the real ids for users.

Instead sort by user created at to avoid any timestamps being included in the pagy queries when calculating which ids to offset and return.

The test fails with the previous sorting but doesn't with the current one.

Amend the sorting on the participants npq query to look at users rather than npq profiles nested model. This ensures that we do not include the timestamp in the distinct query.

### Guidance to review
I have run a script to paginate every v3 endpoint, and the only issue is with NPQ participants. I have run the script with the new ordering and we get no issues

### DIFF
--- a/spec/requests/api/v3/npq_participants_spec.rb
+++ b/spec/requests/api/v3/npq_participants_spec.rb
@@ -68,6 +68,27 @@ RSpec.describe "NPQ Participants API", type: :request do
           expect(JSON.parse(response.body)["data"].size).to eql(1)
         end
 
+        context "When a user has multiple profiles" do
+          let!(:another_npq_application) { create(:npq_application, :accepted, npq_lead_provider:, user: npq_application.user) }
+          let(:ordered_users) { User.order(:created_at).pluck(:id) }
+
+          before do
+            another_npq_application.profile.update!(created_at: 1.day.ago)
+            another_npq_application.user.update!(created_at: 1.day.ago)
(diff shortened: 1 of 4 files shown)


spec/requests/api/v3/npq_participants_spec.rb
```diff
@@ -68,6 +68,27 @@
...
+          let(:ordered_users) { User.order(:created_at).pluck(:id) }
+
+          before do
+            another_npq_application.profile.update!(created_at: 1.day.ago)
+            another_npq_application.user.update!(created_at: 1.day.ago)
+            npq_application.profile.update!(created_at: 4.days.ago)
+            npq_application.user.update!(created_at: 4.days.ago)
+          end
+          it "can return paginated data" do
+            get "/api/v3/participants/npq", params: { page: { per_page: 2, page: 1 } }
```
cwrw, line 82: this test does fail on the old sorting, tidy up tomorrow

ethax-ross (approved):

peteryates (approved):

PR #4944, opened 2024-06-13
Add funded_place attributes accept an application request and response bodies.html.md

mark-avery-dfe:


docs/source/npq/guidance.html.md
```diff
@@ -179,14 +179,14 @@ An example request body is listed below.
 #### Example request body:
 
 ```
-{
-  "data": {
-    "type": "npq-application-accept",
-    "attributes": {
-      "schedule_identifier": "npq-leadership-spring"
```
padv2010, line 186: `V3` includes `schedule_identifier` but `V1` and `V2` don't.
If we are focused on V3 we would need to just add the `funded_place` attribute.

PR #4956, opened 2024-06-17
Tweak npq participant endpoints api docs

leandroalemao:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3082

During PO testing the ticket above, we found some discrepancies between the npq participant serializer and the api docs.

Fix those discrepancies in api v1/v2/v3 docs for npq participant endpoints.

### DIFF
--- a/spec/models/induction_record_spec.rb
+++ b/spec/models/induction_record_spec.rb
@@ -57,7 +57,7 @@ RSpec.describe InductionRecord, type: :model do
       end
 
       it "includes completed induction, current, transferring, transferred, training status withdrawn and training status deferred" do
-        expect(described_class.school_dashboard_relevant).to eq(expected_included)
+        expect(described_class.school_dashboard_relevant).to match_array(expected_included)
       end
     end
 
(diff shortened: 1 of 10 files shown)


config/initializers/devise.rb
```diff
@@ -14,7 +14,7 @@
   # confirmation, reset password and unlock tokens in the database.
   # Devise will use the `secret_key_base` as its `secret_key`
   # by default. You can change it below and use your own secret key.
-  config.secret_key = Rails.application.secrets.secret_key_base
+  config.secret_key = Rails.application.credentials.secret_key_base
```
cwrw, line 17: discussed in person: could we separate this pr into it's own pr and check our credentials have those set correctly

spec/models/induction_record_spec.rb
```diff
@@ -57,7 +57,7 @@
       end
 
       it "includes completed induction, current, transferring, transferred, training status withdrawn and training status deferred" do
-        expect(described_class.school_dashboard_relevant).to eq(expected_included)
+        expect(described_class.school_dashboard_relevant).to match_array(expected_included)
```
cwrw, line 60: discussed in person: does order matter here? if so should we fix the test?

cwrw (commented):
just the comments we discussed in person, rest looks great, thanks for fixing those @leandroalemao !

config/initializers/devise.rb
leandroalemao: Moving this to another PR.. 
Checked all credentials and all have `secret_key_base` set correctly.

cwrw (approved):

PR #4957, opened 2024-06-17
Redirect induction coordinator to schools index when they are not associated to any schools

steventux:
### Context

They can at least see they don't have an association to any schools as the index page will be empty.
We are currently raising a Sentry error and not catching this case.

- Ticket:  https://dfedigital.atlassian.net.mcas.ms/browse/CST-2621
- Sentry error: https://dfe-teacher-services.sentry.io/issues/4486911502 there are currently ~550 errors like this

Check that the induction coordinator is associated to a school before attempting to return a school dashboard path. If they aren't redirect to the schools index.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -34,7 +34,7 @@ module ApplicationHelper
   end
 
   def induction_coordinator_dashboard_path(user)
-    return schools_dashboard_index_path if user.schools.count > 1
+    return schools_dashboard_index_path if user.schools.count != 1
 
     school = user.induction_coordinator_profile.schools.first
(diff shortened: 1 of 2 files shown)


leoapost:
@steventux btw, displaying something like "You are not associated with any schools" could clarify why the page appears empty when SITs are redirected there. Any thoughts?

steventux:
@leoapost At first I thought this might be tricky given how the form accesses the path helper but the errors are consistently thrown in the same way so we may be able to handle this messaging in the `ChooseRolesController` if the user has no schools or alternatively make the `ChooseRoleForm` validate that the user has at least one school if the role is `induction_coordinator`.
Not sure which method is better but I'll try and find the least messy way to do this.

PR #4963, opened 2024-06-18
Ignore declarations in frozen cohort on ChangeSchedule

ethax-ross:
[Jira-3187](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3187)

### Context

For participants that have been migrated out of a payments frozen cohort, we want to ignore any declarations made in a frozen cohort when determining if they can change schedule.

This will allow providers to migrate participants to an extended schedule after they have migrated from the 2021 payments frozen cohort to the 2024 active registration cohort, even if they have billable declarations in the 2021 cohort.

- Ignore declarations in frozen cohort on ChangeSchedule

### Guidance to review

The `RecordDeclaration` service will ensure these declarations are still taken into account when a declaration is created; so if we have a `started` declaration in the 2021 payments frozen cohort the provider will not be able to record a `started` declaration for the same course in 2024:

https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/record_declaration.rb#L217-L229

### DIFF
--- a/app/services/change_schedule.rb
+++ b/app/services/change_schedule.rb
@@ -216,7 +216,7 @@ private
     return unless new_schedule
     return if changing_cohort_due_to_payments_frozen?
 
-    applicable_declarations.each do |declaration|
+    applicable_declarations_excluding_payments_frozen.each do |declaration|
       milestone = new_schedule.milestones.find_by!(declaration_type: declaration.declaration_type)
 
       if declaration.declaration_date <= milestone.start_date.beginning_of_day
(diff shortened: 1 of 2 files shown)


app/services/change_schedule.rb
```diff
@@ -230,8 +230,17 @@ def validate_new_schedule_valid_with_existing_declarations
   end
 
   def applicable_declarations
-    @applicable_declarations ||= participant_profile.participant_declarations
-      .where(state: %w[submitted eligible payable paid])
+    @applicable_declarations ||= begin
```
cwrw, line 233: I think this is also used in `validate_cannot_change_cohort_ecf`, which might relax the validation there to allow providers to change cohort for those frozen participants now 🤔

ethax-ross: Ah yeah you're right. I've pushed an update.

cwrw (approved):
thanks @ethax-ross, after this goes through testing, let's block the merge until we align with the schools team, thanks!

PR #4970, opened 2024-06-19
Enable log-only mode for DfE Analytics in review

peteryates:
### Context

Review apps aren't configured to push DfE-Analytics data to Google Cloud and are spewing lots of noise into the error log. Instead enable log-only mode for review app.


cwrw (approved):

config/initializers/dfe_analytics.rb
```diff
@@ -2,8 +2,9 @@
 
 DfE::Analytics.configure do |config|
   # Whether to log events instead of sending them to BigQuery.
-  #
-  # config.log_only = true
+  if Rails.env.review?
+    config.log_only = true
```
cwrw, line 6: maybe we can try setting `config.enable_analytics` to false

cwrw (approved):

config/initializers/dfe_analytics.rb
```diff
@@ -45,7 +45,7 @@
   # enable analytics. You might want to hook this up to a feature flag or
   # environment variable.
   #
-  config.enable_analytics = proc { FeatureFlag.active?(:dfe_analytics) }
+  config.enable_analytics = proc { !Rails.env.review? && FeatureFlag.active?(:dfe_analytics) }
```
saliceti, line 48: Could we not hardcode the environment name and use an environment variable instead? Could be set [here](https://github.com/DFE-Digital/early-careers-framework/blob/main/terraform/application/application.tf#L18) depending on the environment

peteryates: I'd be happy to do that but think having a feature flag **and** a env var is too much.

I'll add a backlog ticket to make it just use an env var.

saliceti: No worries

peteryates:
On closer inspection we have some more BigQuery stuff floating around, see:

* config/initializers/google_big_query.rb
* app/jobs/stream_big_query_participant_declarations_job.rb
* app/jobs/participant_outcomes/stream_api_requests_to_big_query_job.rb
* app/jobs/npq/stream_big_query_profile_job.rb
* app/jobs/npq/stream_big_query_enrollment_job.rb

Do we even use this stuff any more now we have DfE Analytics? cc @consult1815

PR #4973, opened 2024-06-20
Fix `Rails.application.secrets` is deprecated in favor of `Rails.application.credentials` message

leandroalemao:
### Context

- Ticket: N/A

```
`Rails.application.secrets` is deprecated in favor of `Rails.application.credentials` and will be removed in Rails 7.2.
```

Fix deprecation message above.
Checked all credentials files all have `secret_key_base` set correctly.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -14,7 +14,7 @@ Devise.setup do |config|
   # confirmation, reset password and unlock tokens in the database.
   # Devise will use the `secret_key_base` as its `secret_key`
   # by default. You can change it below and use your own secret key.
-  config.secret_key = Rails.application.secrets.secret_key_base
+  config.secret_key = Rails.application.secret_key_base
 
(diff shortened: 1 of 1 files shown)


steventux (approved):

config/initializers/devise.rb
```diff
@@ -14,7 +14,7 @@
   # confirmation, reset password and unlock tokens in the database.
   # Devise will use the `secret_key_base` as its `secret_key`
   # by default. You can change it below and use your own secret key.
-  config.secret_key = Rails.application.secrets.secret_key_base
+  config.secret_key = Rails.application.secret_key_base
```
cwrw, line 17: just double checking the existing value was `nil`, I'm worried that we might be changing it now (correctly) and will cause invalidation everywhere. Just confirming, let's hold off merging pelase

cwrw (approved):

PR #4974, opened 2024-06-20
Remove NPQ Google Drive integration

peteryates:
DfE no longer uses Google Drive and this integration is redundant.


leoapost (approved):

leandroalemao (approved):
Just a question regarding the 131 EligibilityImport records in the database in prod.. are we gonna use it for something? If we won't so all good.

peteryates:
Google Drive was deactivated a year or more ago so anything in there is now well out of date.

cwrw (commented):
looking good, can we remove the `google-apis-drive_v3` gem as well?

.env.development
```diff
@@ -1,6 +0,0 @@
-GOOGLE_CLIENT_ID=
```
cwrw, line 1: could we also remove those from `.env.template`?

PR #4979, opened 2024-06-21
[CST-2576] Add archiver for undeclared 2021 participants

tonyheadford:
### Context

- Ticket: [Jira](https://dfedigital.atlassian.net/browse/CST-2576)
- Add archiving functionality for undeclared 2021 participants with slightly different safeguards than the unvalidated/undeclared participant archivers

Add 2 new archivers that extend the original "unvalidated" user and profile archivers.


app/services/archive/undeclared_profile.rb
```diff
@@ -0,0 +1,33 @@
...
+    def initialize(participant_profile, cohort_year: 2021, reason: "undeclared participants in 2021", keep_original: false)
+      super(participant_profile, reason:, keep_original:)
+
+      @cohort = Cohort.find_by(start_year: cohort_year)
+    end
+
+    def check_profile_can_be_archived!
+      if profile_not_in_requested_cohort?
+        raise ArchiveError, "Profile #{participant_profile.id} is not in #{cohort.start_year} cohort"
+      elsif profile_has_declarations?
```
ltello, line 18: The method `ParticipantProfile::ECF#archivable?` will do all the checks for you

tonyheadford: That is worrying because there is already another flavour of archiving that this derives from where the checks are slightly different

ltello: Agreed.

app/services/archive/frozen_cohort_profile.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Archive
+  class FrozenCohortProfile < UnvalidatedProfile
```
ethax-ross, line 4: I think the inheritance on these archivers is a bit misleading as the `FrozenCohortProfile/User` archiver _will_ archive validated profiles/users, unless I've missed something?

tonyheadford: I've pulled some common bits out into a module

app/services/archive/frozen_cohort_user.rb
```diff
@@ -0,0 +1,61 @@
...
+      @user = user
+      @reason = reason
+      @keep_original = keep_original
+    end
+
+    def check_user_can_be_archived!
+      if users_excluded_roles.any?
+        raise ArchiveError, "User #{user.id} has excluded roles: #{users_excluded_roles.join(',')}"
+      elsif user_profiles_are_not_all_archivable?
+        raise ArchiveError, "User #{user.id} has profiles in non-frozen cohorts"
```
ltello, line 37: maybe this error messages is not meaningful when the participant has a profile in a frozen cohort that cannot be archived (unfinished 2021 for instance)

ltello (approved):

app/services/archive/frozen_cohort_user.rb
tonyheadford: Yes it's just an indication as previously I checked things separately but with the checks added to ParticipantProfile we only get true/false - but I thought at least it gives you an idea where to look

cwrw (commented):
looks good, thanks @tonyheadford 🙌

PR #4982, opened 2024-06-24
Enable log only mode for dfe analytics in review apps

peteryates:
Review apps aren't configured to push DfE-Analytics data to Google Cloud and are spewing lots of noise into the error log. Instead enable log-only mode for review app.

This is a recreation of #4970 after the build went weird.


leoapost (approved):

peteryates:
This (and #4974)  appears to have solved the Google-related errors in review apps. There are some others in there still caused by the GIAS SOAP interactions, they'll need to be solved separately.

PR #4987, opened 2024-06-25
Allow providers to migrate participants from 2021 to 2024

ethax-ross:
[Jira-3200](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3200)

### Context

The `ChangeSchedule` service already has this functionality, however it was locked down so that providers could not perform the action. It has been decided we want to open up this functionality to LPs going forward, so we can do this by passing `allow_change_to_from_frozen_cohort: true` into the `ChangeSchedule` service.

- Allow providers to migrate participants from 2021 to 2024

### DIFF
--- a/app/controllers/concerns/api/v1/participant_actions.rb
+++ b/app/controllers/concerns/api/v1/participant_actions.rb
@@ -22,7 +22,7 @@ module Api
       end
 
       def change_schedule
-        service = ChangeSchedule.new(action_params)
+        service = ChangeSchedule.new(action_params.merge!(allow_change_to_from_frozen_cohort: true))
 
         serialized_response_for(service)
       end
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/api/v1/participant_actions.rb
```diff
@@ -22,7 +22,7 @@ def resume
       end
 
       def change_schedule
-        service = ChangeSchedule.new(action_params)
+        service = ChangeSchedule.new(action_params.merge!(allow_change_to_from_frozen_cohort: true))
```
cwrw, line 25: I think we might need to merge both prs especially regarding https://github.com/DFE-Digital/early-careers-framework/pull/4963/files#diff-dceb8eaba398eba134e24e36e018bf28066f0a45746bfaaabcfb48cc7ab88b12R238 let's chat tomorrow :)

cwrw (approved):

PR #4995, opened 2024-06-28
Amend accept and change funding service to only accept funded place booleans only

cwrw:
### Context

We currently validate against the funded place attribute regardless of type. So if a string is passed in we compute it as true.
This confused providers as they were passing values such as `"false"` and `"null"` and this was computed as `true`

- Ticket: https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3205

Amend the validations in both services to only accept boolean values otherwise returns a validation error, even if it's a string.

### Guidance to review
Maybe we should allow strings and compute those values, but felt better to lock it down to just booleans

### DIFF
--- a/app/services/npq/application/accept.rb
+++ b/app/services/npq/application/accept.rb
@@ -10,13 +10,18 @@ module NPQ
       attribute :schedule_identifier
       attribute :funded_place
 
+      validates :funded_place,
+                inclusion: {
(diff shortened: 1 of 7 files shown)


ethax-ross (approved):
LGTM

PR #5009, opened 2024-07-02
CST-2657: add new restrictions to the tool that amends the cohort of participants

ltello:
### Context

  3 new restrictions/validations needs to be added to the tool that amends a participant's cohort.

- [Ticket](https://dfedigital.atlassian.net/browse/CST-2657) 

- 2023 or 2024 Participants with NIoT should not be moved into 2021 or 2022 cohorts. 
- Participants with notes in their profile should not be moved cohort.
- Allow any current induction status (except changed) of a participant in a cohort change.


tonyheadford (commented):
The inclusion of the NIoT stuff in `LeadProvider` doesn't feel quite right to me when it could all be encapsulated in the `AmendParticipantCohort` service. Is there any reason for doing it in the model that I'm missing?

ltello:
> The inclusion of the NIoT stuff in `LeadProvider` doesn't feel quite right to me when it could all be encapsulated in the `AmendParticipantCohort` service. Is there any reason for doing it in the model that I'm missing?

Well, I have right the opposite view on this. 
Actually this is about detecting a specific LeadProvider based on whatever characteristic that identifies it. To me it has only to do with the LeadProvider model and it should be encapsulated there so no external class should need to be changed if the identifiers of that lead provider change in the future.

ltello:
Niot functionality extracted into its own module.

tonyheadford (approved):
LGTM

PR #5024, opened 2024-07-09
Stop `DfE::Analytics::SendEvents` from filling logs

peteryates:
These create lots of noise in our logs and aren't much use.

Implementation lifted from [Teaching Vacancies](https://github.com/DFE-Digital/teaching-vacancies/pull/6946) ❤️

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -124,10 +124,16 @@ Rails.application.configure do
 
   if ENV["RAILS_LOG_TO_STDOUT"].present?
     config.rails_semantic_logger.add_file_appender = false
+    config.rails_semantic_logger.filter = proc { |log| log.name != "DfE::Analytics::SendEvents" }
 
     $stdout.sync = true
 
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -124,10 +124,16 @@
 
   if ENV["RAILS_LOG_TO_STDOUT"].present?
     config.rails_semantic_logger.add_file_appender = false
+    config.rails_semantic_logger.filter = proc { |log| log.name != "DfE::Analytics::SendEvents" }
```
peteryates, line 127: I thought about skipping all via something like this but we'd probably miss out on DfE::Analytics errors too (which are in the format `DfE::Analytics::BigQueryApi::SendEventsError`).

```suggestion
    config.rails_semantic_logger.filter = proc { |log| log.name !~ /DfE::Analytics/ }
```

cwrw (approved):

saliceti:
[Remove dfe analytics logs from Logit.io](https://trello.com/c/fxSQSbA6/1907-remove-dfe-analytics-logs-from-logitio)

PR #5027, opened 2024-07-10
[CPDLP-3228] NPQ MVP Prep - Disable NPQ endpoints in ECF specific separation environment

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3228

* `disable_npq_endpoints` feature added
* `NpqApiEndpointRoute` created to constraint NPQ routes
* For NPQ & ECF shared declarations routes:
  * Declarations create action returns error if `course_identifier` includes npq course
  * `ParticipantDeclarations::Index` updated to not return NPQ declarations if feature enabled
  * `Api::V3::ParticipantDeclarationsQuery` updated to not return NPQ declarations if feature enabled


cwrw (commented):
could we also disable NPQ statements as well from the statements joint endpoint?

app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -11,7 +11,14 @@ class ParticipantDeclarationsController < Api::ApiController
       include ApiFilterValidation
 
       def create
-        service = RecordDeclaration.new({ cpd_lead_provider: }.merge(permitted_params["attributes"] || {}))
+        attributes = permitted_params["attributes"] || {}
+
+        if FeatureFlag.active?(:disable_npq_endpoints) && attributes[:course_identifier].to_s.starts_with?("npq-")
```
cwrw, line 16: I understand why we might have this at the controller level, is there an argument of having this check at the record service level instead 🤔

config/routes.rb
```diff
@@ -89,7 +91,7 @@
     end
 
     namespace :v1 do
-      resources :npq_funding, only: [:show], path: "npq-funding", param: :trn
+      resources :npq_funding, constraints: NpqApiEndpointRoute, only: [:show], path: "npq-funding", param: :trn
```
cwrw, line 94: this is part of the internal API could we avoid hiding it?

config/routes.rb
```diff
@@ -113,21 +115,21 @@
...
+      resources :npq_applications, constraints: NpqApiEndpointRoute, only: %i[index show], path: "npq-applications" do
         member do
           post :accept
           post :reject
           put :change_funded_place, path: "change-funded-place"
         end
       end
 
-      resources :npq_profiles, only: %i[show create update], path: "npq-profiles"
+      resources :npq_profiles, constraints: NpqApiEndpointRoute, only: %i[show create update], path: "npq-profiles"
```
cwrw, line 126: same here regarding internal API please

cwrw: unless we want to keep it for live actually 🤔

app/controllers/api/v1/participant_declarations_controller.rb
mooktakim: As we can identify the NPQ declarations, I think it's best to return as soon as possible; that way, we don't interfere with the service classes.

leandroalemao (changes requested):
Hi @mooktakim 
Think `/api/v{1}{2}{3}/participant-declarations/{id}/void` is a shared endpoint as well as the statements one like @cwrw said?
As an idea, maybe we could return a deprecation warning do LPs rather than a 404 error?

cwrw:
> As an idea, maybe we could return a deprecation warning do LPs rather than a 404 error?
Good question, LPs asked for a "sandbox" like environment without the NPQ endpoints, so we have to disable them, also this will probably be used when we switch off the service so we need to completely disable those

app/controllers/api/v1/participant_declarations_controller.rb
mooktakim: The feature check is now done in `RecordDeclaration` service class

mooktakim:
> Hi @mooktakim Think `/api/v{1}{2}{3}/participant-declarations/{id}/void` is a shared endpoint as well as the statements one like @cwrw said? As an idea, maybe we could return a deprecation warning do LPs rather than a 404 error?

@leandroalemao added void endpoints now too

lib/npq_api_endpoint_route.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class NpqApiEndpointRoute
+  def self.matches?(_request)
+    !FeatureFlag.active?(:disable_npq_endpoints)
```
cwrw, line 5: I wonder if we should change this to check the ENV is separation instead and switch it to the feature flag later to avoid hitting the db everytime those URLs are hit? let's discuss

leandroalemao (approved):
Looking good.. thanks @mooktakim 👍 😄

app/services/api/v3/finance/statements_query.rb
```diff
@@ -23,6 +23,10 @@ def statements
               cohort_id: cohorts.map(&:id),
             )
 
+          if NpqApiEndpoint.disable_npq_endpoints?
```
cwrw, line 26: I wonder if we can use `statement_class` and only return `Finance::Statement::ECF` there instead to avoid the extra queries here might be simpler, similar to the declaration query?

cwrw (approved):
thank you @mooktakim ! Looks great, after this is tested we will need to wait for the separation env to be ready then we can switch the config to that from review and merge 🙌

config/environments/migration.rb
```diff
@@ -8,4 +8,9 @@
 
   # Used to handle HTTP_X_WITH_SERVER_DATE header for server side datetime overwrite
   config.middleware.use TimeTraveler
+
+  # Enable/disable aspects of the separation environment
```
cwrw, line 12: sorry could this move to separation instead of migration env? :D

cwrw (approved):

PR #5034, opened 2024-07-11
Set log level to `:warn` when Sidekiq running in server mode

peteryates:
This should reduce the number of entries we're sending on to Logit.io and stop us from going over the limit!

Approach lifted from [this PR](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/2283/files).

- **When Sidekiq running in prod set log level to warn**
- **Disable incorrect Rubocop rules**

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -61,3 +61,9 @@ Rails/SaveBang:
 Style/RedundantFetchBlock:
   Exclude:
     - 'app/middlewares/api_request_middleware.rb'
+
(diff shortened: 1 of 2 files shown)


saliceti (approved):

saliceti:
[Remove dfe analytics logs from Logit.io](https://trello.com/c/fxSQSbA6/1907-remove-dfe-analytics-logs-from-logitio)

PR #5051, opened 2024-07-16
[CPDLP-2764] Introduce 422 errors on NPQ accept application endpoint when there are errors with GAI ID/Conflict

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-2764

* Added `rescue_from Identity::TransferError` to npq_applications controllers to return errors

### DIFF
--- a/app/controllers/api/v1/npq_applications_controller.rb
+++ b/app/controllers/api/v1/npq_applications_controller.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require "csv"
-require "identity/transfer"
 
 module Api
   module V1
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/npq_applications_controller.rb
```diff
@@ -10,6 +11,8 @@ class NPQApplicationsController < Api::ApiController
       include ApiFilter
       include ApiFilterValidation
 
+      rescue_from Identity::TransferError, with: :identity_transfer_error_response
```
cwrw, line 14: maybe we can move this to the api controller instead where the other errors are?

leandroalemao (approved):
Tested here.. and it's looking good tks @mooktakim 👍 😄

cwrw (approved):

PR #5055, opened 2024-07-18
Add referred by return to teaching adviser to edge cases

slawosz:
https://dfedigital.atlassian.net/browse/CPDNPQ-1984

### DIFF
--- a/db/migrate/20240715080739_add_referred_by_return_to_teaching_adviser_to_npq_applications.rb
+++ b/db/migrate/20240715080739_add_referred_by_return_to_teaching_adviser_to_npq_applications.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddReferredByReturnToTeachingAdviserToNPQApplications < ActiveRecord::Migration[7.1]
+  def change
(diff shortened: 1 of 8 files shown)


db/migrate/20240715080739_add_referred_by_return_to_teaching_adviser_to_npq_applications.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddReferredByReturnToTeachingAdviserToNPQApplications < ActiveRecord::Migration[7.1]
+  def change
+    add_column :npq_applications, :referred_by_return_to_teaching_adviser, :string
```
TBenedyk, line 5: Prob best to add an index to this since we're querying on it

peteryates: Are the only possible values `"yes"` and `"no"`?

An index probably won't be much use with such low cardinality.

Also if that's the case might a Boolean field make sense? If more values allowed then clearly not - just couldn't see anything else in this PR.

peteryates (approved):
Looks fine to me overall

PR #5069, opened 2024-07-23
Remove NPQ references from swagger docs on separation env

ethax-ross:
[Jira-3275](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3275)

### Context

We want to remove all references to the NPQ endpoints from the swagger docs in the `separation` environment for providers to test.

- Fix references to NPQ in ECF

These references to NPQ participants should be ECF participants as they are in the context of an ECF path/schema.

- Conditionally remove NPQ references from swagger docs

We want to be able to conditionally remove the references to NPQ from the swagger docs. This includes stripping out paths, schemas and links to schemas as well as NPQ-specific pages.

- Dynamically replace API reference on boot

We want to serve two versions of the API reference depending on the environment we are running. In order to do this we are generating a second version in the `public/api-reference-without-npq` directory and then on boot we will replace `public/api-reference` with this version if applicable.

### Guidance to review

> ⚠️ Environment needs to change from `review` to `separation` prior to merging!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,7 @@ tokens
 /spec/examples.txt
 
 /public/api-reference
+/public/api-reference-without-npq
 
 /app/assets/builds/*
 !/app/assets/builds/.keep
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,10 @@ COPY public /public
 COPY swagger /swagger
 
 WORKDIR docs
+ENV REMOVE_NPQ_REFERENCES="false"
 RUN bundle exec middleman build --build-dir=../public/api-reference
+ENV REMOVE_NPQ_REFERENCES="true"
+RUN bundle exec middleman build --build-dir=../public/api-reference-without-npq
 
 # Stage 1: Download gems and node modules.
 FROM ${BASE_RUBY_IMAGE} AS builder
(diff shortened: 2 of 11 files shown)


cwrw:
nicely done @ethax-ross 👏 !

cwrw:
I found some NPQ course identifiers in declarations, not sure how easy it would be to remove those as well? https://cpd-ecf-review-5069-web.test.teacherservices.cloud/api-reference/reference-v3.html#api-v3-participant-declarations-get-parameters and the single/void endopoints as well I think

ethax-ross:
> I found some NPQ course identifiers in declarations, not sure how easy it would be to remove those as well? https://cpd-ecf-review-5069-web.test.teacherservices.cloud/api-reference/reference-v3.html#api-v3-participant-declarations-get-parameters and the single/void endopoints as well I think

Its a bit tricky doing these; I've pushed a commit but its not easily testable 😬

mooktakim (approved):

cwrw (approved):
🙌

PR #5078, opened 2024-07-26
Fix teaching adviser search

slawosz:


peteryates (approved):
Well spotted, looks sensible

PR #5091, opened 2024-07-31
Remove NPQ references from Swagger JSON

ethax-ross:
[Jira-3358](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3358)

### Context

In addition to scrubbing NPQ references from the guidance/API docs, we also want to remove them from the underlying JSON files.

- Remove NPQ references from JSON swagger files

Add `Preprocessor` to iterate over and remove NPQ references from JSON files when `REMOVE_NPQ_REFERENCES` is `true`.

For simplicity and given that this isn't going to stay in the codebase forever I've opted to do a VCR-style test to check the before/after from running the spec files through the preprocessors.

- Preprocess JSON in LeadProviderApiSpecification

Previously I tried to preprocess the JSON in the middleman build, but it turns out we serve the JSON files from the application rather than the API docs/middleman project.

Hook into `Preprocessor` as part of the `LeadProviderApiSpecification` library.

### Guidance to review

> ⚠️ Temporarily set to remove NPQ from review app docs; will update to separation before merging.


app/lib/lead_provider_api_specification.rb
```diff
@@ -15,6 +15,10 @@ def self.as_hash(version = CURRENT_VERSION)
   end
 
   def self.spec(version = CURRENT_VERSION)
-    YAML.load_file("swagger/#{version}/api_spec.json")
+    swagger_file = "swagger/#{version}/api_spec.json"
```
cwrw, line 18: 🙌

docs/spec/govuk_tech_docs/open_api/renderer_spec.rb
```diff
@@ -1,244 +0,0 @@
-# frozen_string_literal: true
```
cwrw, line 1: not sure we should be deleting this spec?

cwrw (commented):
really nicely done, thanks you @ethax-ross !!

docs/spec/govuk_tech_docs/open_api/renderer_spec.rb
ethax-ross: Whoops, no, I'll re-instate it

app/lib/lead_provider_api_specification.rb
```diff
@@ -15,6 +15,10 @@ def self.as_hash(version = CURRENT_VERSION)
...
+
+    Preprocessor.new(swagger_file).preprocess!
+
+    YAML.load_file(swagger_file)
```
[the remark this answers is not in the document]
ethax-ross: Yeah, bit strange not sure why it was done like that (maybe we used to serve our swagger docs in YAML?), but it works because YAML is a superset of JSON so it will parse it fine

cwrw (approved):

mooktakim (approved):

PR #5103, opened 2024-08-02
Add support and links to API docs footer

ethax-ross:
[Jira-3365](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3365)

### Context

We want to make the footer on the API docs consistent with that on the rest of the website.

- Add support and links to API docs footer

The links are supported by the tech docs gem, however the support paragraph required overriding the footer from the template. Usually in a Middleman template you can just specify a partial in the correct place and it will use it/override for you, however for some reason this doesn't work with the tech docs template. I think because of the way it builds/pulls from the tech docs gem. Instead, I've had to copy the layouts to our repository, rename them and update all pages to use them by default. We can then specify a custom footer.

### Guidance to review

| Before    | After |
| -------- | ------- |
| Screenshot 2024-08-02 at 11 17 09  | Screenshot 2024-08-02 at 11 16 49   |

### DIFF
--- a/docs/config.rb
+++ b/docs/config.rb
@@ -18,6 +18,8 @@ set :images_dir, "api-reference/javascripts"
 
 set :relative_links, true
 
+set :layout, :custom
+
 if ENV["REMOVE_NPQ_REFERENCES"].to_s == "true"
   ignore "/npq.html"
   ignore "/npq/**.*"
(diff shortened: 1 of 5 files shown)


leandroalemao (approved):
Tested locally here.. all good tks 👍 😄

cwrw (approved):
very annoying, thanks for this

PR #5105, opened 2024-08-05
Add validation to ensure NPQContract/Statement is present

ethax-ross:
[Jira-3223](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345&selectedIssue=CPDLP-3223)

### Context

In the accept, change schedule and change funded place actions we want to ensure the NPQContract and (for some) the Statement is present when the `npq_capping` feature is enabled.

- Add validation to ensure NPQContract/Statement is present

Add validation that raises a `Api::Errors::MissingNPQContractOrStatementError`, to be rescued in the API controller and an error message returned.

### Guidance to review

The error is raised so that it can be returned as a `400 bad request` instead of a `422 unprocessable entity`, which would happen if we performed the check as a validation rule in the model.

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -73,11 +73,11 @@ module Api
     end
 
     def identity_transfer_error_response(_exception)
-      render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:bad_request), params: I18n.t(:contact_us_to_resolve_issue)).call }, status: :unprocessable_entity
+      render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:unprocessable_request), params: I18n.t(:contact_us_to_resolve_issue)).call }, status: :unprocessable_entity
     end
 
     def missing_npq_contract_or_statement_response(_exception)
-      render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:bad_request), params: I18n.t(:missing_npq_contract_or_statement)).call }, status: :bad_request
+      render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:unprocessable_request), params: I18n.t(:missing_npq_contract_or_statement)).call }, status: :unprocessable_entity
     end
   end
 end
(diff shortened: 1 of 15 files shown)


app/controllers/api/api_controller.rb
```diff
@@ -74,5 +75,9 @@ def invalid_pagination_response(_exception)
     def identity_transfer_error_response(_exception)
       render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:bad_request), params: I18n.t(:contact_us_to_resolve_issue)).call }, status: :unprocessable_entity
     end
+
+    def missing_npq_contract_or_statement_response(_exception)
+      render json: { errors: Api::ParamErrorFactory.new(error: I18n.t(:bad_request), params: I18n.t(:missing_npq_contract_or_statement)).call }, status: :bad_request
```
cwrw, line 80: this is my bad we should probably go with a 422, let's chat tomorrow

app/services/npq/application/accept.rb
```diff
@@ -179,13 +180,22 @@ def validate_funded_place?
...
       def npq_contract
+        return unless statement
+
         @npq_contract ||=
-          NPQContract.where(
+          NPQContract.find_by(
             cohort_id: cohort.id,
             npq_lead_provider_id: npq_application.npq_lead_provider_id,
             course_identifier: npq_application.npq_course.identifier,
             version: statement.contract_version,
```
cwrw, line 190: do we need to set this to `statement&.contract_version` if the statement doesn't exist?

app/services/npq/application/accept.rb
```diff
@@ -179,13 +180,22 @@ def validate_funded_place?
...
-          ).first
+          )
+      end
+
+      def npq_contract_and_statement_exist
+        return unless FeatureFlag.active?("npq_capping")
+        return if errors.any?
+
+        raise Api::Errors::MissingNPQContractOrStatementError unless npq_contract && statement
```
cwrw, line 198:
```suggestion
        raise Api::Errors::MissingNPQContractOrStatementError unless statement && npq_contract
```

app/services/npq/application/accept.rb
ethax-ross: So before we get the NPQContract we check the statement exists (returning if it doesn't) so we're fine assuming it will always exist here

cwrw (approved):

mooktakim (approved):

PR #5114, opened 2024-08-07
Add release note for ECF standalone service

ethax-ross:
> ⚠️ WIP pending date

### Context

We want to add the release note for the ECF standalone API test environments.

- Add release note for ECF standalone service

### Guidance for review

I'm not sure the links to the environments make sense here as we're linking to `/api` and it'll just 404 in a browser 🤔

### DIFF
--- a/docs/source/release-notes.html.md
+++ b/docs/source/release-notes.html.md
@@ -7,6 +7,34 @@ weight: 8
 
 If you have any questions or comments about these notes, please contact DfE via Slack or email.
 
+## 13 August 2024 
+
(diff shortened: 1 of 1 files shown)


cwrw (approved):
I think re the URLs we are asking them to use the bearer tokens in the guidance before listing those so hopefully they understand that those are API urls.

cwrw:
@ethax-ross content is changing regarding your point about 404s, will update you shortly with new content

mooktakim (approved):

PR #5127, opened 2024-08-13
[CST-2653] Remove usage of SitePrism helper from test

ghost:
- and also remove skipped test

### Context

- Ticket:  https://dfedigital.atlassian.net/browse/CST-2653

Explicitly define helper methods in test rather than using generic objects that use SitePrism


steventux (approved):
Less magic = :100:

PR #5131, opened 2024-08-15
[CPDLP-3387] Generate shared seed data for separation

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3387

* `ValidTestDataGenerator::SeparationSharedData` to create participants with specific ecf id, trn and names. Same participants will be created in the NPQ app.


lib/tasks/valid_test_data_generator.rb
```diff
@@ -45,27 +45,34 @@ def generate_new_participants(school:, count:, sparsity_uplift:, pupil_premium_u
...
+    end
+
+    def create_participant(participant_identity:, school_cohort:, profile_type: :ect, mentor_profile: nil, status: "active", sparsity_uplift: false, pupil_premium_uplift: false)
+      user = participant_identity.user
+      teacher_profile = user.teacher_profile
       schedule = ecf_schedules.sample
-      participant_identity = Identity::Create.call(user:, origin: :ecf)
 
-      lead_provider = school_cohort.lead_provider
+      # lead_provider = school_cohort.lead_provider
```
ethax-ross, line 75: ✂️

lib/tasks/valid_test_data_generator/separation_shared_data.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require "tasks/valid_test_data_generator"
+
+module ValidTestDataGenerator
```
ethax-ross, line 5: Is it worth having tests for this?

ethax-ross (commented):
I tried running this locally after a `db:seed` but I got an error:

```
❯ rake lead_providers:seed_statements_and_applications
rails crake aborted!
NoMethodError: undefined method `each' for nil:NilClass (NoMethodError)

      shared_users_data[lead_provider.name].each do |user_params|
```

I also noticed in the ticket this point, which I don't think is covered, unless I've missed something?:

- Create a list of 30 users in ECF separation environment that have diverging participant id with NPQ separation environment that share a TRN and other data. Those will represent the new world and records

lib/tasks/valid_test_data_generator.rb
```diff
@@ -45,27 +45,33 @@ def generate_new_participants(school:, count:, sparsity_uplift:, pupil_premium_u
       while count.positive?
```
ethax-ross, line 45: It can be done in a separate PR, but I think these test data generators need to be in the same place; I think this one should probably move into `lib/valid_test_data_generator` and be renamed to match the class

mooktakim:
> I tried running this locally after a `db:seed` but I got an error:
> 
> ```
> ❯ rake lead_providers:seed_statements_and_applications
> rails crake aborted!
> NoMethodError: undefined method `each' for nil:NilClass (NoMethodError)
> 
>       shared_users_data[lead_provider.name].each do |user_params|
> ```
> 
> I also noticed in the ticket this point, which I don't think is covered, unless I've missed something?:
> 
> * Create a list of 30 users in ECF separation environment that have diverging participant id with NPQ separation environment that share a TRN and other data. Those will represent the new world and records

In the YAML file the users with `ecf_id` as `nil` will have diverged ID's.

lib/tasks/valid_test_data_generator/separation_shared_data.rb
mooktakim: Added spec

lib/tasks/valid_test_data_generator.rb
mooktakim: Agree, weird place for this. Would be good to move it on a separate PR

db/new_seeds/base/add_lead_providers_and_cips.rb
```diff
@@ -87,3 +87,8 @@
 }.each do |cpd_lead_provider, id|
   FactoryBot.create(:seed_npq_lead_provider, cpd_lead_provider:, id:, name: cpd_lead_provider.name)
 end
+
+CpdLeadProvider.find_each do |clp|
+  lp = LeadProvider.find_or_create_by!(name: clp.name)
+  lp.update!(cpd_lead_provider: clp)
```
cwrw, line 93: isn't this happening on line 63? as we also have to add cohorts to those lead providers?

mooktakim: sorry i didn't meant to commit this, will remove

lib/tasks/valid_test_data_generator.rb
```diff
@@ -45,27 +45,33 @@ def generate_new_participants(school:, count:, sparsity_uplift:, pupil_premium_u
...
+      Identity::Create.call(user:, origin: :ecf)
+    end
+
+    def create_participant(participant_identity:, school_cohort:, profile_type: :ect, mentor_profile: nil, status: "active", sparsity_uplift: false, pupil_premium_uplift: false)
+      user = participant_identity.user
+      teacher_profile = user.teacher_profile
       schedule = ecf_schedules.sample
-      participant_identity = Identity::Create.call(user:, origin: :ecf)
 
-      lead_provider = school_cohort.lead_provider
```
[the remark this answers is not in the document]
mooktakim: Removed as we already set the `lead_provider` in the initializer.

cwrw (approved):

PR #5166, opened 2024-09-05
Fix spec failures after Cohort rollover

cwrw:
### Context

Every Year when we roll into the next Cohort (after September) we get a lot of failures in the specs. We pushed the times to make sure we get time to fix the failures in https://github.com/DFE-Digital/early-careers-framework/pull/5158

This pr reverts that work, and introduces a proper fix for the failures rather than push the start time of the cohort.

We've introduced helpers to wrap around tests to travel to different times in the cohort. To avoid half the test suite failing tomorrow:

- Rename the `mid_cohort` rspec flag to `early_in_cohort` so we can create a new `mid_cohort` flag which is 6 months into a cohort.
- Add the new flag to everywhere we care being in the middle of the cohort which throws all calculations otherwise especially on declarations
- Unfortunately csv data is hardcoded so we need to change those dates. When we have more time those need to become dynamic
- Remove hard coded dates in other specs where we can
- Fix remaining tests with wrapping them with correct timeframe they need to be in, and travelling to declaration milestones

### Guidance to review
Commit by commit is easier

### DIFF
--- a/spec/fixtures/files/importers/cohort_csv_data.csv
+++ b/spec/fixtures/files/importers/cohort_csv_data.csv
@@ -1,2 +1,2 @@
 start-year,registration-start-date,academic-year-start-date,npq-registration-start-date,automatic-assignment-period-end-date,payments-frozen-at
-2025,2025/05/10,2025/09/01,2026/03/31,
+2026,2026/05/10,2026/09/01,2027/03/31,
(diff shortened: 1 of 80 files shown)


spec/fixtures/files/importers/cohort_csv_data.csv
```diff
@@ -1,2 +1,2 @@
 start-year,registration-start-date,academic-year-start-date,npq-registration-start-date,automatic-assignment-period-end-date,payments-frozen-at
-2025,2025/05/10,2025/09/01,2026/03/31,
+2026,2026/05/10,2026/10/01,2027/03/31,
```
ethax-ross: Was this change from Sep 25 to Oct 26 intentional (the month part)?

ethax-ross (approved):
Looks good 👏

spec/fixtures/files/importers/cohort_csv_data.csv
cwrw: oops will fix

ethax-ross (approved):

PR #5184, opened 2024-09-17
[CPDLP-3500] Add data cleanup script to fix applications that exist in ECF not NPQ

leandroalemao:
### Context

- Ticket: [CPDLP-3500](https://dfedigital.atlassian.net/browse/CPDLP-3500)

We've found applications that are duplicated in both ECF and NPQ that need to be deleted in ECF.

Add data cleanup script to fix applications that exist in ECF not NPQ.

[CPDLP-3500]: https://dfedigital.atlassian.net/browse/CPDLP-3500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/backup-and-restore-snapshot-database/action.yml
+++ b/.github/actions/backup-and-restore-snapshot-database/action.yml
@@ -39,7 +39,7 @@ runs:
     - name: Backup database
       shell: bash
       run: |
-        bin/konduit.sh cpd-ecf-${{ inputs.environment }}-web -- pg_dump -E utf8 --exclude-table-data api_requests --exclude-table-data versions --exclude-table-data email_schedules --exclude-table-data emails --compress=1 --clean --if-exists --no-privileges --no-owner --verbose -f backup-${{ inputs.environment }}.sql.gz
+        bin/konduit.sh cpd-ecf-${{ inputs.environment }}-web -- pg_dump -E utf8 --compress=1 --clean --if-exists --no-privileges --no-owner --verbose -f backup-${{ inputs.environment }}.sql.gz
 
     - name: Restore snapshot database
       shell: bash
--- a/.github/actions/backup-and-restore-snapshot-database/action.yml
+++ b/.github/actions/backup-and-restore-snapshot-database/action.yml
@@ -39,7 +39,7 @@ runs:
     - name: Backup database
       shell: bash
       run: |
-        bin/konduit.sh cpd-ecf-${{ inputs.environment }}-web -- pg_dump -E utf8 --compress=1 --clean --if-exists --no-privileges --no-owner --verbose -f backup-${{ inputs.environment }}.sql.gz
+        bin/konduit.sh cpd-ecf-${{ inputs.environment }}-web -- pg_dump -E utf8 --exclude-table-data api_requests --exclude-table-data versions --exclude-table-data email_schedules --exclude-table-data emails --compress=1 --clean --if-exists --no-privileges --no-owner --verbose -f backup-${{ inputs.environment }}.sql.gz
 
     - name: Restore snapshot database
       shell: bash
--- a/app/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq.rb
+++ b/app/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq.rb
@@ -1,0 +1,90 @@
+# frozen_string_literal: true
+
+module Oneoffs::NPQ::DataCleanup
+  class FixApplicationsInECFNotInNPQ
+    attr_reader :npq_application_ids
+
+    def initialize(npq_application_ids:)
+      @npq_application_ids = npq_application_ids
+    end
+
+    def run!(dry_run: true)
+      result = {}
+
+      ActiveRecord::Base.transaction do
+        result = npq_application_ids.each_with_object({}) do |npq_application_id, hash|
+          npq_application = NPQApplication.find_by(id: npq_application_id)
+          next unless npq_application
+
+          # Other applications for the same user/course/lead_provider/cohort
+          similar_applications = similar_applications(npq_application:)
+
+          # Where there is only one accepted similar application
+          # We keep the accepted one and delete the one which does not exist in NPQ.
+          accepted_applications = similar_applications.select(&:accepted?)
+          if accepted_applications.size == 1 && !npq_application.accepted?
+            npq_application.destroy!
+            hash[npq_application_id] = "Delete: #{[npq_application.npq_lead_provider.name, npq_application.user_id, npq_application.id]}"
+            next
+          end
+
+          # Where the application is accepted and all similar ones are rejected.
+          # We keep the accepted one and update NPQ accordingly.
+          # Also delete all the ones which does not exist in NPQ.
+          if npq_application.accepted? && similar_applications.size.positive? && similar_applications.all?(&:rejected?)
+            applications_ids_to_be_deleted = npq_application_ids.select { |id| similar_applications.map(&:id).include?(id) }
+            applications_to_be_deleted = NPQApplication.where(id: applications_ids_to_be_deleted).destroy_all
+            hash[npq_application_id] = "Keep #{npq_application.id}. Delete #{applications_to_be_deleted.map { |app| [app.npq_lead_provider.name, app.user_id, app.id] }}"
+            next
+          end
+
+          # Where the application is pending and all similar ones are also pending.
+          # We keep the latest one and update NPQ accordingly.
+          # Also delete all the ones which does not exist in NPQ.
+          if npq_application.pending? && similar_applications.size.positive? && similar_applications.all?(&:pending?)
+            applications_ids_to_be_deleted = npq_application_ids.select { |id| id != similar_applications.first.id && similar_applications.map(&:id).include?(id) }
+            applications_to_be_deleted = NPQApplication.where(id: applications_ids_to_be_deleted).destroy_all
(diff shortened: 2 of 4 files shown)


app/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq.rb
```diff
@@ -0,0 +1,90 @@
...
+      @npq_application_ids = npq_application_ids
+    end
+
+    def run!(dry_run: true)
+      result = {}
+
+      ActiveRecord::Base.transaction do
+        result = npq_application_ids.each_with_object({}) do |npq_application_id, hash|
+          npq_application = NPQApplication.find_by(id: npq_application_id)
+          next unless npq_application
```
mooktakim, line 17: I think we should raise an error here if the application id doesn't exist, it'll suggest the CSV data is incorrect

app/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    def similar_applications(npq_application:)
+      @similar_applications ||= {}
+      @similar_applications[npq_application.id] ||= user_applications(npq_application:).select do |a|
+        a.id != npq_application.id &&
+          a.npq_course == npq_application.npq_course &&
+          a.cohort == npq_application.cohort &&
+          a.npq_lead_provider == npq_application.npq_lead_provider
+      end
+    end
```
mooktakim, line 88: The memoization isn't needed here, as we only call them once.
```ruby
    def user_applications(npq_application:)
      npq_application.user.npq_applications.order(created_at: :desc)
    end

    def similar_applications(npq_application:)
      user_applications(npq_application:).select do |a|
        a.id != npq_application.id &&
          a.npq_course == npq_application.npq_course &&
          a.cohort == npq_application.cohort &&
          a.npq_lead_provider == npq_application.npq_lead_provider
      end
    end
```

app/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq.rb
```diff
@@ -0,0 +1,90 @@
...
+            npq_application.destroy!
+            hash[npq_application_id] = "Delete: #{[npq_application.npq_lead_provider.name, npq_application.user_id, npq_application.id]}"
+            next
+          end
+
+          # Where the application is accepted and all similar ones are rejected.
+          # We keep the accepted one and update NPQ accordingly.
+          # Also delete all the ones which does not exist in NPQ.
+          if npq_application.accepted? && similar_applications.size.positive? && similar_applications.all?(&:rejected?)
+            applications_ids_to_be_deleted = npq_application_ids.select { |id| similar_applications.map(&:id).include?(id) }
```
mooktakim, line 35: I would replace this like with:
```ruby
similar_application_ids = similar_applications.map(&:id)
applications_ids_to_be_deleted = npq_application_ids & similar_application_ids

```

app/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq.rb
mooktakim: Actually we need this because of line 36, we might delete it before we even get to it on the list.

spec/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        create(:npq_application, :accepted,
+               npq_course: application.npq_course,
+               npq_lead_provider: application.npq_lead_provider,
+               user: application.user,
+               cohort: application.cohort)
+      end
+      let(:npq_application_ids) { [application.id, similar_application.id] }
+
+      it "deletes it" do
+        expect { run }.to change { NPQApplication.count }.by(-1)
```
mooktakim, line 24: We should check the correct one was deleted

mooktakim: Similar for the other expects

mooktakim (changes requested):
Some comments above ^

app/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq.rb
```diff
@@ -0,0 +1,90 @@
...
+          next unless npq_application
+
+          # Other applications for the same user/course/lead_provider/cohort
+          similar_applications = similar_applications(npq_application:)
+
+          # Where there is only one accepted similar application
+          # We keep the accepted one and delete the one which does not exist in NPQ.
+          accepted_applications = similar_applications.select(&:accepted?)
+          if accepted_applications.size == 1 && !npq_application.accepted?
+            npq_application.destroy!
```
cwrw, line 26: If we have duplicates in both NPQ and ECF, how can we ensure we keep those in both as we agreed?

app/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq.rb
```diff
@@ -0,0 +1,90 @@
...
+          npq_application = NPQApplication.find_by(id: npq_application_id)
+          next unless npq_application
+
+          # Other applications for the same user/course/lead_provider/cohort
+          similar_applications = similar_applications(npq_application:)
+
+          # Where there is only one accepted similar application
+          # We keep the accepted one and delete the one which does not exist in NPQ.
+          accepted_applications = similar_applications.select(&:accepted?)
+          if accepted_applications.size == 1 && !npq_application.accepted?
```
ethax-ross, line 25: Just a suggestion, take it or leave it:

```
if similar_applications.select(&:accepted?).one? && !npq_application.accepted?
```

app/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq.rb
```diff
@@ -0,0 +1,90 @@
...
+            npq_application.destroy!
+            hash[npq_application_id] = "Delete: #{[npq_application.npq_lead_provider.name, npq_application.user_id, npq_application.id]}"
+            next
+          end
+
+          # Where the application is accepted and all similar ones are rejected.
+          # We keep the accepted one and update NPQ accordingly.
+          # Also delete all the ones which does not exist in NPQ.
+          if npq_application.accepted? && similar_applications.size.positive? && similar_applications.all?(&:rejected?)
```
ethax-ross, line 34: Again just a suggestion instead of checking `.size.positive` on these you could use `any?`:

```
if npq_application.accepted? && similar_applications.any? && similar_applications.all?(&:rejected?)
```

app/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq.rb
```diff
@@ -0,0 +1,90 @@
...
+
+module Oneoffs::NPQ::DataCleanup
+  class FixApplicationsInECFNotInNPQ
+    attr_reader :npq_application_ids
+
+    def initialize(npq_application_ids:)
+      @npq_application_ids = npq_application_ids
+    end
+
+    def run!(dry_run: true)
```
ethax-ross, line 11: I feel like a broken record today, but this is a pretty big method that might benefit from extracting into smaller methods 🙂

spec/services/oneoffs/npq/data_cleanup/fix_applications_in_ecf_not_in_npq_spec.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+describe Oneoffs::NPQ::DataCleanup::FixApplicationsInECFNotInNPQ do
```
ethax-ross, line 3: Worth adding tests for the `dry_run` flag and asserting on the result returned?

cwrw (commented):
Thanks @leandroalemao. I think generally having a think about this, I'm worried we might miss some strange shapes of data, for instance pending and rejected duplicates, or accepted and pending, might be worth revisiting the logic to ensure we cover all scenarios incase

.github/actions/backup-and-restore-snapshot-database/action.yml
```diff
@@ -39,7 +39,7 @@ runs:
     - name: Backup database
       shell: bash
       run: |
-        bin/konduit.sh cpd-ecf-${{ inputs.environment }}-web -- pg_dump -E utf8 --compress=1 --clean --if-exists --no-privileges --no-owner --verbose -f backup-${{ inputs.environment }}.sql.gz
+        bin/konduit.sh cpd-ecf-${{ inputs.environment }}-web -- pg_dump -E utf8 --exclude-table-data api_requests --exclude-table-data versions --exclude-table-data email_schedules --exclude-table-data emails --compress=1 --clean --if-exists --no-privileges --no-owner --verbose -f backup-${{ inputs.environment }}.sql.gz
```
cwrw, line 42: I think in the snapshot we need to include those as they help us debug especially the api request table so we shouldn't exclude them I don't think

leandroalemao: @cwrw I know that and totally agree and It won't be merged.. as just I added to be able to refresh snapshot quickly and test the script.

mooktakim (approved):

cwrw (approved):
thanks Leandro :)

ethax-ross (approved):

PR #5195, opened 2024-09-24
[CPDLP-3514] Scope `participant_declaration_finder` to completed declaration

leandroalemao:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3514

`NPQApplication` `participant_declaration_finder` is not scoped by completed declarations. It should be as the outcomes belongs to only completed declarations, not to started declarations, which are the first type of declaration for a participant profile, and it's the one that's being fetched using `find_by_participant_profile_id`.

Scope `participant_declaration_finder` by completed declarations as well, so we make sure we select the correct and completed declaration for the participant profile, and then after that the possible outcomes linked to those completed declarations.

### Guidance to review

Before: 
```
Api::V1::NPQ::ApplicationStatusQuery.new(NPQApplication.find("06787ded-3102-4647-ad15-d0d56b70139b").call
-> nil
```

After:
```
Api::V1::NPQ::ApplicationStatusQuery.new(NPQApplication.find("06787ded-3102-4647-ad15-d0d56b70139b").call
-> "passed"

### DIFF
--- a/app/models/npq_application.rb
+++ b/app/models/npq_application.rb
@@ -110,8 +110,8 @@ class NPQApplication < ApplicationRecord
     UK_CATCHMENT_AREA.include?(teacher_catchment)
   end
 
-  def self.participant_declaration_finder(participant_identity_id)
-    ParticipantDeclaration::NPQ.find_by_participant_profile_id(ParticipantProfile.find_by_participant_identity_id(participant_identity_id)&.id)
+  def self.completed_participant_declaration_finder(participant_identity_id)
+    ParticipantDeclaration::NPQ.find_by(declaration_type: "completed", participant_profile_id: ParticipantProfile.find_by_participant_identity_id(participant_identity_id)&.id)
   end
 
   def declared_as_billable?
--- a/app/models/npq_application.rb
+++ b/app/models/npq_application.rb
@@ -111,7 +111,7 @@ class NPQApplication < ApplicationRecord
   end
 
   def self.completed_participant_declaration_finder(participant_identity_id)
-    ParticipantDeclaration::NPQ.find_by(declaration_type: "completed", participant_profile_id: ParticipantProfile.find_by_participant_identity_id(participant_identity_id)&.id)
+    ParticipantDeclaration::NPQ.billable_or_changeable.for_declaration("completed").find_by(participant_profile_id: ParticipantProfile.find_by_participant_identity_id(participant_identity_id)&.id)
   end
 
   def declared_as_billable?
--- a/app/models/npq_application.rb
+++ b/app/models/npq_application.rb
@@ -110,8 +110,10 @@ class NPQApplication < ApplicationRecord
     UK_CATCHMENT_AREA.include?(teacher_catchment)
   end
 
-  def self.completed_participant_declaration_finder(participant_identity_id)
-    ParticipantDeclaration::NPQ.billable_or_changeable.for_declaration("completed").find_by(participant_profile_id: ParticipantProfile.find_by_participant_identity_id(participant_identity_id)&.id)
+  def latest_completed_participant_declaration
+    return unless profile
+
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: app/models/npq_application.rb [hunk 1]
```diff
@@ -110,8 +110,8 @@ def in_uk_catchment_area?
     UK_CATCHMENT_AREA.include?(teacher_catchment)
   end
 
-  def self.participant_declaration_finder(participant_identity_id)
-    ParticipantDeclaration::NPQ.find_by_participant_profile_id(ParticipantProfile.find_by_participant_identity_id(participant_identity_id)&.id)
+  def self.completed_participant_declaration_finder(participant_identity_id)
+    ParticipantDeclaration::NPQ.find_by(declaration_type: "completed", participant_profile_id: ParticipantProfile.find_by_participant_identity_id(participant_identity_id)&.id)
```
cwrw, line 114: we can have multiple completed declarations, one could be voided and then another one could be the correct one. Maybe we should select the billable one?

cwrw (commented):
thanks for fixing this Leandro :)

app/models/npq_application.rb
```diff
@@ -110,8 +110,8 @@ def in_uk_catchment_area?
     UK_CATCHMENT_AREA.include?(teacher_catchment)
   end
 
-  def self.participant_declaration_finder(participant_identity_id)
-    ParticipantDeclaration::NPQ.find_by_participant_profile_id(ParticipantProfile.find_by_participant_identity_id(participant_identity_id)&.id)
+  def self.completed_participant_declaration_finder(participant_identity_id)
+    ParticipantDeclaration::NPQ.billable_or_changeable.for_declaration("completed").find_by(participant_profile_id: ParticipantProfile.find_by_participant_identity_id(participant_identity_id)&.id)
```
cwrw, line 114: Let's chat about this tomorrow, I just realised we aren't scoping the profile to an NPQ profile which could be a big issue, and the service calling this is quite hard to follow.
I would change this whole lookup to something like the following:
```
completed_declaration = npq_application.profile.participant_declarations.npq.complete.billable_or_changeable.order(created_at: :desc).first
```

and we get the outcome for it in the other service:
```
completed_declaration.outcomes.latest.state
```
But yea let's chat tomorrow :)

spec/services/api/v1/npq/application_status_query_spec.rb
```diff
@@ -3,36 +3,40 @@
...
           expect(service.call).to be_nil
         end
       end
     end
+
     context "when no participant declaration exists" do
-      before do
-        allow(NPQApplication).to receive(:participant_declaration_finder).and_return(nil)
-      end
+      before { npq_application.destroy }
```
cwrw, line 38: I think here we should set the declaration to `nil` rather destroy the whole application

spec/requests/api/v1/npq/application_synchronization/index_spec.rb
```diff
@@ -5,28 +5,30 @@
...
+    let(:npq_application) { create(:npq_application, :accepted) }
     let(:pick_application) { NPQApplication.where(id: npq_application.id).pick(:lead_provider_approval_status, :id, :participant_identity_id) }
-    let(:state) do
-      participant_declaration_id = NPQApplication.participant_declaration_finder(pick_application.last)&.id
-      ParticipantOutcome::NPQ.latest_per_declaration&.find_by_participant_declaration_id(participant_declaration_id)&.state
-    end
+    let(:participant_declaration) { create(:npq_participant_declaration, cpd_lead_provider: npq_application.npq_lead_provider.cpd_lead_provider, participant_profile: npq_application.profile) }
+    let!(:participant_outcome) { create(:participant_outcome, participant_declaration:) }
+
+    before { participant_declaration.update!(declaration_type: "completed") }
```
cwrw, line 18: same here

PR #5204, opened 2024-09-30
Removed error on transfer dedupe

TBenedyk:
When we attempt to dedupe users, we currently raise an error if both users have a get_an_identity_id. This causes issues with syncing from NPQ that need to be fixed manually. This is a time consuming and repetitive process.

This PR automates this process and removes the error being raised. Now we select the most recent user as the primary user and remove the get_an_identity_id from the older user. This allows us to merge them and complete the transfer process without error.


app/services/npq/dedupe_participant.rb
```diff
@@ -15,15 +15,30 @@ class DedupeParticipant
     def call
       return if invalid?
 
-      Identity::Transfer.call(from_user: npq_application.user, to_user: primary_user_for_trn)
+      if get_an_identity_id_clash?
+        # If both users have a get_an_identity_id, we transfer to the most recent
```
cwrw, line 19: I think this might cause an issue with flip flopping values as we call identity transfer in other places on primary trn users, let's chat tomorrow might be easier

app/controllers/api/api_controller.rb
```diff
@@ -21,7 +21,6 @@ class ApiController < ActionController::API
     rescue_from Api::Errors::InvalidTrainingStatusError, with: :invalid_training_status_response
     rescue_from Api::Errors::MissingNPQContractOrStatementError, with: :missing_npq_contract_or_statement_response
     rescue_from Pagy::VariableError, with: :invalid_pagination_response
-    rescue_from Identity::TransferError, with: :identity_transfer_error_response
```
cwrw, line 24: maybe we can also remove `identity_transfer_error_response` if it's not used anywhere?

cwrw (approved):
looks great, just the one minor comment

PR #5219, opened 2024-10-07
[CPDLP-3555] Create pending applications in the current ECF/NPQ sandbox

leandroalemao:
### Context

- Ticket: [CPDLP-3555](https://dfedigital.atlassian.net/browse/CPDLP-3555)

Add additional test applications in a pending state to the ECF/NPQ combined sandbox to meet provider needs for more data covering a variety of scenarios, subjects, intakes, and cohorts.

Create pending applications in the current ECF/NPQ sandbox.

[CPDLP-3555]: https://dfedigital.atlassian.net/browse/CPDLP-3555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/tasks/sandbox_data.rake
```diff
@@ -17,13 +21,13 @@ namespace :lead_providers do
 
   desc "create NPQ seed schools and participants for API testing"
   task seed_schools_and_participants_npq: :environment do
-    logger = Logger.new($stdout)
-    logger.formatter = proc do |_severity, _datetime, _progname, msg|
-      "#{msg}\n"
-    end
+    return unless Rails.env.in?(%w[development review sandbox])
```
mooktakim, line 24: could there be a reason for this type of logger formatting? maybe for logit/kibana

leandroalemao: Not sure.. seems code with no usage to me that's why I removed.. best practice is to use `Rails.logger` by default which level is can configured by environment.

lib/tasks/valid_test_data_generator/base_populater.rb
```diff
@@ -0,0 +1,70 @@
...
+      @number_of_participants = number_of_participants
+      @logger = logger
+      # Ignoring ASO course, is an old course which we shouldn't create data
+      @npq_courses = NPQCourse.all.reject { |c| c.identifier == "npq-additional-support-offer" }
+    end
+
+    def create_application!(lead_provider:, school:, npq_course:, cohort:, participant_identity:)
+      NPQApplication.create!(
+        active_alert: "",
+        date_of_birth: Date.new(1990, 1, 1),
```
cwrw, line 37: is this because of the magic dates? or could we change this to be more random, if it's because of magic dates we can keep the same

lib/tasks/sandbox_data.rake
cwrw: looks like it's very old :D

lib/tasks/valid_test_data_generator/npq_lead_provider_populater.rb
```diff
@@ -0,0 +1,104 @@
...
+
+module ValidTestDataGenerator
+  class NPQLeadProviderPopulater < BasePopulater
+    def populate
+      return unless Rails.env.in?(%w[development review sandbox])
+
+      logger.info "NPQLeadProviderPopulater: Started!"
+
+      ActiveRecord::Base.transaction do
+        generate_new_schools!
```
cwrw, line 16: maybe we rename this to show that it is also generating participants as well as schools?

lib/tasks/valid_test_data_generator/base_populater.rb
leandroalemao: Yeah probably magic dates related. 👍

mooktakim (approved):

cwrw (approved):

PR #5234, opened 2024-10-15
[CPDLP-3589] Disable NPQ in admin

leandroalemao:
### Context

We need to disable any NPQ logic and services in the admin interface.

- Ticket: [CPDLP-3589](https://dfedigital.atlassian.net/browse/CPDLP-3589)

- Disable any NPQ tools in the admin interface
- Disable any tabs/routes that show NPQ data
- Add the feature flag to any services that show both, so when disabled will only show ECF

[CPDLP-3589]: https://dfedigital.atlassian.net/browse/CPDLP-3589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/admin/participants/search.rb
+++ b/app/services/admin/participants/search.rb
@@ -40,11 +40,8 @@ private
   end
 
   def type_conditions
-    if type.present?
-      ParticipantProfile.where(type:)
-    else
-      ParticipantProfile.all
-    end
+    scope = NpqApiEndpoint.disabled? ? ParticipantProfile.ecf : ParticipantProfile.all
+    type.present? ? scope.where(type:) : scope
   end
 
   def left_outer_joins
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -259,7 +259,7 @@ Rails.application.routes.draw do
     # Keeping the urls to old guidance urls, but they need to lead to new api-reference ones
     get "/guidance/home", to: redirect("/api-reference")
     get "/guidance/ecf-usage", to: redirect("/api-reference/ecf-usage")
-    get "/guidance/npq-usage", to: redirect("/api-reference/npq-usage")
+    get "/guidance/npq-usage", to: redirect("/api-reference/npq-usage"), constraints: NpqApiEndpoint
     get "/guidance/reference", to: redirect("/api-reference/reference")
     get "/guidance/release-notes", to: redirect("/api-reference/release-notes")
     get "/guidance/help", to: redirect("/api-reference/help")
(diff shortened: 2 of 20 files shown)


cwrw (commented):
@leandroalemao do you mind adding tests please, we need to ensure we have thorough testing for everywhere we add a feature flag

app/controllers/finance/participants_controller.rb
```diff
@@ -14,15 +14,24 @@ def index
       end
     end
 
-    def show; end
+    def show
```
cwrw, line 17: we have a [ticket](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?selectedIssue=CPDLP-3590) to address finance changes, can we please exclude those from this pr?

app/services/admin/participants/search.rb
```diff
@@ -40,11 +40,9 @@ def search_conditions
   end
 
   def type_conditions
-    if type.present?
-      ParticipantProfile.where(type:)
-    else
-      ParticipantProfile.all
-    end
+    scope = ParticipantProfile.all
+    scope = scope.merge(ParticipantProfile.ecf) if NpqApiEndpoint.disabled?
```
ethax-ross, line 44: The `merge` confused me for a second here; would it be more readable without it? something like:

```
scope = NpqApiEndpoint.disabled? ? ParticipantProfile.ecf : ParticipantProfile.all
```

config/routes.rb
```diff
@@ -259,7 +259,7 @@
     # Keeping the urls to old guidance urls, but they need to lead to new api-reference ones
     get "/guidance/home", to: redirect("/api-reference")
     get "/guidance/ecf-usage", to: redirect("/api-reference/ecf-usage")
-    get "/guidance/npq-usage", to: redirect("/api-reference/npq-usage")
+    get "/guidance/npq-usage", to: redirect("/api-reference/npq-usage"), constraints: NpqApiEndpoint
```
mooktakim, line 262: For this type of constraint, i created a request spec to make sure its removed:
https://github.com/DFE-Digital/early-careers-framework/pull/5237/files#diff-290c429c7303749f1b5ef4b8a375f7e24f43d90c12adfc98e819804adc277050R19

In this case you can have route not exist when feature active, and redirect when feature not active.

spec/models/finance/schedule/ecf_spec.rb
```diff
@@ -32,7 +32,7 @@
     end
   end
 
-  Cohort.where.not(start_year: 2020).find_each do |cohort|
+  Cohort.where(start_year: 2021..Cohort.current.start_year).find_each do |cohort|
```
cwrw, line 35: So I understand why this passes now on first run (like in CI) but fails on the second run: on first run because we don't have all the cohorts it doesn't run these tests, but on second one after it set up everything it does run and fails because of what you said 👍 let's keep it in, thank for fixing

cwrw (commented):
Past the test comment from Mook looks good to me 👍

spec/features/admin/participants/participant_steps.rb
```diff
@@ -191,6 +203,10 @@ def and_i_have_added_a_mentor
     Mentors::AddToSchool.call(school: @school, mentor_profile: @participant_profile_mentor)
   end
 
+  def and_i_have_added_a_npq
```
cwrw, line 206:
```suggestion
  def and_i_have_added_an_npq_profile
```

cwrw (approved):

ethax-ross (approved):
Looks good past Abeer's comment 👍

mooktakim (approved):

PR #5233, opened 2024-10-14
[CPDLP-3588] Turn off NPQ in ECF - Swap out app config for feature flag

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3588

### Changes proposed in this pull request

* Updated `NpqApiEndpoint` service class to use feature flag
* Replaced all `env` config with the feature flag `FeatureFlag.active?(:disable_npq)`

PR #5237, opened 2024-10-16
[CPDLP-3590] Turn off NPQ in ECF - Disable NPQ in finance view

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3590

* Updated finance dashboard to hide any NPQ related stuff for:
  * participant drill down
  * statements
  * schedules
* Hidden all NPQ finance routes

### Guidance to review

* Some FIXME's as I forgot to branch off 🤦‍♂️ https://github.com/DFE-Digital/early-careers-framework/pull/5233

### DIFF
--- a/app/forms/finance/choose_payment_breakdown_form.rb
+++ b/app/forms/finance/choose_payment_breakdown_form.rb
@@ -10,10 +10,13 @@ module Finance
     validates :provider, presence: { message: I18n.t("errors.provider.blank") }, on: :choose_provider
 
     def programme_choices
-      [
-        OpenStruct.new(id: "ecf", name: "ECF payments"),
-        OpenStruct.new(id: "npq", name: "NPQ payments"),
-      ]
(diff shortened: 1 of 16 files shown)


leandroalemao (commented):
Think we'll need to add some logic here as well maybe?

https://github.com/DFE-Digital/early-careers-framework/blob/main/app/controllers/finance/participants_controller.rb#L21

So we can't search by npq participants?

mooktakim:
> Think we'll need to add some logic here as well maybe?
> 
> https://github.com/DFE-Digital/early-careers-framework/blob/main/app/controllers/finance/participants_controller.rb#L21
> 
> So we can't search by npq participants?

That controller won't require it as the show view doesn't show any NPQ stuff. It will only redirect to the user.

cwrw (approved):

ethax-ross (approved):

leandroalemao (approved):

PR #5253, opened 2024-10-21
Move API Requests in DFE Analytics to custom events

cwrw:
### Context

- Ticket: n/a

API requests are huge and we need to stop sending them via BQ dfe analytics, but via a custom event

- remove API request from BQ dfe analytics
- add custom event to send this data

### Guidance to review
commit by commit

### DIFF
--- a/config/analytics_custom_events.yml
+++ b/config/analytics_custom_events.yml
@@ -1,0 +1,3 @@
+---
+shared:
+  - persist_api_request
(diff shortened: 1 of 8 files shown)


config/analytics_custom_events.yml
peteryates, line 1: We need to add this line as per [the guidance](https://github.com/DFE-Digital/dfe-analytics?tab=readme-ov-file#custom-events).

peteryates (approved):
Looks good!

peteryates (approved):

PR #5272, opened 2024-10-28
Remove `analytics_pii.yml`

peteryates:
This is no longer required as we're now using config/analytics_hidden_pii.yml and having both might cause confusion and cause us to accidentally upload pseudonymised data.


peteryates:
Closing until DFE-Digital/dfe-analytics/pull/166 has been released.

PR #5275, opened 2024-10-29
[CST-2788] Add mailer to remind SITs to set up training for the current year

leoapost:
### Context

- Ticket: CST-2788 

We need to send reminders to SITs from schools that ran FIP programs in 23/24 but haven’t yet engaged this year to set up their training.

Should exclude:
- CIP and DIY schools 
- independent or independent special schools (even if in receipt of Section-41 funding)
- children's centres
- British Schools Overseas
- Welsh schools

- Add the mailer for the Notify template
- Add the bulk mailer to trigger the emails


peteryates (approved):
Looks good to me.

PR #5278, opened 2024-10-29
CST-2790: frozen cohort triggers must also include unfinished participants with no billable declarations

ltello:
### Context

We have 3 triggers to automatically move participants in 2021 to 2024 due to 2021 cohort being frozen and no more lead provider declarations/payments can be accepted.

Until now only those participants complying with some requirements (unfinished participants) were picked and moved by the triggers implemented in June.

This PR extends those triggers to include any remaining participant in 2021 (not completed induction). 

The treatment will be similar to those unfinished participants but the flag that mark them to the LP won’t be setup for this new group.

- [Ticket: ](https://dfedigital.atlassian.net/browse/CST-2790)


leoapost (approved):
LGTM

PR #5297, opened 2024-11-08
Bump migration env to match production

ethax-ross:
Upgrade the migration webapp/worker instances and memory to match production for the performance testing.


cwrw (approved):

ethax-ross:
Going to re-open as review app is busted

PR #5304, opened 2024-11-11
Disable rate limiting in migration env

ethax-ross:
So we can conduct performance testing.


ethax-ross:
Closing; deployed manually for performance testing which is finished now.

PR #5313, opened 2024-11-13
[CST-2801] Point users to the central Teacher Services privacy policy

leoapost:
### Context

- Ticket: CST-2801

As part of the governance updates we have been asked to point users to the central Teacher Services policy ([link](https://www.gov.uk/government/publications/privacy-information-education-providers-workforce-including-teachers)).

- Update the content in our privacy policy as requested

### Guidance to review
- Publish the new policy with the rake task `rake data:privacy_policy:major`
- Login with a user that haven't accepted the policy yet
- Confirm the new policy is displayed with the correct content
- Confirm accepting the policy works as expected


peteryates (approved):
This
Code-wise this is fine and if we _need_ this page then fair enough, but it now has so little on it it seems a bit pointless having it at all.

Can't we just link direct to the shared one and add the "...by continuing, you agree to our privacy notice" to the link?

peteryates (approved):

PR #5340, opened 2024-12-03
Fix `UnfundedMentorsQuery` to surface correct email

ethax-ross:
[Jira-3788](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3788)

### Context

Ambition flagged 3 separate cases of the unfunded mentors query returning incorrect/old emails, instead of the lates ones from the induction record. We want to update the query to use the induction records so that we surface the correct email.

- Fix UnfundedMentorsQuery email

We are currently surfacing the email from the `ParticipantProfile`, however we instead should be looking it up from the `InductionRecord`.

Update query to join `ParticipantIdentity` on `InductionRecord` and surface `email` from there. Note we need to join on the mentor's latest induction records (the base induction records are relative to the mentee).

- Add release note for UnfundedMentorsQuery email update

Add release note to explain the bug fix for the unfunded-mentors endpoint returning the email from the profile rather than the induction records.

- Relax UnfundedMentorsQuery

To be consistent with the current query and return the same users we need to include `InductionRecord` instances for the mentor where they have a challenged partnership or no lead provider. We also need to `LEFT OUTER JOIN` on the
`participant_identities` as some users do not have a `preferred_identity` on the `InductionRecord` that still exists.

There are also some profiles that don't have any `induction_records` so we need to `LEFT OUTER JOIN` on those as well.

### Guidance for review

The extra join makes this query quite a bit slower.

Tested against the participant `2c6fe2c6-46d9-4b32-9ee8-0ab719b4368a` in the ticket, which now returns with the correct email address from their latest induction record.

Other testing:

```
83 records that are no longer returned in the response
76 records where the email changes (I've eyeballed a lot of these and the changes look sensible)
0 records that are in the new response but not in the old response
```
Of the 83 records that are no longer returned, 82 of them are being excluded because we join the mentor profile induction records and exclude those without a partnership/lead provider or if they have a challenged partnership:

```
ids = [the 83 no longer returned]
User.where(id: ids).map { |u| u.mentor_profile.latest_induction_record.induction_programme.partnership }.compact.reject { |p| p.challenged_at || p.challenge_reason }.count

=> 1
```

The single outlier is `940ed235-1609-4c5d-ae2d-e2d3ac2ca916` and they appear to be omitted as their `latest_induction_record` does not have a `preferred_identity`. Oddly it has a `preferred_identity_id` but it no longer exists:

```
User.find("940ed235-1609-4c5d-ae2d-e2d3ac2ca916").mentor_profile.latest_induction_record.preferred_identity_id
"5352ea4d-0031-45ee-a53b-90151b77ae03"
User.find("940ed235-1609-4c5d-ae2d-e2d3ac2ca916").mentor_profile.latest_induction_record.preferred_identity
nil
```

### DIFF
--- a/spec/services/api/v3/ecf/unfunded_mentors_query_spec.rb
+++ b/spec/services/api/v3/ecf/unfunded_mentors_query_spec.rb
@@ -32,7 +32,7 @@ RSpec.describe Api::V3::ECF::UnfundedMentorsQuery do
       let!(:latest_induction_record) { create(:induction_record, :future_start_date, induction_programme: other_induction_programme, participant_profile: unfunded_mentor_participant_profile, preferred_identity: latest_preferred_identity) }
 
       it "returns the preferred email from the latest induction record" do
-        expect(unfunded_mentor_participant_profile.induction_records.pluck(&:participant_email).uniq.count).to be > 1
+        expect(unfunded_mentor_participant_profile.induction_records.pluck(&:participant_email).uniq.count).to eq(2)
         expect(subject.unfunded_mentors.first.preferred_identity_email).to eq(latest_preferred_identity.email)
       end
     end
--- a/spec/services/api/v3/ecf/unfunded_mentors_query_spec.rb
+++ b/spec/services/api/v3/ecf/unfunded_mentors_query_spec.rb
@@ -24,16 +24,16 @@ RSpec.describe Api::V3::ECF::UnfundedMentorsQuery do
       expect(subject.unfunded_mentors.first.user_id).to eq(unfunded_mentor_participant_profile.user_id)
     end
 
-    context "with preferred identity" do
-      let(:preferred_email) { Faker::Internet.email }
-      let!(:preferred_identity) { create(:participant_identity, :secondary, user: unfunded_mentor_participant_profile.user, email: preferred_email) }
+    context "when the mentor has multiple induction records with different preferred_identity associations" do
+      let(:latest_preferred_identity) { create(:participant_identity, :secondary, user: unfunded_mentor_participant_profile.user) }
+      let(:other_induction_programme) { create(:induction_programme, :fip, partnership: other_partnership) }
+      let(:other_partnership) { create(:partnership, lead_provider: other_lead_provider, cohort:) }
+      let(:other_lead_provider) { create(:lead_provider) }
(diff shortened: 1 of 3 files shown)


cwrw (commented):
looks good thanks Ross from working through this one. I guess just to confirm: we return all the same ids for all providers for this endpoint? and that the email changes all made sense (I know the answer to both but just to check :D)

ethax-ross:
> looks good thanks Ross from working through this one. I guess just to confirm: we return all the same ids for all providers for this endpoint? and that the email changes all made sense (I know the answer to both but just to check :D)

Yep, the IDs are all the same across all providers, just the emails in the spreadsheets that have changed

spec/services/api/v3/ecf/unfunded_mentors_query_spec.rb
```diff
@@ -24,16 +24,16 @@
...
-        expect(subject.unfunded_mentors.first.preferred_identity_email).to eq(unfunded_mentor_participant_profile.participant_identity.email)
+    context "when the mentor has multiple induction records with different preferred_identity associations" do
+      let(:latest_preferred_identity) { create(:participant_identity, :secondary, user: unfunded_mentor_participant_profile.user) }
+      let(:other_induction_programme) { create(:induction_programme, :fip, partnership: other_partnership) }
+      let(:other_partnership) { create(:partnership, lead_provider: other_lead_provider, cohort:) }
+      let(:other_lead_provider) { create(:lead_provider) }
+      let!(:latest_induction_record) { create(:induction_record, :future_start_date, induction_programme: other_induction_programme, participant_profile: unfunded_mentor_participant_profile, preferred_identity: latest_preferred_identity) }
+
+      it "returns the preferred email from the latest induction record" do
+        expect(unfunded_mentor_participant_profile.induction_records.pluck(&:participant_email).uniq.count).to be > 1
```
mooktakim, line 35: Should we check for 2 emails instead of greater than 1?

mooktakim (approved):
One minor comment otherwise 👍

mooktakim (approved):

PR #5343, opened 2024-12-03
[CPDLP-3773] Remove standalone NPQ endpoints (NPQ Post Separation Cleanup) 

leandroalemao:
### Context

- Ticket: [CPDLP-3773](https://dfedigital.atlassian.net/browse/CPDLP-3773)

Now separation has been done, we need to remove any NPQ endpoints from ECF.

- Remove NPQ standalone endpoints only;
- Remove controllers/implementation and routes as well as tests;
- Remove API doc specs as well;
- Shared endpoints will be in another ticket;


db/schema.rb
```diff
@@ -265,7 +265,7 @@
     t.index ["start_year"], name: "index_cohorts_on_start_year", unique: true
   end
 
-  create_table "cohorts_lead_providers", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+  create_table "cohorts_lead_providers", force: :cascade do |t|
```
ethax-ross, line 268: I don't think we should be changing the schema?

ethax-ross (commented):
Looks good - 30k lines! 🔥  is it worth checking the `Gemfile` for anything we can cull that might have been NPQ-specific? 🤔

app/views/finance/participants/_npq_application.html.erb
```diff
@@ -68,14 +68,4 @@
     <% row.with_value { application.updated_at.to_fs(:govuk) } %>
     <% row.with_action(text: :none) %>
   <% end %>
-
-  <% summary_list.with_row do |row| %>
```
mooktakim, line 72: Shouldn't we delete these files instead?
I think we render the partial on `app/views/finance/participants/show.html.erb`

leandroalemao: Well I thought of doing that but was trying to focus on removing the things related to the API endpoints only in this card.

leandroalemao:
> Looks good - 30k lines! 🔥 is it worth checking the `Gemfile` for anything we can cull that might have been NPQ-specific? 🤔

I can see only `gem "countries"` that was used specific for NPQ stuff.. but reckon it'll be removed in another ticket as it's a non-api related thing?

mooktakim (approved):

swagger/v3/api_spec.json
```diff
@@ -161,12 +161,12 @@
         }
       }
     },
-    "/api/v3/partnerships/ecf": {
```
cwrw, line 164: in v3 unfortunately we need to amend this manually otherwise it messes with the order

swagger/v3/api_spec.json
```diff
@@ -5726,1303 +2275,1218 @@
...
-            "example": "unknown"
-          },
-          "has_passed": {
-            "description": "Whether the participant has failed or passed",
-            "type": "boolean",
-            "example": null
+              "ecf-partnership-update"
+            ]
           },
-          "lead_provider_name": {
```
cwrw, line 6907: lead provider name seems to be missing from declaration attributes in `ParticipantDeclarationAttributes` in v3, let's watch out for that

cwrw (commented):
thanks @leandroalemao a few minor comments, but worth being careful with v3 specs and doing manually as a few things are missing and the order changed. Happy to chat :)

spec/requests/api/v1/ecf_participants_spec.rb
```diff
@@ -409,18 +409,6 @@
         expect(response.status).to eq 403
       end
     end
-
-    context "when using LeadProviderApiToken with only NPQ access" do
```
cwrw, line 413: does this fail, can we keep it in for now, keen on not touching scopes yet?

cwrw (approved):
thank you @leandroalemao tremendous effort 🙌

mooktakim (approved):

ethax-ross (approved):

PR #5354, opened 2024-12-10
Remove NPQ from finance dashboards

ethax-ross:
[Jira-3813](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3813)

### Context

Now that we have separated NPQ from ECF, we can remove all of the NPQ-related financial dashboard pages and functionality.

- Remove NPQ from finance statements

Remove all NPQ references from the finance statements.

- Remove NPQ from finance schedules

Remove all NPQ schedules from the finance dashboard. In order to do this we change the `cohort#schedules` relationship to only return ECF schedules (as we render these directly in the view).

- Remove NPQ participants from finance dashboard search

The NPQ-specific data has already been removed from the views by an earlier commit. Excluding NPQ users from the results entirely would be difficult as we would need to update queries to determine the type of user by their profile.
Instead, I've just removed `NPQApplication` from the query for now and they will be excluded as a by-product of removing the NPQ-related user data further down the road.

### Guidance to review

I tried to do this in a number of smaller commits, however it was proving difficult to get each commit to a functional state due to how intertwined all the finance dashboard pages are.

As part of this I've also dropped the step to choose ECF/NPQ statements after selecting `View financial statements' as it was a pointless step with only ECF as the choice after removing NPQ.

The 'Search duplicate records' section appears to be ECF-specific so I haven't done anything there.


cwrw (approved):

leandroalemao (approved):
All tested. Tks @ethax-ross 👍

PR #5371, opened 2024-12-16
[CPDLP-3840] Remove NPQ from record declaration (NPQ Post Separation Cleanup)

leandroalemao:
### Context

- Ticket: [CPDLP-3840](https://dfedigital.atlassian.net/browse/CPDLP-3840)

We couldn't remove NPQ logic from the record declaration service due to the complexity and the fact it's integrated into the factories which mean many tests will break. Instead we have to wait for the main cleanup to happen and then we can remove this logic and service.

- Remove NPQ logic from the record declaration service
- Remove NPQ only declaration factories and everywhere they are used
- Ensure the rest of the test suite passes

[CPDLP-3840]: https://dfedigital.atlassian.net/browse/CPDLP-3840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/record_declaration.rb
```diff
@@ -13,7 +13,7 @@ class RecordDeclaration
   attribute :evidence_held
   attribute :has_passed
 
-  before_validation :validate_if_npq_course_supported
+  before_validation :ensure_npq_course_no_longer_supported
```
mooktakim, line 16: We can just remove this method i think

mooktakim: hmm, I'm thinking this is there to prevent any npq declarations from being sent accidentally. ignore my comment

mooktakim (approved):

ethax-ross (commented):
Looks great 👏 do we need a ticket/to remember to remove the qualified teachers API secrets and big query secrets from the keyvault once this ships?

config/sidekiq_cron_schedule.yml
```diff
@@ -33,10 +33,6 @@ remove_school_mentor_job:
   cron: "45 4 * * *"
   class: "RemoveSchoolMentorJob"
   queue: default
-send_outcomes_to_qualified_teachers_api:
```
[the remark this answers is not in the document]
leandroalemao: Think sidekiq reloads the cron after each deploy but I'll manually double check after we deploy it.

spec/services/api/participant_declarations/index_spec.rb
```diff
@@ -96,30 +96,5 @@
         expect(subject.scope).to include(new_provider_ineligible_declaration)
       end
     end
-
-    context "when NPQ application exists for same lead provider" do
```
cwrw, line 100: same here around avoiding building the declaration using the service but actually build it to keep this test until we delete the data? happy to discuss

spec/services/npq/ehco/targeted_delivery_funding_eligibility_updater_spec.rb
```diff
@@ -1,42 +0,0 @@
-# frozen_string_literal: true
-
-require "rails_helper"
-
-RSpec.describe NPQ::Ehco::TargetedDeliveryFundingEligibilityUpdater do
```
cwrw, line 5: can we delete the corresponding service please?

cwrw (commented):
thanks Leandro a few comments, happy to discuss one of the suggestions

app/services/importers/create_cohort.rb
```diff
@@ -38,7 +38,6 @@ def create_cohort(row)
           start_year:,
           registration_start_date: safe_parse(row["registration-start-date"]),
           academic_year_start_date: safe_parse(row["academic-year-start-date"]),
-          npq_registration_start_date: safe_parse(row["npq-registration-start-date"]),
```
cwrw, line 41: can we please avoid this commit and we can have a different ticket for it? as this is adding too much to an already big pr?

spec/services/statements/mark_as_paid_spec.rb
```diff
@@ -18,8 +18,8 @@
       create(:ect_participant_declaration, :eligible, cpd_lead_provider:)
       VoidParticipantDeclaration.new(create(:ect_participant_declaration, :eligible, cpd_lead_provider:)).call
     end
+
     Statements::MarkAsPayable.new(statement).call
-    Finance::Statement::ECF::Payable.find(statement.id)
```
cwrw, line 22: why did this need to be deleted?

cwrw (approved):
thanks Leandro

mooktakim (approved):

config/sidekiq_cron_schedule.yml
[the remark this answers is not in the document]
ethax-ross: Ah I'm thinking of DelayedJob cron in NPQ reg I think 👍

ethax-ross (approved):

PR #5372, opened 2024-12-16
[CPDLP-3815] NPQ Post Separation Cleanup - Remove NPQ API disable feature flag from ECF

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3815

* removed `disable_npq` feature flag.
* removed `NpqApiEndpoint`.


leandroalemao (approved):
Looks good @mooktakim 👍 😄 tks

app/services/record_declaration.rb
```diff
@@ -286,13 +285,4 @@ def create_participant_outcome!
   def check_mentor_completion
     ParticipantDeclarations::HandleMentorCompletion.call(participant_declaration:)
   end
-
-  def validate_if_npq_course_supported
```
ethax-ross, line 290: Do we not want to keep this? 🤔

config/routes.rb
```diff
@@ -198,7 +196,6 @@
     # Keeping the urls to old guidance urls, but they need to lead to new api-reference ones
     get "/guidance/home", to: redirect("/api-reference")
     get "/guidance/ecf-usage", to: redirect("/api-reference/ecf-usage")
```
ethax-ross, line 198: A bunch of these other pages don't seem to exist either, we should maybe ditch the redirects at some point. There are also some translations we could bin here

https://github.com/DFE-Digital/early-careers-framework/blob/main/config/locales/en.yml#L600

spec/models/finance/schedule/ecf_spec.rb
```diff
@@ -32,7 +32,7 @@
     end
   end
 
-  Cohort.where(start_year: 2021..).find_each do |cohort|
+  Cohort.where(start_year: 2021..2024).find_each do |cohort|
```
ethax-ross, line 35: Why did this need to change, of interest?

ethax-ross (commented):
Looks good 👍 I think there's quite a bit of overlap with Leandro's PR, might be worth one branching off the other

app/services/oneoffs/npq/migrate_declarations_between_statements.rb
```diff
@@ -1,180 +0,0 @@
-# frozen_string_literal: true
-
-require "has_recordable_information"
-
-module Oneoffs::NPQ
```
cwrw, line 5: didn't those go with the other pr 🤔

app/services/record_declaration.rb
cwrw: agree to keep

config/routes.rb
cwrw: I think let's keep the other redirects for now as not sure about those for now and they are not NPQ related. 
In terms of other translations I think we need to do another sweep after those tickets to see whats left and break down into tickets to avoid too many changes

cwrw (commented):
looks good, a few minor comments. I agree with @ethax-ross around overlap, let's all chat tomorrow to agree way forward :)

app/services/record_declaration.rb
mooktakim: Yes, will change this, i saw it on Leandro's PR👍

spec/models/finance/schedule/ecf_spec.rb
mooktakim: We discussed this on call. The schedules.csv file creates up to 2024. We will change this after we add 2025 to the csv. I'll add a comment.

app/jobs/participant_outcomes/stream_api_requests_to_big_query_job.rb
```diff
@@ -1,36 +0,0 @@
-# frozen_string_literal: true
-
-module ParticipantOutcomes
-  class StreamApiRequestsToBigQueryJob < ApplicationJob
```
cwrw, line 4: can we remove this from `ParticipantOutcomeApiRequest`?

cwrw (approved):

PR #5378, opened 2024-12-18
Fix GitHub codespaces

ethax-ross:
[Jira-3847](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3847)

### Context

The GitHub codespace for ECF is not building/running. We want to fix it so that we content editors can use it to make changes easily.

- Fix GitHub codespaces

Update to bring it inline with NPQ registration codespace configuration, as this contains various fixes to get codespaces to run and the applications are very similar. Notable fixes:

- Expose Postgres on the correct/default port
- Setup Postgres user for `vscode`
- Manually switch bundler version prior to `bundle install`
- Build API guidance as part of `boot.sh`
- Disable CSRF protection (for some reason it doesn't work in codespaces)
- Boot the app with foreman so we get JS/CSS watching

### Guidance to review

Switch to the branch in GitHub and boot up a new codespace, check that:

- Rails boots/is accessible once finished.
- API guidance works.
- Logging in works.
- Changing some CSS/JS then refreshing updates the page.


leandroalemao (approved):
@ethax-ross Tested here and looks like it's working well. 👏  tks

cwrw (approved):
🙌

PR #5379, opened 2024-12-19
Update `ParticipantDeclarationsQuery` to include `completed` previous

ethax-ross:
[Jira-3880](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CPDLP/boards/87?assignee=712020%3A5b5839dc-e4d6-4df5-9dd6-627cc49fe345%2Cunassigned&selectedIssue=CPDLP-3880)

### Context

Currently, the `ParticipantDeclarationsQuery` filters declarations based only on an `active` induction_status. This prevents declarations from being surfaced for participants with a `completed` induction status.

- Update `ParticipantDeclarationsQuery` to include `completed` previous declarations

Update the `ParticipantDeclarationsQuery` to include `completed` declarations for a previous provider in addition to `active`.

### Guidance to review

TODO: testing

### DIFF
--- a/app/services/api/v3/participant_declarations_query.rb
+++ b/app/services/api/v3/participant_declarations_query.rb
@@ -6,7 +6,7 @@ module Api
       include Api::Concerns::FilterCohorts
       include Api::Concerns::FilterUpdatedSince
 
-      RELEVANT_INDUCTION_STATUS = %w[active completed].freeze
+      RELEVANT_INDUCTION_STATUS = Api::ParticipantDeclarations::Index::RELEVANT_INDUCTION_STATUS
 
       attr_reader :cpd_lead_provider, :params
 
--- a/app/services/api/v3/participant_declarations_query.rb
+++ b/app/services/api/v3/participant_declarations_query.rb
@@ -6,7 +6,7 @@ module Api
       include Api::Concerns::FilterCohorts
       include Api::Concerns::FilterUpdatedSince
 
-      RELEVANT_INDUCTION_STATUS = Api::ParticipantDeclarations::Index::RELEVANT_INDUCTION_STATUS
+      RELEVANT_INDUCTION_STATUS = %w[active completed].freeze
 
       attr_reader :cpd_lead_provider, :params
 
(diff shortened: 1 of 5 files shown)


app/services/api/v3/participant_declarations_query.rb
```diff
@@ -81,7 +83,7 @@ def ecf_previous_declarations_scope
           )
           .where.not(participant_profile: { induction_records: { induction_programmes: { partnership_id: nil } } }) # Exclude any nil partnerships (eg: cip)
           .where(participant_profile: { induction_records: { induction_programme: { partnerships: { lead_provider_id: lead_provider&.id } } } })
-          .where(participant_profile: { induction_records: { induction_status: "active" } }) # only want induction records that are the winning latest ones
+          .where(participant_profile: { induction_records: { induction_status: RELEVANT_INDUCTION_STATUS } }) # only want induction records that are the winning latest ones
```
cwrw, line 86: can we also add this to v1/v2 please in `app/services/api/participant_declarations/index.rb`?

cwrw (approved):

app/services/api/v3/participant_declarations_query.rb
```diff
@@ -6,6 +6,8 @@ class ParticipantDeclarationsQuery
       include Api::Concerns::FilterCohorts
       include Api::Concerns::FilterUpdatedSince
 
+      RELEVANT_INDUCTION_STATUS = %w[active completed].freeze
```
leandroalemao, line 9: Maybe set this higher up or in a model maybe? so we don't need to declare `RELEVANT_INDUCTION_STATUS` twice in difference places?

mooktakim: I'll set it to `Api::ParticipantDeclarations::Index::RELEVANT_INDUCTION_STATUS` that way its shared.

cwrw: I wonder if we should avoid that as those are services that support different versions of the API, and if we disable a version we would like to avoid extra dependencies? Maybe we can live with duplication or maybe moving it to the declaration model?

mooktakim: I would prefer to keep the duplication instead of moving to declaration model, as its within the context.

cwrw: I agree, @leandroalemao what do you think?

leandroalemao (approved):

PR #5380, opened 2024-12-19
Spike: mentor declarations using STI

ethax-ross:
Spike to look into the necessary changes to implement the notion of 'mentor declarations' using the existing STI modelling.


ethax-ross:
Closing to fix review app

PR #5386, opened 2024-12-19
[CPDLP-3843] Remove token access for NPQ (NPQ Post Separation Cleanup)

leandroalemao:
### Context

- Ticket: [CPDLP-3843](https://dfedigital.atlassian.net/browse/CPDLP-3843)

Now we’ve removed all the data we need to remove access to NPQ tokens in ECF in the API.

Remove the NPQ API token setup from the ECF API.

[CPDLP-3843]: https://dfedigital.atlassian.net/browse/CPDLP-3843?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/concerns/api_token_authenticatable.rb
```diff
@@ -50,6 +50,6 @@ def check_access_scope
   end
 
   def access_scope
```
cwrw, line 52: if we have a look at controllers like `DataStudio::SchoolRolloutController` these have:
```
def access_scope
  ApiToken.where(private_api_access: true)
end
```
Which we want to keep. Can we keep this as is and keep the level the same everywhere and just remove the NPQ access please instead?

db/legacy_seeds/dummy_structures.rb
```diff
@@ -238,15 +238,11 @@
   # The tokens below have different unhashed version to avoid worrying about clever cryptographic attacks
   if Rails.env.review?
     EngageAndLearnApiToken.find_or_create_by!(hashed_token: "dfce9a34c6f982e8adb4b903f8b6064682e6ad1f7858c41ed8a0a7468abc8896")
-    NPQRegistrationApiToken.find_or_create_by!(hashed_token: "1dae3836ed90df4b796eff1f4a4713247ac5bc8a00352ea46eee621d74cd4fcf")
```
cwrw, line 241: maybe we can also delete `NPQRegistrationApiToken` class as well? let me know what you think

cwrw (approved):

PR #5374, opened 2024-12-18
[CPDLP-3814] NPQ Post Separation Cleanup - Remove NPQ specific tables from ECF

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3814

### Changes proposed in this pull request

* Remove the following models and tables:
  * NPQ applications
  * NPQ contracts
  * NPQ courses
  * NPQ lead providers
  * NPQ application exports

PR #5401, opened 2025-01-03
[CPDLP-3818] Remove data in shared ecf declaration tables for npq

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3818
- Branched off: https://github.com/DFE-Digital/early-careers-framework/pull/5374

* Removed and dropped table `ParticipantOutcomeApiRequest`
* Removed and dropped table `ParticipantOutcome::NPQ`
* With migration deleted NPQ declarations data from:
  * `DeclarationState`
  * `ParticipantDeclarationAttempt`
  * `ParticipantDeclaration`


leandroalemao (commented):
@mooktakim All looks good.. a minor comment on cleaning [en.yml](https://github.com/DFE-Digital/early-careers-framework/blob/main/config/locales/en.yml#L186) file as well if you want.

app/jobs/detect_sidekiq_metrics_issues_job.rb
```diff
@@ -3,7 +3,7 @@
 class DetectSidekiqMetricsIssuesJob
   include Sidekiq::Worker
 
-  SIDEKIQ_QUEUE_NAMES = %w[priority_mailers mailers default participant_outcomes big_query].freeze
```
cwrw, line 6: maybe we can remove outcomes from fixtures? `fixtures/files/api_reference/api_spec.json`

cwrw (approved):
looks good just one minor comment

app/validators/future_date_validator.rb
```diff
@@ -3,7 +3,7 @@
 class FutureDateValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     if value && value > Time.zone.now
-      record.instance_of?(::ParticipantOutcome::NPQ) ? record.errors.add(attribute, I18n.t(:future_date_outcomes, attribute:)) : record.errors.add(attribute, I18n.t(:future_date, attribute:))
```
ethax-ross, line 6: I think we can drop `future_date_outcomes` from the translations file

ethax-ross: There are also some other translations we can probably bin:

https://github.com/DFE-Digital/early-careers-framework/blob/main/config/locales/en.yml#L186
https://github.com/DFE-Digital/early-careers-framework/blob/main/config/locales/en.yml#L706

Maybe we just do a sweep of these separately though 🤔

ethax-ross (approved):

app/validators/future_date_validator.rb
cwrw: I think those are in the sweep later so its ok :)

PR #5402, opened 2025-01-03
[CPDLP-3899] Ensure Void/Show declarations are scoped to ECF declarations

leandroalemao:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3899

As part of our clean up of NPQ in ECF, we are scoping all endpoints and queries to ECF only. We need to also scope void/show declaration to ECF as well.

- For voiding declarations in v1/v2/v3 scope the queries to only allow voiding ECF declarations only and not NPQ;
- Tried to replace `participant_declaration_for_lead_provider` by `participant_declaration_query_scope` but the queries of the second are slightly different taking in consideration `induction records` & `partnerships` as well. That would allow past lead providers to void declarations currently with other lead providers;
- Single declaration endpoints are already scoped by ECF;

### DIFF
--- a/app/controllers/api/v1/participant_declarations_controller.rb
+++ b/app/controllers/api/v1/participant_declarations_controller.rb
@@ -71,7 +71,7 @@ module Api
       end
 
       def participant_declaration_for_lead_provider
-        @participant_declaration_for_lead_provider ||= ParticipantDeclaration.for_lead_provider(cpd_lead_provider).find(params[:id])
+        @participant_declaration_for_lead_provider ||= ParticipantDeclaration::ECF.for_lead_provider(cpd_lead_provider).find(params[:id])
       end
 
       def permitted_params
--- a/app/controllers/api/v3/participant_declarations_controller.rb
+++ b/app/controllers/api/v3/participant_declarations_controller.rb
@@ -99,7 +99,7 @@ module Api
       end
 
       def participant_declaration_for_lead_provider
-        @participant_declaration_for_lead_provider ||= ParticipantDeclaration.for_lead_provider(cpd_lead_provider).find(params[:id])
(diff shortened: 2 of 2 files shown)


app/controllers/api/v1/participant_declarations_controller.rb
```diff
@@ -71,7 +71,7 @@ def participant_declaration_query_scope
       end
 
       def participant_declaration_for_lead_provider
-        @participant_declaration_for_lead_provider ||= ParticipantDeclaration.for_lead_provider(cpd_lead_provider).find(params[:id])
+        @participant_declaration_for_lead_provider ||= ParticipantDeclaration::ECF.for_lead_provider(cpd_lead_provider).find(params[:id])
```
cwrw, line 74: can we add specs for this please?

leandroalemao: There's already specs for this scenario.. any other scenario you have in mind so I can add specs?

https://github.com/DFE-Digital/early-careers-framework/blob/main/spec/requests/api/v1/participant_declarations_spec.rb#L922
https://github.com/DFE-Digital/early-careers-framework/blob/main/spec/requests/api/v2/participant_declarations_spec.rb#L640
https://github.com/DFE-Digital/early-careers-framework/blob/main/spec/requests/api/v3/participant_declarations_spec.rb#L776

ethax-ross: Is it to ensure NPQ declarations are no longer returned maybe? 🤔 I'm not sue its worth adding specs for those given we're nuking the data (or have already removed it I'm not sure?)

ethax-ross (approved):

app/controllers/api/v1/participant_declarations_controller.rb
cwrw: yea fair enough if NPQ is going we can't really test it

cwrw (approved):

mooktakim (approved):

PR #5412, opened 2025-01-08
Prevent retained declarations for mentors in funded cohorts

ethax-ross:
[Jira-3907](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3907)

### Context

We have a new `mentor_funding` attribute on `Cohort`. Going forward we want to prevent mentors from adding `retained-x` declarations to a `mentor_funding` cohort. We should continue to allow `retained-x` declarations for ECTs.

- Prevent retained declarations for mentors in funded cohorts

Update `RecordDeclaration` with appropriate validation.

### Guidance to review

Error message updated to Mark's copy.

### DIFF
--- a/app/services/record_declaration.rb
+++ b/app/services/record_declaration.rb
@@ -30,6 +30,7 @@ class RecordDeclaration
   validate :output_fee_statement_available
   validate :validate_milestone_exists
   validate :validates_billable_slot_available
+  validate :validate_only_started_or_completed_if_mentor
   validates :course_identifier, course: true
   validates :cpd_lead_provider, induction_record: true
 
@@ -203,6 +204,14 @@ private
 
   def original_participant_declaration
     @original_participant_declaration ||= participant_declaration.duplicate_declarations.order(created_at: :asc).first
+  end
+
+  def validate_only_started_or_completed_if_mentor
+    return if declaration_type&.in?(%w[started completed])
+    return unless participant_profile && participant_profile.mentor?
+    return unless cohort && cohort.mentor_funding?
+
+    errors.add(:declaration_type, I18n.t(:retained_declaration_not_allowed_for_mentor))
   end
 
   def validates_billable_slot_available
--- a/spec/services/record_declaration_spec.rb
+++ b/spec/services/record_declaration_spec.rb
@@ -331,9 +331,8 @@ RSpec.describe RecordDeclaration do
     }
   end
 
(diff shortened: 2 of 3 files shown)


app/services/record_declaration.rb
```diff
@@ -205,6 +206,14 @@ def original_participant_declaration
     @original_participant_declaration ||= participant_declaration.duplicate_declarations.order(created_at: :asc).first
   end
 
+  def validate_not_retained_if_mentor
+    return unless declaration_type&.match?(/retained-\d/)
+    return unless participant_profile && participant_profile.mentor?
+    return unless cohort && cohort.mentor_funding?
```
cwrw, line 212: should this come first just to shorten the if statements we need to go through?

spec/services/record_declaration_spec.rb
```diff
@@ -418,6 +430,26 @@
       it_behaves_like "creates a participant declaration"
       it_behaves_like "creates participant declaration attempt"
       it_behaves_like "checks for mentor completion event"
+
+      context "when the declaration_type is retained" do
```
cwrw, line 434: can we add a spec for started or completed to show they go through? (unless that's already tested above with the flag

cwrw (commented):
looks good just some minor comments

app/services/record_declaration.rb
```diff
@@ -205,6 +206,14 @@ def original_participant_declaration
     @original_participant_declaration ||= participant_declaration.duplicate_declarations.order(created_at: :asc).first
   end
 
+  def validate_not_retained_if_mentor
+    return unless declaration_type&.match?(/retained-\d/)
```
leandroalemao, line 210: Are we allowing extended declarations to be declared as well?

app/services/record_declaration.rb
ethax-ross: Ah actually we can't do this as cohort delegates to participant profile -> schedule and if the profile is `nil` it throws an error

app/services/record_declaration.rb
ethax-ross: I believe so, @cwrw can you confirm?

cwrw: ah good catch Leandro, we only allow started and completed declarations for mentors 2025, so everything else should raise an error

ethax-ross: I'll update the PR 👍

leandroalemao (approved):

PR #5429, opened 2025-01-10
Switch to Mentor and ECT types for ParticipantDeclaration 

ethax-ross:
[Jira-3897](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3897)

### Context

The original/complete PR is #5404.
Depends on #5428.

Now that we have `temp_type` populating and backfilled, we can switch to use `Mentor` and `ECT` declaration sub-types, backfilling the `type` field quickly from the `temp_type` field to avoid downtime/data inconsistencies.

- Add Mentor and ECT types for ParticipantDeclaration

Going forward we need to differentiate between Mentor and ECT declarations. In order to do this we are leveraging the existing STI setup and adding two new classes that inherit from the `ParticipantDeclaration::ECF` class.

- Update specs and factories for new declaration types

Update specs and factories to use the new declaration types so that they are correctly referenced through the new relationship types.

- Fix assurance report to return ECT and Mentor

Previously this was scoped to just ECF declarations; we now need to scope it to include both ECT and Mentor declarations and include the `type` so that Rails correctly type-casts them in the results.

- Update RecordDeclaration to use correct declaration type

When creating a declaration we go through the relationship so it will by typed correctly as a by-product. The only place we need to specify the declaration type explicitly is when validating that there is a billable slot available.

- Add migration to populate type from temp_type

Add a migration to populate the `type` field from the backfilled/populated `temp_type`, which will contain the `ECT` or `Mentor` declaration type.

- Add callbacks and validation to ensure profile/declaration type consistency

Add a callback to `ParticipantProfile::ECF` to ensure if the type is changed then the declarations also reflect the correct type.

Add validation to `ParticipantDeclaration::ECF` to ensure that their type cannot be changed, resulting in a mismatch with the `ParticipantProfile`.

- Switch inheritance column to temp_type for ECF declarations

Until we can safely backfill the `type` column, we are using `temp_type`, which has already been backfilled and is kept up to date.

### Guidance to review

Once this has shipped we can remove the `temp_type` field. #5433

### DIFF
--- a/app/models/participant_declaration/ecf.rb
+++ b/app/models/participant_declaration/ecf.rb
@@ -3,7 +3,7 @@
 class ParticipantDeclaration::ECF < ParticipantDeclaration
   has_many :statements, class_name: "Finance::Statement::ECF", through: :statement_line_items
 
-  before_save :set_temp_type
+  validate :validate_against_profile_type
 
   def ecf?
     true
@@ -16,10 +16,10 @@ class ParticipantDeclaration::ECF < ParticipantDeclaration
       (sparsity_uplift || pupil_premium_uplift)
   end
 
-  def set_temp_type
+  def validate_against_profile_type
     return unless participant_profile
-    return unless temp_type.nil?
+    return if participant_profile.type.demodulize == type.demodulize
 
-    self.temp_type = participant_profile.type.sub("ParticipantProfile", "ParticipantDeclaration")
+    errors.add(:type, I18n.t(:declaration_type_must_match_profile_type))
   end
 end
--- a/app/models/participant_declaration/ecf.rb
+++ b/app/models/participant_declaration/ecf.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class ParticipantDeclaration::ECF < ParticipantDeclaration
+  self.inheritance_column = :temp_type
+
   has_many :statements, class_name: "Finance::Statement::ECF", through: :statement_line_items
 
   validate :validate_against_profile_type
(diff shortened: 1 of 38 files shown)


leandroalemao (approved):
Hi @ethax-ross Looks good tks 👍

app/models/participant_profile/ecf.rb
```diff
@@ -203,8 +203,8 @@ def sync_status_with_induction_record
     induction_record&.update!(mentor_profile:) if saved_change_to_mentor_profile_id?
   end
 
-  def update_declaration_temp_types!
-    participant_declarations.update_all(temp_type: type.sub("ParticipantProfile", "ParticipantDeclaration"))
+  def update_declaration_types!
+    participant_declarations.update_all(type: type.sub("ParticipantProfile", "ParticipantDeclaration"))
```
cwrw, line 207: I think this should be temp type 🤔

cwrw: or actually we should be changing both

app/models/participant_declaration/ecf.rb
```diff
@@ -16,10 +18,10 @@ def uplift_paid?
       (sparsity_uplift || pupil_premium_uplift)
   end
 
-  def set_temp_type
```
cwrw, line 19: shouldn't we continue setting temp type, otherwise how will new declarations get this value set 🤔

app/models/participant_profile/ecf.rb
ethax-ross: yeah this is a mistake it should be `temp_type`, I'll update. We don't need to update `type` as we're going to backfill over it anyway in the next PR

cwrw (approved):

mooktakim (approved):

PR #5433, opened 2025-01-13
Add migration to populate type from temp_type

ethax-ross:
[Jira-3897](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3897)

### Context

The original/complete PR is #5404.
Depends on #5429.

Now that we are using the `temp_type` field, we can backfill the `type` field and maintain it ahead of finally switching over.

- Add migration to populate type from temp_type

Add a migration to populate the `type` field from the backfilled/populated `temp_type`, which will contain the `ECT` or `Mentor` declaration type.

- Populate type when declaration changes

To ensure `type` remains up to date while we are still referencing `temp_type` we need to update it `after_commit`.

### Guidance to review

Once this has shipped we can switch back to the `type` field and remove the `temp_type` field. #5430

### DIFF
--- a/app/models/participant_declaration/ecf.rb
+++ b/app/models/participant_declaration/ecf.rb
@@ -7,6 +7,8 @@ class ParticipantDeclaration::ECF < ParticipantDeclaration
 
   validate :validate_against_profile_type
 
+  before_save :populate_type
+
   def ecf?
     true
   end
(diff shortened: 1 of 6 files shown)


mooktakim (approved):

leandroalemao (approved):

cwrw (approved):
🙌

PR #5447, opened 2025-01-16
Add workflow for refreshing the Knapsack manifest

ethax-ross:
[Jira-3946](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3946)

### Context

As we add/remove tests the manifest becomes gradually outdated and as a result Knapsack can't distribute the test suite across workers as efficiently.

We want to add a workflow to recompute the Knapsack manifest and raise a pull request with the changes. This can be triggered manually and will also run on a weekly schedule.

- Add workflow for refreshing the Knapsack manifest

The action to create a pull request also supports assigning a team, but in order to do this you need to generate a personal access token. To avoid coupling the action to a user account I haven't done this, but if we have access to a generic 'dfe user' we could do this in the future.

### Guidance to review

Example pull request: https://github.com/DFE-Digital/early-careers-framework/pull/5444

If an existing PR is already open from the previous week it will update it rather than creating another PR.

Action details for raising a PR: https://github.com/marketplace/actions/create-pull-request

Example CI run: https://github.com/DFE-Digital/early-careers-framework/actions/runs/12792397881/job/35662752136

### DIFF
--- a/.github/workflows/refresh_knapsack_manifest.yml
+++ b/.github/workflows/refresh_knapsack_manifest.yml
@@ -1,0 +1,57 @@
+name: "Refresh Knapsack manifest"
+
+on:
+  schedule:
+    - cron: '0 8 * * 1' # Monday at 8am
+  workflow_dispatch: 
+
+jobs:
+  refresh_manifest:
+    name: Refresh Knapsack manifest
+    runs-on: ubuntu-20.04
+
+    env:
+      RAILS_ENV: test
+      DB_USERNAME: postgres
+      DB_PASSWORD: ""
+      DB_HOSTNAME: 127.0.0.1
+      DB_PORT: 5432
+
+    services:
+      postgres:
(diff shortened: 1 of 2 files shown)


cwrw (approved):

.github/workflows/refresh_knapsack_manifest.yml
```diff
@@ -0,0 +1,58 @@
+name: "Refresh Knapsack manifest"
+
+on:
+  schedule:
+    - cron: '0 8 * * 1' # Monday at 8am
+  workflow_dispatch: 
+  push: # Temp for testing
```
leandroalemao, line 7: All good @ethax-ross 👍 Just a reminder to remove this line before merging.. tks 😄

leandroalemao (approved):

mooktakim (approved):

PR #5449, opened 2025-01-16
[CPDLP-3934] Add no uplift logic to call off contracts

leandroalemao:
### Context

- Ticket: [CPDLP-3934](https://dfedigital.atlassian.net/browse/CPDLP-3934)

We would like to drive no uplifts in 2025 cohort for ECTs/Mentors by setting the uplift db field `uplift_amount` to nil.

- Add no uplift logic to call off contracts
- Tweak `Importers::CreateCallOffContract`
- Tweak `Oneoffs::UpdateContracts`

[CPDLP-3934]: https://dfedigital.atlassian.net/browse/CPDLP-3934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/finance/ecf/statement_calculator.rb
+++ b/app/services/finance/ecf/statement_calculator.rb
@@ -144,6 +144,8 @@ module Finance
       end
 
       def uplift_additions_count
+        return 0.0 unless statement.contract.include_uplift_fees?
+
         output_calculator.uplift_breakdown[:additions]
       end
 
@@ -165,11 +167,9 @@ module Finance
         previous_uplift_count = output_calculator.uplift_breakdown[:previous_count]
         previous_uplift_amount = previous_uplift_count * uplift_fee_per_declaration
 
-        delta_uplift_count = output_calculator.uplift_breakdown[:count]
-
         # uplift_clawback_deductions is a negative number so doing a double negative --
         # we're adding it back as we had subtracted from adjustments_total
-        delta_uplift_amount = delta_uplift_count * uplift_fee_per_declaration - uplift_clawback_deductions
+        delta_uplift_amount = uplift_count * uplift_fee_per_declaration - uplift_clawback_deductions
 
         available = [(statement.contract.uplift_cap - previous_uplift_amount), 0].max
 
--- a/app/services/finance/ecf/statement_calculator.rb
+++ b/app/services/finance/ecf/statement_calculator.rb
@@ -148,14 +148,20 @@ module Finance
       end
 
       def uplift_deductions_count
+        return 0 unless statement.contract.include_uplift_fees?
+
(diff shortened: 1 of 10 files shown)


app/services/importers/create_call_off_contract.rb
```diff
@@ -79,17 +79,19 @@ def rows
...
-        uplift_amount: contract_data[:uplift_amount],
         recruitment_target: contract_data[:recruitment_target],
         revised_target: contract_data[:revised_target],
         set_up_fee: contract_data[:set_up_fee],
         monthly_service_fee: contract_data[:monthly_service_fee],
         raw: contract_data.to_json,
-      )
+      }
+
+      attributes.merge!(uplift_amount: contract_data[:uplift_amount]) unless CallOffContract::COHORTS_WITH_NO_UPLIFT.include?(cohort.start_year)
```
mooktakim, line 93: I'm not sure about coupling this to `COHORTS_WITH_NO_UPLIFT`
The CSV will set a nil value for `uplift_amount` and therefore will disable it.

cwrw: I think I agree, the `nil` value will be the flag instead of the global variable, let us know your thoughts :)

leandroalemao: Agree.. but I had to keep `COHORTS_WITH_NO_UPLIFT` in the importer context at least so we know to not include uplift fees for 2025 cohort when creating contracts seed data without a CSV.

app/services/finance/ecf/statement_calculator.rb
```diff
@@ -148,14 +148,20 @@ def uplift_additions_count
       end
 
       def uplift_deductions_count
+        return 0 unless statement.contract.include_uplift_fees?
```
ethax-ross, line 151: The tests covering this and line 163 pass without the guard statement; is there a way we can isolate them in the spec?

app/services/finance/ecf/statement_calculator.rb
```diff
@@ -148,14 +148,20 @@ def uplift_additions_count
       end
```
cwrw, line 148: we have uplift additions count and uplift count, do we use those anywhere?

cwrw: nvm all good

cwrw (approved):

app/services/finance/ecf/statement_calculator.rb
ethax-ross: Not sure this is fixed? if you comment out lines 153/165 I would expect tests to fail but they don't

PR #5461, opened 2025-01-21
[CPDLP-3913] Remove any remnants of NPQ in ECF app

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3913

* Removed REMOVE_NPQ_REFERENCES and related
* Removed NPQ css
* Removed NPQ form services
* Removed NPQ logic from Cohort
* Removed NPQ locales
* Removed NPQ payment calculators
* From rails credentials, removed:
  * `QUALIFIED_TEACHERS_API_URL`
  * `QUALIFIED_TEACHERS_API_KEY`
* Added `npq_registration_start_date` to ignored columns for `Cohort`
* Removed feature `npq_capping`


app/models/cohort.rb
```diff
@@ -22,10 +21,6 @@ def self.active_registration_cohort
     where(registration_start_date: ..Date.current).order(start_year: :desc).first
   end
 
-  def self.active_npq_registration_cohort
-    where(npq_registration_start_date: ..Date.current).order(start_year: :desc).first.presence || current
```
leandroalemao, line 26: Will we drop `npq_registration_start_date` from `cohorts` table as well?

leandroalemao (commented):
Would it be nice to remove feature flag as well?

`disable_npq`
`npq_capping`

app/models/cohort.rb
```diff
@@ -3,7 +3,6 @@
 class Cohort < ApplicationRecord
   has_paper_trail
 
-  NPQ_PLUS_1_YEAR = 2020
```
cwrw, line 6: I wonder if instead of removing this we should rename it, as not sure we should remove this logic 🤔 let me know you're thoughts

app/models/cohort.rb
mooktakim: We're keeping the columns for now, just removing the logic.

app/models/cohort.rb
mooktakim: We're only using this for `npq_plus_one_or_earlier?` and used in `app/forms/schools/add_participants/base_wizard.rb`

mooktakim:
> Would it be nice to remove feature flag as well?
> 
> `disable_npq` `npq_capping`

good question about the feature flags. For now left it. Will discuss

app/models/cohort.rb
cwrw: I think maybe removing the column might be a good shout, let's discuss tomorrow

app/models/cohort.rb
cwrw: Since we are using it in the base wizard I wonder if we keep this logic which says if the cohort is equal to or less than 2020 we default to the current cohort, since it's used on the ECF side. Just as an extra guard maybe incase there is some bad ecf data somewhere?

cwrw (commented):
thanks Mook a few minor comments, also maybe we can look to remove some NPQ content in `app/views/sandbox/show.html.erb`, `db/new_seeds/base/add_lead_providers_and_cips.rb`, otherwise looking great :)

ethax-ross (commented):
Looks good, couple of files I think can go:

https://github.com/DFE-Digital/early-careers-framework/blob/main/spec/fixtures/files/api_reference/api_spec_no_npq.json
https://github.com/DFE-Digital/early-careers-framework/blob/main/spec/fixtures/files/api_reference/api_spec.json

And a few references kicking about still (these are just a few but there are more on searching):

https://github.com/DFE-Digital/early-careers-framework/blob/1931220555219c072cfe0de17cf656f7d7ce95b1/spec/factories/participant_identity.rb#L10-L12
https://github.com/DFE-Digital/early-careers-framework/blob/1931220555219c072cfe0de17cf656f7d7ce95b1/app/models/participant_identity.rb#L12
https://github.com/DFE-Digital/early-careers-framework/commit/cdb02bcd47448db0fdaaad76800697246a53f53b

cwrw:
I was thinking we leave the participant identity origin, as it will still exist and we don't intent on ever updating it/backfilling it

app/models/cohort.rb
mooktakim: Reverted the change. I renamed the method `npq_plus_one_or_earlier?` to `start_year_2020_or_earlier?`
Let me know what you think.

cwrw (approved):
thanks Mook!

PR #5476, opened 2025-01-24
Switch from Selenium to Cuprite

ethax-ross:
> ⚠️ WIP

### Context

We have noticed that our CI is hanging randomly on the feature specs that are using the Javascript Selenium webdriver. It appears to be a HTTP timeout that Selenium experiences, though its not easy to replicate or debug.

- Switch from Selenium to Cuprite

In an effort to fix the issue we can move to Cuprite, which is a pure Ruby driver for Capybara. An issue with doing this is that the latest versions of `axe-core-api` have not been updated for Cuprite support (they expect Selenium). It does, however, work with versions prior to `4.3` so we can pin the version.

### Guidance to review

This should get rid of the flakiness as its removing Selenium, which is where it falls down. I'm still not clear on why Selenium is flakey though.

There are some nuances with how Cuprite behaves so the tests have minor updates to reflect that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -173,14 +173,15 @@ group :development do
 end
 
 group :test do
+  gem "axe-core-api", "4.2"
   gem "axe-core-rspec"
   gem "capybara", "~> 3.40"
+  gem "cuprite"
   gem "jsonapi-rspec"
   gem "launchy"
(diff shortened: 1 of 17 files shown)


ethax-ross:
Closing as cuprite seems flakey

PR #5478, opened 2025-01-27
Attempt to fix/mitigate flakey Capybara feature specs

ethax-ross:
### Context

We are finding that our tests hang in CI randomly on a consistent error:

```
Net::ReadTimeout:
       Net::ReadTimeout with #<TCPSocket:(closed)>
```

We want to investigate why this happens and see if we can fix it to make our CI more reliable.

- Minor tweaks to Capybara/chrome settings

In an attempt to fix/reduce the impact of feature spec failures where we get a `Net:HTTP` timeout error on multiple tests, resulting in a long time until the suite fails this commit will:

```
- Add additional args to reduce flakiness (based on various GitHub issues around similar problems)
- Increase the client open/read timeout
- Remove the allowed external domain (in case its rate limiting us/causing the timeouts)
- Increase the default Capybara wait time
```

- Split feature specs in CI and fail fast

So that we aren't waiting a long time to see a failure when we get a random Net:HTTP timeout issue, we can split the feature specs and pass `--fail-fast` so that the job will fail on the first feature spec that fails. Otherwise we can end up waiting a long time for each test to fail and as a result the workflow won't fail for over an hour.

### Guidance to review

This is a bit of a 'throw various fixes at it and see what sticks' approach, as its not easy to replicate or debug.

I've re-ran the rspec tests 3 times without failure.

If we find its still flakey after this, it might be worth using a gem like `rspec-retry` to retry the javascript feature specs (this could be a sticking plaster fix, but given we're decommissioning ECF shortly it could be a quick win to keep us going/reduce the impact of the issue).

### DIFF
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -7,16 +7,16 @@ on:
         type: boolean
         required: false
         default: true
-
+        
 jobs:
-  backend-tests:
+  tests:
     name: Run rspec
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
-        ci_node_total: [8]
-        ci_node_index: [0, 1, 2, 3, 4, 5, 6, 7]
+        ci_node_total: [6]
+        ci_node_index: [0, 1, 2, 3, 4, 5]
 
     env:
       RAILS_ENV: test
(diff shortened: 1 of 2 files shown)


leandroalemao (approved):

.github/workflows/rspec.yml
```diff
@@ -28,9 +28,59 @@ jobs:
...
+
+      - name: Run tests
+        env:
+          CI_NODE_TOTAL: ${{ matrix.ci_node_total }}
+          CI_NODE_INDEX: ${{ matrix.ci_node_index }}
+        run: |-
+          bundle exec rake "knapsack:rspec[--tag ~type:feature]"
+
+  feature-tests:
+    name: Run rspec
```
cwrw, line 63: can we name this run rspec feature, to distinguish the two?

cwrw (approved):

mooktakim (approved):

PR #5486, opened 2025-01-27
Send correct school name in withdrawn participant email

ethax-ross:
[Jira-3979](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3979)

### Context

When we withdraw a participant we send an email with details of the school they were withdrawn from. Currently, however, we take the participant profile -> school, which is not always the school relevant to the withdraw action. We should instead be taking the school from the relevant induction record, in case the participant has transferred schools in the past.

- Send correct school name in withdrawn participant email

Update `WithdrawParticipant` action to take the school from the `relevant_induction_record` instead of the `participant_profile`.

### Guidance to review

There is a temp commit that adds a GOV.UK notify API key for testing to the review credentials.

There are scripts attached to the ticket to setup the test scenario.

### DIFF
--- a/app/services/withdraw_participant.rb
+++ b/app/services/withdraw_participant.rb
@@ -32,10 +32,10 @@ class WithdrawParticipant
       participant_profile.training_status_withdrawn!
     end
 
-    induction_coordinator = participant_profile.school.induction_coordinator_profiles.first
+    induction_coordinator = relevant_induction_record.school.induction_coordinator_profiles.first
     if induction_coordinator.present?
       SchoolMailer.with(
-        withdrawn_participant: participant_profile,
+        induction_record: relevant_induction_record,
         induction_coordinator:,
         partnership: relevant_induction_record.partnership,
       ).fip_provider_has_withdrawn_a_participant.deliver_later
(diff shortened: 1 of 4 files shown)


mooktakim (approved):

app/services/withdraw_participant.rb
```diff
@@ -32,10 +32,10 @@ def call
       participant_profile.training_status_withdrawn!
     end
 
-    induction_coordinator = participant_profile.school.induction_coordinator_profiles.first
+    induction_coordinator = relevant_induction_record.school.induction_coordinator_profiles.first
     if induction_coordinator.present?
       SchoolMailer.with(
-        withdrawn_participant: participant_profile,
+        withdrawn_participant: relevant_induction_record,
```
cwrw, line 38: should we rename this param to induction record rather than withdrawn participant?

spec/mailers/school_mailer_spec.rb
```diff
@@ -314,14 +314,11 @@
     let(:participant_profile) { create(:ect_participant_profile, training_status: "withdrawn", school_cohort:, user: create(:user, email: "john.clemence@example.com")) }
     let(:sit_profile) { create(:induction_coordinator_profile, schools: [school_cohort.school]) }
     let(:partnership) { create(:partnership) }
-
-    before do
-      create(:induction_record, participant_profile:, partnership:)
-    end
+    let!(:induction_record) { create(:induction_record, participant_profile:, partnership:) }
```
cwrw, line 317: just to confirm this is linked to a different school to the profile?

ethax-ross: Yeah its different by default; I'll make it explicit in the spec though in case that changes in the future

cwrw (approved):

cwrw (approved):

leandroalemao (approved):
Hi @ethax-ross Looking good tks

PR #5487, opened 2025-01-27
[CPDLP-3968] Amend void declarations service to include ineligible statement line items

leandroalemao:
### Context

- Ticket: [CPDLP-3968](https://dfedigital.atlassian.net/browse/CPDLP-3968)

Contract managers are not able to see voids for ineligible declarations in the CSV extracts.

- Amend void declarations service to include `ineligible` statement line items, so the record state is updated to `voided`;
- Removed `submitted` line items from the logic as it's not a valid state for `StatementLineItem`. Checked prod and there's no `submitted` line itens currently in the db;
- Improved tests coverage to cover all 3 states in the logic;

[CPDLP-3968]: https://dfedigital.atlassian.net/browse/CPDLP-3968?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/services/void_participant_declaration_spec.rb
+++ b/spec/services/void_participant_declaration_spec.rb
@@ -52,22 +52,6 @@ RSpec.describe VoidParticipantDeclaration do
       end
     end
 
-    context "when declaration is payable" do
-      let(:participant_declaration) do
-        create(
-          :ect_participant_declaration,
-          :payable,
-          cpd_lead_provider:,
-          participant_profile:,
-        )
-      end
-
-      it "can be voided" do
-        subject.call
-        expect(participant_declaration.reload).to be_voided
-      end
-    end
-
     context "when declaration is paid" do
       let(:participant_declaration) do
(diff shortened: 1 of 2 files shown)


spec/services/void_participant_declaration_spec.rb
```diff
@@ -62,9 +62,12 @@
         )
       end
 
-      it "can be voided" do
+      let(:line_item) { participant_declaration.statement_line_items.first }
```
cwrw, line 65: Maybe we group the specs correctly, as the below context for `when declaration is attached to a statement` will include the line items instead, where we can check all types there perhaps?

cwrw (approved):

ethax-ross (approved):
Looks good 👍

mooktakim (approved):

PR #5496, opened 2025-01-29
Remove unused StreamBigQueryParticipantDeclarationsJob

ethax-ross:
[Jira-3986](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3986)

### Context

We want to remove the `StreamBigQueryParticipantDeclarationsJob` job as it is no longer used.

- Remove the `StreamBigQueryParticipantDeclarationsJob` job as it is no longer used.

### Guidance to review

Also removes `big_query` queue and the scheduled cron job as well as the gem we were using.

Removes `GOOGLE_BQ_SERVICE_ACCOUNT_BASE64` from credentials.


config/sidekiq.yml
```diff
@@ -4,6 +4,5 @@
   - priority_mailers
   - mailers
   - default
-  - big_query
```
leandroalemao, line 7: Do we need to remove this queue via sidekiq web as well? Or just removing here will do the trick?

ethax-ross: I believe removing it here should be enough; we can check the UI afterwards though to make sure it all looks right 👍

leandroalemao (approved):
Hi @ethax-ross Looks good thx 👍

cwrw (approved):

PR #5505, opened 2025-01-31
[CPDLP-3937] Add mentor funding statement and output calculators

leandroalemao:
### Context

- Ticket: [CPDLP-3937](https://dfedigital.atlassian.net/browse/CPDLP-3937)

We would like implement mentor funding in the finance dashboards. 

- Add new output mentor calculators to calculate mentor funding;
- Add a new mentor statement calculator;


app/controllers/finance/ecf/statements_controller.rb
```diff
@@ -6,8 +6,12 @@ class StatementsController < BaseController
       def show
         @ecf_lead_provider = lead_provider_scope.find(params[:payment_breakdown_id])
         @statement = @ecf_lead_provider.statements.find(params[:id])
-        @calculator = StatementCalculator.new(statement: @statement)
+        @calculator = ECT::StatementCalculator.new(statement: @statement)
         set_important_message(title: t("finance.statements.payment_authorisations.banner.title"), content: t("finance.statements.payment_authorisations.banner.content", statement_marked_as_paid_at: @statement.marked_as_paid_at.strftime("%-I:%M%P on %-e %b %Y"))) if authorising_for_payment_banner_visible?(@statement)
+
+        if @statement.cohort.mentor_funding?
```
cwrw, line 12: I think we have a ticket for the FE finance calculators, can we please leave out the changes from this one?

app/views/finance/ecf/statements/show.html.erb
```diff
@@ -142,6 +142,143 @@
     </div>
   </div>
 </div>
+<% if @statement.cohort.mentor_funding? %>
```
cwrw, line 145: same here: we have a ticket for the FE finance calculators, can we please leave out the changes from this one?

app/controllers/finance/ecf/statements_controller.rb
leandroalemao: Ah yeah tks @cwrw I committed those ones.. forgot to remove it before pusing.. I had added just to be able to see calculators results in the page. tks

app/components/finance/statements/ecf_details_table.rb
```diff
@@ -9,7 +9,7 @@ class ECFDetailsTable < BaseComponent
 
       def initialize(statement:)
         @statement = statement
-        @calculator = Finance::ECF::StatementCalculator.new(statement: @statement)
+        @calculator = Finance::ECF::ECT::StatementCalculator.new(statement: @statement)
```
cwrw, line 12: I think the old calculators will support ECF in the old world and ECTs in the new world. Maybe we can keep them as is with ECF instead? and pass in the class in the initialiser instead, and default to both otherwise?

cwrw: this might help us avoiding `Finance::ECF::ECT`, and instead we'll have `Finance::ECF` for old world and ECTs in new world, but controlled by passing in the class to the calculator, let me know your thoughts

app/services/finance/ecf/output_calculator.rb
```diff
@@ -3,249 +3,16 @@
 module Finance
   module ECF
     class OutputCalculator
```
cwrw, line 5: I think from the spike the output calculator isn't shared, maybe we can keep that same assumption as there is nothing to share between mentor and ECTs? And we have different ECF and mentor output calculators?

mooktakim (changes requested):
I think we should keep the existing `Finance::ECF::StatementCalculator` as is, but create a new `Finance::ECF::ECT::StatementCalculator` which maybe inherits.
We can make the decision on which one to use in the controller. 
That way we keep the existing code unaffected, and the new way can be separate.

app/services/finance/ecf/mentor/output_calculator.rb
```diff
@@ -0,0 +1,75 @@
...
+          hash[:"#{declaration_type.underscore}_additions"] = current_output_count
+          hash[:"#{declaration_type.underscore}_subtractions"] = refundable_output_count
+          hash
+        end
+
+        def current_billable_count_for_declaration_type(declaration_type)
+          statement
+            .billable_statement_line_items
+            .joins(:participant_declaration)
+            .where(participant_declarations: { declaration_type:, type: "ParticipantDeclaration::Mentor" })
```
ethax-ross, line 61: Worth extracting this to a scope and merging it into the query?

app/services/finance/ecf/mentor/output_calculator.rb
```diff
@@ -0,0 +1,75 @@
...
+                         completed_event_percentage
+                       end
+
+          percentage * statement.mentor_contract.payment_per_participant
+        end
+
+      private
+
+        def started_event_percentage
+          0.5
```
ethax-ross, line 33: Would these be better as constants? I've also not seen the `event` terminology before, would something like `STARTED_DECLARATION_FEE_PERCENTAGE = 0.5` make more sense? or maybe wrap it up with declaration types to kill another method, such as:

```
DECLARATION_TYPE_FEE_PROPORTIONS = {
  started: 0.5,
  completed: 0.5.
}
```

app/services/finance/ecf/mentor/statement_calculator.rb
```diff
@@ -0,0 +1,113 @@
+# frozen_string_literal: true
+
+module Finance
+  module ECF
+    module Mentor
+      class StatementCalculator
+        class << self
+          def event_types
+            %i[
```
ethax-ross, line 9: Should we couple/delegate to the output calculator here as it will always need to be the same types?

app/services/finance/ecf/mentor/statement_calculator.rb
```diff
@@ -0,0 +1,113 @@
...
+          end
+        end
+
+        attr_reader :statement
+
+        def initialize(statement:)
+          @statement = statement
+        end
+
+        def vat
```
ethax-ross, line 29: `total_vat` so its not confusable with `vat_rate` or total including vat?

app/services/finance/ecf/mentor/statement_calculator.rb
```diff
@@ -0,0 +1,113 @@
...
+          total * vat_rate
+        end
+
+        def voided_declarations
+          statement.participant_declarations.voided.where(type: "ParticipantDeclaration::Mentor")
+        end
+
+        event_types.each do |event_type|
```
ethax-ross, line 37: I think `declaration_types` makes more sense here as well to mirror the field

app/services/finance/ecf/mentor/statement_calculator.rb
```diff
@@ -0,0 +1,113 @@
...
+          define_method "#{event_type}_fee_for_declaration" do
+            fee_for_declaration(type: event_type)
+          end
+
+          define_method "additions_for_#{event_type}" do
```
ethax-ross, line 42: Would it be a big change to get rid of these dynamic method names? open to opinions, but I find they make searching/navigating the code much harder

app/services/finance/ecf/statement_calculator.rb
```diff
@@ -5,46 +5,52 @@
...
-          a: 0,
-          b: 1,
-          c: 2,
-          d: 3,
-        }
-      end
-
-      attr_reader :statement
+      COHORT_WITH_NO_MENTOR_FUNDING_DECLARATION_TYPES = ["ParticipantDeclaration::ECT", "ParticipantDeclaration::Mentor"].freeze
+      COHORT_WITH_MENTOR_FUNDING_DECLARATION_TYPES = ["ParticipantDeclaration::ECT"].freeze
```
ethax-ross, line 9: Following these is a bit confusing 🤔 can the `StatementCalculator` infer the declaration types by checking the `mentor_funding?` flag on the `cohort` more directly?

cwrw: I was thinking we inject the type when we'd like from where we instantiate those services instead where we'd be using the flag rather than in those services

app/services/finance/ecf/output_calculator.rb
```diff
@@ -3,10 +3,11 @@
...
-      attr_reader :statement
+      attr_reader :statement, :participant_declaration_types
 
-      def initialize(statement:)
+      def initialize(statement:, participant_declaration_types: StatementCalculator::COHORT_WITH_NO_MENTOR_FUNDING_DECLARATION_TYPES)
```
cwrw, line 8: can we add specs for this please?

cwrw (commented):
looking good a few minor comments

app/models/participant_declaration.rb
```diff
@@ -64,6 +64,8 @@ class ParticipantDeclaration < ApplicationRecord
 
   scope :ect, -> { where(participant_profile_id: ParticipantProfile::ECT.select(:id)) }
   scope :mentor, -> { where(participant_profile_id: ParticipantProfile::Mentor.select(:id)) }
+  scope :ects, -> { where(type: "ParticipantDeclaration::ECT") }
+  scope :mentors, -> { where(type: "ParticipantDeclaration::Mentor") }
```
mooktakim, line 68: I think we can replace the first two now that we have type column with ECT/Mentor

spec/services/finance/ecf/output_calculator_spec.rb
```diff
@@ -1,1258 +1,5 @@
...
-            extended_3_subtractions: 0,
-          },
-        ]
-
-        expect(first_statement_calc.banding_breakdown.map { |e| e.slice(*extended_keys) }).to eql(expected)
-      end
-    end
-  end
+RSpec.describe Finance::ECF::OutputCalculator do
+  it_behaves_like "a Finance ECF output calculator", :mentor_participant_declaration
```
cwrw, line 4: is it possible to pass in both ect/mentor here instead?

app/services/finance/ecf/mentor/statement_calculator.rb
```diff
@@ -35,18 +35,20 @@ def voided_declarations
...
         event_types.each do |event_type|
-          define_method "#{event_type}_fee_for_declaration" do
+          define_method "#{event_type}_fee_per_declaration" do
             fee_for_declaration(type: event_type)
           end
 
           define_method "additions_for_#{event_type}" do
-            hash = {}.merge(*output_calculator.output_breakdown)
-            hash[:"#{event_type}_additions"] * output_calculator.fee_for_declaration(type: event_type)
+            output_calculator.output_breakdown.sum do |hash|
```
cwrw, line 43: do you mind explaining why we needed those changes as it's not explained in the commit message please?

leandroalemao: Just for easy understanding... I've improved the commit message.

cwrw (approved):

mooktakim (approved):

app/services/finance/ecf/mentor/statement_calculator.rb
leandroalemao: This is just the calculated `vat` figure..

app/services/finance/ecf/mentor/statement_calculator.rb
leandroalemao: yeah I tried that but a class method needs to be there in order to use in the `define_method` loop..

ethax-ross: You could keep the methods and just avoid redefining the values, so something like:

```
def declaration_types
  OutputCalculator.supported_declaration_types
end
```

Then in the `StatementCalculator`:

```
def self.supported_declaration_types
  DECLARATION_TYPE_FEE_PROPORTIONS.stringify_keys.keys
end
```

I'm assuming that these will always be the same value and it would just make it clear that if you want to add a new declaration type to the statement calculator then the output calculator has to change.

Just an thought, I don't feel that strongly about it.

app/services/finance/ecf/mentor/statement_calculator.rb
ethax-ross: Sure, I just meant the method name is maybe a bit ambiguous; if I came across something like this I would need to look at the implementation of `vat` to know if this is the vat rate the total vat amount:

```
<%= @statement_calculator.vat %>
```

Maybe its just how it is in my head, but if I read VAT I assume/see 20% 🙂

ethax-ross (commented):
Looks good, can we cover the lines sonarcloud flags as untested then I'll give it a ✅

mooktakim (approved):

ethax-ross (approved):

PR #5521, opened 2025-02-07
[CPDLP-3958] Remove extra lead providers from seeds

leandroalemao:
### Context

- Ticket: [CPDLP-3958](https://dfedigital.atlassian.net/browse/CPDLP-3958)

We currently have a set of known providers in production, however in seeds we create many other providers unnecessarily and the data for contracts and statements is not set up properly.

Ensure all seed data and participants for both ECTs/Mentors is attached to only the known lead providers.

[CPDLP-3958]: https://dfedigital.atlassian.net/browse/CPDLP-3958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/lead_provider.rb
+++ b/app/models/lead_provider.rb
@@ -2,6 +2,16 @@
 
 # This is actually ECFLeadProvider in all but name. See https://github.com/DFE-Digital/early-careers-framework/issues/698
 class LeadProvider < ApplicationRecord
+  ALL_PROVIDERS = [
+    "Ambition Institute",
+    "Best Practice Network",
+    "Capita",
+    "Education Development Trust",
+    "National Institute of Teaching",
+    "Teach First",
+    "UCL Institute of Education",
+  ].freeze
+
   belongs_to :cpd_lead_provider, optional: true
 
   has_many :participant_declarations, through: :cpd_lead_provider, class_name: "ParticipantDeclaration::ECF"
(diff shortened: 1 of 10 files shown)


app/models/lead_provider.rb
```diff
@@ -2,6 +2,16 @@
 
 # This is actually ECFLeadProvider in all but name. See https://github.com/DFE-Digital/early-careers-framework/issues/698
 class LeadProvider < ApplicationRecord
+  ALL_PROVIDERS = [
```
cwrw, line 5: I wonder if this list can be in the seeds not in the main model? as we already define them there and this avoids confusion?

ethax-ross (commented):
Looks good past Abeer's commet 👍

spec/services/participants/check_and_set_completion_date_spec.rb
```diff
@@ -4,13 +4,17 @@
...
   let(:school) do
     NewSeeds::Scenarios::Schools::School
       .new
       .build
+      .with_partnership_in(cohort:, lead_provider:)
       .chosen_fip_and_partnered_in(cohort:)
       .school
   end
+  let(:lead_provider_1) { create(:lead_provider, name: "Capita") }
+  let(:school_1) { NewSeeds::Scenarios::Schools::School.new.build.with_partnership_in(cohort:, lead_provider: lead_provider_1).chosen_fip_and_partnered_in(cohort:) }
```
mooktakim, line 17: `school_1` doesn't seem to be used, not needed maybe?

leandroalemao: Great spot tks 👍 😄

ethax-ross (approved):

PR #5535, opened 2025-02-12
Update api docs for unfunded mentors

ethax-ross:
[Jira-4013](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4013)

### Context

The new ECF framework requires providers to only submit started and completed declarations for mentors and started, retained and completed for ECT’s. This is different to how our current spec presents information. We’ll need our spec to dual run business logic from pre-2025 and 2025 onwards.

- Add release note for unfunded mentors

Add a release note to details the changes for recording declarations against unfunded mentors.

- Allow anyOf on data attributes in API docs

Currently, if you set a `data` attribute up with an `anyOf` value set it will output the "It conforms to" text as well as the "This conforms to" text. This is due to `data` being a special attribute whereby we usually link to another, single schema.

To get around this and allow `anyOf` for `data` attributes we only want to output the "It conforms to" links if there is _not_ an `anyOf` value set present.

- Update POST /api/participant-declarations API docs for mentor funding

The API docs for creating a declaration need to change to inform lead providers that they will only be able to use certain payloads depending on the cohort that is being referenced and the type of participant in question.

For pre-2025 cohorts the request schemas remain the same (started, retained, extended and completed). For 2025 and later when a declaration is being made for a mentor we will only accept started and completed.

### Guidance to review

This ticket changed a bit on picking it up; adding the copy from the ticket to the API docs directly isn't how swagger wants to work, so instead we need to abstract each scenario in its own schema.

[Link to reworked ParticipantDeclarationRequest schema](https://cpd-ecf-review-5535-web.test.teacherservices.cloud/api-reference/reference-v3.html#schema-participantdeclarationrequest)
[Pre-2025 schema](https://cpd-ecf-review-5535-web.test.teacherservices.cloud/api-reference/reference-v3.html#schema-participantdeclarationpre2025datarequest)
[Post-2024 ECT schema](https://cpd-ecf-review-5535-web.test.teacherservices.cloud/api-reference/reference-v3.html#schema-participantdeclarationpost2024ectdatarequest)
[Post 2024 mentor schema](https://cpd-ecf-review-5535-web.test.teacherservices.cloud/api-reference/reference-v3.html#schema-participantdeclarationpost2024mentordatarequest)

### DIFF
--- a/swagger/v1/api_spec.json
+++ b/swagger/v1/api_spec.json
@@ -2784,8 +2784,8 @@
           }
         }
       },
-      "ParticipantDeclarationDataRequest": {
-        "description": "A participant declaration data request",
+      "ParticipantDeclarationPost2024ECTDataRequest": {
+        "description": " A participant declaration data request for ECT participants from cohort 2025 onwards",
         "type": "object",
         "properties": {
           "type": {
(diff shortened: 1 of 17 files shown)


swagger/v1/api_spec.json
```diff
@@ -2784,8 +2784,58 @@
...
+          "type": {
+            "type": "string",
+            "enum": [
+              "participant-declaration"
+            ]
+          },
+          "attributes": {
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/ECFParticipantDeclarationStartedAttributesRequest"
```
leandroalemao, line 2828: For mentors, do we need to differentiate the parameters as well? ie: `course_identifier` can be either `ecf-induction` or `ecf-mentor` or it should be `ecf-mentor` only?

ethax-ross: Good question; this would lead to doubling the number of attribute request schemas, so I'd say not if we can help it, but I'll defer to @cwrw 🙂

cwrw: I think let's leave it as is, although good point to avoid duplication

swagger/v2/component_schemas/ParticipantDeclarationRequest.yml
```diff
@@ -4,4 +4,7 @@ required:
   - data
 properties:
   data:
-    $ref: "#/components/schemas/ParticipantDeclarationDataRequest"
```
cwrw, line 7: do we need to do something similar with renaming the `ParticipantDeclarationDataRequest` component in v2/v3?

cwrw (commented):
looks great, just the one question. After all is done do you mind running this by Mark, to get his thoughts? thanks

swagger/v2/component_schemas/ParticipantDeclarationRequest.yml
ethax-ross: Ah yes, I should have removed that file for v2 and v3 as its redundant now. I'll update 👍

leandroalemao (approved):
Looks good tks @ethax-ross

cwrw (approved):
looks great thanks Ross

mooktakim (approved):

PR #5541, opened 2025-02-13
Address brakeman warnings

ethax-ross:
[Jira-3987](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-3987)

### Context

We have a number of `brakeman` security warnings in the app that we want to review and address where appropriate.

- Fix dynamic render path brakeman warning

We have a warning due to dynamically rendering an unsanitised page parameter.

Extract template resolving to `TemplateResolver` service and only allow rendering of valid templates in the `app/views/page` directory.

- Remove unused permitted filter attributes

We only appear to filter on `query` and `status` in the `AppropriateBodies::ParticipantsFilter` service.

The `role` attribute is flagged by Brakeman (albeit as a false positive), but as its not used we can just remove it to silence the warning.

- Sanitize confirmation page rendering

We only want to allow rendering certain partials to avoid dynamic render paths.

Switch `expect` to `allow` for `exceptions_app` helper so that we can use it in feature specs (the method is called multiple times in the feature specs).

- Add brakeman.ignore file

A few of the `brakeman` warnings are false positives that we can ignore.

Add `config/brakeman.ignore` file to allow ignoring warnings. We're using the default file path here for the ignore file so we shouldn't need to update any of the commands that run `brakeman`.

### Guidance to review

I started this as a spike/investigation, but figured it wasn't going to be much more work to just address them with solutions and we can then use this PR for discussion.

### DIFF
--- a/app/controllers/appropriate_bodies/participants_controller.rb
+++ b/app/controllers/appropriate_bodies/participants_controller.rb
@@ -45,7 +45,7 @@ module AppropriateBodies
     end
 
     def filter_params
-      params.permit(:query, :role, :academic_year, :status)
+      params.permit(:query, :status)
     end
   end
 end
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -2,10 +2,29 @@
 
 class PagesController < ApplicationController
   def show
-    render template: "pages/#{params[:page].tr('-', '_')}"
+    template = template_resolver.resolve(params[:page])
+
+    raise_not_found! unless template
+
+    render template:
   end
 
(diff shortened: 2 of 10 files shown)


mooktakim (approved):

leandroalemao (approved):
Tested locally.. all good tks @ethax-ross

cwrw (commented):
thanks Ross for this, all solutions seems sensible however looking to test further on the review app to ensure there are no side effects. Will ping tomorrow to confirm

cwrw (approved):

PR #5543, opened 2025-02-13
[CPDLP-3915] Add new participant declarations evidence_held validator

leandroalemao:
### Context

- Ticket: [CPDLP-3915](https://dfedigital.atlassian.net/browse/CPDLP-3915)

The 2025 policy requires a more detailed evidence type to reduce use of “other” by lead providers.

- Create a new `evidence_held` validator;
- Add new set of evidence types depending upon the declaration and participant profile types;


app/validators/evidence_held_validator.rb
```diff
@@ -0,0 +1,88 @@
...
+      # extended-3
+      9 => [1, 2, 3, 7],
+    },
+      "ParticipantProfile::Mentor" => {
+        # started
+        1 => [1, 2, 3, 7],
+        # completed
+        5 => [4, 5],
+      },
+  }.freeze
```
mooktakim, line 53: Using numbers for the map would be difficult to maintain and easily understand. 
I'm thinking its better to break it up and maybe use case statements:
```ruby
def mentor_evidences(declaration_type)
  case declaration_type
  when "started"
    %w[
      training-event-attended
      self-study-material-completed
      materials-engaged-with-offline
      other
    ]
  when "completed"
    %w[
      75-percent-engagement-met
      75-percent-engagement-met-reduced-induction
    ]
  end
end

def ect_evidences(declaration_type)
  case declaration_type
  when "started", "retained-1", "retained-3", "retained-4", "extended-1", "extended-2", "extended-3"
    %w[
      training-event-attended
      self-study-material-completed
      materials-engaged-with-offline
      other
    ]
  when "retained-2"
    %w[
      75-percent-engagement-met
      75-percent-engagement-met-reduced-induction
    ]
  when "completed"
    %w[
      75-percent-engagement-met
      75-percent-engagement-met-reduced-induction
      one-term-induction
    ]
  end
end
```

mooktakim: You'll then be able to check with `record.evidence_held.in?(ect_evidences(record.declaration_type))`

app/validators/evidence_held_validator.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def self.evidence_held_map
+    EVIDENCE_HELD_MAP.map { |sa| { sa.first => sa.last.map { |a| { DECLARATION_TYPES.key(a.first) => a.last.map { |b| VALID_EVIDENCE_HELD.key(b) } } }.reduce(&:merge) } }.reduce(&:merge)
```
mooktakim, line 56: and you won't need this

app/validators/evidence_held_validator.rb
```diff
@@ -0,0 +1,88 @@
...
+
+private
+
+  def validate_evidence_held?(record)
+    return if record.declaration_type == "started" && record.evidence_held.blank?
+
+    record.participant_profile && record.participant_profile.is_a?(ParticipantProfile::ECF)
+  end
+
+  def has_evidence_held?(record)
```
cwrw, line 75: maybe we can include presence in the method name for clarity?

spec/validators/evidence_held_validator_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+        context "when `evidence_held` is not present" do
+          let(:evidence_held) { nil }
+
+          it "has a meaningful error", :aggregate_failures do
+            expect(subject).to be_invalid
+            expect(subject.errors.messages_for(:evidence_held)).to include("Enter a '#/evidence_held' value for this participant.")
+          end
+        end
+
+        context "when `evidence_held` is present" do
```
cwrw, line 157: I wonder if we should fallback to the other validations in record declaration for declaration types for mentors being not supported, to avoid having an extra error over that one being returned 🤔

spec/validators/evidence_held_validator_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+          let(:evidence_held) { "other" }
+
+          it "is valid" do
+            expect(subject).to be_valid
+          end
+
+          context "when `evidence_held` is invalid" do
+            let(:evidence_held) { "anything" }
```
cwrw, line 135: maybe worth adding another test to an ect only value to test that that errors as well?

app/validators/evidence_held_validator.rb
```diff
@@ -0,0 +1,75 @@
...
+    return if record.evidence_held.present?
+
+    record.errors.add(:evidence_held, I18n.t(:missing_evidence_held))
+  end
+
+  def valid_evidence_held?(record)
+    return if record.errors.any?
+
+    method = record.participant_profile.ect? ? :ect_evidences : :mentor_evidences
+    return if record.evidence_held.in?(send(method, record.declaration_type) || [])
```
mooktakim, line 71: Better to not use `send`, instead:
```ruby
evidences = if record.participant_profile.ect?
  ect_evidences(record.declaration_type)
else
  mentor_evidences(record.declaration_type)
end

return if record.evidence_held.in?(evidences)
```

mooktakim (approved):

app/validators/evidence_held_validator.rb
```diff
@@ -0,0 +1,110 @@
...
+private
+
+  def cohort_with_detailed_evidence_types?(record)
+    return unless record.schedule && record.cohort
+
+    record.cohort.detailed_evidence_types?
+  end
+
+  def validate_evidence_held?(record)
+    return unless record.participant_profile && record.participant_profile.is_a?(ParticipantProfile::ECF)
```
ethax-ross, line 73: `return unless record.participant_profile&.ect?`

ethax-ross (approved):

app/services/record_declaration.rb
```diff
@@ -20,13 +20,7 @@ class RecordDeclaration
...
   validates :declaration_type, presence: { message: I18n.t(:missing_declaration_type) }
   validates :declaration_date, future_date: true, declaration_date: true, allow_blank: true
-  validates :evidence_held,
-            presence: { message: I18n.t(:missing_evidence_held), if: :validate_evidence_held? },
-            inclusion: {
-              in: ->(record_declaration) { record_declaration.participant_profile.class::VALID_EVIDENCE_HELD },
-              if: :validate_evidence_held?,
-              message: I18n.t(:invalid_evidence_type),
-            }
+  validates :evidence_held, evidence_held: true
```
cwrw, line 23: could we move this down as the last validation to line 30 to avoid it returning with other validations incase?

cwrw (commented):
looks great and tested, only some minor comments otherwise ✅

cwrw (approved):
nicely done, thanks Leandro 🙌

mooktakim (approved):

PR #5550, opened 2025-02-14
[CPDLP-3956] Middleware to catch SessionRestoreError and clear the session

mooktakim:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3956

* New middleware `SessionRestoreErrorRescueMiddleware`
* Added to `applications.rb`
* Rescue from `ActionDispatch::Session::SessionRestoreError` error, delete session cookie and reset

### DIFF
--- a/app/middlewares/session_restore_error_rescue_middleware.rb
+++ b/app/middlewares/session_restore_error_rescue_middleware.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class SessionRestoreErrorRescueMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    @app.call(env)
+  rescue ActionDispatch::Session::SessionRestoreError
+    req = ActionDispatch::Request.new(env)
+    req.cookies.delete(session_key)
+    req.env.delete("rack.session")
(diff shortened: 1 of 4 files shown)


app/middlewares/session_restore_error_rescue_middleware.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class SessionRestoreErrorRescueMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    @app.call(env)
+  rescue ActionDispatch::Session::SessionRestoreError
+    req = ActionDispatch::Request.new(env)
```
cwrw, line 11: discussed in person: maybe we can try deleting the session instead, if we can get the id

mooktakim: Updated code to only delete the session from session store and leave cookies unchanged

leandroalemao (approved):
Looks good 👏 tks

cwrw (approved):
looks great thanks Mook 👍

ethax-ross (approved):
Looks good 👍

PR #5564, opened 2025-02-21
Remove shared sandbox seed data

ethax-ross:
### Context

As NPQ separation has finished we no longer need the shared sandbox seed data.

- Remove shared sandbox seed data

Remove seed and associated YML file. Update `MentorEctGenerator` seed which was referencing it when it shouldn't have been/doesn't need to.


leandroalemao (approved):

leandroalemao (commented):
Should we remove the call from here as well?

https://github.com/DFE-Digital/early-careers-framework/blob/main/lib/tasks/sandbox_data.rake#L32

leandroalemao (changes requested):
https://github.com/DFE-Digital/early-careers-framework/blob/main/lib/tasks/sandbox_data.rake#L32

ethax-ross:
> https://github.com/DFE-Digital/early-careers-framework/blob/main/lib/tasks/sandbox_data.rake#L32

Whoops, yep, could have sworn I had done that - good spot, I'll push a fix

leandroalemao (approved):
All looks good Tks @ethax-ross

mooktakim (approved):

cwrw (approved):

PR #5565, opened 2025-02-21
Exclude 2025 and later participants from delivery partners

ethax-ross:
[Jira-4060](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4060)

### Context

We want to exclude participants in the 2025 cohort from the delivery partners participants view. We also want to hide the 2025 and later cohorts from the academic year dropdown.

This is intended to encourage lead providers to carry out their contractual duties and ensure there is no ownness on the DfE to share details of participants in 2025 and later cohorts.

- Exclude 2025 and later participants from delivery partners

### Guidance to review

Participant created in 2025 cohort for `Koelpin-Homenick` DP:

```
school_cohort = FactoryBot.create(:school_cohort, cohort: Cohort.find_by!(start_year: 2025))
delivery_partner = DeliveryPartner.find("8d2875ca-4727-48b7-8adf-9a47fa9debdc")
lead_provider = LeadProvider.find_by(name: "Ambition Institute")
partnership = FactoryBot.create(:partnership, lead_provider: FactoryBot.create(:lead_provider, name: "LP3"), delivery_partner:)
induction_programme = FactoryBot.create(:induction_programme, partnership:)
participant_profile = FactoryBot.create(:ect_participant_profile, training_status: "deferred")
FactoryBot.create(:induction_record, participant_profile:, induction_programme:, school_cohort:)
```

### DIFF
--- a/app/services/delivery_partners/participants_filter.rb
+++ b/app/services/delivery_partners/participants_filter.rb
@@ -4,6 +4,8 @@ module DeliveryPartners
   class ParticipantsFilter
     attr_reader :collection, :params, :training_record_states
 
+    LATEST_COHORT_TO_RETURN = 2024
+
     def initialize(collection:, params:, training_record_states:)
       @collection = collection
       @params = params
@@ -11,7 +13,7 @@ module DeliveryPartners
     end
 
     def scope
-      scoped = collection
+      scoped = filter_academic_year(collection, params[:academic_year])
 
       if params[:query].present?
         scoped = filter_query(scoped, params[:query])
@@ -19,10 +21,6 @@ module DeliveryPartners
 
       if params[:role].present?
         scoped = filter_role(scoped, params[:role])
-      end
-
-      if params[:academic_year].present?
-        scoped = filter_academic_year(scoped, params[:academic_year])
       end
 
       if params[:status].present?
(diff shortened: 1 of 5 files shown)


app/services/delivery_partners/participants_filter.rb
```diff
@@ -61,7 +63,13 @@ def filter_role(scoped, role)
       end
     end
 
+    def filter_out_cohorts(scoped)
+      scoped.joins(:cohort).where("start_year <= ?", LATEST_COHORT_TO_RETURN)
```
leandroalemao, line 67: Should we do this in `app/services/delivery_partners/induction_records_query.rb` instead?

We're seeing timeouts on the AB participants filter which has a similar logic to DP filter.. please see https://dfedigital.atlassian.net/browse/CPDLP-4045. 

Maybe DP participants has lesser records in comparison with AB, so no timeout are raised so we should be fine doing it here. Needs some assessment.

app/services/delivery_partners/participants_filter.rb
```diff
@@ -61,7 +63,13 @@ def filter_role(scoped, role)
...
+    end
+
     def filter_academic_year(scoped, academic_year)
```
cwrw, line 70: I wonder if we could always filter by academic year even if filter not passed, default being filtering by all years or the one passed in, rather than adding a new `filter_out_cohorts` method?

app/services/delivery_partners/participants_filter.rb
ethax-ross: I think it makes sense to keep the filtering all in one place; if I was looking for it I would look in `ParticipantsFilter`. As we're not introducing new functionality here (we already support filtering on cohort) it shouldn't cause any new performance issues - happy to discuss if you feel differently though?

cwrw (approved):

mooktakim (approved):

leandroalemao (approved):

PR #5570, opened 2025-02-24
[CPDLP-3943] Create eligible schools and partnerships on the mentor/ect valid test data generators

leandroalemao:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-3943

We want to create participants linked to eligible schools as only eligible schools are visible in the admin dashboards and returned in the API responses.

- Create eligible schools when running the mentor/ect valid test data generators;
- Make sure we select the eligible school when creating the school cohort in the generator;

### Guidance to review

```
  names = [
    "Ambition Institute",
  ]
  cohorts = Cohort.between_years(2021, 2025)

  names.product(cohorts) do |name, cohort|
    ValidTestDataGenerators::ECFLeadProviderPopulater.call(name:, cohort:, total_schools: 1, participants_per_school: 0)
    ValidTestDataGenerators::MentorECTGenerator.call(name:, cohort:, number: 5)
  end
end
```

### DIFF
--- a/app/services/valid_test_data_generators/mentor_ect_generator.rb
+++ b/app/services/valid_test_data_generators/mentor_ect_generator.rb
@@ -51,13 +51,17 @@ module ValidTestDataGenerators
 
     def find_or_create_school_cohort
       school = partnership.school
-      school_cohort = SchoolCohort.find_or_create_by!(school:, cohort:, induction_programme_choice: "full_induction_programme")
-      InductionProgramme.find_or_create_by!(school_cohort:, partnership:, training_programme: "full_induction_programme")
+      school_cohort = SchoolCohort.find_or_create_by!(school:, cohort:) do |sc|
+        sc.induction_programme_choice = "full_induction_programme"
+      end
+      school.school_local_authorities.create!(local_authority: LocalAuthority.all.sample, start_year: cohort.start_year)
+      Induction::SetCohortInductionProgramme.call(school_cohort:,
+                                                  programme_choice: "full_induction_programme")
       school_cohort
     end
 
     def partnership
-      lead_provider.partnerships.find_by(cohort:)
+      lead_provider.partnerships.joins(:school).merge!(School.eligible_or_cip_only).find_by(cohort:)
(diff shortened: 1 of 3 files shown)


app/services/valid_test_data_generators/mentor_ect_generator.rb
```diff
@@ -51,13 +51,17 @@ def schedule
 
     def find_or_create_school_cohort
       school = partnership.school
-      school_cohort = SchoolCohort.find_or_create_by!(school:, cohort:, induction_programme_choice: "full_induction_programme")
-      InductionProgramme.find_or_create_by!(school_cohort:, partnership:, training_programme: "full_induction_programme")
+      school_cohort = SchoolCohort.find_or_create_by!(school:, cohort:) do |sc|
```
ethax-ross, line 54: What's the reason we are only matching on school/cohort but set programme choice to FIP on creation if it doesn't exist?

cwrw (commented):
looks good past Ross's questions, do you mind running this on the review app so we can test :)

app/services/valid_test_data_generators/mentor_ect_generator.rb
leandroalemao: Because I was having issues on that line because there's a unique constraint in school & cohort..

`SchoolCohort.find_or_create_by!(school:, cohort:, induction_programme_choice: "full_induction_programme")` might try to create a new record for the same school & cohort if `induction_programme_choice` is other than `full_induction_programme` in an existing record.

cwrw (approved):

app/services/valid_test_data_generators/mentor_ect_generator.rb
ethax-ross: Ah I see, that makes sense thanks 👍

ethax-ross (commented):
Looks good, the sandbox seeds take a long time to run from scratch now - not sure if that is because of this or past changes but its about an hour locally.

PR #5575, opened 2025-02-27
Support tagging in release notes

ethax-ross:
[Jira-4076](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4076)

### Context

We want the ability to tag release notes so that providers can easily distinguish the type of change being released.

- Support tagging in release notes

Add custom `MarkdownRenderer` that can interpret our tag syntax, for example: `[#tag-1 #tag-2]`.

Update Middleman config to use our custom markdown rendering.

Add CSS for slightly smaller govuk-style tags.

Add page title to body class so that we can easily add targetted CWSS.

Update latest release notes with tags.

### Guidance to review

Example of all tags (review app only has tags on the latest two release notes, as per the ticket):

Screenshot 2025-02-27 at 10 55 09

### DIFF
--- a/docs/source/api-reference/stylesheets/screen.css.scss
+++ b/docs/source/api-reference/stylesheets/screen.css.scss
@@ -14,6 +14,17 @@ $govuk-assets-path: "/api-reference/assets/govuk/assets/";
   padding: 3px 0;
 }
 
+body.release-notes .technical-documentation {
+  .tag-group {
+    margin: -22px 0 22px 0;
+  
+    .govuk-tag {
+      font-size: 0.8em;
+      margin: 0 10px 8px 0;
+    }
+  }
+}
+
 @include govuk-media-query(desktop) {
   .flexbox {
     .app-pane__toc {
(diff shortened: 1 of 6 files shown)


leandroalemao (approved):
Hi @ethax-ross Excelent!! Tks

docs/source/release-notes.html.md
```diff
@@ -39,6 +41,8 @@ These changes align with the updated frameworks and payment guidelines shared wi
 
 ## 8 January 2025
 
+[#new-feature]
```
cwrw, line 44: I think this is a bug fix, regardless of what the ticket says :D

docs/source/api-reference/stylesheets/screen.css.scss
```diff
@@ -14,6 +14,17 @@ $govuk-assets-path: "/api-reference/assets/govuk/assets/";
   padding: 3px 0;
 }
 
+body.release-notes .technical-documentation {
+  h2 {
+    margin-bottom: 5px;
```
cwrw, line 19: happy to chat about this, but do we not want to keep the margin on h2 as is IF there are no tags? as now the text is quite close to the header?

ethax-ross: Ah yeah good point, I'll change it so the margin only appears for the headings with tags 👍

cwrw (approved):

PR #5579, opened 2025-02-28
fix: Adding permissions for restoring snapshot DB

shaheislamdfe:
### Context

Fix to add permissions for restoring Snapshot DB

Adding write and contents permission, this will fix the workflow

### DIFF
--- a/.github/actions/backup-and-restore-snapshot-database/action.yml
+++ b/.github/actions/backup-and-restore-snapshot-database/action.yml
@@ -28,6 +28,13 @@ runs:
   using: composite
 
   steps:
+    - name: Set Environment variables
+      id: set_env_var
+      shell: bash
+      run: |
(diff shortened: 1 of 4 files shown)


.github/actions/backup-and-restore-snapshot-database/action.yml
```diff
@@ -28,6 +28,13 @@ runs:
   using: composite
 
   steps:
+    - name: Set Environment variables
+      id: set_env_var
+      shell: bash
+      run: |
+        tf_vars_file=terraform/application/workspace_variables/${{ inputs.environment }}.tfvars.json
+        echo "namespace=$(jq -r '.namespace' ${tf_vars_file})" >> $GITHUB_ENV
```
saliceti, line 36: We normaly follow a convention to use uppercase for environment variable names. Can you rename it in the workflows?

cwrw (approved):

PR #5580, opened 2025-03-03
Add `voided_at` and `voided_by_user` to `ParticipantDeclaration`

ethax-ross:
[Jira-4084](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4084)

### Context

We want to maintain a record of who voided a declaration when it happens through the user interface.

- Add `voided_at` and `voided_by_user` to `ParticipantDeclaration` 

These will be `nil` when a lead provider has voided the declaration via the API.

### Guidance to review

The `voided_at` column appears to have been part of the model years ago, which is why it was later ignored - see #2320 - reversing that change here as we re-introduce it.

### DIFF
--- a/app/models/participant_declaration.rb
+++ b/app/models/participant_declaration.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ParticipantDeclaration < ApplicationRecord
-  self.ignored_columns = %w[statement_type statement_id voided_at temp_type]
+  self.ignored_columns = %w[statement_type statement_id temp_type]
 
   ARCHIVABLE_STATES = %w[ineligible voided submitted].freeze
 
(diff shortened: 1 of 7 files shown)


spec/models/participant_declaration_spec.rb
```diff
@@ -22,6 +23,18 @@
     it { is_expected.to validate_presence_of(:declaration_date) }
     it { is_expected.to validate_presence_of(:declaration_type) }
     it { is_expected.to validate_presence_of(:cohort) }
+
+    context "when voided_at is present" do
```
cwrw, line 27: is it worth adding a test that if both are `nil` that the declaration is valid?

ethax-ross: I'll add explicit tests to ensure the fields are not validated when nil 👍

app/models/participant_declaration.rb
```diff
@@ -45,6 +46,8 @@ class ParticipantDeclaration < ApplicationRecord
   delegate :fundable?, to: :participant_profile, allow_nil: true
 
   validates :course_identifier, :user, :cpd_lead_provider, :declaration_date, :declaration_type, :cohort, presence: true
+  validates :voided_by_user, presence: true, if: :voided_at
+  validates :voided_at, presence: true, if: :voided_by_user
```
leandroalemao, line 50: Doesn't it create a dead lock condition? Not sure.. as it validates the other when the first present and vice-versa?

ethax-ross: Not sure I follow, I can't see how it could deadlock? It'll process the validation rules in order

mooktakim (approved):

PR #5582, opened 2025-03-03
Update `VoidParticipantDeclaration` to support `voided_by_user`

ethax-ross:
> ⚠️ depends on #5580

### Context

When we void a declaration manually via the admin interface we want to track the user that performed the void and the time that it was voided. 

- Update `VoidParticipantDeclaration` to support `voided_by_user`

We have `voided_at` and `voided_by_user` as attributes on the `ParticipantDeclaration` model that we can use.

Add a `voided_by_user` required parameter to the `VoidParticipantDeclaration` service initializer and update current usage of the service to pass `nil` (to retain current behaviour).

When we come to expose this in the admin interface we can pass the currently signed in user. If a user is provided we will populated `voided_by_user` and set `voided_at` to the current time.

### Guidance to review

In the ticket and during our discussion we agreed to wrap this in a form object model, but on implementing it I don't think this is worthwhile. When we come to expose this in the front end we're not going to be sending any data from the browser and having an extra form model will just create confusion (if we forgot to use this model in the console we would end up losing the voided by user details).

By keeping it in the `VoidParticipantDeclaration` service it means we only have a single service/point of use and we can call this service from the action that processes the web link, passing in the `current_user`.

I thought about extracting the `track_voiding_user` and its tests to a concern/shared context, but decided it was probably more hassle than its worth.

### DIFF
--- a/app/services/finance/clawback_declaration.rb
+++ b/app/services/finance/clawback_declaration.rb
@@ -8,14 +8,16 @@ module Finance
     validate :validate_refundable_state
     validate :output_fee_statement_available
 
-    def initialize(participant_declaration)
+    def initialize(participant_declaration, voided_by_user:)
       self.participant_declaration = participant_declaration
+      self.voided_by_user = voided_by_user
     end
 
     def call
       return if invalid?
 
       ApplicationRecord.transaction do
+        track_voiding_user
         DeclarationState.awaiting_clawback!(participant_declaration)
         DeclarationStatementAttacher.new(participant_declaration).call
         ParticipantDeclarations::HandleMentorCompletion.call(participant_declaration:)
@@ -24,7 +26,13 @@ module Finance
 
   private
 
-    attr_accessor :participant_declaration
+    attr_accessor :participant_declaration, :voided_by_user
+
+    def track_voiding_user
+      return unless voided_by_user
+
+      participant_declaration.update!(voided_by_user:, voided_at: Time.zone.now)
+    end
 
     def output_fee_statement_available
       cohort = participant_declaration.cohort
--- a/app/services/void_participant_declaration.rb
+++ b/app/services/void_participant_declaration.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 class VoidParticipantDeclaration
-  def initialize(participant_declaration)
+  def initialize(participant_declaration, voided_by_user:)
     self.participant_declaration = participant_declaration
+    self.voided_by_user = voided_by_user
   end
 
   def call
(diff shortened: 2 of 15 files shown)


app/services/finance/clawback_declaration.rb
```diff
@@ -8,14 +8,16 @@ class ClawbackDeclaration
     validate :validate_refundable_state
     validate :output_fee_statement_available
 
-    def initialize(participant_declaration)
+    def initialize(participant_declaration, voided_by_user:)
```
cwrw, line 11: can voiding by `nil` be the default so we don't need to change the interface everywhere?

app/services/void_participant_declaration.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 class VoidParticipantDeclaration
-  def initialize(participant_declaration)
+  def initialize(participant_declaration, voided_by_user:)
```
cwrw, line 4: same here regarding making it default

app/services/void_participant_declaration.rb
```diff
@@ -31,6 +32,7 @@ def make_voided
     raise Api::Errors::InvalidTransitionError, I18n.t(:declaration_not_voidable) unless participant_declaration.voidable?
 
     ApplicationRecord.transaction do
+      track_voiding_user
```
cwrw, line 35: I'm in two minds around where to have this, maybe we only have it in the void declaration service and instead of implementing it twice in each service we check if the clawback or voided was successful and track the voided user here once for both?

app/services/finance/clawback_declaration.rb
ethax-ross: So I did this intentionally; if someone goes to call this service by the rails console I wanted it to throw out an error if they forgot to specify a `voided_by_user`, rather than defaulting to the API-type behaviour. It's less likely to to be misused this way IMO

app/services/void_participant_declaration.rb
ethax-ross: I opted to put it in both services to make it more robust, in case `Finance::ClawbackDeclaration` was called independently (we do this in `Importers::Clawbacks`, for example). It's less likely this is going to be called manually from the rails console I guess, so happy to limit it to just the `VoidParticipantDeclaration` service if you feel that would be better?

app/services/finance/clawback_declaration.rb
cwrw: fair enough

app/services/void_participant_declaration.rb
cwrw: let's keep it

cwrw (approved):

mooktakim (approved):

leandroalemao (approved):

PR #5587, opened 2025-03-04
Remove sentry url from CSP config

leandroalemao:
### Context

- Ticket: N/A

PR to rollback changes made in the PR https://github.com/DFE-Digital/early-careers-framework/pull/5586

- Remove Sentry URL from CSP `connect_src` config;


cwrw (approved):

ethax-ross (approved):

leandroalemao:
Think we don't need this any more as it was created just in case we had to rollback these changes.

PR #5596, opened 2025-03-05
Enable voiding of declarations via the finance dashboard

ethax-ross:
[Jira-4088](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4088)

### Context

At present declarations can only be voided directly by a Lead Provider or through a developer after a support request being made. This ends up causing lengthy lead times before a resolution can be reached, either delaying payment or creating need for further manual intervention.

To alleviate the issues we want to allow finance users to be able to void applicable declarations through the finance dashboard.

- Extract voidable states into constant

Extracting and exposing as a constant as we need to be able to reference these states externally to determine if a declaration is voidable.

- Add ability to void declarations in finance dashboard

Add `Void` link to the participant declarations if they are in a voidable state (if paid they must also have an output fee statement available).

Add summary page to detail declaration prior to voiding.

On voiding, redirect to the participant page with a success message.

If there is an error when voiding, redirect to the participant page with a failure message.

- Update void error message styling

Switch to a govuk summary error type display after talking to Mark about the styling.

### Guidance to review

| Void link    | Void page | Success | Failure |
| -------- | ------- | -------- | ------- |
| Screenshot 2025-03-05 at 13 27 29  | Screenshot 2025-03-05 at 13 28 12   |  Screenshot 2025-03-05 at 13 28 36  | Screenshot 2025-03-07 at 11 50 21  |

### DIFF
--- a/app/assets/stylesheets/application.sass.scss
+++ b/app/assets/stylesheets/application.sass.scss
@@ -63,3 +63,8 @@ $govuk-new-link-styles: true;
 .govuk-summary-list__key--align-top {
   vertical-align: top;
 }
+
+.govuk-error-summary__list li {
+  color: #d4351c;
+  font-weight: bold;
+}
--- a/app/controllers/finance/ecf/participant_declarations/void_controller.rb
+++ b/app/controllers/finance/ecf/participant_declarations/void_controller.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Finance
+  module ECF
+    module ParticipantDeclarations
+      class VoidController < BaseController
+        before_action :set_participant_profile_and_declaration
+
+        def new; end
+
+        def create
+          VoidParticipantDeclaration.new(@declaration, voided_by_user: current_user).call
+          set_success_message(**I18n.t("finance.void_declaration.success"))
+        rescue Api::Errors::InvalidTransitionError
+          set_important_message(**I18n.t("finance.void_declaration.failure"))
+        ensure
+          redirect_to finance_participant_path(@participant_profile.user)
+        end
+
+      private
+
+        def set_participant_profile_and_declaration
+          @participant_profile = ParticipantProfile.find(params[:participant_profile_id])
+          @declaration = @participant_profile.participant_declarations.includes(:cpd_lead_provider).find(params[:participant_declaration_id])
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/finance/ecf/participant_declarations/void_controller.rb
+++ b/app/controllers/finance/ecf/participant_declarations/void_controller.rb
@@ -12,7 +12,7 @@ module Finance
           VoidParticipantDeclaration.new(@declaration, voided_by_user: current_user).call
           set_success_message(**I18n.t("finance.void_declaration.success"))
         rescue Api::Errors::InvalidTransitionError
-          set_important_message(**I18n.t("finance.void_declaration.failure"))
+          set_alert_message(**I18n.t("finance.void_declaration.failure"))
         ensure
           redirect_to finance_participant_path(@participant_profile.user)
         end
--- a/app/views/finance/ecf/participant_declarations/void/new.html.erb
+++ b/app/views/finance/ecf/participant_declarations/void/new.html.erb
@@ -1,0 +1,50 @@
+<% content_for :before_content, govuk_back_link(text: "Back", href: finance_participant_path(@participant_profile.user)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">Void declaration</h1>
+
+    <%= 
+      govuk_summary_list do |summary_list|
+        summary_list.with_row do |row|
+          row.with_key { "Participant ID" }
+          row.with_value { @participant_profile.user_id }
+        end
+        
+        summary_list.with_row do |row|
+          row.with_key { "Declaration ID" }
+          row.with_value { @declaration.id }
+        end
+
+        summary_list.with_row do |row|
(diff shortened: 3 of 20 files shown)


app/controllers/finance/ecf/participant_declarations/void_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      class VoidController < BaseController
+        before_action :set_participant_profile_and_declaration
+
+        def new; end
+
+        def create
+          VoidParticipantDeclaration.new(@declaration, voided_by_user: current_user).call
+          set_success_message(**I18n.t("finance.void_declaration.success"))
+        rescue Api::Errors::InvalidTransitionError
+          set_important_message(**I18n.t("finance.void_declaration.failure"))
```
cwrw, line 15: I think this might need to be red, worth checking with Mark as well?

app/controllers/finance/ecf/participant_declarations/void_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+        ensure
+          redirect_to finance_participant_path(@participant_profile.user)
+        end
+
+      private
+
+        def set_participant_profile_and_declaration
+          @participant_profile = ParticipantProfile.find(params[:participant_profile_id])
```
cwrw, line 23: can't we infer the profile from the declaration? just wondering why we'll need to have both ids sent when we only need to the declaration?

spec/helpers/finance_helper_spec.rb
```diff
@@ -75,6 +75,68 @@
...
+
+    (ParticipantDeclaration.states.keys - ParticipantDeclaration::VOIDABLE_STATES).each do |state|
+      context "when the declaration is #{state}" do
+        let(:declaration) { create(:ect_participant_declaration, state) }
+
+        it { is_expected.not_to be_voidable_declaration(declaration) }
+      end
+    end
+
+    context "when the declaration is paid and there is a next output fee statement" do
```
cwrw, line 97: can we add a test when there is no output fee statement but it's paid?

app/views/finance/ecf/participant_declarations/void/new.html.erb
```diff
@@ -0,0 +1,45 @@
...
+        end
+
+        summary_list.with_row do |row|
+          row.with_key { "Declaration date" }
+          row.with_value { @declaration.declaration_date.to_fs(:govuk) }
+        end
+
+        summary_list.with_row do |row|
+          row.with_key { "Lead provider" }
+          row.with_value { @declaration.cpd_lead_provider.name }
```
cwrw, line 31: maybe we can include the course identifier as well?

cwrw (commented):
very cleanly and nicely done, just some minor comments

app/controllers/finance/ecf/participant_declarations/void_controller.rb
ethax-ross: Yeah we could, but I opted to stay consistent with other similar routes here that include the participant profile ID. Happy to change it but we may want to do it as part of a wider refactoring to keep them consistent, for example:

```
/finance/participant_profiles/:participant_profile_id/ecf/induction_records/:induction_record_id/change_training_status/new
```

spec/helpers/finance_helper_spec.rb
ethax-ross: That should be covered by the test above: `(ParticipantDeclaration.states.keys - ParticipantDeclaration::VOIDABLE_STATES)`

app/views/finance/ecf/participant_declarations/void/new.html.erb
ethax-ross: I'll add that in 👍

leandroalemao (approved):
Awesome work @ethax-ross tks!!

config/routes.rb
```diff
@@ -449,6 +449,11 @@
     resources :participants, only: %i[index show]
     resources :participant_profiles, only: [] do
       namespace :ecf do
+        resource :participant_declarations, only: [] do
+          get ":participant_declaration_id/void/new", to: "participant_declarations/void#new", as: :new_void
+          post ":participant_declaration_id/void", to: "participant_declarations/void#create", as: :create_void
+        end
```
mooktakim, line 455: Should we follow a more restful route?
```ruby
namespace :ecf do
  resource :participant_declarations, only: [] do
    resources :voids, only: [:new, :create]
  end
end
```

Hmm, maybe we should rename controller to "MakeVoidsController"

I suppose we could also do it with singular resource:
```ruby
    resource :void, only: [:new, :create]
```

app/controllers/finance/ecf/participant_declarations/void_controller.rb
cwrw: makes sense, let's keep it

app/assets/stylesheets/application.sass.scss
```diff
@@ -63,3 +63,8 @@ $govuk-new-link-styles: true;
 .govuk-summary-list__key--align-top {
   vertical-align: top;
 }
+
+.govuk-error-summary__list li {
```
cwrw, line 67: I'm surprised this isn't the default?

cwrw (commented):
looks and works great thanks @ethax-ross, my only question would be should we be changing the alert banner across the whole app for this, or limit it to this tool? Let's chat in person otherwise 👍

config/routes.rb
ethax-ross: Yeah I agree the route structure isn't ideal; I opted for consistency here though to keep it the same as the change training status action

mooktakim (approved):

cwrw (approved):
nicely done!

PR #5607, opened 2025-03-07
Add service to mark call off contracts as unused

ethax-ross:
[Jira-4125](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4125)

### Context

Due to historical/manual changes to contract data over the years we are in a position where we have a number of contracts that are either duplicates or are unused (i.e. not referenced by any financial statements).

- Add not_flagged_as_unused scope to CallOffContract

We will be referencing the used/unused status of call off contracts outside the oneoff service to prefix them as `unused_`, so it makes sense to keep the scope and terminology in the model itself.

I expect when we come to migrate the contract data we will add a `not_flagged_as_unused` scope to get the contracts that are applicable to migrate over.

- Add service to mark CallOffContract records as unused

Add `Oneoffs::MarkContractsAsUnused` service to find unused/duplicated contracts and prefix their version with `unused_`. This will enable us to identify them and exclude them from the data migration to ECF 2.0.

We identify orphaned contracts by finding those without a corresponding statement (matching on lead provider, cohort and contract version).

Duplicate contracts are initially identified by comparing lead provider, cohort and version and we then perform a deep-comparison of the `CallOffContract` and `ParticipantBand` attributes.

Any contracts that match on lead provider/cohort/contract version but differ when we perform the deep comparison are not marked as unused and we instead record how many of these we encounter during the process.

### Guidance to review

```
Oneoffs::MarkContractsAsUnused.new.perform_change(dry_run: true)

["~~~ DRY RUN ~~~",
 "Marking 246 contracts as unused (no associated finance statement)",
 "Marking 86 contracts as unused (duplicates)"]
```

These numbers appear to align with Mook's original spike.

### DIFF
--- a/app/models/call_off_contract.rb
+++ b/app/models/call_off_contract.rb
@@ -1,10 +1,14 @@
 # frozen_string_literal: true
 
 class CallOffContract < ApplicationRecord
+  UNUSED_VERSION_PREFIX = "unused_"
+
   belongs_to :lead_provider
   belongs_to :cohort
 
   has_many :participant_bands
+
+  scope :not_flagged_as_unused, -> { where.not("version LIKE ?", "#{UNUSED_VERSION_PREFIX}%") }
 
   def total_contract_value
     participant_bands.map(&:contract_value).sum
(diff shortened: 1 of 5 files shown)


mooktakim (approved):

leandroalemao (approved):
@ethax-ross Nicely done tks

PR #5612, opened 2025-03-10
Optimize `ParticipantsFilter` query

ethax-ross:
[Jira-4129](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4129)

### Context

Since updating the `ParticipantsFilter` to filter on cohort by default (to exclude 2025 and later) we have noticed timeouts on the corresponding page in the dashboard. We want to make this query more performant so that we no longer see timeouts in Sentry.

- Optimize `ParticipantsFilter` query

It looks like when we join on `cohort` it ends up pulling in a bunch of other tables (as it has to go through induction programme -> school cohort -> cohort) and the query that active record ends up with is quite large and slow.

To avoid this, we can instead use a subquery to find the induction programmes that are associated with relevant cohorts and then filter on those, avoiding the many joins.

### Guidance to review

This appears to be more performant; the underlying query is much smaller and on doing some contrived performance testing against sandbox it seems to take less time:

- Without cohort filtering; ~2.3s
- With joining on cohort to filter: ~3.2s
- Using a subquery on induction programme: ~2.5s

Benchmark script:

```
require 'benchmark'

include Pagy::Backend

results = Benchmark.measure {
  delivery_partner = DeliveryPartner.find("****")
  induction_records = DeliveryPartners::InductionRecordsQuery.new(delivery_partner:).induction_records
  training_record_states = DetermineTrainingRecordState.call(induction_records:)
  filter = DeliveryPartners::ParticipantsFilter.new(collection: induction_records, params: {}, training_record_states:)

  pagy(filter.scope.order(updated_at: :desc), page: 1, limit: 3000)
}

puts results.real
```

Test to verify before/after changes don't alter results:

```
delivery_partners = DeliveryPartner.all

results = {}

delivery_partners.each do |delivery_partner|
  induction_records = DeliveryPartners::InductionRecordsQuery.new(delivery_partner:).induction_records
  training_record_states = DetermineTrainingRecordState.call(induction_records:)

  filter_old = DeliveryPartners::ParticipantsFilter.new(collection: induction_records, params: {}, training_record_states:, new_logic: false)
  filter_new = DeliveryPartners::ParticipantsFilter.new(collection: induction_records, params: {}, training_record_states:, new_logic: true)

  old_ids = filter_old.scope.pluck(:id)
  new_ids = filter_new.scope.pluck(:id)

  results[delivery_partner.name] = old_ids == new_ids
end

results.values.uniq

irb(main):068> results.values.uniq
=> [true]
```

### DIFF
--- a/app/services/delivery_partners/induction_records_query.rb
+++ b/app/services/delivery_partners/induction_records_query.rb
@@ -35,7 +35,7 @@ module DeliveryPartners
           :school,
           user: :teacher_profile,
           induction_programme: [partnership: :lead_provider],
-          participant_profile: %i[ecf_participant_eligibility ecf_participant_validation_data],
+          participant_profile: %i[teacher_profile ecf_participant_eligibility ecf_participant_validation_data],
         )
         .select(
           "induction_records.*",
--- a/app/services/delivery_partners/participants_filter.rb
+++ b/app/services/delivery_partners/participants_filter.rb
@@ -60,10 +60,18 @@ module DeliveryPartners
     end
 
     def filter_academic_year(scoped, academic_year)
-      return scoped.joins(:cohort).where("start_year <= ?", LATEST_COHORT_TO_RETURN) if academic_year.blank?
+      return scoped.where.not(induction_programme: induction_programmes_to_ignore) if academic_year.blank?
       return scoped.none unless academic_year.to_i.in?(academic_year_options.map(&:id))
 
-      scoped.includes(:cohort).where(cohort: { start_year: academic_year })
+      scoped.where(induction_programme: induction_programmes_to_include(academic_year))
+    end
+
+    def induction_programmes_to_ignore
+      InductionProgramme.joins(school_cohort: :cohort).where("cohorts.start_year > ?", LATEST_COHORT_TO_RETURN)
+    end
+
+    def induction_programmes_to_include(academic_year)
+      InductionProgramme.joins(school_cohort: :cohort).where(cohort: { start_year: academic_year })
     end
 
     def filter_status(scoped, status)


leandroalemao (commented):
Tested locally and seems all good but I found a N+1 query in the main branch that I couldn't replicate in this branch.. not sure why.. maybe it fixed itself?

N+1 query is [here](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/serializers/delivery_partners/induction_records_serializer.rb#L22) as the `teacher_profile` is fetched from `participant_profile` and in the query we're including `teacher_profile` from user see:

https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/delivery_partners/induction_records_query.rb#L36

maybe we could do as the AB one:
https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/appropriate_bodies/induction_records_query.rb#L27

ethax-ross:
> Tested locally and seems all good but I found a N+1 query in the main branch that I couldn't replicate in this branch.. not sure why.. maybe it fixed itself?
> 
> N+1 query is [here](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/serializers/delivery_partners/induction_records_serializer.rb#L22) as the `teacher_profile` is fetched from `participant_profile` and in the query we're including `teacher_profile` from user see:
> 
> https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/delivery_partners/induction_records_query.rb#L36
> 
> maybe we could do as the AB one: https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/appropriate_bodies/induction_records_query.rb#L27

Good spot - I can replicate this on my branch as well; its not to do with my change but I've added a commit to fix it. It only happens on the CSV download endpoint which uses the serializer. The web view seems to read the TRN from the user -> teacher profile (rightly or wrongly) so I've left the preload in for that as well.

cwrw (approved):

leandroalemao (approved):

PR #5615, opened 2025-03-11
Update libxml2 to patch vulnerability

ethax-ross:
See https://security.snyk.io/vuln/SNYK-ALPINE318-LIBXML2-9380739


ethax-ross:
Closing - we'll bump the versionn of alpine instead

PR #5628, opened 2025-03-12
Prevent mentors changing cohort due to payments frozen

ethax-ross:
[Jira-4130](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4130)

### Context

Currently we allow both ECTs and mentors to change schedule from a payments frozen cohort to the active registration cohort if they have billable declarations in the frozen cohort.

Going forward we will be marking these mentors as complete which will exclude them from being eligible to transfer from a payments frozen cohort, but for a belt and braces approach we also want to guard against this explicitly in the `ChangeSchedule` service.

- Prevent mentors changing cohort due to payments frozen

Add guard clause to prevent mentors from transferring due to payments being frozen. We will still allow transferring out of a payments frozen cohort if there are no billable declarations against the participant (though they won't be flagged as `cohort_changed_after_payments_frozen` in this scenario).

### Guidance to review

Depends on the updated scopes in #5629.

### DIFF
--- a/app/services/change_schedule.rb
+++ b/app/services/change_schedule.rb
@@ -78,7 +78,7 @@ class ChangeSchedule
 private
 
   def changing_cohort_due_to_payments_frozen?
-    return false unless participant_profile
+    return false unless participant_profile&.ect?
     return false unless allow_change_to_from_frozen_cohort
     return true if participant_profile.unfinished_with_billable_declaration?(cohort:)
 
--- a/spec/services/change_schedule_spec.rb
+++ b/spec/services/change_schedule_spec.rb
@@ -73,6 +73,11 @@ RSpec.shared_examples "changing cohort and continuing training" do
   context "when there are no billable declarations and allow_change_to_from_frozen_cohort is true" do
     let(:allow_change_to_from_frozen_cohort) { true }
 
+    before { cohort.update!(payments_frozen_at: Time.zone.now) }
+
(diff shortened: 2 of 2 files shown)


app/models/participant_profile/ecf.rb
```diff
@@ -52,17 +52,15 @@ def self.ransackable_associations(_auth_object = nil)
     %w[cohort participant_identity school user teacher_profile induction_records]
   end
 
-  def self.unfinished_with_billable_declaration(cohort:, restrict_to_participant_ids: [])
-    billable_states = %w[eligible payable paid].freeze
-
-    return none unless cohort == Cohort.active_registration_cohort
+  def self.unfinished_with_billable_declaration(cohort: Cohort.active_registration_cohort, restrict_to_participant_ids: [])
+    return none if cohort.blank?
```
ltello, line 56:
```suggestion
    return none unless cohort.start_year == 2024
```

ltello (changes requested):

app/models/participant_profile/ecf.rb
cwrw: this is going to be rebased from leandro's pr so hopefully we will resolve there

mooktakim (approved):

ltello (approved):

leandroalemao (approved):

PR #5643, opened 2025-03-18
[CPDLP-4149] Pin chromedriver version locally

leandroalemao:
### Context

- Ticket: https://dfedigital.atlassian.net/browse/CPDLP-4149

A possible race condition between newer versions of chromedriver and selenium is causing some specs to fail intermittently on CI with the failure `Net::ReadTimeout` and locally with `Selenium::WebDriver::Error::UnknownError`.

Move chromeversion pin to features helper file instead, so feature specs will pass locally as well on the CI, without the errors described above.

### DIFF
--- a/spec/features_helper.rb
+++ b/spec/features_helper.rb
@@ -15,10 +15,13 @@ Capybara.register_driver :chrome_headless do |app|
   http_client.read_timeout = 120
   http_client.open_timeout = 120
 
+  # Pin Chrome version due a possible race condition between newer versions of chromedriver and selenium
+  # causing some specs to fail intermittently on CI with the failure Net::ReadTimeout
+  # See https://github.com/teamcapybara/capybara/issues/2770
   Capybara::Selenium::Driver.new(
     app,
     browser: :chrome,
-    options: Selenium::WebDriver::Options.chrome(args:),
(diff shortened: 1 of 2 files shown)


spec/features_helper.rb
```diff
@@ -15,10 +15,13 @@
...
   http_client.open_timeout = 120
 
+  # Pin Chrome version due a possible race condition between newer versions of chromedriver and selenium
+  # causing some specs to fail intermittently on CI with the failure Net::ReadTimeout
+  # See https://github.com/teamcapybara/capybara/issues/2770
   Capybara::Selenium::Driver.new(
     app,
     browser: :chrome,
-    options: Selenium::WebDriver::Options.chrome(args:),
+    options: Selenium::WebDriver::Options.chrome(args:, browser_version: "127.0.6533.119"),
```
cwrw, line 24: discussed in person: we'd like to find a long term fix for the issues locally, thanks Leandro

leandroalemao: Took a different approach and changes the browser from chrome to firefox.. no need to pin a version.. tests passing locally and also on the CI.. I can amend the PR title if the new approach is accepted.

mooktakim (approved):

ethax-ross (approved):

cwrw (approved):
thanks Leandro let's go with it :)

PR #5679, opened 2025-04-01
Clear evidence_held if not applicable when recording declaration

ethax-ross:
[Jira-4140](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4140)

### Context

The `EvidenceHeldValidator` will only validate the evidence held if it applies to the declaration being recorded.

It applies for detailed evidence types if the declaration type is not `started` or the `evidence_held` value is provided.

It applies for normal evidence types if the declaration type is not `started`.

This means if the declaration is started and for normal evidence types and the `evidence_held` value is provided it will not be validated but it will end up being persisted against the declaration.

We want to prevent lead providers persisting bad/incorrect `evidence_held` values against declarations.

- Clear evidence_held if not applicable when recording declaration

We can avoid this by either returning a new error message when an `evidence_held` is provided or by sanitizing the field as part of the validation process (clearing it if it is not required).

I opted for the latter in this commit in order to keep the behaviour of the API consistent with what lead providers expect at the moment. Its not ideal as the validator shouldn't be modifying the value in my opinion, but its the most pragmatic solution (a better solution would require a larger refactoring of the validator).

### Guidance to review

Checking the snapshot database we will have some clean up work to do after this goes in:

```
ParticipantDeclaration
  .joins(:cohort)
  .where(cohort: { detailed_evidence_types: false }, declaration_type: :started)
  .where.not(evidence_held: nil)
  .group(:evidence_held)
  .count

  => 
  {""=>11166,
   "other"=>408,
   "self-study-material completed"=>6,
   "self-study-material-completed"=>14141,
   "training_event_attendance"=>1,
   "training-event-attended"=>13455,
   "yes"=>143}
```

### DIFF
--- a/app/validators/evidence_held_validator.rb
+++ b/app/validators/evidence_held_validator.rb
@@ -54,7 +54,7 @@ class EvidenceHeldValidator < ActiveModel::Validator
       evidence_held_present?(record)
       valid_detailed_evidence_held?(record)
     else
-      return unless validate_evidence_held?(record)
+      return record.evidence_held = nil unless validate_evidence_held?(record)
 
       evidence_held_present?(record)
       valid_evidence_held?(record)
--- a/spec/requests/api/v1/participant_declarations_spec.rb
+++ b/spec/requests/api/v1/participant_declarations_spec.rb
@@ -394,14 +394,15 @@ RSpec.describe "participant-declarations endpoint spec", type: :request, mid_coh
               .to include({ "title" => "participant_id", "detail" => "Your update cannot be made as the '#/participant_id' is not recognised. Check participant details and try again." })
           end
         end
-        context "with unpermitted parameter" do
+
+        context "with parameter that is not required" do
           before { params[:data][:attributes][:evidence_held] = "test" }
 
-          it "creates the declaration" do
+          it "ignores a parameter that is not required" do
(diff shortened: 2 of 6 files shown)


leandroalemao (commented):
So are we nullifying evidence_held when it's not required even when it's a valid value? ie: "other"? Think we could only nullify when it's an invalid value then? ie: "self-study-material complete"

cwrw (approved):
Thanks Ross, not sure about cleanup and if it's required, let's check with Shahad

mooktakim (approved):

ethax-ross:
> So are we nullifying evidence_held when it's not required even when it's a valid value? ie: "other"? Think we could only nullify when it's an invalid value then? ie: "self-study-material complete"

@leandroalemao sorry I missed this -- yeah, so it will nullify `evidence_held` when its not required, even if its a valid value. I did it this way because I figured we don't want an `evidence_held` to be specified at all if the declaration is not applicable for one. If we want to accept any `evidence_held` value if its valid regardless of if the declaration type (and if the more detailed/other validations don't apply) its going to be a fairly big refactor of the validator I think.

leandroalemao (approved):

PR #5680, opened 2025-04-01
Optimise the appropriate body participants query

ethax-ross:
> ⚠️ WIP

### Context

We are seeing timeouts in Sentry from the appropriate body participants page. This is down to the underlying query being inefficient. We want to optimise the query to prevent timeouts.

- Shift ordering of AB participants into query

Instead of ordering in the controller we can shift the sort into the query, which makes more sense and enables us to more easily test it.


cwrw (approved):
🙌

leandroalemao (commented):
Hi @ethax-ross Tested all good tks 👍

leandroalemao (approved):

mooktakim (approved):

PR #5689, opened 2025-04-03
Update the 'About the API' guidance

mark-avery-dfe:
### Context

- Ticket: CPDLP-3993

### Changes proposed in this pull request:

Tightening up the copy so that it's clear and more concise.

### DIFF
--- a/docs/source/index.html.md
+++ b/docs/source/index.html.md
@@ -5,24 +5,12 @@ weight: 1
 
 # About the API
 
-The Department of Education (DfE) has developed an API for the Manage teacher continuing professional development service. 
-
-It lets lead providers view, submit and update data associated with Early Career Framework (ECF) based training. The data submitted is used to facilitate payment from the DfE to providers.
-
-Once an ECF participant has been registered to the service, data associated with them becomes available via the API.
-
-Once integrated with the API, providers can, for example, view and update participant details, or notify DfE that they have completed training by submitting a declaration.
+Lead providers must use this API to view, submit and update participant data so they receive accurate payments from the Department for Education (DfE) for their ECF-based training.
 
 ## API versions and updates
 
-The DfE works to continually improve the API service and new functionality occasionally becomes available. Guidance is reviewed and updated as necessary.
+We regularly make improvements and add new functionality to the API. View our [release notes](/api-reference/release-notes) to find out about the latest updates.
 
-If the API changes in a way that is backwards-incompatible, a new version of the API will be published. This is specified in the URL `/api/v{n}/`. For example, `/api/v1/` or  `/api/v2/` and so on. 
+If we change the API so it’s no longer able to work with older data or functionality, we’ll publish a new version. This is specified in the URL `/api/v{n}/`. For example, `/api/v1/` or `/api/v2/` and so on. We recommend lead providers use the latest version of the API.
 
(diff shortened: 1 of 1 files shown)


docs/source/index.html.md
```diff
@@ -5,24 +5,12 @@ weight: 1
...
-Once integrated with the API, providers can, for example, view and update participant details, or notify DfE that they have completed training by submitting a declaration.
+Lead providers must use this API to view, submit and update participant data so they receive accurate payments from the Department for Education (DfE) for their ECF-based training.
 
 ## API versions and updates
 
-The DfE works to continually improve the API service and new functionality occasionally becomes available. Guidance is reviewed and updated as necessary.
-
-If the API changes in a way that is backwards-incompatible, a new version of the API will be published. This is specified in the URL `/api/v{n}/`. For example, `/api/v1/` or  `/api/v2/` and so on. 
-
-When the DfE publishes a new API version, only one previous version will remain supported. For example, when a `v4` is released then `v2` will be discontinued.
```
cwrw, line 22: I think we need to keep this line as we committed to this previous version rule?

docs/source/index.html.md
```diff
@@ -5,24 +5,12 @@ weight: 1
...
 # About the API
 
-The Department of Education (DfE) has developed an API for the Manage teacher continuing professional development service. 
-
-It lets lead providers view, submit and update data associated with Early Career Framework (ECF) based training. The data submitted is used to facilitate payment from the DfE to providers.
-
-Once an ECF participant has been registered to the service, data associated with them becomes available via the API.
-
-Once integrated with the API, providers can, for example, view and update participant details, or notify DfE that they have completed training by submitting a declaration.
+Lead providers must use this API to view, submit and update participant data so they receive accurate payments from the Department for Education (DfE) for their ECF-based training.
```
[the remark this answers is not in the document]
mark-avery-dfe: How's about this?

Lead providers must use this API to view, submit and update participant data so they receive accurate payments from the Department for Education (DfE) for their ECF-based training.

Once a participant has been registered to the service by their school, data associated with them becomes available to lead providers via the API.

docs/source/index.html.md
mark-avery-dfe: Cool. Let's go with: 

## API versions and updates

We regularly make improvements and add new functionality to the API. View our [release notes](/api-reference/release-notes) to find out about the latest updates.

If we change the API so it’s no longer able to work with older data or functionality, we’ll publish a new version. This is specified in the URL `/api/v{n}/`. For example, `/api/v1/` or `/api/v2/` and so on. We recommend lead providers use the latest version of the API.

When we publish a new API version, only one previous version will remain supported. For example, when a `v4` is released then `v2` will be discontinued.

Note, there'll be an exception to this rule for `v3`. We're supporting `v1` for an extended period while we work with providers on transition plans.

When we make non-breaking updates (sometimes referred to as backwards compatible updates), we will not re-version the API. An example of a non-breaking change would be when we introduce a new attribute without removing an existing one.

docs/source/index.html.md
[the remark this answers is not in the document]
cwrw: looks good thanks

docs/source/index.html.md
cwrw: looks good thanks

mark-avery-dfe (commented):
Sorted

cwrw (approved):

PR #5691, opened 2025-04-04
Add ProgrammeTypeMappings service

ethax-ross:
As part of the work to reword programme type changes across the user-facing portions of the application we want to centralise the mappings to/from new/old values we display.

Add the `ProgrammeTypeMappings` service to centralise the programme type changes.


app/services/programme_type_mappings.rb
```diff
@@ -0,0 +1,31 @@
...
+    def training_programme(training_programme:)
+      return training_programme.humanize unless mappings_enabled?
+
+      case training_programme
+      when "full_induction_programme", "school_funded_fip"
+        "Provider Led"
+      else
+        "School Led"
+      end
+    end
```
mooktakim, line 29: `humanize` should be outside of this service class. I'm thinking:
```ruby
    def training_programme(training_programme:)
      return training_programme unless mappings_enabled?

      case training_programme
      when "full_induction_programme", "school_funded_fip"
        "provider_led"
      when "core_induction_programme", "design_our_own"
        "school_led"
      else
        training_programme
      end
    end
```

ethax-ross: yeah that's a good point, I'll update 👍

cwrw (approved):

spec/services/programme_type_mappings_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+          it { is_expected.to eq("provider_led") }
+        end
+
+        context "when training_programme is 'core_induction_programme'" do
+          let(:training_programme) { build(:induction_programme, :cip).training_programme }
+
+          it { is_expected.to eq("school_led") }
+        end
+
+        context "when training_programme is 'design_our_own'" do
```
cwrw, line 77: looks like we have genuine failures here in the spec

app/services/programme_type_mappings.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class ProgrammeTypeMappings
+  class << self
+    def mappings_enabled?
+      FeatureFlag.active?(:programme_type_changes_2025)
+    end
+
+    def withdrawl_reason(reason:)
```
cwrw, line 9:
```suggestion
    def withdrawal_reason(reason:)
```

spec/services/programme_type_mappings_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      before { FeatureFlag.deactivate(:programme_type_changes_2025) }
+
+      it { is_expected.not_to be_mappings_enabled }
+    end
+  end
+
+  describe "mappings" do
+    before { allow(described_class).to receive(:mappings_enabled?) { mappings_enabled } }
+
+    describe ".withdrawl_reason" do
```
cwrw, line 23:
```suggestion
    describe ".withdrawal_reason" do
```

spec/services/programme_type_mappings_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        context "when training_programme is 'full_induction_programme'" do
+          let(:training_programme) { build(:induction_programme, :fip).training_programme }
+
+          it { is_expected.to eq(training_programme) }
+        end
+
+        context "when training_programme is 'school_funded_fip'" do
+          let(:training_programme) { build(:induction_programme, :school_funded_fip).training_programme }
+
+          it { is_expected.to eq(training_programme) }
```
cwrw, line 102: could we add specs for the other examples please

leandroalemao (approved):

cwrw (approved):

mooktakim (approved):

PR #5695, opened 2025-04-04
Remap withdrawl reason when `programme_type_changes_2025` feature is enabled

ethax-ross:
[Jira-4186](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4186)

### Context

We want to remap the withdrawl reason `school-left-fip` to `switched-to-school-led` when the `programme_type_changes_2025` feature flag is enabled. This is part of a wider renaming of programme types.

- Map withdrawl reason to new value in v3 ParticipantSerializer

As part of the programme type changes we want to map the `switched-to-school-led` withdrawl reason to `school-left-fit` when the `programme_type_changes_2025` feature flag is enabled.

Update the v3 `ParticipantSerializer` to map the withdrawl reason. The v1/2 serializers don't appear to expose the withdrawl reason.

- Update WithdrawParticipant with program type changes

When the `programme_type_changes_2025` feature is enabled we want to support sending the new `switched-to-school-led` withdrawl reason in place of `school-left-fip`.

Avoids changing the `WITHDRAW_REASONS` constant to prevent unintended changes
elsewhere.

- Update ChangeTrainingStatusForm to support new programme types

When the `programme_type_changes_2025` feature is enabled we want to display and accept the new withdrawl reason types in the change training status form.

- Update AssuranceReport::CsvSerializer with programme type changes

As with other places, we want to remap the withdrawl reason when the `programme_type_changes_2025` feature is enabled.

Confusingly, the `AssuranceReport::Query` will only surface withdrawl reasons in the `Training Status Reason` column, even if the current state of the participant is not withdrawn.

### Guidance for review

I thought about extracting the `WITHDRAWN_REASONS` into a method so we could do the remapping there instead of both in the form and csv serializer, but I figured since the feature is temporary its better to keep it consistent with `DEFERRAL_REASONS` in a constant for the short-term.

### DIFF
--- a/spec/forms/finance/ecf/change_training_status_form_spec.rb
+++ b/spec/forms/finance/ecf/change_training_status_form_spec.rb
@@ -1,38 +1,112 @@
 # frozen_string_literal: true
 
-RSpec.describe Finance::ECF::ChangeTrainingStatusForm, type: :model do
+RSpec.shared_context "changing training status form" do
   subject(:form) { described_class.new(params) }
 
-  let(:cpd_lead_provider)   { create(:cpd_lead_provider, :with_lead_provider) }
+  let(:cpd_lead_provider) { create(:cpd_lead_provider, :with_lead_provider) }
   let(:induction_programme) { create(:induction_programme, :fip) }
-  let!(:induction_record)   { Induction::Enrol.call(participant_profile:, induction_programme:) }
-  let(:params)              { { participant_profile:, training_status: "deferred", reason: "bereavement", induction_record: } }
+  let!(:induction_record) { Induction::Enrol.call(participant_profile:, induction_programme:) }
+  let(:params) { { participant_profile:, training_status:, reason: "bereavement", induction_record: } }
+  let(:training_status) { "deferred" }
 
+  describe "validations" do
+    it { is_expected.to validate_inclusion_of(:training_status).in_array(ParticipantProfile.training_statuses.values) }
+    it { is_expected.to validate_inclusion_of(:reason).in_array(ParticipantProfile::DEFERRAL_REASONS) }
+
+    context "when the training_status is 'withdrawn'" do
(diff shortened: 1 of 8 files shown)


spec/serializers/api/v3/ecf/participant_serializer_spec.rb
```diff
@@ -115,15 +115,37 @@ module ECF
...
-                ect_profile.participant_profile_states.last.update!(state: "withdrawn", reason: "other")
+                ect_profile.participant_profile_states.last.update!(state: "withdrawn", reason:)
               end
 
               it "includes a withdrawal object" do
                 expect(ecf_enrolments[0][:withdrawal][:reason]).to eq("other")
                 expect(ecf_enrolments[0][:withdrawal][:date]).to eql(ect_profile.participant_profile_state.created_at.rfc3339)
               end
+
+              context "when the withdrawl reason is 'school-left-fip'" do
```
leandroalemao, line 130: `withdrawal`

spec/forms/finance/ecf/change_training_status_form_spec.rb
```diff
@@ -1,64 +1,119 @@
...
+      before { allow(ProgrammeTypeMappings).to receive(:mappings_enabled?) { false } }
+
+      it { expect(form.reason_options["withdrawn"]).not_to include("switched-to-school-led") }
+      it { expect(form.reason_options["withdrawn"]).to include("school-left-fip") }
+    end
+  end
+
+  describe "#save" do
+    context "valid params" do
+      it "should change training status" do
```
leandroalemao, line 86: it "changes training status"

leandroalemao (approved):
Just a few very minor comments... tks

mooktakim (approved):

cwrw (approved):

PR #5698, opened 2025-04-07
[CPDLP-4158] Update appropriate bodies views for the new programme types

leandroalemao:
### Context

- Ticket: [CPDLP-4158](https://dfedigital.atlassian.net/browse/CPDLP-4158)

We need to update a few places we mention programme types to the new values.

- `AppropriateBodies::Participants::TableRow` to show provider/school led if feature flag is on instead of CIP/FIP;
- Update `AppropriateBodies::InductionRecordsSerializer` to show provider/school led if feature flag is on instead of CIP/FIP;
- Update mentions to FIP/CIP in ABs documentation. Please see [PR#433](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/433);
### Guidance to review

Review app

[CPDLP-4158]: https://dfedigital.atlassian.net/browse/CPDLP-4158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/programme_type_mappings.rb
+++ b/app/services/programme_type_mappings.rb
@@ -30,40 +30,9 @@ class ProgrammeTypeMappings
       end
     end
 
-    def training_programme_friendly_name(name, short: false)
-      long_names = if mappings_enabled?
-                     {
-                       "full_induction_programme" => "Provider led",
-                       "core_induction_programme" => "School led",
-                       "design_our_own" => "School led",
-                       "school_funded_fip" => "Provider led",
-                     }
-                   else
-                     {
-                       "full_induction_programme" => "Full induction programme",
-                       "core_induction_programme" => "Core induction programme",
-                       "design_our_own" => "Design our own",
-                       "school_funded_fip" => "School funded full induction programme",
-                     }
-                   end
-
-      short_names = if mappings_enabled?
-                      {
-                        "full_induction_programme" => "Provider led",
-                        "core_induction_programme" => "School led",
-                        "design_our_own" => "School led",
-                        "school_funded_fip" => "Provider led",
-                      }
-                    else
-                      {
-                        "full_induction_programme" => "FIP",
-                        "core_induction_programme" => "CIP",
-                        "design_our_own" => "Design our own",
-                        "school_funded_fip" => "School funded FIP",
-                      }
-                    end
-
-      short ? short_names.fetch(name) : long_names.fetch(name)
+    def training_programme_friendly_name(training_programme, length: :long)
+      translation_key = training_programme(training_programme:)
+      I18n.t("training_programme.#{length}.#{translation_key}")
     end
   end
 end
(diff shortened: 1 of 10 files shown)


app/services/programme_type_mappings.rb
```diff
@@ -29,5 +29,41 @@ def training_programme(training_programme:)
         training_programme
       end
     end
+
+    def training_programme_friendly_name(name, short: false)
+      long_names = if mappings_enabled?
```
ethax-ross, line 34: Would this logic make more sense in a view helper called directly from the view component its used in, instead of in the service/`InductionRecord` model? the copy could even go in the translations file so we're not redefining the mappings, something like:

```
en:
  training_programme:
    short:
      full_induction_programme: FIP
      provider_led: Provider led
      ...
    long:
      full_induction_programme: Full induction programme
      provider_led: Provider led
      ...
```

Then the helper:

```
def training_programme_friendly_name(training_programme, length: :long)
  translation_key = ProgrammeTypeMapping.training_programme(training_programme:)
  I18n.t("training_programme_#{length}_#{translation_key}")
end
```

leandroalemao: Tks @ethax-ross I used your suggestion but kept it in the service/model as the same method is used in a serializer as well.. not only in views.

spec/models/induction_record_spec.rb
```diff
@@ -464,4 +464,30 @@
...
+    subject { induction_record.induction_type }
+
+    context "when enrolled in cip " do
+      let(:induction_programme) { create(:induction_programme, :cip) }
+      let(:induction_record) { Induction::Enrol.call(participant_profile: create(:ect_participant_profile), induction_programme:) }
+
+      it "calls ProgrammeTypeMappings with correct params" do
+        expect(ProgrammeTypeMappings).to receive(:training_programme_friendly_name).with("core_induction_programme", length: :short)
+
+        subject
```
ethax-ross, line 478: Worth adding an expectation on the return values here as well for good measure?

spec/serializers/appropriate_bodies/induction_records_serializer_spec.rb
```diff
@@ -35,6 +35,21 @@ module AppropriateBodies
           induction_tutor: induction_record.school.contact_email,
         )
       end
+
+      context "when the programme_type_changes_2025 feature flag is enabled" do
+        before { FeatureFlag.activate(:programme_type_changes_2025) }
+
+        it "returns valid hash" do
```
ethax-ross, line 42: `returns the correct induction_type`

ethax-ross (approved):
Looks good, just two very minor suggestions

mooktakim (approved):

cwrw (commented):
thanks Leandro let's chat tomorrow about the new requirements for dashified content, sorry about that

PR #5705, opened 2025-04-09
Fix failing analytics upsert jobs on review deploy

ethax-ross:
[Jira-4196](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4196)

### Context

We see a lot of errors in logit when review apps are deployed. These are to do with jobs that are running where the data the job was serialized with no longer exists in the database. We want to fix this to prevent review deploys from spamming the logs/eating into our quota.

- Run analytics job after_commit instead of after_save

There is a race condition when the various models that have an `update_analytics` callback are persisted in a transaction (as is the case with some of the seed data).

The `after_save` will enqueue the job before the transaction is commited; if a worker picks up the job it attempts to deserialize and verify the model argument and raises an error as it hasn't been commited to the database yet.

Using `after_commit` instead of `after_save` ensures the model will be in the database. As all changes aren't readily available to `after_commit` (`previous_changes` only retains changes from the last save, which is often empty) we are leveraging `ar_transaction_changes`, which retains all changes from a commit across multiple saves.

- Avoid serializing entire objects for job arguments

We currently pass in (and therefore serialize) whole objects for the various analytic job parameters. This is inefficient and also causes an issue if the object that was serialized is deleted before the job is picked up by a worker (as it can't then deserialize it).

To avoid this issue we can just pass the `id` of the object, which will take up much less space in Redis, and load it into memory when the worker picks up the job. We can guard against the object no longer existing to avoid potential race
conditions/errors.

### Guidance to review

We get these errors when deploying both a fresh review app and when we do a redeploy.

The first commit fixed 90% of the errors we see; when I checked the errors I was able to find the objects in the database it complained were not there, so I expect it's down to the job being queued on save rather than after commit and is the result of a race condition.

The remaining 10% were all for the participant profile analytics jobs. I haven't been able to verify this as there's a lot going on in the seeds, but for these when I checked the database they objects didn't exist and I expect some are being deleted somewhere during seeding causing the error when the worker gets around to the job. To prevent this I changed to serialize only the id of the object and return if it no longer exists. This probably makes the first commit unnecessary, but doing it `after_commit` feels like the better approach regardless so I left the change in.

You can verify this change by re-running the review deploy and checking the logs [here](https://kibana-uk1.logit.io/s/b2876053-0173-44fc-983d-99567292ba31/app/data-explorer/discover#?_a=(discover:(columns:!(app.message,kubernetes.deployment.name,app.payload.job_class,app.exception.cause.message),isDirty:!t,sort:!(!('@timestamp',desc),!(app.payload.job_class,desc))),metadata:(indexPattern:'filebeat-*',view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2025-04-09T13:47:00.000Z',to:now))&_q=(filters:!(('$state':(store:appState),exists:(field:app.exception.stack_trace),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:app.exception.stack_trace,negate:!f,type:exists,value:exists)),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:kubernetes.namespace,negate:!f,params:(query:cpd-development),type:phrase),query:(match_phrase:(kubernetes.namespace:cpd-development))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:kubernetes.container.name,negate:!f,params:(query:cpd-ecf-review-5705-worker),type:phrase),query:(match_phrase:(kubernetes.container.name:cpd-ecf-review-5705-worker)))),query:(language:kuery,query:''))).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,9 @@ gem "ar-uuid", "~> 0.2.3"
 # Use Puma as the app server
 gem "puma", "~> 6.0"
 
+# after_commit track changes
+gem "ar_transaction_changes"
+
 # Soft delete
 gem "discard", "~> 1.4"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -118,6 +118,8 @@ GEM
     amazing_print (1.7.2)
     ar-uuid (0.2.3)
       activerecord
+    ar_transaction_changes (1.1.9)
+      activerecord (>= 5.2.0)
     ast (2.4.3)
     auto_strip_attributes (2.6.0)
       activerecord (>= 4.0)
@@ -745,6 +747,7 @@ DEPENDENCIES
   activerecord-session_store (~> 2.2)
   amazing_print
   ar-uuid (~> 0.2.3)
+  ar_transaction_changes
   auto_strip_attributes (~> 2.6)
   axe-core-rspec
   bootsnap (~> 1.18)
--- a/app/jobs/analytics/upsert_ecf_appropriate_body_job.rb
+++ b/app/jobs/analytics/upsert_ecf_appropriate_body_job.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 class Analytics::UpsertECFAppropriateBodyJob < ApplicationJob
-  def perform(appropriate_body:)
-    Analytics::ECFAppropriateBodyService.upsert_record(appropriate_body)
+  def perform(appropriate_body_id:)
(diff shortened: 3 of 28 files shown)


cwrw (approved):
🙌

leandroalemao (approved):

mooktakim (approved):

PR #5736, opened 2025-04-24
[CPDLP-4218] Fix memory issue when seeding review apps

leandroalemao:
### Context

- Ticket: [CPDLP-4218](https://dfedigital.atlassian.net/browse/CPDLP-4218)

It’s been flagged a few times that seeds aren’t running to completion in review apps. Sometimes for instance API tokens aren’t created or anything after that.

- Refactor `add_declarations.rb` to improve performance.
- Batch Processing: Breaking the large transaction into smaller batches to prevent timeouts:
  - Lead providers batch
  - Participants batch
  - Cohort batch 
- Pre-loading Data: Loading collections upfront to reduce repeated database queries
- Bulk Operations: Creating related records in groups rather than individually
- Query Reduction: Using indexing and in-memory operations to minimize database hits
- Optimized Time Handling: Reducing the number of time travel operations

### Guidance to review

Review app

[CPDLP-4218]: https://dfedigital.atlassian.net/browse/CPDLP-4218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/db/new_seeds/base/add_declarations.rb
+++ b/db/new_seeds/base/add_declarations.rb
@@ -51,10 +51,10 @@ cpd_lead_providers.each_slice(LEAD_PROVIDER_BATCH_SIZE) do |cpd_lead_provider_ba
       # Create participants in a single transaction per cohort
       ActiveRecord::Base.transaction do
         participant_count.times do
-          FactoryBot.create(:ect, :eligible_for_funding, cohort:, lead_provider:)
-          FactoryBot.create(:mentor, :eligible_for_funding, cohort:, lead_provider:)
-          FactoryBot.create(:ect, :eligible_for_funding, :with_extended_schedule, cohort:, lead_provider:)
-          FactoryBot.create(:mentor, :eligible_for_funding, :with_extended_schedule, cohort:, lead_provider:)
+          FactoryBot.create(:ect, :eligible_for_funding, cohort:, lead_provider:, user: FactoryBot.create(:user, full_name: Faker::Name.name))
+          FactoryBot.create(:mentor, :eligible_for_funding, cohort:, lead_provider:, user: FactoryBot.create(:user, full_name: Faker::Name.name))
+          FactoryBot.create(:ect, :eligible_for_funding, :with_extended_schedule, cohort:, lead_provider:, user: FactoryBot.create(:user, full_name: Faker::Name.name))
+          FactoryBot.create(:mentor, :eligible_for_funding, :with_extended_schedule, cohort:, lead_provider:, user: FactoryBot.create(:user, full_name: Faker::Name.name))
         end
       end
     end
--- a/db/new_seeds/base/add_declarations.rb
+++ b/db/new_seeds/base/add_declarations.rb
@@ -87,47 +87,6 @@ cpd_lead_providers.find_each(batch_size: LEAD_PROVIDER_BATCH_SIZE) do |cpd_lead_
 
     # Process each schedule group
     profiles_by_schedule.each do |schedule, profiles|
-      # Pre-create all needed milestones for this schedule in one batch
-      needed_milestone_types = Set.new
-
-      profiles.each do |participant_profile|
-        profile_declaration_types = declaration_types.dup
-
-        # Filter out existing declaration types
-        existing_types = participant_profile.participant_declarations.pluck(:declaration_type)
-        profile_declaration_types.reject! { |type| type.in?(existing_types) }
-
-        # Filter based on mentor/schedule conditions
-        if participant_profile.mentor? && schedule.cohort.mentor_funding?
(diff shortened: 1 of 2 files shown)


ethax-ross (commented):
This generally makes sense, but can you add a description/some context for the changes in the first commit, I'm not sure I understand what the issue was - were we loading too much into memory and needed to batch more as well as reducing the overall quantities?

leandroalemao:
> This generally makes sense, but can you add a description/some context for the changes in the first commit, I'm not sure I understand what the issue was - were we loading too much into memory and needed to batch more as well as reducing the overall quantities?

Hi @ethax-ross Yeah basically that but keeping the same quantities. Added some more context in the PR description as well. Tksss

db/new_seeds/base/add_declarations.rb
```diff
@@ -1,72 +1,209 @@
...
+        FactoryBot.create(:ecf_extended_schedule, cohort:)
+      end
+    end
+  end
+end
+
+# Process lead providers in batches
+cpd_lead_providers = CpdLeadProvider.joins(:lead_provider).to_a
+
+cpd_lead_providers.each_slice(LEAD_PROVIDER_BATCH_SIZE) do |cpd_lead_provider_batch|
```
cwrw, line 43: do we really need to batch lead providers, there should only be 7?

db/new_seeds/base/add_declarations.rb
```diff
@@ -1,72 +1,209 @@
...
+      cohort_ids_batch.each do |cohort_id|
+        cohort = cohorts.find { |c| c.id == cohort_id }
+        FactoryBot.create(:ecf_schedule, cohort:)
+      end
+    end
+  end
+end
+
+if cohort_ids_without_extended_schedule.any?
+  cohort_ids_without_extended_schedule.each_slice(COHORT_BATCH_SIZE) do |cohort_ids_batch|
```
cwrw, line 30: same with cohorts its only a handful?

db/new_seeds/base/add_declarations.rb
```diff
@@ -1,72 +1,209 @@
...
-            cohort: schedule.cohort,
-          )
+    # Process participant profiles in batches
+    active_profiles = lead_provider.active_ecf_participant_profiles.to_a
+
+    # Pre-calculate possible states and declaration types to avoid repeated calculations
+    states = ParticipantDeclaration.states.keys
+    declaration_types = %w[started retained-1 retained-2 retained-3 retained-4 extended-1 extended-2 extended-3 completed]
+
+    active_profiles.each_slice(PARTICIPANT_BATCH_SIZE) do |profile_batch|
```
cwrw, line 69: wondering if we can use `find_each(batch_size: 50)` to avoid preloading all profiles into memory?

db/new_seeds/base/add_declarations.rb
```diff
@@ -1,72 +1,209 @@
...
+              profile_type: participant_profile.ect? ? :ect : :mentor,
+            }
+          end
+
+          # Create declarations in a single transaction
+          next unless declarations_to_create.any?
+
+          ActiveRecord::Base.transaction do
+            declarations_to_create.each do |declaration_data|
+              milestone_date = declaration_data.delete(:milestone_date)
```
cwrw, line 195: is there a reason we're deleting this data?

leandroalemao: Hi @cwrw It's just deleting from the hash which goes as a param to the factory and using the value to `travel_to`

db/new_seeds/base/add_declarations.rb
leandroalemao: Hi @cwrw I've kept the batch for lead providers loop as the participants loop is inside of it.. so we process participant in batches for a batch of lead providers at the time.

db/new_seeds/base/add_declarations.rb
```diff
@@ -1,72 +1,201 @@
...
+    # Process each schedule group
+    profiles_by_schedule.each do |schedule, profiles|
+      # Pre-create all needed milestones for this schedule in one batch
+      needed_milestone_types = Set.new
+
+      profiles.each do |participant_profile|
+        profile_declaration_types = declaration_types.dup
+
+        # Filter out existing declaration types
+        existing_types = participant_profile.participant_declarations.pluck(:declaration_type)
```
cwrw, line 97: we have a bit of duplication here with similar logic below on line 146 and continuing, can we try and avoid that?

leandroalemao: Hi @cwrw I've removed the loop as the milestones are being created with the schedules at the top of the script so no need for that inner loop anymore. tks

cwrw (approved):

PR #5746, opened 2025-04-30
Closing 2022 cohort

ltello:
### Context

[Ticket](https://dfedigital.atlassian.net/browse/CST-2850)

- add feature flag for closing 2022 cohort
- prevent mentors in a frozen cohort from being moved by the automatic cohort change triggers
- define 2024 as the only cohort to move ECTs from a frozen cohort.
- add seed data to test participants in 2022

### DIFF
--- a/app/forms/schools/add_participants/base_wizard.rb
+++ b/app/forms/schools/add_participants/base_wizard.rb
@@ -493,7 +493,7 @@ module Schools
       end
 
       def cohort_for_transfer
-        cohort = Cohort.active_registration_cohort
+        cohort = Cohort.destination_from_frozen_cohort
         return cohort if existing_participant_profile&.unfinished?(cohort:)
 
         existing_participant_cohort || existing_participant_profile&.schedule&.cohort
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -3,6 +3,7 @@
 class Cohort < ApplicationRecord
   has_paper_trail
 
+  DESTINATION_START_YEAR_FROM_A_FROZEN_COHORT = 2024
   START_YEAR_2020 = 2020
 
   INITIAL_COHORT_START_DATE = Date.new(2021, 9, 1)
@@ -29,6 +30,10 @@ class Cohort < ApplicationRecord
 
   def self.current
     starting_within(Date.current - 1.year + 1.day, Date.current)
+  end
+
+  def self.destination_from_frozen_cohort
+    where(start_year: DESTINATION_START_YEAR_FROM_A_FROZEN_COHORT).first
   end
 
   def self.next
--- a/app/models/cohort.rb
+++ b/app/models/cohort.rb
@@ -33,7 +33,7 @@ class Cohort < ApplicationRecord
   end
 
   def self.destination_from_frozen_cohort
-    where(start_year: DESTINATION_START_YEAR_FROM_A_FROZEN_COHORT).first
+    find_by!(start_year: DESTINATION_START_YEAR_FROM_A_FROZEN_COHORT)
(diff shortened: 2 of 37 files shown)


app/forms/schools/add_participants/base_wizard.rb
```diff
@@ -493,7 +493,7 @@ def cohort_for_ect_with_induction_start_date
       end
 
       def cohort_for_transfer
```
tonyheadford, line 495: I assume this is only used when it's a "frozen" transfers?  The naming isn't clear to me so I'm not sure.

tonyheadford (approved):
LGTM, one comment with a query as I wasn't sure. This also should should get a review from @cwrw as she has much more context than me regarding the business rules.

app/models/cohort.rb
```diff
@@ -31,6 +32,10 @@ def self.current
     starting_within(Date.current - 1.year + 1.day, Date.current)
   end
 
+  def self.destination_from_frozen_cohort
+    find_by!(start_year: DESTINATION_START_YEAR_FROM_A_FROZEN_COHORT)
```
cwrw, line 36: we have a similar constant here `ParticipantProfile::ECF::DESTINATION_COHORT_WHEN_MOVING_PARTICIPANTS_TO_FROM_A_FROZEN_COHORT` maybe we can use that instead of defining another one in cohort?

app/models/participant_profile/ecf.rb
```diff
@@ -67,7 +63,7 @@ def self.unfinished_with_billable_declaration(cohort:, restrict_to_participant_i
       .where("participant_declarations.state IN (?) AND declaration_type != ?", Finance::StatementLineItem::BILLABLE_STATES, "completed")
       .where.not(id: completed_billable_declarations.select(:participant_profile_id))
       .distinct
-
+    query = query.where(type: "ParticipantProfile::ECT") if FeatureFlag.active?(:closing_2022)
```
cwrw, line 66: I think last time we agreed we are going to add guard clauses to where this is relevant, unfortunately this is used in the API for change schedule, and we already implemented it by adding a guard clause not a feature flag here: https://github.com/DFE-Digital/early-careers-framework/blob/e4e99184e8a320a3b27e5efc70ab1e0405e71ece/app/services/change_schedule.rb#L81 I don't we can use feature flags as this will reverse our already live implementation?

app/models/participant_profile/ecf.rb
```diff
@@ -180,14 +176,15 @@ def unfinished_with_billable_declaration?(cohort:)
   end
 
   def unfinished_with_no_billable_declaration?(cohort:)
+    return false if mentor? && FeatureFlag.active?(:closing_2022)
```
cwrw, line 179: same here with feature flag

app/models/participant_profile/ecf.rb
```diff
@@ -180,14 +176,15 @@ def unfinished_with_billable_declaration?(cohort:)
...
     return false if completed_training?
-    return false unless cohort == Cohort.active_registration_cohort
+    return false unless cohort&.start_year == Cohort::DESTINATION_START_YEAR_FROM_A_FROZEN_COHORT
     return false unless schedule&.cohort&.payments_frozen?
 
-    participant_declarations.none? { |declaration| declaration.billable? && declaration.completed? }
+    participant_declarations.none?(&:billable?)
```
cwrw, line 184: why are we removing the completed?

ltello: Because it should have been removed last year. It is unnecessary

cwrw (approved):

PR #5767, opened 2025-05-08
Fix Dockerfile warnings

cwrw:
### Context

We are getting warnings from the docker file run for:
- Absolute paths
- Secrets in ARG
- JSON arguments in commands

- Absolute paths: set absolute path for work dir by adding forward slash
- Secrets in ARGs: unfortunately because we are setting the `SECRET_KEY_BASE` it will always fail even if its a placeholder and not really a secret. I had to disable it for the whole file
- JSON arguments: set the arguments in JSON format however due to us using `&&` we had to also set the shell otherwise it will fail see https://docs.docker.com/reference/build-checks/json-args-recommended/#explicitly-specify-the-shell

### Guidance to review

Keen to hear any other ways we can resolve the JSON arguments as tried many variations without the shell but didn't work and kept failing

Build warning can be found here: https://docs.docker.com/reference/build-checks/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ COPY docs /docs
 COPY public /public
 COPY swagger /swagger
 
-WORKDIR docs
+WORKDIR /docs
 RUN bundle exec middleman build --build-dir=../public/api-reference
 
 # Stage 1: Download gems and node modules.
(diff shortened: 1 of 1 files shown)


ethax-ross (approved):
Looks good. The other way I've seen people work around the asset precompilation requiring the secret key base is to do something like this in the config:

```
config.require_master_key = true unless Rake.application.top_level_tasks.include?('assets:precompile')
```

ethax-ross (approved):

PR #5771, opened 2025-05-08
Update analytics_blocklist.yml

t-phillips-dfe:
Updated blocklist to allow user ID and name to be streamed through to the ECF Voided Declaration Dashboard

### Context
I need user name and ID streamed through to ECF big query to allow user names to appear on the ECF Voided Declarations Dashboard

### Changes proposed in this pull request
Removed user_id, name, created_at and updated_at from the block list

### Guidance to review
Please get in touch with Theo Phillips if there are any problems / concerns

PR #5773, opened 2025-05-08
Update user analytics fields

peteryates:
### Context

Allow user ID and name to be streamed through to the ECF Voided Declaration Dashboard and added created_at and updated_at dates to users table

Fixes #5770, #5771, #5772

### DIFF
--- a/config/analytics_hidden_pii.yml
+++ b/config/analytics_hidden_pii.yml
@@ -8,3 +8,6 @@
   :emails:
   - to
   - personalisation
+  :users:
+  - id
+  - full_name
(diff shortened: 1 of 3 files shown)


cwrw (approved):

config/analytics_hidden_pii.yml
```diff
@@ -8,3 +8,6 @@
   :emails:
   - to
   - personalisation
+  :users:
+  - id
+  - full_name
```
cwrw, line 13: just to confirm it's ok to send this under hidden pii?

peteryates: Yeah, it'll be encrypted and masked in BQ and only visible to people with special privileges.

PR #5795, opened 2025-05-16
Fix the link to guidance on how a mentor gets a TRN in ECF1.

claire-hughez:
### Context

- Ticket:  https://github.com/DFE-Digital/register-ects-project-board/issues/1577

Fix the link for how to get a TRN.

### Guidance to review

Make sure the link works to the new guidance.

I don't think cleaning up the pages matters no much given we're hopefully decommissioning ECF1 relatively soon.


peteryates (approved):

tonyheadford (approved):
We had been trying to put external urls in the [routes file here](https://github.com/DFE-Digital/early-careers-framework/blob/114162a1c09bed323213bfe1a9aa1cc97bc0e39a/config/routes.rb#L22) so that would be better but it's not a blocker

PR #5798, opened 2025-05-16
feat(2234): add staging to deploy workflow

pasitconsulting:
### Context

[- Ticket: ](https://trello.com/c/xzhTdSFq/2234-ecf-plan-for-dr-test-may-2025)

add staging envt to list of envts in workflow_dispatch for deploy workflow

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,6 +14,7 @@ on:
         type: choice
         options:
           - review
+          - staging
   push:
     branches:
       - main


.github/workflows/deploy.yml
```diff
@@ -14,6 +14,7 @@ on:
         type: choice
         options:
           - review
+          - staging
```
RMcVelia, line 17: I'm not sure we need this. There is a manual deployment workflow which includes staging i.e. deploy_to_environment.yml

PR #5805, opened 2025-05-20
chore(2234):  ecf plan for dr test may 2025 -  - fix to app-name in restore job

pasitconsulting:
### Context

- Ticket:  https://trello.com/c/xzhTdSFq/2234-ecf-plan-for-dr-test-may-2025

- correct envt var
- add staging to list of envts

### Guidance to review

before fix:-
https://github.com/DFE-Digital/early-careers-framework/actions/runs/15069149480

after fix:-
https://github.com/DFE-Digital/early-careers-framework/actions/runs/15140555006

https://github.com/DFE-Digital/early-careers-framework/actions/runs/15183210172
Screenshot from 2025-05-22 10-47-45


.github/workflows/deploy.yml
```diff
@@ -14,6 +14,7 @@ on:
         type: choice
         options:
           - review
+          - staging
```
cwrw, line 17: isn't this similar to https://github.com/DFE-Digital/early-careers-framework/pull/5798/files#r2093067771 and we don't need it?

cwrw: we have a manual deployment action here instead https://github.com/DFE-Digital/early-careers-framework/actions/workflows/deploy_to_environment.yml which includes staging

shaheislamdfe (approved):

global_config/staging.sh
```diff
@@ -3,6 +3,6 @@ AZURE_SUBSCRIPTION=s189-teacher-services-cloud-test
 CONFIG=staging
 CONFIG_SHORT=st
 CONFIG_LONG=staging
-DEPLOY_ENV=${CONFIG}
+DEPLOY_ENV=staging
```
RMcVelia, line 6: this doesn't fix the workflow issue, maybe better to use 'app-name: ${{ env.SERVICE_NAME }}-${{ inputs.environment }}-web' instead of app-name: ${{ env.SERVICE_NAME }}-${CONFIG}-web as per the template?

pasitconsulting: amended accordingly

RMcVelia (approved):

PR #5861, opened 2025-06-18
[1751] Amend record declaration validation to fix the participant profile state bug

leandroalemao:
### Context

The issue arises from transfers, either admin or SITs where partnerships aren't created yet so a participant profile state does not have a lead provider OR manual intervention where there is no participant profile state at all. The participant profile state is used in validating if a participant had been withdrawn at the time the declaration date is set. We use the participant profile state and not induction records as induction records are sometimes hard to trust and more to query.

- Ticket: [Issue#1751](https://github.com/DFE-Digital/register-ects-project-board/issues/1751)

- Amend the validation in the record declaration service
- Validation to checking induction records in case the latest participant profile state is withdrawn however the induction record at the time of the declaration state is active.

### DIFF
--- a/docs/source/release-notes.html.md
+++ b/docs/source/release-notes.html.md
@@ -7,7 +7,7 @@ weight: 8
 
 If you have any questions or comments about these notes, please contact DfE via Slack or email.
 
-## 20 June 2025
+## 24 June 2025
 
 [#bug-fix]
 


spec/services/record_declaration_spec.rb
```diff
@@ -76,6 +76,40 @@
 
       it { is_expected.to be_valid }
     end
+
+    context "when `relevant_induction_record` is active" do
```
cwrw, line 80: I wonder if we can make the tests closer to what is actually happening, as in participant profile states without a cpd lead provider for active, but actual withdrawn states etc.?

cwrw:
Hey sorry @leandroalemao can you also add the release note please

PR #5862, opened 2025-06-19
Fix school led materials issues in UI

tonyheadford:
### Context

Spotted a couple of issues where we are showing "Materials supplier" and referring to School-led training wrongly in the UI

Using the feature flag method, removed the view of "Materials supplier" from schools doing CIP/School-led and also from the view of school cohorts an admin user sees.

### Guidance to review

| Pre-2025 | 2025+ |
| ---- | ---- |
| image | image |
| image | image |
| image | image |


leandroalemao (approved):
Think we can assume the `programme_type_changes_2025` feature flag won't be toggled off anymore? can we then assume it's true and remove it? maybe one task for the future..

PR #5873, opened 2025-06-26
[1754] Tweak TrainingRecordState service class

leandroalemao:
### Context

- Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1754

Mentor left school but is being displayed as `Training or eligible for training` on the old delivery partner dashboard.
The logic determining the displayed status appears to not considering the induction status (in this case, "leaving"). As a result, it mistakenly attributes the new school's mentee record to the previous school and returns `:active_mentoring` as the participant's status. We need to incorporate induction status more accurately.

When a `InductionRecord` has induction status = `leaving` for a delivery partner:
- Take in consideration in the logic when a mentor left a school -> `mentoring?` method will return false instead.
- Then return `:left` on `mentoring_state` when mentor is not mentoring and has been transferred.

### Guidance to review

- Snapshot db.
- Same participant can be seen in the old and new delivery partner dashboard.

Logged in as old delivery partner:
image

Logged in as new delivery partner:
image

### DIFF
--- a/app/services/training_record_state.rb
+++ b/app/services/training_record_state.rb
@@ -44,6 +44,7 @@
 # mentoring_state:
 #   active_mentoring
 #   active_mentoring_ero
+#   left
 #   not_a_mentor
 #   not_yet_mentoring
 #   not_yet_mentoring_ero
@@ -299,6 +300,9 @@ class TrainingRecordState
     in { mentoring: false, previous_participation_reason: true }
       :not_yet_mentoring_ero
 
+    in { mentoring: false, transferred: true }
+      :left
+
     else
       :not_yet_mentoring
     end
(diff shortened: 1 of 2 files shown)


app/services/training_record_state.rb
```diff
@@ -299,6 +299,9 @@ def mentoring_state
     in { mentoring: false, previous_participation_reason: true }
       :not_yet_mentoring_ero
 
+    in { mentoring: false, transferred: true }
```
cwrw, line 302: I wonder if transferred should be it's own check rather than rolling it into mentoring? and it can come before the mentor check? as including in mentoring seems like we're combining two concepts? let's chat through this and then walk through it with Shahad :)

cwrw (commented):
can we add leaving to the comments leandro otherwise all good

PR #5893, opened 2025-07-10
Remove old participant registration routes and helpers

tonyheadford:
### Context

Way back we used to request that participants (ECT and mentors) accessed the service to register their personal details.

There were also some guidance referring to "Get a TRN" that seems to be the highest ranked Google item when searching this term instead of the correct Gov.uk guidance.

This PR removes these.

Remove the routes and controllers for the `/participants` registration area.

### Guidance to review

You should no longer be able to sign in, swap role or access the area `/participants/start-registration` or `/participants/get-a-trn`


gen-sole:
Can a dev please review

app/forms/choose_role_form.rb
```diff
@@ -63,33 +60,18 @@ def has_no_role
 private
 
   def rejected_roles
-    %w[induction_coordinator early_career_teacher mentor teacher].freeze
+    %w[early_career_teacher mentor teacher].freeze
```
ltello, line 63: Weird `induction_coordinator` was rejected

ltello (approved):
LGTM

app/forms/choose_role_form.rb
tonyheadford: It was added in again in the subsequent code (if the user was a SIT) as this was trying to make sure only one "school" role was selected, looks weird now but it was doing the right thing

PR #5960, opened 2025-08-27
[2064] Fix declarations api v3 query bug

leandroalemao:
### Context

- Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2064

A LP that a ppt has transferred to or trained with should be able to see historical decs made by another lead provider. This is so they can see how much engagement has taken place. 

We have identified a bug where participants with leaving and withdrawn inductions result in providers missing out declarations. Where IR are active these can be found but there's a bug in the service that does not permit this.

- Fix the bug described above.
- Add release note on the bug fix

### Guidance to review

Snapshot db api calls before/after using declaration id examples added to the ticket.


mooktakim (approved):

ethax-ross (approved):
Looks good thanks 👍

cwrw (approved):

PR #5984, opened 2025-09-14
Fix rollover to 2025

cwrw:
## Context

The cohort rolls over in the 1st of September in the specs. This means `Cohort.current` moves to 2025. However specs seem to fail especially around frozen cohorts, and hardcoded cohorts.

## Changes proposed in this pull request
- Revert DFE-Digital/early-careers-framework#5966
- Hardcode frozen payment specs to cohort 2024 as that was the current previous cohort, and we aren't moving participants to 2025 
- Add new cohort to specs where we need to for cohort creation, as the pre-generated spec cohorts are up to 2026
- Set mid cohort flags on specs that require it
- Excluded cohort is no longer the latest cohort, include all later excluded cohorts in specs as the excluded one was 2024
- Hard code cohort to avoid cohort previous being similar to 2024 cohort

## Guidance to review
commit by commit

### DIFF
--- a/spec/factories/cohorts.rb
+++ b/spec/factories/cohorts.rb
@@ -8,7 +8,7 @@ FactoryBot.define do
   factory :cohort do
     start_year { Faker::Number.unique.between(from: 2022, to: 2100) }
     registration_start_date { Date.new(start_year.to_i, 6, 5) }
-    academic_year_start_date { Date.new(start_year.to_i, 10, 1) }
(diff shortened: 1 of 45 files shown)


ethax-ross (commented):
Looks great, just one question

leandroalemao (approved):
All good 👍

mooktakim (approved):

PR #6000, opened 2025-09-24
Set `voided_by_user` as event actor on voided declarations

ethax-ross:
### Context

Currently, declarations that have been voided by a user in the finance dashboard still show up as voided by a lead provider in the audit trail. We want to show the user that voided the declaration here instead.

- Set `voided_by_user` as event actor on voided declarations

Update the `HistoryBuilder` to set the `voided_by_user` as the `ParticipantEvent#user` when building the audit trail.

### Guidance to review

Find a voidable declaration:

```
ParticipantDeclaration.all.find(&:voidable?).id
ParticipantDeclaration.all.find(&:voidable?).email
```

Search for it in the finance dashboard and void it.

Go to the user's audit log in the admin dashboard.

| Before | After |
| -- | -- |
| Screenshot 2025-09-24 at 11 08 35 | Screenshot 2025-09-24 at 11 08 24 |

### DIFF
--- a/app/services/participants/history_builder.rb
+++ b/app/services/participants/history_builder.rb
@@ -125,7 +125,7 @@ private
       declaration.declaration_states.each do |declaration_state|
         type = "#{declaration.declaration_type.capitalize}Declaration"
 
-        actor = declaration.voided_by_user if declaration.voided? || declaration.awaiting_clawback?
+        actor = declaration.voided_by_user if declaration_state.voided? || declaration_state.awaiting_clawback?
         actor ||= declaration.cpd_lead_provider.name
 
         value = declaration_state.state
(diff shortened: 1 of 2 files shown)


app/services/participants/history_builder.rb
```diff
@@ -124,7 +124,10 @@ def record_participant_declaration_events(declarations)
 
       declaration.declaration_states.each do |declaration_state|
         type = "#{declaration.declaration_type.capitalize}Declaration"
-        actor = declaration.cpd_lead_provider.name
+
+        actor = declaration.voided_by_user if declaration.voided?
```
cwrw, line 128: should we also include awaiting clawback/clawed back?

ethax-ross: Ah yes, so looking at the services we set the user when it moves to `awaiting_clawback` as well as `voided`, but I don't think it applies to `clawedback` as that happens via a job? I'll update it for `awaiting_clawback, let me know if that makes sense to leave `clawedback` as-is

cwrw (approved):

leandroalemao (approved):

PR #6016, opened 2025-10-03
Remove CSV partnership upload code

tonyheadford:
### Context

We want to prevent the use of the CSV upload functionality for Lead Providers to report partnerships with schools as it should no longer be used in favour of the API and it does not have all of the validations present in the API.

Remove routes, links to, code and specs related to CSV reporting of partnerships

### Guidance to review

Sign in as a provider. You should not be able to see any "Confirm your schools ..." links on the dashboard or see the "Confirm more schools" button on the "Your schools" page.


spec/models/partnership_csv_upload_spec.rb
```diff
@@ -1,166 +0,0 @@
-# frozen_string_literal: true
-
-require "rails_helper"
```
cwrw, line 3: I think since we are keeping the model this can stay?

spec/services/create_partnership_csv_upload_spec.rb
```diff
@@ -1,93 +0,0 @@
-# frozen_string_literal: true
-
-require "rails_helper"
-
-RSpec.describe CreatePartnershipCsvUpload do
```
cwrw, line 5: can we remove the related service to this spec?

cwrw:
looks good, some minor comments, I think worth getting someone from product to test it (maybe Nathan) before we merge (after confirming with providers :))

tonyheadford:
> looks good, some minor comments, I think worth getting someone from product to test it (maybe Nathan) before we merge (after confirming with providers :))

Thanks, yes Matt has already checked the review app

cwrw (approved):

PR #6157, opened 2026-04-14
Update maintenance banner for April service migration window

leoapost:
### Context

- Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3676

- update the maintenance banner window

### Guidance to review
  - enable the `maintenance_banner` feature flag in the target environment
  - verify the banner shows: `This service will be unavailable from 12pm on 27 April to 9am on 28 April.`

### DIFF
--- a/app/components/banners/maintenance_component.rb
+++ b/app/components/banners/maintenance_component.rb
@@ -4,7 +4,7 @@ module Banners
   class MaintenanceComponent < ApplicationComponent
     HOUR_FORMAT = "%-l%P"
     DAY_FORMAT = "%-d %B"
-    MAINTENANCE_WINDOW = Time.zone.local(2024, 11, 27, 19)..Time.zone.local(2024, 11, 27, 22)
+    MAINTENANCE_WINDOW = Time.zone.local(2026, 4, 27, 12)..Time.zone.local(2026, 4, 28, 9)
 
     def render?
       FeatureFlag.active?(:maintenance_banner) && maintenance_window_ends_in_future?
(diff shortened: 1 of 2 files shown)


avinhurry (commented):
This may be an hour out in production. I couldn't find any app timezone config so rails will use UTC. Since these dates are in BST, we'd be an hour late.

https://guides.rubyonrails.org/configuring.html

> active_support.initialize_time_zone: Sets the default time zone for the application based on the config.time_zone setting, which defaults to "UTC".

leoapost:
@avinhurry if it helps, the banner will be activated today by us. It's also activated on the review app if you would like to see how it looks.

peteryates:
Yeah in production if we get the current time it's in UTC 

```
Time.zone.now
=> Tue, 14 Apr 2026 13:59:56.779670643 UTC +00:00
```

But we're now in BST (so locally I get this):

```
Time.zone.now
=> 2026-04-14 15:02:44.615078442 BST +01:00
```

As it's only related to when the banner goes up, one which will be shown for a fortnight, I don't think it's really that important. Changing the TZ in ECF1 might impact other stuff we don't know about, so no really keen to do it now.

So long as we're happy if it appears an hour later, let's go with it.

leoapost:
That's all correct, however the banner will be activated manually by us today rather than scheduled, so I think the timezone shouldn't matter in this case as long as the message displays  the correct times. It's already live on the review app if you would like to take a look.

avinhurry (approved):

